Issue #48, opened 2013-10-02
Absolute values in objective/constraints


joehuchette:
Essentially, this would be some way to add absolute value terms to the model without forcing the user to deal with the proliferation of auxiliary variables/constraints themselves. More generally, this approach could be applied to other things that can be modeled via LP but require some kind of problem transformation (e.g. convex piecewise linear), but absolute value seems the most immediate (and useful, for my research).

iamed2:
Can we bring this back since that branch never got merged?

mlubin:
@iamed2, what about using @IainNZ example definition of ``abs`` above?

omus:
Gets quite ugly when you are using arrays and doesn't work when you are using `abs` with different variables:

```julia
using JuMP
import JuMP: GenericAffExpr

function abs{V<:GenericAffExpr}(v::Array{V})
	m = first(first(v).vars).m
	@defVar(m, aux[1:length(v)] >= 0)
	@addConstraint(m, aux .>= v)
	@addConstraint(m, aux .>= -v)
	return aux
end

m = Model()
@defVar(m, a[1:10] >= 0)
@defVar(m, b[1:10] <= 0)
@defVar(m, c[1:10] >= 10)
@defVar(m, d[1:10] <= -10)
```

```julia
julia> sum(abs(a + b))
aux[1] + aux[2] + aux[3] + aux[4] + aux[5] + aux[6] + aux[7] + aux[8] + aux[9] + aux[10]

julia> sum(abs(c + d))
aux[1] + aux[2] + aux[3] + aux[4] + aux[5] + aux[6] + aux[7] + aux[8] + aux[9] + aux[10]
```

mlubin:
What doesn't work? Don't worry about the names of the variables overlapping, they're distinct variables.

omus:
According do the documentation @defVar will [redefine variables without warning](http://jump.readthedocs.org/en/latest/refvariable.html). `aux` isn't being overwritten by the second `abs` call?

mlubin:
``@defVar(m,x)`` is equivalent to something like ``x = Variable(m)``. It redefines *Julia* variables in the local scope, not JuMP variables.

omus:
Interesting. Maybe I'm missing something but why isn't the following model respecting the constraint given? Both `a` and `b` both contain all zeros.

```julia
using JuMP
import JuMP: GenericAffExpr

function abs{V<:GenericAffExpr}(v::Array{V})
[...]

PR #156, opened 2014-04-03
WIP: SDP and Block structure

joehuchette:
This pull request adds SDP functionality to JuMP. Primal, dual, and matrix constraints can be used in modeling. Currently only Mosek support is available, although that will change soon. Left remaining before a merge:
- [ ] Extensive tests on types, operators, etc.
- [x] Construct useful example suite
- [x] Documentation
- [x] ~~Macros for ``setObjective``, ``addConstraint``~~
- [x] Absolute values (#48)
- [x] Handle MatrixExpr/NormExpr in setObjective
- [x] matrix operations on MatrixVar/MatrixExpr's

I'm still not quite sure what functionality will be most useful in practice (for instance, converting Lyapanov equations seems like it would be quite handy), so this should probably be considered an experimental feature for a release cycle or two, at the very least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ notifications:
     email: false
 env:
     matrix: 
-        - PKGADD="Cbc;Clp" JULIAVERSION="juliareleases" 
+        - PKGADD="Cbc;Clp" JULIAVERSION="julianightlies" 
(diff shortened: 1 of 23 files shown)


IainNZ:
:star2: *sdp branch :one: gone but never forgotten* :star2:

PR #275, opened 2014-09-26
remove 'presolve' hook

mlubin:
@joehuchette, if this isn't really needed, I'd prefer to remove this functionality in favor of `buildInternalModel(); ...; solve()`. How badly does this break things?
Ref discussion in 28c8a5d741a9a5f89ced780f20d96245be009904.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -25,7 +25,7 @@ export
     getNumVars, getNumConstraints, getObjectiveValue, getObjective,
     getObjectiveSense, setObjectiveSense, writeLP, writeMPS, setObjective,
     addConstraint, addSOS1, addSOS2, solve,
-    getInternalModel, setPresolve, buildInternalModel,
+    getInternalModel, buildInternalModel,
     # Variable
     setName, getName, setLower, setUpper, getLower, getUpper,
     getValue, setValue, getDual,
@@ -77,8 +77,6 @@ type Model
     # JuMPDict list
     dictList::Vector
 
-    # presolve callback function
-    presolve
     # storage vector for merging duplicate terms
     indexedVector::IndexedVector{Float64}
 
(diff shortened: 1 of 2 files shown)


joehuchette:
@juan-pablo-vielma pointed out that this was being used in CPLEX.jl to implement branch/incumbent callbacks. Any opposition to restoring?

mlubin:
Could you use the ``buildInternalModel(); ...; solve()`` pattern instead?
You'll need to wrap ``solve`` with a new name, but that seems to be the standard approach currently for extensions.

joehuchette:
It seems kludgy to me to force the user to use something like ``solve_with_extra_callbacks(m::JuMP.Model)`` since this isn't really an extension, more like an extra feature that lives in one of the solver packages.

In other words, something like
```
solve(m::StochasticModel)
```
is kosher, but since we can't add a new method to dispatch on ``m::Model``, you lose the generic-ness in user code.

mlubin:
Any code using incumbent or branching callbacks is already not generic. What about
```
solvecpx(m, branchcb = callback1, incumbentcb = callback2)
```
instead of
```
setBranchCallback(m, callback1)
setIncumbentCallback(m, callback2)
solve(m)
```
If we're planning on documenting the CPLEX callbacks and making it official that's one thing, but as it is now this would get things working without adding extra kludge to JuMP itself.
I do think we could use a way to plug into ``solve()`` for extensions, but a presolve hook isn't really sufficiently general.

joehuchette:
I mean it ideally would _look_ generic---the fact that CPLEX is the only solver to support it, and where exactly the code lives should be immaterial to the end user. What would a more general ``solve`` hook look like? I can reshape the code if you have something in mind, but as it is right now this serves a real purpose, seems reasonably general and useful, and is a pretty tiny addition to the codebase.

mlubin:
Something like ``setSolveHook`` that gets called instead of the default solve if it's set, with a special keyword like ``solve(m, ignore_solve_hook=true)`` so that you could call JuMP's solve from inside of the solve hook.
Would this work for JuMPeR, @IainNZ?

joehuchette:
Sounds reasonable to me.

IainNZ:
Could we have a setHook with a solve or print option? That'd be the two awkward things with "RobustModel" resolved.

PR #286, opened 2014-10-16
WIP: documentation for extensions

mlubin:
Let the extensions begin! The only question is if we're sure that we want to document the internals at this point.

### DIFF
--- a/doc/extensions.rst
+++ b/doc/extensions.rst
@@ -1,0 +1,80 @@
+.. _extensions:
+
+--------------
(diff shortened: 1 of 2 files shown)


joehuchette:
Reminder to add docs for solve/print hooks here

joehuchette:
We should probably finish up sometime soon for the paper.

mlubin:
This won't be merged, might be a useful example for future documentation

Issue #161, opened 2014-04-16
Conditions in defVar, addConstraint indexing


joehuchette:
Contrived example:
```
@defVar(m, x[i=1:10,j=1:10;i+j==8])
```
Sudoku example (set initial solutions):
```
@addConstraint(m, inits[row=1:9,col=1:9,v=initgrid[i,j];v!=0], x[row, col, v] == 1)
```

mlubin:
Is this feasible now given the parser changes in 0.4?

joehuchette:
I think so?
```jl
julia> dump(:(x[i=1:3;isodd(i)]))
Expr
  head: Symbol typed_vcat
  args: Array(Any,(3,))
    1: Symbol x
    2: Expr
      head: Symbol =
      args: Array(Any,(2,))
        1: Symbol i
        2: Expr
          head: Symbol :
          args: Array(Any,(2,))
          typ: Any
      typ: Any
    3: Expr
      head: Symbol call
      args: Array(Any,(2,))
        1: Symbol isodd
        2: Symbol i
      typ: Any
  typ: Any

julia> dump(:(x[i=1:3,j=1:3;isodd(i)]))
Expr
  head: Symbol typed_vcat
  args: Array(Any,(4,))
    1: Symbol x
    2: Expr
      head: Symbol parameters
      args: Array(Any,(1,))
        1: Expr
          head: Symbol call
          args: Array(Any,(2,))
          typ: Any
      typ: Any
    3: Expr
      head: Symbol =
      args: Array(Any,(2,))
        1: Symbol i
        2: Expr
          head: Symbol :
          args: Array(Any,(2,))
          typ: Any
      typ: Any
    4: Expr
      head: Symbol =
      args: Array(Any,(2,))
        1: Symbol j
        2: Expr
          head: Symbol :
          args: Array(Any,(2,))
          typ: Any
      typ: Any
  typ: Any
```

mlubin:
Closed by #277

yeesian:
:+1:

PR #312, opened 2014-11-19
deprecate @defConstrRef

mlubin:
Now that the three argument versions of ``@addConstraint`` and ``@addNLConstraint`` work in general, I don't see much of a point of keeping this around. Users can always create their own collections of constraint references if needed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ JuMP release notes
 Unreleased
 ----------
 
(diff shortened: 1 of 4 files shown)


mlubin:
Tagging this for 0.10 assuming we can resolve #161

IainNZ:
Needs docs of some sort to make it clearer what to do in different situations I think

mlubin:
The syntax from #161 only works on 0.4, so definitely can't do this until we drop 0.3.

mlubin:
Guess this isn't going anywhere anytime soon

Issue #341, opened 2015-01-04
method for changing variable type


mlubin:
This has come up a few times, we should export a method for this.

joehuchette:
``setVarType``?

PR #343, opened 2015-01-04
[RFC] add setVarType

joehuchette:
ref #341. Will add docs + tests + NEWS if it looks alright.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,12 @@
 JuMP release notes
 ==================
 
+Unversioned
+-----------
+  
+  * Added the ``setCategory`` function to change a variables category (e.g. continuous or binary) 
+  after construction, and ``getCategory`` to retrieve the variable category.
+
 Version 0.7.1 (January 2, 2015)
 -------------------------------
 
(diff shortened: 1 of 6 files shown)


mlubin:
Type is a heavily overloaded word, what about ``setCategory``? No need to put ``Var`` in the name since the input is a Variable object.

joehuchette:
Sounds good to me. Added everything else. Printing is still a bit broken:
```
julia> @defVar(m, x[1:3]);

julia> setCategory(x[2], :Int);

julia> print(m)
Min 0
Subject to
 x[i] free for all i in {1,2,3}
```

IainNZ:
Printing thing is hella awkward...

Can this be made to work on `JuMPContainers` too? So you can set all the categories? I think this works on bounds already, or did at one point.

joehuchette:
I think I got it figured out; pretty much you just "forget" that variables are in a JuMPContainer if they're not all the same category.

joehuchette:
Win32 AppVeyor timed out, gonna merge anyway

Issue #349, opened 2015-01-08
Error messages when trying to use nonlinearities in the linear/quad macros


tkelman:
This is on JuMP master, Julia 0.3.5-pre+123. (totally artificial example)
```
julia> m = Model();

julia> @defVar(m, 0 <= x1); # I'm sure this one is a well-known annoyance
ERROR: assertion failed: isa(var,Expr)

julia> @defVar(m, x1 >= 0);

julia> @defVar(m, x2 >= 0);

julia> @setObjective(m, Max, x1/x2)
ERROR: Cannot divide by variable
 in / at /home/tkelman/.julia/v0.3/JuMP/src/operators.jl:21

julia> @addConstraint(m, log(x1) - log(x2) >= 1)
ERROR: `log` has no method matching log(::Variable)
```

If there's any reasonable low-performance-impact way of detecting this kind of mistake and suggesting the user should be using `@setNLObjective` and `@addNLConstraint`, would be a bit more user-friendly.

I'd argue the linear/quadratic versions are more specialized and should maybe be aliased or renamed to `@setLinObjective` `@addLinConstraint` etc. I recognize the development history and implementation differences behind the separate names as they are, but to newcomers they don't necessarily make a lot of sense. I've seen people coming from Yalmip frequently forgetting to add the `NL` names.

mlubin:
We could provide some basic overloads for ``log(::Variable)``, ``log(::AffExpr)``, ``log(::QuadExpr)``, etc... and print a useful error message. The name ``addLinConstraint`` seems off since it accepts linear and quadratic. It's in the long-term plans to accept both linear and nonlinear expressions in ``@addConstraint``, but that will require a complete rewrite of the macros with heavy abuse of staged functions.

tkelman:
Could be aliased to `addQuadConstraint`... guess I can be patient and see how much better things get in 0.4, at which point error-message overloads wouldn't be necessary?

mlubin:
What will change in 0.4?
We should just implement these now to address the usability issue.

joehuchette:
Let's implement the fallbacks now and start thinking about how best to consolidate everything under ``@addConstraint`` now (though that necessarily won't get released until after v0.4).

tkelman:
> What will change in 0.4?

Anything that relies on staged functions?

mlubin:
This is mostly resolved by #351, but let's see if we can also improve the error message for
```
@defVar(m, 0 <= x1)
```

[...]

PR #351, opened 2015-01-08
add fallback methods for nonlinear functions on built-in types

joehuchette:
ref #349 

How does the error message look, @tkelman?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,3 +3,4 @@ MathProgBase 0.3.8 0.4.0-
 ReverseDiffSparse
 ArrayViews
 Compat
+Calculus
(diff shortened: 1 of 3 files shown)


tkelman:
Maybe replace "constraint" by "problem" in

> Are you trying to build a nonlinear constraint?

since this error could also occur in `@setObjective`, right?

mlubin:
Looks good. Also we should give a similar message for division.

mlubin:
And exponents?

joehuchette:
AppVeyor 32bit timed out again...

tkelman:
> AppVeyor 32bit timed out again...

There's some 0.4-only regression we need to figure out in how long it takes WinRPM to download the repodata xml files. It used to be 64-bit-only but then something changed to make it 32-bit-only. I think if you test AppVeyor on releases it shouldn't take as long, depending whether you want to check the combination of Windows and 0.4-only features regularly.

PR #360, opened 2015-01-17
[RFC]: add solvehook/printhook functionality

joehuchette:
Ref #275, is this what you had in mind @mlubin?

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -28,7 +28,7 @@ export
     getObjectiveSense, setObjectiveSense, setSolver,
     writeLP, writeMPS, setObjective,
     addConstraint, addSOS1, addSOS2, solve,
-    getInternalModel, buildInternalModel,
+    getInternalModel, buildInternalModel, setSolveHook,
     # Variable
     setName, getName, setLower, setUpper, getLower, getUpper,
     getValue, setValue, getDual, setCategory, getCategory, 
@@ -78,6 +78,10 @@ type Model
     cutcallback
     heurcallback
 
+    # hook into a solve call...function of the form f(m::Model; kwargs...), 
+    # where kwargs get passed along to subsequent solve calls
+    solvehook
+
     # List of JuMPContainer{Variables} associated with model
     dictList::Vector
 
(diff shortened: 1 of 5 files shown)


IainNZ:
I'd really like a printing hook too, so can we have either
```julia
setSolveHook(
setPrintHook(
```
or
```julia
setHook(:solve
setHook(:print
```
(I guess with value dispatching in Julia 0.4 they'd be almost identical!)

joehuchette:
What would this be used for? Any calls to printing of JuMP models/variables/constraints, etc?

IainNZ:
A JuMP extension (like JuMPeR) can't currently use `print` because print doesn't do anything with extensions. So I have `printRobust` and `solveRobust`. One option with printing would be to have `print(::Model)` try to call a print on each extensions object in the extension dict, OR we can do the hook thing for printing too. I'm going to use this solve hook thing to get rid of `solveRobust`, iff the solve hook can return e.g. a false that'll break out of JuMP's solve without it doing anything.

joehuchette:
I think a hook for ``print(::Model)`` makes more sense than trying to deal with everything in the extension dict directly. I _think_ the solve hook functionality will work with JuMPeR if you just add ``ignore_solve_hook=true`` to each of the ``solve`` calls in ``solveRobust``.

mlubin:
I don't mind two separate hooks for solve and print. We'll eventually make this nice but let's just experiment a bit first to see what works.

joehuchette:
Added printing hook as well.

IainNZ:
Yeah I think this is cool, and really smooths nicely over one of the more awkward things with extensions (not being able to dispatch on the model type).

waTeim:
say, how general is "printing" do you have some ideas on visualizing?

joehuchette:
Not really, but we're definitely interested in any suggestions!

waTeim:
Well, something I've been playing around with recently.  Idea is Julia + atom-shell + d3.  If something could export something (JSON in a IOBuffer maybe), then it could be read and then yielded by a generator and then rendered fairly arbitrarily.

mlubin:
LGTM. We're sort of emulating virtual methods but that's okay.

IainNZ:
Will this be 0.7.4 or 0.8?

joehuchette:
I'd prefer 0.7.4 so I can start using it. I didn't add docs or NEWS, but I'm thinking it would fit best in the extension docs PR #286

joehuchette:
Also thinking that maybe I shouldn't have exported the new functions

PR #363, opened 2015-01-21
Some little coverage improvements

IainNZ:
Seems like a decent number of the zeros a spurious, which is a shame.


IainNZ:
0.8% improvement, wooo

PR #368, opened 2015-01-26
use MPB methods for accessing number of variables and constraints.

mlubin:
There's a bit of ambiguity when a model has both SOS and nonlinear constraints, but I think we can live with this for now.
CC @tkelman

### DIFF
--- a/doc/refmodel.rst
+++ b/doc/refmodel.rst
@@ -36,8 +36,10 @@ Methods
 
 **General**
 
-* ``getNumVars(m::Model)`` - returns the number of variables associated with the ``Model m``.
-* ``getNumConstraints(m::Model)`` - returns the number of constraints associated with the ``Model m``.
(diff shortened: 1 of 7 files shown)


doc/refmodel.rst
```diff
@@ -36,8 +36,10 @@ Methods
 
 **General**
 
-* ``getNumVars(m::Model)`` - returns the number of variables associated with the ``Model m``.
-* ``getNumConstraints(m::Model)`` - returns the number of constraints associated with the ``Model m``.
+* ``MathProgBase.numvar(m::Model)`` - returns the number of variables associated with the ``Model m``.
```
IainNZ: We're exporting these names, so why does it need to be fully qualified?

mlubin: Good point, actually the re-export doesn't work because ``numvar`` are never imported into the JuMP namespace. Should we just not export them?

PR #370, opened 2015-01-28
Fixed variables were broken for nonlinear solvers that did not support MINLP

joehuchette:


test/nonlinear.jl
```diff
@@ -60,6 +60,20 @@ context("With solver $(typeof(nlp_solver))") do
             [1.000000, 4.742999, 3.821150, 1.379408], 1e-5)
 end; end; end
 
+facts("[nonlinear] Accepting fixed variables") do
+for nlp_solver in nlp_solvers
```
mlubin: This should be ``convex_nlp_solvers`` so that the test also runs with Mosek.

mlubin:
LGTM. Could you squash since there's a commit in there with broken tests?

mlubin:
Ok, feel free to merge once CI passes.

PR #377, opened 2015-01-30
[WIP] add OsilSolver to tests

joehuchette:
@tkelman my testlog is in [this gist](https://gist.github.com/joehuchette/b1b4b4bd201ae32d531b).

### DIFF
--- a/test/solvers.jl
+++ b/test/solvers.jl
@@ -18,6 +18,7 @@ ipt = isdir(Pkg.dir("Ipopt"))
 nlo = isdir(Pkg.dir("NLopt"))
 kni = isdir(Pkg.dir("KNITRO"))
 eco = isdir(Pkg.dir("ECOS"))
+osl = isdir(Pkg.dir("OptimizationServices"))
 # Load them
 if grb; import Gurobi; end
 if cpx; import CPLEX; end
@@ -28,6 +29,7 @@ if ipt; import Ipopt; end
 if nlo; import NLopt; end
 if kni; import KNITRO; end
 if eco; import ECOS; end
+if osl; import OptimizationServices; end
 # Create solver lists
 # LP solvers
 lp_solvers = Any[]
@@ -38,6 +40,7 @@ cbc && push!(lp_solvers, Clp.ClpSolver())
 glp && push!(lp_solvers, GLPKMathProgInterface.GLPKSolverLP())
 ipt && push!(lp_solvers, Ipopt.IpoptSolver(print_level=0))
 eco && push!(lp_solvers, ECOS.ECOSSolver(verbose=false))
+osl && push!(lp_solvers, OptimizationServices.OsilSolver())
 # MILP solvers
 ip_solvers = Any[]
 grb && push!(ip_solvers, Gurobi.GurobiSolver(OutputFlag=0))
@@ -45,6 +48,7 @@ cpx && push!(ip_solvers, CPLEX.CplexSolver(CPX_PARAM_SCRIND=0))
 mos && push!(ip_solvers, Mosek.MosekSolver(LOG=0))
 cbc && push!(ip_solvers, Cbc.CbcSolver(logLevel=0))
 glp && push!(ip_solvers, GLPKMathProgInterface.GLPKSolverMIP())
+osl && push!(ip_solvers, OptimizationServices.OsilSolver())
 # Support semi-continuous, semi-integer solvers
 semi_solvers = Any[]
 grb && push!(semi_solvers, Gurobi.GurobiSolver(OutputFlag=0))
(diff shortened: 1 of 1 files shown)


mlubin:
How does OS choose the default solver?

tkelman:
It looks at the variable types and whether there are any nonlinearities. You can build it with any of the Osi linear solvers, though I'm not doing anything special about the commercial ones or downloading GLPK right now.

For MINLP I believe it always defaults to Bonmin, so you'd have to do `OsilSolver(solver="couenne")` to test Couenne.

Looks like a little bit of code that I had for correcting Bonmin and Couenne returning "other" status for a user-limit condition doesn't work right on linear problems, I'll look into it.

tkelman:
https://github.com/tkelman/OptimizationServices.jl/commit/8b9246c54b36325ea9bdd4578eb0cdbb9ae45941 should help with the `startswith` method error, now I'm getting a bunch more output to look through.

tkelman:
Most of the output was telling me that I haven't implemented duals (constraints or bounds), quadratic objectives or constraints yet. Though HS071 doesn't solve correctly, the lexer doesn't like the osil file for some reason.

tkelman:
Fixed a few more bugs. Now it's saying the MINLP test problem is infeasible... here's the osil it generates, looks like a correct translation? https://gist.github.com/3919bc2f421c207fcc48

tkelman:
Ah, interesting. Couenne solves it, Bonmin doesn't. I think I'll need to re-do my implementation of solver options since the way I have them done now isn't working.

tkelman:
I implemented duals. The "[probmod] Test bound modification on binaries" test revealed what looks like a bug in OS with Cbc.

mlubin:
Good to see that our tests are doing a good job of shaking out the bugs :)

IainNZ:
Indeed! Makes me want to write more tests for things that are, in theory, covered. We've talked about fuzzing things, I think we fuzz expression generation now right?

tkelman:
And moar test problems once I write the `osil2jump` function. Though we probably need to be careful about the licenses on existing test sets, which is an endlessly confusing issue.

tkelman:
This'll need updating for the rename to CoinOptServices

joehuchette:
Updated, [new test log here](https://gist.github.com/joehuchette/2bb2e5b932f62599beca). I'm seeing a log of things like
```
Results written to file /Users/huchette/.julia/v0.4/CoinOptServices/.osil/results.osrl
/Users/huchette/.julia/v0.4/CoinOptServices/.osil/results.osrl:2: namespace warning : xmlns: URI os.optimizationservices.org is not absolute
<osrl xmlns="os.optimizationservices.org"   xmlns:xsi="http://www.w3.org/2001/XM
                                         ^
```
and
```
<?xml version="1.0" encoding="UTF-8"?><?xml-stylesheet type="text/xsl" href="OSrL.xslt"?>
<osrl xmlns="os.optimizationservices.org"   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="os.optimizationservices.org http://www.optimizationservices.org/schemas/2.0/OSrL.xsd" >
<general>
<generalStatus type="error">
</generalStatus>
<message>PARSER ERROR:  Input is either not valid or well formed: </obj> element missing
Here are 20 characters currently being pointed to in the input string: coef idx="0">1.0</co
See line number: 12
</message>
</general>
</osrl>
```

tkelman:
I'll try to figure out how to suppress the "Results written to file" message and the xml namespace warning (that part is coming from libxml2 by way of LightXML.jl).

I think the "element missing" should be fixed on CoinOptServices master (I'm working on quadratics before making another tag), but to be sure, can you run just one of the problems that generates that message then gist `Pkg.dir("CoinOptServices", ".osil", "problem.osil")` ?

tkelman:
Quadratic objectives should now work on CoinOptServices master. The MIQP from "[qcqpmodel] Test quad objective (discrete)" is another case of Couenne solving it fine but Bonmin failing, at least with the default options OS sets when it calls Bonmin. Ipopt is exiting at a continuous node where the bounds are infeasible, which I would think Bonmin should expect to happen but apparently not with these options. And osol is picky, my initial solver options implementation isn't working yet.

I'll see how much work it is to get solver options and quadratic constraints working.

joehuchette:
It's kind of scary that Bonmin's breaking down on our relatively simple test cases.

@tkelman is there any changes I should make to this PR or is it good to merge? It's not hooked up on Travis so it won't break anything while you iron out the last few kinks on your end.

tkelman:
I think it would be fine if it weren't going through OS, since I think OS overrides the default options.

I am aiming to make that last commit revert-worthy, doesn't matter to me in the meantime whether this gets merged now or later. I'm going to push quad constraints in a couple minutes, but fixing the options might take longer.

mlubin:
> It would probably be fine if it weren't going through OS, since I think OS overrides the default options.

One reason why I'm afraid of touching options from JuMP...

tkelman:
I was going to say your MINLP test problem might just be challenging, but nah it looks like Ipopt.jl solves the continuous relaxation of it no sweat (introducing a nonnegative variable for one of the expressions inside the `log` can help but might not be necessary). There might be something more fundamentally wrong with derivative calculations in OS, since running Ipopt through osil has serious trouble on the continuous relaxation of that problem. I'll have to run the derivative checker and see what's up.

mlubin:
That's surprising. Let us know what you find.

tkelman:
I know there's been a long-standing problem with maximization objectives, but this is a min so it's something else.

PR #378, opened 2015-01-31
logic fixes

mlubin:
Relatedly, something seems fishy with fixed variables. We shouldn't be sending ``:Fixed`` status to MPB because that's not a valid variable type. It doesn't look like we filter this out anywhere?


joehuchette:
We should probably just set the variable type to ``:Cont`` when we pass it through MBP and ensure the bounds are equal.

PR #381, opened 2015-02-01
Add a few more tests for lower bounds of binaries

tkelman:
OS some bugs with the combination of binary variables and bounds. Thanks for having the tests that identified the bug. This adds the same check but for lower bounds, which OS wasn't doing anything about but it should be.


joehuchette:
Thanks!

PR #384, opened 2015-02-06
Pass on kwargs to solve hook

IainNZ:
This is pretty essential to make JuMPeR work as it is right now. Without this I'll need to have some option-setting lines before the solve, which doesn't feel good.
If this is OK, can it be v0.7.5?

### DIFF
--- a/src/solvers.jl
+++ b/src/solvers.jl
@@ -1,6 +1,6 @@
-function solve(m::Model; suppress_warnings=false, ignore_solve_hook=(m.solvehook==nothing))
+function solve(m::Model; suppress_warnings=false, ignore_solve_hook=(m.solvehook==nothing), kwargs...)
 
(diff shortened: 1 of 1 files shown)


mlubin:
LGTM.

joehuchette:
Yep, looks good. What do you need the new release for? (Not that I'm opposed, just wondering)

IainNZ:
Trying to tag a JuMPeR v0.1  :D

joehuchette:
:+1: 

BTW what's the deal with the release failure on PkgEvaluator?

IainNZ:
Wooah... only on 0.3 too. I haven't pushed the run from this morning but I manually checked and it has happened again...

IainNZ:
Figured it out - was because I had precompiled JuMP into Julia 0.3, and I don't build it every night. This was a baffling error!

IainNZ:
AppVeyor failure seems to be due to download issues.

PR #388, opened 2015-02-10
[RFC] Add info callback and refactor callback code

joehuchette:
Adds informational callback through the standard interface. Also refactors the callback handling code to allow multiple callbacks to be registered; it is the responsibility of the solver to handle multiple callbacks added to the same model. Thoughts? Maybe it's better to merge together the callbacks at the JuMP level.

cc @mlubin @IainNZ

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,13 @@
 JuMP release notes
 ==================
 
+Unversioned
+-----------
+
+  * **Breaking change**: ``set___Callback`` family deprecated in favor of ``add___Callback``.
+  * Multiple callbacks of the same type can be registered.
+  * Added support for informational callbacks via ``addInfoCallback``.
+
 Version 0.7.4 (February 4, 2015)
 --------------------------------
 
--- a/doc/callbacks.rst
+++ b/doc/callbacks.rst
@@ -280,6 +280,32 @@ including::
     cbgetstate(cb)
 
 
+Informational Callbacks
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Sometimes it can be useful to track solver progress without actually changing the algorithm by adding cuts or heuristic solutions. In these cases, informational callbacks can be added, wherein statistics can be tracked via the ``cbget`` functions discussed in the previous section. Informational callbacks are added to a JuMP model with the ``addInfoCallback(m::Model, f::Function)`` function.
+
+For a simple example, we can add a function that tracks the best bound and incumbent objective value as the solver progresses through the branch-and-bound tree::
+
+    # build model ``m`` up here
+
+    type NodeData
+        node::Int
+        obj::Float64
+        bestbound::Float64
+    end
+
+    bbdata = NodeData[]
+
+    function infocallback(cb)
+        node      = cbgetexplorednodes(cb)
+        obj       = cbgetobj(cb)
+        bestbound = cbgetbestbound(cb)
+        push!(bbdata, NodeData(node,obj,bestbound))
+    end
+    setInfoCallback(m, infocallback)
+
+
 Code Design Considerations
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -70,9 +70,10 @@ type Model
     solver::MathProgBase.AbstractMathProgSolver
     internalModelLoaded::Bool
     # callbacks
-    lazycallback
-    cutcallback
-    heurcallback
+    callbacks
+    # lazycallback
+    # cutcallback
+    # heurcallback
 
     # hook into a solve call...function of the form f(m::Model; kwargs...),
     # where kwargs get passed along to subsequent solve calls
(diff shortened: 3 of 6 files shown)


mlubin:
I don't think solvers should have to deal with registering multiple callbacks, that doesn't exist in any low-level solver interface. It's much easier to handle it inside JuMP.

IainNZ:
+1 to handle it at JuMP level

Also the `setXCallback` name is odd if it can lead to multiple callbacks: `addXCallback` seems natural.

joehuchette:
Ok, I'll handle it in JuMP. I agree new names would be more clear, but we'd have to deprecate the old ones and I'm pretty sure the vast majority of users won't ever add multiple callbacks. Not sure if it's worth it.

mlubin:
It's just a deprecation with a visible warning, not a breaking change, so I'd say go for it.

joehuchette:
Updated. The handling is slightly more complex than it could be, but I've been thinking about this approach for a while since it allows extensions to attach callbacks directly to JuMP without adding a solve hook.

mlubin:
A bit weird but I guess it works.
@IainNZ, is it easy to rerun the callback overhead benchmark you put together.
One thing that this PR breaks is the passing along the return status from callbacks back to the solver. This wasn't documented but could be used to exit the solver: https://github.com/JuliaOpt/Gurobi.jl/blob/8498bed9a35d3b66723fe09e0101fee2d089dd0b/src/GurobiSolverInterface.jl#L474. We should make sure this is still possible in some form.

IainNZ:
Here is the benchmark script

```julia
using JuMP
using Gurobi

function solve_sphere(N,R,c,tol)
    m = Model(solver=GurobiSolver(OutputFlag=0,MIPGap=0.01,Threads=1))
    @defVar(m, -R <= x[1:N] <= R, Int)
    @setObjective(m, Max, sum{c[i]*x[i],i=1:N})

    callbacks = 0
    function normcb(cb)
        callbacks += 1
        n = getValue(x)[:]
        # Calculate length of vector
        len = norm(n)
        # Only add constraint if this vector is outside
        # the sphere by more than the tolerance
        (len <= R + tol) && return
        # Want to add a plane that is tangent to the sphere
        # where this vector intersects the boundary of the
        # sphere. Our x is the normal of the plane, and
        # the vector intersects the boundary at (R/||n||)*n
        # So plane equation is:
        # <n,x> = <n,x_0> = <n, (R/||n||)*n> = R||n||
        @addLazyConstraint(cb, sum{n[i]*x[i],i=1:N} <= R*len)
    end
    setLazyCallback(m, normcb)

    tic()
    solve(m)
    t = toq()

    #@show getObjectiveValue(m)
    #@show norm(getValue(x)[:])
    #@show callbacks

    return t
end

# Warmup JIT
@time solve_sphere(2,100.0,[1.0,1.0],1e-6)

# Run benchmark
N = 20
R = 50.0
c = zeros(N)
for ind in 1:N
    c[ind] = ind%2==0 ? -ind^2 : ind^2
end

times = Float64[]
for run in 1:20
    @time push!(times, solve_sphere(N,R,c,1e-6))
end
@show minimum(times)
@show median(times)
@show mean(times)
```

IainNZ:
> since it allows extensions to attach callbacks directly to JuMP without adding a solve hook.

You mean like additional callbacks on top of whatever the user added?

joehuchette:
Yeah, like in CPLEX I can add methods for ``attach_callbacks(::JuMP.Model, ::CPLEXBranchCallback)`` and just have ``registercallbacks`` handle everything for me.

joehuchette:
This PR:
```
minimum(times) = 10.432456053
median(times) = 10.7896223025
mean(times) = 11.27911690735
```
JuMP master:
```
minimum(times) = 10.522120258
median(times) = 11.513886100499999
mean(times) = 11.71897603945
```

joehuchette:
Yeah, not sure what to do about the return types. FWIW there was some type instability in the lazy callback as implemented. Not sure if it really matters, but I'm seeing a speedup on this branch of ~0.5 seconds and that's the only reason I can come up with.

IainNZ:
With those times, I feel like we are pretty much at Gurobi C++, kind of neat

mlubin:
Well we could either check for a specific return value from each function and return on the first ``:Exit`` or add a method to MPB to signal an error from inside a callback.

joehuchette:
If we're just trying to help the callbacks fail gracefully, why not just encourage users to throw errors inside the callbacks themselves?

mlubin:
That makes more sense, we could put a try/catch around the callbacks and return ``:Exit`` or whatever from JuMP.

IainNZ:
Try-catch is only slow if it throws right?

mlubin:
Certainly less overhead than calling a function though an untyped object anyway.

joehuchette:
I'm not sure try/catch is the best approach, since then we also have to handle any errors the callback code itself produces (BoundsErrors, etc). I'd probably rather have a ``stopSolver(cb)`` function, rather than mess around with return values.

joehuchette:
Also, we should document the different callback states that can be returned more explicitly in MPB, since they affect code behavior.

joehuchette:
Going to merge, I'll open a separate issue for handling early exits in MPB.

NEWS.md
```diff
@@ -1,6 +1,13 @@
 JuMP release notes
 ==================
 
+Unversioned
+-----------
+
+  * **Breaking change**: ``set___Callback`` family deprecated in favor of ``add___Callback``.
```
mlubin: This isn't breaking, just prints warning.

PR #390, opened 2015-02-13
RFC: another try at nonlinear subexpressions, now embeddable

mlubin:
The syntax is the same as before but now can be embedded inside ``sum{}`` and ``prod{}``
```
@defNLExpr(entropy[i=1:N], -x[i]*log(x[i]))
@setNLObjective(m, Max, sum{entropy[i], i = 1:N})
```
The syntax is a bit misleading because we actually just create a single ParametricExpression object that needs index parameters like ``i`` but doesn't do anything with the index sets ``1:N``. This is probably better than inventing a new syntax, and later on we can add assertions that indices are valid. As it is now, you're free to use ``entropy[x]`` even when ``x`` isn't in ``1:N``.
CC @tkelman

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.3
 MathProgBase 0.3.8 0.4.0-
-ReverseDiffSparse 0.1.11-
+ReverseDiffSparse 0.2.1
 ArrayViews
 Compat
 Calculus
--- a/doc/nlp.rst
+++ b/doc/nlp.rst
@@ -71,12 +71,6 @@ the syntax for linear and quadratic expressions. We note some important points b
 
     @defNLExpr(myexpr[i=1:n], sin(x[i]))
     @addNLConstraint(m, myconstr[i=1:n], myexpr[i] <= 0.5)
-- Nonlinear expression objects currently cannot appear inside of ``sum{}`` or ``prod{}``. For example::
-
-    @defNLExpr(myexpr[i=1:n], sin(x[i]))
-    # This is not supported!
-    @addNLConstraint(m, myconstr, sum{myexpr[i], i=1:n} <= 0.5)
-
 
 Performance: Solution time
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -34,7 +34,8 @@ export
 # Macros and support functions
     @addConstraint, @addConstraints, @defVar,
     @defConstrRef, @setObjective, addToExpression, @defExpr,
-    @setNLObjective, @addNLConstraint, @addNLConstraints
+    @setNLObjective, @addNLConstraint, @addNLConstraints,
+    @defNLExpr
 
 include("JuMPDict.jl")
 #include("JuMPArray.jl")
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -572,3 +572,18 @@ macro addNLConstraint(m, x, extra...)
 
     return assert_validmodel(m, code)
 end
+
+macro defNLExpr(x, extra...)
+    # Two formats:
+    # - @defNLExpr(a*x <= 5)
+    # - @defNLExpr(myref[a=1:5], sin(x^a))
+    length(extra) > 1 && error("in @defNLExpr: too many arguments.")
+    # Canonicalize the arguments
+    c = length(extra) == 1 ? x        : nothing
+    x = length(extra) == 1 ? extra[1] : x
+
+    refcall, idxvars, idxsets, idxpairs = buildrefsets(c)
+    varname = isexpr(refcall,:ref) ? refcall.args[1] : refcall
+    macrocall = Expr(:macrocall, symbol("@parametricExpr"), [esc(v) for v in idxvars]..., esc(x))
+    return :($(varname) = $macrocall)
+end
--- a/test/nonlinear.jl
+++ b/test/nonlinear.jl
@@ -215,6 +215,21 @@ context("With solver $(typeof(nlp_solver))") do
     @fact getValue(x) + getValue(y) => roughly(-1/3, 1e-3)
 end; end; end
 
+facts("[nonlinear] Test maximization objective (embedded expressions)") do
+for nlp_solver in convex_nlp_solvers
+context("With solver $(typeof(nlp_solver))") do
+    m = Model(solver=nlp_solver)
+    @defVar(m, -2 <= x <= 2); setValue(x, -1.8)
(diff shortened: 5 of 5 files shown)


IainNZ:
Will try to break

mlubin:
Please do

IainNZ:
So the older issue with indices is fixed, although the Goddard rocket problem is still too hard for IPOPT it seems (at least with the chosen starting point). The code looks really cool IMO:

```julia

# Forces
# Drag(h,v) = Dc v^2 exp( -hc * (h - h0) / h0 )
@defNLExpr(drag[j=0:n], D_c*(v[j]^2)*exp(-h_c*(h[j]-h_0)/h_0))
# Grav(h)   = go * (h0 / h)^2
@defNLExpr(grav[j=0:n], g_0*(h_0/h[j])^2)

# Dynamics
for j in 1:n
    # h' = v
    @addNLConstraint(mod, h[j] - h[j-1] == Δt*v[j-1])

    # v' = (T-D(h,v))/m - g(h)
    @addNLConstraint(mod, v[j] - v[j-1] == Δt*(
                    (T[j-1] - drag[j-1])/m[j-1] - grav[j-1]))

    # m' = -T/c
    @addNLConstraint(mod, m[j] - m[j-1] == -Δt*T[j-1]/c)
end
```

IainNZ:
Also an interesting note - can pretty much copy the model from AMPL. The above I posted above is the rectangular version, I'm going to use the trapezoidal version from the original .mod and it might even solve - it was too painful to write before.

IainNZ:
D: still infeasible with IPOPT, but look at dat modeling
```julia
# Dynamics - using trapezoidal rule
for j in 1:n
    # h' = v
    @addNLConstraint(mod, h[j] == h[j-1] + 0.5*Δt*(v[j] + v[j-1]))

    # v' = (T-D(h,v))/m - g(h)
    @addNLConstraint(mod, v[j] == v[j-1] + 0.5*Δt*(
                    (T[j]   - drag[j]   - m[j]  *grav[j]  )/m[j] +
                    (T[j-1] - drag[j-1] - m[j-1]*grav[j-1])/m[j-1]))

    # m' = -T/c
    @addNLConstraint(mod, m[j] == m[j-1] - 0.5*Δt*(T[j] + T[j-1])/c)
end
```

tkelman:
what's the initial guess?

IainNZ:
Bonus @joehuchette style:
```julia
@addNLConstraints mod begin
    # h' = v
    hdash[j=1:n], h[j] == h[j-1] + 0.5*Δt*(v[j] + v[j-1])
    # v' = (T-D(h,v))/m - g(h)
    vdash[j=1:n], v[j] == v[j-1] + 0.5*Δt*(
                            (T[j]   - drag[j]   - m[j]  *grav[j]  )/m[j] +
                            (T[j-1] - drag[j-1] - m[j-1]*grav[j-1])/m[j-1])
    # m' = -T/c
    mdash[j=1:n], m[j] == m[j-1] - 0.5*Δt*(T[j] + T[j-1])/c
end
```

IainNZ:
@tkelman I've put up a gist https://gist.github.com/IainNZ/36b966eb59096b7646ba with
- my implementation
- the original AMPL mod
- the original AMPL dat

from COPS3

tkelman:
Was Ipopt able to solve the problem when called by AMPL? A .nl would be useful, if not too large.

IainNZ:
No idea, the COPS3 paper didn't test with IPOPT (I don't think it existed yet).

tkelman:
I think it existed in 2004 but was still in Fortran. http://list.coin-or.org/pipermail/ipopt/2002-May/000000.html

Interesting data point, going through osil takes about the same number of iterations and also returns infeasible, but takes over 90 seconds in function evaluations. 30 in Jacobian, 60 in Hessian. This might be due to my patch working around a bug, not sure. I should probably compile a few old versions of OSSolverService and see if things used to be faster.

mlubin:
@IainNZ, the @joehuchette style formulation is fixed, check out ReverseDiffSparse master.

joehuchette:
I don't even particularly like that syntax ¯\_(ツ)_/¯

joehuchette:
It's really great how this is shaping up, the rocket model looks fantastic now.

mlubin:
@tkelman, JuMP spends 0.3 seconds on NLP evaluations on my machine, is OS really 300x slower?

tkelman:
About 250x on this same machine. Something is probably very wrong.

IainNZ:
Heres another COPS problem using sub expressions, also doesn't solve with IPOPT (although solution is close to correct if I compare plots), also doesn't sum sub expressions. Might have interesting hessian though?

https://gist.github.com/IainNZ/84383f90c80ad0d7432b

IainNZ:
OK I made one that uses this! Although not in a very exciting way mathematically. Its a very satisfying notebook though:

https://gist.github.com/IainNZ/eef68592a7d272074e12

http://nbviewer.ipython.org/gist/IainNZ/eef68592a7d272074e12

tkelman:
OS 2.4.4 from 2012 is just as slow as the latest versions. Getting these old releases to build on Ubuntu 14.04 is surprisingly difficult...

mlubin:
@IainNZ, cool!
There's a typo:
```
@setNLObjective(mod, Max, 0.5*sum{segarea[i],i=1:n-1})
```
should be
```
@setNLObjective(mod, Max, 0.5*sum{segarea[i],i=1:nv-1})
```

IainNZ:
Why is Travis failing on this?

mlubin:
Need to tag new ReverseDiffSparse. I'm still doing some bug fixes.

mlubin:
Passing now. I don't see any harm in merging since it won't break anything that wasn't already broken by bumping ReverseDiffSparse.

IainNZ:
Indeed, seems like if there any regressions then they won't be easily found.

JackDunnNZ:
Sorry for coming in late here, missed this but thought I would add in the stats for going via .nl. Running Ipopt as an NLSolver using Iain's formulation, it detects infeasibility quickly:
    
    Number of objective function evaluations             = 48
    Number of objective gradient evaluations             = 15
    Number of equality constraint evaluations            = 48
    Number of inequality constraint evaluations          = 0
    Number of equality constraint Jacobian evaluations   = 30
    Number of inequality constraint Jacobian evaluations = 0
    Number of Lagrangian Hessian evaluations             = 27
    Total CPU secs in IPOPT (w/o function evaluations)   =      0.216
    Total CPU secs in NLP function evaluations           =      0.068

.nl file for reference here: https://gist.github.com/JackDunnNZ/c6b557002e4907e898fe

PR #397, opened 2015-02-18
don't run AD if only expression graphs are requested

mlubin:
Ref JuliaOpt/MathProgBase.jl#50
CC @tkelman @JackDunnNZ


tkelman:
cool

Issue #340, opened 2015-01-04
IndexedVector and AffExpr


mlubin:
I'm trying to use an ``IndexedVector`` where the coefficients are ``AffExpr`` but this doesn't quite work with the ``if val != zero(T)`` test which unintentionally creates a constraint object.
@IainNZ, have you run into this for JuMPeR?

joehuchette:
Here's a crazy idea: what about removing all the methods for ``==`` and friends to construct constraints, and force users to use the macros? I don't see any reason why you would need to use the function version of ``addConstraint`` anymore, and there are all kinds of weird issues that crop up like this with having equality overloaded.

mlubin:
Hmm. That could work for JuMP and ``AffExpr`` but in extensions we may still need to use operator overloading for constraints, at least until it's easy to extend the macros.

mlubin:
I like the idea though.

IainNZ:
I haven't encountered this in JuMPeR because I haven't tried - the places I would use it I just don't (e.g. printing).

Macros seem to work 100% with JuMPeR (although perf is ??? but I don't care) so I think I'd be OK with that too. Something to think about. Can we replace the equality test with `isequal` or something?

IainNZ:
This reminds me of my PR to Julia relating to sparse matrices - someone suggested creating a new `Base.iszero` for this reason.

joehuchette:
@IainNZ it looks like you only had to add [this](https://github.com/IainNZ/JuMPeR.jl/blob/05a398b678b07245945356d92e63886feefaf8ed/src/robustmacro.jl#L105-L109) to JuMPeR to make macros work, is that right? We probably make the code in JuMP more general to allow packages to use macros without touching the macro code themselves.

joehuchette:
In particular, can't we just change the ``addToExpression`` fallback to the one in JuMPeR?

mlubin:
In JuMPChance we need keyword arguments to ``addConstraint``:
```
addConstraint(m, z*x <= -1, with_probability=0.05,
    uncertainty_budget_mean=1, uncertainty_budget_variance=1)
```
I guess the macros could accept these and pass them along also.

mlubin:
[...]

PR #401, opened 2015-02-22
[RFC] Remove comparison operators

joehuchette:
Ref #340, this passes all tests except for some of the operators ones, where there's something issues with ``a^Tx <= b`` becoming ``-a^Tx >= -b``. Still needs docs and all that, as well as deprecations for the operators.

cc @mlubin @IainNZ

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -32,7 +32,8 @@ export
     affToStr, quadToStr, conToStr, chgConstrRHS,
 
 # Macros and support functions
-    @addConstraint, @addConstraints, @LinearConstraint, @LinearConstraints,
+    @addConstraint, @addConstraints,
+    @LinearConstraint, @LinearConstraints, @QuadConstraint, @QuadConstraints,
     @defVar, @defConstrRef, @setObjective, addToExpression, @defExpr,
     @setNLObjective, @addNLConstraint, @addNLConstraints,
     @defNLExpr
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -32,8 +32,8 @@ export
     affToStr, quadToStr, conToStr, chgConstrRHS,
 
 # Macros and support functions
-    @addConstraint, @addConstraints, @defVar,
-    @defConstrRef, @setObjective, addToExpression, @defExpr,
+    @addConstraint, @addConstraints, @LinearConstraint, @LinearConstraints,
+    @defVar, @defConstrRef, @setObjective, addToExpression, @defExpr,
     @setNLObjective, @addNLConstraint, @addNLConstraints,
     @defNLExpr
 
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -166,10 +166,11 @@ macro addLazyConstraint(cbdata, x)
     if length(x.args) == 3 # simple comparison
         lhs = :($(x.args[1]) - $(x.args[3])) # move everything to the lhs
         newaff, parsecode = parseExpr(lhs, :aff, [1.0])
+        sense = _canonicalize_sense(x.args[2])
         quote
             aff = AffExpr()
             $parsecode
-            constr = $(x.args[2])($newaff,0)
+            constr = _construct_constraint!($newaff, $(quot(sense)))
             addLazyConstraint($cbdata, constr)
         end
     else
@@ -199,10 +200,11 @@ macro addUserCut(cbdata, x)
     if length(x.args) == 3 # simple comparison
         lhs = :($(x.args[1]) - $(x.args[3])) # move everything to the lhs
         newaff, parsecode = parseExpr(lhs, :aff, [1.0])
+        sense = _canonicalize_sense(x.args[2])
         quote
             aff = AffExpr()
             $parsecode
-            constr = $(x.args[2])($newaff,0)
+            constr = _construct_constraint!($newaff, $(quot(sense)))
             addUserCut($cbdata, constr)
         end
     else
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -212,10 +212,11 @@ macro LinearConstraint(x)
         sense in valid_senses ||
             error("in @LinearConstraint ($(string(x))): expected comparison operator (<=, >=, or ==).")
         lhs = :($(x.args[1]) - $(x.args[3]))
-        # newaff, parsecode = parseExpr(lhs, :q, [1.0])
         return quote
             newaff = @defExpr($(esc(lhs)))
-            _construct_constraint!(newaff,$(quot(sense)))
+            c = _construct_constraint!(newaff,$(quot(sense)))
+            isa(c, LinearConstraint) || error("Constraint in @LinearConstraint is really a $(typeof(c))")
+            c
         end
     elseif length(x.args) == 5
         # Ranged row
@@ -224,7 +225,6 @@ macro LinearConstraint(x)
         end
         lb = x.args[1]
         ub = x.args[5]
-        # newaff, parsecode = parseExpr(x.args[3],:aff, [1.0])
         return quote
             if !isa($(esc(lb)),Number)
                 error(string("in @LinearConstraint (",$(string(x)),"): expected ",$(string(lb))," to be a number."))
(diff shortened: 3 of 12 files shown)


joehuchette:
On second thought, maybe deprecation isn't the right choice so we can immediately use ``==`` for something more meaningful.

src/JuMP.jl
```diff
@@ -32,8 +32,8 @@ export
     affToStr, quadToStr, conToStr, chgConstrRHS,
 
 # Macros and support functions
-    @addConstraint, @addConstraints, @defVar,
-    @defConstrRef, @setObjective, addToExpression, @defExpr,
+    @addConstraint, @defConstraint, @addConstraints, @defConstraints,
```
mlubin: Why is ``@defConstraint`` needed? Who needs to use this?

mlubin:
For a significant breaking change like this, we definitely need to deprecate and likely leave the operators around for one or two release cycles.

src/JuMP.jl
joehuchette: Me: https://github.com/joehuchette/CutPools.jl/blob/badf1e364ad197d2c8b46678e34fec63b635be14/src/CutPools.jl#L12

mlubin: Why not use the explicit constructor?

joehuchette: Why not use nicer syntax?

joehuchette: I have some code that treats constraints as first-class (add them to formulations, use in callbacks, etc), so it's nice to have syntax to construct them without adding them to a model.

mlubin: This isn't really user facing though, and it seems to have added a lot of complexity to the ``addConstraint`` macro implementation. Do you also need JuMPDicts of constraint objects?

joehuchette: My codes looks like
```jl
function foo()
    return [
        dˣ[i,j] ≤ cˣ[i] - cˣ[j] + 2*(Lˣ - lbˣ[i] - lbˣ[j])*(1 - zˣ[j,i])
        dˣ[i,j] ≤ cˣ[j] - cˣ[i] + 2*(Lˣ - lbˣ[i] - lbˣ[j])*(1 - zˣ[i,j])
        dʸ[i,j] ≤ cʸ[i] - cʸ[j] + 2*(Lʸ - lbʸ[i] - lbʸ[j])*(1 - zʸ[j,i])
        dʸ[i,j] ≤ cʸ[j] - cʸ[i] + 2*(Lʸ - lbʸ[i] - lbʸ[j])*(1 - zʸ[i,j])
    ]
end
```
so I was planning on making something like
```jl
function foo()
    return @defConstraints(begin
        dˣ[i,j] ≤ cˣ[i] - cˣ[j] + 2*(Lˣ - lbˣ[i] - lbˣ[j])*(1 - zˣ[j,i])
        dˣ[i,j] ≤ cˣ[j] - cˣ[i] + 2*(Lˣ - lbˣ[i] - lbˣ[j])*(1 - zˣ[i,j])
        dʸ[i,j] ≤ cʸ[i] - cʸ[j] + 2*(Lʸ - lbʸ[i] - lbʸ[j])*(1 - zʸ[j,i])
        dʸ[i,j] ≤ cʸ[j] - cʸ[i] + 2*(Lʸ - lbʸ[i] - lbʸ[j])*(1 - zʸ[i,j])
    end)
end
```
(though it currently wouldn't). Open to other suggestions, though.

mlubin: I think I'd be happier with a more modular implementation which didn't touch ``@addConstraint`` and just called ``@defExpr`` on both sides of the comparison.

joehuchette: We could implement ``@defConstraint`` so that it returns a collection of constraints (or JuMPContainers), and ``@addConstraint`` as a wrapper around this that just adds all the constraints to a model.

mlubin: I'm not really convinced that we should provide a nice syntax for this. Someone who doesn't understand JuMP's internals could easily read ``@defConstraint`` as defining a constraint that will be enforced by the model (i.e., added to the model). The distinction between ``@defConstraint`` and ``@addConstraint`` is a whole new unintuitive concept that we would need to add to the manual.

joehuchette: I'm open to other names, but I still feel it's important to be able to construct constraints using "JuMP syntax" without necessarily attaching them to models.

mlubin: How about
```
@LinearConstraint( sum{a[i]*x[i],i=1:N} <= C)
```
as a pun on the constructor?
Is the JuMPDict functionality important?

joehuchette: That should work.

joehuchette:
Updated to remove ``@defConstraint`` and add ``@LinearConstraint`` that returns a ``LinearConstraint`` and ``@LinearConstraints`` that returns a vector of ``LinearConstraint``s.

src/JuMP.jl
```diff
@@ -471,6 +471,18 @@ function addConstraint(m::Model, c::LinearConstraint)
     return ConstraintRef{LinearConstraint}(m,length(m.linconstr))
 end
 
+function construct_constraint(aff::Union(Variable,AffExpr),lb,ub)
+    if lb == -Inf
+        return LinearConstraint(aff-aff.constant, -Inf, ub-aff.constant)
```
mlubin: Unnecessary temporary allocation here with ``aff-aff.constant``.

src/callbacks.jl
```diff
@@ -166,10 +166,17 @@ macro addLazyConstraint(cbdata, x)
     if length(x.args) == 3 # simple comparison
         lhs = :($(x.args[1]) - $(x.args[3])) # move everything to the lhs
         newaff, parsecode = parseExpr(lhs, :aff, [1.0])
+        lb, ub = 0.0, 0.0
+        if x.args[2] == :(<=) || x.args[2] == :≤
```
mlubin: Why do we need this new logic here? What about a ``construct_constraint`` method which takes a sense symbol?

src/macros.jl
```diff
@@ -168,10 +174,78 @@ macro addConstraint(m, x, extra...)
...
+
+    if length(x.args) == 3
+        # Simple comparison - move everything to the LHS
+        if !((x.args[2] == :<=) || (x.args[2] == :≤) ||
+             (x.args[2] == :>=) || (x.args[2] == :≥) ||
+             (x.args[2] == :(==)))
+            error("in @LinearConstraint ($(string(x))): expected comparison operator (<=, >=, or ==).")
+        end
+        lhs = :($(x.args[1]) - $(x.args[3]))
+        newaff, parsecode = parseExpr(lhs, :q, [1.0])
```
mlubin: Just call ``@defExpr`` here. There's no need to interact with the low-level ``parseExpr`` code.

src/macros.jl
```diff
@@ -168,10 +174,78 @@ macro addConstraint(m, x, extra...)
...
+        lb, ub = 0.0, 0.0
+        if x.args[2] == :(<=) || x.args[2] == :≤
```
mlubin: This logic is duplicated in 4 places now.

joehuchette:
If this approach looks alright to everyone, I can go ahead in finishing up the PR.

mlubin:
Yes, looks good so far.

joehuchette:
I think this is ready to go now.

joehuchette:
The only thing I'm somewhat unhappy with is that ``@LinearConstraint`` really builds either linear or quadratic constraints, and there is no analogous ``@QuadConstraint``, but I think I can live with this inconsistency.

src/operators.jl
```diff
@@ -244,14 +260,14 @@ end
...
 
-for sgn in (:<=, :(==), :>=)
-    for term in (:Real, :Variable, :AffExpr)
-        @eval begin
-            $(sgn)(a::JuMPContainer, b::$(term)) = error("Cannot construct constraint with a JuMPDict term")
-            $(sgn)(a::$(term), b::JuMPContainer) = error("Cannot construct constraint with a JuMPDict term")
-        end
-    end
-end
+# for sgn in (:<=, :(==), :>=)
```
mlubin: Just delete these lines? Will be subsumed by vectorized operations.

test/operator.jl
```diff
@@ -131,6 +164,9 @@ facts("[operator] Testing basic operator overloads") do
     @fact conToStr(aff ≤ 1) => "7.1 x $leq -1.5"
     @fact conToStr(aff == 1) => "7.1 x $eq -1.5"
     @fact conToStr(aff ≥ 1) => "7.1 x $geq -1.5"
```
mlubin: You can remove the lines that will throw deprecation warnings.

joehuchette: Might as well keep testing them as long as we have them.

test/print.jl
```diff
@@ -468,11 +468,11 @@ facts("[print] expressions") do
     io_test(REPLMode, mod.linconstr[end], "x[1] + 2 y[2,3] $le 3")
     io_test(IJuliaMode, mod.linconstr[end], "x_{1} + 2 y_{2,3} \\leq 3")
 
-     addConstraint(mod, (x[1]+x[2])*(y[2,2]+3.0) <= 1)
-    io_test(REPLMode, mod.quadconstr[end], "x[1]*y[2,2] + x[2]*y[2,2] + 3 x[1] + 3 x[2] - 1 $le 0")
-    io_test(IJuliaMode, mod.quadconstr[end], "x_{1}\\timesy_{2,2} + x_{2}\\timesy_{2,2} + 3 x_{1} + 3 x_{2} - 1 \\leq 0")
+    # @addConstraint(mod, (x[1]+x[2])*(y[2,2]+3.0) <= 1)
```
mlubin: Why commented out?

test/operator.jl
mlubin: Not if they print out tons of annoying warnings, let's see

test/print.jl
joehuchette: Not sure, think it just failed on an older commit.

test/operator.jl
joehuchette: I'm using ``Base.warn_once``, so it's only like 5 lines

mlubin:
We could add a ``QuadConstraint`` macro, and add a type assertion when using ``@LinearConstraint``.

mlubin:
Docs need updating also.

joehuchette:
Yeah, that seems fair. Where? I thought I got the docs already.

mlubin:
Did you check if the operators are referenced anywhere else in the docs?

joehuchette:
Think so, but I might have missed something.

test/print.jl
joehuchette: Ahh, this was originally w/o the macro, so the macro version fails on v0.3. I'll only run it on v0.4.

mlubin:
Some of the examples (qcp.jl) also use operators.

mlubin:
In the future, could you squash commits with failing tests? This could cause trouble if we want to run ``git bisect`` sometime later.

joehuchette:
Ahh, yep.

IainNZ:
This is a pretty big day in JuMP history!

joehuchette:
Hopefully it doesn't break too much...

mlubin:
We likely need to leave the deprecations in for a good year or so to avoid heavy breakage of existing code.

PR #404, opened 2015-02-25
[RFC] Vectorized operations on JuMPArrays

joehuchette:
```jl
julia> using JuMP

julia> m=Model();

julia> @defVar(m,x[1:3]);

julia> A = rand(3,3)
3x3 Array{Float64,2}:
 0.487805  0.775064  0.300965
 0.135576  0.962211  0.89139
 0.94714   0.107853  0.865879

julia> x'A*x
1-element Array{JuMP.GenericQuadExpr{Float64,JuMP.Variable},1}:
 0.48780514045744394 x[1]² + 0.9106396785074213 x[1]*x[2] + 0.9622110354003406 x[2]² + 1.2481052489643654 x[1]*x[3] + 0.9992426781705144 x[2]*x[3] + 0.8658785594081322 x[3]²
```
No idea how efficient this is (probably not very), but that doesn't seem like it should be a prime concern at this point. TODO:

* [ ] Fallback that collects/reshapes ``AbstractArrays``
* [x] Add specialized methods for sparse matrices
* [x] Vectorized constraints


mlubin:
Plus tests. No need to worry about efficiency at this point.

joehuchette:
Hah, I wasn't even thinking about tests/docs at this point, just trying to get something working.

tkelman:
```
julia> x'A*x
1-element Array{JuMP.GenericQuadExpr{Float64,JuMP.Variable},1}:
```
This is wonderful. You guys rock. cc @smukoehler

joehuchette:
Thoughts on using elementwise comparisons like ``.>=`` for vectorized constraints? I think it's more clear (and opens up ``>=`` for semidefinite), though ``@addConstraint(m, x'A*x .>= 1)`` is an interesting corner case.

mlubin:
@joehuchette, I like that, it gives us a syntactic way to differentiate scalar and vector constraints.

IainNZ:
Regarding the corner case, I know its horribly un-Julia but I'd be tempted to embrace our DSL-ness by  MATLAB-ing it and treating 1-by-1 as a scalar...

tkelman:
I think dot-operators for vector constraints would make sense. It's a bit unfamiliar for Matlab/Python/Ampl users but so are the `@` symbols for macros initially.

mlubin:
@IainNZ treating 1-by-1 as a scalar means that we can't assume that we have scalar expressions when ``>=`` appears, which would have solved an issue with the macros that @joehuchette tried to work around with https://github.com/JuliaOpt/JuMP.jl/commit/924d991409da74af349bb0d66893e5e0cdb99f4f.

joehuchette:
I think that it's actually good to keep the expression parsing code relatively agnostic and only enforce the operators are right via multi dispatch. We should benchmark but I don't think there will be an appreciable performance hit from my commutativity fix.

I think special-casing the 1x1 case makes things a little cleaner visually, but it might confuse users wondering why 
```jl
a = Array(Int,1,1)
a >= 1
```
doesn't work. And it will require extra code.

mlubin:
Philosophically I don't have an objection to ``x'Qx <= 1`` just working, since we're a DSL and we're not even overriding the operators. But I think that if we ever want to rework how we parse things (e.g., merging with nonlinear code?), it's going to be a huge help to have a syntactic hint that differentiates between the vector and scalar case. We can always add it in later, but it's harder to take it away once it's established.

mlubin:
Just realized that we won't have this syntax hint when setting the objective, so @joehuchette's commutativity fix is still needed.

mlubin:
On second thought, we could require that the objective be a scalar anyway.

joehuchette:
Any good test cases for vectorized constraints would be much appreciated

IainNZ:
Here you go:

```julia
using JuMP, Gurobi

function bestsubset(n,p,X,y,K,M)
    mod = Model(solver=GurobiSolver(OutputFlag=0))
    @defVar(mod, β[1:p])
    @defVar(mod, z[1:p], Bin)
    @setObjective(mod, Min, (y-X*β).^2)
    @addConstraint(mod, eye(p)*β .<=  M*eye(p)*z)
    @addConstraint(mod, eye(p)*β .>= -M*eye(p)*z)
    @addConstraint(mod, sum(z) == K)
    solve(mod)
    return getValue(β)[:]
end

# y = 100 x1  +  50 x2  +  10 x3  +  1 x4
srand(1000)
n = 1000
p = 4
X = rand(n,p)
y = X * [100, 50, 10, 1] + 20*randn(n)
@show bestsubset(n,p,X,y,2,500)
```

joehuchette:
:+1: 
```jl
n = 1000
p = 4
K = 2
M = 500
X = rand(n,p)
y = X * [100, 50, 10, 1] + 20*randn(n)
mod = Model()
@defVar(mod, β[1:p])
@defVar(mod, z[1:p], Bin)
@setObjective(mod, Min, (y-X*β)'*(y-X*β) )
@addConstraint(mod, eye(p)*β .<=  M*eye(p)*z)
@addConstraint(mod, eye(p)*β .>= -M*eye(p)*z)
@addConstraint(mod, sum(z) == K)
solve(mod)
```
```jl
julia> @show getValue(β)[:]
(getValue(β))[:] = [107.54138090112164,54.0818239607563,0.0,0.0]
4-element Array{Float64,1}:
 107.541
  54.0818
   0.0
   0.0
```

joehuchette:
I rly hope ``@addConstraint(mod, eye(p)*β .<=  M*eye(p)*z)`` doesn't become a thing, though...

IainNZ:
Haha yeah, that wasn't in version 1 of it by I thought I'd through it in for good measure.

IainNZ:
Are we going to collapse singleton dimensions, e.g. is 
```
@addConstraint(mod, A*x[:,:,1] .<= b)
```
ok? Its so hard to think like this, but extra without that.

mlubin:
Slicing like that should just return a ``Matrix{Variable}``.

IainNZ:
Here is one, from Convex.jl
http://nbviewer.ipython.org/github/JuliaOpt/Convex.jl/blob/master/examples/control.ipynb

joehuchette:
@tkelman do you have any nice test cases in mind for this branch? I'm reasonably happy with the interface but want to stress-test it some more. (Support for sparse matrices is not there either, but that can be added independently now or later).

tkelman:
Not written in Julia at the moment. Most of the test cases I would run on would look almost identical to that example from Convex.jl, but with matrices for the dynamics and cost function.

smukoehler:
Here's some Matlab code for a Cessna Aircraft that should be easy to translate. Here's a link to some slides that explain what the problem is: https://www.kirp.chtf.stuba.sk/pc11/data/workshops/mpc/MPC_PC11_Lecture1.pdf

```
% Create LQR
A = [-1.2822 0 0.98 0;0 0 1 0;-5.4293 0 -1.8366 0;-128.2 128.2 0 0];
B = [-0.3;0;-17;0];
G = ss(A,B,[0 1 0 0;0 0 0 1],zeros(2,1))
Ts = 0.15;
Gd = c2d(G,Ts)
[K,P] = dlqr(Gd.a,Gd.b,Q,R);

%MPC controller code for first part

[n,m] = size(B);  
[K,P] = dlqr(A,B,Q,R);
H = blkdiag(kron(eye(N),Q),P);
H = blkdiag(H,kron(eye(N+1),R));
f = zeros((N+1)*(n+m),1);
  
Ain = [];
bin = [];

auxA = kron(diag(ones(N,1),-1),A);
auxB = kron(diag(ones(N,1),-1),B);
  
Aeq = [eye((N+1)*n)-auxA -auxB];
beq = [eye(n);zeros(N*n,n)]*x;
  
lb = [repmat(xmin,(N+1),1);repmat(umin,N+1,1)];
ub = [repmat(xmax,(N+1),1);repmat(umax,N+1,1)];
    
XU = quadprog(H,f,Ain,bin,Aeq,beq,lb,ub,optimset('Display','off'))
u = XU((N+1)*n+1);

% Compare with LQ&
if 1
    [u -K*x]
end
```

joehuchette:
Cool, thanks for the example. It made me realize we need to add support for concatenation over JuMPArrays. I've added this in the latest commit, following [Michael Grant's tip](https://groups.google.com/d/msg/julia-users/icTP-l-odtw/RxCL2QKZp-gJ); it seems a little brittle but hopefully works fine enough. You can do things like this now:
```jl
julia> @defVar(m,x[1:3]);

julia> @defVar(m,y[1:3,1:3]);

julia> [1 2x'
        x 2y-x*x']
4x4 Array{JuMP.GenericQuadExpr{Float64,JuMP.Variable},2}:
 1     2 x[1]                 2 x[2]                 2 x[3]
 x[1]  -x[1]² + 2 y[1,1]      -x[1]*x[2] + 2 y[1,2]  -x[1]*x[3] + 2 y[1,3]
 x[2]  -x[1]*x[2] + 2 y[2,1]  -x[2]² + 2 y[2,2]      -x[2]*x[3] + 2 y[2,3]
 x[3]  -x[1]*x[3] + 2 y[3,1]  -x[2]*x[3] + 2 y[3,2]  -x[3]² + 2 y[3,3]
```

mlubin:
Scary stuff

joehuchette:
I'm pretty much just rewriting the SDP branch at this point...

joehuchette:
Pending the CI, I think this is now RFC. I don't have specialization for sparse matrices yet, though.

joehuchette:
Hmm concatenation is messy on 0.3, looks like the unexported interface is different between 0.3 and 0.4...

mlubin:
We can just target 0.4 if it's too messy.

joehuchette:
Yeah, not really sure I see a way to do it on v0.3 without really hacking into Base and changing concat behavior globally, which I don't want to do. This is only an issue with concatenation of JuMPArrays, though, the rest of the changes are unaffected.

joehuchette:
So I changed up concatenation to work on v0.3 as well, but the git history is sufficiently complicated that merging and rebasing has become complicated...

IainNZ:
Can you squash these commits back to a single commit, but still keepign the base what it orginally was. Then you could rebase on head and just have one big conflict to go through?

joehuchette:
[This is what I end up with](https://github.com/JuliaOpt/JuMP.jl/compare/matrices3?expand=1); only the last commit is from this PR.

joehuchette:
Got it! This is RFC now.

mlubin:
Github still says that it's not mergeable.

joehuchette:
Not _automatically_ mergeable, I moved around some files that might have it confused. I was primarily just looking to squash it without adding a bunch of redundant commits, we can still merge from the CLI.

test/qcqpmodel.jl
```diff
@@ -90,6 +139,22 @@ context("With solver $(typeof(solver))") do
     @fact modQ.objVal => roughly(sqrt(1/2), 1e-6)
     @fact norm([getValue(x), getValue(y), getValue(t)] - [0.5,0.5,sqrt(1/2)]) => roughly(0.0,1e-3)
 
+    # # Vectorized version
```
mlubin: Why is this commented out?

mlubin:
Let's check out the coveralls report once travis finishes.

test/qcqpmodel.jl
joehuchette: https://github.com/JuliaOpt/MathProgBase.jl/issues/61

mlubin: You can tag MPB.

mlubin:
https://coveralls.io/builds/2093584/source?filename=src%2Foperators.jl
There's a good number of new untested lines.

IainNZ:
I'd really like to do a really close read of this before merging, its just a lot of new code

joehuchette:
Yes please please do. Granted, about half of the new LOCs are test data, but still.

doc/refexpr.rst
```diff
@@ -129,3 +129,47 @@ references in this structure, e.g.::
       myCons[i] = @addConstraint(m, x[i] >= i)
     end
 
+Vectorized operations
+^^^^^^^^^^^^^^^^^^^^^
+
+JuMP supports vectorized expressions and constraints for linear and quadratic models. Although this syntax may
+be familiar for users coming from other modeling languages, we caution that this syntax can be slow---especially
```
mlubin: s/other/MATLAB-based/
It won't be familiar to AMPL/GAMS users.

doc/refexpr.rst
```diff
@@ -129,3 +129,47 @@ references in this structure, e.g.::
...
+for large operations---and that it often can destroy the sparse, algebraic structure that makes modeling
```
mlubin: What does "destroy" mean here? If you use sparse matrices, it shouldn't destroy the structure.

doc/refexpr.rst
```diff
@@ -129,3 +129,47 @@ references in this structure, e.g.::
...
+languages useful. Nevertheless, the syntax often proves useful, for example in constraints involving small,
+dense matrix-vector products.
+
+Linear algebraic operators have been added to give meaning to expressions like ``A*x`` where ``A`` is a matrix
```
mlubin: I would just describe the functionality instead of using the past tense "have been added"

src/JuMP.jl
```diff
@@ -329,7 +340,7 @@ typealias AffExpr GenericAffExpr{Float64,Variable}
 AffExpr() = AffExpr(Variable[],Float64[],0.0)
 
 Base.isempty(a::AffExpr) = (length(a.vars) == 0 && a.constant == 0.)
-Base.convert(::Type{AffExpr}, v::Variable) = AffExpr([v], [1.], 0.)
+Base.convert(::Type{AffExpr}, v::Variable) = AffExpr(Variable[v], [1.], 0.)
```
mlubin: Why is this change needed?

src/JuMP.jl
```diff
@@ -143,6 +143,17 @@ function setObjectiveSense(m::Model, newSense::Symbol)
 end
 setObjective(m::Model, something::Any) =
     error("in setObjective: needs three arguments: model, objective sense (:Max or :Min), and expression.")
+
+# # Hacky little fallbacks so that arrays of length 1 act as scalars for adding constraints and setting objective
```
mlubin: No need to leave in commented code. But do we give a nice message now when a vector is given for the objective?

src/operators.jl
```diff
@@ -261,6 +261,186 @@ function Base.dot{N}(lhs::JuMPArray{Float64,N},rhs::JuMPArray{Float64,N})
     return sum(lhs.innerArray .* rhs.innerArray)
 end
 
+Base.promote_type(::Type{Variable},::Type{Union()})  = Variable
```
mlubin: Why do we need promotions from ``Union()``?

src/operators.jl
```diff
@@ -261,6 +261,186 @@ function Base.dot{N}(lhs::JuMPArray{Float64,N},rhs::JuMPArray{Float64,N})
...
+        ret[i,j] = q
+    end
+    return ret
+end
+
+typealias JuMPTypes Union(Variable,AffExpr,QuadExpr)
+
+(*)(lhs::AbstractArray, rhs::OneIndexedArray) = (*)(full(lhs), rhs.innerArray)
+(*)(lhs::OneIndexedArray, rhs::AbstractArray) = (*)(lhs.innerArray, full(rhs))
+(*){T<:JuMPTypes}(lhs::SparseMatrixCSC, rhs::Array{T}) = (*)(full(lhs),rhs)
```
mlubin: *cringe*
I don't feel good about releasing this without sparse matrix support.

src/v0.3/concatenation.jl
```diff
@@ -0,0 +1,162 @@
...
+
+typealias BaseTypes Union(Number,AbstractArray)
+typealias CatTypes Union(Number,JuMPTypes,AbstractArray)
+typealias CatTypesFull Union(CatTypes,OneIndexedArray)
+
+Base.vcat(X::CatTypes...) = Base.cat(1, X...)
+Base.hcat(X::CatTypes...) = Base.cat(2, X...)
+
+###############################################################################
+# copied from Base at https://github.com/JuliaLang/julia/blob/0c24dca65c031820b91721139f0291068086955c/base/abstractarray.jl#L613-L668
```
mlubin: We need to include a copyright and license notice if we're copying code from julia. It can just be in a comment right here.

src/v0.4/parseExpr_staged.jl
```diff
@@ -131,8 +139,12 @@ stagedfunction addToExpression_reorder(ex, args...)
     # how many of the multiplicands are variables?
     n_var = mapreduce(t -> (t == Variable || t == AffExpr), +, args)
     has_quad = mapreduce(t -> (t == QuadExpr), |, args)
+    # has_vect = any(t -> (t <: Array), args)
     if n_var == 0 && !has_quad
         #println("No variables")
+        # println(:(addToExpression(ex, 1.0, (*)(args...))))
+        # println("ex = $ex")
+        # println("args = $args")
```
mlubin: Dangling debug code?

test/model.jl
```diff
@@ -473,6 +489,60 @@ context("With solver $(typeof(solver))") do
...
+    function bestsubset(solver,X,y,K,M,integer)
+        mod = Model(solver=solver)
+        @defVar(mod, β[1:p])
+        if integer
+            @defVar(mod, z[1:p], Bin)
+        else
+            @defVar(mod, 0 <= z[1:p] <= 1)
+        end
+        obj = (y-X*β)'*(y-X*β)
+        @setObjective(mod, Min, obj[1])
```
mlubin: Just write ``dot(y-X*β,y-X*β)``?

IainNZ: That came from me, it was a stress test to see what would happen

IainNZ:
Are we still possibly going to move `JuMPContainer` into its own package? I'm really into the idea and this kinda relates to that.

mlubin:
I'm not opposed, but it shouldn't block this.

src/JuMP.jl
joehuchette: I think it was causing a stack overflow in the concatenation code, but I'm not sure if it's necessary anymore.

src/operators.jl
joehuchette: Ambiguity warnings at one point.

src/v0.3/concatenation.jl
joehuchette: The full MIT license?

src/operators.jl
```diff
@@ -261,6 +261,189 @@ function Base.dot{N}(lhs::JuMPArray{Float64,N},rhs::JuMPArray{Float64,N})
...
+
+function multiply!(ret, A, x)
+    m, n = size(A,1), size(A,2)
+    for i in 1:m, j in 1:n
+        ret[i] += A[i,j] * x[j]
+    end
+    return ret
+end
+
+function multiply!(ret::Array{AffExpr}, lhs, rhs)
```
IainNZ: Could this be loosened to GenericAffExpr?

IainNZ: e.g. `multiply!{T<:GenericAffExpr}(ret::Array{T}, lhs, rhs)`

IainNZ:
I'm trying to figure out how much of this can be leveraged for JuMPeR...

IainNZ:
Just had another request for this, will definitely be nice for people transitioning from YALMIP-land I think

joehuchette:
After generals on Wednesday :)

tkelman:
Several people in my lab have been eagerly awaiting this. (I've been encouraging them to speak up on github issues.)

joehuchette:
Coverage.jl says this is at 78.5% right now.

joehuchette:
(Though that might be artificially inflated since a lot of the new operators are defined in an ``@eval`` block).

mlubin:
I'm getting a lot of ambiguity warnings:
```
Warning: New definition 
    promote_type(Type{R<:Real},Type{Variable}) at /home/mlubin/.julia/v0.3/JuMP/src/operators.jl:271
is ambiguous with: 
    promote_type(Type{T},Type{T}) at promotion.jl:101.
To fix, define 
    promote_type(Type{_<:Real},Type{_<:Real})
before the new definition.
Warning: New definition 
    promote_type(Type{R<:Real},Type{GenericAffExpr{Float64,Variable}}) at /home/mlubin/.julia/v0.3/JuMP/src/operators.jl:283
is ambiguous with: 
    promote_type(Type{T},Type{T}) at promotion.jl:101.
To fix, define 
    promote_type(Type{_<:Real},Type{_<:Real})
before the new definition.
Warning: New definition 
    promote_type(Type{R<:Real},Type{GenericQuadExpr{Float64,Variable}}) at /home/mlubin/.julia/v0.3/JuMP/src/operators.jl:292
is ambiguous with: 
    promote_type(Type{T},Type{T}) at promotion.jl:101.
To fix, define 
    promote_type(Type{_<:Real},Type{_<:Real})
before the new definition.
Warning: New definition 
    promote_type(Type{T<:JuMPArray{T,N,true}},Type{S<:JuMPArray{T,N,true}}) at /home/mlubin/.julia/v0.3/JuMP/src/v0.3/concatenation.jl:17
is ambiguous with: 
    promote_type(Type{T},Type{T}) at promotion.jl:101.
To fix, define 
    promote_type(Type{_<:JuMPArray{T,N,true}},Type{_<:JuMPArray{T,N,true}})
before the new definition.
```

mlubin:
I'd like to try to get this in or almost in before the weekend so we can focus on the rest of the 0.10 issues as well.

joehuchette:
I think it's almost ready, I can try to finish up on Thursday/Friday. Maybe we can schedule a code review for the weekend?

mlubin:
Sounds good

IainNZ:
Thursday/Friday code bashing is my plan

joehuchette:
Mmk I think this is ready for code review / everyones best tries at breaking things.

IainNZ:
Got this on 0.3:
```
Warning: New definition 
    promote_rule(Type{T},Type{NAtype}) at /Users/idunning/.julia/v0.3/DataArrays/src/natype.jl:37
is ambiguous with: 
    promote_rule(Type{T<:JuMPArray{T,N,true}},Type{S}) at /Users/idunning/.julia/v0.3/JuMP/src/v0.3/concatenation.jl:12.
To fix, define 
    promote_rule(Type{_<:JuMPArray{T,N,true}},Type{NAtype})
before the new definition.
```

IainNZ:
I made a MPC example
http://nbviewer.ipython.org/gist/IainNZ/822d9a9f25936966d8e1
I had an issue with matrix multiplication with variables, but I just had to switch the order of the indices (I put time as the first dimension, so my vectors were transposed the wrong way). 
I actually hit this though:
http://stackoverflow.com/questions/30242576/how-can-i-write-code-to-work-with-1d-matrices-in-julia

joehuchette:
Yeah I'm not sure of a way around that ambiguity warning, short of adding extra methods conditional on DataArrays, which is hacky.

joehuchette:
Maybe I can only define concatenation in terms of ``Number``s, ``AbstractArray``s, and JuMP types? Though I feel that might also be liable to introduce ambiguity warnings.

tkelman:
@IainNZ that example looks really great. My advisor was just asking me today about the best package to use for plotting animations, and we were struggling with getting PyCall to work on a Windows machine that didn't have Visual Studio. I'll want to show him that notebook once it doesn't require too much Pkg fiddling to run it. Worth setting the solver to Ipopt just so anyone can run it?

IainNZ:
Yeah I'll do that for the final version. I'm planning on implementing the Cessna example too.

src/vectorized_macros_overloads.jl
```diff
@@ -0,0 +1,39 @@
+#  Copyright 2015, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+typealias VectTypes Union(JuMPTypes,Real)
+
+# Remember: it is only safe to mutate the first argument!
+
+function addToExpression{T<:VectTypes}(x::AffExpr, y::Real, z::Array{T})
```
mlubin: When are these methods used? Why would we have empty ``AffExpr``s?

mlubin: Never mind, I see it in the macros.

NEWS.md
```diff
@@ -5,6 +5,8 @@ Unversioned
 -----------
 
   * Support for conditions in indexing ``@defVar`` and ``@addConstraint`` constructs, e.g. ``@defVar(m, x[i=1:5,j=1:5; i+j >= 3])
+  * Support for vectorized operations on Variables, expressions, and JuMPArrays with indexing over unit ranges starting at one. See
```
mlubin: I would say "unit-step ranges"

src/JuMP.jl
```diff
@@ -493,6 +515,10 @@ function addConstraint(m::Model, c::LinearConstraint)
     end
     return ConstraintRef{LinearConstraint}(m,length(m.linconstr))
 end
+addConstraint(m::Model, c::Array{LinearConstraint}) =
+    error("Vectorized constraint added without elementwise comparisons. Try using one of (.<=,.>=,.==).")
```
mlubin: I'd reword this as:
"The operators <=, >=, and == can only be used to specify scalar constraints. If you are trying to add a vectorized constraint, use the element-wise dot comparison operators (.<=, .>=, or .==) instead".

mlubin:
I commented out the promotion rule which was causing ambiguities:
```
Base.promote_rule{T<:OneIndexedArray,S}(::Type{T},::Type{S}) = DummyJuMPArray
```
and the tests still passed. So where is this used?

joehuchette:
I have no idea how, but that does seem to work. Something with ``promote_rule`` that I don't understand:
```jl
julia> @defVar(m, x[1:3]);

julia> promote_rule(typeof(x), typeof(1))
None

julia> Base.promote_rule{T<:JuMP.OneIndexedArray,S}(::Type{T},::Type{S}) = JuMP.DummyJuMPArray;

julia> promote_rule(typeof(x), typeof(1))
DummyJuMPArray (constructor with 1 method)
```

joehuchette:
I think the ambiguity is actually cased by the
```jl
Base.promote_rule{T<:OneIndexedArray}(::Type{T},::Type{Union()}) = DummyJuMPArra
```
method, which can be safely removed at this point as I only added it in the first place to avoid ANOTHER ambiguity warning...

mlubin:
I'm still getting ambiguity warnings with DataArrays on the latest commit.

joehuchette:
Yeah, I'm not sure of a good way around that, unfortunately. DataArrays folks seem to know that what we're doing isn't quite alright, too: https://github.com/JuliaStats/DataArrays.jl/blob/d78a6d3141a7ab4c1365a52e9bafcdc74b8672c1/src/natype.jl#L36-L37

src/JuMP.jl
```diff
@@ -414,6 +433,9 @@ function setObjective(m::Model, sense::Symbol, q::QuadExpr)
     setObjectiveSense(m, sense)
 end
 
+# Copy an affine expression
```
IainNZ: Wrong comment, but probably can just drop the comment anyway

joehuchette: Lol whoops

src/macros.jl
```diff
@@ -128,19 +128,28 @@ function assert_validmodel(m, macrocode)
     end
 end
 
-
-const valid_senses = [:(<=), :≤, :(>=), :≥, :(==)]
```
IainNZ: This is the only thing that broke JuMPeR, which is nothing in the scheme of things! I do have screens of ambiguities to fix though...

test/qcqpmodel.jl
```diff
@@ -19,23 +19,46 @@ context("With solver $(typeof(solver))") do
...
+         0.5 -1    0]
+    @addConstraint(modV, A*x .<= zeros(2))
+    @fact solve(modV) => :Optimal
+    @fact modV.objVal => roughly( 247.0, 1e-5)
+    @fact getValue(x)[:] => roughly([2.0, 3.0, 4.0], 1e-6)
 
     modQ = Model(solver=solver)
     @defVar(modQ, 1.1*i <= x[i=1:3] <= 2.5*i, Int)
-    setObjective(modQ, :Max, -10*x[1]*x[1] - 3*x[1]*x[2] - 5*x[2]*x[2] - 9*x[3]*x[3])
+    @setObjective(modQ, :Max, -10*x[1]*x[1] - 3*x[1]*x[2] - 5*x[2]*x[2] - 9*x[3]*x[3])
```
IainNZ: Can drop the `:`

test/qcqpmodel.jl
```diff
@@ -45,12 +68,22 @@ context("With solver $(typeof(solver))") do
     modQ = Model(solver=solver)
     @defVar(modQ, 0.5 <= x <= 2 )
     @defVar(modQ, 0 <= y <= 30 )
-    setObjective(modQ, :Min, (x+y)*(x+y) )
+    @setObjective(modQ, :Min, (x+y)*(x+y) )
```
IainNZ: Another `:`

mlubin:
Given all the internal restructuring, we should rerun some benchmarks before merging.

joehuchette:
Mmk @IainNZ the ``addToExpression`` methods in JuMPeR are no longer necessary!

PR #413, opened 2015-03-10
Relicense code and tests under mozilla public license version 2

### LINKED ISSUE #410: MPL?
Saw some discussion [here](https://news.ycombinator.com/item?id=9156644) where [MPL](https://tldrlegal.com/license/mozilla-public-license-2.0-(mpl-2)) came up. It makes much more sense to me than LGPL, especially for Julia code.
Under MPL:
- If someone embeds JuMP in a commercial application, they must release the source code and any modifications they've made at the source level.

Under LGPL:
- All of the above, and they must "link" to JuMP dynamically and let users "link" in their own version of JuMP.

It's not clear at all what "linking" means for Julia code, and if it's not clear to us, I don't see how anyone else could confidently interpret the license. It doesn't help us to have a licence that's difficult to interpret and apply.

mlubin:
closes #410

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,16 +1,381 @@
-Copyright (c) 2014: Iain Dunning, Miles Lubin, Joseph Huchette, and John Lepird.
+Copyright (c) 2015: Iain Dunning, Joey Huchette, Miles Lubin, and contributors
 
-The JuMP Julia module is licensed under the **[LGPL]** version 3:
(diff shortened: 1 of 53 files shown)


mlubin:
I didn't put MPL headers on the code in ``examples/``. Should these be more liberally licensed? Does it matter?

joehuchette:
Probably doesn't matter

mlubin:
Ready to go

PR #419, opened 2015-03-19
a slight change in the explanation of setValue.

chkwon:


mlubin:
Thanks for catching this!

chkwon:
my pleasure.

Issue #405, opened 2015-02-26
Easily accessing solve time in JuMP


Shuvomoy:
Per Miles's suggestion I am opening this issue.

Currently  solve time in JuMP has to be accessed either by looking into the solver print out, or by running `@time begin 
solve (modelName) 
end`,
which gives a slightly higher value. It might be better if it could be accessed more directly.

mlubin:
So MathProgBase has ``getsolvetime``, we just have to expose it and make sure it's actually implemented by solvers. PRs accepted.

mlubin:
You can also try ``MathProgBase.getsolvetime(getInternalModel(jumpmodel))``.

IainNZ:
Would be awesome if you could submit this feature @Shuvomoy, a good start to future larger contributions maybe? :D

mlubin:
Closed by #422, thanks to @odow

Issue #356, opened 2015-01-10
Should getNumConstraints count quadratic and nonlinear constraints?


tkelman:
Slightly misleading to only return the number of linear constraints from this function, isn't it?

mlubin:
Good point, these methods haven't gotten much attention. I'd vote for separating into linear/quadratic/nonlinear constraints.

joehuchette:
What about 
```
(num_lin, num_quad, num_nl) = getNumConstrs
```
?

mlubin:
Seems reasonable to me, only issue is that it's a breaking change.

tkelman:
Maybe a dumb question, but is there a good reason for these to not use the MathProgBase names/interfaces?

IainNZ:
I'm fairly ok with the breaking change without deprecation, as I doubt this is used by casual users of JuMP, so they'll be capable of updating easily enough.

mlubin:
@tkelman, not really a good reason for not using the MPB names here, except to match the JuMP naming style. These functions are low-level enough that maybe it doesn't matter.

joehuchette:
I've only ever used these functions when touching the MPB interface for other reasons (e.g. playing around with the constraint matrix), so it probably makes sense to just adopt the MPB naming conventions

PR #422, opened 2015-03-31
implement getsolvetime. issue #405

odow:
Exposes ```MathProgBase.getsolvetime``` and checks if implemented by solver. https://github.com/JuliaOpt/JuMP.jl/issues/405

### DIFF
--- a/doc/refmodel.rst
+++ b/doc/refmodel.rst
@@ -40,6 +40,7 @@ Methods
 * ``MathProgBase.numlinconstr(m::Model)`` - returns the number of linear constraints associated with the ``Model m``.
 * ``MathProgBase.numquadconstr(m::Model)`` - returns the number of quadratic constraints associated with the ``Model m``.
 * ``MathProgBase.numconstr(m::Model)`` - returns the total number of constraints associated with the ``Model m``.
+* ``MathProgBase.getsolvetime(m::Model)`` - returns the solve time from the solver if it is implemented.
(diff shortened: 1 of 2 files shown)


mlubin:
LGTM. The only question is if we should follow the MPB naming convention here, given that we decided to do so for ``numvar`` etc (#356). I'd lean towards extending the MPB method instead of defining a new name if it just accesses or imitates a property of the MPB solver.

src/solvers.jl
```diff
@@ -431,3 +431,16 @@ function merge_duplicates{CoefType,IntType<:Integer}(::Type{IntType},aff::Generi
     return indices, coeffs
 
 end
+
+# get the solve time
+function getSolveTime(m::Model)
+    if !m.internalModelLoaded
+        Base.warn("Model not solved")
+        return nothing
```
mlubin: I think it's more Julia-like (for type stability reasons) to throw an exception when the value isn't available instead of returning ``nothing``.

mlubin:
@IainNZ @joehuchette, comments on the naming issue?

IainNZ:
I'm leaning towards MathProgBase version, for now.

joehuchette:
I think I agree with Iain in theory, except for the fact that the current way to call it is so verbose with ``getInternalSolver``.

mlubin:
@joehuchette, this wrapper lets you call it without ``getInternalSolver``

joehuchette:
Understood, I'm thinking that I would just prefer a shorter alias for ``getInternalSolver``

mlubin:
@joehuchette, the wrapper method lets you give a nice error message when the solver doesn't implement ``getsolvetime``, which is pretty useful.

mlubin:
@odow, we can merge this as is. Could you squash the commits?

odow:
@mlubin  squashed.

mlubin:
Thanks!

PR #423, opened 2015-04-08
fix writeLP issue with negative coeffs

sbebo:
Cplex doesn't like .lp files containing stuff like "C1 VAR1 + -C2 VAR2".

### DIFF
--- a/src/writers.jl
+++ b/src/writers.jl
@@ -221,11 +221,19 @@ function writeLP(m::Model, fname::String)
     write(f, " obj: ")
     nnz = length(objaff.coeffs)
     for ind in 1:(nnz-1)
-        print_shortest(f, objaff.coeffs[ind])
-        @printf(f, " VAR%d + ", objaff.vars[ind].col)
+        if ind == 1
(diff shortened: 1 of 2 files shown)


mlubin:
Thanks for this. Do you know why the commit author isn't associated with your github ID?

sbebo:
What do you mean? Looks fine to me.

mlubin:
Okay, seems fine now. Will merge after CI finishes.

sbebo:
By the way, printing to file I noticed another issue: expressions where the same variable appears more than once cause an error in the MPS file ("repeated row"). (In the LP files this is not an issue, they work fine even with repeated variables.)
Shouldn't be hard to fix, one just needs to aggregate the cofficients before writing to file -- if I have 5 spare minutes.

mlubin:
@sbebo, I would've thought that wouldn't be a problem given the transpose we do here: https://github.com/JuliaOpt/JuMP.jl/blob/47f75c144bc4594b295866db860cc85bef48b5ab/src/writers.jl#L74. Could you open a separate issue for this?

PR #431, opened 2015-04-23
generalize defExpr for extensions

mlubin:
This makes a difference when building up expressions in an extension which don't have ``AffExpr``s as constant terms. E.g.,

```
using JuMPChance, JuMP
m = ChanceModel()
@defIndepNormal(m, x, mean=1, var=1)
# before
typeof(@defExpr(x+1)) # -> GenericAffExpr{GenericAffExpr{Float64,Variable},IndepNormal} (constructor with 2 methods)
# after
typeof(@defExpr(x+1)) # -> GenericAffExpr{Float64,IndepNormal}
```
This only works with 0.4, unfortunately.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -363,14 +363,21 @@ macro defExpr(args...)
 
     refcall, idxvars, idxsets, idxpairs = buildrefsets(c)
     newaff, parsecode = parseExpr(x, :q, [1.0])
-    code = quote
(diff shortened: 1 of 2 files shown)


IainNZ:
Nice, I hadn't even tried these for JuMPeR yet. Don't worry, we won't have to support 0.3 much longer, right? ;)

PR #432, opened 2015-04-23
pass along keyword arguments in addConstraint macro

mlubin:
This now works:
```
@addConstraint(m, z*x <= -1, with_probability=0.05)
@addConstraint(m, z*x <= -1; with_probability=0.05) # equivalent
```

``@addConstraints`` *kinda* works, but you need to put parentheses because of operator precedence
```
@addConstraints m begin
    (z*x <= -1, with_probability=0.05)
end
```

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -144,7 +144,28 @@ function _construct_constraint!(quad::QuadExpr, sense::Symbol)
     return QuadConstraint(quad, sense)
 end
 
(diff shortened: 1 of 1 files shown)


mlubin:
Eventually, we should have a dummy extension as part of the unit tests, but I'm okay with this being just tested through JuMPChance for now.

joehuchette:
+1 to a dummy extension

Issue #383, opened 2015-02-06
Occasional ambiguity warnings?


IainNZ:
Has anyone else seen non-deterministic ambiguity warnings?

joehuchette:
Are you on 0.3?

IainNZ:
Yeah 0.3.3

mlubin:
Haven't seen this here.

joehuchette:
I haven't seen any, but I usually use 0.4. Do you often use some other (unrelated) package at the same time?

IainNZ:
Its quite possible. The fact you've never seen suggests... something. I'll reopen if I can recreate, I've always been distracted when its happening.

IainNZ:
```
Warning: New definition
    *(S<:Union(Variable,GenericAffExpr{Float64,Variable},GenericQuadExpr{Float64,Variable}),T<:GenericQuadExpr{Float64,Variable}) at /Users/idunning/.julia/v0.3/JuMP/src/operators.jl:266
is ambiguous with:
    *(T<:GenericQuadExpr{Float64,Variable},S<:Union(Variable,GenericAffExpr{Float64,Variable},GenericQuadExpr{Float64,Variable})) at /Users/idunning/.julia/v0.3/JuMP/src/operators.jl:264.
To fix, define
    *(T<:GenericQuadExpr{Float64,Variable},T<:GenericQuadExpr{Float64,Variable})
before the new definition.
```

IainNZ:
Stochastic though - JuMPeR always involed.

mlubin:
Well it seems like the warning is correct.

IainNZ:
I just got this in something not involving JuMPeR at all, btw. Am I the only one who ever sees it?!

yeesian:
I've seen it before, and just saw it again:
```
Warning: New definition 
    *(S<:Union(Variable,GenericQuadExpr{Float64,Variable},GenericAffExpr{Float64,Variable}),T<:GenericQuadExpr{Float64,Variable}) at /Users/yeesian/.julia/v0.3/JuMP/src/operators.jl:275
is ambiguous with: 
    *(T<:GenericQuadExpr{Float64,Variable},S<:Union(Variable,GenericQuadExpr{Float64,Variable},GenericAffExpr{Float64,Variable})) at /Users/yeesian/.julia/v0.3/JuMP/src/operators.jl:273.
To fix, define 
[...]

PR #438, opened 2015-05-05
fix ambiguity warnings for * in quadexpr

yeesian:
See #383.

Just a shot in the dark.. by following the messages, since I can't reproduce the warnings. Do I just make it error out here?

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -272,6 +272,8 @@ end
 
 *{T<:QuadExpr,S<:Union(Variable,AffExpr,QuadExpr)}(::T,::S) =
     error( "*(::$T,::$S) is not defined. $op_hint")
+(*)(lhs::QuadExpr, rhs::QuadExpr) =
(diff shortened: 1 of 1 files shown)


src/operators.jl
```diff
@@ -272,6 +272,8 @@ end
 
 *{T<:QuadExpr,S<:Union(Variable,AffExpr,QuadExpr)}(::T,::S) =
     error( "*(::$T,::$S) is not defined. $op_hint")
+(*)(lhs::GenericQuadExpr{Float64,Variable}, rhs::GenericQuadExpr{Float64,Variable}) =
```
mlubin: You can use the ``QuadExpr`` alias instead of ``GenericQuadExpr{Float64,Variable}``

mlubin:
LGTM, could you squash the commits?

joehuchette:
:+1:

Issue #442, opened 2015-05-13
Accessing solver stats (node count)


sbebo:
I was trying to extract the number of nodes at the end of a branch-and-bound. It looks like there's no easy way to do it in JuMP? 

I tried to look at MathProgBase, and it looks like `getnodecount` should do the trick. But `getnodecount(m.internalModel)` doesn't work for Cplex or Gurobi. I had to use `CPLEX.getnodecnt`, or even `CPLEX.get_node_count` and `Gurobi.get_node_count` passing the "raw model". 

Other MathProgBase methods work fine, like `getobjval` or `getobjbound`; other ones, like `getobjgap`, don't. I guess it's just a (boring) matter of implementing the missing method(s) in the solver interfaces. But what I mean is: if I'm using JuMP, I'd rather not use MathProgBase at all, if possible. It would be nice to expose at least a subset (if not all?) of these methods in JuMP.

mlubin:
No reason this hasn't been implemented other than nobody asking for it yet.

The two quick steps are:
1. Implement ``MathProgBase.getnodecount`` in Gurobi.jl and CPLEX.jl
2. Implement ``MathProgBase.getnodecount(::JuMP.Model)`` (like in #422).

We'd gladly accept a PR for this.

sbebo:
Cool, will do.

mlubin:
Fixed by #444

PR #444, opened 2015-05-14
Expose getnodecount #442

sbebo:
Wrapper for  `MathProgBase.getnodecount `.


joehuchette:
Great, thanks for this. Looks good to me.

joehuchette:
A sneaky way to handle a lot of the remaining ``MathProgBase.get*(::JuMP.Model)`` methods at once would be to use an ``@eval`` loop, since the body of of the functions will end up looking much the same (like here for ``getnodecount`` and ``getsolvetime``).

mlubin:
@joehuchette, I was just thinking that as well. We can do that after merging though.

joehuchette:
Thanks @sbebo!

PR #445, opened 2015-05-17
implement defVars macro

mlubin:
Missing docs. Let me know what you guys think about the ``->`` syntax. I think the end result is pretty nice.
@davidanthoff

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -39,7 +39,7 @@ export
     @addConstraint, @addConstraints,
     @LinearConstraint, @LinearConstraints, @QuadConstraint, @QuadConstraints,
     @defVar, @defConstrRef, @setObjective, addToExpression, @defExpr,
-    @setNLObjective, @addNLConstraint, @addNLConstraints,
+    @setNLObjective, @addNLConstraint, @addNLConstraints, @defVars,
     @defNLExpr
 
 include("JuMPDict.jl")
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -427,35 +427,36 @@ for (mac,sym) in [(:LinearConstraints, symbol("@LinearConstraint")),
 end
 
 for (mac,sym) in [(:addConstraints,  symbol("@addConstraint")),
-                  (:addNLConstraints,symbol("@addNLConstraint"))]
+                  (:addNLConstraints,symbol("@addNLConstraint")),
+                  (:defVars, symbol("@defVar"))]
     @eval begin
         macro $mac(m, x)
             x.head == :block || error("Invalid syntax for @$mac")
             @assert x.args[1].head == :line
             code = quote end
             for it in x.args
-                if it.head == :line
+                if isexpr(it, :line)
                     # do nothing
-                elseif it.head == :comparison # regular constraint
-                    mac = Expr(:macrocall,$(quot(sym)), esc(m), esc(it))
-                    code = quote
-                            $code
-                            $mac
-                            end
-                elseif it.head == :tuple # constraint ref or kwargs
+                elseif isexpr(it, :tuple) # line with commas
                     for i in 1:length(it.args)
-                        if isexpr(it.args[i], :(=))
-                            it.args[i] = Expr(:kw, it.args[i].args[1], esc(it.args[i].args[2]))
+                        if isexpr(it.args[i], :(->))
+                            kwname = it.args[i].args[1]
+                            @assert isexpr(it.args[i].args[2], :block)
+                            @assert length(it.args[i].args[2].args) == 2
+                            @assert isexpr(it.args[i].args[2].args[1], :line)
+                            kwval = it.args[i].args[2].args[2]
+                            it.args[i] = Expr(:kw, kwname, esc(kwval))
                         else
                             it.args[i] = esc(it.args[i])
                         end
                     end
-                    mac = Expr(:macrocall,$(quot(sym)), esc(m), it.args...)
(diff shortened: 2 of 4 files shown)


IainNZ:
So its only `->` for `@defVars`?

joehuchette:
Yeah, I'm not sure I like using ``->``, but only in this one macro.

mlubin:
``@defVars`` and [parameters to constraints](https://github.com/mlubin/JuMPChance.jl/blob/090efed3829f1aea90ec4c92dae3dd8d873a4b18/test/runtests.jl#L297). We don't have many choices for the syntax, ``=`` just doesn't parse like we would want it to.

mlubin:
We can allow ``->`` in ``@defVar`` for consistency

yeesian:
I'm confused why it is `->` instead of `<-` (if I understood it to be like R's assignment syntax)

mlubin:
@yeesian, ``x <- y`` is parsed as ``x < -y``, so that wouldn't be an option. Agreed that it makes more sense though.

IainNZ:
For me, the utility isn't outweighing the extra syntax/mental overhead...

mlubin:
I guess we could try to manually override the precedence of ``=``

joehuchette:
This also doesn't support the column-generation stuff (which would make working around the ``=`` parsing trickier).

mlubin:
This does support column generation:
```
@defVars m begin
    0 <= z <= 5, objective->100.0, inconstraints->[maincon], coefficients->[1.0]
end
```

joehuchette:
Ahh whoops, it just passes it along to ``@defVar`` I guess. It's worth taking a stab at working around the parsing, but things like this might get tricky:
```jl
julia> a = quote
       x[1:3], start = 3, coeffs=3
       end
quote  # none, line 2:
    (x[1:3],start) = (3,coeffs) = 3
end
```

mlubin:
Well if it's possible to build a 1-1 mapping to undo the precedence of ``=``, it wouldn't be terrible to have the 20 lines of code to do so.
@JeffBezanson, here's an issue for us DSL builders

davidanthoff:
Love the general utility of this. Would be great if ``=`` could be used instead of ``->``.

mlubin:
What do we think about:
```
@defVars m begin
    0 ≤ x ≤ 1
    y ≥ 2; Int
    z ≤ 3; start = 10
    q; Bin; start = 0.5
end
```
We can keep the `=` for parameters if we use a semicolon as a separator. (Suggested by @StefanKarpinski)

joehuchette:
Still not crazy about it, but it's the best proposal so far. I think it'd be better if we only use semicolons for "keyword arguments" like ``start``:
```jl
@defVars(m, begin
    0 ≤ x ≤ 1
    y ≥ 2, Int
    z ≤ 3; start = 10
    q, Bin; start = 0.5
```

mlubin:
Having more than one type of separator is pretty confusing

joehuchette:
Using commas in ``@addConstraint``, ``@defVar``, etc. and semicolons in ``@defVars`` is even more confusing. Mixing like this also matches with the distinction between regular/keyword arguments in julia syntax.

mlubin:
We could switch to all semicolons in ``@addConstraint`` and ``@defVar``.

> Mixing like this also matches with the distinction between regular/keyword arguments in julia syntax.

You can go pretty far as a julia user without understanding this distinction, it only shows up when you're defining a function that takes in keyword arguments.

joehuchette:
Another not totally satisfying solution that will work right now:
```jl
@defVars(m, begin
    0 ≤ x ≤ 1
    y ≥ 2, Int
    z ≤ 3, (start = 10)
    q, Bin, (start = 0.5, col)
end)
```

mlubin:
The parentheses don't look too bad to me, but the use should be consistent for all modifiers, e.g.,
```
@defVars(m, begin
    0 ≤ x ≤ 1
    y ≥ 2, (Int)
    z ≤ 3, (start = 10)
    q, (Bin, start = 0.5, col)
end)
```

The parsed version of
```
y ≥ 2, (Int)
```
is the same as
```
y ≥ 2, Int
```
so there's no much we can do in that case.

However, there's a difference between
```
q, Bin, (start = 0.5, col)
```
and
```
q, (Bin, start = 0.5, col)
```
and I prefer the latter.

mlubin:
Comments on the parentheses syntax?
@davidanthoff @yeesian @chriscoey

yeesian:
I actually find the parentheses syntax quite pleasing (given the constraints in options available at the moment). I also prefer having the variable type in the parens

cc @JackDunnNZ

yeesian:
One question: will the ordering of the arguments in the parentheses matter? (i.e. `vartype` comes before everything else, etc)

I guess the concern is whether there's a good way to distinguish between what goes into the parens and what doesn't. If we go by "non-parens for regular, parens for keyword args", it might lead to expectations that the arguments in the parens should be unordered.

mlubin:
The ordering of the arguments inside the parentheses should not matter.

davidanthoff:
I guess it is fine. Not ideal, but I also don't have a better idea :)

mlubin:
Seems like a local optimum. I like it because it's easy to explain (put anything except the variable declaration in parentheses) and easy to parse.

mlubin:
See https://github.com/JuliaOpt/JuMP.jl/pull/688 for the new implementation with parentheses as discussed

PR #446, opened 2015-05-17
implement dictionary of variable names and getVar

mlubin:
Needs docs. Doesn't seem like there should be any performance penalty, but worth double checking.
@davidanthoff

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -32,6 +32,7 @@ export
     # Variable
     setName, getName, setLower, setUpper, getLower, getUpper,
     getValue, setValue, getDual, setCategory, getCategory,
+    getVar,
     # Expressions and constraints
     affToStr, quadToStr, conToStr, chgConstrRHS,
 
@@ -95,6 +96,8 @@ type Model
 
     nlpdata#::NLPData
 
+    varDict::Dict{Symbol,Any} # dictionary from variable names to variable objects
+
     # Extension dictionary - e.g. for robust
     # Extensions should define a type to hold information particular to
     # their functionality, and store an instance of the type in this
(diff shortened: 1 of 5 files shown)


IainNZ:
Seems nice and simple!

src/JuMP.jl
```diff
@@ -730,6 +733,28 @@ function Variable(m::Model,lower::Number,upper::Number,cat::Symbol,objcoef::Numb
...
+# handle dictionary of variables
+function registervar(m::Model, varname::Symbol, value)
+    if haskey(m.varDict, varname)
+        m.varDict[varname] = nothing # indicate duplicate variable
+    else
+        m.varDict[varname] = value
+    end
+    return value
+end
+registervar(m::Model, varname, value) = value # variable name isn't a simple symbol, ignore
```
joehuchette: When would this get called?

joehuchette:
:+1:

test/variable.jl
```diff
@@ -27,6 +27,15 @@ facts("[variable] constructors") do
...
     @fact getUpper(x[-4,"Green"]) => 5.5
     @fact getValue(x[-3,"Blue"]) => -2
+    @fact isequal(getVar(mcon, :lbonly),lbonly) => true
+    @fact isequal(getVar(mcon, :ubonly),ubonly) => true
+    @fact isequal(getVar(mcon, :onerangeub)[-7],onerangeub[-7]) => true
+    @defVar(mcon, lbonly)
+    @fact_throws ErrorException getVar(mcon, :lbonly)
+    @fact_throws ErrorException getVar(mcon, :foo)
+    d = Dict()
+    @defVar(mcon, d["bar"][1:10] == 1)
```
mlubin: @joehuchette, it would get called here

joehuchette: Hmm. Maybe instead of symbols you could use strings, in which case this would work? Not sure it's worth it as this is kind of a corner case.

mlubin: I don't think there's much of a need to support that in the first pass

joehuchette: Agreed, we could always add it in without breaking anything later if there's a need.

davidanthoff:
Looks perfect, this will help us a lot :)

mlubin:
Updated ``NEWS`` and added a line in the docs.

tkelman:
This looks pretty handy indeed.

PR #447, opened 2015-05-17
Add AmplNLWriter to nonlinear tests

JackDunnNZ:
After JuliaOpt/CoinOptServices.jl@7771b4dee8b356f93a50feb148d33eba08c38292 by @tkelman we can use the solvers from the osil tests in the NL writer as well. (Ipopt will be possible soon too after JuliaOpt/Ipopt.jl#35)

Unfortunately Couenne doesn't seem to turn off its output even when the right args are passed to it, so it spews a bit of output.


joehuchette:
:+1:

PR #451, opened 2015-05-23
Fix 0.4-dev warnings

yuyichao:
This fixes certain deprecation warnings on `0.4`.

Also use `@__FILE__` instead of `Pkg.dir` to find current dir so that the tests can be run in any directories.


mlubin:
LGTM, thanks!

PR #455, opened 2015-05-25
fix #454

### LINKED ISSUE #454: @defExpr weirdness
On 0.4:

```
julia> using JuMP

julia> m = Model();

julia> @defVar(m,x[1:2]);

julia> @defExpr(y[i=1:2], sum{x[i]; i==1})
ERROR: Three argument form of @defExpr only supports linear expressions
 in anonymous at /home/mlubin/.julia/v0.4/JuMP/src/macros.jl:514
```

The two issues with this are:
- There is no three-argument form of ``@defExpr``. We mean the two-argument form with a nonscalar expression variable.
- The check for ``isa($newaff,AffExpr)`` should be replaced with a check for if the result can be converted to an AffExpr.

mlubin:


mlubin:
Seems to pass except for an unrelated SCS issue (JuliaOpt/SCS.jl#26)

PR #457, opened 2015-05-27
Warn only once when getValue(x) doesn't work

joehuchette:
I was tired of getting pages of warnings that filled up the buffer on my terminal.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -278,11 +278,15 @@ function setValue(v::Variable, val::Number)
     end
 end
 
+# internal method that doesn't print a warning if the value is NaN
+_getValue(v::Variable) = v.m.colVal[v.col]
+
 function getValue(v::Variable)
-    if isnan(v.m.colVal[v.col])
-        warn("Variable $(getName(v))'s value not defined. Check that the model was properly solved.")
+    ret = _getValue(v)
+    if isnan(ret)
+        Base.warn("Variable value not defined. Check that the model was properly solved.")
     end
-    return v.m.colVal[v.col]
+    ret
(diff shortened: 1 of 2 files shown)


mlubin:
This has been annoying for me too. You still might want to keep the variable name in there for concreteness.

joehuchette:
I'm not sure if it'll work then; looks like ``warn_once`` [caches the string](https://github.com/JuliaLang/julia/blob/c18f18d14aacadac588e92a76226e5145448791b/base/util.jl#L361-L371).

mlubin:
Oh. I'd add that the message will not repeat then, so that lack of a message doesn't imply that the problem was solved correctly

joehuchette:
Think I like this implementation better.

IainNZ:
Quite a nice little patch

joehuchette:
Looks like SCS is broken on OS X nightlies:
```
while loading /Users/travis/build/JuliaOpt/JuMP.jl/test/model.jl, in expression starting on line 179
     - With solver SCS.SCSSolver
dyld: lazy symbol binding failed: Symbol not found: _validateLinSys
  Referenced from: /Users/travis/.julia/v0.4/Homebrew/deps/usr/lib/libscsdir64.dylib
  Expected in: flat namespace
dyld: Symbol not found: _validateLinSys
  Referenced from: /Users/travis/.julia/v0.4/Homebrew/deps/usr/lib/libscsdir64.dylib
  Expected in: flat namespace
/Users/travis/build.sh: line 41:  4213 Trace/BPT trap: 5       julia test/runtests.jl

```

PR #458, opened 2015-05-28
Update docs

odow:
Not sure if you want to include this or not.

Highlight the fact that there is no bounds checking of @defNLExpr
See https://groups.google.com/forum/#!topic/julia-opt/_ucw349fpIk


mlubin:
It's fair to put this in the docs. I'll push a fix with slightly different wording.

PR #459, opened 2015-06-01
Make dot more generic (and safer)

joehuchette:
Also adds a missing method so that e.g. ``+(::OneIndexedArray, ::OneIndexedArray)`` works as expected.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -228,6 +228,8 @@ end
 #  - dot
 #############################################################################
 
+typealias JuMPTypes Union(Variable,AffExpr,QuadExpr)
+
 Base.sum(j::JuMPArray) = sum(j.innerArray)
 Base.sum(j::JuMPDict)  = sum(values(j.tupledict))
 Base.sum(j::JuMPArray{Variable}) = AffExpr(vec(j.innerArray), ones(length(j.innerArray)), 0.0)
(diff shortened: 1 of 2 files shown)


joehuchette:
SCS :'(

src/operators.jl
```diff
@@ -241,26 +243,25 @@ function Base.sum{S,T}(affs::Array{GenericAffExpr{S,T}})
...
-    size(lhs.innerArray) == size(rhs.innerArray) || error("Incompatible dimensions")
-    return QuadExpr(vec(lhs.innerArray), vec(rhs.innerArray), ones(length(lhs.innerArray)), AffExpr())
+Base.dot{T,S,N}(lhs::OneIndexedArray{T,N},rhs::OneIndexedArray{S,N}) = _dot(lhs.innerArray, rhs.innerArray)
+Base.dot{T,S,N}(lhs::OneIndexedArray{T,N},rhs::Array{S,N}) = _dot(lhs.innerArray, rhs)
+Base.dot{T,S,N}(lhs::Array{T,N},rhs::OneIndexedArray{S,N}) = _dot(lhs, rhs.innerArray)
+Base.dot{T<:JuMPTypes,S,N}(lhs::Array{T,N},rhs::Array{S,N}) = _dot(lhs,rhs)
+Base.dot{T<:JuMPTypes,S<:JuMPTypes,N}(lhs::Array{T,N},rhs::Array{S,N}) = _dot(lhs,rhs)
+Base.dot{T,S<:JuMPTypes,N}(lhs::Array{T,N},rhs::Array{S,N}) = _dot(lhs,rhs)
+
+function _dot{T,S}(lhs::Array{T}, rhs::Array{S})
```
mlubin: This is slow in every case, can't we do better? At least ``dot(Vector{Float64},OneIndexedArray{Variable})`` should be very fast.

joehuchette: I'd rather just get rid of multidimensional dot, to be honest, considering
it's not in Base.

tkelman: there is now a `vecdot` in 0.4

joehuchette: Cool, I thought I remembered something about to be added. I'll change the naming then.

PR #461, opened 2015-06-03
dot -> vecdot for multidimensional arrays

joehuchette:
@mlubin

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -243,6 +243,13 @@ function Base.sum{S,T}(affs::Array{GenericAffExpr{S,T}})
     return new_aff
 end
 
+if VERSION < v"0.4-"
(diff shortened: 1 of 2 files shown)


mlubin:
Leave in a deprecation warning for this use of ``dot``?

mlubin:
Is vecdot part of Base in 0.3?

joehuchette:
No, I thought for some reason that adding a method like this would work. Is it too tacky to define it ourselves with something like
```jl
eval(Base,:(vecdot(x,y) = ...))
```

mlubin:
Just do ``import Base.vecdot`` on 0.4 and ``export vecdot`` on 0.3.

mlubin:
What about the deprecation?

PR #464, opened 2015-06-14
Added examples based on problems in the AMPL book

lluang:
JuMP versions of models described in:

   AMPL: A Modeling Language for Mathematical Programming, 2nd ed
   by Robert Fourer, David Gay, and Brian W. Kernighan

### DIFF
--- a/examples/multi.jl
+++ b/examples/multi.jl
@@ -1,4 +1,9 @@
 #############################################################################
+#  Copyright 2015, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
 # Author: Louis Luangkesorn
 # Date: February 26, 2015
 #
--- a/examples/multi.jl
+++ b/examples/multi.jl
@@ -1,0 +1,107 @@
+#############################################################################
+# Author: Louis Luangkesorn
+# Date: February 26, 2015
+#
+# JuMP implementation of the multicommodity transportation model
+#   AMPL: A Modeling Language for Mathematical Programming, 2nd ed
+#   by Robert Fourer, David Gay, and Brian W. Kernighan
+#   4-1 multi.mod and multi.dat
+#############################################################################
+
+function PrintSolution(status, Trans, ORIG, DEST, PROD)
+    println("RESULTS:")
+    if status == :Optimal
+      for i = 1:length(ORIG)
+        for j = 1:length(DEST)
+          for p = 1:length(PROD)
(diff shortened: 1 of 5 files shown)


joehuchette:
Wow, this is fantastic. Thanks @lluang!

Hopefully there aren't any licensing issues (@mlubin?).

mlubin:
Very cool! Unfortunately there are copyright issues. A model translated from AMPL could be considered a derived work, so I'm cautious to include these models with the JuMP distribution. @lluang, perhaps you could contact Bob Fourer to ask for his opinion/permission to distribute the translated models? Another option is to put these examples in the [juliaopt-notebooks](https://github.com/JuliaOpt/juliaopt-notebooks) collection.

joehuchette:
How is this substantially different from translating a model from e.g. an academic paper?

IainNZ:
I feel like we've had this whole discussion before. My position was that I don't think you can copyright a mathematical model.

mlubin:
You can copyright an implementation of a model, e.g., an ampl ``mod`` file, and any direct translation of that model could be considered a derivative work. IANAL though.

lluang:
I'm asking Bob Fourer on his thoughts. Note that there really is no such thing as a direct translation of a model, because the languages are different. There is correspondence, but only because both JuMP and AMPL representations are of the same model (i.e. constraints, variables, objective, etc.)

lluang:
From Bob Fourer (below)

How do you want to do this (the files in question have attribution to either the original AMPL journal article or the AMPL book bibliographic information included in the title block, so that is taken care of.

Louis

" 
Hi Louis,

I am not sure whether permission is needed to distribute conversions of published model examples into a different modeling language. But in any case I am willing to arrange permission for your JuMP versions, especially since a reference to the original source is being given.

-- Bob
﻿"

joehuchette:
Is this sufficient @mlubin?

mlubin:
Looks good to me. @lluang, could you add an MPL header to these files? Something like:
```
#  Copyright 2015, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
#  This Source Code Form is subject to the terms of the Mozilla Public
#  License, v. 2.0. If a copy of the MPL was not distributed with this
```
You can place your name in the copyright line if you like.

lluang:
Done!

examples/multi.jl
```diff
@@ -0,0 +1,112 @@
...
+
+function PrintSolution(status, Trans, ORIG, DEST, PROD)
+    println("RESULTS:")
+    if status == :Optimal
+      for i = 1:length(ORIG)
+        for j = 1:length(DEST)
+          for p = 1:length(PROD)
+            print(" $(PROD[p]) $(ORIG[i]) $(DEST[j]) = $(getValue(Trans[i,j, p])) \t")
+          end
+        println()
```
mlubin: It looks like the indentation is off on this line.

joehuchette:
Thanks again @lluang!

Issue #305, opened 2014-11-13
Non-range indexing


yeesian:
I was looking around for a way to express

```
sum(e in E) x[e] <= 5
```

succinctly, and wasn't aware of the following:

``` julia
@defVar(m, x[E], Bin)
@addConstraint(m, sum{x[e], e in E} <= 5}
```

Going through [the documentation](http://jump.readthedocs.org/en/release-0.6/refexpr.html), I got the sense that only range indices (e.g. `x[i:j]` or `x[i:j:k]`) are supported.

mlubin:
I consider this resolved by the generator syntax. If you know julia syntax it should be obvious that you can use any iterable set.

PR #475, opened 2015-07-07
Clarify use of iterables in sum{} in quickstart

iamed2:
See https://groups.google.com/forum/#!topic/julia-opt/UGPkMD1xpqc

I first assumed that these also were restricted ranges of the form `1:N`. Updated to clarify they can be any iterable.


yeesian:
Thanks for this! (Y) Will help close #305, which I didn't get around to

IainNZ:
Nice. Yeah the docs are... not great.

Issue #468, opened 2015-06-25
MPB indices


mlubin:
We don't currently provide any well-defined way to map from MPB variable indices into JuMP variables. This seems to be a big gap in the abstraction. Look at the IIS example:

```
    println("Irreducible Inconsistent Subsystem (IIS)")
    println("Variable bounds:")
    for i in 1:numvar
        v = Variable(m, i)
        if iislb[i] != 0 && iisub[i] != 0
            println(getLower(v), " <= ", getName(v), " <= ", getUpper(v))
        elseif iislb[i] != 0
            println(getName(v), " >= ", getLower(v))
        elseif iisub[i] != 0
            println(getName(v), " <= ", getUpper(v))
        end
    end
```

What about exporting a function which takes a model and MPB variable index and returns a corresponding JuMP variable?

mlubin:
(Title changed since I changed my mind halfway through writing up the issue.)

mlubin:
What if we guarantee and document that ``Variable(m, idx)`` is valid when ``idx`` is an MPB index? If we ever need to use a nontrivial map, we can make the appropriate internal changes to keep ``Variable(m,idx)`` working.

mlubin:
Same for ``ConstraintRef{LinearConstraint}`` also

mlubin:
I think this is resolved by #479 and https://github.com/JuliaOpt/JuMP.jl/commit/794c91157de5d4b2917ce5a27b6a1a264f1f9d3a

PR #479, opened 2015-07-10
implement getLinearIndex

mlubin:
This fills a significant hole in the abstraction. Ref https://github.com/JuliaOpt/juliaopt-notebooks/pull/11  and #468. We can bikeshed the name.
CC @rgiordan

### DIFF
--- a/doc/refvariable.rst
+++ b/doc/refvariable.rst
@@ -95,7 +95,7 @@ This form of constructing variables is not considered idiomatic JuMP code.
     The reference to the first set of variables has been lost, although they will remain
     in the model.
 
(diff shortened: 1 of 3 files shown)


mlubin:
@joehuchette, same failure in SCS. Any chance you could extract the problem in conic form and add it to  the [SDP tests](https://github.com/JuliaOpt/MathProgBase.jl/blob/9414e08725c72edeaf538bffafcb448903bb8be3/test/conicinterface.jl#L333)? May be easier to debug from there.

joehuchette:
ac78ec11a3e4f412f7c7662def84b231f74dcd5a will hopefully help identify which instance is failing with SCS.

PR #480, opened 2015-07-14
warnings for some common performance traps

mlubin:
This covers some of the common cases but not everything. I think we should wait for user feedback before expanding the scope of performance warnings like this.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -101,6 +101,9 @@ type Model
 
     varDict::Dict{Symbol,Any} # dictionary from variable names to variable objects
 
+    getvalue_counter::Int # number of times we call getValue on a JuMPContainer, so that we can print out a warning
+    operator_counter::Int # number of times we add large expressions
+
     # Extension dictionary - e.g. for robust
     # Extensions should define a type to hold information particular to
     # their functionality, and store an instance of the type in this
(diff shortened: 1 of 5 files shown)


mlubin:
Added void returns to avoid unnecessary allocations, otherwise don't see any performance regressions

joehuchette:
This doesn't play nicely with the vectorized stuff, which uses [this fallback](https://github.com/JuliaOpt/JuMP.jl/blob/794c91157de5d4b2917ce5a27b6a1a264f1f9d3a/src/v0.4/parseExpr_staged.jl#L180).

```jl
using JuMP

A = rand(250,250)
B = rand(250,250)
b = rand(250)

m = Model()
@defVar(m, x[1:250])

@addConstraint(m, A*x + B*x .<= b) # gives warning
```

mlubin:
Hmm, any suggestions on how to improve the warning? Bump the threshold for ``operator_counter`` to 10000?

joehuchette:
Actually I think the real problem should be in the ``_multiply!`` functions this calls, let me play around with it.

joehuchette:
But yeah, bumping it up seems reasonable as well.

joehuchette:
This particular instance is fixed with dac340296a354b7b65071499cd8debfefbd6bbea

mlubin:
At least the warnings help us find cases where we're doing too much operator overloading.

joehuchette:
The example above still gives warnings on v0.3

joehuchette:
(because ``@addConstraint`` generates different code on v0.3 and v0.4)

mlubin:
Hmm. We could disable the performance warnings on 0.3.

joehuchette:
I'm gonna see if I can do something clever here, but otherwise yeah, let's do that and bump up the counter too.

PR #482, opened 2015-07-15
WIP: rewrite .travis.yml for new docker infrastructure

mlubin:
CC @tkelman

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,30 +1,25 @@
-language: cpp
-compiler:
-    - gcc
+language: julia
 os:
     - linux
     - osx
+julia:
+    - release
+    - nightly
 notifications:
     email: false
-env:
-    matrix:
-        - PKGADD="FactCheck;Cbc;Clp;GLPKMathProgInterface;Ipopt;ECOS;SCS" JULIAVERSION="juliareleases"
-        - PKGADD="FactCheck;GLPKMathProgInterface;ECOS" JULIAVERSION="julianightlies"
-before_install:
-    - curl http://julialang.org/install-julia.sh | sh -s $JULIAVERSION
-    - if [ `uname` = "Linux" ]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,29 +1,23 @@
...
+        - libglpk-dev
 script:
-    - julia -e 'Pkg.init(); Pkg.clone(pwd())'
-    - if [[ `uname` = "Darwin" ]]; then julia -e 'Pkg.add("Ipopt")'; fi # installs from binaries
-    - if [[ `uname` = "Linux" ]]; then julia -e 'Pkg.add("SCS")'; fi # hack to avoid broken SCS on OSX nightly
-    - julia -e 'for p in split(ENV["PKGADD"],";") Pkg.add(p) end'
-    - if [ $JULIAVERSION = "julianightlies" ]; then julia test/runtests.jl; fi
-    - if [ $JULIAVERSION = "juliareleases" ]; then julia --code-coverage test/runtests.jl; fi
+    - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+    - julia -e 'Pkg.clone(pwd())'
```
tkelman: you'll need to `Pkg.build` I suspect, also check CoinOptServices for how I do dep caching - I was supposed to write a julia-users post about that but completely forgot about it

tkelman: nevermind on `Pkg.build`, jump doesn't have a `deps/build.jl` file so it isn't needed

mlubin: How does dep caching deal with the situation where we bump, e.g., Cbc, and Bindeps doesn't decide to build a new version because an old version is present? Can we easily manually clear the cache?

tkelman: Should be a button to clear the cache under settings, caches, though I'm not seeing one right now, maybe after a month or so they auto-clear it

mlubin:
Seems like a strict improvements, caching can be done later.

PR #489, opened 2015-07-25
add parens around tests for `in`

tkelman:
precedence of `=>` changed in https://github.com/JuliaLang/julia/pull/12285

### DIFF
--- a/test/expr.jl
+++ b/test/expr.jl
@@ -79,8 +79,8 @@ facts("[expr] Test in(::Variable, ::AffExpr)") do
     m = Model()
     @defVar(m, x[1:3])
     @defVar(m, y)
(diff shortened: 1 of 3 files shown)


mlubin:
Now scs isn't downloading, guess we need to tag that too. Sourceforge seems pretty useless these days

tkelman:
Yeah, very much so. I haven't been very good about tagging but tried to at least fix master for the majority of the packages that I've regrettably inflicted sourceforge on.

mlubin:
I think we're back to just the normal issues with Ipopt broken on OS X and SCS giving the wrong answers...

mlubin:
Thanks!

Issue #490, opened 2015-07-26
Copying a Model and adding new constraints generates an error when calling solve(m)


h3nnn4n:
[Here](https://gist.github.com/h3nnn4n/fb3dee142158b7ab9229) I have a small snippet of code that reproduces an error on a bigger piece of code. 

The code is:

```julia
using JuMP

## Simple LP model
m = Model()

@defVar(m, 0 <= x <= 2 )
@defVar(m, 0 <= y <= 30 )

@setObjective(m, Max, 5x + 3*y )
@addConstraint(m, 1x + 5y <= 3.0 )

## The model is copied
n = copy(m)

## New constraint using the same variables is added to the copy
@addConstraint(n, 2x + 3y <= 3.0 )

solve(m)

## Error Here
solve(n)
```

and the error:

```bash
h3nnn4n@clusterfuck ~/julia_bnb (git)-[master] % julia jump_copy_error.jl
ERROR: Variable not owned by model present in constraints
 in prepConstrMatrix at /home/h3nnn4n/.julia/v0.3/JuMP/src/solvers.jl:163
 in solveLP at /home/h3nnn4n/.julia/v0.3/JuMP/src/solvers.jl:221
 in solve at /home/h3nnn4n/.julia/v0.3/JuMP/src/solvers.jl:39
 in include at ./boot.jl:245
 in include_from_node1 at loading.jl:128
 in process_options at ./client.jl:285
 in _start at ./client.jl:354
while loading /home/h3nnn4n/julia_bnb/bug.jl, in expression starting on line 21
```

The error happens on both Julia 0.3 and 0.4 using the following packages:
```julia
julia> Pkg.status()
6 required packages:
 - GLPK                          0.2.16
 - GLPKMathProgInterface         0.1.13
 - JuMP                          0.9.2
```

IainNZ:
The error message is correct, the variables belong to the initial model, not the copy.
I guess what you want is a way to get a handle on the equivalent variables for the copy?

h3nnn4n:
Yes. 

In the code I am working on I have
```julia
type instance
    m :: Model
    weight :: Array{Int}
    profit :: Array{Int}
    size :: Int
    capacity :: Int
    x
end
```
When I copy the whole thing the model `m` keeps the reference to the original `instance.x`. [Here](https://gist.github.com/h3nnn4n/1e9f040841c4443d9644) is the whole code in case you want to take a closer look.

h3nnn4n:
[...]

PR #491, opened 2015-07-26
Fixes for copy(::Model)

joehuchette:
ref #490

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -203,9 +203,31 @@ function Base.copy(source::Model)
     # Variables
     dest.numCols = source.numCols
     dest.colNames = source.colNames[:]
+    dest.colNamesIJulia = source.colNamesIJulia[:]
     dest.colLower = source.colLower[:]
     dest.colUpper = source.colUpper[:]
     dest.colCat = source.colCat[:]
+
+    # callbacks and hooks
+    if !isempty(dest.callbacks)
+        error("Copying callbacks is not supported")
+    end
+    if source.solvehook != nothing
+        dest.solvehook = copy(source.solvehook)
+    end
+    if source.printhook != nothing
+        dest.printhook = copy(source.printhook)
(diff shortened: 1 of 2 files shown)


src/JuMP.jl
```diff
@@ -214,6 +230,19 @@ function Base.copy(source::Model)
     return dest
 end
 
+# Copy methods that I'm not sure where else to put
```
IainNZ: I'd just put them in with the Variable section

IainNZ:
Related style of code I have in JuMPeR:
https://github.com/IainNZ/JuMPeR.jl/blob/aa24323313ba0db9d238419fe9fe4f318c9c91bd/src/solve.jl#L11-L24

Kinda funny that we had the variable copy-to-new-model code in JuMP but not this PR.

joehuchette:
I'm kind of tempted to turn of OS X travis until we get Homebrew figured out...

src/JuMP.jl
```diff
@@ -203,10 +203,26 @@ function Base.copy(source::Model)
     # Variables
     dest.numCols = source.numCols
     dest.colNames = source.colNames[:]
+    dest.colNamesIJulia = source.colNamesIJulia[:]
     dest.colLower = source.colLower[:]
     dest.colUpper = source.colUpper[:]
     dest.colCat = source.colCat[:]
 
+    # callbacks and hooks
+    dest.callbacks = copy(source.callbacks)
```
mlubin: Is it safe to copy callbacks? They refer to variables from the old model.

joehuchette: Since callbacks are wrapped in types, not plain functions, I think this will result in a no method error, except if a user explicitly defines a copy method for their custom callback.

joehuchette: Err, guess not. I'll just make this an error if callbacks are present.

PR #492, opened 2015-07-28
fix uninitialized b vector

mlubin:
@joehuchette, this seems to fix the issues with SCS. The components of the ``b`` vector for SOC constraints were garbage.


joehuchette:
Oh man

joehuchette:
Good catch. Maybe it's safe to just do ``b = zeros(numRows)`` as well?

mlubin:
Well, zero is just as much garbage if it's not what's intended. I found this by printing out the vectors, in which case it's useful to quickly see that there are uninitialized values.

PR #493, opened 2015-07-28
[RFC] Smarter vectorization in @addConstraint

joehuchette:


joehuchette:
Fixes the operator overloading warning on v0.3 for
```jl
using JuMP
N = 110

A = rand(N,N);
B = rand(N,N);
C = rand(N,N);

MyModel = Model()

@defVar(MyModel, x[1:N,1:N] >= 0)
@defVar(MyModel, y[1:N,1:N] >= 0)
@defVar(MyModel, z[1:N,1:N] >= 0)


@addConstraint(MyModel, ones(N,N) .>= A*x + B*y)
```

src/v0.3/parseExpr_0.3.jl
```diff
@@ -132,26 +132,67 @@ function addToExpression{C,V}(quad::GenericQuadExpr{C,V},c::Number,x::GenericQua
     quad
 end
 
-function addToExpression{N}(ex::Array{AffExpr,N}, c::Number, x::Array{AffExpr,N})
-    size(ex) == size(x) || error("Incompatible sizes: $(size(ex)) + $(size(x))")
+function chkdims(x,y)
+    (ndims(x) ≤ 2 &&
```
mlubin: Why is this restricted to two dimensions?

joehuchette: I didn't want to touch tensor multiplication stuff, but I guess stuff like ``@addConstraint(m, X + Y .== Z)`` should be allowed.

mlubin: Okay, let's leave that as a lazy feature

test/runtests.jl
```diff
@@ -13,6 +13,7 @@
 using JuMP
 using Compat
 using FactCheck
+using Base.Test
```
mlubin: Extra line here?

mlubin:
Are all of these methods covered by tests?

test/runtests.jl
joehuchette: ``@test_approx_eq`` is used somewhere in the nonlinear tests, and was failing locally when this wasn't here. Not sure why I haven't seen it before...

joehuchette:
RFC now. New methods are covered.

I'm excited for when we don't have to carry around two versions of the source...

PR #494, opened 2015-07-29
test quadratic MPS writer

mlubin:
This fixes an apparent bug that variables which had 0 linear objective coefficients didn't appear in the list of columns, which Gurobi didn't like. It also seems that we don't need to duplicate the off-diagonal terms in the ``QMATRIX`` section (@cmaes?).

### DIFF
--- a/src/writers.jl
+++ b/src/writers.jl
@@ -13,7 +13,7 @@ function writeMPS(m::Model, fname::String)
     # Objective and constraint names
     gc_enable(false)
     write(f,"ROWS\n")
-    write(f," N  CON$(numRows+1)\n")
+    write(f," N  OBJ\n")
     hasrange = false
     for c in 1:numRows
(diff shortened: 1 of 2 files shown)


mlubin:
Cplex seems to say that you *do* need to provide the lower and upper triangle entry for off-diagonal terms, but gurobi doesn't. Should we care?

cmaes:
From http://www.gurobi.com/documentation/6.0/refman/mps_format.html

The next section in an MPS file is the optional QUADOBJ section, which contains quadratic objective terms. Each line in this section represents a single non-zero value in the *lower triangle* of the Q matrix. The names of the two variable that participate in the quadratic term are found first (starting in columns 5 and 15 in fixed format), followed by the numerical value of the coefficient (in column 25 in fixed format). By convention, the Q matrix has an implicit one-half multiplier associated with it. Here's an example containing three quadratic terms:

PR #498, opened 2015-07-29
better error messages inside callbacks

### LINKED ISSUE #496: better error messages inside callbacks
It would be nice to give better error messages for:
- ``@defVar(cb,...)`` where ``cb`` is the callback object
- ``@addConstraint(cb,...)`` where ``cb`` is the callback object
- ``@addLazyConstraint`` with quadratic expressions

CC @kaarthiksundar

mlubin:
closes #496


IainNZ:
Nice little addition

PR #501, opened 2015-07-29
Fix undefined variable error in zero(Var)

IainNZ:


IainNZ:
Having `one(Variable)` give you an affine expression of 1s is kinda wild when you think about it, but if it makes everything work...

PR #506, opened 2015-08-04
implement MPB interface for hessian-vector products

mlubin:
@dpo, feel free to try this out (and make sure that it's correct).
I haven't done any benchmarking.

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -67,7 +67,7 @@ function MathProgBase.initialize(d::JuMPNLPEvaluator, requested_features::Vector
     for feat in requested_features
         if !(feat in [:Grad, :Jac, :Hess, :HessVec, :ExprGraph])
             error("Unsupported feature $feat")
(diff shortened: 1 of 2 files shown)


src/nlp.jl
```diff
@@ -332,6 +333,35 @@ function MathProgBase.eval_jac_g(d::JuMPNLPEvaluator, J, x)
...
+import DualNumbers: Dual, epsilon
+
+function MathProgBase.eval_hesslag_prod(
+    d::JuMPNLPEvaluator,
+    h::Vector{Float64}, # output vector
+    x::Vector{Float64}, # current solution
+    v::Vector{Float64}, # rhs vector
+    σ::Float64,         # multiplier for objective
+    μ::Vector{Float64}) # multipliers for each constraint
+r
```
tkelman: what's `r` ?

mlubin: Good question. Not sure how that was passing tests.

PR #507, opened 2015-08-04
Add error messages for using NLExpr in quad/linear macros

IainNZ:
A user just hit this, I figured we can return a better error message.

### DIFF
--- a/src/v0.3/parseExpr_0.3.jl
+++ b/src/v0.3/parseExpr_0.3.jl
@@ -132,6 +132,23 @@ function addToExpression{C,V}(quad::GenericQuadExpr{C,V},c::Number,x::GenericQua
     quad
 end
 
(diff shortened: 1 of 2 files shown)


IainNZ:
Ambiguity warnings were not good, but I followed the warning messages and managed to address them.

mlubin:
Seems like a good idea

IainNZ:
Tangentially related, `@defNLExpr` prints out nasty innards when at the REPL, it'd almost be better just to have `(nonlinear expression)`, like Julia does with `(anonymous function)`

IainNZ:
Looks like Appveyor had download issues, and Travis is backlogged on OSX, but otherwise, good to merge?

joehuchette:
lgtm

PR #508, opened 2015-08-05
document querying derivatives from a JuMP model

### LINKED ISSUE #352: document getting an AbstractNLPEvaluator from a JuMP model
We should have a documented method to extract an AbstractNLPEvaluator from a JuMP model. This will make it easy to explicitly evaluate derivatives or extract expression graphs from a model for experimental purposes.

CC @tkelman

mlubin:
Closes #352 
CC @tkelman @dpo

### DIFF
--- a/doc/nlp.rst
+++ b/doc/nlp.rst
@@ -170,5 +170,35 @@ For example::
     # ∇f[getLinearIndex(x)] == cos(2.0)
     # ∇f[getLinearIndex(y)] == cos(3.0)
 
+The ordering of constraints in a JuMP model corresponds to the following ordering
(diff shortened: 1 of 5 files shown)


dpo:
That's great! The notebook you pointed me to earlier also has good examples worth mentioning.

mlubin:
CC @rgiordan

mlubin:
@dpo, I've expanded the docs and implemented ``buildInternalModel``, so now it should be easy to extract the standard-form representation from a JuMP model.

dpo:
Many thanks, Miles! I'll be able to give it a whirl tomorrow.

PR #512, opened 2015-08-06
Typo rand->randn for normal distribution?

anriseth:
The example says it computes the mle of a *normal* distribution. I assume you want to use 
data = randn(n) and not rand?


mlubin:
Thanks!

PR #513, opened 2015-08-06
RFC: Changes to allow JuMPeR to use matrix operations

IainNZ:
* Introduce a new abstract type, to allow Uncertain to be a "JuMPType" basically
* Tighten a few operator overloads, to avoid ambiguities
* Introduce a new function, `_multiply_type`, to determine the storage for matrix multiply

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -244,7 +244,8 @@ setPrintHook(m::Model, f) = (m.printhook = f)
 ###############################################################################
 # Variable class
 # Doesn't actually do much, just a pointer back to the model
-immutable Variable <: ReverseDiffSparse.Placeholder
+abstract AbstractJuMPScalar <: ReverseDiffSparse.Placeholder
+immutable Variable <: AbstractJuMPScalar
     m::Model
     col::Int
 end
@@ -957,9 +958,9 @@ operator_warn(lhs,rhs) = nothing
 ##########################################################################
 # Behavior that's uniform across all JuMP "scalar" objects
 
-typealias JuMPTypes Union(Variable,
+typealias JuMPTypes Union(AbstractJuMPScalar,
                           Norm,
-                          AffExpr,
+                          GenericAffExpr,
                           QuadExpr,
                           SOCExpr)
 typealias JuMPScalars Union(Number,JuMPTypes)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -39,10 +39,6 @@
 (-){T<:GenericSOCExpr}(lhs::Number, rhs::T) = T(copy(rhs.norm),    -rhs.coeff, lhs-rhs.aff)
 (*){T<:GenericSOCExpr}(lhs::Number, rhs::T) = T(copy(rhs.norm), lhs*rhs.coeff, lhs*rhs.aff)
 
-# This is not well defined if variable types are different, but needed to avoid ambiguities
-(+)(lhs::GenericAffExpr, rhs::GenericAffExpr) = (+)(promote(lhs,rhs)...)
-(-)(lhs::GenericAffExpr, rhs::GenericAffExpr) = (-)(promote(lhs,rhs)...)
-
 # Variable
 (+)(lhs::Variable) = lhs
 (-)(lhs::Variable) = AffExpr([lhs],[-1.0],0.0)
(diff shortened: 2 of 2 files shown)


src/JuMP.jl
```diff
@@ -244,7 +244,8 @@ setPrintHook(m::Model, f) = (m.printhook = f)
 ###############################################################################
 # Variable class
 # Doesn't actually do much, just a pointer back to the model
-immutable Variable <: ReverseDiffSparse.Placeholder
+abstract AbstractVariable <: ReverseDiffSparse.Placeholder
```
IainNZ: This is to allow `JuMPTypes` to be more general. It may also allow for other code savings in JuMPeR, but just focussed on linalg right now

IainNZ: I don't like the name though

joehuchette:
I've been thinking about this for a few weeks now, and how about something like
```jl
abstract AbstractScalar <: ReverseDiffSparse.Placeholder
immutable Variable <: AbstractScalar ... end
type GenericAffExpr <: AbstractScalar ... end
type JuMPeR.Uncertain <: AbstractScalar ... end
```
The only thing I'm unsure about is the ``<: ReverseDiffSparse.Placeholder``. Maybe that could go to a trails-like interface anyway? Seems a bit odd to be doing it through types like this.

joehuchette:
Or maybe name it something else like ``AbstractJuMPScalar`` so then we can do ``typealias AbstractScalar Union(AbstractJuMPScalar, Number)``

src/operators.jl
```diff
@@ -496,13 +496,13 @@ end
 
 # Don't do size checks here in _return_array, defer that to (*)
 function _return_array{R,S}(A::AbstractArray{R}, x::AbstractArray{S,1})
-    Q = (R <: JuMPTypes && S <: JuMPTypes) ? QuadExpr : AffExpr
+    Q = promote_type(R,S)
```
joehuchette: This works the same way on v0.3 and v0.4? If so that's great, I was never happy about this.

IainNZ: Things are not working entirely, but are at least working partially

joehuchette:
LGTM. Does it make sense to make ``AffExpr <: AbstractJuMPScalar``?

src/operators.jl
```diff
@@ -118,16 +114,16 @@ function (^)(lhs::Union(Variable,AffExpr), rhs::Integer)
...
+(-){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::V) = GenericAffExpr{C,V}(vcat(lhs.vars,rhs),vcat(lhs.coeffs,-one(C)),lhs.constant)
 (*)(lhs::AffExpr, rhs::Variable) = (*)(rhs,lhs)
 (/)(lhs::AffExpr, rhs::Variable) = error("Cannot divide affine expression by a variable")
 # AffExpr--Norm
 (+){C,V}(lhs::GenericAffExpr{C,V}, rhs::GenericNorm{2,C,V}) = GenericSOCExpr{C,V}(copy(rhs),  one(C), copy(lhs))
 (-){C,V}(lhs::GenericAffExpr{C,V}, rhs::GenericNorm{2,C,V}) = GenericSOCExpr{C,V}(copy(rhs), -one(C), copy(lhs))
 # AffExpr--AffExpr
-(+){T<:GenericAffExpr}(lhs::T, rhs::T) = (operator_warn(lhs,rhs); GenericAffExpr(vcat(lhs.vars,rhs.vars),vcat(lhs.coeffs, rhs.coeffs),lhs.constant+rhs.constant))
-(-){T<:GenericAffExpr}(lhs::T, rhs::T) = GenericAffExpr(vcat(lhs.vars,rhs.vars),vcat(lhs.coeffs,-rhs.coeffs),lhs.constant-rhs.constant)
+(+){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::GenericAffExpr{C,V}) = (operator_warn(lhs,rhs); GenericAffExpr(vcat(lhs.vars,rhs.vars),vcat(lhs.coeffs, rhs.coeffs),lhs.constant+rhs.constant))
```
mlubin: Why change this? The previous definition seems fine.

src/operators.jl
```diff
@@ -39,10 +39,6 @@
 (-){T<:GenericSOCExpr}(lhs::Number, rhs::T) = T(copy(rhs.norm),    -rhs.coeff, lhs-rhs.aff)
 (*){T<:GenericSOCExpr}(lhs::Number, rhs::T) = T(copy(rhs.norm), lhs*rhs.coeff, lhs*rhs.aff)
 
-# This is not well defined if variable types are different, but needed to avoid ambiguities
-(+)(lhs::GenericAffExpr, rhs::GenericAffExpr) = (+)(promote(lhs,rhs)...)
```
mlubin: Removing these definitions broke JuMPChance. I have:
```
Base.promote_rule(::Type{AffExpr},::Type{CCAffExpr}) = CCAffExpr
```
which saves a bunch of method definitions.

IainNZ: It was causing ambiguity  problems, not sure how to bring it back.

src/operators.jl
IainNZ: Ambiguity warnings, again. Plus, new definition also seemed fine, and this was a minimal  change that I couldn't see causing any issues that couldn't be fixed with the new abstract type

IainNZ:
To be clear, the only things I think really needed to be done was the notion of `_multiply_type` and widening `JuMPTypes`. Everything else was ambiguity related, so if you have a refactoring I'm definitely into it.

Issue #509, opened 2015-08-05
Look into iteration performance


mlubin:
Opening this so we don't forget the discussion in #488

joehuchette:
Closed with #516

PR #516, opened 2015-08-08
[RFC] Faster iteration for JuMPContainers

joehuchette:
ref #509, currently iteration for JuMPArrays is very inefficient, since the contract is that iteration returns ``(key..., val)`` for any JuMPContainer. First, this should move to returning ``key=>val`` to mirror changes in Base. 

Second, I'm wondering if we should just encourage people to use ``vals(d)`` for efficient iteration over the entries. More radically, we could just change iteration to act like ``vals(d)`` and encourage people that want the current behavior to use ``zip(keys(d), vals(d))``.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,7 +13,7 @@ Unversioned
   * Introduce warnings for two common performance traps: too many calls to ``getValue()`` on a collection of variables and use of the ``+`` operator in a loop to sum expressions.
   * Second-order cone constraints can be written directly with the ``norm()`` and ``norm2{}`` syntax.
   * Implement MathProgBase interface for querying Hessian-vector products.
-  * Iteration over ``JuMPContainers`` is deprecated; instead, use the ``keys`` and ``values`` functions, and ``zip(keys(d),values(d))`` for the old behavior.
+  * Iteration over ``JuMPContainer``s is deprecated; instead, use the ``keys`` and ``values`` functions, and ``zip(keys(d),values(d))`` for the old behavior.
 
 Version 0.9.2 (June 27, 2015)
 -----------------------------
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,7 +12,8 @@ Unversioned
   * Dual solutions are now available for general nonlinear problems. You may call ``getDual`` on a reference object for a nonlinear constraint, and ``getDual`` on a variable object for Lagrange multipliers from active bounds.
   * Introduce warnings for two common performance traps: too many calls to ``getValue()`` on a collection of variables and use of the ``+`` operator in a loop to sum expressions.
   * Second-order cone constraints can be written directly with the ``norm()`` and ``norm2{}`` syntax.
-  * Implement MathProgBase interface for querying Hessian-vector products
+  * Implement MathProgBase interface for querying Hessian-vector products.
+  * Iteration over ``JuMPContainers`` is deprecated; instead, use the ``keys`` and ``values`` functions, and ``zip(keys(d),values(d))`` for the old behavior.
 
 Version 0.9.2 (June 27, 2015)
 -----------------------------
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -51,7 +51,7 @@ export
     @setNLObjective, @addNLConstraint, @addNLConstraints,
     @defNLExpr
 
-include("JuMPDict.jl")
+include("JuMPContainer.jl")
 #include("JuMPArray.jl")
 include("utils.jl")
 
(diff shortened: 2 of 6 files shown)


mlubin:
It would be nice if iteration over ``OneIndexedArray``s matched iteration over julia ``Array``s, now that we're trying to have ``OneIndexedArray``s behave more predictably. That does create a consistency issue for the other ``JuMPContainer`` types. What if we disallow direct iteration over other ``JuMPContainer`` types and ask people to use ``keys`` and ``values`` explicitly?

joehuchette:
That sounds great to me. I'm guessing it'll have to be deprecated, though.

mlubin:
I think it's okay to break things for ``OneIndexedArray`` since that's new in 0.10 anyway. We should deprecate iteration over ``JuMPContainer`` other types.

joehuchette:
@mlubin @IainNZ do we have an iteration benchmark sitting around somewhere?

mlubin:
Not that I know of besides the ``getValue`` discussion: https://github.com/JuliaOpt/JuMP.jl/pull/488

joehuchette:
```jl
using JuMP

function bench1(N)
    m = Model()
    @defVar(m, x[1:N,1:N])
    cnt = 0
    for tmp in x
        cnt += 1
    end
end
```
master:
```
julia> @time bench1(10);
  0.000284 seconds (1.59 k allocations: 70.469 KB)

julia> @time bench1(100);
  0.025973 seconds (178.60 k allocations: 7.204 MB)

julia> @time bench1(1000);
  2.706214 seconds (19.95 M allocations: 636.033 MB, 7.60% gc time)
```
PR:
```
julia> @time bench1(10);
  0.000068 seconds (288 allocations: 28.359 KB)

julia> @time bench1(100);
  0.004143 seconds (20.13 k allocations: 2.650 MB)

julia> @time bench1(1000);
  0.584545 seconds (2.00 M allocations: 148.447 MB, 28.02% gc time)
```

mlubin:
Pretty good improvement

joehuchette:
For reference, just the allocation on PR:
```jl
function bench2(N)
    m = Model()
    @defVar(m, x[1:N,1:N])
    nothing
end
```
```
julia> @time bench2(10);
  0.000039 seconds (187 allocations: 20.469 KB)

julia> @time bench2(100);
  0.000987 seconds (10.13 k allocations: 1.887 MB)

julia> @time bench2(1000);
  0.157737 seconds (1.00 M allocations: 72.153 MB, 19.97% gc time)
```
So still some extra allocation somewhere (index sets?).

joehuchette:
But this looks like a regression for JuMPArrays...
```jl
function bench3(N)
    m = Model()
    @defVar(m, x[2:(N+1),1:N])
    cnt = 0
    for tmp in x
        cnt += 1
    end
end
```
master:
```
julia> @time bench3(10);
  0.000298 seconds (1.59 k allocations: 70.469 KB)

julia> @time bench3(100);
  0.025858 seconds (178.60 k allocations: 7.204 MB)

julia> @time bench3(1000);
  2.862607 seconds (19.96 M allocations: 656.064 MB, 7.97% gc time)
```
PR:
```
julia> @time bench3(10);
  0.000428 seconds (1.89 k allocations: 76.797 KB)

julia> @time bench3(100);
  0.042028 seconds (228.09 k allocations: 8.112 MB)

julia> @time bench3(1000);
  4.172662 seconds (32.40 M allocations: 841.226 MB, 8.40% gc time)
```

mlubin:
Isn't that deprecated?

joehuchette:
Yes

joehuchette:
In any case, ``values`` iteration is faster:
```jl
julia> function bench5(N)
           m = Model()
           @defVar(m, x[1:N,1:N])
           cnt = 0
           for tmp in values(x)
               cnt += 1
           end
       end
bench5 (generic function with 1 method)
```

```
julia> @time bench5(10);
  0.000042 seconds (188 allocations: 20.484 KB)

julia> @time bench5(100);
  0.001463 seconds (10.13 k allocations: 1.887 MB)

julia> @time bench5(1000);
  0.208572 seconds (1.00 M allocations: 92.153 MB, 40.04% gc time)
```

mlubin:
So what does iteration on a ``OneIndexedArray`` do after this PR? Give a deprecation warning or iterate over ``values()``?

joehuchette:
Iterate over values, but something seems a bit off with the timings in that case.

joehuchette:
Hmm:
```
julia> Base.return_types(start, (JuMP.OneIndexedArray{Float64,1},))
1-element Array{Any,1}:
 Any

julia> Base.return_types(next, (JuMP.OneIndexedArray{Float64,1},Int))
1-element Array{Any,1}:
 Any
```

mlubin:
``OneIndexedArray`` is an abstract type

mlubin:
Oh, the benchmark is a bit weird because ``@defVar`` generates a new type inside the function itself (at macro expansion time). Seems like that would easily throw type inference. What about moving ``@defVar`` outside of the benchmark?

joehuchette:
If that throws inference off, I feel like we have bigger problems to worry about...

mlubin:
Well, check out ``@code_warntype`` to confirm

joehuchette:
Just looked, it seems alright. I fixed up a missing method so now ``for _ in x`` and ``for _ in values(x)`` gets the same performance behavior.

joehuchette:
If there are no objections, I'll merge in the next day or two.

mlubin:
LGTM

PR #532, opened 2015-08-14
Fix #520 by not adding zeros to constraint matrix

### LINKED ISSUE #520: Gurobi seems to think there are small coefficients somewhere, or large coefficients
I have a fairly simple & small model, and when I solve it with Gurobi it says:

```
Optimize a model with 20 rows, 40 columns and 75 nonzeros
Coefficient statistics:
  Matrix range    [0e+00, 5e+02]
  Objective range [2e-01, 5e+01]
  Bounds range    [1e+00, 1e+00]
  RHS range       [7e+01, 1e+02]
Warning: Model contains large matrix coefficient range
         Consider reformulating model or setting NumericFocus parameter
         to avoid numerical issues.
```

However, printing the A matrix after `prepConstrMatrix` reveals nothing out of the ordinary, and printing the model to an LP and loading that with gurobipy also doesn't trigger the warning. Sebastien reported to me a similar message where Gurobi said it was removing variables with very small coefficients, even though he couldn't find any by printing out. Is there somewhere we could be injecting noise?

IainNZ:
Tossed in an inbounds for good measure

### DIFF
--- a/src/solvers.jl
+++ b/src/solvers.jl
@@ -232,11 +232,13 @@ function prepConstrMatrix(m::Model)
             end
             addelt!(tmprow,vars[ind].col,coeffs[ind])
         end
-        for i in 1:tmprow.nnz
-            nnz += 1
+        @inbounds for i in 1:tmprow.nnz
             idx = tmpnzidx[i]
+            elt = tmpelts[idx]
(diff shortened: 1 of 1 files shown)


mlubin:
That works

PR #534, opened 2015-08-14
Re-org and removal of redundant methods

IainNZ:
This PR is mostly just reorganization, mostly because I felt JuMP.jl was organized in a way that no longer made sense and was over a 1000 lines long.

If you don't hate it, I'll merge it.


mlubin:
LGTM, let's make sure travis approves.

PR #535, opened 2015-08-15
Strip zeros from affine part of quadratic expressions

IainNZ:
```
julia> using JuMP

julia> m=Model()
Feasibility problem with:
 * 0 linear constraints
 * 0 variables
Solver set to Default

julia> @defVar(m, 0 <= x[1:3] <= 1)
0 ≤ x[i] ≤ 1 for all i in {1,2,3}

julia> @defVar(m, y >= 0)
y

julia> @addConstraint(m, dot(x,x) <= y*y)
x[1]² + x[2]² + x[3]² - y² + 0 ≤ 0
```

That extra 0 is actually a `0y`, which causes this model to die if passed through the conic MPB path. I fixed it here, because its generally a useful thing I figure. I don't know the `0y` appears in the first place.

### DIFF
--- a/src/solvers.jl
+++ b/src/solvers.jl
@@ -138,11 +138,19 @@ function addQuadratics(m::Model)
                 error("Variable not owned by model present in constraints")
             end
         end
-        affidx  = Cint[v.col for v in qconstr.terms.aff.vars]
+        n_aff = length(qconstr.terms.aff.coeffs)
+        affcoef = Float64[]; sizehint!(affcoef, n_aff)
+        affidx  = Cint[];    sizehint!(affidx,  n_aff)
+        for (coeff, var) in qconstr.terms.aff
+            if coeff != 0.0
+                push!(affcoef, coeff)
+                push!(affidx, var.col)
+            end
+        end
         var1idx = Cint[v.col for v in qconstr.terms.qvars1]
         var2idx = Cint[v.col for v in qconstr.terms.qvars2]
-        if applicable(MathProgBase.addquadconstr!, m.internalModel, affidx, qconstr.terms.aff.coeffs, var1idx, var2idx, qconstr.terms.qcoeffs, s, -qconstr.terms.aff.constant)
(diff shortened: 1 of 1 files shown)


mlubin:
What about just eliminating duplicates in the linear part? May also be useful for some solvers.

IainNZ:
I wasn't really sure how the existing merge_duplicates function fit in, and I noticed indexedvector wasn't already in use here, so I kept it simple

joehuchette:
Interesting that it doesn't happen outside the macros:
```jl
julia> dot(x,x) <= y*y
WARNING: The comparison operator <= has been deprecated for constructing constraints. Use the macro form @addConstraint instead.
x[1]² + x[2]² + x[3]² - y² ≤ 0
```

IainNZ:
@mlubin did you look at the Travis logs? They seem completely unrelated to this PR

IainNZ:
That aside, I need this (and only this!) for JuMPeR to be completely switched over to `norm2`. I think it'd be good to do more a general job, as part of some refactoring of solve.jl. So I'll merge once we confirm 100% the failure isn't from this (it passes locally).

joehuchette:
The problem comes from this method:
```
function addToExpression{C,V}(quad::GenericQuadExpr{C,V},c::GenericAffExpr{C,V},x::V)
    append!(quad.qvars1, c.vars)
    append!(quad.qvars2, fill(x,length(c.vars)))
    append!(quad.qcoeffs, c.coeffs)
    addToExpression(quad.aff,c.constant,x)
    quad
end
```
In particular, ``c.constant == 0`` in the last line. Why not fix this problem at the source and just drop 0 terms?

IainNZ:
That works for me to, I just couldn't figure out where it was coming from last night :)

IainNZ:
Can you submit a PR for that?

joehuchette:
Yep

IainNZ:
#536 is cleaner. Checking at this stage can be handled in more comprehensive matter in a later PR.

PR #539, opened 2015-08-16
Improve typeinf for dot ops

IainNZ:
```
using JuMP, JuMPeR
m=RobustModel()
@defUnc(m, u[1:5])
a = [3.0, 0.0, 0.0, 2.0, 1.0]
a.*u  # This was an Array{Any}
```
The benefit is that you no longer get tossed into generic base `vecnorm` code, which matters if you aren't used `normX{...}`.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -581,16 +581,67 @@ for (dotop,op) in [(:.+,:+), (:.-,:-), (:.*,:*), (:./,:/)]
     @eval begin
         $dotop(lhs::Real,rhs::JuMPTypes) = $op(lhs,rhs)
         $dotop(lhs::JuMPTypes,rhs::Real) = $op(lhs,rhs)
-        $dotop{T<:JuMPTypes}(lhs::Real,rhs::Array{T}) = map(c->$op(lhs,c), rhs)
-        $dotop{T<:JuMPTypes}(lhs::Array{T},rhs::Real) = map(c->$op(c,rhs), lhs)
+        function $dotop{T<:JuMPTypes}(lhs::Real,rhs::Array{T})
+            arr_type = typeof($op(lhs,one(T)))
+            arr = Array(arr_type, size(rhs))
+            @inbounds for i in eachindex(rhs)
+                arr[i] = $op(lhs,rhs[i])
+            end
+            return arr
+        end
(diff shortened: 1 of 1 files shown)


src/operators.jl
```diff
@@ -581,16 +581,67 @@ for (dotop,op) in [(:.+,:+), (:.-,:-), (:.*,:*), (:./,:/)]
...
+                arr[i] = $op(lhs[i],rhs)
+            end
+            return arr
+        end
         $dotop(lhs::Real,rhs::OneIndexedArray) = $dotop(lhs, rhs.innerArray)
         $dotop(lhs::OneIndexedArray,rhs::Real) = $dotop(lhs.innerArray, rhs)
-        $dotop{T<:JuMPTypes}(lhs::T,rhs::Array) = map(c->$op(lhs,c), rhs)
-        $dotop{T<:JuMPTypes}(lhs::Array,rhs::T) = map(c->$op(c,rhs), lhs)
+        function $dotop{T<:JuMPTypes}(lhs::T,rhs::Array)
+            length(rhs) == 0 && return Array(eltype(rhs),size(rhs))
```
joehuchette: Is this type stable?

joehuchette: What about
```jl
Array(typeof(one(T) * one(eltype(rhs))), size(rhs))
```
?

IainNZ: eltype might be Any, so it'd throw an error. Base has same problem, its not possible to be type stable in the sense of the output type being decidable by only the input types. But at least now it'll be tight for the non empty case, unlike before.

joehuchette: What about adding a method for ``Array{Any}`` then?

IainNZ: That'd look just like this though, returns ArrayAny if its empty but will be tighter if possible. The whole point of doing this is to avoid returning ArrayAnys if possible. I also don't think type stability is going to make difference here.

IainNZ:
I think this is needed not just for JuMPeR, but of course for JuMP for the same reasons. I'm going to YOLO-merge, and we can also take a second pass at improving performance if it ever matters.

joehuchette:
What about all the other uses of ``map`` in this file?

IainNZ:
I haven't hit any problems with them yet. I think the real catch was with the way `Base.vecnorm` was behaving here.

PR #543, opened 2015-08-20
Stop redundant SOS constraints being added

IainNZ:
Came up during my code review. If you have an LP with Gurobi as the solver, add some continuous variables and an SOS constraint, and just keep hitting solve, you'll add SOS constraints each time to the internal model. I think this is all thats required to fix this, as the addSOS constraint methods already do problem modification if possible, and reset internalModelLoaded if not.


joehuchette:
Test?

IainNZ:
Couldn't think of a way to do it in general, can do it for Gurobi though?

IainNZ:
Given we can't ask the MPB interface how many SOS constraints its got, I'm just going to punt on tests for now

PR #552, opened 2015-08-29
parsing bugfix, closes #551

### LINKED ISSUE #551: bug in addConstraint
On 0.4 only:
```
julia> m = Model();

julia> @defVar(m, x)
x

julia> f(z) = z

julia> @addConstraint(m, x ≤ f(5)^2)
ERROR: MethodError: `constructconstraint!` has no method matching constructconstraint!(::Float64, ::Symbol)
Closest candidates are:
  constructconstraint!(::JuMP.Variable, ::Symbol)
  constructconstraint!(::JuMP.GenericAffExpr{Float64,JuMP.Variable}, ::Symbol)
  constructconstraint!(::JuMP.GenericAffExpr{Float64,JuMP.Variable}, ::Any, ::Any)
```
This can also give the wrong answer sometimes:
```
julia> using JuMP

julia> m = Model();

julia> @defVar(m, x)
x

julia> f(z) = z

julia> @addConstraint(m, x ≤ f(x)^2)
x² ≤ 0
```

mlubin:
Looks like this code was never really tested? I think we need to verify code coverage before releasing 0.10, given the huge amount of new code.

### DIFF
--- a/src/v0.4/parseExpr_staged.jl
+++ b/src/v0.4/parseExpr_staged.jl
@@ -529,7 +529,7 @@ function parseExpr(x, aff::Symbol, lcoeffs::Vector, rcoeffs::Vector, newaff::Sym
             s = gensym()
             newaff_, parsed = parseExprToplevel(x.args[2], s)
             push!(blk.args, :($s = 0.0; $parsed))
-            push!(blk.args, :($newaff = $newaff_*$newaff_))
+            push!(blk.args, :($newaff = $(Expr(:call,:*,lcoeffs...,newaff_,newaff_,rcoeffs...)) + $aff))
             return newaff, blk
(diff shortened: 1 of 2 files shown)


IainNZ:
Does codecov say its tested? I've been ignoring it for a while

IainNZ:
Do we need the version check in Travis? I wonder if its smart enough to merge them

mlubin:
Actually this code has been around since 0.7: https://github.com/JuliaOpt/JuMP.jl/commit/dcf17f595734fa847b2d0036c25611e5f323e74b

mlubin:
And it's tested but just not on the path that caused the bug...

IainNZ:
ah damn thats worst-case scenario

PR #561, opened 2015-09-01
Tweak buildInternalModel for compile speed

IainNZ:
Unfortunately this is all I could get. Next thing to hit would be to precompile `loadproblem!` for JuMP's inputs and their own internal model types. Thats a bit messy.

```
using Gurobi

macro howlong(stage)
    esc(quote
        println($stage, " ", (time_ns()-start_t)/10^9)
        start_t = time_ns()
    end)
end

start_t = time_ns()
@howlong "start"
using JuMP
@howlong "using"
m = Model(solver=GurobiSolver(OutputFlag=0))
@howlong "model"
N=1000
@defVar(m, x[1:N])
@howlong "defvr"
@addConstraint(m, sum{i*x[i],i=1:N} >= N)
@howlong "addcn"
solve(m)
@howlong "solve"
```

```
# Master
using 0.280585613
solve 1.886616573

# After buildInternalModel change
using 0.271496954
solve 1.433308388
```

The other methods could be precompiled, but they didn't really seem to register, possibly are precompiled just by precompiled buildInternalModel? Also `solve` with the kwargs didn't really help much either.

### DIFF
--- a/src/solvers.jl
+++ b/src/solvers.jl
@@ -74,7 +74,7 @@ function solve(m::Model; suppress_warnings=false,
     traits = ProblemTraits(m)
 
     # Build the MathProgBase model from the JuMP model
-    buildInternalModel(m, traits, suppress_warnings=suppress_warnings)
+    buildInternalModel(m, traits, suppress_warnings)
 
     # If the model is a general nonlinear, use different logic in
     # nlp.jl to solve the problem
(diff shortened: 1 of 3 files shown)


joehuchette:
``buildInternalModel`` is exported and documented, so this is a breaking change.

IainNZ:
ah...
maybe I can make it not one, let me see

mlubin:
Is the motivation for removing keyword arguments just so that it's easier to precompile?

IainNZ:
kwargs (seem to) cause type inference trouble, so even without precompilation they seem to be best avoided if you want to extract maximum compile speed. Similar how to my refactoring of printing made time-to-print quicker on 0.3 even without the benefit of precompilation, although that was due to too many arguments, not kwargs.

IainNZ:
Optional arguments are different (they just generate more dummy methods).

mlubin:
Not sure the API breakage is worth the extra 0.4 seconds. We at least need to keep the fallback with one argument.

IainNZ:
I'm also inclined just to leave this as-is.

PR #569, opened 2015-09-03
Remove variable 'categories'

IainNZ:
This is a left over from very early naming choices I made, we use type everywhere pretty much, even in MPB. Here is me paying for past sins.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -38,7 +38,7 @@ export
     # Variable
     setName, getName, setLower, setUpper, getLower, getUpper,
     getValue, setValue, getDual, setCategory, getCategory,
-    getVar,
+    getVar, setType, getType,
     getLinearIndex,
     # Expressions and constraints
     affToStr, quadToStr, exprToStr, conToStr, chgConstrRHS,
@@ -74,7 +74,7 @@ type Model
     colNamesIJulia::Vector{UTF8String}
     colLower::Vector{Float64}
     colUpper::Vector{Float64}
-    colCat::Vector{Symbol}
+    colType::Vector{Symbol}
 
     # Variable cones of the form, e.g. (:SDP, 1:9)
     varCones::Vector{@compat Tuple{Symbol,Any}}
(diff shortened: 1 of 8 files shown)


mlubin:
``getType`` seems like it could be confused with a generic julia function (besides the camel case)

mlubin:
That's why we have ``getCategory``

joehuchette:
Yeah, is this just a renaming?

IainNZ:
This is just a renaming. In MPB we use type, and almost everywhere inside JuMP we use `vartype`, `v_type`, `t`, except for here. It makes it awkward to talk about - I'd have preferred to use category throughout, but I guess it didn't stick so now we're in a strange spot.

IainNZ:
So, what you guys thinking?

joehuchette:
I dunno, seems like a lot of churn for a change I'm not really convinced is better than what's there now.

mlubin:
Where are all the places that we're currently using type over category? A grep for vartype shows:
```
type GenericQuadExpr{CoefType,VarType}
```
I know MPB uses vartype, but that doesn't mean we have to.

IainNZ:
At a glance:
- MPB
- `t` in `@defVar`, `writers.jl`
- `setCategory` itself refers to it as a `type` in its error message + arguments, and we refer to a list of `var_types`: https://github.com/JuliaOpt/JuMP.jl/blob/e0164566e953616c6c9e4bd0eacc3d5b03be6372/src/JuMP.jl#L419
- `vartypes_without_fixed` https://github.com/JuliaOpt/JuMP.jl/blob/e0164566e953616c6c9e4bd0eacc3d5b03be6372/src/solvers.jl#L471-L480

IainNZ:
If you prefer category, we can go the other way as well. I don't think its good how it is now.

IainNZ:
I kind of like category, I just figured that maybe you guys didn't.

mlubin:
I actually prefer category.

PR #572, opened 2015-09-04
More cats

IainNZ:
First commit is needed, second commit is optional

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -415,10 +415,10 @@ function getDual(v::Variable)
     return v.m.redCosts[v.col]
 end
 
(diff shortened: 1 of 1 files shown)


joehuchette:
:+1:

src/solvers.jl
```diff
@@ -468,7 +468,7 @@ function prepConstrMatrix(m::Model)
     A = rowmat'
 end
 
-function vartypes_without_fixed(m::Model)
+function cats_without_fixed(m::Model)
```
mlubin: Maybe ``varcats`` instead of ``cats``? Otherwise I'm happy with ``vartypes`` since we're talking about translating into MPB format.

IainNZ:
Yeah I'm just going to stick with `setCategory` change

PR #576, opened 2015-09-05
Remove special-cased matmul in addToExpression...

joehuchette:
...and fix some bugs it revealed.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -470,30 +470,30 @@ end
 # Let's be conservative and only define arithmetic for the basic types
 typealias ArrayOrSparseMat{T} Union(Array{T}, SparseMatrixCSC{T})
 
-for op in [:+, :-, :*, :/]; @eval begin
+for op in [:+, :-]; @eval begin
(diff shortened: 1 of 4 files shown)


src/operators.jl
```diff
@@ -501,13 +501,45 @@ for op in [:+, :-, :*, :/]; @eval begin
     end
 end; end
 
-# Special-case sparse matrix scalar multiplicaiton
+for op in [:*, :/]; @eval begin
+    function $op{T<:JuMPTypes}(lhs::Number,rhs::Array{T})
```
IainNZ: Whats the intuition for these things not working for SparseMatrices?

joehuchette:
```jl
m = Model()
@defVar(m, x)
A = SparseMatrixCSC(3, 3, [1:4;], [1:3;], [x,x,x])
B = 3A
```
You want ``B`` to be sparse as well, which ``similar`` gives you, but then ``eachindex`` does iteration over all elements, not just nonzeros.

PR #579, opened 2015-09-06
Revert "generalize addToExpression (0 -> zero(C))"

mlubin:
This commit broke cases where affine expressions are coefficients of another expression, i.e., JuMPChance. We can't use ``!=`` like this until we kill the deprecated comparison operators. It's also a big performance issue to create and throw away expressions all the time.

Is there a reason why we don't want to compare ``Number`` types with zero? It's pretty much guaranteed to always be defined.

### DIFF
--- a/src/v0.3/parseExpr_0.3.jl
+++ b/src/v0.3/parseExpr_0.3.jl
@@ -91,9 +91,9 @@ function addToExpression(quad::GenericQuadExpr,c::Number,x::Number)
 end
 
 function addToExpression{C,V}(quad::GenericQuadExpr{C,V},c::V,x::V)
-    push!(quad.qvars1, x)
-    push!(quad.qvars2, c)
-    push!(quad.qcoeffs, 1.0)
+    push!(quad.qvars1, c)
+    push!(quad.qvars2, x)
+    push!(quad.qcoeffs, one(C))
     quad
 end
 
--- a/src/v0.3/parseExpr_0.3.jl
+++ b/src/v0.3/parseExpr_0.3.jl
@@ -34,7 +34,7 @@ function addToExpression(aff::GenericAffExpr,c::Number,x::Number)
 end
 
 function addToExpression{C,V}(aff::GenericAffExpr{C,V}, c::Number, x::V)
-    if c != zero(C)
+    if c != 0
         push!(aff.vars,   x)
         push!(aff.coeffs, c)
     end
@@ -42,7 +42,7 @@ function addToExpression{C,V}(aff::GenericAffExpr{C,V}, c::Number, x::V)
 end
 
 function addToExpression{C,V}(aff::GenericAffExpr{C,V},c::Number,x::GenericAffExpr{C,V})
-    if c != zero(C)
+    if c != 0
         append!(aff.vars, x.vars)
         append!(aff.coeffs, c*x.coeffs)
         aff.constant += c*x.constant
@@ -68,7 +68,7 @@ addToExpression{C}(aff::GenericAffExpr{C,Variable},c::Variable,x::GenericAffExpr
                                 addToExpression(aff,x.constant,c))
 
 function addToExpression{C,V}(aff::GenericAffExpr{C,V}, c::Number, x::GenericQuadExpr{C,V})
-    if c == zero(C)
+    if c == 0
         convert(GenericQuadExpr{C,V}, aff)
     else
         GenericQuadExpr{C,V}(copy(x.qvars1),
@@ -79,7 +79,7 @@ function addToExpression{C,V}(aff::GenericAffExpr{C,V}, c::Number, x::GenericQua
 end
 
 function addToExpression{C,V}(quad::GenericQuadExpr{C,V},c::Number,x::V)
-    if c != zero(C)
+    if c != 0
         push!(quad.aff,convert(C,c),x)
     end
     quad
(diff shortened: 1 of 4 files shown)


src/v0.3/parseExpr_0.3.jl
```diff
@@ -91,14 +91,14 @@ function addToExpression(quad::GenericQuadExpr,c::Number,x::Number)
 end
 
 function addToExpression{C,V}(quad::GenericQuadExpr{C,V},c::V,x::V)
-    push!(quad.qvars1, c)
-    push!(quad.qvars2, x)
-    push!(quad.qcoeffs, one(C))
+    push!(quad.qvars1, x)
+    push!(quad.qvars2, c)
+    push!(quad.qcoeffs, 1.0)
```
IainNZ: This doesn't seem valid, `one(C)` was correct

IainNZ: I guess it'll convert though

mlubin: Yes, it'll convert. We can tidy this up after the breakage is fixed.

IainNZ:
JuMPeR works fine on JuMP master, so not sure why this didn't break JuMPeR if it broke JuMPChance.

IainNZ:
JuMPeR also works on this PR too.

IainNZ:
Can you give an example of some JuMPChance code that broke, maybe I'm missing something in my test coverage?

mlubin:
```
julia> using JuMPChance, JuMP

julia> m = ChanceModel();

julia> @defIndepNormal(m, z[1:2], mean=0, var=1);

julia> @defVar(m, x[1:2]);

julia> ex = z[1]*x[1]
(x[1])*z[1] + 0

julia> @defExpr z[2]*x[1] + ex
WARNING: The comparison operator == has been deprecated for constructing constraints. Use the macro form @addConstraint instead.
ERROR: MethodError: `!` has no method matching !(::JuMP.GenericRangeConstraint{JuMP.GenericAffExpr{Float64,JuMP.Variable}})
 in addToExpression at /home/mlubin/.julia/v0.4/JuMP/src/v0.4/parseExpr_staged.jl:77
 in addToExpression_reorder at /home/mlubin/.julia/v0.4/JuMP/src/v0.4/parseExpr_staged.jl:254
```

IainNZ:
Ah its because I already removed operator overloads in JuMPeR

mlubin:
I think this is using ``AffExpr``'s operator overloads though

IainNZ:
I don't follow... there isn't a GenericAffExpr fallback for `==`... or do you mean something else?

mlubin:
My coefficient type is ``AffExpr``, which has a definition for ``==``:
```
julia> using JuMP

julia> m = Model();

julia> @defVar(m, x);

julia> 3x != 0
WARNING: The comparison operator == has been deprecated for constructing constraints. Use the macro form @addConstraint instead.
ERROR: MethodError: `!` has no method matching !(::JuMP.GenericRangeConstraint{JuMP.GenericAffExpr{Float64,JuMP.Variable}})
 in != at operators.jl:27
```

IainNZ:
Ohh, right. Must be that coefficient thing then.

src/v0.4/parseExpr_staged.jl
```diff
@@ -86,11 +86,11 @@ function addToExpression{C,V}(aff::GenericAffExpr{C,V},c::Number,x::GenericAffEx
 end
 
 # help w/ ambiguity
-addToExpression{C,V<:Number}(aff::GenericAffExpr{C,V}, c::Number, x::Number) = aff + c*x
-
+addToExpression{C,V<:Number}(aff::GenericAffExpr{C,V}, c::Number, x::Number) =
+    __addToExpression__(aff, c, x)
```
joehuchette: This change should stay

joehuchette:
Can't we just remove the comparison operators now? 

It seems like the compiler should be easily able to optimize this down to just the appropriate checks, but maybe that's not the case.

mlubin:
> Can't we just remove the comparison operators now?

Should be okay for 0.11, not for 0.10.2.

> It seems like the compiler should be easily able to optimize this down to just the appropriate checks, but maybe that's not the case.

I really doubt the julia compiler will optimize away creation of temporary expressions here.

Comparing ``Number`` types with zero is perfectly valid, what's the motivation for making it more complicated?

src/v0.4/parseExpr_staged.jl
mlubin: Updated to keep this

src/v0.3/parseExpr_0.3.jl
```diff
@@ -91,14 +91,14 @@ function addToExpression(quad::GenericQuadExpr,c::Number,x::Number)
 end
 
 function addToExpression{C,V}(quad::GenericQuadExpr{C,V},c::V,x::V)
-    push!(quad.qvars1, c)
-    push!(quad.qvars2, x)
+    push!(quad.qvars1, x)
+    push!(quad.qvars2, c)
```
joehuchette: Also this change, though I guess it's mostly cosmetic

joehuchette:
> Comparing Number types with zero is perfectly valid, what's the motivation for making it more complicated?

This seems more correct and less weirdly error-prone if there's some custom ``GenericAffExpr`` that does something really crazy, but that's speculative.

mlubin:
But ``GenericAffExpr`` isn't a subtype of Number. When would ``c != 0`` give a different answer than ``c != zero(c)``?

joehuchette:
I'm not sure, maybe something like ``x-x``, depending on how ``==`` is implemented? My only point being that this approach seemed easier to reason about, but I'm fine reverting for the reasons you've laid out.

PR #580, opened 2015-09-09
Move to 0.5-defensive Travis settings

IainNZ:
I saw @tkelman do this

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ os:
     - linux
     - osx
 julia:
-    - release
+    - 0.3
+    - 0.4
     - nightly
 notifications:
     email: false


tkelman:
Though I think Example.jl isn't on the docker workers so that job of mine is still queued. You'll only be getting RC's with `0.4`, you could leave `nightly` in unless you think Julia master is going to start going nuts soon.

tkelman:
Failure is due to COIN-OR's not being back up yet. RC1 looks like it works though.

IainNZ:
Yeah I suppose we can take out 0.5 if it does go nuts, might as well keep in for now. We'll be dropping 0.3 support soon enough anyway.

>You'll only be getting RC's with 0.4

And 0.4.0 right, eventually?

tkelman:
You're deleting nightly here currently. And yes, once 0.4.0, 0.4.1, etc are ready then `0.4` will use whatever's newest.

IainNZ:
But there isn't such a thing as 0.4-nightly anymore right, nightly=0.5?

tkelman:
correct, that's what I meant by "you'll only be getting RC's"

mlubin:
All passing now

IainNZ:
Nice. So @mlubin, do you think 0.3/0.4, or 0.3/0.4/nightly?

mlubin:
I don't see the harm in running with nightly.

IainNZ:
Indeed. I guess its weird right now, does `release` point to 0.3 and `nightly` point to 0.5, with no one running Travis against the RC? I guess RC and 0.5 are approximately equal, but still, kind of amusing.

Issue #582, opened 2015-09-11
Potentially confusing scoping behavior in macros


joehuchette:
```jl
using JuMP
m = Model()
@defVar(m, x[1:3])

for i in 1:3
    @addConstraint(m, sum{x[i], i=1} == 1)
    @show i
end
```
Gives
```
i = 1
i = 1
i = 1
```
This doesn't qualify as a bug necessarily, but is potentially confusing from a users perspective.

mlubin:
That's a feature of julia, but I think it would be reasonable for the macros to introduce a ``let`` block around for loops and mark any indexing variables as local, if there's no performance penalty. Would require a whole new chunk of code in the macros.

joehuchette:
Closed by #584

PR #584, opened 2015-09-12
Wrap indices used in macros in let/local

joehuchette:
Ref #582, this seems like a reasonable change in that it makes JuMP behave more like a DSL. Haven't added v0.3 support yet, nor done perf testing to see if this slows things down.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -92,11 +92,15 @@ function getloopedcode(c::Expr, code, condition, idxvars, idxsets, idxpairs, sym
         vname = expr.args[1].args[1]
         expr.args[1] = :tmp
         code = quote
-            for $i in $(idxsets[1]), $j in $(idxsets[2])
-                $i <= $j || continue
-                $expr
-                $vname[$i,$j] = tmp
-                $vname[$j,$i] = tmp
+            let
+                $(localvar(i))
+                $(localvar(j))
+                for $i in $(idxsets[1]), $j in $(idxsets[2])
+                    $i <= $j || continue
+                    $expr
+                    $vname[$i,$j] = tmp
+                    $vname[$j,$i] = tmp
+                end
             end
         end
     else
@@ -108,8 +112,11 @@ function getloopedcode(c::Expr, code, condition, idxvars, idxsets, idxpairs, sym
         end
         for (idxvar, idxset) in zip(reverse(idxvars),reverse(idxsets))
             code = quote
-                for $(esc(idxvar)) in $idxset
-                    $code
+                let
+                    $(localvar(esc(idxvar)))
+                    for $(esc(idxvar)) in $idxset
+                        $code
+                    end
                 end
             end
         end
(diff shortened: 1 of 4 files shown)


src/macros.jl
```diff
@@ -130,6 +137,26 @@ end
 
 getloopedcode(c, code, condition, idxvars, idxsets, idxpairs, sym) = code
 
+localvar(x::Expr) = Expr(:block, _localvar(x)...)
+_localvar(x::Symbol) = :(local $(esc(x)))
+function _localvar(x::Expr)
```
mlubin: What does this do?

joehuchette: Handles this gross case from the tests:
```jl
@defVar(m, tri_3[(i,j)=[(i,i+2) for i in 1:5],k=i:j])
```
and gives you
```jl
local i
local j
for (i,j) = [(i,i + 2) for i = 1:5]
    ...
end
```

src/macros.jl
```diff
@@ -130,6 +137,26 @@ end
...
+    @assert x.head == :escape
+    args = Any[]
+    for t in x.args
+        if isa(t, Symbol)
+            push!(args, :(local $(esc(t))))
+        else
+            @assert isa(t, Expr)
+            if t.head in (:tuple, :vect, :vcat, :hcat, :row)
```
mlubin: How would all these different expression types come up?

joehuchette: ``(i,j)``, ``[i,j]``, ``[i;j]``, ``[i j]``, and ``[i j; k l]`` (for the last two)

mlubin: ``for [i,j] in X`` is valid syntax?

joehuchette: Guess not

joehuchette:
Not seeing any performance regressions. On master:
```
test/perf❯ j speed.jl                          JuMP/git/master
PMEDIAN BUILD MIN=0.642807454  MED=0.806141951
PMEDIAN WRITE MIN=2.21950597  MED=2.448302863
CONT5 BUILD   MIN=0.282755884  MED=0.338333321
CONT5 WRITE   MIN=1.964233636  MED=1.988693638

test/perf❯ j macro.jl                          JuMP/git/master
  Running N=20...
    N=20 min 0.008439511
    N=20 min 0.021468948
  Running N=50...
    N=50 min 0.159582901
    N=50 min 0.410900603
  Running N=100...
    N=100 min 1.538156
    N=100 min 3.739682569
```
and this PR:
```
test/perf❯ j speed.jl               JuMP/git/localvars
PMEDIAN BUILD MIN=0.620035225  MED=0.805018357
PMEDIAN WRITE MIN=2.225878204  MED=2.530214445
CONT5 BUILD   MIN=0.265568488  MED=0.329455468
CONT5 WRITE   MIN=1.961389032  MED=1.985305759

test/perf❯ j macro.jl                       JuMP/git/localvars
  Running N=20...
    N=20 min 0.008394612
    N=20 min 0.021188721
  Running N=50...
    N=50 min 0.165039593
    N=50 min 0.412274769
  Running N=100...
    N=100 min 1.512589045
    N=100 min 3.707197417
```

mlubin:
Could you compare with 0.3 just for sanity check?

Also need lots of new tests here.

test/macros.jl
```diff
@@ -389,3 +389,44 @@ facts("[macros] Special-case binary multiplication in addToExpression_reorder (#
...
+    m = Model()
+    cnt = 4
+    for i in 5:8
+        @defVar(m, x[i=1:3,j=1:3] ≤ i)
+        cnt += 1
+        @fact i --> cnt
+    end
+    cnt = 4
+    for i in 5:8
+        @defVar(m, y[i=2:4,j=1:3] ≤ i)
```
mlubin: Check the upper bounds on ``y``?

src/macros.jl
```diff
@@ -130,6 +137,26 @@ end
...
+    args = Any[]
+    for t in x.args
+        if isa(t, Symbol)
+            push!(args, :(local $(esc(t))))
+        else
+            @assert isa(t, Expr)
+            if t.head == :tuple
+                append!(args, map(_localvar, t.args))
+            else
+                error("Internal error; please file an issue")
```
mlubin: How? Maybe add the URL?

joehuchette: Will add that separately.

mlubin:
I'm a bit worried that this will subtly break something. I think it needs a bit of time to settle in. Not sure if it should go into 0.10.2.

joehuchette:
Agreed, since this is technically breaking.

joehuchette:
Should maybe get a NEWS entry then

PR #586, opened 2015-09-18
Fix 0.4 deprecated bindings

yuyichao:


yuyichao:
Looks like AppVeyor is having a hard time printing all the deprecation warning of `Union(args...)` from `LibExpat.jl` .....

src/nlp.jl
```diff
@@ -593,7 +593,7 @@ function solvenlp(m::Model, traits; suppress_warnings=false)
 end
 
 # getValue for nonlinear subexpressions
-function getValue(x::Union(ReverseDiffSparse.ParametricExpressionWithParams,ReverseDiffSparse.ParametricExpression{0}))
+@compat function getValue(x::Union{ReverseDiffSparse.ParametricExpressionWithParams,ReverseDiffSparse.ParametricExpression{0}})
```
mlubin: What's the effect of applying ``@compat`` to the whole function? Would it be better to do:
```
function getValue(x::@compat(Union{ReverseDiffSparse.ParametricExpressionWithParams,ReverseDiffSparse.ParametricExpression{0}}))
```

yuyichao: I did it this way mainly because it adds less when there's [multiple `Union`'s](https://github.com/JuliaOpt/JuMP.jl/pull/586/files#diff-188d49b6a9fe32e3c270a948c830cbcfL443) (or other things to be transformed) in one line. I think both of them should't have too much issue in general and I can't say if one of them is better than the other.

mlubin:
Thanks!

Issue #346, opened 2015-01-07
Use staged functions in ``@variable``


joehuchette:
Currently we don't recognize this as something we can store as a `JuMPArray`:

``` julia
I = 1:3
@defVar(m, x[I])
```

It should be fairly straightforward to add a staged function in the generated macro code that takes the `JuMPArray` branch based on the type of `I`.

mlubin:
As I alluded to in #337 we can't easily implement this until we move over to the dormant ``JuMPArray.jl`` implementation because we're currently generating custom types inside of the macros. That's sketchy enough as it is, I don't think we should try to generate types at runtime within staged functions.

joehuchette:
Should we try to bump the ntuple inlining thing?

mlubin:
Not sure if it will be helpful, it's already a well-known issue.

mlubin:
Don't think this is needed for JuMP 0.10, it'll be easier to do after we drop support for 0.3.

joehuchette:
I meant to tag it 0.15. Thinking about this more, we won't be able to return a ``Vector{Variable}`` for my ``I=1:3`` example above without checking the endpoint at runtime, which will ruin type stability. The best we could do with generated functions is to return a JuMPArray. Is it still worth it? If this is mostly about performance, rather than semantics, type instability seems like a show-stopper.

mlubin:
We should be looking at the new arbitrary-indexed arrays for that: https://github.com/JuliaLang/julia/pull/16260
So definitely after we drop support for 0.4

chriscoey:
any new thoughts on this? #933 was added to 1.0 milestone, should this be also?

joehuchette:
d032934b86197c6814973d8d7679f5368eb42a66

PR #589, opened 2015-09-27
Use staged functions in ``@defVar``

joehuchette:
Ref #346, this adds a runtime version of our ``@gendict`` macro so that
```jl
@defVar(m, x[1:3])
```
and
```jl
T = 1:3
@defVar(m, x[T])
```
produce the same object (that is, ``x::Vector{Variable}``). This implementation is type unstable, which I think is inevitable, given that we produce different types for ``T=1:3`` and ``T=2:3``. In the benchmarks updates I pushed, this lead to about a 4x performance hit in the worst-case, though there may be more serious performance hits in other settings.

This PR takes the current behavior, which has opaque difference in semantics, with something that has the same behavior, but opaque differences in performance. If the performance hit is just 4x, I think this may be a reasonable change (and could get better without all the evaling once we switch to proper ``JuMPArray``s), but I'm not certain.

### DIFF
--- a/src/JuMPContainer.jl
+++ b/src/JuMPContainer.jl
@@ -69,7 +69,7 @@ Base.isempty(d::JuMPContainer) = isempty(_innercontainer(d))
 # 0:K -- range with compile-time starting index
 # S -- general iterable set
 export @gendict
-macro gendict(instancename,T,idxpairs,idxsets...)
+macro gendict(instancename,T,idxsets...)
     N = length(idxsets)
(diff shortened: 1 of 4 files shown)


mlubin:
Why don't we just switch to the JuMPArray code now instead of further hacking? Did you want to get this into the 0.10 series?

joehuchette:
This seems a little too feature-y to put in 0.10.x. I'm fine trying to change JuMPArrays now, but that won't subsume this, just change some of the implementation.

PR #592, opened 2015-09-28
getObjective method implemented.

Vgrunert:


src/JuMP.jl
```diff
@@ -199,6 +199,8 @@ end
 @Base.deprecate getNumVars(m::Model) MathProgBase.numvar(m)
 @Base.deprecate getNumConstraints(m::Model) MathProgBase.numlinconstr(m)
 
+getCost(m::Model) = m.redCosts
```
mlubin: This is already available through ``getDual(::Variable)``

Vgrunert: Not exactly. The getDual returns the costs of one variable, whereas the redCosts returns the costs associated with the entire model.

mlubin:
Could you describe a bit what you're trying to do and why these new methods are needed?

src/JuMP.jl
mlubin:
```
n = MathProgBase.numvar(model)
z = zeros(n)
for i in 1:n
    z[i] = getDual(Variable(model,i))
end
```
Accomplishes this using the existing documented methods.

Vgrunert:
Yes. The getObjective function was added simply because it was not implemented so far. I tried to use it but could not, so I added it. The getCost function is useful to get an overview of the reduced costs / shadow costs of the entire model, compared to the getDual function which provides the costs of a particular variable.

src/JuMP.jl
Vgrunert: Fair enough. You decide of course. I just thought that it's a nice convenience function. But please consider the getObjective function, as it is being exported but not implemented so far.

mlubin:
Yes, exported but not implemented is a bit bizarre, not sure how that happened. @IainNZ

Vgrunert:
Well Thanks for your consideration. I was also thinking about adding a few docstrings in the near future, so until then all the best.

mlubin:
I think it's a good idea to provide ``getObjective``, along with way to access constraints (https://github.com/JuliaOpt/JuMP.jl/issues/462)

IainNZ:
Yes to getObjective, no to getCost, docstrings is good but lets hold off for now until we think about style.

Vgrunert:
Fair enough.

joehuchette:
I think we should merge da361e5829421ce53c015c371601668cd7c35046

mlubin:
Needs docs and tests and handling for nonlinear objectives

Vgrunert:
Sry, I am not familiar with github. Is there anything on my part that needs to be done? I though that the request is already out and has to be accepted.

mlubin:
You can make changes to your PR at any time by modifying the corresponding branch on your fork (in this case ``master``). If you're up to it you could tidy the PR by removing the ``getCost`` changes and adding documentation for ``getObjective``, otherwise we'll do it when we get a chance.

PR #593, opened 2015-09-29
remove deprecated comparison operators for building constraints

mlubin:
Removing code is very satisfying

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -235,57 +235,8 @@ end
 # SOCExpr--QuadExpr
 # SOCExpr--SOCExpr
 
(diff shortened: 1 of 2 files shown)


IainNZ:
Are we losing any test coverage?

mlubin:
Not all combinations of the new meaning of ``==`` are covered, but I figured we could do that later.

IainNZ:
I'm more worried about the old meaning, the removed tests. We can still test those right, with a macro added?

mlubin:
Those are already there

mlubin:
Look right below all of the removed tests

IainNZ:
Hah, of course, sorry

Issue #587, opened 2015-09-23
Simpler syntax to relax integer variables


adowling2:
I'd like to see simpler syntax (one line?) to relax/unrelax all integer variables in a MI(N)LP. For example, 

"relax(m)" and "unrelax(m)" where m is a JuMP model

(P.S. My apologies if I placed this feature request in the wrong place.)

The suggestion offered here (https://groups.google.com/forum/#!topic/julia-opt/L6gn60ex6ww) could be used to relax/unrelax individual variables.

mlubin:
I'm not sure I want to store an extra state variable in the JuMP model representing whether we're relaxed or not. What about a ``solve_relaxed`` which solves the continuous relaxation of the given model?

adowling2:
```solve_relaxed``` sounds even better than ```relax(m)``` and ```unrelax(m)```

yeesian:
@JackDunnNZ suggested `solve(m, relaxation=true)`.

mlubin:
That doesn't sound too bad either. Let's keep bikeshedding it

mlubin:
``relaxation`` doesn't quite specify *which* relaxation, would be nice to have this more explicit.

yeesian:
Also from jack: 
ampl has `option relax_integrality`

joehuchette:
+1 to ``relaxation=true``

mlubin:
I kinda like ``relax_integrality`` more

joehuchette:
What about SOS constraints?

mlubin:
Anyone who knows what SOS constraints are shouldn't be too confused

mlubin:
Are you asking if we need to explicitly add in the relaxation of the SOS constraints?

joehuchette:
My point is that there's a natural relaxation for everything JuMP supports, but it doesn't necessarily correspond to relaxing integrality when it comes to SOS or semicontinuous variables, for example.

mlubin:
Relaxation means lots of different things to different people. It's clear in the context of MIP that it means relaxing integrality/SOS/etc, but in the context of nonconvex (MI)QCQP it might be an SDP relaxation. ``solve(m, relaxation=true)`` seems too ambiguous.

joehuchette:
Maybe ambiguous, but it's the most natural choice, and it's not misleadingly immaculate like ``relax_integrality``

mlubin:
``solve(m, continuous_relaxation=true)``?

[...]

PR #597, opened 2015-10-05
Add relaxation kwarg to solve

joehuchette:
Ref #587, still open to a bikeshed to the naming convention if @IainNZ wants to chime in.


test/model.jl
```diff
@@ -738,3 +738,23 @@ facts("[model] Test getValue on OneIndexedArrays") do
...
+    m = Model()
+    @defVar(m, 1 <= x <= 2, SemiCont)
+    @defVar(m, 1.5 <= y <= 2, Int)
+    @defVar(m, z, Bin)
+    @defVar(m, 0.5 <= w <= 0.5, Int)
+    @defVar(m, 1 <= v <= 2)
+    @defVar(m, -2 <= t <= -1, SemiInt)
+
+    @setObjective(m, Min, x + y + z + w + v - t)
+    solve(m, relaxation=true)
```
mlubin: What happens if you call ``solve()`` again with ``relaxation=false``? Do we clear the relaxation data?

joehuchette: Seems to work (returns infeasible with CPLEX).

mlubin: For simplicity, what about setting ``internalModelLoaded=false`` after a solve with ``relaxation=true``?

joehuchette: Works with me

test/model.jl
```diff
@@ -738,3 +738,25 @@ facts("[model] Test getValue on OneIndexedArrays") do
...
+    @fact getValue(x) --> 0
+    @fact getValue(y) --> 1.5
+    @fact getValue(z) --> 0
+    @fact getValue(w) --> 0.5
+    @fact getValue(v) --> 1
+    @fact getValue(t) --> 0
+    @fact getObjectiveValue(m) --> 0 + 1.5 + 0 + 0.5 + 1 + 0
```
mlubin: Could you add tests with some switching between ``true`` and ``false``?

mlubin:
The name ``relaxation`` still doesn't sit well with me. It will definitely send some people to the docs asking "which relaxation?"

joehuchette:
Not sure that's a bad thing, or avoidable, really. I'm warming up to ``continuous_relaxation``, though it doesn't fix the "lookup in the docs" issue.

IainNZ:
I think there is no fixing the "lookup in docs" situation. But, of course, someone would have to look in the docs to find it in the first place. I'd say relaxing integrality is by the most common usage for people who use JuMP, and I don't think it'll be scrutinized heavily. If we get into more general relaxations later, we can cross that bridge then.

One option is `relaxation=:Continuous` I guess if we want to avoid booleans.

src/solvers.jl
```diff
@@ -168,6 +170,8 @@ function solve(m::Model; suppress_warnings=false,
         end
     end
 
+    relaxation && (m.internalModelLoaded = false)
```
mlubin: Maybe add a comment here like: ``# don't keep relaxed model in memory``? I had trouble reading this line on the first try.

mlubin:
Let's just go with ``relaxation=true`` for now, can always deprecate later. Needs docs before merging.

doc/refmodel.rst
```diff
@@ -43,7 +43,7 @@ Methods
...
-* ``solve(m::Model;  suppress_warnings=false)`` - solves the model using the selected solver (or a default for the problem class), and takes two optional arguments that are disabled by default. Setting ``suppress_warnings`` to ``true`` will suppress all JuMP-specific output (e.g. warnings about infeasibility and lack of dual information) but will not suppress solver output (which should be done by passing options to the solver).
+* ``solve(m::Model; suppress_warnings=false, relaxation=false)`` - solves the model using the selected solver (or a default for the problem class), and takes two optional arguments that are disabled by default. Setting ``suppress_warnings`` to ``true`` will suppress all JuMP-specific output (e.g. warnings about infeasibility and lack of dual information) but will not suppress solver output (which should be done by passing options to the solver). Setting ``relaxation=true`` solves the standard continuous relaxation for the model: that is, integrality is dropped, SOS constraints are not enforced, and semi-continuous and semi-integer variables with bounds ``[l,u]`` are replaced with bounds ``[min(l,0),max(u,0)]``.
```
mlubin: Should clarify that SOS is special ordered set. Many people will read it as sum of squares.

PR #603, opened 2015-10-25
Fix test and 0.5 breakage

yuyichao:
* Fix test to support solvers in `LOAD_PATH`. (Ref https://github.com/JuliaOpt/MathProgBase.jl/pull/89)
* Support the new parsing of `in` (Ref https://github.com/JuliaLang/julia/pull/13078 and https://github.com/dcjones/Compose.jl/pull/164)

    The tests pass locally without the fix for `in` so I suspect it is not covered in the test.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -39,11 +39,15 @@ function buildrefsets(expr::Expr)
         end
     end
     for s in c.args
-        if isa(s,Expr) && (s.head == :(=) || s.head == :in)
-            idxvar = s.args[1]
-            idxset = esc(s.args[2])
-            push!(idxpairs, IndexPair(s.args[1],s.args[2]))
-        else
+        parse_done = false
+        if isa(s, Expr)
+            parse_done, idxvar, _idxset = tryParseIdxSet(s::Expr)
+            if parse_done
+                idxset = esc(_idxset)
+                push!(idxpairs, IndexPair(idxvar, _idxset))
+            end
+        end
+        if !parse_done
             idxvar = gensym()
             idxset = esc(s)
             push!(idxpairs, IndexPair(nothing,s))
--- a/src/parseExpr_staged.jl
+++ b/src/parseExpr_staged.jl
@@ -3,7 +3,29 @@
 #  License, v. 2.0. If a copy of the MPL was not distributed with this
 #  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-#addToExpression(ex::Number, c::Number) = ex + c
+# `a in b` is a comparison after JuliaLang/julia#13078
+const in_is_compare = VERSION >= v"0.5.0-dev+901"
+
+function tryParseIdxSet(arg::Expr)
+    if arg.head === :(=) || (!in_is_compare && arg.head === :in)
+        @assert length(arg.args) == 2
+        return true, arg.args[1], arg.args[2]
+    elseif (in_is_compare && arg.head === :comparison &&
+            length(arg.args) == 3 && arg.args[2] === :in)
+        return true, arg.args[1], arg.args[3]
+    end
+    return false, nothing, nothing
+end
+
+function parseIdxSet(arg::Expr, msg="Invalid loop condition")
+    parse_done, idxvar, idxset = tryParseIdxSet(arg)
+    if parse_done
+        return idxvar, idxset
+    end
+    error(msg)
+end
+
+# addToExpression(ex::Number, c::Number) = ex + c
 
 addToExpression(ex::Number, c::Number, x::Number) = ex + c*x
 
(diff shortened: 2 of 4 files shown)


yuyichao:
I only have the bare minimum solvers installed to finish running the test so it's entirely possible that the `in` expression is covered but I couldn't run the test locally.....

mlubin:
Thanks for the fix. Which case is affected by the parsing change? Could you add a test?

src/macros.jl
```diff
@@ -39,10 +39,17 @@ function buildrefsets(expr::Expr)
...
     end
     for s in c.args
-        if isa(s,Expr) && (s.head == :(=) || s.head == :in)
+        if isa(s, Expr) && ((s::Expr).head === :(=) || (s::Expr).head === :in)
+            s = s::Expr
             idxvar = s.args[1]
             idxset = esc(s.args[2])
-            push!(idxpairs, IndexPair(s.args[1],s.args[2]))
+            push!(idxpairs, IndexPair(s.args[1], s.args[2]))
+        elseif (isa(s, Expr) && (s::Expr).head === :comparison &&
```
mlubin: Better to wrap this in ``VERSION`` conditions so that we know to remove a branch when we drop support for 0.4

yuyichao:
> Thanks for the fix. Which case is affected by the parsing change? Could you add a test?

Unfortunately, I'm not familiar with JuMP enough to test that. I found this when doing a grep after submitting dcjones/Compose.jl#164 (and I think this should be affected since `(e::Expr) === :in` should not happen anymore).

yuyichao:
Version check added.

mlubin:
I'd like to find a failure case before we merge this.

yuyichao:
You are right. When trying to make such a test (by modifying existing ones) I noticed that there're other places in the code base that is implicitly assuming `:(a in b) == Expr(:in, :a, :b)`. I'll try to add tests and fix those too ....

mlubin:
The cases that seem affected are ``sum{ x[i], i in X }``.

yuyichao:
That is the case that is not yet fixed by this PR which is implicitly assuming what `a in b` looks like.

mlubin:
How is it possible that this doesn't cause any tests to fail?

yuyichao:
Because that syntax is not tested?

yuyichao:
The test failure when this triggers is also pretty obscure ....

```
ERROR: LoadError: LoadError: MethodError: `length` has no method matching length(::Function)
 [inlined code] from /usr/share/julia/site/v0.5/JuMP/src/parseExpr_staged.jl:331
 in anonymous at /home/yuyichao/projects/contrib/JuMP.jl/test/macros.jl:311
 in facts at /usr/share/julia/site/v0.5/FactCheck/src/FactCheck.jl:448
while loading /home/yuyichao/projects/contrib/JuMP.jl/test/macros.jl, in expression starting on line 308
while loading /home/yuyichao/projects/contrib/JuMP.jl/test/runtests.jl, in expression starting on line 24
```

yuyichao:
I update the PR to fix other implicit assumptions about `Expr(:in, ...)`. (I did a `git grep '\.args'`) so hopefully this is the complete list. I've also added a few assertions so hopefully there can be slightly better error messages when there's certain parser changes.

I've added one test for this. This is definitely not enough (e.g. it doesn't trigger the fix in the original PR AFAICT) so I'll see if I can add more.....

yuyichao:
OK. I've updated the test and verified that all the changes in this PR are required to make the test pass (at least one of the test will fail if any of the use of `tryParseIdxSet` or `parseIdxSet` is reverted). I've also add it to the parsing test directly to make sure the expression is always parsed in the way we expected.

test/solvers.jl
```diff
@@ -12,19 +12,28 @@
...
+glp = package_installed(:GLPKMathProgInterface)
+ipt = package_installed(:Ipopt)
+nlo = package_installed(:NLopt)
+kni = package_installed(:KNITRO)
+eco = package_installed(:ECOS)
+osl = package_installed(:CoinOptServices)
+scs = package_installed(:SCS)
+nlw = package_installed(:AmplNLWriter)
 # Load them
 if grb; import Gurobi; end
```
IainNZ: Aren't all these redundant now? If the package is installed, it was `import`ed by the above code.

yuyichao: Should I remove them except `Cbc`/`Clp` below?

IainNZ: I guess so, unless I'm not understanding something about this change

yuyichao: I renamed the function to make it clear that it actually load the package instead of just checking its existence and removed most of the import below.

yuyichao:
CI all pass. ;-)

src/parseExpr_staged.jl
```diff
@@ -3,7 +3,29 @@
...
+    end
+    return false, nothing, nothing
+end
+
+function parseIdxSet(arg::Expr, msg="Invalid loop condition")
+    parse_done, idxvar, idxset = tryParseIdxSet(arg)
+    if parse_done
+        return idxvar, idxset
+    end
+    error(msg)
```
mlubin: I'd prefer an error message like ``invalid syntax $arg``. Given that the ``msg`` argument isn't used, it can be dropped.

mlubin:
I'll tidy up and take it from here. Thanks for the fix!

src/parseExpr_staged.jl
yuyichao: That makes sense. I add a `msg` parameter because I saw many places that needs to be fixed. Turns out they are almost the same so I end up didn't have to supply different messages for different ones.

joehuchette:
Thanks, @yuyichao!

Issue #179, opened 2014-05-13
Rewrite operator overloads using promotion rules


mlubin:
It seems like we could cut down a lot of the operator overloading code and make Julia's generic operations work correctly by implementing promotion rules. 

Ref JuliaLang/Julia#6672

mlubin:
Is this still useful? Did we come to the conclusion that promotion rules don't quite work for the case where ``a*b`` has a different type from ``a+b``?

mlubin:
Not much going on here. We'll reconsider if we find a new reason to refactor the operator overloads.

PR #604, opened 2015-10-28
DNMY: rely on promote_op for many operations

timholy:
Requires backporting of https://github.com/JuliaLang/julia/pull/13803, do not merge yet. Once it gets backported, do not merge until the REQUIRE file has its julia version bumped. (I'm not sure which julia version it will end up in, or I'd add that change now.)

I noticed a few "Array"-type operations missing. While it would have been easier to write specialized methods to handle the operations I needed, it seemed best to try to get to the point where we supply the right traits and then rely on methods in Base.

### DIFF
--- a/test/operator.jl
+++ b/test/operator.jl
@@ -465,7 +465,7 @@ module TestHelper # weird scoping behavior with FactCheck...
         return x
     end
 
-    vec_eq(x,y) = vec_eq([x;], [y;])
+    vec_eq(x,y) = vec_eq(convert(Array, [x;]), convert(Array, [y;]))
 
     function vec_eq(x::Array, y::Array)
         size(x) == size(y) || return false
@@ -575,6 +575,14 @@ context("Vectorized arithmetic") do
                                      x[2] + 3x[3]]) --> true
 
     @fact TestHelper.vec_eq(@defExpr(A*x/2), A*x/2) --> true
+
+    x2 = x[1:2]
+    Xcov = [x[1]^2    x[1]*x[2];
+            x[1]*x[2] x[2]^2]
+    @fact TestHelper.vec_eq(x2 .* x2', Xcov) --> true
+
+    X = reshape(x2, 2, 1)
+    @fact TestHelper.vec_eq(X * X', Xcov) --> true
 end
 
 context("Dot-ops") do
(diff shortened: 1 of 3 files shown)


mlubin:
Very cool, thanks! Ref #179
Could we also shift to this approach for scalar operations?

joehuchette:
This looks great, thanks!

test/operator.jl
```diff
@@ -465,7 +465,7 @@ module TestHelper # weird scoping behavior with FactCheck...
         return x
     end
 
-    vec_eq(x,y) = vec_eq([x;], [y;])
+    vec_eq(x,y) = vec_eq(convert(Array, [x;]), convert(Array, [y;]))
```
joehuchette: Why is this change needed?

timholy:
I think this basically closes #179.

I'm not exactly certain what you mean by shifting for scalar operations; unfortunately right now you have to basically implement scalar operations for both instances and types, e.g.,
```jl
(+)(lhs::Number, rhs::GenericAffExpr) = GenericAffExpr(copy(rhs.vars),copy(rhs.coeffs),lhs+rhs.constant)
Base.promote_op{R<:Real}(::AddFun, ::Type{R}, ::Type{AffExpr}) = AffExpr
```
But all those are scalar objects. Once you do that, the array operations follow automatically. (And you really only need the type-implementation for use in the array operations.)

Someday, someone will write a `ReturnTypes.jl` package that is an independent implementation of type-inference. Once it's working well, many of us will lobby to put this in base, and if that happens then we won't need the type-implementation. See the extensive discussion in https://github.com/JuliaLang/julia/pull/12292#issuecomment-124322950.

test/operator.jl
timholy: `isa(x, SparseMatrixCSC) == true`

mlubin:
@timholy, I mean falling back on promotion rules to convert Number + AffExpr into AffExpr + AffExpr. Though there's some loss of efficiency there.

timholy:
Oh, now I understand. That could be good for `+` and `-`, but (to state what's certainly obvious to you) I think for `*` you'd lose important type information, since `Number*AffExpr -> AffExpr` and not `QuadExpr`. Not sure it would be worth it for just `+` and `-`.

mlubin:
@timholy, any interest in updating this PR?

mlubin:
This is stale, but we still have https://github.com/JuliaOpt/JuMP.jl/issues/179 open

PR #605, opened 2015-10-29
Unrecognized keyword arguments to solve should be handled

joehuchette:


joehuchette:
```
julia> solve(m, this_should_throw=true)
ERROR: Unrecognized keyword arguments Any[(:this_should_throw,true)]
```

joehuchette:
Open to suggestions

mlubin:
Ah, I see the issue. I think this is okay, maybe improve the printing a bit

joehuchette:
Open to suggestions on printing. Though it's breaking, this should maybe be backported, I got tripped up trying to pass ``relaxation=true`` on release-0.10.

mlubin:
Just print out the names of the unrecognized keywords instead of ``Any[(:this_should_throw,true)]``?

PR #610, opened 2015-11-05
Make solve/buildInternalModel typestable

joehuchette:
I doubt this was causing any performance issues, but this seems like a reasonable restriction to me.


mlubin:
LGTM

PR #612, opened 2015-11-05
Change `iround` to `round` in docs

JackDunnNZ:
Not sure how much longer these docs will be around, but may as well get rid of deprecated things...


joehuchette:
:+1:

PR #613, opened 2015-11-05
allow access to constraint bounds without going though MPB

mlubin:
Based on user feedback, forcing everyone to go through MathProgBase (i.e., implementing a dummy solver) to get the full specification of the problem seems to be an extra level of complexity that's not always needed, e.g., for demonstrating how to implement nonlinear optimization solvers in a class setting. I've been worried about potential misuse of allowing people to avoid MPB, but I made it clear in the docs that using this API is not the right way to make a real solver.

Open to bikeshedding the name.

CC @dpo

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -498,9 +498,8 @@ function MathProgBase.constr_expr(d::JuMPNLPEvaluator,i::Integer)
     end
 end
 
-function _buildInternalModel_nlp(m::Model, traits)
 
-    @assert isempty(m.sdpconstr) && isempty(m.socconstr)
+function _buildInternalModel_nlp(m::Model, traits)
 
     linobj, linrowlb, linrowub = prepProblemBounds(m)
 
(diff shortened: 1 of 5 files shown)


mlubin:
CC @tkelman

dpo:
That's really useful. But why not just `getConstraintBounds`? The current name suggests I won't get the bounds on the linear constraints.

mlubin:
``getConstraintBounds`` is probably a better name, I just don't know (and don't want to think about) what the method should do outside of the context of NLP. Specifically, we haven't yet defined any interaction or ordering between conic, SOS, and nonlinear constraints. We could just throw an error in this case until we've decided how to define it.

dpo:
I see. Yes I suppose a message indicating that the functionality isn't yet implemented for SOCP, etc., is appropriate. Will the function also return bounds on the linear constraints though?

mlubin:
Yes, it does

mlubin:
Renamed to ``getConstraintBounds``. Docs aren't good but maybe we'll put that off for the upcoming rewrite.

mlubin:
@IainNZ, @joehuchette, :+1: or :-1: ?

joehuchette:
:+1:

src/nlp.jl
```diff
@@ -498,32 +498,21 @@ function MathProgBase.constr_expr(d::JuMPNLPEvaluator,i::Integer)
     end
 end
 
-function _buildInternalModel_nlp(m::Model, traits)
-
-    @assert isempty(m.sdpconstr) && isempty(m.socconstr)
+getConstraintBounds(m::Model) = getConstraintBounds(m, ProblemTraits(m))
 
-    linobj, linrowlb, linrowub = prepProblemBounds(m)
```
joehuchette: This probably shouldn't go in nlp.jl

src/nlp.jl
```diff
@@ -498,32 +498,21 @@ function MathProgBase.constr_expr(d::JuMPNLPEvaluator,i::Integer)
...
+function getConstraintBounds(m::Model,traits::ProblemTraits)
```
joehuchette: This probably shouldn't go in nlp.jl

mlubin: The logic of flattening out constraints into linear then quadratic then nonlinear has so far been constrained to ``nlp.jl``, I think.

joehuchette: Seems like kind of a stretch...

mlubin: Fair enough

IainNZ:
:+1:

mlubin:
Moved the new code to ``solvers.jl``

PR #617, opened 2015-11-15
implement getObjective()

mlubin:
Here you go, @Vgrunert


mlubin:
An issue came up here where ``quad != quad.aff`` even when the quadratic part is empty, although arguably they should compare as equal. That's a separate issue though.

Also funny is that this method was already documented!

PR #618, opened 2015-11-18
Fix typo in mindistortion.jl example problem documentation

goretkin:


mlubin:
Thanks! Haven't looked at this in a while...

goretkin:
;) stumbled by accident upon it...

PR #619, opened 2015-11-19
fix Variable constructor

mlubin:
https://groups.google.com/forum/#!topic/julia-opt/BzkfyW5xNyc

There's no performance issue with allowing wider arguments, it's just our job to provide a ``UTF8String`` from the macros if we want to avoid conversions.


joehuchette:
Any idea why this was failing?

mlubin:
@joehuchette, the default value ``""`` isn't a ``UTF8String``

joehuchette:
Were these not getting called by the macros, so that we only noticed they were broken when someone tried to call them explicitly?

mlubin:
Yep, wasn't tested.

PR #630, opened 2015-12-07
hacks to support Mosek's new MPB update

mlubin:
CC @ulfworsoe


mlubin:
Mosek still isn't passing tests, see https://github.com/JuliaOpt/Mosek.jl/pull/67

mlubin:
Seems safe to merge this

PR #631, opened 2015-12-08
add fallback for appending constants to AffExprs

mlubin:


IainNZ:
If we're doing this, which is fair enough, should maybe add the `::V` version too

mlubin:
I guess that wouldn't hurt.

IainNZ:
Would be a confusion local minimum I figure

mlubin:
Done

Issue #633, opened 2015-12-09
AffExpr iteration doesn't play nicely with mapreduce


joehuchette:
As noticed on the [mailing list](), the current behavior for iteration over an AffExpr returning (coeff,var) doesn't work with the fallbacks for things like sum, mapreduce, etc.:
```jl
using JuMP
m = Model()
@defVar(m, x)
sum(2x) # => (2.0,x)
```
cc @IainNZ since I think you added this behavior?

mlubin:
I think we should deprecate the current behavior. How about something like:
```
for (coeff,var) in linearterms(aff)
...
end
```

IainNZ:
I like that proposal.
So we should remove the iteration protocol? What will `sum(2x)` do then?

mlubin:
Should probably behave like iterating over a scalar, so ``sum(2x)`` will return ``2x``.

joehuchette:
Closed by #636

PR #636, opened 2015-12-22
Change AffExpr iteration (#633)

joehuchette:
Changes:
* Remove idiosyncratic AffExpr iteration
* Remove in(::Variable, ::AffExpr)
* Make AffExpr, QuadExpr <: AbstractJuMPScalar

The first seemed uncontroversial, but I'm not sure how the other two will go over. I don't think it's possible to "deprecate" iteration if we want to replace the behavior with "number-like" iteration instead.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,11 @@
 JuMP release notes
 ==================
 
+Unversioned
+-----------
+
+  * Replaced iteration over ``AffExpr`` with ``Number``-like scalar iteration; previous iteration behavior is now available via ``linearterms(::AffExpr)``.
+
 Version 0.11.1 (December 1, 2015)
 ---------------------------------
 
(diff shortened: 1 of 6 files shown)


mlubin:
No deprecation period?

joehuchette:
I don't see how to do that without removing iteration completely, which will break things like ``sum(2x)`` anyway. I don't see many users relying on this behavior in its current state.

mlubin:
What happened to keeping the iteration behavior as ``linearterms(aff)``? Even if there are a small number of people using this, breaking their code with no useful error message *and* forcing them to restructure the code to access the members of ``AffExpr`` is a bit rough.

joehuchette:
I'm fine with keeping the functionality under a different name (though I think it should iterate over a "reduced" version of the expression), but I still don't see how a deprecation would fit into this.

mlubin:
We could decide to deprecate it with a warning to use ``linearterms`` instead, and then in the next release change the behavior. ``sum(2x)`` is already broken, so it's a balance between breaking existing code and fixing this bug. I could be convinced to just fix the bug.
CC @IainNZ

IainNZ:
I'd be surprised if someone other than me is using it, possibly people who have written oracles for JuMPeR? But, thats so experimental that I think its OK to break (people can pin package versions, after all)

joehuchette:
If it were just a matter of adding a temporary ``sum(::AffExpr)`` method that would be one thing, but I think any Base code that uses an iteration fallback will be similarly broken, so I'd rather just fix it now.

mlubin:
Fair enough.

Issue #627, opened 2015-12-05
setName is not working well on this example


emreyamangil:
The following example doesn't print correctly the labels of the variables, is this a known issue? many thanks!!!

julia> using JuMP

julia> using CPLEX

julia> numBins = [2,3,1]
3-element Array{Int64,1}:
 2
 3
 1

julia> m = Model(solver=CplexSolver())
Feasibility problem with:
 * 0 linear constraints
 * 0 variables
Solver set to CPLEX.Cplex

julia> N = 3
3

julia> t = Array(Vector{Variable},N)
3-element Array{Array{JuMP.Variable,1},1}:
 #undef
 #undef
 #undef

julia> for k in 1:N
               @defVar(m, t[k][j=1:numBins[k]], Bin)
       end

julia> for k = 1:N
       for j = 1:numBins[k]
       setName(t[k][j], "t[$k][$j]")
       end
       end

julia> println(m)
Min 0
Subject to
 t[k][j] ∈ {0,1} ∀ j ∈ {1,2}
 t[k][j] ∈ {0,1} ∀ j ∈ {1,2,3}
 t[k][j] ∈ {0,1} ∀ j ∈ {1}

mlubin:
I think this is a bug, but you can do
```
@defVar(m, t[k=1:N,1:numBins[k]], Bin)
```
instead.

emreyamangil:
Thanks Miles!! Will certainly switch to this asap!

mlubin:
@joehuchette, another issue here is that the base name for the ``JuMPArray`` shouldn't be ``"t[k]"``

joehuchette:
That's true, but that seems a much trickier thing to handle corrrectly

emreyamangil:
Hey Joey and Miles! First of all many thanks for your time, though I have bad news :(

The following works with setName

``````````````````````````````````
using JuMP



type mymodel
        x
        m
        myx
        function mymodel()
                m = new()
                m.m = Model()
                m.x = Any[]
                return m
        end
end

function addvar(m::mymodel, myind)
        push!(m.x, @defVar(m.m, x[1:10, 1:2]))
        for i = 1:10
                for j = 1:2
                        setName(x[i,j], "x[$myind][$i,$j]")
                end
        end
        @addConstraint(m.m, sum{x[i,1], i = 1:10} <= 10)
        @addConstraint(m.m, sum{x[i,2], i = 1:10} <= 5)
[...]

PR #637, opened 2015-12-26
Short-circuit compact printing if user provides a custom names

joehuchette:
Ref #627. I'm not super pleased with this solution, but I can't think of a more reasonable approach. It shouldn't affect performance except in the case where you call ``setName`` frequently.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -74,6 +74,8 @@ type Model
     colUpper::Vector{Float64}
     colCat::Vector{Symbol}
 
+    customNames::Vector
+
     # Variable cones of the form, e.g. (:SDP, 1:9)
     varCones::Vector{Tuple{Symbol,Any}}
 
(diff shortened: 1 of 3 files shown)


mlubin:
Is there any way we can avoid storing extra state inside ``Model``?

joehuchette:
Open to suggestions

mlubin:
What about removing the corresponding entry from ``dictList`` if any elements of a JuMP container are renamed?

joehuchette:
I can try, not sure how much else that'll break

joehuchette:
Not sure how to do that, actually, since you call ``setName(::Variable, ::String)``, we'd have to look up the corresponding JuMPContainer somehow

mlubin:
We could call ``fill_var_names`` the first time that someone calls ``setName``

mlubin:
then clear ``dictList``

mlubin:
The reason we have ``dictList`` is that we want to delay generating the names if possible. But if someone calls ``setName`` it means that we'll likely need to generate the names anyway, so no harm in doing it sooner.

joehuchette:
AFAICT we need ``dictList`` (or something like it) to do the compacted printing like ``x[i] <=1, i in 1:3``. Note that, if you call ``setName(x[2], "foo")``, the compacted printing no longer works.

mlubin:
Ok, not opposed to the current fix

joehuchette:
Not sure why appveyor is failing, but it looks unrelated (and scary).

PR #638, opened 2015-12-28
RFC: NLP rewrite using ReverseDiffSparse2

### LINKED ISSUE #304: Add support for parameters
I would actually like to see support for model parameters, a la:

````
@defPar(m, p1)
````

Then ``p1`` could be used in any algebraic expression as if it was a constant, but the value for it could be changed before a ``solve(m)`` call. So I guess ``setValue`` would have to work with ``p1``, but that is probably all.

The scenario I have is that I need to solve the same model very, very many times, but each time I want to vary one of the parameters of the model.

In my mind there is no need for the underlying solver to even know that ``p1`` is not just a constant, so I have no need for duals or anything like that for ``p1``. Microsoft's solver foundation had support for this and it was really handy.

### LINKED ISSUE #414: bounds checking on nonlinear expression indices
I've noticed some counter-intuitive behavior when accessing nonlinear expressions using single indices.  It seems that you are allowed to set indexed nonlinear expressions one at a time and sum them, but that later assignments overwrite the whole array.  Here's an example:
```
m = Model()
@defVar(m, baz)
@defNLExpr(foo[1], 5 * baz)
@defNLExpr(foo[2], -3 * baz)
@defNLExpr(bar, sum{foo[i], i=1:2})
setValue(baz, 2)
ReverseDiffSparse.getvalue(bar, m.colVal) # Expect 4, get -12
```
You get different behavior with ordinary expressions:
```
m = Model()
@defVar(m, baz)
@defExpr(foo[1], 5 * baz)
@defExpr(foo[2], -3 * baz)

julia> foo[1]
ERROR: key not found: (1,)
 in getindex at /home/rgiordan/.julia/v0.3/JuMP/src/JuMPDict.jl:30

julia> foo[2]
-3 baz
```
It's not hard to avoid using this kind of expression, though it would be convenient.  For example, this works:
```
m = Model()
@defVar(m, baz)
@defNLExpr(foo[s=1:2], (s == 1) * 5 * baz + (s == 2) * (-3 * baz))
@defNLExpr(bar, sum{foo[i], i=1:2})
setValue(baz, 2)
ReverseDiffSparse.getvalue(bar, m.colVal) 
```
However, I think the single indices should at least fail with an error rather than allowing you to quietly overwrite the rest of the array.

### LINKED ISSUE #474: memory issues with JuMP/Ipopt
I am trying to run a medium scale nonlinear, nonconvex optimization problem with a dense Hessian (approximately 40 decision variables) using JuMP with Ipopt backend. Unfortunately, I run into huge memory issues - memory usage is stable around 3 GB for quite some time, but then all of a sudden shoots up to my physical memory capacity (8 GB), at which point thrashing occurs.
The code is pretty big, so I have kept it as a gist:

https://gist.github.com/gajjanag/97ecb4f452172d64d7d7

Method is invoked via: solve_model().

Package dependencies: JuMP, Ipopt, MathProgBase (obvious), Distributions.

### LINKED ISSUE #567: Trouble with @setNLObjective from
Hi, 

I'm trying to do an optimization with an array of different objectives within JuMP. But I get a type error when running nonlinear solves:
```julia
using JuMP
m = Model()
@defVar(m, 0 <= x <= 1)
@defExpr(f1, -x)
@defNLExpr(f2, -x)

fs = [f1, f2]

@setObjective(m, :Min, fs[1]) #Works
solve(m)

@setNLObjective(m, :Min, f2) #Works
solve(m)

@setNLObjective(m, :Min, fs[2]) #Fails
solve(m)
```
The last line raises the error
```julia
ERROR: type: eval_f: in typeassert, expected Float64, got ParametricExpression{0}
 in eval_f at /home/asbjorn/.julia/v0.3/JuMP/src/nlp.jl:246
 in initialize at /home/asbjorn/.julia/v0.3/JuMP/src/nlp.jl:222
 in loadnonlinearproblem! at /home/asbjorn/.julia/v0.3/Ipopt/src/IpoptSolverInterface.jl:269
 in _buildInternalModel_nlp at /home/asbjorn/.julia/v0.3/JuMP/src/nlp.jl:540
 in buildInternalModel at /home/asbjorn/.julia/v0.3/JuMP/src/solvers.jl:186
 in solve at /home/asbjorn/.julia/v0.3/JuMP/src/solvers.jl:77

```

### LINKED ISSUE #571: NLExpressions and multiplication / division with numbers
I guess this is an issue with the complexity of dealing with nonlinear expressions, so this is more of a note for people who might have the same problems:

The Hessian sparsity changes drastically depending on how you set your objectives in nonlinear optimisation. Division is bad, so is multiplication with something like (1 - a) for some scalar a.
The way around these issues is using multiplication with dummy-variables. In the following, let nf2 be a ParametricExpression defined by @defNLExpr:

```julia
julia> @setNLObjective(m, :Min, nf2*(1-a))
julia> solve(m)
This is Ipopt version 3.11.9, running with linear solver mumps.
NOTE: Other linear solvers might be more efficient (see Ipopt documentation).

Number of nonzeros in equality constraint Jacobian...:        0
Number of nonzeros in inequality constraint Jacobian.:      355
Number of nonzeros in Lagrangian Hessian.............:    61075
*snip*

julia> num = (1-a)
julia> @setNLObjective(m, :Min, nf2*num)
julia> solve(m)
This is Ipopt version 3.11.9, running with linear solver mumps.
NOTE: Other linear solvers might be more efficient (see Ipopt documentation).

Number of nonzeros in equality constraint Jacobian...:        0
Number of nonzeros in inequality constraint Jacobian.:      355
Number of nonzeros in Lagrangian Hessian.............:     1495
```

Likewise, if you want `@setNLObjective(m, :Min, nf2 / a)`. It is better to use `b = 1/a` and  `@setNLObjective(m, :Min, nf2 * b)`

### LINKED ISSUE #585: Incorrect parsing/evaluation of NLExpressions
Using a NLExpr in the definition of another NLExpr does not seem to work correctly all the time.
I've tried to define an expression $v(x)^TAv(x)$, where $v$ is a nonlinear vector field in $x$ and $A$ is a matrix. In the example below, you can see that the evaluation of the functions f2 and f3 are different, while I expect it to be the same.

```julia 
using JuMP
n = 2
x0 = [2.25, 2.4]
β = [-2 0.15; 0.23 -2.1]
Σ = [0.02 0.007; 0.007 0.014]
x0 = [1.5, 2.0]
m = Model()
@defVar(m, x[i=1:n] >= 0, start=x0[i])

@defNLExpr(f1[i=1:n], prod{x[j]^β[j,i], j = 1:n})
@defNLExpr(f2, sum{Σ[i,j]*f1[i]*f1[j], i=1:n, j=1:n}) # bug
@defNLExpr(f3, sum{Σ[i,j]*prod{x[k]^β[k,i], k = 1:n}*prod{x[k]^β[k,j], k = 1:n}, i=1:n, j=1:n})

println(getValue(f2))
println(getValue(f3))
```
Output:
``` julia
0.0021786119842496466
0.008103475980157478
```

### LINKED ISSUE #614: Very slow JuMP + Knitro for nonlinear problem
I started using Julia with Knitro through Julia's JuMP for mathematical programming. I noticed that a small nonlinear problem with 33 variables and just over 300 non-zeros in Hessian takes quite a lot of time to generate and solve. While slow generation can be attributed to Julia's JuMP, Knitro itself takes as much as 15 minutes to solve the above problem, though it takes only 12 iterations? Would you be able to have a look? I can send you the code.

The analogous problem in AMPL+KNITRO takes a few seconds to generate and solve.

mlubin:
The long awaited NLP rewrite is now a ~~WIP~~ RFC. ~~The net 200 line addition is pretty good considering that it replaces the 400+ lines of expression parsing in ReverseDiffSparse.~~

Still to do:
- [x] Nonlinear subexpressions
- [x] Replace ReverseDiffSparse with ReverseDiffSparse2 and tag
- [x] Re-implement hessian-vector products
- [x] Re-implement expression graph output
- [x] Parameters
- [x] Error message on NLP resolve
- [x] Switch to ``NaNMath``
- [x] Deprecation warning for change in ``@defNLExpr`` (*missing backtraces*)
- [x] Support for ``ifelse``
- [x] Support for ``abs``
- [x] ``getValue`` for expressions
- [x] Add unit test for pow domain error
- [x] Docs

CC @anriseth ~~(don't bother testing until subexpressions are working)~~

Closes #614 
Closes #571
Closes #585
Closes #414
Closes #567 
Closes #474 
Closes #304

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -16,14 +16,18 @@ importall Base.Operators
 
 import MathProgBase
 
-using ReverseDiffSparse, Calculus
-import ArrayViews
-const subarr = ArrayViews.view
+using Calculus
+using ReverseDiffSparse
+
+function __init__()
+    ENABLE_NLP_RESOLVE[1] = false
+end
 
 export
 # Objects
     Model, Variable, Norm, AffExpr, QuadExpr, SOCExpr, AbstractJuMPScalar,
     LinearConstraint, QuadConstraint, SDPConstraint, SOCConstraint,
+    NonlinearConstraint,
     ConstraintRef, LinConstrRef,
     JuMPNLPEvaluator,
 # Functions
@@ -48,7 +52,7 @@ export
     @SOCConstraint, @SOCConstraints,
     @defVar, @defConstrRef, @setObjective, addToExpression, @defExpr,
     @setNLObjective, @addNLConstraint, @addNLConstraints,
-    @defNLExpr
+    @defNLExpr, @defNLParam
 
 include("JuMPContainer.jl")
 include("utils.jl")
@@ -296,7 +300,7 @@ setPrintHook(m::Model, f) = (m.printhook = f)
 abstract JuMPConstraint
 # Abstract base type for all scalar types
 # In JuMP, used only for Variable. Useful primarily for extensions
-abstract AbstractJuMPScalar <: ReverseDiffSparse.Placeholder
+abstract AbstractJuMPScalar
 
 Base.start(::AbstractJuMPScalar) = false
 Base.next(x::AbstractJuMPScalar, state) = (x, true)
@@ -312,7 +316,6 @@ immutable Variable <: AbstractJuMPScalar
 end
 
 getLinearIndex(x::Variable) = x.col
-ReverseDiffSparse.getplaceindex(x::Variable) = getLinearIndex(x)
 Base.isequal(x::Variable,y::Variable) = (x.col == y.col) && (x.m === y.m)
 
 Variable(m::Model, lower, upper, cat::Symbol, name::AbstractString="", value::Number=NaN) =
@@ -706,6 +709,18 @@ Base.eltype{T<:JuMPTypes}(::T) = T
 Base.size(::JuMPTypes) = ()
 Base.size(x::JuMPTypes,d::Int) = 1
 Base.ndims(::JuMPTypes) = 0
+
+##########################################################################
+# Types used in the nonlinear code
+immutable NonlinearExpression
+    m::Model
+    index::Int
+end
+
+immutable NonlinearParameter
+    m::Model
+    index::Int
+end
 
 ##########################################################################
 # Operator overloads
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -373,14 +373,6 @@ function setValue(v::Variable, val::Number)
     end
 end
 
-function setValue(set::Array{Variable}, val::Array)
-    promote_shape(size(set), size(val)) # Check dimensions match
-    for I in eachindex(set)
-        setValue(set[I], val[I])
-    end
-    nothing
-end
-
 # internal method that doesn't print a warning if the value is NaN
 _getValue(v::Variable) = v.m.colVal[v.col]
 
@@ -422,7 +414,6 @@ function getValue(arr::Array{Variable})
     end
     ret
 end
-
 
 # Dual value (reduced cost) getter
 function getDual(v::Variable)
@@ -695,9 +686,23 @@ end
 operator_warn(lhs,rhs) = nothing
 
 ##########################################################################
+# Types used in the nonlinear code
+immutable NonlinearExpression
+    m::Model
+    index::Int
+end
+
+immutable NonlinearParameter <: AbstractJuMPScalar
+    m::Model
+    index::Int
+end
+
+
+##########################################################################
 # Behavior that's uniform across all JuMP "scalar" objects
 
 typealias JuMPTypes Union{AbstractJuMPScalar,
+                          NonlinearExpression,
                           Norm,
                           GenericAffExpr,
                           QuadExpr,
@@ -710,17 +715,6 @@ Base.size(::JuMPTypes) = ()
 Base.size(x::JuMPTypes,d::Int) = 1
 Base.ndims(::JuMPTypes) = 0
 
-##########################################################################
-# Types used in the nonlinear code
-immutable NonlinearExpression
-    m::Model
-    index::Int
-end
-
-immutable NonlinearParameter
-    m::Model
-    index::Int
-end
 
 ##########################################################################
 # Operator overloads
@@ -737,6 +731,17 @@ include("callbacks.jl")
 include("print.jl")
 # Nonlinear-specific code
 include("nlp.jl")
+
+getValue{T<:JuMPTypes}(arr::Array{T}) = map(getValue, arr)
+
+function setValue{T<:AbstractJuMPScalar}(set::Array{T}, val::Array)
+    promote_shape(size(set), size(val)) # Check dimensions match
+    for I in eachindex(set)
+        setValue(set[I], val[I])
+    end
+    nothing
+end
+
 
 ##########################################################################
 end
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -5,6 +5,9 @@
 
 using Base.Meta
 
+issum(s::Symbol) = (s == :sum) || (s == :∑) || (s == :Σ)
+isprod(s::Symbol) = (s == :prod) || (s == :∏)
+
 include("parseExpr_staged.jl")
 
 ###############################################################################
@@ -954,7 +957,6 @@ macro addNLConstraint(m, x, extra...)
         code = quote
             c = NonlinearConstraint(@processNLExpr($(esc(lhs))), $lb, $ub)
             push!($m.nlpdata.nlconstr, c)
-            push!($m.nlpdata.nlconstrlist, c.terms)
             $(refcall) = ConstraintRef{NonlinearConstraint}($m, length($m.nlpdata.nlconstr))
         end
     elseif length(x.args) == 5
@@ -972,7 +974,6 @@ macro addNLConstraint(m, x, extra...)
             end
             c = NonlinearConstraint(@processNLExpr($(esc(x.args[3]))), $(esc(lb)), $(esc(ub)))
             push!($m.nlpdata.nlconstr, c)
-            push!($m.nlpdata.nlconstrlist, c.terms)
             $(refcall) = ConstraintRef{NonlinearConstraint}($m, length($m.nlpdata.nlconstr))
         end
     else
@@ -992,17 +993,59 @@ macro addNLConstraint(m, x, extra...)
     return assert_validmodel(m, code)
 end
 
-macro defNLExpr(x, extra...)
-    # Two formats:
-    # - @defNLExpr(a*x <= 5)
-    # - @defNLExpr(myref[a=1:5], sin(x^a))
-    length(extra) > 1 && error("in @defNLExpr: too many arguments.")
-    # Canonicalize the arguments
-    c = length(extra) == 1 ? x        : nothing
-    x = length(extra) == 1 ? extra[1] : x
+macro defNLExpr(args...)
+    if length(args) <= 2
+        s = IOBuffer()
+        print(s,args[1])
+        if length(args) == 2
+            print(s,",")
+            print(s,args[2])
+        end
+        msg = """
+        in @defNLExpr($(takebuf_string(s))): three arguments are required.
+        Note that the syntax of @defNLExpr has recently changed:
+        The first argument should be the model to which the expression is attached.
+        The second is the name of the expression (or collection of expressions).
+        The third is the expression itself.
+        Example:
+        @defNLExpr(m, my_expr, x^2/y)
+        @defNLExpr(m, my_expr_collection[i=1:2], sin(z[i])^2)
+        Support for the old syntax (with the model omitted) will be removed in an upcoming release.
+        """
+        Base.warn(msg)
+        m = :(__last_model[1])
+        if length(args) == 2
+            c = args[1]
+            x = args[2]
+        else
+            c = nothing
+            x = args[1]
+        end
+    else
+        @assert length(args) == 3
+        m, c, x = args
+        m = esc(m)
+    end
 
-    refcall, idxvars, idxsets, idxpairs = buildrefsets(c)
-    varname = isexpr(refcall,:ref) ? refcall.args[1] : refcall
-    macrocall = Expr(:macrocall, symbol("@parametricExpr"), [esc(v) for v in idxvars]..., esc(x))
-    return :($(varname) = $macrocall)
+    refcall, idxvars, idxsets, idxpairs, condition = buildrefsets(c)
+    code = quote
+        $(refcall) = NonlinearExpression($m, @processNLExpr($(esc(x))))
+    end
+    return assert_validmodel(m, getloopedcode(c, code, condition, idxvars, idxsets, idxpairs, :NonlinearExpression))
+end
+
+# syntax is @defNLParam(m, p[i=1] == 2i)
+macro defNLParam(m, ex)
+    m = esc(m)
+    @assert isexpr(ex, :comparison)
+    @assert length(ex.args) == 3
+    @assert ex.args[2] == :(==)
+    c = ex.args[1]
+    x = ex.args[3]
+
+    refcall, idxvars, idxsets, idxpairs, condition = buildrefsets(c)
+    code = quote
+        $(refcall) = newparameter($m, $(esc(x)))
+    end
+    return assert_validmodel(m, getloopedcode(c, code, condition, idxvars, idxsets, idxpairs, :NonlinearParameter))
 end
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -880,6 +880,7 @@ function _buildInternalModel_nlp(m::Model, traits)
     if m.internalModelLoaded
         @assert isa(nldata.evaluator, JuMPNLPEvaluator)
         d = nldata.evaluator
+        fill!(d.last_x, NaN)
         if length(nldata.nlparamvalues) == 0 && !ENABLE_NLP_RESOLVE[1]
             # no parameters and haven't explicitly allowed resolves
             # error to prevent potentially incorrect answers
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -3,17 +3,44 @@
 #  License, v. 2.0. If a copy of the MPL was not distributed with this
 #  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-typealias NonlinearConstraint GenericRangeConstraint{ReverseDiffSparse.SymbolicOutput}
+include("nlpmacros.jl")
+
+import DualNumbers: Dual, epsilon
+
+type NonlinearExprData
+    nd::Vector{NodeData}
+    const_values::Vector{Float64}
+end
+
+typealias NonlinearConstraint GenericRangeConstraint{NonlinearExprData}
 
 type NLPData
     nlobj
     nlconstr::Vector{NonlinearConstraint}
-    nlconstrlist::ReverseDiffSparse.ExprList
+    nlexpr::Vector{NonlinearExprData}
     nlconstrDuals::Vector{Float64}
+    nlparamvalues::Vector{Float64}
     evaluator
 end
 
-NLPData() = NLPData(nothing, NonlinearConstraint[], ExprList(), Float64[], nothing)
+function NonlinearExpression(m::Model,ex::NonlinearExprData)
+    initNLP(m)
+    nldata::NLPData = m.nlpdata
+    push!(nldata.nlexpr, ex)
+    return NonlinearExpression(m, length(nldata.nlexpr))
+end
+
+function newparameter(m::Model,value::Number)
+    initNLP(m)
+    nldata::NLPData = m.nlpdata
+    push!(nldata.nlparamvalues, value)
+    return NonlinearParameter(m, length(nldata.nlparamvalues))
+end
+
+getValue(p::NonlinearParameter) = p.m.nlpdata.nlparamvalues[p.index]::Float64
+setValue(p::NonlinearParameter,v::Number) = (p.m.nlpdata.nlparamvalues[p.index] = v)
+
+NLPData() = NLPData(nothing, NonlinearConstraint[], NonlinearExprData[], Float64[], Float64[], nothing)
 
 Base.copy(::NLPData) = error("Copying nonlinear problems not yet implemented")
 
@@ -32,20 +59,54 @@ function getDual(c::ConstraintRef{NonlinearConstraint})
     return nldata.nlconstrDuals[c.idx]
 end
 
+type FunctionStorage
+    nd::Vector{NodeData}
+    adj::SparseMatrixCSC{Bool,Int}
+    const_values::Vector{Float64}
+    forward_storage::Vector{Float64}
+    reverse_storage::Vector{Float64}
+    grad_sparsity::Vector{Int}
+    hess_I::Vector{Int} # nonzero pattern of hessian
+    hess_J::Vector{Int}
+    rinfo::Coloring.RecoveryInfo # coloring info for hessians
+    seed_matrix::Matrix{Float64}
+    linearity::Linearity
+    dependent_subexpressions::Vector{Int} # subexpressions which this function depends on, ordered for forward pass
+end
+
+type SubexpressionStorage
+    nd::Vector{NodeData}
+    adj::SparseMatrixCSC{Bool,Int}
+    const_values::Vector{Float64}
+    forward_storage::Vector{Float64}
+    reverse_storage::Vector{Float64}
+    forward_hessian_storage::Vector{Dual{Float64}}
+    reverse_hessian_storage::Vector{Dual{Float64}}
+end
+
 type JuMPNLPEvaluator <: MathProgBase.AbstractNLPEvaluator
     m::Model
     A::SparseMatrixCSC{Float64,Int} # linear constraint matrix
+    parameter_values::Vector{Float64}
     has_nlobj::Bool
     linobj::Vector{Float64}
-    eval_f_nl::Function
-    eval_fgrad_nl::Function
-    eval_h_nl::Function
-    nnz_hess_obj::Int
-    jac_I::Vector{Int}
-    jac_J::Vector{Int}
-    hess_I::Vector{Int}
-    hess_J::Vector{Int}
-    requested_hessian::Bool
+    objective::FunctionStorage
+    constraints::Vector{FunctionStorage}
+    subexpressions::Vector{SubexpressionStorage}
+    subexpression_order::Vector{Int}
+    subexpression_forward_values::Vector{Float64}
+    subexpression_reverse_values::Vector{Float64}
+    subexpressions_as_julia_expressions::Vector{Any}
+    last_x::Vector{Float64}
+    jac_storage::Vector{Float64} # temporary storage for computing jacobians
+    # storage for computing hessians
+    want_hess::Bool
+    forward_storage_hess::Vector{Dual{Float64}} # length is of the longest expression
+    reverse_storage_hess::Vector{Dual{Float64}} # length is of the longest expression
+    forward_input_vector::Vector{Dual{Float64}} # length is number of variables
+    reverse_output_vector::Vector{Dual{Float64}}# length is number of variables
+    subexpression_hessian_forward_values::Vector{Dual{Float64}} # length is number of subexpressions
+    subexpression_hessian_reverse_values::Vector{Dual{Float64}} # length is number of subexpressions
     # timers
     eval_f_timer::Float64
     eval_g_timer::Float64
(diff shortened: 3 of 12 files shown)


mlubin:
CC @rgiordan @davidanthoff as well

src/JuMP.jl
```diff
@@ -296,7 +297,7 @@ setPrintHook(m::Model, f) = (m.printhook = f)
 abstract JuMPConstraint
 # Abstract base type for all scalar types
 # In JuMP, used only for Variable. Useful primarily for extensions
-abstract AbstractJuMPScalar <: ReverseDiffSparse.Placeholder
```
joehuchette: :+1:

src/nlp.jl
```diff
@@ -3,17 +3,25 @@
...
+include("nlpmacros.jl")
+
+import DualNumbers: Dual, epsilon
+
+type NonlinearExprData
+    nd::Vector{NodeData}
+    const_values::Vector{Float64}
+end
+
+typealias NonlinearConstraint GenericRangeConstraint{NonlinearExprData}
```
joehuchette: export?

tkelman:
Will this be more amenable to precompilation? @jgonzal13 was looking at minimizing the load time from julia startup to nonlinear solve.

mlubin:
@tkelman absolutely

mlubin:
We're going to have to break ``@defNLExpr`` by adding the model as the first argument, since subexpressions are now attached to models.

mlubin:
When this is merged, it will close 11% of open issues on JuMP

mlubin:
Subexpressions are up and running

src/nlp.jl
```diff
@@ -32,20 +53,52 @@ function getDual(c::ConstraintRef{NonlinearConstraint})
     return nldata.nlconstrDuals[c.idx]
 end
 
+type FunctionStorage
+    nd::Vector{NodeData}
+    adj::SparseMatrixCSC{Bool,Int}
```
joehuchette: Could this be ``::SparseMatrixCSC{Void,Int}`` instead?

mlubin: I haven't tested if this works with the ``sparse()`` constructor, but the storage for the ``Vector{Bool}`` for ``nzval`` is already 1/8th of the storage for the ``Vector{Int}`` for ``rowval``.

IainNZ:
Re: `@defNLExpr` breakage, are we just going to throw an error explaining the problem if two arguments are given?

mlubin:
I'm hoping we can make it a deprecation warning.

src/nlp.jl
mlubin: done

mlubin:
Not breaking existing code using ``@defNLExpr`` will be hard...
Here's my stab at a useful error message, which includes a backtrace.
```julia
using JuMP
function foo()
    @defNLExpr(a,b+y)
end
foo()
```
yields
```
ERROR: LoadError: in @defNLExpr(a,b + y): three arguments are required.
Note that the syntax of @defNLExpr has recently changed:
The first argument should be the model to which the expression is attached.
The second is the name of the expression (or collection of expressions).
The third is the expression itself.
Example:
@defNLExpr(m, my_expr, x^2/y)
@defNLExpr(m, my_expr_collection[i=1:2],sin(z[i])^2)

 in foo at /home/mlubin/test.jl:3
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:304
 in process_options at ./client.jl:280
 in _start at ./client.jl:378
while loading /home/mlubin/test.jl, in expression starting on line 5
```

@davidanthoff, how disruptive would this change be?

tkelman:
Is it possible to hack a deprecation via something like `get_last_model()` ?

davidanthoff:
I don't think we use ``@defNLExpr`` anywhere... I hope to try this branch sometime next week with the existing code we have.

mlubin:
@tkelman, possibly. Is there a way to print a backtrace after a ``warn``/``warn_once``?

davidanthoff:
I should generally add that I don't mind at all if in a case like this the old API use throws an error. Those kinds of things are really fixed fast, in my experience. I get nervous if the semantics of such a call change, i.e. when the old API call seems to work, but really has a new effect in a new version. But that doesn't seem to be the case here.

mlubin:
@davidanthoff, right, this change is pretty straightforward. The potentially scary change here is with "sneaky reoptimization" (https://github.com/JuliaOpt/JuMP.jl/pull/315) which simply will stop working on this branch. It will be replaced by proper nonlinear parameters as requested in https://github.com/JuliaOpt/JuMP.jl/issues/304, but unfortunately there's no feasible way to give a warning or error message.

mlubin:
Actually wasn't too hard to hack in a deprecation warning for ``@defNLExpr``. It's missing a backtrace though, which makes it potentially less useful than ``error``.

davidanthoff:
Having the backtrace for old ``@defNLExpr`` calls would be really useful, so maybe throwing an error is better.

We rely heavily on the sneaky reoptimization in one project, but solving that via proper parameters would be much preferred on my end. No worries about missing warnings, as long as we know that this is coming we can easily adapt.

tkelman:
There are quite a few users you don't hear much from on github who are also relying heavily on the current vector-parameters for nonlinear data modification. So all data will now get compiled in at the first solve now? That's going to break a lot of code, but not necessarily in immediately visible ways.

mlubin:
@tkelman, yes, the semantics with data will be the same as with linear expressions. Constant values will be captured when expressions are defined. The breakage is unfortunate, but users can always pin to an old release if they need to.

tkelman:
Yes but that doesn't happen automatically. People who aren't paying attention to github or the mailing list will do Pkg.update, then without any warning silently get very different results that aren't doing what they thought they should.

It seems there's now zero point in re-solving a nonlinear model unless the new parameter syntax is used and a value is changed, or maybe a solver option or something. It might make sense to have all nonlinear re-solves warn loudly.

mlubin:
> It seems there's now zero point in re-solving a nonlinear model unless the new parameter syntax is used and a value is changed, or maybe a solver option or something. It might make sense to have all nonlinear re-solves warn loudly.

Yes, this could be a good idea. Will have to test a bit to make sure it's not too annoying. However, a perfectly valid reason to resolve an identical model is to use multiple starting points.

tkelman:
I think you're underestimating how bad this breakage is. I'm concerned about getting a phone call that an autonomous vehicle started violating safety constraints because modified problem data no longer had any effect on path planning after someone ran Pkg.update without doing enough simulations first. I don't think jump is running in closed loop experiments on real roads quite yet, but it will be soon. Not solving the same problem any more, with no immediate way of telling the difference, is a drastic enough thing that I'd consider making this update explicitly opt in, restrict it to julia nightly or something.

mlubin:
What about throwing an error on any nonlinear resolve unless a certain option is passed to the ``Model`` constructor?

mlubin:
(with instructions on how to pin the old version of JuMP to get things working again)

tkelman:
Yes, that would be suitably loud, and easy to circumvent.

davidanthoff:
Throwing an error on resolve (with an option to override) seems a good solution.

mlubin:
I'm also a bit spooked by the idea of JuMP controlling autonomous vehicles on the road...

tkelman:
You'd be more spooked by the existing matlab/simulink s function running on a dspace embedded system with a copy of npsol that someone ran f2c on 10 years ago, and hand coded derivatives or finite differences for everything. Knowing the people on Tesla's autopilot team, what they're using is even less pretty.

mlubin:
I can believe that

mlubin:
The basic infrastructure for parameters is now [in place](https://github.com/JuliaOpt/JuMP.jl/blob/fb9cb71998105ece9fc9cc63fe46f08952388f1c/test/nonlinear.jl#L102). Next I'm going to fight with git to merge ReverseDiffSparse2 into ReverseDiffSparse and tag so that we can get travis passing.

davidanthoff:
Excellent! How will vector valued parameters work?

Have fun with git ;)

mlubin:
@davidanthoff, as you'd expect, ``@defNLParam(m, p[i=1:N,j=S] == coef[i]*X[j])``.

davidanthoff:
Looks great. And then a ``setValue`` call would take an Array, and copy those values into wherever parameter values are stored in a model?

mlubin:
Some of that is handled by JuMPContainer fallbacks already. If you find a missing method we can add it.

mlubin:
And travis is now passing

mlubin:
We now have an [error message](https://github.com/JuliaOpt/JuMP.jl/commit/8139077b80ceb2b4eca1fc362a5328649a60c4cd#diff-19c54d6865b3a7cd7eb343daa4256ea0R887) when attempting to re-solve an NLP. See [this test](https://github.com/JuliaOpt/JuMP.jl/commit/8139077b80ceb2b4eca1fc362a5328649a60c4cd#diff-a72fb81e6e171bf2197a110437d212f0R513) for how it works.

mlubin:
This is an RFC now. Only things really missing are docs and smoothing over the ``@defNLExpr`` error/warning.

IainNZ:
I'll review/play with it tomorrow

anriseth:
This looks great. I'll test the changes on my code and see how things work out

davidanthoff:
I ran this branch with the most involved model we have, both with the current JuMP master and this branch here. I show output for both branches below (I removed the iteration output detail).

From master:
````
This is Ipopt version 3.12.2, running with linear solver ma57.

Number of nonzeros in equality constraint Jacobian...:    34587
Number of nonzeros in inequality constraint Jacobian.:     2124
Number of nonzeros in Lagrangian Hessian.............:    32391

Total number of variables............................:    14942
                     variables with only lower bounds:     4801
                variables with lower and upper bounds:     1140
                     variables with only upper bounds:       60
Total number of equality constraints.................:    12098
Total number of inequality constraints...............:      708
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:      708

Number of Iterations....: 48

                                   (scaled)                 (unscaled)
Objective...............: -2.3086798148916714e+003  -2.3086798148916714e+003
Dual infeasibility......:  1.1368683772161603e-013   1.1368683772161603e-013
Constraint violation....:  3.3196556614711881e-011   3.3196556614711881e-011
Complementarity.........:  2.5059063914226807e-009   2.5059063914226807e-009
Overall NLP error.......:  2.5059063914226807e-009   2.5059063914226807e-009


Number of objective function evaluations             = 53
Number of objective gradient evaluations             = 49
Number of equality constraint evaluations            = 53
Number of inequality constraint evaluations          = 53
Number of equality constraint Jacobian evaluations   = 49
Number of inequality constraint Jacobian evaluations = 49
Number of Lagrangian Hessian evaluations             = 48
Total CPU secs in IPOPT (w/o function evaluations)   =      2.379
Total CPU secs in NLP function evaluations           =      6.112

EXIT: Optimal Solution Found.
 13.518691 seconds (27.13 M allocations: 847.751 MB, 3.71% gc time)
````

And with this branch
````
This is Ipopt version 3.12.2, running with linear solver ma57.

Number of nonzeros in equality constraint Jacobian...:    34587
Number of nonzeros in inequality constraint Jacobian.:     2124
Number of nonzeros in Lagrangian Hessian.............:    19201

Total number of variables............................:    14942
                     variables with only lower bounds:     4801
                variables with lower and upper bounds:     1140
                     variables with only upper bounds:       60
Total number of equality constraints.................:    12098
Total number of inequality constraints...............:      708
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:      708


Number of Iterations....: 46

                                   (scaled)                 (unscaled)
Objective...............: -2.3086798148916710e+003  -2.3086798148916710e+003
Dual infeasibility......:  1.7053025658242404e-013   1.7053025658242404e-013
Constraint violation....:  2.7466739993542433e-010   2.7466739993542433e-010
Complementarity.........:  2.5059157373242014e-009   2.5059157373242014e-009
Overall NLP error.......:  2.5059157373242014e-009   2.5059157373242014e-009


Number of objective function evaluations             = 50
Number of objective gradient evaluations             = 47
Number of equality constraint evaluations            = 50
Number of inequality constraint evaluations          = 50
Number of equality constraint Jacobian evaluations   = 47
Number of inequality constraint Jacobian evaluations = 47
Number of Lagrangian Hessian evaluations             = 46
Total CPU secs in IPOPT (w/o function evaluations)   =      2.218
Total CPU secs in NLP function evaluations           =      1.926

EXIT: Optimal Solution Found.
  4.654671 seconds (4.64 M allocations: 188.112 MB, 5.60% gc time)
````

Overall this new branch seems about a factor 3 faster! Amazing, that of course is fantastic news.

But probably hard to tease out from this test what the real cause is: certainly the sparsity detection for the Lagrangian Hessian seems way better, but I assume that the evaluation code for the various callbacks also looks different, and so one probably can't tell which of these two explains this crazy speedup. But hey, I don't really care ;)

I seem to remember that in the paper that has performance comparisons with AMPL, JuMP was at worst 2.7 times slower than AMPL. Have you rerun those tests with this branch? If the factor 3 speedup I am seeing is seen for other models as well, JuMP might now be faster than AMPL for some problems?

In any case, congrats, this is a fantastic branch.

**Comment by IainNZ** [comment]
Tried it on the rocket problem, the version in `juliaopt-notebooks`

On this branch:
```
Total CPU secs in IPOPT (w/o function evaluations)   =      0.286
Total CPU secs in NLP function evaluations           =      0.720
```

On JuMP master:
```
Total CPU secs in IPOPT (w/o function evaluations)   =      0.297
Total CPU secs in NLP function evaluations           =      0.315
```

**Comment by mlubin** [comment]
@davidanthoff, thanks for testing!
This branch eliminates the difference between expertly tuned models (e.g., the models which only the JuMP developers know how to write) and more typical user models (e.g., https://github.com/JuliaOpt/JuMP.jl/issues/614), so there will be a huge decrease in the performance variability, and I'd say an overall improvement for most models. Unfortunately the expertly tuned models have gotten a bit slower, so it's not fair to say that we're faster than AMPL yet. I'm planning on doing some low-hanging performance work after the branch is merged.

**Comment by mlubin** [comment]
On top of the derivative time, there should be a dramatic decrease in the time between ``solve()`` and when Ipopt starts solving.

File: src/macros.jl [hunk 4]
```diff
@@ -5,6 +5,9 @@
 
 using Base.Meta
 
+issum(s::Symbol) = (s == :sum) || (s == :∑) || (s == :Σ)
```
philipithomas: I was going to suggest casting the variable to lowercase so that you didn't have to handle comparing to both uppercase and lowercase, but I was disappointed to discover that Julia does not handle case transformations for Greek. 

```julia
lowercase("∑") == "∑" # False
lowercase("E") == "e" # True
```

src/macros.jl
```diff
@@ -992,17 +993,52 @@ macro addNLConstraint(m, x, extra...)
...
-macro defNLExpr(x, extra...)
-    # Two formats:
-    # - @defNLExpr(a*x <= 5)
-    # - @defNLExpr(myref[a=1:5], sin(x^a))
-    length(extra) > 1 && error("in @defNLExpr: too many arguments.")
-    # Canonicalize the arguments
-    c = length(extra) == 1 ? x        : nothing
-    x = length(extra) == 1 ? extra[1] : x
+macro defNLExpr(args...)
+    if length(args) == 2
```
philipithomas: Why is this not `<=`?

src/macros.jl
```diff
@@ -992,17 +993,52 @@ macro addNLConstraint(m, x, extra...)
...
+        print(s,args[2])
+        msg = """
+        in @defNLExpr($(takebuf_string(s))): three arguments are required.
+        Note that the syntax of @defNLExpr has recently changed:
+        The first argument should be the model to which the expression is attached.
+        The second is the name of the expression (or collection of expressions).
+        The third is the expression itself.
+        Example:
+        @defNLExpr(m, my_expr, x^2/y)
+        @defNLExpr(m, my_expr_collection[i=1:2],sin(z[i])^2)
```
philipithomas: Pedantic, but a space after the final comma would improve legibility so that it is clear that there are 3 args

src/nlp.jl
```diff
@@ -32,20 +59,54 @@ function getDual(c::ConstraintRef{NonlinearConstraint})
...
+    objective::FunctionStorage
+    constraints::Vector{FunctionStorage}
+    subexpressions::Vector{SubexpressionStorage}
+    subexpression_order::Vector{Int}
+    subexpression_forward_values::Vector{Float64}
+    subexpression_reverse_values::Vector{Float64}
+    subexpressions_as_julia_expressions::Vector{Any}
+    last_x::Vector{Float64}
+    jac_storage::Vector{Float64} # temporary storage for computing jacobians
+    # storage for computing hessians
```
philipithomas: More of a style critique, but all other comments are inline with the variable.

src/nlp.jl
```diff
@@ -83,147 +201,87 @@ function MathProgBase.initialize(d::JuMPNLPEvaluator, requested_features::Vector
...
-            hJ = []
-            d.eval_h_nl = (x,y,z) -> nothing
-            nnz_hess += length(d.m.obj.qvars1)
-        end
-        d.nnz_hess_obj = length(hI)
 
-        hess_I = zeros(Int,nnz_hess)
-        hess_J = zeros(Int,nnz_hess)
+    if d.has_nlobj
+        @assert length(d.m.obj.qvars1) == 0 && length(d.m.obj.aff.vars) == 0
```
philipithomas: Just for maintainability, it might be good to break this into two lines (separate assertions) so that it is clear which length is incorrect should either trigger an error.

src/nlp.jl
```diff
@@ -274,10 +366,13 @@ end
...
 function MathProgBase.eval_g(d::JuMPNLPEvaluator, g, x)
     tic()
+    if d.last_x != x
+        forward_eval_all(d,x)
+    end
     A = d.A
     for i in 1:size(A,1); g[i] = 0.0; end
-    #fill!(subarr(g,1:size(A,1)), 0.0)
-    A_mul_B!(subarr(g,1:size(A,1)),A,x)
+    #fill!(sub(g,1:size(A,1)), 0.0)
```
philipithomas: Is this intentionally commented?

src/nlp.jl
```diff
@@ -328,7 +429,25 @@ function MathProgBase.eval_jac_g(d::JuMPNLPEvaluator, J, x)
...
+        end
+
+        for k in 1:length(nzidx)
+            J[idx+k-1] = grad_storage[nzidx[k]]
+        end
+        idx += length(nzidx)
+    end
 
     d.eval_jac_g_timer += toq()
     #print("x = ");show(x);println()
```
philipithomas: Same - is this intentionally left commented?

Also raises the idea of adding in a proper logging library so that you could make something like this a debug statement

src/macros.jl
tkelman: one of these is greek capital letter sigma, the other is n-ary summation

src/nlp.jl
```diff
@@ -463,9 +771,64 @@ function quadToExpr(q::QuadExpr,constant::Bool)
...
+        return ex
+    elseif nod.nodetype == LOGIC
+        op = nod.index
+        opsymbol = logic_operators[op]
+        children_idx = nzrange(adj,k)
+        lhs = tapeToExpr(children_arr[first(children_idx)], nd, adj, const_values, parameter_values, subexpressions)
+        rhs = tapeToExpr(children_arr[last(children_idx)], nd, adj, const_values, parameter_values, subexpressions)
+        return Expr(opsymbol, lhs, rhs)
+    else
+        error()
```
philipithomas: Is this now an empty "else" statement?

src/nlp.jl
IainNZ: This is a comment about the following 7 or so lines, not the `want_hess`, if thats what you mean

src/macros.jl
philipithomas: My greek sucks. Please disregard. (Though `lowercase("∑") == "∑"` is still frustratingly evaluating to `true`)

src/nlp.jl
philipithomas: Ah thanks, nevermind.

src/nlp.jl
philipithomas: In general, rather than this:

```julia
else
    error()
end
```

it's preferred to do this
```julia
# skip else
end
error()
```

This is because it still triggers an error if one of the preceding statements forgot to include a `return`.

davidanthoff:
@mlubin The last line of of each output log that I posted shows a ``@time`` output around the ``solve`` call. I'm not sure whether it makes sense to take that time, and then subtract the Ipopt reported times spent in Ipopt and Ipopt reported times spent in function evaluation, but if you do that, there is an improvement of the residual time of almost a factor 10! I guess that is the dramatic improvement in the time until ``solve`` actually starts calling Ipopt?

mlubin:
@davidanthoff, yes, that's a reasonable way to measure the JuMP overhead in processing the model.

anriseth:
_The following breaks on nlprewrite, but works on v0.11.1._

I get an error in ReverseDiffSparse when running the model below.
It works with AmplNLWriter.

```julia
using JuMP, AmplNLWriter, Ipopt
#m = Model(solver=IpoptNLSolver()) # Works
m = Model(solver=IpoptSolver()) # Fails
@defVar(m, x[i=1:5])
@defNLExpr(m, f, 3x[1]+2x[2]-x[3]/3+0.01*(x[4]-x[5])^3)
@addNLConstraint(m, x[1]+2x[2]-x[3]-0.5x[4]+x[5]==2)
@addNLConstraint(m, 4x[1]-2x[2]+0.8x[3]+0.6x[4]+0.5x[5]^2 == 0)
@addNLConstraint(m, sum{x[i]^2, i=1:5} <= 10)
@setNLObjective(m, :Min, f)
solve(m)
```

Gives the error

```julia
This is Ipopt version 3.11.9, running with linear solver mumps.
NOTE: Other linear solvers might be more efficient (see Ipopt documentation).

Number of nonzeros in equality constraint Jacobian...:       10
Number of nonzeros in inequality constraint Jacobian.:        5
Number of nonzeros in Lagrangian Hessian.............:        9

Total number of variables............................:        5
                     variables with only lower bounds:        0
                variables with lower and upper bounds:        0
                     variables with only upper bounds:        0
Total number of equality constraints.................:        2
Total number of inequality constraints...............:        1
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        1

iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  0.0000000e+00 2.00e+00 1.08e+00  -1.0 0.00e+00    -  0.00e+00 0.00e+00   0
WARNING: Problem in step computation; switching to emergency mode.
   1r 0.0000000e+00 2.00e+00 9.99e+02   0.3 0.00e+00  20.0 0.00e+00 0.00e+00R  1
ERROR: LoadError: DomainError:
 in pow at /home/asbjorn/.julia/v0.4/DualNumbers/src/dual.jl:231
 in forward_eval at /home/asbjorn/.julia/v0.4/ReverseDiffSparse/src/forward.jl:59
 in hessian_slice at /home/asbjorn/.julia/v0.4/JuMP/src/nlp.jl:640
 in eval_hesslag at /home/asbjorn/.julia/v0.4/JuMP/src/nlp.jl:593
 in eval_h_cb at /home/asbjorn/.julia/v0.4/Ipopt/src/IpoptSolverInterface.jl:90
 in eval_h_wrapper at /home/asbjorn/.julia/v0.4/Ipopt/src/Ipopt.jl:147
while loading /home/asbjorn/test.jl, in expression starting on line 11
```

IainNZ:
Can you provide a starting value? Might be that going via NL is starting from a different point

anriseth:
Setting a startvalue doesn't seem to help. 
I've tried starting at zeros(5), -ones(5) and at the optimal value given from NLWriter.

mlubin:
@anriseth, thanks for finding that. Not sure why the behavior changed with the rewrite, but will be fixed with https://github.com/JuliaDiff/DualNumbers.jl/pull/34

src/nlp.jl
mlubin: Good point in general, but either assertion failing here would lead to the same debugging anyway

src/nlp.jl
mlubin: Yep

src/nlp.jl
mlubin: This is old debugging code. Probably safe to remove

src/nlp.jl
mlubin: Good pattern, I like it

mlubin:
All known issues are fixed and I've updated the docs. Unless there are any objections, seems ready to merge. Afterwards I'll put out a call for testers on julia-opt (and a heads-up for the breaking changes).

philipithomas:
:+1: Looks good to merge

joehuchette:
gogogogogogo

davidanthoff:
Yay!!!

joehuchette:
file jan 11 5 45 58 pm

PR #641, opened 2016-01-04
getValue on nonlinear arrays

anriseth:
This enables getValue to be called on nonlinear vector-expressions, the same way it's done for AffExpr.

(I hope opening a PR to branch nlprewrite works together with #638)

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -50,7 +50,7 @@ export
     @SOCConstraint, @SOCConstraints,
     @defVar, @defConstrRef, @setObjective, addToExpression, @defExpr,
     @setNLObjective, @addNLConstraint, @addNLConstraints,
-    @defNLExpr
+    @defNLExpr, @defNLParam
 
 include("JuMPContainer.jl")
 include("utils.jl")
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -17,7 +17,7 @@ importall Base.Operators
 import MathProgBase
 
 using Calculus
-using ReverseDiffSparse2
+using ReverseDiffSparse
 import ArrayViews
 const subarr = ArrayViews.view
 
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -19,6 +19,10 @@ import MathProgBase
 using Calculus
 using ReverseDiffSparse
 
+function __init__()
+    ENABLE_NLP_RESOLVE[1] = false
+end
+
 export
 # Objects
     Model, Variable, Norm, AffExpr, QuadExpr, SOCExpr, AbstractJuMPScalar,
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -373,14 +373,6 @@ function setValue(v::Variable, val::Number)
     end
 end
 
-function setValue(set::Array{Variable}, val::Array)
-    promote_shape(size(set), size(val)) # Check dimensions match
-    for I in eachindex(set)
-        setValue(set[I], val[I])
-    end
-    nothing
-end
-
 # internal method that doesn't print a warning if the value is NaN
 _getValue(v::Variable) = v.m.colVal[v.col]
 
@@ -422,7 +414,6 @@ function getValue(arr::Array{Variable})
     end
     ret
 end
-
 
 # Dual value (reduced cost) getter
 function getDual(v::Variable)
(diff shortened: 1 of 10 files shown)


src/JuMP.jl
```diff
@@ -738,5 +729,26 @@ include("print.jl")
 # Nonlinear-specific code
 include("nlp.jl")
 
+
+typealias _getTypes Union{NonlinearExpression,NonlinearParameter, AffExpr}
```
mlubin: Why isn't ``Variable`` listed here?

src/JuMP.jl
```diff
@@ -738,5 +729,26 @@ include("print.jl")
...
+function getValue{T<:_getTypes}(arr::Array{T})
```
mlubin: Couldn't we just define this as ``map(getValue,arr)``? I don't think this is performance critical code and the performance difference from using map will be disappearing soon.

mlubin:
Looks good, just needs some tidying. CC @joehuchette 
I'd also like to squash some commits in #638 before we merge this.

src/JuMP.jl
anriseth: Just forgot it, thanks

mlubin: Check that it's not already defined somewhere else

anriseth: Actually, it was because the `getValue(Array{Variable})` function seems to be doing a lot more checking and Warning printing than was done on the original `getValue(Array{AffExpr})`. 

I didn't want to remove all that from `getValue(Array{Variable})`

mlubin: Ah, good point

src/JuMP.jl
anriseth: Sure, I'll do that. 

Out of interest, what are the changes in `map` that are coming up (reference)?

mlubin: https://github.com/JuliaLang/julia/pull/13412

mlubin:
I wonder why AppVeyor didn't run on this PR.

src/JuMP.jl
```diff
@@ -738,5 +729,19 @@ include("print.jl")
 # Nonlinear-specific code
 include("nlp.jl")
 
+
+typealias _getTypes Union{NonlinearExpression,NonlinearParameter, AffExpr}
+getValue{T<:_getTypes}(arr::Array{T}) = map(getValue, arr)
```
joehuchette: I think I would prefer it if this was ``T <: JuMPTypes`` (with ``NonlinearExpression <: JuMPTypes`` etc. elsewhere). I think that would also help with the problem with more specific behavior for ``Array{Variable}`` mentioned above (since this would be a stricter type signature).

src/JuMP.jl
```diff
@@ -738,5 +729,19 @@ include("print.jl")
...
+
+typealias _setTypes Union{NonlinearParameter, Variable}
+function setValue{T<:_setTypes}(set::Array{T}, val::Array)
```
joehuchette: Similar to above, I think I would prefer using ``T<:AbstractJuMPScalar`` (with ``NonlinearParameter <: AbstractJuMPScalar`` elsewhere), rather than this type union.

src/JuMP.jl
mlubin: Having ``NonlinearExpression <: JuMPTypes`` would define a lot of operator overloads which don't necessarily make sense. We could do ``Union{JuMPTypes, NonlinearExpression}`` though.

joehuchette: Are you sure? ``JuMPTypes`` appears a lot in the operator overload code, but for scalar ops it looks like the signatures are of the form ``{C,V<:JuMPTypes}(lhs::V, rhs::GenericAffExpr{C,V})``, which is already nonsense if ``V==NonlinearExpression``. Then the vectorized ops all fall back on these scalar ops.

mlubin: The definitions like ``function $op{T<:JuMPTypes}(lhs::Number,rhs::Array{T})`` could easily be triggered by accident and lead to confusion, I think.

joehuchette: But that just falls back on ``$op(las, rhs[I])``, which will give a no method error.

joehuchette:
```jl
m = Model()
@defNLParam(m, x == 1)
2 + [x,x] # => ERROR: MethodError: `.+` has no method matching .+(::Int64, ::JuMP.NonlinearParameter)
```

mlubin: Fair enough

anriseth: I've made the changes. Haven't dealt much with (sub)types in Julia before, so let me know if I'm doing it wrong.

joehuchette:
Looks great to me, thanks for doing this!

tkelman:
> I wonder why AppVeyor didn't run on this PR.

Target branch isn't in the whitelist.
https://github.com/JuliaOpt/JuMP.jl/blob/e489af906e5e95969ff6c22afaacf081f06a4767/appveyor.yml#L8-L11

joehuchette:
@mlubin are you alright merging this into your branch?

mlubin:
@joehuchette, let me squash my branch first

mlubin:
Cherry-picked into https://github.com/JuliaOpt/JuMP.jl/pull/638, so this is resolved

PR #651, opened 2016-01-21
support for simplifying nonlinear expressions with fixed variables

mlubin:
This gives a 3x speedup on the acpower benchmark, which basically closes the performance regression from the NLP rewrite on this model. If we integrate ForwardDiff (CC @jrevels), that should finally give us a speedup over the old code.

The simplification is a user option via ``Model(simplify_nonlinear_expressions=true)`` since there's a trade-off... if we simplify the nonlinear expression and eliminate fixed variables, then we have to prepare the AD from scratch on the next solve. I don't have a good sense for a reasonable default. Eventually we could make the code smarter to detect which kinds of changes occur in the model and avoid redoing the AD in certain cases. For the moment we can recommend users to try this option to speed up derivative evaluations.

CC @davidanthoff @tkelman

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -114,6 +114,7 @@ type Model
     indexedVector::IndexedVector{Float64}
 
     nlpdata#::NLPData
+    simplify_nonlinear_expressions::Bool
 
     varDict::Dict{Symbol,Any} # dictionary from variable names to variable objects
     varData::ObjectIdDict
@@ -133,7 +134,7 @@ type UnsetSolver <: MathProgBase.AbstractMathProgSolver
 end
 
 # Default constructor
-function Model(;solver=UnsetSolver())
+function Model(;solver=UnsetSolver(), simplify_nonlinear_expressions::Bool=false)
     if !isa(solver,MathProgBase.AbstractMathProgSolver)
         error("solver argument ($solver) must be an AbstractMathProgSolver")
     end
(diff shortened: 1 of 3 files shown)


mlubin:
Also, if fixed variables are eliminated then we provide their derivatives as zero. That doesn't seem to bother Ipopt or KNITRO but it's technically incorrect. The more correct approach would be to renumber the variables so that the solvers aren't aware of the fixed variables.

tkelman:
Ipopt is smart enough to remove fixed variables, knitro probably is too. There's a fixed_variable_treatment (or similar) option that can be changed around.

mlubin:
@tkelman, that's what I thought, but I wouldn't want to depend on the solver being that smart if we make this the default.

davidanthoff:
Nice! One question: from a user's point of view, what is the difference between a parameter and a fixed variable? Not entirely clear to me right now...

mlubin:
Philosophically, I don't like the idea that fixing a variable can change
the class of the problem, e.g. nonlinear to linear, nonconvex to convex.
Fixed variables are just normal variables with the lower bound equal to the
upper bound, so it's well-posed to ask for the dual of a fixed variable for
sensitivity analysis. (Right now that isn't implemented when you eliminate
them using this new option.)

Parameters are a convenience that let you avoid regenerating a model from
scratch when the data change. You can't ask for the dual of a parameter.

davidanthoff:
So, essentially, parameters could also have been implemented by adding them as vars, and then fixing them... Thanks for clarifying, this all seems good.

mlubin:
> So, essentially, parameters could also have been implemented by adding them as vars, and then fixing them...

If you're assuming that JuMP will always eliminate them, yes. If they're passed to the solver as fixed variables, it's possible that they could affect convergence if the solver isn't too smart.

PR #652, opened 2016-01-22
add inline documentation

abhijithch:
Have added inline documentation for most of the functions and types exported by JuMP module.


IainNZ:
Hi @abhijithch. I see that this PR just copies the manual into the source, and indeed shows that even the manual is out-of-date as some of those docstrings are now incorrect. At this time we are not adding inline docstrings until the issue of how to keep the manual and the inline docstrings synchronized.

PR #668, opened 2016-02-10
RFC: support for user-defined nonlinear functions

mlubin:
``autodiff=true`` makes a beautiful comeback: 

```julia
mysquare(x) = x^2
myf(x,y) = (x-1)^2+(y-2)^2

registerNLFunction(:myf, 2, myf, autodiff=true)
registerNLFunction(:mysquare, 1, mysquare, autodiff=true)

m = Model()

@defVar(m, x[1:2] >= 0.5)
@setNLObjective(m, Min, myf(x[1],mysquare(x[2])))

status = solve(m)
@show getValue(x)
```

Only major caveat at the moment is that we don't yet support hessians for multivariate functions. Mostly looking to get feedback on the syntax. Interesting usage examples would also be appreciated for the documentation.

CC @jrevels @davidanthoff @anriseth @tkelman @dpo

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -39,6 +39,7 @@ export
     addConstraint, addSOS1, addSOS2, solve,
     getInternalModel, buildInternalModel, setSolveHook, setPrintHook,
     getConstraintBounds,
+    registerNLFunction,
     # Variable
     setName, getName, setLower, setUpper, getLower, getUpper,
     getValue, setValue, getDual, setCategory, getCategory,
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -5,8 +5,6 @@
 
 include("nlpmacros.jl")
 
-import DualNumbers: Dual, epsilon
-
 type NonlinearExprData
     nd::Vector{NodeData}
     const_values::Vector{Float64}
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -104,9 +104,11 @@ type JuMPNLPEvaluator <: MathProgBase.AbstractNLPEvaluator
     subexpressions_as_julia_expressions::Vector{Any}
     last_x::Vector{Float64}
     jac_storage::Vector{Float64} # temporary storage for computing jacobians
+    user_output_buffer::Vector{Float64} # temporary storage for user-defined functions
     # storage for computing hessians
     # these Float64 vectors are reinterpreted to hold multiple epsilon components
     # so the length should be multiplied by the maximum number of epsilon components
+    disable_2ndorder::Bool # don't offer Hess or HessVec
     want_hess::Bool
     forward_storage_ϵ::Vector{Float64} # (longest expression)
     partials_storage_ϵ::Vector{Float64} # (longest expression)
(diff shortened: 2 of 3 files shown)


mlubin:
CC @fqiang @cnpetra

davidanthoff:
Cool! Will there also be a way to provide my own derivative implementations? Say something like this

````julia
f(x)=x[1]^2
function df(x, grad)
    grad[1] = 2x[1]
end

registerNLFunction(:f, 1, function=f, gradient=df)
````

Or something like that? The application I have in mind is a case where (I think) I might have a custom implementation for the gradient that could be very efficient, so I might not even want to use any autodiff for my specific function...

I should say that the implementation in the PR here will work great for what I have in mind right now, the suggestion in this comment is just something I have been thinking about, might also be a dead end.

**Comment by mlubin** [comment]
@davidanthoff,

Yes, it's already there. The syntax for univariate and multivariate functions is a bit different. If the dimension is 2 or greater, the gradient vector will come in as the first argument. The input values never come in as a vector, they're always splatted. We may eventually provide a vector-input version but I think the current version makes more sense for beginning users.
```
f(x)=x^2
df(x) = 2x

registerNLFunction(:f, 1, f, df)

myf(x,y) = (x-1)^2+(y-2)^2
function myf_g(grad,x,y)
    grad[1] = 2(x-1)
    grad[2] = 2(y-2)
end
registerNLFunction(:myf_2, 2, myf, (g,x,y) -> (g[1] = 2(x-1); g[2] = 2(y-2)))
```

**Comment by mlubin** [comment]
Another option is to assume vector-form input if a gradient function is provided. I'm open to suggestions here.

**Comment by IainNZ** [comment]
Exciting day

**Comment by cnpetra** [comment]
Vector input would be very useful when using simulation-based evaluations of function/constraint and its gradient.

Regarding user-defined Hessians,  it would be definitely helpful to support user-defined Hessian applies (Hessian-vector products) functions (again for simulation-based setups).

**Comment by mlubin** [comment]
The vector input mode is mostly a syntax issue (unless the dimension gets very large), because you can always define your function as:
```julia
function foo(x...)
   ...
end
```

tkelman:
Isn't splatting still kind of slow for large vectors? I'd want to express functions of separate vectors, say state, control, parameters etc so something more general than a single vector input and less verbose than listing every scalar input would be ideal.

mlubin:
The feature requests start piling up :)

> Isn't splatting still kind of slow for large vectors?

Yes, in the first pass I'm aiming to support the low-dimensional use cases.

> I'd want to express functions of separate vectors, say state, control, parameters etc so something more general than a single vector input and less verbose than listing every scalar input would be ideal.

I could imagine something like:
```
function foo(x,y)
 ...
end
registerNLFunction(:foo, (2,2), foo, autodiff=true)
```

which will pass two vectors of length 2. The trouble is that JuMP's syntax doesn't allow vector-valued nodes in the nonlinear expression graph, so you wouldn't be able to call it like ``foo(x,y)`` in the JuMP nonlinear expression where x and y are vectors of variables. It's worth thinking about but there's no way I can spend time on making this work anytime soon.

mlubin:
Docs will be forthcoming

PR #677, opened 2016-02-22
take in model object as first argument to @defExpr

mlubin:
All calls to ``@defExpr`` will now take three arguments. I added a version of ``@JuMP.Expression`` to replace the one-argument ``@defExpr``. Not sure I want to export this at this point.


mlubin:
Any objections?

joehuchette:
:

PR #678, opened 2016-02-22
debugging work for travis failure on master

mlubin:
Travis has been failing on Julia nightly for a while, would be nice if it didn't.

Current failure mode is:
```
[callback] Test user cuts
  > With solver GLPKMathProgInterface.GLPKInterfaceMIP.GLPKSolverMIP
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

CC @tkelman

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ julia:
 notifications:
     email: false
 sudo: false
+dist: trusty
 addons:
     apt_packages:
         - gfortran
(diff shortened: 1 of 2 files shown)


mlubin:
So disabling GLPK seems to fix the issue on travis, but I can't reproduce the hanging locally using the same julia binary as travis. Also weird that travis has been failing in the same way on both OS X and Linux...

PR #679, opened 2016-02-23
pass model through nonlinear expression parsing

mlubin:
@joehuchette, this should be a good start

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -912,7 +912,7 @@ macro setNLObjective(m, sense, x)
     code = quote
         initNLP($m)
         setObjectiveSense($m, $(esc(sense)))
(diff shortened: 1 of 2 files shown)


IainNZ:
?

mlubin:
@IainNZ,
```
julia> m = Model(); m2 = Model();

julia> @defVar(m, x)

julia> @addNLConstraint(m2, x <= 1)
ERROR: Variable in nonlinear expression does not belong to corresponding model
 in parseNLExpr_runtime at /home/mlubin/.julia/v0.4/JuMP/src/nlpmacros.jl:151
```

before this would silently give the wrong answer. Also we need this in place for the nonlinear version of StochJuMP.

IainNZ:
Oh wow!

PR #688, opened 2016-03-03
implement defVars macro

mlubin:
Now with parentheses syntax, see the tests. The parsing *should* work fine for valid expressions, but we will probably need to add a bit more to catch syntax errors.

Closes https://github.com/JuliaOpt/JuMP.jl/pull/445
Closes https://github.com/JuliaOpt/JuMP.jl/issues/440

### DIFF
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -456,20 +456,20 @@ end
 facts("[macros] @defVars and @addConstraints") do
     m = Model()
     @defVars m begin
-        0 ≤ x ≤ 1
+        0 ≤ x[i=1:2] ≤ i
         y ≥ 2, Int, (start = 0.7)
         z ≤ 3, (start=10)
         q, (Bin, start=0.5)
     end
     @addConstraints m begin
-        0 ≤ x + y ≤ 1
-        x + z ≤ 2
+        0 ≤ x[1] + y ≤ 1
+        x[1] + z ≤ 2
         y + z ≥ 3
         y*z ≤ 1
     end
-    @fact m.colUpper --> [1.0, Inf,  3.0, 1.0]
-    @fact m.colLower --> [0.0, 2.0, -Inf, 0.0]
-    @fact m.colCat --> [:Cont, :Int, :Cont, :Bin]
+    @fact m.colUpper --> [1.0, 2.0, Inf,  3.0, 1.0]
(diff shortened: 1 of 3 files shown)


test/macros.jl
```diff
@@ -452,3 +452,31 @@ facts("[macros] Issue #621") do
...
     @fact conToStr(con) --> "3 x² + x + 1 $leq 0"
 end
+
+facts("[macros] @defVars and @addConstraints") do
+    m = Model()
+    @defVars m begin
+        0 ≤ x ≤ 1
+        y ≥ 2, Int
+        z ≤ 3, (start=10)
+        q, (Bin, start=0.5)
```
joehuchette: Is
```jl
q, Bin, (start=0.5)
```
valid as well here?

mlubin: Yes

mlubin:
@IainNZ, any objections?

test/macros.jl
```diff
@@ -452,3 +452,32 @@ facts("[macros] Issue #621") do
     con = @QuadConstraint(a+q*3 <= 0)
     @fact conToStr(con) --> "3 x² + x + 1 $leq 0"
 end
+
+facts("[macros] @defVars and @addConstraints") do
+    m = Model()
+    @defVars m begin
+        0 ≤ x ≤ 1
```
IainNZ: Maybe have a variable with indicies and nontrivial bounds, just to be safe with parsing?

IainNZ:
No objections on technical merit, bit nervous about having yet another syntax

mlubin:
This is the future™

test/macros.jl
mlubin: Updated

IainNZ:
Hah ok. Just had a longish conversation with a couple of the first years who were scared that JuMP was doing weird scoping things, not sure this eases those anxieties. But YOLO.

joehuchette:
What were the scoping concerns?

IainNZ:
Just a lack of confidence that a `@defVar`'d variable was a just a plain-old Julia variable, and the notion of ownership between a variable and a model. Also a desire for `getVar`, because they have 20 or so different sets of variables in their model and were struggling to break down their model construction in a way they felt happy with.

mlubin:
I've been toying with an 
```
x = ....
```
syntax within macros but haven't been able to come up with something reasonable.

PR #689, opened 2016-03-03
Add getValue for norms

IainNZ:
Tackling the big problems

### DIFF
--- a/src/norms.jl
+++ b/src/norms.jl
@@ -71,6 +71,7 @@ _build_norm(Lp, terms::Vector{GenericAffExpr}) = _build_norm(Lp, [terms...])
 # Alias for AffExprs. Short-hand used in operator overloads, etc.
 typealias Norm{P} GenericNorm{P,Float64,Variable}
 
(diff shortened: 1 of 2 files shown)


src/norms.jl
```diff
@@ -71,6 +71,23 @@ _build_norm(Lp, terms::Vector{GenericAffExpr}) = _build_norm(Lp, [terms...])
 # Alias for AffExprs. Short-hand used in operator overloads, etc.
 typealias Norm{P} GenericNorm{P,Float64,Variable}
 
+function getValue{P,C,V}(n::GenericNorm{P,C,V})
+    term_vals = getValue(n.terms)
+    if P == 2
+        return sqrt(sumabs2(term_vals))
+    elseif P == 1
```
mlubin: Is there a reason to carry around all of this untested code below for ``P != 2``?

IainNZ: I was thinking of JuMPeR, but then I just realized you can't take `getValue` of `Uncertain` anyway. I'll remove it.

IainNZ: Oh I guess I can just do a oneliner version anyway

PR #701, opened 2016-03-12
Add support for setting base name in defVar

joehuchette:
Could be useful for extensions:
```jl
julia> i = 3;

julia> @defVar(m, x[1:3], basename="foo-$i")
3-element Array{JuMP.Variable,1}:
 foo-3[1]
 foo-3[2]
 foo-3[3]

julia> x[2]
foo-3[2]
```
cc @juan-pablo-vielma

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -162,7 +162,8 @@ end
 
 getname(c::Symbol) = c
 getname(c::Void) = ()
-getname(c::Expr) = c.args[1]
+getname(c::AbstractString) = c
+getname(c::Expr) = (c.head == :string ? c : c.args[1])
 
 validmodel(m::Model, name) = nothing
 validmodel(m::MathProgBase.MathProgCallbackData, name) = error("Expected $name to be a JuMP model, but it is a callback object. Use of this macro is not supported within callbacks.")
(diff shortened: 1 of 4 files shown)


mlubin:
LGTM. Docs and tests?

IainNZ:
Thats pretty neat, I could use this.
Is there a logic behind calling it `basename`?

joehuchette:
Not really, I just thought "name" wasn't quite right. Open to suggestions.

src/macros.jl
```diff
@@ -872,36 +877,34 @@ macro defVar(args...)
...
         code = quote
             $(esc(idxsets[1].args[1].args[2])) == $(esc(idxsets[2].args[1].args[2])) || error("Cannot construct symmetric variables with nonsquare dimensions")
             (issym($lb) && issym($ub)) || error("Bounds on symmetric variables must be symmetric")
             $looped
-            push!($(m).dictList, $varname)
+            push!($(m).dictList, $escvarname)
         end
         if sdp
-            code = :($code; push!($(m).varCones, (:SDP, first($varname).col : last($varname).col)))
+            code = :($code; push!($(m).varCones, (:SDP, first($quotvarname).col : last($quotvarname).col)))
```
joehuchette: should be ``escvarname``

joehuchette:
CIs broken, so....

PR #706, opened 2016-03-17
switch to run-time lookups for user-defined functions

### LINKED ISSUE #705: Scope issue with registerNLFunction?
The following code does not work with JuMP 0.12:

```
using JuMP

function mysquare(x)
	return x^2
end

function generateModel!(m)

	registerNLFunction(:square, 1, mysquare, autodiff=true)

	@defVar(m, 0 <= x <= 5)
	@addNLConstraint(m, square(x) >= 1)
	@setObjective(m, Min, x)

	return nothing

end

m = Model()

generateModel!(m)
```

`ERROR: LoadError: Unrecognized function square used in nonlinear expression.`

Interestingly, this works if the model generation is NOT placed in a function. Bug or feature?

mlubin:
CC @adowling2, could you confirm this fixes #705?

### DIFF
--- a/src/nlpmacros.jl
+++ b/src/nlpmacros.jl
@@ -16,9 +16,18 @@ function parseNLExpr(m, x, tapevar, parent, values)
             code = :(let; end)
             block = code.args[1]
             @assert isexpr(block, :block)
-            haskey(univariate_operator_to_id,x.args[1]) || error("Unrecognized function $(x.args[1]) used in nonlinear expression.")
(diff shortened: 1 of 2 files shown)


adowling2:
Sure. What command do I need to run in Julia to grab/apply this patch from GitHub?

mlubin:
``git fetch; git checkout ml/runtime`` inside of the JuMP directory should do the trick

adowling2:
I can confirm this change fixed the error. I tested it both on a small example (#705) and a larger problem.

PR #711, opened 2016-03-23
fix for parsing changes on julia 0.5

mlubin:
This gets tests passing on 0.5 modulo https://github.com/JuliaOpt/JuMP.jl/issues/700. Note that we're dropping support for old builds of 0.5.


mlubin:
Ref https://github.com/JuliaLang/julia/pull/15524

PR #723, opened 2016-04-06
support for _locally_ valid lazy constraints and user cuts

madanim:
First, thanks a lot for all these wonderful and truely useful tools!

Having re-coded some algorithm using _local_ lazy constraints* with Julia/JuMP, here is a suggestion of very small additions handling them in JuMP/MathProgBase/CPLEX (joint additions, cf the forks at https://github.com/madanim). As far as I know, this feature is only avalaible in CPLEX and XPress MP.

It simply consists in duplicating the current mechanics for lazy constraints and user cuts, and using the CPLEX naming convention, appending 'local' or 'Local' where applicable: e.g. '@addLazyConstraintLocal' instead of '@addLazyConstraint', etc (in the CPLEX C Api the variant is 'cutcallbackaddlocal' instead of 'cutcallbackadd').

Do you think this way of proceeding is consistent with the packages design, or would it be preferable to handle the feature by passing a parameter value (as in AIMMS) throughout the 'call chain' instead - e.g. for maintenance purposes ? And displaying an error message when used with solvers not supporting the feature ?

Thanks a lot!,

Kind Regards,

Mehdi

*http://www.ibm.com/support/knowledgecenter/SS9UKU_12.6.0/com.ibm.cplex.zos.help/refcallablelibrary/mipapi/cutcallbackaddlocal.html

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -162,6 +162,7 @@ const sensemap = Dict(:(<=) => '<', :(==) => '=', :(>=) => '>')
 
 ## Lazy constraints
 export addLazyConstraint, @addLazyConstraint
+export addLazyConstraintLocal, @addLazyConstraintLocal
 
 macro addLazyConstraint(cbdata, x)
     cbdata = esc(cbdata)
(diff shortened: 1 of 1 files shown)


IainNZ:
Hi @madanim 
Thanks for bringing this to our attention. It definitely seems like a reasonable thing to add. We'll look over it and digest it. You mentioned Xpress - we (JuliaOpt people at MIT) haven't had any exposure to it. Have you worked on Julia bindings for it (if that is a thing that is possible)?

mlubin:
Thanks for this, very solid contribution. I'm leaning towards passing in additional parameters to ``@addLazyConstraint`` etc versus exporting new names from JuMP. Are there any other flags besides local/global which might be useful in other contexts to pass through?

madanim:
Nice!

@IainNZ  Unfortunately, I have virtually no exposure to it, though some colleagues are using it and may be interested in such features (?). I think Xpress is less used by academics, yet it is among the most competitive besides CPLEX and Gurobi ...
Correction to my previous comment: SCIP seems also to handle locally valid cuts ...

@mlubin  I do no see right now other important flags to consider which would be solver-independant. I would also go for argument passing. However, assuming the argument would preferably be optional, I do not know if there is a way to make a macro argument/expression optional, as for functions (it seems there is no polymorphism for macros ? Or this is wrong ? Maybe there is some simple way to circumvent this? I'm relatively new to Julia ...).

Kind Regards,

Mehdi

mlubin:
@madanim, there is polymorphism for macros in julia 0.5, but you can also just declare a macro to take varargs and do something different depending on the number of arguments, e.g.,
```
@macro foo(args...)
    if length(args) == 3
        ...
    else
        ...
    end
end
```

mlubin:
Anyway I think it's okay to have separate "local" methods in MathProgBase as you've proposed, but at the JuMP level I'd prefer to expose them via extra arguments to the corresponding macros.

madanim:
@mlubin Thanks for the answer (I just noticed that such 'manual dispatches' are indeed used in 'macros.jl' ...)

Since macros with keyword arguments will be supported in Julia 0.5 (as mentionned by tkelman in #726), the extra arguments approach seems perfect.

mlubin:
@madanim, things will be quieter on our end now that JuMP 0.13 is released. If you implement the changes we discussed we can go ahead and push this through.

madanim:
@mlubin 
thanks for the follow-up! I created a new pull request, sorry for this: I used Pkg.submit with a new branch, etc -->  #763

mlubin:
Closing in favor of #763

Issue #715, opened 2016-03-24
add syntax for more general set inclusion constraints


chriscoey:

chriscoey:
na na na na na na na na POWER CONE!

joehuchette:
What would nice syntax for this look like?

chriscoey:
Maybe we should sit down and discuss this is the broader context of conic modeling in JuMP.

chriscoey:
For reference, here is Convex.jl's exp constraint:
https://github.com/JuliaOpt/Convex.jl/blob/master/src/constraints/exp_constraints.jl

and the log and exp atoms:
https://github.com/JuliaOpt/Convex.jl/blob/master/src/atoms/exp_cone/log.jl
https://github.com/JuliaOpt/Convex.jl/blob/master/src/atoms/exp_cone/exp.jl

emreyamangil:
+1

mlubin:
Possible syntax idea:
```
@constraint(m, A*x <= b, cone = XXX)
```
would constrain ``b - A*x`` to be in cone ``XXX``. As in ``@SDconstraint``, arguments would be vectors by default, no ``.<=`` needed.

chriscoey:
it's a little weird but i guess it lines up somewhat with the `Ax \leq_{\mathcal{K}} b` syntax that Boyd's group uses?

the thing here is that very often you would be writing something like `(t,x) \in SOC` or `(x,y,z) \in Exp`, or `(1, b-Ax) \in SOC` etc. that seems to me the most common way of using these cones

chriscoey:
as an example, let's use this little convex.jl model from pajarito tests
```
P = minimize(-3x - y,
    x >= 1,
    y >= -2,
    3x + 2y <= 30,
    exp(y^2) + x <= 7)
```
I guess in JuMP it would be nice (for now) to write:
```
@variable(m, x >= 1)
@variable(m, y >= -2)
@constraint(m, 3x+2y <= 30)
@variable(m, w)
@constraint(m, w + x <= 7)
@variable(m, u)
@constraint(m, (1, w, u) in :Exp)        # :Exp is (a,b,c): cl(b >= a*exp(c/a), a > 0)
@constraint(m, (1/2, u, y) in :Sqr)      # :Sqr is (a,b,c): cl(b >= a*(c/a)^2/2 = c^2/(2a), a > 0)
```
or to cut out `w`:
```
@variable(m, x >= 1)
@variable(m, y >= -2)
[...]

PR #726, opened 2016-04-09
RFC: big macro rename

mlubin:
We're getting rid of camelCase, it's happening. 

So far I've implemented the change for the main macros.
``@defVar`` -> ``@variable``
``@addConstraint`` -> ``@constraint``
``@setObjective`` -> ``@objective``
``@setNLObjective`` -> ``@NLobjective``
``@addNLConstraint`` -> ``@NLconstraint``
``@addSDPConstraint`` -> ``@SDconstraint``
``@defVars`` -> ``@variables``
``@addConstraints`` -> ``@constraints``
``@addNLConstraints`` -> ``@NLconstraints``

HS109 now looks like:
```julia
m = Model()
@variables m begin
    L[i] <= x[i=1:9] <= U[i], (start = 0.0)
end

@NLobjective(m, Min, 3 * x[1] + 1e-6 * x[1]^3 + 2 * x[2] + .522074e-6 * x[2]^3)

@NLconstraints(m, begin
    x[4] - x[3] + 0.55 >= 0
    x[3] - x[4] + 0.55 >= 0
    2250000 - x[1]^2 - x[8]^2 >= 0
    2250000 - x[2]^2 - x[9]^2 >= 0
    x[5] * x[6] * sin(-x[3] - .25) + x[5] * x[7] * sin(-x[4] - .25) +
        2 * b * x[5]^2 - a * x[1] + 400 * a == 0

    x[5] * x[6] * sin(x[3] - .25) + x[6] * x[7] * sin(x[3] - x[4] - .25) +
        2 * b * x[6]^2 - a * x[2] + 400 * a == 0

    x[5] * x[7] * sin(x[4] - .25) + x[6] * x[7] * sin(x[4] - x[3] - .25) +
        2 * b * x[7]^2 + 881.779 * a == 0

    a * x[8] + x[5] * x[6] * cos(-x[3] - .25) +
        x[5] * x[7] * cos(-x[4] - .25) - 200 * a - 2 * c * x[5]^2 +
        .7533e-3 * a * x[5]^2 == 0

    a * x[9] + x[5] * x[6] * cos(x[3] - .25) +
        x[6] * x[7] * cos(x[3] - x[4] - .25) - 2 * c * x[6]^2 +
        .7533e-3 * a * x[6]^2 - 200 * a == 0

    x[5] * x[7] * cos(x[4] - .25) + x[6] * x[7] * cos(x[4] - x[3] - .25) -
        2 * c * x[7]^2 + 22.938 * a + .7533e-3 * a * x[7] ^2 == 0
end)
```

### DIFF
--- a/examples/basic.jl
+++ b/examples/basic.jl
@@ -20,11 +20,11 @@ using JuMP
 
 m = Model()
 
-@defVar(m, 0 <= x <= 2)
-@defVar(m, 0 <= y <= 30)
+@variable(m, 0 <= x <= 2)
+@variable(m, 0 <= y <= 30)
 
-@setObjective(m, Max, 5x + 3y)
-@addConstraint(m, 1x + 5y <= 3.0)
+@objective(m, Max, 5x + 3y)
+@constraint(m, 1x + 5y <= 3.0)
 
 print(m)
 
--- a/examples/bnatt350.jl
+++ b/examples/bnatt350.jl
@@ -47,7 +47,7 @@ u = MathProgBase.getconstrUB(m_internal)
 vtypes = MathProgBase.getvartype(m_internal)
 
 # populate JuMP model with data from internal model
-@defVar(mod, x[1:n])
+@variable(mod, x[1:n])
 for i in 1:n
     setLower(x[i], xlb[i])
     setUpper(x[i], xub[i])
@@ -55,9 +55,9 @@ for i in 1:n
 end
 At = A' # transpose to get useful row-wise sparse representation
 for i in 1:At.n
-    @addConstraint( mod, l[i] <= sum{ At.nzval[idx]*x[At.rowval[idx]], idx = At.colptr[i]:(At.colptr[i+1]-1) } <= u[i] )
+    @constraint( mod, l[i] <= sum{ At.nzval[idx]*x[At.rowval[idx]], idx = At.colptr[i]:(At.colptr[i+1]-1) } <= u[i] )
 end
-@setObjective(mod, Min, sum{ c[i]*x[i], i=1:n })
+@objective(mod, Min, sum{ c[i]*x[i], i=1:n })
 
 function myheuristic(cb)
     fp = open("data/bnatt350.sol", "r")
--- a/examples/cannery.jl
+++ b/examples/cannery.jl
@@ -33,18 +33,18 @@ function solveCannery(plants, markets, capacity, demand, distance, freight)
   nummarkets = length(markets)
   cannery = Model()
 
-  @defVar(cannery, ship[1:numplants, 1:nummarkets] >= 0)
+  @variable(cannery, ship[1:numplants, 1:nummarkets] >= 0)
 
   # Ship no more than plant capacity
-  @addConstraint(cannery, xyconstr[i=1:numplants],
+  @constraint(cannery, xyconstr[i=1:numplants],
                    sum{ship[i,j], j=1:nummarkets}<=capacity[i])
 
   # Ship at least market demand
-  @addConstraint(cannery, xyconstr[j=1:nummarkets],
+  @constraint(cannery, xyconstr[j=1:nummarkets],
                sum{ship[i,j], i=1:numplants}>=demand[j])
 
   # Minimize transporatation cost
-  @setObjective(cannery, Min,
+  @objective(cannery, Min,
               sum{distance[i,j]* freight * ship[i,j], i=1:numplants, j=1:nummarkets})
 
   solution = solve(cannery)
--- a/examples/cluster.jl
+++ b/examples/cluster.jl
@@ -34,20 +34,20 @@ end
 mod = Model()
 
 # Z >= 0, PSD
-@defVar(mod, Z[1:m,1:m], SDP)
-@addConstraint(mod, Z .>= 0)
+@variable(mod, Z[1:m,1:m], SDP)
+@constraint(mod, Z .>= 0)
 
 # min Tr(W(I-Z))
-@setObjective(mod, Min, trace(W * (eye(m,m) - Z)))
+@objective(mod, Min, trace(W * (eye(m,m) - Z)))
 
 # Z e = e
-@addConstraint(mod, Z*ones(m) .== ones(m))
+@constraint(mod, Z*ones(m) .== ones(m))
 #for i = 1:m
 #    addConstraint(mod, sum([ Z[i,j] for j = 1:m]) == 1)
 #end
 
 # Tr(Z) = k
-@addConstraint(mod, trace(Z) == k)
+@constraint(mod, trace(Z) == k)
 
 solve(mod)
 
--- a/examples/corr_sdp.jl
+++ b/examples/corr_sdp.jl
@@ -20,27 +20,27 @@ using JuMP
 
 m = Model()
 
-@defVar(m, X[1:3,1:3], SDP)
+@variable(m, X[1:3,1:3], SDP)
 
 # Diagonal is 1s
-@addConstraint(m, X[1,1] == 1)
-@addConstraint(m, X[2,2] == 1)
-@addConstraint(m, X[3,3] == 1)
+@constraint(m, X[1,1] == 1)
+@constraint(m, X[2,2] == 1)
+@constraint(m, X[3,3] == 1)
 
 # Bounds on the known correlations
-@addConstraint(m, X[1,2] >= -0.2)
-@addConstraint(m, X[1,2] <= -0.1)
-@addConstraint(m, X[2,3] >=  0.4)
-@addConstraint(m, X[2,3] <=  0.5)
+@constraint(m, X[1,2] >= -0.2)
+@constraint(m, X[1,2] <= -0.1)
+@constraint(m, X[2,3] >=  0.4)
+@constraint(m, X[2,3] <=  0.5)
 
 # Find upper bound
-@setObjective(m, Max, X[1,3])
+@objective(m, Max, X[1,3])
 solve(m)
 println("Maximum value is ", getValue(X)[1,3])
 @assert +0.8719 <= getValue(X)[1,3] <= +0.8720
 
 # Find lower bound
-@setObjective(m, Min, X[1,3])
+@objective(m, Min, X[1,3])
 solve(m)
 println("Minimum value is ", getValue(X)[1,3])
 @assert -0.9779 >= getValue(X)[1,3] >= -0.9799
--- a/examples/diet.jl
+++ b/examples/diet.jl
@@ -55,16 +55,16 @@ function SolveDiet()
     m = Model()
 
     # Variables for nutrition info
-    @defVar(m, minNutrition[i] <= nutrition[i=1:numCategories] <= maxNutrition[i])
+    @variable(m, minNutrition[i] <= nutrition[i=1:numCategories] <= maxNutrition[i])
     # Variables for which foods to buy
-    @defVar(m, buy[i=1:numFoods] >= 0)
+    @variable(m, buy[i=1:numFoods] >= 0)
 
     # Objective - minimize cost
-    @setObjective(m, Min, dot(cost, buy))
+    @objective(m, Min, dot(cost, buy))
 
     # Nutrition constraints
     for j = 1:numCategories
-        @addConstraint(m, sum{nutritionValues[i,j]*buy[i], i=1:numFoods} == nutrition[j])
+        @constraint(m, sum{nutritionValues[i,j]*buy[i], i=1:numFoods} == nutrition[j])
     end
 
     # Solve
@@ -73,7 +73,7 @@ function SolveDiet()
     PrintSolution(status, foods, buy)
 
     # Limit dairy
-    @addConstraint(m, buy[8] + buy[9] <= 6)
+    @constraint(m, buy[8] + buy[9] <= 6)
     println("Solving dairy-limited problem...")
     status = solve(m)
     PrintSolution(status, foods, buy)
--- a/examples/iis.jl
+++ b/examples/iis.jl
@@ -65,10 +65,10 @@ End of IIS
 function iis_example()
     m = Model(solver=GurobiSolver())
 
-    @defVar(m, 0 <= x <= 1)
-    @defVar(m, 0 <= y <= 1)
+    @variable(m, 0 <= x <= 1)
+    @variable(m, 0 <= y <= 1)
 
-    @addConstraint(m, x+y >= 3)
+    @constraint(m, x+y >= 3)
 
     status = solve(m)
     @assert status == :Infeasible
--- a/examples/knapsack.jl
+++ b/examples/knapsack.jl
@@ -20,17 +20,17 @@ using JuMP
 # Maximization problem
 m = Model()
 
-@defVar(m, x[1:5], Bin)
+@variable(m, x[1:5], Bin)
 
 profit = [ 5, 3, 2, 7, 4 ]
 weight = [ 2, 8, 4, 2, 5 ]
 capacity = 10
 
 # Objective: maximize profit
-@setObjective(m, Max, dot(profit, x))
+@objective(m, Max, dot(profit, x))
 
 # Constraint: can carry all
-@addConstraint(m, dot(weight, x) <= capacity)
+@constraint(m, dot(weight, x) <= capacity)
 
 # Solve problem using MIP solver
 status = solve(m)
--- a/examples/maxcut_sdp.jl
+++ b/examples/maxcut_sdp.jl
@@ -24,9 +24,9 @@ function solve_maxcut_sdp(n, W)
 
     # Solve the SDP relaxation
     m = Model()
-    @defVar(m, X[1:n,1:n], SDP)
-    @setObjective(m, Max, dot(L,X))
-    @addConstraint(m, diag(X) .== 1)
+    @variable(m, X[1:n,1:n], SDP)
+    @objective(m, Max, dot(L,X))
+    @constraint(m, diag(X) .== 1)
     solve(m)
 
     # Cholesky the result
--- a/examples/mindistortion.jl
+++ b/examples/mindistortion.jl
@@ -42,18 +42,18 @@ D = [0.0 1.0 1.0 1.0
      1.0 2.0 0.0 2.0
      1.0 2.0 2.0 0.0]
 
-@defVar(m, cSq >= 1.0)
+@variable(m, cSq >= 1.0)
 
-@defVar(m, Q[1:4,1:4], SDP)
+@variable(m, Q[1:4,1:4], SDP)
 
 for i in 1:4
     for j in (i+1):4
-        @addConstraint(m, D[i,j]^2 <= Q[i,i] + Q[j,j] - 2Q[i,j])
-        @addConstraint(m, Q[i,i] + Q[j,j] - 2Q[i,j] <= D[i,j]^2*cSq )
+        @constraint(m, D[i,j]^2 <= Q[i,i] + Q[j,j] - 2Q[i,j])
+        @constraint(m, Q[i,i] + Q[j,j] - 2Q[i,j] <= D[i,j]^2*cSq )
     end
 end
 
-@setObjective(m, Min, cSq)
+@objective(m, Min, cSq)
 
 solve(m)
 
--- a/examples/minellipse.jl
+++ b/examples/minellipse.jl
@@ -40,10 +40,10 @@ W = [1.0 0.0
      0.0 1.0];
 
 mod = Model()
-@defVar(mod, X[1:2,1:2], SDP)
-@setObjective(mod, Min, trace(W*X))
+@variable(mod, X[1:2,1:2], SDP)
+@objective(mod, Min, trace(W*X))
 for i = 1:m
-    @addSDPConstraint(mod, X >= As[i])
+    @SDconstraint(mod, X >= As[i])
 end
 solve(mod)
 
--- a/examples/mle.jl
+++ b/examples/mle.jl
@@ -13,10 +13,10 @@ data = randn(n)
 
 m = Model()
 
-@defVar(m, μ, start = 0.0)
-@defVar(m, σ >= 0.0, start = 1.0)
+@variable(m, μ, start = 0.0)
+@variable(m, σ >= 0.0, start = 1.0)
 
-@setNLObjective(m, Max, (n/2)*log(1/(2π*σ^2))-sum{(data[i]-μ)^2, i=1:n}/(2σ^2))
+@NLobjective(m, Max, (n/2)*log(1/(2π*σ^2))-sum{(data[i]-μ)^2, i=1:n}/(2σ^2))
 
 solve(m)
 
@@ -27,7 +27,7 @@ println("var(data) = ", var(data))
 println("MLE objective: ", getObjectiveValue(m))
 
 # constrained MLE?
-@addNLConstraint(m, μ == σ^2)
+@NLconstraint(m, μ == σ^2)
 
 solve(m)
 println("\nWith constraint μ == σ^2:")
--- a/examples/multi.jl
+++ b/examples/multi.jl
@@ -77,28 +77,28 @@ multi = Model()
 
 #  VARIABLES
 
-@defVar(multi, Trans[1:numorig, 1:numdest, 1:numprod] >= 0)
+@variable(multi, Trans[1:numorig, 1:numdest, 1:numprod] >= 0)
 
 #  OBJECTIVE
 
 length(cost)
 
-@setObjective(multi, Max,
+@objective(multi, Max,
               sum{sum{sum{cost[j, i, p] * Trans[i,j, p],
                         i=1:numorig}, j=1:numdest}, p=1:numprod})
 
 #  CONSTRAINTS
 
 # Supply constraint
-@addConstraint(multi, xyconstr[i=1:numorig, p=1:numprod],
+@constraint(multi, xyconstr[i=1:numorig, p=1:numprod],
                sum{Trans[i,j,p], j=1:numdest} == supply[p,i])
 
 # Demand constraint
-@addConstraint(multi, xyconstr[j=1:numdest, p=1:numprod],
+@constraint(multi, xyconstr[j=1:numdest, p=1:numprod],
                sum{Trans[i,j,p], i=1:numorig} == demand[p,j])
 
 # Total shipment constraint
-@addConstraint(multi, xyconstr[i=1:numorig, j=1:numdest],
+@constraint(multi, xyconstr[i=1:numorig, j=1:numdest],
                sum{Trans[i,j,p], p=1:numprod} - limit[i][j] <= 0)
 limit[2][3]
 status = solve(multi)
--- a/examples/noswot.jl
+++ b/examples/noswot.jl
@@ -41,7 +41,7 @@ u = MathProgBase.getconstrUB(m_internal)
 vtypes = MathProgBase.getvartype(m_internal)
 
 # populate JuMP model with data from internal model
-@defVar(mod, x[1:n])
+@variable(mod, x[1:n])
 for i in 1:n
     setLower(x[i], xlb[i])
     setUpper(x[i], xub[i])
@@ -49,21 +49,21 @@ for i in 1:n
 end
 At = A' # transpose to get useful row-wise sparse representation
 for i in 1:At.n
-    @addConstraint( mod, l[i] <= sum{ At.nzval[idx]*x[At.rowval[idx]], idx = At.colptr[i]:(At.colptr[i+1]-1) } <= u[i] )
+    @constraint( mod, l[i] <= sum{ At.nzval[idx]*x[At.rowval[idx]], idx = At.colptr[i]:(At.colptr[i+1]-1) } <= u[i] )
 end
-@setObjective(mod, Min, sum{ c[i]*x[i], i=1:n })
+@objective(mod, Min, sum{ c[i]*x[i], i=1:n })
 
 function mycutgenerator(cb) # valid cuts
     x_val = getValue(x)
     println("in callback")
-    @addUserCut(cb, x[62]-x[63] <= 0)
-    @addUserCut(cb, x[63]-x[64] <= 0)
-    @addUserCut(cb, x[64]-x[65] <= 0)
-    @addUserCut(cb, 2.08x[52] + 2.98x[62] +   3.47x[72] + 2.24x[82] +  2.08x[92] + 0.25x[51] + 0.25x[61] + 0.25x[71] + 0.25x[81] + 0.25x[91] <= 20.25)
-    @addUserCut(cb, 2.08x[54] + 2.98x[64] +   3.47x[74] + 2.24x[84] +  2.08x[94] + 0.25x[53] + 0.25x[63] + 0.25x[73] + 0.25x[83] + 0.25x[93] <= 20.25)
-    @addUserCut(cb, 2.08x[56] + 2.98x[66] + 3.4722x[76] + 2.24x[86] +  2.08x[96] + 0.25x[55] + 0.25x[65] + 0.25x[75] + 0.25x[85] + 0.25x[95] <= 20.25)
-    @addUserCut(cb, 2.08x[58] + 2.98x[68] +   3.47x[78] + 2.24x[88] +  2.08x[98] + 0.25x[57] + 0.25x[67] + 0.25x[77] + 0.25x[87] + 0.25x[97] <= 20.25)
-    @addUserCut(cb, 2.08x[60] + 2.98x[70] +   3.47x[80] + 2.24x[90] + 2.08x[100] + 0.25x[59] + 0.25x[69] + 0.25x[79] + 0.25x[89] + 0.25x[99] <= 16.25)
+    @usercut(cb, x[62]-x[63] <= 0)
+    @usercut(cb, x[63]-x[64] <= 0)
+    @usercut(cb, x[64]-x[65] <= 0)
+    @usercut(cb, 2.08x[52] + 2.98x[62] +   3.47x[72] + 2.24x[82] +  2.08x[92] + 0.25x[51] + 0.25x[61] + 0.25x[71] + 0.25x[81] + 0.25x[91] <= 20.25)
+    @usercut(cb, 2.08x[54] + 2.98x[64] +   3.47x[74] + 2.24x[84] +  2.08x[94] + 0.25x[53] + 0.25x[63] + 0.25x[73] + 0.25x[83] + 0.25x[93] <= 20.25)
+    @usercut(cb, 2.08x[56] + 2.98x[66] + 3.4722x[76] + 2.24x[86] +  2.08x[96] + 0.25x[55] + 0.25x[65] + 0.25x[75] + 0.25x[85] + 0.25x[95] <= 20.25)
+    @usercut(cb, 2.08x[58] + 2.98x[68] +   3.47x[78] + 2.24x[88] +  2.08x[98] + 0.25x[57] + 0.25x[67] + 0.25x[77] + 0.25x[87] + 0.25x[97] <= 20.25)
+    @usercut(cb, 2.08x[60] + 2.98x[70] +   3.47x[80] + 2.24x[90] + 2.08x[100] + 0.25x[59] + 0.25x[69] + 0.25x[79] + 0.25x[89] + 0.25x[99] <= 16.25)
 end  # End of callback function
 
 # # Tell JuMP/CPLEX to use our callback function
--- a/examples/optcontrol.jl
+++ b/examples/optcontrol.jl
@@ -27,19 +27,19 @@ let
 
     m = Model()
 
-    @defVar(m, -1 <= t[1:(ni+1)] <= 1)
-    @defVar(m, -0.05 <= x[1:(ni+1)] <= 0.05)
-    @defVar(m, u[1:(ni+1)])
+    @variable(m, -1 <= t[1:(ni+1)] <= 1)
+    @variable(m, -0.05 <= x[1:(ni+1)] <= 0.05)
+    @variable(m, u[1:(ni+1)])
 
-    @setNLObjective(m, Min, sum{ 0.5*h*(u[i+1]^2 + u[i]^2) + 0.5*alpha*h*(cos(t[i+1]) + cos(t[i])), i = 1:ni})
+    @NLobjective(m, Min, sum{ 0.5*h*(u[i+1]^2 + u[i]^2) + 0.5*alpha*h*(cos(t[i+1]) + cos(t[i])), i = 1:ni})
 
     # cons1
     for i in 1:ni
-        @addNLConstraint(m, x[i+1] - x[i] - (0.5h)*(sin(t[i+1])+sin(t[i])) == 0)
+        @NLconstraint(m, x[i+1] - x[i] - (0.5h)*(sin(t[i+1])+sin(t[i])) == 0)
     end
     # cons2
     for i in 1:ni
-        @addConstraint(m, t[i+1] - t[i] - (0.5h)*u[i+1] - (0.5h)*u[i] == 0)
+        @constraint(m, t[i+1] - t[i] - (0.5h)*u[i+1] - (0.5h)*u[i] == 0)
     end
 
     solve(m)
--- a/examples/prod.jl
+++ b/examples/prod.jl
@@ -143,53 +143,53 @@ minv = [[dem[p][t+1] * checkpro(p,t, pro, pir, rir) for t=numperiods] for p=1:nu
 prod = Model()
 
 ###  VARIABLES  ###
-@defVar(prod, Crews[0:lastperiod] >= 0)
+@variable(prod, Crews[0:lastperiod] >= 0)
 # Average number of crews employed in each period
 
-@defVar(prod, Hire[numperiods] >= 0)
+@variable(prod, Hire[numperiods] >= 0)
 # Crews hired from previous to current period
 
-@defVar(prod, Layoff[numperiods]>= 0)
+@variable(prod, Layoff[numperiods]>= 0)
 # Crews laid off from previous to current period
 
-@defVar(prod, Rprd[1:numprd, numperiods] >=0)
+@variable(prod, Rprd[1:numprd, numperiods] >=0)
 # Production using regular-time labor, in 1000s
 
-@defVar(prod, Oprd[1:numprd, numperiods]>=0)
+@variable(prod, Oprd[1:numprd, numperiods]>=0)
 # Production using overtime labor, in 1000s
 
-@defVar(prod, Inv[1:numprd, numperiods, 1:life] >=0)
+@variable(prod, Inv[1:numprd, numperiods, 1:life] >=0)
 # a numperiods old -- produced in period (t+1)-a --
 # and still in storage at the end of period t
-@defVar(prod, Short[1:numprd, numperiods]>=0)
+@variable(prod, Short[1:numprd, numperiods]>=0)
 			# Accumulated unsatisfied demand at the end of period t
 
 ###  CONSTRAINTS  ###
 
-@addConstraint(prod, xyconstr[t=numperiods],
+@constraint(prod, xyconstr[t=numperiods],
                sum{pt[p] * Rprd[p,t], p=1:numprd} <= sl * dpp[t] * Crews[t])
 # Hours needed to accomplish all regular-time
 # production in a period must not exceed
 # hours available on all shifts
 
-@addConstraint(prod, xyconstr[t=numperiods],
+@constraint(prod, xyconstr[t=numperiods],
                sum{pt[p] * Oprd[p,t], p:1:numprd}  <= ol[t])
 # Hours needed to accomplish all overtime
 # production in a period must not exceed
 # the specified overtime limit
 
-@addConstraint(prod, Crews[firstperiod-1]==iw)
+@constraint(prod, Crews[firstperiod-1]==iw)
 # Use given initial workforce
 
-@addConstraint(prod,xyconstr[t=numperiods],
+@constraint(prod,xyconstr[t=numperiods],
                Crews[t]== Crews[t-1] + Hire[t]-Layoff[t])
 # Workforce changes by hiring or layoffs
 
-@addConstraint(prod, xyconstr[t=numperiods], cmin[t] <= Crews[t])
-@addConstraint(prod, xyconstr[t=numperiods], Crews[t] <= cmax[t])
+@constraint(prod, xyconstr[t=numperiods], cmin[t] <= Crews[t])
+@constraint(prod, xyconstr[t=numperiods], Crews[t] <= cmax[t])
 # Workforce must remain within specified bounds
 
-@addConstraint(prod, xyconstr[p=1:numprd],
+@constraint(prod, xyconstr[p=1:numprd],
                Rprd[p, firstperiod] + Oprd[p,firstperiod] + Short[p, firstperiod]-Inv[p, firstperiod, 1]
                  == max(0,dem[p][firstperiod]-iinv[p]))
 # 'first demand requirement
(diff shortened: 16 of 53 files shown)


chriscoey:
The names are nicer. But idk what y'all have against camels

mlubin:
Camels can't compete with llamas.

joehuchette:
Just noticed an ambiguity with the new syntax:
```jl
@constraint(m, NLP, x+y == 1)
```
and
```jl
@constraint(m, notNLP, x+y == 1)
```
have different behaviors.

mlubin:
Do we want to put the parsing mode as the first argument?

joehuchette:
I don't think that completely solves the ambiguity. What if you name your model ``NLP``?

mlubin:
The syntax
```
@addConstraint(m, notNLP, x+y == 1)
```
wasn't tested at all by the way, was broken by this change and all tests passed.

joehuchette:
What about a keyword argument like ``kind=NLP``?

mlubin:
Oh, the deprecated version still worked but ``@constraint(m, notNLP, x+y == 1)`` does not.

mlubin:
Eh, ``kind=NLP`` is pretty ugly.

joehuchette:
What about ``@constraint(m, :NLP, x+y==1)``?

mlubin:
I could go for that. Would we still allow
```julia
@constraints m NLP begin

end
```
since it's not ambiguous there? May cause more confusion though.

mlubin:
How about this:
```julia
@constraint(m, NLP, x == 1) # error, don't allow constraint references named NLP
@constraint(m, notNLP, x == 1) # ok, linear constraint named notNLP
@constraint(m, :NLP, xx, sqrt(x) == 1) # ok, nonlinear constraint named xx

@constraints m NLP begin # ok
    xx, sqrt(x) == 1
end

@constraints m :NLP begin # also ok
    xx, sqrt(x) == 1
end

@constraints m begin # error, don't allow constraint references named NLP
    NLP, x == 1
end
```

joehuchette:
I'm not a fan of special-casing things to this degree. I would almost just rather remove creating singleton constraint references like this, since you could just as easily write
```jl
notNLP = @constraint(m, NLP, x+y == )
```

mlubin:
So we'd have this?
```julia
@constraint(m, NLP, notNLP, x+y == 1) # doesn't work
@constraint(m, NLP, notNLP[1:1], x+y == 1) # works
```

joehuchette:
Yes. Not totally satisfying, but seems more principled than "don't name things NLP or SDP"

mlubin:
The assignment syntax
```julia
notNLP = @constraint(m, NLP, x+y == 1)
```
doesn't seem to be very popular, I'm hesitant to start forcing people to use it. It increases the cognitive load since before you didn't have to be aware that this was valid.

joehuchette:
Is the ``@constraint(m, NLP, x+y == 1)`` syntax particularly popular right now?

mlubin:
Calling @davidanthoff @dpo

joehuchette:
Special-casing ``NLP`` and ``SDP`` also means we can't easily let extensions use this syntax

mlubin:
The special case is just an error message to help avoid what I imagine could be a common user mistake of using ``NLP`` instead of ``:NLP``. We could still allow extensions but just not provide this checking, you're own your own if there's an extension which defines a ``:GP`` parser and you call your linear constraint ``GP``.

mlubin:
We could also just go and dump ``@constraint`` in favor of ``@constraints`` everywhere.

mlubin:
*OR* we could make the parsing mode a required parameter and not default to the LPQP parsing. That would also resolve the ambiguity.

mlubin:
That's perhaps not too crazy, and would make @tkelman happy as NLP will no longer be a second-class citizen.

IainNZ:
You're thinking outside the box, and its scary outside the box
#MILPforever

tkelman:
Is the label really needed? Can't linearity be detected automatically? SDP could use the latex inequality symbols maybe? Would extensions absolutely have to plug in to the same names?

mlubin:
@tkelman, automatic detection is really out of the question for now, it would require a complete rewrite of how JuMP represents expressions because the NLP pathway can't process vectorized linear expressions.

> Would extensions absolutely have to plug in to the same names?

Not sure I understood, which same names?

tkelman:
same names = `@constraint` et al. If you're not doing linearity detection, I don't see the benefit of trying to use the same names yet.

mlubin:
Extensions are free to define their own macros, of course. The idea for merging SDP/NLP/default now is that the names ``@NLPconstraint`` and ``@NLPconstraints`` are ugly while
```julia
@constraints m NLP begin
    x^3 <= 5
    sum{sqrt(y[i]),i=1:N} <= 10
end
```
seems pretty decent.

mlubin:
Maybe it's not a great idea to build the syntax around forcing you to specify a parsing mode so that we can leave the door open for a good "automatic" mode in the future.

tkelman:
I don't think NLPconstraint is that ugly. If it were called NLconstraint it would be less annoying of an update to make to existing code. If the manual dispatch has to be there, then it's not really the same function.

diam:
I would go for at least to set of macros:
- one (or several) low level macro (or function?) with ambiguous key-value arguments.
  These macro should be stable while allowing to expand more feature in the future.
  They have not to be pretty.
- a set of macro based on the previous ones, and specialized for one or several
  need 

Also, using a symbol as a constant seems more robust and natural and less intrusive for me.
I know that :NLP has local meaning, and that it is like a string parameter. But what is the name NLP? An integer? An object? What is its namespace and range?

While I knowing nothing about semi-definite programming, What if I make some application specific constants just for my need like SDP for Simple Defined Paper and MDP Multi Defined Paper?
The I call `using JuMP` for a linear solver. Will my constants SDP... be overloaded?

-- Maurice




Le 9 avr. 2016 � 20:33, Miles Lubin <notifications@github.com> a �crit :

dpo:
I do think it's nice to be able to force a parsing mode because algorithms can give special treatment to linear constraints (e.g., work in the nullspace, cache their Jacobian, project easily, etc.)

I'm also looking forward to automatic detection in the future, and if it is in the cards, I'm leaning towards the optional keyword argument. That'll let me either go the easy route with
```
@constraint(m, sum{x[i], i=1:n} == 1)
```
or force the mode, e.g.,
```
@constraint(m, mode=:nonlinear, sum{x[i], i-1:n} == 1)
```

mlubin:
I'm convinced that we should go with symbols like ``:NLP`` and ``:SDP`` over names ``NLP``, ``SDP``. The keyword arguments seem overly verbose, however. It seems like it will be painful to replace ``@addNLConstraint(m, ...)`` with ``@constraint(m, mode=:nonlinear, ...)```.

mlubin:
```julia
@constraints m :nonlinear begin
    x^3 <= 5
    sum{sqrt(y[i]),i=1:N} <= 10
end
```
or
```julia
@constraints m :NLP begin
    x^3 <= 5
    sum{sqrt(y[i]),i=1:N} <= 10
end
```
?

joehuchette:
I'm alright with symbols. Does this mean we should change things to ``@variable(m, x, :Int)`` as well?

Also, can we agree on a way to write blocks of constraints? I much prefer
```jl
@constraints(m, :NLP, begin
    x^3 <= 5
    sum{sqrt(y[i]),i=1:N} <= 10
end)
```
but I think the worst case is us mixing between the two.

mlubin:
Are you referring to the parentheses? We can't force people to use them because Julia parses the macro the same without them.

chriscoey:
I'm with Joey. The parentheses in @constraints make it more consistent,
which translates to easier to remember and understand.

Also maybe we want @variable (m, :Int, x) ? Then symbol or "type" always
comes after model. And we could write a @variables syntax just like
@constraints. If necessary, have to put :Cont for continuous variables,
which isn't so bad since being explicit is good for the modeler.

I think consistency between @variable, @constraints, @objective, and
simplicity, should be the goal.
On Apr 10, 2016 12:11 PM, "Joey Huchette" <notifications@github.com> wrote:

I'm alright with symbols. Does this mean we should change things to
@variable(m,
x, :Int) as well?

Also, can we agree on a way to write blocks of constraints? I much prefer

@constraints(m, :NLP, begin

    x^3 <= 5
    sum{sqrt(y[i]),i=1:N} <= 10
end)

but I think the worst case is us mixing between the two.

—
You are receiving this because you commented.

Reply to this email directly or view it on GitHub
<https://github.com/JuliaOpt/JuMP.jl/pull/726#issuecomment-208007536>

joehuchette:
Yes. I understand they're equivalent, I'm talking about the code that we (developers) write, put in docs/examples, etc.

chriscoey:
In the examples I've seen, there haven't been parentheses. So I think the
examples should use the parenthesis format, if there is no difference to
the parser

mlubin:
I don't have a strong opinion on the parenthesis, fine with me if we use them.

mlubin:
@chriscoey, be careful with using macro names in your responses. Unless it's ``escaped`` it pings the corresponding github user.

chriscoey:
oops. last time i respond from email.

mlubin:
Switching to symbols for variable categories makes sense, but not sure it's quite analogous with parsing modes for constraints. Maybe I'm wrong, but it seems somewhat of a corner case to be declaring a long list of variables that are all integer or binary, whereas a long list of nonlinear constraints seems normal.

mlubin:
Also variable categories are just one of many properties which you can set inside ``@variable`` (also ``start``, ``basename``), and you can change these properties later. Parsing modes aren't just properties of the constraint and you could get an error if you don't put down the correct mode.

chriscoey:
true true

mlubin:
So
```julia
@constraints(m, :nonlinear, begin
    x^3 <= 5
    sum{sqrt(y[i]),i=1:N} <= 10
end)
```
or
```julia
@constraints(m, :NLP, begin
    x^3 <= 5
    sum{sqrt(y[i]),i=1:N} <= 10
end)
```
?

IainNZ:
```julia
for j in 1:n
    # h' = v
    @addNLConstraint(mod,
        h[j] == h[j-1] + 0.5*Δt*(v[j]+v[j-1]))

    # v' = (T-D(h,v))/m - g(h)
    @addNLConstraint(mod,
        v[j] == v[j-1] + 0.5*Δt*(
            (T[j  ] - drag[j  ] - m[j  ]*grav[j  ])/m[j  ] +
            (T[j-1] - drag[j-1] - m[j-1]*grav[j-1])/m[j-1] ))

    # m' = -T/c
    @addNLConstraint(mod,
        m[j] == m[j-1] - 0.5*Δt*(T[j] + T[j-1])/c)
end
```

```julia
@constraints(m, :nonlinear, begin
    # h' = v
    d_h[j=1:n],  h[j] == h[j-1] + 0.5*Δt*(v[j]+v[j-1]))
    # v' = (T-D(h,v))/m - g(h)
    d_v[j=1:n],  v[j] == v[j-1] + 0.5*Δt*(
                    (T[j  ] - drag[j  ] - m[j  ]*grav[j  ])/m[j  ] +
                    (T[j-1] - drag[j-1] - m[j-1]*grav[j-1])/m[j-1] ))
    # m' = -T/c
    d_m[j=1:n],  m[j] == m[j-1] - 0.5*Δt*(T[j] + T[j-1])/c)
end
```

tkelman:
Basically none of the code I've seen anyone in my lab write ever uses the multi constraint block macros, just as a data point. Adding individual constraints inside loops over data or indices ends up a lot more natural to most people I work with, rather than ampl style index sets.

On Julia 0.5 macros support keyword arguments automatically so you wouldn't have to reimplement the dispatch. I still think having to manually implement dispatch is a good sign you should just use separate names for now.

IainNZ:
One liner:
Now:
`@addNLConstraint(mod, h[j] == h[j-1] + 0.5*Δt*(v[j]+v[j-1]))`
After:
`@constraint(mod, :nonlinear, h[j] == h[j-1] + 0.5*Δt*(v[j]+v[j-1]))`
Fairly small difference

joehuchette:
```jl
@addNLConstraint(mod, h[j] == h[j-1] + 0.5*Δt*(v[j]+v[j-1]))
```
or
```
@constraint(mod, :nonlinear, h[j] == h[j-1] + 0.5*Δt*(v[j]+v[j-1]))
```
```
@constraint(mod, mode=:NLP, h[j] == h[j-1] + 0.5*Δt*(v[j]+v[j-1]))
```

mlubin:
Summary of conclusions from the JuMP developers' command center:
- ``:nonlinear`` and ``:SDP`` symbols to indicate parsing mode (keyword argument version is too long)
- We prefer the macro syntax with parentheses
- Warn when somebody names their constraint reference ``nonlinear`` or ``SDP``
- We should add a multiline macro for defining expressions

dpo:
`@constraint(m, mode=:nonlinear, ...)`, `@constraint(m, mode=:linear, ...)`, etc.?

mlubin:
This PR will focus on renaming the macros, then we'll have follow-up PRs for the methods.

mlubin:
@dpo,
```julia
@constraint(m, :nonlinear, sqrt(x) == 1)
```
and
```julia
@constraint(m, x + y == 1)
```
The default will be the current linear/quadratic mode.

tkelman:
`@NLconstraint` would be shorter, simpler to implement, and an easier find-replace on existing code. What's the downside that makes it really necessary to move parsing mode from the macro name to an argument?

mlubin:
> What's the downside that makes it really necessary to move parsing mode from the macro name to an argument?

We're trying to come up with a syntax that we're proud of. ``@NLconstraint`` doesn't really meet the bar for me.

mlubin:
Updated the PR to accept ``:nonlinear`` and ``:SDP`` symbols.

mlubin:
CC @anriseth @yeesian @JackDunnNZ

tkelman:
Consider it temporary until automatic linearity detection becomes feasible? Then it could be an optional thing to force the solver into one behavior or the other, but most of the time you wouldn't need to deal with it.

If it ends up as a mode parameter, the first thing any of my code that uses JuMP is going to do is define a hacky version of `@NLconstraint` so I don't have to deal with it. The parameter is also inherently not extensible, you'd have to teach JuMP's single `@constraint` macro how to interpret any new symbol tags. GP's or chance constraints etc can't tie into this.

mlubin:
> If it ends up as a mode parameter, the first thing any of my code that uses JuMP is going to do is define a hacky version of ``@NLconstraint`` so I don't have to deal with it. 

That seems pretty reasonable. I don't think it's hacky at all, the implementation will be a single line.

> The parameter is also inherently not extensible, you'd have to teach JuMP's single ``@constraint`` macro how to interpret any new symbol tags. GP's or chance constraints etc can't tie into this.

The plan for extensibility is to let extensions register themselves when they're loaded using, e.g., a dictionary in the JuMP module. This should work except in the cases where the code is compiled before the extension is registered.

mlubin:
Or more simply we'd have a parsing method that takes ``Val{:nonlinear}``, ``Val{:GP}``, etc. which extensions could implement.

joehuchette:
What about just having Tony's one-line ``@NLconstraint`` macro live in JuMP? Is it so bad to have two ways to do the same thing?

mlubin:
Maybe unexported.

mlubin:
CC @harshangrjn @emreyamangil

davidanthoff:
I agree with @tkelman, I find the use of different macros for linear and nonlinear constraints less verbose, cleaner and generally simpler. I understand the "proud" argument, but maybe keep that moment for a later stage when automatic detection is implemented?

I actually REALLY like the visuals of adding constraints without the brackets, i.e.

````julia
@NLconstraint m begin
  blah
  blah
end
````

Just as a counter data point to @tkelman's, we use the multi constraint block macros a lot.

Overall, very nice!

**Comment by mlubin** [comment]
@davidanthoff, would you be opposed to making ``@constraint(m, :nonlinear, ...)`` the default and leaving ``@NLconstraint`` defined but unexported? With the rest of the big renaming we're planning on unexporting advanced functions.

I know ``@NLconstraint`` feels comfortable for expert JuMP users, but I really don't want to fall into the trap of getting stuck with bad syntax; the window for making syntax changes like this is closing as JuMP's popularity grows. Automatic detection is not coming anytime soon, possibly post 1.0.

**Comment by mlubin** [comment]
My main opposition to ``@NLconstraint`` is the ``NL``. If there's consensus around a self-explanatory name like ``@nonlinearconstraint`` then I'd be more comfortable putting it out there as the official syntax.

**Comment by davidanthoff** [comment]
I'm not a huge fan of keeping alternative naming around as unexported, I'd rather just adopt the "official" stance, even if I would have preferred a different choice. At the end of the day these are really just small syntactic things that are more about stylistic preferences than anything else, and I could definitely live with any choice made.

I don't mind ``@nonlinearconstraint``. I'm not clear what the general julia naming guidelines are around underscore right now, i.e. whether ``@nonlinear_constraint`` would be better or not?

**Comment by mlubin** [comment]
I just updated most of the tests. Have a look and see if you like the new syntax.

**Comment by mlubin** [comment]
Some before/after highlights:
Before:
```julia
    @defVar(m, X[1:3,1:3], SDP)
    @addSDPConstraint(m, X <= 1/2*eye(3,3))
    @defVar(m, Y[1:5,1:5], Symmetric)
    @addSDPConstraint(m, -ones(5,5) <= Y)
    @addSDPConstraint(m, Y <= 2*ones(5,5))
    @defVar(m, Z[1:4,1:4], Symmetric)
    @addSDPConstraint(m, ones(4,4) >= Z)
```
After:
```julia
    @variable(m, X[1:3,1:3], SDP)
    @constraint(m, :SDP, X <= 1/2*eye(3,3))
    @variable(m, Y[1:5,1:5], Symmetric)
    @constraint(m, :SDP, -ones(5,5) <= Y)
    @constraint(m, :SDP, Y <= 2*ones(5,5))
    @variable(m, Z[1:4,1:4], Symmetric)
    @constraint(m, :SDP, ones(4,4) >= Z)
```
Multiple macro alternative:
```julia
    @variable(m, X[1:3,1:3], SDP)
    @SDPconstraint(m, X <= 1/2*eye(3,3))
    @variable(m, Y[1:5,1:5], Symmetric)
    @SDPconstraint(m, -ones(5,5) <= Y)
    @SDPconstraint(m, Y <= 2*ones(5,5))
    @variable(m, Z[1:4,1:4], Symmetric)
    @SDPconstraint(m, ones(4,4) >= Z)
```
Before:
```julia
    @defVar(m, 1 <= x[i=1:4] <= 5, start=initval[i])
    @setNLObjective(m, Min, x[1]*x[4]*(x[1]+x[2]+x[3]) + x[3])
    @addNLConstraint(m, x[1]*x[2]*x[3]*x[4] >= 25)
    @addNLConstraint(m, sum{x[i]^2,i=1:4} == 40)
```
After:
```julia
    @variable(m, 1 <= x[i=1:4] <= 5, start=initval[i])
    @objective(m, :nonlinear, Min, x[1]*x[4]*(x[1]+x[2]+x[3]) + x[3])
    @constraint(m, :nonlinear, x[1]*x[2]*x[3]*x[4] >= 25)
    @constraint(m, :nonlinear, sum{x[i]^2,i=1:4} == 40)
```
Multiple macro alternative:
```julia
    @variable(m, 1 <= x[i=1:4] <= 5, start=initval[i])
    @nonlinearobjective(m, Min, x[1]*x[4]*(x[1]+x[2]+x[3]) + x[3])
    @nonlinearconstraint(m, x[1]*x[2]*x[3]*x[4] >= 25)
    @conlinearconstraint(m, sum{x[i]^2,i=1:4} == 40)
```

chriscoey:
chiming in with stylistic thoughts (but not much understanding of how JuMP works under the hood):
- can we just have `@var, @obj, @con`? same length so more readable models, more text fits on one line, less to type every time
- if you are using symbols eg :SDP in constraint, maybe should use symbols in the variable too, like `@variable(m, X[1:n,1:n], :SDP)`
- I prefer to use :PSD that :SDP actually. looking ahead we may want to allow specifying conic models in a standard conic form (even though currently it's not considered natural for real-world problems). cones are the future! we may want to support exponential cone and power cone, so perhaps we could use syntax that is extensible for :EXP and :POW. which are analogous to :PSD not :SDP
- this one might make less sense... - for constraints, since you don't always specify what type it is (SDP, nonlinear etc), maybe the type should go at the end after the comparison expr. this is more analogous to how variable is defined. if you add a cont variable and a linear constraint, you don't usually write the type. or else, change @variable: put the optional variable type matter before the variable definition. the argument against this is that variable type stuff can be changed, whereas constraint parsing mode can't. but maybe you might have a nonlinear constraint that you want to change to a conic type constraint, to test different algorithms or whatever, so maybe you want to allow constraint parsing type to change

chriscoey:
but yes I want to reiterate that I think it would be great to think ahead about conic model input/parsing. convex.jl, though it's very slow and too matlab-y, allows us to input DCP models in a very straightforward way. I'd like JuMP to be able to do the same one day.

chriscoey:
My last gripe... the terminology "nonlinear" in this context has always been confusing for me. And I've read the docs at some point in the past. 

So we can specify constraints that are nonlinear (as in, having terms in them that are literally not linear) with or without the :nonlinear. Sometimes we want conic constraints and sometimes we want general smooth nonlinear ones. I know it's about how they are stored and what algorithms can be used. But what is the real distinction here, and can we have a clearer terminology for it?

Furthermore, if we are inputting non-conic constraints that we want to solve with boring old non-conic derivative based solvers, why isn't this a property of the whole model rather than the constraint? If we want a conic model (stored in MPB conic format) then why don't we have :conic be a property of the model, and all constraints will be parsed as such (if possible, otherwise error is thrown). If instead we want a general smooth possibly nonconvex model, then all constraints should be parsed as such, so maybe the model should be of type :nonconic or :general or something?

diam:
Le 14 avr. 2016 � 04:20, Miles Lubin <notifications@github.com> a �crit :

> @objective(m, :nonlinear, Min, x[1]*x[4]*(x[1]+x[2]+x[3]) + x[3])
> ...

> @variable(m, Y[1:5,1:5], Symmetric)

Is it possible to have symbol also for Min and so?

and

Le 14 avr. 2016 � 18:03, Chris C. <notifications@github.com> a �crit :

> � can we just have @var, @obj, @con? same length so more readable models, more text fits on one line, less to type every time

I would also like if the short names were provided as alternative for the standard ones (but @cst instead of @con?).

And to avoid user conflicts, one could require some magic macro (not sure about the possible syntaxe):

  @jump_use_shortnames
  @jump_use_oldnames
  @jump_use_testnames
  ...

Or even one could require a special "using Jump::shortnames"?


-- maurice

JackDunnNZ:
Throwing in my vote for offering short names `@var`, `@con`, `@obj` in some form. I also would prefer symbols in variables and objective (it's less magic that way), but understand it might confuse current users. That said, there's no better time to make a change like this.

chriscoey:
Yeah I'm for symbols everywhere, objective sense included

mlubin:
I'm okay with an extra import for using short names. I'm not sure if they're sufficiently self-explanatory to make them the default and only syntax though. Symbols in the objective already work. I could go for symbols in variables, but that's something that can be done separately from this PR.

The term "nonlinear" means in preparation for derivative-based nonlinear solvers. Don't know of a better name but open to suggestions.

> If instead we want a general smooth possibly nonconvex model, then all constraints should be parsed as such, so maybe the model should be of type :nonconic or :general or :whatever, and then there isn't a need to specify individual constraint parsing modes?

This is a good idea, but the parsing mode need to be decidable at compile time (so that we know how to parse it and generate code) and not a runtime property of the model. We could generate code for all possible parsing modes and then decide which code to use at runtime, but I expect this would be asking too much of the compiler since we already have received complaints about 1000-line functions using JuMP taking 10 minutes to compile.

mlubin:
The blocker right now is to decide whether to have separate macros or symbols for constraint parsing modes.
```
@constraint(m, :nonlinear, ...)
@con(m, :nonlinear, ...) # short form
```
or
```
@nonlinearconstraint(m, ...)
@nonlincon(m, ...) # short form
```
or
```
@NLPconstraint(m, ...)
@NLPcon(m, ...) # short form
```
or
...

I'm on the fence about whether the symbols are sufficiently better than the separate macros that it's worth the change.

chriscoey:
I'm still complaining about "nonlinear"

mlubin:
Is NLP better? It's analogous to SDP.

emreyamangil:
I feel like constraint(m, :nonlinear) makes it worse than addNLconstraint. My confusion was exactly what chris is saying, if its a conic constraint addconstraint works with norm or dot. There are three models conic nonlinear and linearquadratic, why not have separste macros for each instead of intersecting at places? Or tags like nonlinear conic linearquadratic??

chriscoey:
No my confusion arises because to myself and beginners, nonlinear means not
linear. In Freund's class, called NLP, you learn stuff that's beyond LP,
both conic and nonconic stuff, for example. A name that actually
distinguishes would be helpful.

See my earlier post, I don't understand why this parsing mode is specified
per constraint rather than by the whole model. That said, I haven't dug
into the code yet. I feel, if you want a model to be stored in mpb conic
format, you should specify that once. If you want derivatives calculated
for all your constraints, you specify that once.

I am open to being enlightened! I am confused, despite having read the docs
and not being a complete noob.

emreyamangil:
For example, if you do addconstraint(m, dot(x,x) <= 2yz) generates a linearquadratic constraint, but i was hoping for rsoc

chriscoey:
Yeah, what Emre said

chriscoey:
For example, could one specify when they call Model() that they want
everything to be parsed into a conic model? If the constraints or objective
are not conic representable, raise an error.

chriscoey:
Or have conic be default, and make them specify :general or :smooth or
:differentiable or something in Model() and avoid having any constraint
parsing types?

dpo:
> For example, could one specify when they call Model() that they want
everything to be parsed into a conic model? If the constraints or objective
are not conic representable, raise an error.

I think it's useful to specify the parse mode per constraint because it's often useful to evaluate how efficiently a solver is able to take advantage of some knowledge about the structure of constraints. A simple example is a solver that will give special treatment to linear constraints. If the parse mode is specified at the model level, that would be lost.

I'm not a fan of the "NLP", "SDP", etc., acronyms. They mean "nonlinear program(ming)", "semidefinite program(ming)", etc. They refer to the structure of an entire problem, not so much to the structure of an expression. I much prefer `:nonlinear`, `:semidefinite` (or `:conic`), `:linear`, etc., and perhaps `:quadratic` (not sure if that's in JuMP already).

To many optimizers, "nonlinear" means "not linear and not quadratic either", hence the acronyms QP and QCQP. It also typically implies "nonconvex".

mlubin:
As I mentioned, specifying the parsing mode at the model level is not technically feasible. The information needs to be available at macro evaluation time, not runtime.

At some point in the *not* near future we may unify the parsing modes into a general expression tree and then have a second pass that does a structural analysis like DCP for conic or AD for NLP, which would resolve some of these issues.

mlubin:
@chriscoey, none of ``:general``, ``:smooth``, or ``:differentiable`` really encompass NLP, because you're free to use nonsmooth functions like ``ifelse`` and it's up to the solver if it wants to converge or not.

mlubin:
Nonlinear means JuMP is going to call out to a solver which implements the [nonlinear interface](http://mathprogbasejl.readthedocs.org/en/latest/nlp.html) from MathProgBase. That's our definition.

chriscoey:
Ok, so tabling what you said for the future, what about the issue of
ensuring a constraint is parsed conically... If I want it to go in mpb
conic format? Like Emre's example. It would be a move in the right
direction for eventually implementing dcp no?

mlubin:
> what about the issue of ensuring a constraint is parsed conically... If I want it to go in mpb
conic format?

I'd put that under issue https://github.com/JuliaOpt/JuMP.jl/issues/715, I don't see that as blocking this. If we choose symbols for parsing modes then it will look like ``@constraint(m, :conic, ...)`` and if we choose separate macros then it will look like ``@conicconstraint(m, ...)``.

chriscoey:
I'm a little concerned about what is the default constraint parsing mode when you don't specify one though, going back to Emre's issue. Aside from this, as long you're confident that the syntax changes are pretty future-proof for now, given what we have discussed briefly about conic and dcp (which we should discuss more in #715, then I'm satisfied with these syntax changes.

My preference lies with the symbols I think. `@constraint(m, :conic, ...)` should be OK, though down the track it would be nice to just have :conic as a property of the model as I mentioned, though I understand now why this isn't a very straightforward change. We would need `@objective(m, :conic, ...)` too I suppose.

tkelman:
I like the idea of `using JuMP.shortnames` pulling in optional shorthands. Maybe also have it import the optional everything-in-the-name version of the macros? Those aren't really "short," but doesn't seem necessary to make 2 different modules for alternate macro names.

mlubin:
So far I've seen more support for the separate macro versions as opposed to symbols for parsing modes. Leaving aside any proposed feature requests like automatic detection, conic support, DCP, symbols for variable types and objective sense, free kittens, etc., let's have an up/down vote on the following naming scheme, which is just a 1-1 mapping from existing macro names plus a short form.
```
@variable # == @defVar
@variables # == @defVars
@var # short form
@vars

@constraint # == @addConstraint
@constraints # == @addConstraints
@con # short form
@cons

@objective # == @setObjective
@obj # short form

@nonlinearconstraint # == @addNLConstraint
@nonlinearconstraints # == @addNLConstraints
@nlcon # short form
@nlcons

@nonlinearobjective # == @setNLObjective
@nlobj # short form

@semidefiniteconstraint # == @addSDPConstraint
@semidefiniteconstraints # == @addSDPConstraints (oops, this never existed)
@sdcon # short form
@sdcons
```

You can use the github reactions to register thumbs up or thumbs down, but if you give a thumbs down please post a constructive response with an alternative. Voting will "close" in 24 hours.

joehuchette:
I propose ``NLobjective`` and ``NLobjective``, and none of the short form versions.

emreyamangil:
I am pro separate macros, structured macros makes things easier to read imho (unless dcp overhaul happens, which is a long term goal), though i would love to have a addConicConstraint and separate linearquadratic and conic for good, thanks!!

emreyamangil:
And free kittens please!

mlubin:
CC @semartin @evanfields @daisy12321

IainNZ:
> I propose NLobjective, NLobjective, and SDPconstraint), and none of the short form versions.

If we are going with that proposal, I'm definitely having similar thoughts to this. But its not a dealbraker for me.

tkelman:
@ashwincarvalho I was mentioning this to you last night. And heads up to @jgonzal13 too, you'll notice this with some future deprecation warnings when it first gets released, but have a look at the proposals.

mlubin:
So I'm not a fan of the ``@NLconstraint`` style because it's a compromise that doesn't really satisfy anyone in particular. If you want to write a pretty Jupyter notebook or a book (CC @chkwon), it makes sense to use a more explicit and readable name like ``@nonlinearconstraint``. If you're interested in brevity then you'll be happy with ``@nlcon``. I think we'll be increasing the overall utility by offering both options (with the short versions requiring a special import) and letting people decide for themselves. We don't need to be in the business of imposing a certain aesthetic on the code unless we have a compelling solution, which we don't.

chkwon:
What about those for expressions? `@defExpr`, `@defNLExpr`, etc. I guess these will follow the new naming convention once it is settled down. 

One comment regarding `:nonlinear` and  `:SDP` as optional arguments:  To understand those "optional" arguments, I guess one needs to understand the whole structure of JuliaOpt. When different macro names are used, they are just enforced to use those macros; so less confusion I guess. But when the macro name is same but in some cases (linear quadratic) we don't need those optional arguments, in some other cases, we need. 

Would a method like the following work? When calling the model, one needs to declare the model type. linear, nonlinear, SDP, etc. Then all other macros determine the types of constraints and objective functions automatically. Something like

```julia
m = Model(type=:NLP, solver=IpoptSolver())
@objective(m, x[1] + x[2]^3 + ...)
@constraint(m, x[1] + x[2]^3 + ...)
```

Would there be cases where different types of constraints are mixed?

mlubin:
> What about those for expressions? ``@defExpr``, ``@defNLExpr``, etc. I guess these will follow the new naming convention once it is settled down.

Yes, these will follow from the naming scheme we choose for the "main" macros discussed so far.

> Would a method like the following work? When calling the model, one needs to declare the model type. 

This is discussed above. As far as I understand, it's not technically feasible at the moment because the parsing mode needs to be known at macro evaluation time and cannot be a runtime property of the model. @dpo also argues that it would be useful to separate the linear constraints from the nonlinear constraints in a model.

JackDunnNZ:
Is there a particular reason the short names would require a special import rather than just coexisting? It might create issues if people are copy/pasting code from different examples with different imports.

joehuchette:
I don't see a world in which having two different names for exact same thing is a net gain. The only thing more confusing than short, cryptic names is a situation where they're used only half the time, seemingly at random. What do we say to someone who reads @chkwon book with ``@nonlinearconstraint``, and then goes to @JackDunnNZ research code and sees a bunch of ``@nlcon`` everywhere?

> We don't need to be in the business of imposing a certain aesthetic on the code unless we have a compelling solution, which we don't.

``@NLconstraint`` is a great balance between brevity and clarity. Has anyone been confused by the ``NL`` in ``@addNLConstraint`` in the past? AFAICT the only objection is one of aesthetics.

mlubin:
> What do we say to someone who reads @chkwon book with @nonlinearconstraint, and then goes to @JackDunnNZ research code and sees a bunch of @nlcon everywhere?

Hopefully they'll see ``using JuMP.ShortNames`` at the top of the file and then figure it out and move on

joehuchette:
>> What do we say to someone who reads @chkwon book with @nonlinearconstraint, and then goes to @JackDunnNZ research code and sees a bunch of @nlcon everywhere?

> Hopefully they'll see using JuMP.ShortNames at the top of the file and then figure it out and move on

This is _a lot_ to expect from a new user...

emreyamangil:
how about conicconstraint?? Would that be painful to add or break some portions of jump?? (This is directly tying to the ease of access to jump models on pajarito)

mlubin:
It's up to @JackDunnNZ how to write his research code and judge if he thinks potential readers will be confused by the short names. This is such a small aspect of whether a code base is readable that I don't think we should pretend to have control here.

Having to explicitly import the short names will hopefully encourage standalone scripts e.g., notebooks, stackexchange, etc, to either use the long-form names or explain where the short names are coming from.

mlubin:
@emreyamangil, please keep discussion of conic constraints to https://github.com/JuliaOpt/JuMP.jl/issues/715

chriscoey:
I think short names make things a lot more readable. Sometimes my jump
macro expressions get so long that they take up multiple lines and I have
to tab in  and it just becomes a big mess. In notebooks, you want to see
the model, not the jump syntax. The 3 letter abbreviations will make it
easier to run your eyes over a model in a presentation or notebook
involving Julia code, and beginners don't want to have to type such long
expressions all the time when other modeling packages like ampl just allow
you to type constraints line by line without macro or function calls.
On Apr 15, 2016 12:12 PM, "Miles Lubin" <notifications@github.com> wrote:

@emreyamangil <https://github.com/emreyamangil>, please keep discussion of
conic constraints to #715 <https://github.com/JuliaOpt/JuMP.jl/issues/715>

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/JuliaOpt/JuMP.jl/pull/726#issuecomment-210524680>

chriscoey:
Plus, var, obj and con are all obviously variable, objective, constraint.
It's not ambiguous to modelers.

mlubin:
``@nlcon`` and ``@sdcon`` are not obvious, but yes, I think the short names could prove to be popular and we'd be missing out on a boost to the overall utility of JuMP by not allowing them.

joehuchette:
> This is such a small aspect of whether a code base is readable that I don't think we should pretend to have control here.

This is such a weird argument to me. If the short names don't pose a significant readability issue, let's just use them exclusively?

If we're going to bless two different names for the exact same functionality with inclusion in JuMP, what principled argument is there to stop at two? The current names are already used in a bunch of code, notebooks, stackexchange answers, etc. Let's just keep them as the default, and you can access the long form names with ``JuMP.LongNames``.

mlubin:
@joehuchette, you mean keep the camelCase names?

joehuchette:
Yes. If you'd like not to give them preferential treatment, we can ask users to write ``JuMP.camelCase`` at the top of their script. I happen to prefer ``@NLConstraint`` and friends, so we can namespace those under ``JuMP.JoeysNames``. We can also accept PRs for other naming conventions that certain users prefer. 

Having to explicitly import the a particular naming will hopefully encourage standalone scripts e.g., notebooks, stackexchange, etc, to either use the current camelCase names or explain where their prefered names are coming from.

diam:
Just an idea. I think you could provide two levels of macro:

1. the standard low level macros.
   They must be **explicit**, **stable** and **extensible**.
      @variable(m, key1=value1, key2=value2, ...)
   Symbols could be used as an (unique) abbreviation of some frequently used key value pairs.

2. the optional hight level macros.
   @var(m, ...)
   These macro could be clever enough to guess some reasonable default 
   parameters (key=value pairs like :min for direction=Minimize), but one 
   should be able to override them if needed.

These two levels could instead be defined with longer names:

   @defvariable(m, key1=value1, key2=value2, ...)

for the internal stable macros (or functions?), and 

   @variable(m, ...)

for the clever user level ones.
This would allow to require some `using JuMP.ShortNames` for the (clever) 
abbreviations.

Why requiring some using is important?
Because:
1. one can provide **several** set of macro for various domains:
     using JuMP.ShortNamesForConicProgramming
2. this would allow to provide some day a yet more clever but incompatible 
   set of macros:
     using JuMP.ShortNamesTest
   Then make the old deprecated, but yet usable by calling
     using JuMP.YesIRealyWantToUseTheseOldMacros


-- Maurice
  

Le 15 avr. 2016 � 18:03, Joey Huchette <notifications@github.com> a �crit :

anriseth:
I too think the shortnames are cluttering things up. One name per
macro/function call would keep things more consistent, and the proposed
names are good in my opinion.

Also: abbreviations are more likely to be ambiguous. If we start
abbreviating everything, it may be difficult for future functionality to
both have a sensible abbreviation and be easily distinguished from existing
calls.

mlubin:
> If we're going to bless two different names for the exact same functionality with inclusion in JuMP, what principled argument is there to stop at two?

The principled argument is that two is enough and any more is ridiculous, of course. I think the short names are worth putting out there because they could become popular and useful, but it's an experiment and I'm not convinced either way about the readability.

mlubin:
@anriseth so you support ``@nonlinearconstraint`` over ``@NLconstraint``?

joehuchette:
> The principled argument is that two is enough and any more is ridiculous, of course. I think the short names are worth putting out there because they could become popular and useful, but it's an experiment and I'm not convinced either way about the readability.

I don't think that's particularly principled, but OK. Does this not imply that it would be ridiculous to have ``@constraint``, ``@con``, and ``@addConstraint`` coexisting at the same time? We should pick one, wait until after the deprecation period is over, and reevaluate then.

My point here is that there's a reason we deprecate things when we change naming conventions, rather than just leaving the old names around: having redundant ways the same exact thing just leads to horribly confusing code. I think we're on the same page w.r.t. the short names: I would use them all the time, but there's a real usability concern. The solution isn't to roll them out as "experimental names," though.

chriscoey:
How valid are the usability concerns for @var, @con, @obj? I don't yet see
why there is such a concern. I personally would be fine with only one set
of names, but I would prefer the short names rather than the long names if
there is only to be one set.

However, a reason to temporarily roll short names out as experimental would
be to see if any usability issues actually arise. If we get
complaints/issues, roll back, but if people like the added readability and
ease of typing shorter names, deprecate old names.

mlubin:
@chriscoey, be mindful of the github shout-outs with @...

chriscoey:
omg i did it again i promised never to respond by email but i did

IainNZ:
I'm against short names, because I think they encourage bad habits in software development and would not pass any style guide that I'm aware of.
To be precise, I'm very strongly against them as default, and against including them in general.
It feels anti-JuMP as well, given the amount of effort put in over the years to represent best development practices in Julia, from documentation, testing, deprecations, and meaningful use of semvar.

I realize the standard for research code is typically lower, but it feels bad to be actively endorsing them by including them in JuMP core. I don't think its a big deal splitting constraints across lines to stay within 80 columns, and I think with the multiple-constraint syntax this is even less of a problem as you're using multiple lines anyway.

I'm happy for someone to make a package that just exports a whole heap of wrappers, but I don't like being a party to it.

anriseth:
@mlubin I prefer @nonlinearconstraint, yes.

I appreciate the argument for shortnames to increase "readability" of long,
complicated expressions. My way around that has been to define multiple
subexpressions. I think that's a better approach in many cases, which also
keeps things more readable and flexible.

chriscoey:
OK so @IainNZ for readability sake, could we instead extend the begin...end multiple lines of constraints and variables functionality, rather than shortening names? if we have constraints of different types, we have to use lots of macros separately and it's not super readable especially with long names of differing lengths. BUT if we allowed different variable and constraint types in ONE call to `@vars` and `@cons` (or even "named objectives", and define multiple objectives in one go), using symbols on each line to say what type or parse mode to use, then things would be super readable and the long names would be fine

chriscoey:
That suggestion takes us back to `@constraints, @variables, @objectives` and symbols for variable types, objective sense, parsing mode that go in each line of the `begin... end`

mlubin:
You can already define variables of different types in ``@variables``.

chriscoey:
ok I haven't used it yet, but great, so no need to ever have multiple `@defVar, @defVars` currently?

the same would be good for `@constraints` then, for parsing mode if possible.

then the naming issue would be bunk for me. i'd be fine with long ones if I only ever have to write them once, and JuMP code would be super readable

mlubin:
And today at most two different types of constraints can coexist in a single model (linearquadratic, nonlinear, semidefinite), so I don't really see the gain in being able to change the parsing mode within a constraint block versus just using two constraint blocks.

chriscoey:
"today", what about in future?

mlubin:
Maybe in 5 years three different types of constraints will be able to coexist?

chriscoey:
lol. OK. well look i guess i feel less strongly now, but I think we should start using the block definitions in documentation then. to showcase how readable and straightforward JuMP code can be with multiline variable, constraint (even one day objective?) definitions

IainNZ:
Yes I think the plan is to move more towards block definitions.

chriscoey:
for me, if we make a concerted push towards block definitions, then I don't care if I have to type `@nonlinearconstraintpleasethankyougreat` once

mlubin:
Ok, official short names are scrapped for now. Final decision point is:
```julia
@nonlinearconstraint
@nonlinearconstraints
@semidefiniteconstraint
@semidefiniteconstraints
```
versus
```julia
@NLconstraint
@NLconstraints
@SDPconstraint
@SDPconstraints
```
Make your opinions known...

chriscoey:
I don't like SDPconstraint because of the P. Fine with
semidefiniteconstraint or SDconstraint

chriscoey:
Or PSDconstraint is fine

joehuchette:
I'm not sure PSD is technically correct (or precise, at least), as you could write something like ``@PSDconstraint(m, A <= 0)``.

chriscoey:
But with respect to the inequality itself, it's correct

mlubin:
The better way to explain it is that it goes through the semidefinite parsing mode.

chriscoey:
Then I vote for SDconstraint

anriseth:
@nonlinear and @semidefinite gets +1 from me

chkwon:
I agree that official short names will hurt. I agree with @chriscoey and prefer

```julia
@NLconstraint
@NLconstraints
@SDconstraint
@SDconstraints
```

(without P in SDP)

semartin:
I prefer the long version  (especially when we can just stack all the constraints) and also agree with @IainNZ that short names are bad. My personal favorites are
```julia
@nonlinearconstraint
@nonlinearconstraints
@semidefiniteconstraint
@semidefiniteconstraints
```

tkelman:
> when we can just stack all the constraints

Not always, especially when problem structure and data needs to be programmatically generated or loaded. I prefer `NLconstraint` and `SDconstraint` over the longer forms.

davidanthoff:
I'm really glad that there will be only one macro for each thing, rather than a short and long version!

I have no opinion on ``nonlinear`` vs ``NL``, both seem fine to me.

mlubin:
Just to add more confusion to the mix, what about:
```
@NLconstraint
@nonlinearconstraints
@SDconstraint
@semidefiniteconstraints
```
If you're using the block versions then you shouldn't care about the extra typing. If you're using the single line versions then the extra verbosity is annoying. Irregular plurals ftw?

chriscoey:
nah, i say either all super long versions, or all medium long versions

chkwon:
IMHO, `@NLconstraint` looks a little better than `@nonlinearconstraint`, not only it is shorter, but also it resembles `@constraint` more. 

We all know that it will be best that the unified `@constraint` macro handles all cases; but looks impossible. Then I think variants of `@constraint` should at least *look* like the original. I don't know why, but to me, `@NLconstraint` looks like a variant, while `@nonlinearconstraint` looks like an entirely new macro. Perhaps because I am just already familiar with the current `@addNLConstraint` macro. Anyway, this is why I prefer `@NLconstraint` to `@nonlinearconstraint`, although the latter seems to make more sense linguistically and aesthetically. 

One question: if 
```julia
m = Model(:nonlinear)
```
is impossible, would
```julia
using JuMP.Nonlinear
m = Model()
@objective(m, ...)
@constraint(m, ...)
```
be possible? I guess this would potentially limit the mixed constraint types.

tkelman:
Or `NonlinearModel()`, `SDPModel()` or `ConicModel()` ? That would address the compile-time property, but not the ability to specify parsing mode on a per constraint basis. You could do the parsing mode as a parameter to a single `@constraint` macro (the early plan here) in that scheme where asking for an LQ (or conic?) constraint or objective in a nonlinear model would be the non-default opt-in case with non-default syntax.

mlubin:
@tkelman types aren't available at macro expansion time, not sure if that would solve the issue.

mlubin:
@chkwon that's possible but too restrictive, I think. It means you can make only nonlinear models in a scope that imports ``JuMP.Nonlinear``.

emreyamangil:
My +1 goes for NLconstraint and SDconstraint, writing their names explicitly makes it way too long and i think NL and SD is explanatory enough.

evanfields:
I need to read through this more - which I'll do when I get a chance - but for now I'll just note that as a fairly novice JuMP user (and optimizer), I like that macro names are verbs. `@defVar` tells me what I'm doing, whereas a name like `@variable` makes it less clear whether I'm getting, setting, etc.

mlubin:
@evanfields, the names are now declarative. You're declaring a decision variable, declaring constraints attached to a model, declaring the objective, etc.

mlubin:
I'm going to redo this PR with ``@NLconstraint`` and ``@SDconstraint``, and we'll use that as a basis for the next step of the consensus process. Thanks everyone for your input so far!

mlubin:
Updated the PR, should be in a mergable state but code reviews are welcome. The final list of renamed macros [here](https://github.com/JuliaOpt/JuMP.jl/blob/bigrename/src/deprecated.jl). It's actually quite easy to update code to the new names, here's what I did to update the tests and examples:

```
sed -i s/@defVar/@variable/g *.jl
sed -i s/@addConstraint/@constraint/g *.jl
sed -i s/@addNLConstraint/@NLconstraint/g *.jl
sed -i s/@addSDPConstraint/@SDconstraint/g *.jl
sed -i s/@defExpr/@expression/g *.jl
sed -i s/@defNLExpr/@NLexpression/g *.jl
sed -i s/@defNLParam/@NLparameter/g *.jl
sed -i s/@setObjective/@objective/g *.jl
sed -i s/@setNLObjective/@NLobjective/g *.jl
sed -i s/@addLazyConstraint/@lazyconstraint/g *.jl
sed -i s/@addUserCut/@usercut/g *.jl
sed -i s/@defConstrRef/@constraintrefs/g *.jl
```

Voice your support or objections, planning on merging soon!

joehuchette:
docs?

mlubin:
Planning on doing the docs in a single pass after everything else is renamed

PR #733, opened 2016-04-20
explicit warning for developer docs, closes #727

### LINKED ISSUE #727: warning for developer docs
We should add a banner on the RTD version of the docs indicating that it's developer documentation and pointing to the last versioned docs. This is particular important given the big renaming.

mlubin:
Here's my little hack. We do need to merge this before we can tell if it works correctly or not.


mlubin:
Seems to work

PR #735, opened 2016-04-25
allow anonymous variables in @variable

joehuchette:


mlubin:
Examples?
Note there's a merge conflict

joehuchette:
```jl
julia> m = Model();

julia> x = @variable(m)
##7145

julia> y = @variable(m, [1:3])
3-element Array{JuMP.Variable,1}:
 col_2
 col_3
 col_4

julia> z = @variable(m, [i=1:4; isodd(i)])
# printing error...
```

mlubin:
I like it. Depends on https://github.com/JuliaOpt/JuMP.jl/pull/736 if you want to specify bounds?

mlubin:
Re: the printing, maybe we should just allow these for ``Array{Variable}`` and not jump containers at the start? Or is that just a bad idea?

joehuchette:
Yes you'd need #736 for the bounds. I did that first and branched off it, so that's probably the merge conflict.

I think it's a bad idea to split things like that. I'd rather just have ugly printing that doesn't throw an error

joehuchette:
Also I'm not 100% sure the hygiene here is right

mlubin:
Agreed we should support constructing jump containers.

> Also I'm not 100% sure the hygiene here is right

Needs tests :)

IainNZ:
👍

chriscoey:
I literally just needed this

chriscoey:
I literally still need this

joehuchette:
Lol I just need to motivate myself to clean this up

mlubin:
Time to revive this?

joehuchette:
@mlubin @IainNZ @chriscoey wanna take a look now?

IainNZ:
Did we settle on full `lowerbound`/`upperbound`?
It also looks a bit weird having kwargs for those but not the category, but not a deal breaker

joehuchette:
We can add category as a keyword argument pretty easily. Annoying corner case: what does
```jl
x = @variable(m, Int)
```
do?

mlubin:
We should add a category keyword argument but doesn't need to be in this PR.

chriscoey:
The category keyword would have been useful for me recently actually. If it's not hard to get in here, it would be nice to have soon. See #741

test/macros.jl
```diff
@@ -538,3 +538,44 @@ facts("[macros] LB/UB kwargs") do
...
     @fact macroexpand(:(@variable(m, 0 <= j <= 1, upperbound=1))).head --> :error
 end
+
+facts("[macros] Anonymous versions of macros") do
+    m = Model()
+    x = @variable(m, [1:3], lowerbound=1.0)
+    y = @variable(m, [2:3,1:3], upperbound=1.0, lowerbound=0.0, Bin)
+    z = @variable(m, [[:red,:blue]], Int)
+    w = @variable(m, [i=1:4,j=1:4;isodd(i+j)], SemiCont)
+    # v = @variable(m, [i=1:3,j=1:3], Symmetric, lowerbound = eye(3)[i,j])
```
mlubin: Why is this commented?

mlubin:
This breaks extensions, right?

test/macros.jl
joehuchette: The named version was already broken, I have a fix upcoming.

joehuchette:
To the extent they were relying on the unexported macro code, maybe. I think the changes should be backwards compatible, but who knows.

PR #736, opened 2016-04-25
Specify variable bounds with keyword arguments in @variable

joehuchette:


mlubin:
I like it. Needs docs and tests and we can just merge to proceed with https://github.com/JuliaOpt/JuMP.jl/pull/735.
CC @chriscoey @davidanthoff @yeesian, this touches on the discussion in https://github.com/JuliaOpt/JuMP.jl/issues/692 about alignment when complicated bounds are present

mlubin:
What the...

joehuchette:
That's a pretty big picture to convey pretty much zero information @codecov-io...

PR #742, opened 2016-04-27
JuMPConstraint -> AbstractConstraint

IainNZ:
Would also be happy with just `Constraint` but this feels more consistent.


mlubin:
Ok with me

IainNZ:
All green... except codecov?!

PR #743, opened 2016-04-27
get rid of printing shims

mlubin:


mlubin:
This makes much more sense

IainNZ:
Beautiful. I ran into this when updating JuMPeR, couldn't believe how simple it was...
A lot of the complexity that obscured this arises from supporting IJulia anyway, makes me wonder its worth it.

Issue #730, opened 2016-04-15
Indexing 0-based JuMPArray: end keyword throws errors


dourouc05:
When I ran code on an updated JuMP (0.12), indexing arrays of variables at [end] stopped working (while this code perfectly worked in JuMP 0.11). Here is a 

```
using JuMP

m = Model()
@defVar(m, x[0:2], Bin)
@setObjective(m, Max, x[1])
solve(m)

getValue(x[end])
getValue(x)[end]
```

The errors I get (for both examples) are: 

```
ERROR: Failed attempt to index JuMPArray along dimension 1: 3 ∉ 0:2
 in error at error.jl:21
```

(By the way, is it normal this error does not show where it was triggered from my code?)

joehuchette:
Sorry for the slow reply. Unfortunately I don't think there's much we can do here: the compiler rewrites ``x[end]`` to something like ``x[size(x,1)]``. Are you sure this worked on v0.11?

What do you mean by "this error does not show"? This might be a bug, either on our end or with julia, depending on the behavior.

mlubin:
@joehuchette, I think it uses the ``endof`` method, may we shouldn't [define](https://github.com/JuliaOpt/JuMP.jl/blob/4bbca12b8ba771bb2fc6574791234341992124c8/src/JuMPContainer.jl#L130) it?

mlubin:
The current behavior is dangerous:
```
julia> @variable(m, x[2:4]);

julia> x[end]
x[3]
```

joehuchette:
I'm fine with that

mlubin:
I wonder if we can return a special object from ``endof`` analogous to ``Colon()``

mlubin:
No, that doesn't work because that would break ``end-1``

dourouc05:
@joehuchette: to be precise, I tried a larger piece of code that used to work with JuMP 0.11, and it was broken when I retried with JuMP 0.12.

mlubin:
``end`` on multidimensional arrays uses ``size()``, so it's impossible for us to fix this

mlubin:
Maybe the solution is to define ``size`` as an error and implement ``JuMP.size`` if you actually want to access the dimensions?

PR #748, opened 2016-04-28
disable unsafe 'end' syntax

mlubin:
We can reimplement ``size`` once https://github.com/JuliaLang/julia/issues/16104 is resolved, but in the meantime I don't think we have a choice but to turn ``end`` into an error because it can lead to surprisingly incorrect user code.

Ref #730


mlubin:
@joehuchette, okay with this fix?

joehuchette:
Thanks, Miles. I can cherry-pick to release-0.12 and push a patch release tomorrow, if you don't beat me to it.

mlubin:
I'm planning on releasing 0.13 tomorrow unless there are any objections. Do we actually need to patch 0.12?

PR #763, opened 2016-05-10
handling the "local cuts" variant for lazy constraints and user cuts

madanim:
@mlubin

Thanks for the follow-up about local cuts. Here is the suggestion with an optional keyword arg for the lazyconstraint/usercut macros, to specify if the cuts are local or not. The corresponding suggestions in MathProgBase/CPLEX follow but are indeed identical to the previous ones (first pull request). Apologies for having created a new pull request instead of updating the last one.

I gave it a try here starting from the latest master branches today and reusing what is done in the @variable macro in case other (optional) kw args should be added later. The suggestions are really minor, and it works (provided that CPLEX does the job on its side). However, it may not fit everyone's taste regarding names, etc. 

N.B. A small encoutered issue is related to the deprecation of 'MathProgBase.updatemodel!' which is still needed with the current release of the CPLEX package 'as is', so it has been temporarily added back (one line in MathProgBase/LinearQuadratic.jl) to focus on the locally valid cuts question...

--> Many thanks in advance for any modifications/comments.

### DIFF
--- a/doc/callbacks.rst
+++ b/doc/callbacks.rst
@@ -32,6 +32,8 @@ to the callback management code inside JuMP. Next you will do whatever
 analysis of the solution you need to inside your function to generate the new
 constraint before adding it to the model with
 ``@lazyconstraint(cb, myconstraint)``.
+There is an optional keyword option ``localcut`` to ``@lazyconstraint``, which indicates if the lazy constraint that will be added will only apply at the current node and the tree rooted at that node.
+For example, ``@lazyconstraint(cb, myconstraint, localcut=true)``. By default, ``localcut`` is set to ``false``.
 Finally we notify JuMP that this function should be used for lazy constraint
 generation using the ``addlazycallback(m, myLazyConGenerator)`` function
 before we call ``solve(m)``.
@@ -131,8 +133,10 @@ single argument, e.g. ``function myUserCutGenerator(cb)``, where cb is a referen
 to the callback management code inside JuMP. Next you will do whatever
 analysis of the solution you need to inside your function to generate the new
 constraint before adding it to the model with the JuMP macro
-``@addusercut(cb, myconstraint)`` (same limitations as addConstraint).
-Finally we notify JuMP that this function should be used for lazy constraint
+``@usercut(cb, myconstraint)`` (same limitations as addConstraint).
+There is an optional keyword option ``localcut`` to ``@usercut``, which indicates if the user cut that will be added will only apply at the current node and the tree rooted at that node.
+For example, ``@usercut(cb, myconstraint, localcut=true)``. By default, ``localcut`` is set to ``false``.
+Finally we notify JuMP that this function should be used for user cut
 generation using the ``addcutcallback(m, myUserCutGenerator)`` function
 before we call ``solve(m)``.
 
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -160,8 +160,23 @@ const sensemap = Dict(:(<=) => '<', :(==) => '=', :(>=) => '>')
 ## Lazy constraints
 export @lazyconstraint
 
-macro lazyconstraint(cbdata, x)
-    cbdata = esc(cbdata)
+macro lazyconstraint(args...)
+    cbdata = esc(args[1])
+    x = args[2]
+    extra = vcat(args[3:end]...)
+    # separate out keyword arguments
+    kwargs = filter(ex->isexpr(ex,:kw), extra) # filtering expressions corresponding to kw args specs
+    extra = filter(ex->!isexpr(ex,:kw), extra) # others
+
+    localcut_val = false # by default, the lazy constraint is global
+    for ex in kwargs
+        kwarg = ex.args[1]
+        if kwarg == :localcut
+            localcut_val = esc(ex.args[2])   # excepted if otherwise specified ...
+        else error("in @lazyconstraint($(join(args,','))), invalid keyword argument: $(kwarg)")
+        end
+    end
+
     if VERSION < v"0.5.0-dev+3231"
         x = comparison_to_call(x)
     end
@@ -174,14 +189,14 @@ macro lazyconstraint(cbdata, x)
             aff = zero(AffExpr)
             $parsecode
             constr = constructconstraint!($newaff, $(quot(sense)))
-            addlazyconstraint($cbdata, constr)
+            addlazyconstraint($cbdata, constr, localcut=$localcut_val)
         end
     else
         error("Syntax error in @lazyconstraint, expected one-sided comparison.")
     end
 end
 
-function addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData, constr::LinearConstraint)
+function addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData, constr::LinearConstraint; localcut::Bool=false)
     if length(constr.terms.vars) == 0
         MathProgBase.cbaddlazy!(cbdata, Cint[], Float64[], sensemap[sense(constr)], rhs(constr))
         return
@@ -189,17 +204,38 @@ function addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData, constr::Li
     assert_isfinite(constr.terms)
     m::Model = constr.terms.vars[1].m
     indices, coeffs = merge_duplicates(Cint, constr.terms, m.indexedVector, m)
-    MathProgBase.cbaddlazy!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    if localcut
+        MathProgBase.cbaddlazylocal!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    else
+        MathProgBase.cbaddlazy!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    end
+
 end
 
-addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData,constr::QuadConstraint) = error("Quadratic lazy constraints are not supported.")
+addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData,constr::QuadConstraint; localcut::Bool=false) = error("Quadratic lazy constraints are not supported.")
 
 ## User cuts
 export addusercut
 export @usercut
 
-macro usercut(cbdata, x)
-    cbdata = esc(cbdata)
+macro usercut(args...)
+    cbdata = esc(args[1])
+    x = args[2]
+    extra = vcat(args[3:end]...)
+    # separate out keyword arguments
+    kwargs = filter(ex->isexpr(ex,:kw), extra) # filtering expressions corresponding to kw args specs
+    extra = filter(ex->!isexpr(ex,:kw), extra) # others
+
+    localcut_val = false # by default, the user cut is global
+    for ex in kwargs
+        kwarg = ex.args[1]
+        if kwarg == :localcut
+            localcut_val = esc(ex.args[2])   # excepted if otherwise specified ...
+        else error("in @usercut($(join(args,','))), invalid keyword argument: $(kwarg)")
+        end
+    end
+
+    #cbdata = esc(cbdata)
     if VERSION < v"0.5.0-dev+3231"
         x = comparison_to_call(x)
     end
@@ -212,14 +248,14 @@ macro usercut(cbdata, x)
             aff = zero(AffExpr)
             $parsecode
             constr = constructconstraint!($newaff, $(quot(sense)))
-            addusercut($cbdata, constr)
+            addusercut($cbdata, constr, localcut=$localcut_val)
         end
     else
         error("Syntax error in @usercut, expected one-sided comparison.")
     end
 end
 
-function addusercut(cbdata::MathProgBase.MathProgCallbackData, constr::LinearConstraint)
+function addusercut(cbdata::MathProgBase.MathProgCallbackData, constr::LinearConstraint; localcut::Bool=false)
     if length(constr.terms.vars) == 0
         MathProgBase.cbaddcut!(cbdata, Cint[], Float64[], sensemap[sense(constr)], rhs(constr))
         return
@@ -227,7 +263,11 @@ function addusercut(cbdata::MathProgBase.MathProgCallbackData, constr::LinearCon
     assert_isfinite(constr.terms)
     m::Model = constr.terms.vars[1].m
     indices, coeffs = merge_duplicates(Cint, constr.terms, m.indexedVector, m)
-    MathProgBase.cbaddcut!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    if localcut
+        MathProgBase.cbaddcutlocal!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    else
+        MathProgBase.cbaddcut!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    end
 end
 
 ## User heuristic
--- a/test/callback.jl
+++ b/test/callback.jl
@@ -42,6 +42,38 @@ context("With solver $(typeof(lazysolver))") do
     @fact getvalue(y) --> roughly(2.0, 1e-6)
 end; end; end
 
+facts("[callback] Test local lazy constraints") do
+for lazylocalsolver in lazylocal_solvers
+context("With solver $(typeof(lazylocalsolver))") do
+    entered = [false,false, false]
+
+    weights = [24714888; 21118272; 5063487; 23450813; 5598179; 4049178; 13516450; 8385365; 9684076; 31317634; 14084148; 21750211; 29261668; 17996589; 12115244]
+    values = weights
+    W = floor(sum(weights)/2) # knsapsack instance generated according to criteria in 'Hard knapsack Instances' (Chvatal, Op. Res., 1980), with weights composed of random integers in [1:10^(nbItems / 2)]
+    mod = Model(solver=lazylocalsolver)
+    @variable(mod, 0 <= x[1:length(weights)] <= 1, Int)
+    @objective(mod, Max, dot(x, values))
+    @constraint(mod, dot(x, weights) <=  W)
+
+    function mycb_localzero(cb)
+        nodesexpl = MathProgBase.cbgetexplorednodes(cb)
+        if  entered[1] == false && nodesexpl >= 1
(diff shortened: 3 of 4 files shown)


src/callbacks.jl
```diff
@@ -198,8 +217,23 @@ addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData,constr::QuadConstrai
 export addusercut
 export @usercut
 
-macro usercut(cbdata, x)
-    cbdata = esc(cbdata)
+macro usercut(args...)
+  cbdata = esc(args[1])
```
mlubin: Please follow JuMP's 4-space indentation style

src/callbacks.jl
```diff
@@ -160,8 +160,22 @@ const sensemap = Dict(:(<=) => '<', :(==) => '=', :(>=) => '>')
...
+    cbdata = esc(args[1])
+    x = args[2]
+    extra = vcat(args[3:end]...)
+    # separate out keyword arguments
+    kwargs = filter(ex->isexpr(ex,:kw), extra) #filtering expressions corresponding to kw args specs
+    extra = filter(ex->!isexpr(ex,:kw), extra) #others
+
+    localcut_val=false #by default, the lazy constraint is global
+    for ex in kwargs
+      kwarg = ex.args[1]
```
mlubin: Also 4-space indentation here

src/callbacks.jl
```diff
@@ -174,22 +188,27 @@ macro lazyconstraint(cbdata, x)
...
     if length(constr.terms.vars) == 0
         MathProgBase.cbaddlazy!(cbdata, Cint[], Float64[], sensemap[sense(constr)], rhs(constr))
         return
     end
     assert_isfinite(constr.terms)
     m::Model = constr.terms.vars[1].m
     indices, coeffs = merge_duplicates(Cint, constr.terms, m.indexedVector, m)
-    MathProgBase.cbaddlazy!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    if(localcut==true)
+      MathProgBase.cbaddlazylocal!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
```
mlubin: And here

src/callbacks.jl
```diff
@@ -174,22 +188,27 @@ macro lazyconstraint(cbdata, x)
...
+function addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData, constr::LinearConstraint; localcut::Bool=false) #
     if length(constr.terms.vars) == 0
         MathProgBase.cbaddlazy!(cbdata, Cint[], Float64[], sensemap[sense(constr)], rhs(constr))
         return
     end
     assert_isfinite(constr.terms)
     m::Model = constr.terms.vars[1].m
     indices, coeffs = merge_duplicates(Cint, constr.terms, m.indexedVector, m)
-    MathProgBase.cbaddlazy!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    if(localcut==true)
```
mlubin: It's better julia style to just write ``if localcut``

src/callbacks.jl
```diff
@@ -198,8 +217,23 @@ addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData,constr::QuadConstrai
...
+    cbdata = esc(args[1])
+    x = args[2]
+    extra = vcat(args[3:end]...)
+    # separate out keyword arguments
+    kwargs = filter(ex->isexpr(ex,:kw), extra) #filtering expressions corresponding to kw args specs
+    extra = filter(ex->!isexpr(ex,:kw), extra) #others
+
+    localcut_val=false #by default, the user cut is global
+    for ex in kwargs
+      kwarg = ex.args[1]
```
mlubin: Also indentation here

src/callbacks.jl
```diff
@@ -212,22 +246,26 @@ macro usercut(cbdata, x)
...
     if length(constr.terms.vars) == 0
         MathProgBase.cbaddcut!(cbdata, Cint[], Float64[], sensemap[sense(constr)], rhs(constr))
         return
     end
     assert_isfinite(constr.terms)
     m::Model = constr.terms.vars[1].m
     indices, coeffs = merge_duplicates(Cint, constr.terms, m.indexedVector, m)
-    MathProgBase.cbaddcut!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    if(localcut==true)
+      MathProgBase.cbaddcutlocal!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
```
mlubin: And here

madanim:
I think it's OK now

src/callbacks.jl
```diff
@@ -174,22 +188,27 @@ macro lazyconstraint(cbdata, x)
...
-            addlazyconstraint($cbdata, constr)
+            addlazyconstraint($cbdata, constr, localcut=$localcut_val)
         end
     else
         error("Syntax error in @lazyconstraint, expected one-sided comparison.")
     end
 end
 
-function addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData, constr::LinearConstraint)
+function addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData, constr::LinearConstraint; localcut::Bool=false) #
```
mlubin: Dangling ``#`` here?

src/callbacks.jl
```diff
@@ -160,8 +160,22 @@ const sensemap = Dict(:(<=) => '<', :(==) => '=', :(>=) => '>')
...
 export @lazyconstraint
 
-macro lazyconstraint(cbdata, x)
-    cbdata = esc(cbdata)
+macro lazyconstraint(args...)
+    cbdata = esc(args[1])
+    x = args[2]
+    extra = vcat(args[3:end]...)
+    # separate out keyword arguments
+    kwargs = filter(ex->isexpr(ex,:kw), extra) #filtering expressions corresponding to kw args specs
```
mlubin: Nitpicking, but our convention is to put a space after ``#`` for comments.

madanim:
thanks for the info and I understand it's better to keep things clean, the stylistic fixes are in the last update.

mlubin:
Could you add a mention of the ``localcut`` option in the [documentation](http://www.juliaopt.org/JuMP.jl/0.13/callbacks.html) and a test [here](https://github.com/JuliaOpt/JuMP.jl/blob/master/test/callback.jl)? Since CPLEX is the only solver which supports local cuts at the moment, you can add a check to run the test only for CPLEX. Adding a test is your insurance that we won't unintentionally break this feature in the future.

madanim:
Hello @mlubin, here is a proposition for the documentation and tests which has been committed to the pull-request branch. Concerning the tests:

1. Regarding, local lazy constraints, the test consists in using a small ""hard"" knapsack instance (generated according to criteria proposed by Chvatal in a paper entitled 'Hard Knaspack Instances' ...) hard coded in the test, for which many nodes and several feasible solutions are enumerated by CPLEX (with cuts generation and heuristics deactivated). Only one lazy constraint is added at the node of the first feasible solution found, constraining all the variables to be zero, but being applied only locally, it eventually doesn't preclude the existence of a non-null ""optimal"" solution ... It's not perfect but seems to be a simple and rather acceptable way to test if the solver behaves appropriately. (More challenging instances providing improved guarantees could certainly be considered...).

2. Regarding user cuts, it is mutatis mutandis the same but one problem is that such cuts being 'purgeable by default' (set as such by CPLEX.jl), there is no guarantee that they will be used/remain in the cut pools (making then the test trivially satisfied)... However, as of now, the test fails if the user cut is not local, as expected.

3. N.B. 'cbgetexplorednodes' is currently not exported in CplexSolverInterface.jl, which is why calling it without 'CPLEX.' before it is currently commented in the added test sections in JuMP/test/callback.jl. Should I pull a new request regarding the export and change the test sections accordingly ?

Do you think this is OK ?

madanim:
Sorry for the last typo, due to two 'working versions'... Running the test with Pkg.test, the additions are OK: there are a few other issues related to COIN Optimization Services but maybe due to the current package versions on my machine.

doc/callbacks.rst
```diff
@@ -131,8 +133,10 @@ single argument, e.g. ``function myUserCutGenerator(cb)``, where cb is a referen
 to the callback management code inside JuMP. Next you will do whatever
 analysis of the solution you need to inside your function to generate the new
 constraint before adding it to the model with the JuMP macro
-``@addusercut(cb, myconstraint)`` (same limitations as addConstraint).
-Finally we notify JuMP that this function should be used for lazy constraint
+``@usercut(cb, myconstraint)`` (same limitations as addConstraint).
```
mlubin: Good catch.

test/callback.jl
```diff
@@ -66,6 +99,39 @@ context("With solver $(typeof(cutsolver))") do
...
+    values = weights
+    W = floor(sum(weights)/2) # knsapsack instance generated according to criteria in 'Hard knapsack Instances' (Chvatal, Op. Res., 1980), with weights composed of random integers in [1:10^(nbItems / 2)]
+    mod = Model(solver=cutlocalsolver)
+    @variable(mod, 0 <= x[1:length(weights)] <= 1, Int)
+    @objective(mod, Max, dot(x, values))
+    @constraint(mod, dot(x, weights) <=  W)
+
+    global _cutcount_ = 0
+    function mycb_localzero(cb)
+        nodesexpl = CPLEX.cbgetexplorednodes(cb)
```
mlubin: This should be ``MathProgBase.cbgetexplorednodes(cb)``

mlubin:
Seems reasonable. How hard are these "hard" problems? The test shouldn't take more than a few seconds to run. User cuts are definitely hard to test well. I think any criteria that shows the callback is entered and the cuts are local is fine with me.

test/callback.jl
```diff
@@ -42,6 +42,39 @@ context("With solver $(typeof(lazysolver))") do
...
+
+    weights = [24714888; 21118272; 5063487; 23450813; 5598179; 4049178; 13516450; 8385365; 9684076; 31317634; 14084148; 21750211; 29261668; 17996589; 12115244]
+    values = weights
+    W = floor(sum(weights)/2) # knsapsack instance generated according to criteria in 'Hard knapsack Instances' (Chvatal, Op. Res., 1980), with weights composed of random integers in [1:10^(nbItems / 2)]
+    mod = Model(solver=lazylocalsolver)
+    @variable(mod, 0 <= x[1:length(weights)] <= 1, Int)
+    @objective(mod, Max, dot(x, values))
+    @constraint(mod, dot(x, weights) <=  W)
+
+    global lazy_cutcount_ = 0
```
mlubin: See the approach used with the ``entered`` array. This is better than using global variables. It would be something like ``lazy_cutcount_ = [0]`` and then update ``lazy_cutcount_[1]`` inside the callback.

test/callback.jl
```diff
@@ -66,6 +99,39 @@ context("With solver $(typeof(cutsolver))") do
...
+
+    weights = [24714888; 21118272; 5063487; 23450813; 5598179; 4049178; 13516450; 8385365; 9684076; 31317634; 14084148; 21750211; 29261668; 17996589; 12115244]
+    values = weights
+    W = floor(sum(weights)/2) # knsapsack instance generated according to criteria in 'Hard knapsack Instances' (Chvatal, Op. Res., 1980), with weights composed of random integers in [1:10^(nbItems / 2)]
+    mod = Model(solver=cutlocalsolver)
+    @variable(mod, 0 <= x[1:length(weights)] <= 1, Int)
+    @objective(mod, Max, dot(x, values))
+    @constraint(mod, dot(x, weights) <=  W)
+
+    global _cutcount_ = 0
```
mlubin: Same comment about globals

madanim:
Oops, fixes are uploaded. Concerning the test, it takes less than half a second on a standard laptop, yet a few hundred nodes are explored, and several feasible solutions are found. I think the instance is OK. If the cuts are global, which is not the case of the test, for the lazy constraint, a decade of nodes are explored until a feasible solution is found and then rejected with x constrained to be null. For user cuts, this happens at the first explored node after the root node. When e.g. lazy constraints are local as in the test, 5 solutions are found, and the test is validated as expected...

src/callbacks.jl
```diff
@@ -160,8 +160,22 @@ const sensemap = Dict(:(<=) => '<', :(==) => '=', :(>=) => '>')
...
+    # separate out keyword arguments
+    kwargs = filter(ex->isexpr(ex,:kw), extra) # filtering expressions corresponding to kw args specs
+    extra = filter(ex->!isexpr(ex,:kw), extra) # others
+
+    localcut_val = false # by default, the lazy constraint is global
+    for ex in kwargs
+        kwarg = ex.args[1]
+        if kwarg == :localcut
+            localcut_val = esc(ex.args[2])   # excepted if otherwise specified ...
+        end
```
mlubin: Maybe throw an informative error here if users provide a keyword argument that isn't ``localcut``?

src/callbacks.jl
```diff
@@ -198,8 +217,23 @@ addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData,constr::QuadConstrai
...
+    # separate out keyword arguments
+    kwargs = filter(ex->isexpr(ex,:kw), extra) # filtering expressions corresponding to kw args specs
+    extra = filter(ex->!isexpr(ex,:kw), extra) # others
+
+    localcut_val = false # by default, the user cut is global
+    for ex in kwargs
+        kwarg = ex.args[1]
+        if kwarg == :localcut
+            localcut_val = esc(ex.args[2])   # excepted if otherwise specified ...
+        end
```
mlubin: Same comment here

mlubin:
Basically ready to go on my end modulo those two comments. @joehuchette do you sign off?

madanim:
done! 
However, I don't get why there is an issue when I'm  adding the corresponding @fact_throws tests as follows, just besides adding the local cut in the test:
@fact_throws ErrorException @usercut(cb, sum{x[i], i=1:length(weights)} <= 0, badkwarg=true)
I then get an 'ErrorLoad', etc, and then everything is stopped. 
-> Certainly a mistake on my side, but everything looks similar to the case of other errors thrown by e.g. the @lazyconstraint macro...

N.B.During tests, FactCheck complains about some other @fact_throws when testing lazy/user cuts: "Expected: ErrorException, Occurred: MethodError" ...

mlubin:
You can't catch an error thrown inside a macro with ``@fact_throws`` because the error is thrown at compile time. See [here](https://github.com/JuliaOpt/JuMP.jl/blob/0241ebcedb8051f04b27bf6d6d991e5a485f8f8f/test/macros.jl#L517) for an example of a workaround

madanim:
OK, thanks! workaround now applied accordingly

joehuchette:
This looks good to me, thanks for the thorough PR @madanim! I'm going to restart Travis before merging, it still seems to be failing on julia 0.4 for some reason.

joehuchette:
In particular, I'm seeing
```jl
[callback] Test lazy constraints
  > With solver GLPKMathProgInterface.GLPKInterfaceMIP.GLPKSolverMIP
    Failure :: (line:-1) :: With solver GLPKMathProgInterface.GLPKInterfaceMIP.GLPKSolverMIP :: wrong exception was thrown
      Expression: @lazyconstraint cb x ^ 2 <= 1
        Expected: ErrorException
        Occurred: MethodError
    Failure :: (line:-1) :: With solver GLPKMathProgInterface.GLPKInterfaceMIP.GLPKSolverMIP :: wrong exception was thrown
      Expression: @lazyconstraint cb x ^ 2 <= 1
        Expected: ErrorException
        Occurred: MethodError
Out of 8 total facts:
  Verified: 6
  Failed:   2
```

src/callbacks.jl
```diff
@@ -174,22 +189,27 @@ macro lazyconstraint(cbdata, x)
...
+function addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData, constr::LinearConstraint; localcut::Bool=false)
     if length(constr.terms.vars) == 0
         MathProgBase.cbaddlazy!(cbdata, Cint[], Float64[], sensemap[sense(constr)], rhs(constr))
         return
     end
     assert_isfinite(constr.terms)
     m::Model = constr.terms.vars[1].m
     indices, coeffs = merge_duplicates(Cint, constr.terms, m.indexedVector, m)
-    MathProgBase.cbaddlazy!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    if localcut == true
```
joehuchette: Can just be ``if localcut``

mlubin:
Yes, we need this passing tests on julia 0.4 (0.5 is broken) on appveyor and travis before merging.

madanim:
Yep OK, I naively though this was not related to the PR, but indeed it was, sorry for that.

The fix is in the update --> the wrong Exception type was thrown with the following line already in the master branch:

@fact_throws ErrorException @lazyconstraint(cb, x^2 <= 1)

... because the following (now added) method for 'addlazyconstraint' was missing, which throws an error for Quadr. constr and local cuts kwarg :

addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData,constr::QuadConstraint; localcut::Bool=false) = error("Quadratic lazy constraints are not supported.")


("INFO: JuMP tests passed" on my machine :-) )

src/callbacks.jl
```diff
@@ -174,32 +189,54 @@ macro lazyconstraint(cbdata, x)
...
-    MathProgBase.cbaddlazy!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    if localcut
+        MathProgBase.cbaddlazylocal!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    else
+        MathProgBase.cbaddlazy!(cbdata, indices, coeffs, sensemap[sense(constr)], rhs(constr))
+    end
+
 end
 
 addlazyconstraint(cbdata::MathProgBase.MathProgCallbackData,constr::QuadConstraint) = error("Quadratic lazy constraints are not supported.")
```
mlubin: This method should be removed now?

mlubin:
Almost there! Just remove the now-redundant definition I noted above. The final challenge will be to [squash](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit) your 14 commits into one. (Github is supposed to be able to do this automatically now, but I don't see the option.)

madanim:
seems to be done! Is it OK ? (thanks a lot for all the guidance, and I'll try to be more careful/efficient in case of another PR ...)

mlubin:
Thanks again for the contribution!

PR #765, opened 2016-05-12
avoid generating rows for variables already in varcones

mlubin:
JuMP currently generates varcones **and** constraint rows when bounds are zero, which doesn't make any sense.
This is still a WIP because it breaks reduced costs. We now need to get the duals for the var cones when computing the reduced costs.

CC @chriscoey @emreyamangil

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.4
-MathProgBase 0.5 0.6
+MathProgBase 0.5.1 0.6
 ReverseDiffSparse 0.5.3 0.6
 ForwardDiff 0.1 0.2
 Calculus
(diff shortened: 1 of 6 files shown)


chriscoey:
Also need to remove all 0 rows in zero cones - this fix doesn't do that

mlubin:
@chriscoey: that's https://github.com/JuliaOpt/JuMP.jl/issues/766

emreyamangil:
Cool^2!!

chriscoey:
geomean(Cool, SomeonePleaseFix766)!!

mlubin:
This should be good to go after https://github.com/JuliaLang/METADATA.jl/pull/5312. I think our coverage is lacking though... tests pass whether or not we flip the signs of the reduced costs.

PR #770, opened 2016-05-17
fix updating linear objective for nonlinear models

### LINKED ISSUE #704: @setObjective not updating objective for nonlinear program
I'd like to resolve a NLP after changing the objective. With JuMP 0.12 I've found @setObective does not actually update the model. Here is a minimum working example:

```
using JuMP
using Ipopt

EnableNLPResolve()

m = Model(solver=IpoptSolver())

@defVar(m, 0 <= x <= 1)
@defVar(m, 0 <= y <= 1)
@defVar(m, z1)
@defVar(m, z2)

@addNLConstraint(m, x^2 + y^2 >= 1/2)
@addConstraint(m, z1 == x + 0.1*y)
@addConstraint(m, z2 == 0.1*x + y)


@setObjective(m, Min, z1)
solve(m)

println("x = ",string(getValue(x)))
println("y = ",string(getValue(y)))
println("z1 = ",string(getValue(z1)))
println("z2 = ",string(getValue(z2)))


@setObjective(m, Min, z2)
solve(m)

println("x = ",string(getValue(x)))
println("y = ",string(getValue(y)))
println("z1 = ",string(getValue(z1)))
println("z2 = ",string(getValue(z2)))
```

Unfortunately the second solve gives the same results are the first, which suggests the model is not being updated.

mlubin:
fixes #704

### DIFF
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -94,6 +94,11 @@ end
 
 setobjective(m::Model, sense::Symbol, x::Variable) = setobjective(m, sense, convert(AffExpr,x))
 function setobjective(m::Model, sense::Symbol, a::AffExpr)
+    if isa(m.internalModel, MathProgBase.AbstractNonlinearModel)
+        # Give the correct answer when changing objectives in an NLP.
+        # A better approach would be to update and reuse the evaluator
+        m.internalModelLoaded = false
+    end
     setobjectivesense(m, sense)
     m.obj = convert(QuadExpr,a)
 end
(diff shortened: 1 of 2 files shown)


src/affexpr.jl
```diff
@@ -94,6 +94,11 @@ end
 
 setobjective(m::Model, sense::Symbol, x::Variable) = setobjective(m, sense, convert(AffExpr,x))
 function setobjective(m::Model, sense::Symbol, a::AffExpr)
+    if isa(m.internalModel, MathProgBase.AbstractNonlinearModel)
+        # Give the correct answer when changing objectives in an NLP.
+        # A better approach would be to update and reuse the evaluator
+        m.internalModelLoaded = false
```
tkelman: yikes, won't this throw away all constraint info?

mlubin: Yes, but it's better than giving an incorrect answer. We do exactly the same thing when a nonlinear objective is changed with ``@NLobjective`` or if a new constraint is added.

PR #776, opened 2016-05-24
Add getconstraint (close #639)

### LINKED ISSUE #639: Add getVar analogous to get constraint name
JuMP can retrieve a variable (column) ID by using getVar that enables code Modularization.

However JuMP currently does not support an equivalent for retrieving a constraint (line) ID, that is extremely important for code modularization that uses features such as: getDual and changeRHS.

The first step is probably adding a field, lineNames, in the Model type to keep track of the added constraint names just like the already existing colNames.

This is connected to this [topic](https://groups.google.com/forum/#!topic/julia-opt/K1_tzElY9M4) in julia-opt

Might be linked to this issue #462 also.

joehuchette:


src/JuMP.jl
```diff
@@ -660,6 +677,16 @@ function getvariable(m::Model, varname::Symbol)
     end
 end
 
+function getconstraint(m::Model, conname::Symbol)
+    if !haskey(m.conDict, conname)
+        error("No variable with name $conname")
+    elseif m.conDict[conname] === nothing
+        error("Multiple variables with name $conname")
```
mlubin: s/variables/constraints/

mlubin:
Shouldn't this work for ``@NLconstraint`` and ``@SDconstraint`` also?

joehuchette:
This is better implemented on top of #735  at this point.

joehuchette:
Rebased onto master

mlubin:
Bump. @joehuchette, could you fix the typo above, get this working for ``@NLconstraint``, and add tests?

joehuchette:
Should be able to get to it this weekend.

PR #777, opened 2016-05-25
Better error messages in @variable

joehuchette:


mlubin:
Do we have reasonable coverage for these error cases?

mlubin:
Codecov does say that 100% of the diff is hit

PR #782, opened 2016-06-01
disallow colons in variable index sets

mlubin:
I feel like there's a mental block on the indexing with colon issue, hopefully this pushes us forward. Letting people index with colons as keys just seems like a terrible idea in any case, given the syntax confusion.

We can even merge and let it hang around for a release before we implement slicing.

Ref https://github.com/JuliaOpt/JuMP.jl/issues/643

### DIFF
--- a/src/JuMPContainer.jl
+++ b/src/JuMPContainer.jl
@@ -63,6 +63,14 @@ end
 
 Base.isempty(d::JuMPContainer) = isempty(_innercontainer(d))
 
+coloncheck(args::Number...) = nothing
+
+function coloncheck(args...)
+    if any(t -> t == Colon(), args)
+        error("Colons not allowed as keys in JuMP containers.")
+    end
+end
+
 # generate and instantiate a type which is indexed by the given index sets
 # the following types of index sets are allowed:
 # 0:K -- range with compile-time starting index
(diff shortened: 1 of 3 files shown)


joehuchette:
Why make this a generated function?

mlubin:
> Why make this a generated function?

This way the check gets lifted to compile time in most cases.

joehuchette:
Only in the case with well-typed index sets, where the compiler should be
able to optimize away ``isa(foo, Colon)`` anyway. In the case of
heterogeneous arrays, this will potentially be much more expensive than a
runtime check.

mlubin:
What are you proposing instead, and why would it be faster than generated functions?

joehuchette:
Pathological case:
```jl
using JuMP

type Foo{T} end

N = 100
I = [Foo{T} for T in 1:N]

m = Model()
@time @variable(m, x[I,I;true])
```
First run (i.e. compile time) takes 16 seconds on this branch and 0.5 seconds on master (this is on v0.5; v0.4 is much slower for both). Just make it a runtime check:
```jl
function coloncheck(args...)
    if any(t -> t == Colon, args)
        return error("Colons not allowed as keys in JuMP containers.")
    else
        return nothing
    end
end
```
This gives roughly the same performance as current master.

mlubin:
How about this version?

joehuchette:
LGTM

mlubin:
Did you check for regressions?

joehuchette:
No regressions on my example case.

PR #794, opened 2016-06-22
Write out default bounds in MPS writer. Closes #792

### LINKED ISSUE #792: writeMPS doesn't output Lower bounds by default
I know _.mps_ is not used so much, but I tried to create one, for my problem using the `writeMPS()` function. Solvers like CBC, Gurobi (on NEOS server) did not accept the _.mps_ generated by JuMP.

I then checked with the .mps I had generated using python PuLP and found that PuLP writes the lower bounds by default.

I modified the `writeMPS` function to also write the lower bounds by default, and then it worked fine.

Sam

odow:
Solvers such as Gurobi require an explicit statement of variable bounds in MPS format even when that variable is `0 <= x <= Inf`.


mlubin:
Thanks, @odow!

PR #805, opened 2016-07-13
fixes for julia 0.5

mlubin:
First stab at getting tests passing on Julia 0.5. Needs more work.


mlubin:
Passing now on 0.5 modulo OS X nightlies being broken.

PR #807, opened 2016-07-18
support for generators instead of curly syntax in Julia 0.5

mlubin:
Now that Julia master has support for generator syntax with filters (https://github.com/JuliaLang/julia/issues/550), we can now support this:
```jl
@constriant(m, sum(x[i] for i in 1:N if i != 2) <= 1)
```
This PR enables support for generator syntax wherever we used to use curly braces. Once we drop support for Julia 0.4, we should also deprecate the curly syntax.

### DIFF
--- a/src/parseExpr_staged.jl
+++ b/src/parseExpr_staged.jl
@@ -670,6 +670,8 @@ function parseExpr(x, aff::Symbol, lcoeffs::Vector, rcoeffs::Vector, newaff::Sym
             return parseExpr(numerator, aff, lcoeffs,vcat(esc(:(1/$denom)),rcoeffs),newaff)
         elseif isexpr(x,:call) && length(x.args) >= 2 && isexpr(x.args[2],:generator)
             return newaff, parseGenerator(x,aff,lcoeffs,rcoeffs,newaff)
+        elseif isexpr(x,:call) && length(x.args) >= 2 && isexpr(x.args[2],:flatten)
+            flatten_error(x.args[2])
         elseif x.head == :curly
             return newaff, parseCurly(x,aff,lcoeffs,rcoeffs,newaff)
         else # at lowest level?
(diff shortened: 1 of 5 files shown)


mlubin:
Everything's passing. Will merge in a few days if there are no objections, followed by a post to julia-opt calling for testers for the new syntax.

joehuchette:
We're not losing any functionality here, correct? If so, I'm all for it.

mlubin:
No loss of functionality, if this syntax had been around 3 years ago we would have definitely chosen it over the curly braces.

src/parseExpr_staged.jl
```diff
@@ -443,6 +443,152 @@ function parseNorm(normp::Symbol, x::Expr, aff::Symbol, lcoeffs, rcoeffs, newaff
...
+        cond = x.args[2].args[1]
+        # generate inner loop code first and then wrap in for loops
+        inneraff, innercode = parseExpr(x.args[1], aff, lcoeffs, rcoeffs, aff)
+        code = quote
+            if $(esc(cond))
+                $innercode
+            end
+        end
+        for level in length(x.args[2].args):-1:2
+            _idxvar, idxset = parseIdxSet(x.args[2].args[level]::Expr)
```
joehuchette: Why is the type assert needed?

mlubin: Dunno, just following the previous code here

joehuchette:
What about deprecating the curly bracket syntax?

mlubin:
> What about deprecating the curly bracket syntax?

I was thinking about holding off on that until 0.15, if 0.14 is the last release that supports 0.4. People who want to be able to use the same JuMP models on Julia 0.4 and 0.5 without annoying deprecation warnings can then use JuMP 0.14.

mlubin:
This is also a more invasive change than the big renaming, since it requires more than find and replace. Hopefully @tkelman won't tell us that we'll be driving cars off the road because of it.

IainNZ:
Its going to be nice to not have to explain the `{` anymore!

PR #809, opened 2016-07-26
Fix various v0.5 deprecation warnings + add minimum version requirement for Compat

jrevels:


src/operators.jl
```diff
@@ -555,7 +555,7 @@ end
 const op_hint = "Are you trying to build a nonlinear problem? Make sure you use @NLconstraint/@NLobjective."
 for (func,_) in Calculus.symbolic_derivatives_1arg(), typ in [:Variable,:AffExpr,:QuadExpr]
     errstr = "$func is not defined for type $typ. $op_hint"
-    @eval Base.($(quot(func)))(::$typ) = error($errstr)
+    @eval Base.$(func)(::$typ) = error($errstr)
```
mlubin: I'm surprised this works without compat on 0.4

mlubin:
Thanks!

mlubin:
So the tests are [still printing](https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/147574017#L607) deprecation warnings for ``writemime`` even though JuMP doesn't use writemime anymore. Any ideas?

jrevels:
It's coming from the use of `writemime` [here](https://github.com/JuliaOpt/JuMP.jl/blob/master/test/print.jl#L23). Looking at Compat's documentation, it seems there is a `@compat` fix for *defining* `show` in place of `writemime` (which I applied in this PR), but I don't see a fix in the docs for *using* `show` in place of `writemime`.

Might need to make a PR to Compat for that, unless I missed something (maybe `Compat.show` works?).

PR #810, opened 2016-07-29
deprecation fixes for julia 0.5

mlubin:
Thanks @jrevels


src/JuMP.jl
```diff
@@ -23,8 +23,6 @@ using Compat
 
 import Compat: UTF8String, view
 
-@compat import Base.show
-
```
jrevels: I thought this was still needed for the `writemime` patch? See [here](https://github.com/JuliaLang/Compat.jl/pull/219).

mlubin: Probably. I broke 0.4 with this, need to fix up.

mlubin:
JuMP will be deprecation-free after this, I think. Solver wrappers need work though.

PR #811, opened 2016-07-29
accept nonlinear expressions in Expr form, no macros needed

mlubin:
This provides a new way to input nonlinear expressions by providing Julia ``Expr`` objects. I can imagine this being useful for extensions or some cases of embedding JuMP where the nonlinear expressions are generated programmatically. There's no need to force expressions into macros anymore. See the new tests for examples.

CC @jalving


mlubin:
If there are no objections to the naming, I'll add some docs and merge this.

mlubin:
Docs are done

PR #814, opened 2016-07-31
Permit info callbacks in states other than MIPInfo

yeesian:
The use-case is to allow for the extraction of intermediate solutions with information callbacks.

ref: https://github.com/JuliaOpt/MathProgBase.jl/pull/115, https://github.com/JuliaOpt/Gurobi.jl/pull/59, https://github.com/JuliaOpt/CPLEX.jl/pull/80, https://github.com/JuliaOpt/GLPKMathProgInterface.jl/pull/32

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -18,6 +18,7 @@ type HeuristicCallback <: JuMPCallback
 end
 type InfoCallback <: JuMPCallback
     f::Function
+    when::Symbol
 end
 
 type CallbackAbort <: Exception end
@@ -38,9 +39,22 @@ function addheuristiccallback(m::Model, f::Function)
     m.internalModelLoaded = false
     push!(m.callbacks, HeuristicCallback(f))
 end
-function addinfocallback(m::Model, f::Function)
+
+function _unspecifiedstate()
+    warn("""Info Callbacks without a when clause will currently default
+         to fire only in the :Intermediate state to preserve its behavior
+         with previous versions of JuMP.
+
+         This behavior will be deprecated in subsequent versions of JuMP, so
+         please rewrite all invocations of addinfocallbacks(m, f) to
+         addinfocallbacks(m, f, when = :Intermediate) instead.
+         """)
+    :Intermediate
+end
+
+function addinfocallback(m::Model, f::Function; when::Symbol = _unspecifiedstate())
     m.internalModelLoaded = false
-    push!(m.callbacks, InfoCallback(f))
+    push!(m.callbacks, InfoCallback(f, when))
 end
 
 function lazycallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vector{LazyCallback})
@@ -138,10 +152,16 @@ end
 
 function infocallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vector{InfoCallback})
     state = MathProgBase.cbgetstate(d)
-    @assert state == :MIPInfo
+    if state == :MIPSol
+        MathProgBase.cbgetmipsolution(d,m.colVal)
+    elseif state == :MIPNode
+        MathProgBase.cbgetlpsolution(d,m.colVal)
+    end
     try
         for cb in cbs
-            cb.f(d)
+            if cb.when == state
+                cb.f(d)
+            end
         end
     catch y
         if isa(y, CallbackAbort)
--- a/test/callback.jl
+++ b/test/callback.jl
@@ -216,8 +216,9 @@ for infosolver in info_solvers
 context("With solver $(typeof(infosolver))") do
     nodes      = Int[]
     objs       = Float64[]
+    objvals    = Float64[]
     bestbounds = Float64[]
-    entered = [false,false]
+    entered = fill(false, 2)
 
     N = 10000
     include(joinpath("data","informational.jl"))
@@ -226,16 +227,22 @@ context("With solver $(typeof(infosolver))") do
     @objective(mod, Max, dot(r1,x))
     @constraint(mod, c[i=1:10], dot(r2[i],x) <= rhs[i]*N/10)
     # Test that solver fills solution correctly
-    function myinfo(cb)
+    function myinfo1(cb)
+        @assert MathProgBase.cbgetstate(cb) == :Intermediate
         entered[1] = true
         push!(nodes,      MathProgBase.cbgetexplorednodes(cb))
         push!(objs,       MathProgBase.cbgetobj(cb))
         push!(bestbounds, MathProgBase.cbgetbestbound(cb))
     end
-    addinfocallback(mod, myinfo)
-    addinfocallback(mod, cb -> (entered[2] = true))
+    function myinfo2(cb)
+        @assert MathProgBase.cbgetstate(cb) == :MIPSol
+        push!(objvals, dot(r1,JuMP.getvalue(x[:])))
+    end
+    addinfocallback(mod, myinfo1, when = :Intermediate)
+    addinfocallback(mod, myinfo2, when = :MIPSol)
+    addinfocallback(mod, cb -> (entered[2] = true), when = :Intermediate)
     @fact solve(mod) --> :Optimal
-    @fact entered --> [true,true]
+    @fact entered --> fill(true, 2)
     mono_node, mono_obj, mono_bestbound = true, true, true
     for n in 2:length(nodes)
         mono_node &= (nodes[n-1] <= nodes[n] + 1e-8)
@@ -247,6 +254,9 @@ context("With solver $(typeof(infosolver))") do
     @fact mono_node      --> true
     @fact mono_obj       --> true
     @fact mono_bestbound --> true
+    for (s1,s2) in zip(objvals[1:end-1], objvals[2:end])
+        @fact  s1 <= s2 --> true
+    end
 end; end; end
 
 facts("[callback] Callback exit on CallbackAbort") do
@@ -275,6 +285,6 @@ cbc && facts("[callback] Solver doesn't support callbacks") do
     addheuristiccallback(mod, mycb)
     @fact_throws ErrorException solve(mod)
     mod = Model(solver=Cbc.CbcSolver())
-    addinfocallback(mod, mycb)
+    addinfocallback(mod, mycb, when = :Intermediate)
     @fact_throws ErrorException solve(mod)
 end
--- a/doc/callbacks.rst
+++ b/doc/callbacks.rst
@@ -287,7 +287,7 @@ including::
 Informational Callbacks
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-Sometimes it can be useful to track solver progress without actually changing the algorithm by adding cuts or heuristic solutions. In these cases, informational callbacks can be added, wherein statistics can be tracked via the ``cbget`` functions discussed in the previous section. Informational callbacks are added to a JuMP model with the ``addinfocallback(m::Model, f::Function)`` function.
+Sometimes it can be useful to track solver progress without actually changing the algorithm by adding cuts or heuristic solutions. In these cases, informational callbacks can be added, wherein statistics can be tracked via the ``cbget`` functions discussed in the previous section. Informational callbacks are added to a JuMP model with the ``addinfocallback(m::Model, f::Function; when::Symbol)`` function, where the `when` argument should be one of ``:MIPNode``, ``:MIPSol`` or ``:Intermediate`` (listed under ``cbgetstate()`` in the `MathProgBase documentation <https://mathprogbasejl.readthedocs.io/en/latest/lpqcqp.html#cbgetstate>`_)
 
 For a simple example, we can add a function that tracks the best bound and incumbent objective value as the solver progresses through the branch-and-bound tree::
 
@@ -308,7 +308,7 @@ For a simple example, we can add a function that tracks the best bound and incum
         bestbound = MathProgBase.cbgetbestbound(cb)
         push!(bbdata, NodeData(time(),node,obj,bestbound))
     end
-    addinfocallback(m, infocallback)
+    addinfocallback(m, infocallback, when = :Intermediate)
 
     solve(m)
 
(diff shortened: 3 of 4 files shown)


test/callback.jl
```diff
@@ -228,9 +228,11 @@ context("With solver $(typeof(infosolver))") do
     # Test that solver fills solution correctly
     function myinfo(cb)
         entered[1] = true
-        push!(nodes,      MathProgBase.cbgetexplorednodes(cb))
-        push!(objs,       MathProgBase.cbgetobj(cb))
-        push!(bestbounds, MathProgBase.cbgetbestbound(cb))
+        if MathProgBase.cbgetstate(cb) == :Other
+            push!(nodes,      MathProgBase.cbgetexplorednodes(cb))
```
mlubin: What does this mean? It's invalid to call ``MathProgBase.cbgetexplorednodes`` if state is not ``:Other``? How are people supposed to know what is valid to call inside an info callback?

yeesian: Previously, the callback will only fire when the state is `:MIPInfo` (which is supposed to be `:Other` in MPB-land). It's to let current tests pass, since Gurobi.jl does not allow for `MPB.cbgetobj` to work when the state is `:MIPSol`.

mlubin:
I don't really get this ``:Other`` thing. Can we update the MPB docs to make info callbacks more coherent?

yeesian:
From my reading (primarily of the master callback functions),

- [GLPKMathProgInterface.jl](https://github.com/JuliaOpt/GLPKMathProgInterface.jl/blob/a07e520389e2414dacc06f8978e761cc07197376/src/GLPKInterfaceMIP.jl#L90-L167) and [Gurobi.jl](https://github.com/JuliaOpt/Gurobi.jl/blob/d6ce6f827acde1d5dc8d9002f81189ed20fbcb3a/src/GurobiSolverInterface.jl#L477-L530) sets `state` to one of `:MIPNode`, `:MIPSol` or `:MIPInfo`
- [CPLEX.jl](https://github.com/JuliaOpt/CPLEX.jl/blob/e06d89bb5e3837b922e707ad07047f2c78e04672/src/CplexSolverInterface.jl#L447-L471) sets `state` to one of `:MIPNode`, `:MIPSol` or [`:MIPBranch`](https://github.com/JuliaOpt/CPLEX.jl/blob/e06d89bb5e3837b922e707ad07047f2c78e04672/src/CplexSolverInterface.jl#L600)  or [`:MIPIncumbent`](https://github.com/JuliaOpt/CPLEX.jl/blob/e06d89bb5e3837b922e707ad07047f2c78e04672/src/CplexSolverInterface.jl#L678)

Which is how I imagine MPB came to have the current situation of `:MIPNode`, `:MIPSol` or `:Other`.

The PR at https://github.com/JuliaOpt/Gurobi.jl/pull/59 is to make the state information available in the info callback so that you can extract info at each MIPSol. In the process, `:MIPInfo` has been changed to `:Other` (for `Gurobi.jl`) in accordance with [MPB's guidelines](https://mathprogbasejl.readthedocs.io/en/latest/lpqcqp.html#cbgetstate).

I don't pretend this fixes the situation for the other solvers, or that all other state information should be coerced into `:Other`. But that requires a larger investigation into the different kinds of states and callbacks to support at MPB-level, which is beyond the scope of this PR.

mlubin:
This PR together with JuliaOpt/Gurobi.jl#59 breaks user code using Gurobi with info callbacks, I assume. If we're going to do this, it should be an obvious improvement and not unnecessarily introduce confusing things like ``:Other``. So it's certainly within the scope of this PR to make info callbacks more coherent in general, otherwise we'll have to break user code again in the future when this happens.

mlubin:
I'm ok with removing the ``@assert state == :MIPInfo`` if that helps you out, but not with the changes to the tests.

yeesian:
To summarise the issues here:

1. Opening info callbacks to fire on all states (instead of an unstated/limited subset; see point 2) changes their current behavior (as you've intuited/pointed-out). This might be a breaking change (if users relied on implicit and unstated assumptions of where [the infocallback function](https://mathprogbasejl.readthedocs.io/en/latest/lpqcqp.html#setinfocallback!) is supposed to fire).
2. Currently, info callback state info is munged into `:Other` (so it only fires when gurobi is in state `CB_MIP`. If we remove `@assert state == :MIPInfo`, the tests will need to be changed [like in this PR] to preserve meaning.)

---

We can proceed in a number of ways. Two possibilities that comes to mind:

- make it the responsibility of the user to check which state the info callback is in (to determine which methods are available) the way it is done in the info callbacks test in this PR. This ought to have been the case for defensive users in the current setup (see point 1 above).
- have an additional argument to the info callback function, e.g.  `infocallback(m, mycb, [:MIPSol])` to specify the states in which the info callback function will fire. That way, the current `infocallback(m, mycb)` can be aliased to `infocallback(m, mycb, [:Other])` without breaking anything.

---

Some other auxiliary comments:

* Currently `Gurobi.jl` breaks on `cbgetobj()` in `state == :MIPSol` (which will get resolved in time, I presume), which caused point 1 (above) to become visible in the tests.
* Perhaps there is scope in MPB for the three states `:MIPNode`, `:MIPSol`, and `:Other` the way it is done now. One possibility is to introduce a secondary state that is solver specific, so that all the other info isn't munged into `:Other`. This can be introduced in a lazy/iterative way in subsequent PRs.

mlubin:
Could we start out by thinking about what people might want to do with info callbacks and how we can make it possible and easy to do so? Checking ``state == :Other`` before calling accessor functions is just  terribly unintuitive and doesn't necessarily make sense for all solvers.

yeesian:
> Could we start out by thinking about what people might want to do with info callbacks and how we can make it possible and easy to do so?

I'll personally prefer start with a couple of documented examples of "here's how you can do X (with info callbacks)" that interested users can copy+paste. And maybe generalise only after we elicit a couple of use-cases, rather than upfront trying to figure it all out.

The anxiety over `if MathProgBase.cbgetstate(cb) == :Other` should be directed towards

- the [issue with `cbgetobj()` in Gurobi.jl](https://github.com/JuliaOpt/Gurobi.jl/blob/d6ce6f827acde1d5dc8d9002f81189ed20fbcb3a/src/GurobiSolverInterface.jl#L411-L416), and
- the lack of precedence in documented behavior for callback states beyond `MIPSol` and `MIPNode`.

and not with this PR (to permit info callbacks to fire in other states of the B&B tree).

---

> Checking state == :Other before calling accessor functions is just terribly unintuitive and doesn't necessarily make sense for all solvers.

Having a generic fallback `:Other` would allow for users to have more customized callbacks (depending on their knowledge of their particular solver; and users can choose to ignore it if they prefer). I happen to see it as a net plus, rather than a minus, in terms of usability.

The check of `MathProgBase.cbgetstate(cb) == :Other` starts to feel less "special"/discomfiting when there are infocallbacks that keep track of different things at different states. For example, one thing I have found myself needing (that the current setup of infocallbacks don't allow for) is keeping track of the intermediate solutions in the B&B tree.

Here's how that can be done (with this PR):

```julia
function myinfo(cb)
    entered[1] = true
    if MathProgBase.cbgetstate(cb) == :Other
        push!(nodes,      MathProgBase.cbgetexplorednodes(cb))
        push!(objs,       MathProgBase.cbgetobj(cb))
        push!(bestbounds, MathProgBase.cbgetbestbound(cb))
    end
    if MathProgBase.cbgetstate(cb) == :MIPSol
        # Update all variable values so that `getvalue` works
        MathProgBase.cbgetmipsolution(cb, mod.colVal)
        println(JuMP.getvalue(x[1:15])) # print just 15 values to not overwhelm the example screen
    end
end
```
Giving something like
```julia
[callback] Test informational callback
  > With solver Gurobi.GurobiSolver
[1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,0.0,1.0,0.0,0.0,0.0,0.0,0.0]
[0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0]
[-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0]
[-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0]
[-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0,-0.0]
5 facts verified.
```
Alternatively, I can store them all into an array, rather than printing them out. And again, I'll like to reiterate that the need to check for `:Other` is due to the [issue with `cbgetobj()` in Gurobi.jl](https://github.com/JuliaOpt/Gurobi.jl/blob/d6ce6f827acde1d5dc8d9002f81189ed20fbcb3a/src/GurobiSolverInterface.jl#L411-L416), and is otherwise not required.

mlubin:
How about being clear about when we want the info callback to be called? Something like
```jl
addinfocallback(mod, myinfo, when = :Feasible)   # MIPSOL in Gurobi
addinfocallback(mod, myinfo, when = :Fractional) # MIPNODE
addinfocallback(mod, myinfo, when = :BnBNode)    # MIP
addinfocallback(mod, myinfo, when = :Iteration)  # For barrier/simplex iterations, etc..
```

For ``:Feasible`` and ``:Fractional``, JuMP can load the solution into the variables as it does for the lazy and cut callbacks.

yeesian:
This is all pretty new to me, so I'm not entirely certain about the choice of `when` clauses in your proposal above. But sure, I'll be okay with that.

yeesian:
Although I imagine it'll be a fairly advanced/motivated user who'll go far enough to want to dig into
```julia
addinfocallback(mod, myinfo, when = :Feasible)   # MIPSOL in Gurobi
addinfocallback(mod, myinfo, when = :Fractional) # MIPNODE
addinfocallback(mod, myinfo, when = :BnBNode)    # MIP
addinfocallback(mod, myinfo, when = :Iteration)  # For barrier/simplex iterations, etc..
```
and said user will be quite frustrated if it is hamstrung to a pre-selected set of options, that artificially hamstrung them from what they'll otherwise be able to do.

My primary consideration will be to allow for solver-specific state/information, before we settle on a more general set of solver independent descriptions. And I don't yet see that with the proposal above, unless we are to make the `when` clauses solver-specific (like we do with solver-creation options).

yeesian:
Thinking about it more, I remain unconvinced that 
```julia
addinfocallback(mod, myinfo1, when = :Feasible)   # MIPSOL in Gurobi
addinfocallback(mod, myinfo2, when = :Fractional) # MIPNODE
addinfocallback(mod, myinfo3, when = :BnBNode)    # MIP
addinfocallback(mod, myinfo4, when = :Iteration)  # For barrier/simplex iterations, etc..
```
is any better than
```julia
"when::Symbol can be any of :Feasible, :Fractional, :BnBNode, or :Iteration"
function myinfo(cb, when::Symbol)
    when == :Feasible && myinfo1(cb)
    when == :Fractional && myinfo2(cb)
    when == :BnBNode && myinfo3(cb)
    when == :Iteration && myinfo4(cb)
end    

addinfocallback(mod, myinfo)
```
Both require the user to think about when the infocallback should be fired, and the latter results in a lot less work for the master callback function to keep track of all possible infocallback functions (`myinfo1` to `myinfo4`) and dispatch accordingly.

src/callbacks.jl
```diff
@@ -38,9 +39,9 @@ function addheuristiccallback(m::Model, f::Function)
     m.internalModelLoaded = false
     push!(m.callbacks, HeuristicCallback(f))
 end
-function addinfocallback(m::Model, f::Function)
+function addinfocallback(m::Model, f::Function, when::Symbol)
```
mlubin: Keyword argument?

yeesian:
This PR is ready for review, after https://github.com/JuliaOpt/GLPKMathProgInterface.jl/pull/32 is merged.

joehuchette:
I think I'm alright with this change, but throwing the error in ``addinfocallback`` if the ``where`` kwarg is omitted seems too breaking. Can we make it a deprecation warning instead?

yeesian:
Deprecation warning added. Now just pending the changes in 

- [ ] https://github.com/JuliaOpt/MathProgBase.jl/pull/115
- [ ] https://github.com/JuliaOpt/Gurobi.jl/pull/59
- [ ] https://github.com/JuliaOpt/CPLEX.jl/pull/80
- [x] https://github.com/JuliaOpt/GLPKMathProgInterface.jl/pull/32

src/callbacks.jl
```diff
@@ -41,7 +41,15 @@ function addheuristiccallback(m::Model, f::Function)
 end
 function addinfocallback(m::Model, f::Function; when::Symbol = :Undefined)
     if when == :Undefined
-        error("Info Callbacks without a `when` clause are unsupported.")
+        when = :Intermediate
```
joehuchette: Doesn't really matter, but I think you could do something like:
```jl
function _not_specified()
    warn("""Info Callbacks without a when clause will currently default
              to fire only in the :Intermediate state to preserve its behavior
              with previous versions of JuMP.
 
             This behavior will be deprecated in subsequent versions of JuMP, so
             please rewrite all invocations of addinfocallbacks(m, f) to
             addinfocallbacks(m, f, when = :Intermediate) instead.
             """)
    :Intermediate
end

function addinfocallback(m::Model, f::Function; when::Symbol = _not_specified())
...
```
to avoid the corner case where I might pass in ``when=:Undefined``.

joehuchette:
What's the status here? Any objections @mlubin? If we're good to go, would you mind squashing @yeesian?

mlubin:
No objections, just note that we have to tag GLPKMathProgInterface CPLEX, Gurobi, and JuMP all at the same time. JuMP release is currently broken against CPLEX master because of this change.

joehuchette:
Is this breaking enough that we can't backport to the release branch?

mlubin:
It's fine to tag master+this as 0.14.1.

joehuchette:
Restarted Travis, will merge if passing.

mlubin:
If this passes before GLPK is tagged then something is broken

chriscoey:
This would be really useful for Pajarito. Can I help? Would be very nice to have this feature available ASAP.

joehuchette:
I can tag everything tomorrow

PR #820, opened 2016-08-05
tidy up expression graphs

mlubin:
For https://github.com/JuliaOpt/AmplNLWriter.jl/issues/39
CC @JackDunnNZ

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -1018,6 +1018,9 @@ end
 
 # currently don't merge duplicates (this isn't required by MPB standard)
 function affToExpr(aff::AffExpr, constant::Bool)
+    if length(aff.vars) == 0 && !constant
+        return 0
+    end
     ex = Expr(:call,:+)
     for k in 1:length(aff.vars)
         push!(ex.args, Expr(:call,:*,aff.coeffs[k],:(x[$(aff.vars[k].col)])))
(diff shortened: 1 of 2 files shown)


mlubin:
@JackDunnNZ, my example [here](https://github.com/JuliaOpt/AmplNLWriter.jl/issues/39#issuecomment-237339527) still gives an error:
```
ERROR: MethodError: `process_expression!` has no method matching process_expression!(::Int64, ::Dict{Int64,Float64}, ::Array{Symbol,1})
```

and your example [here](https://github.com/JuliaOpt/AmplNLWriter.jl/issues/39#issuecomment-226366919) now seems to crash ipopt.

JackDunnNZ:
Ipopt seems to die if there's no objective, changing the example to:

```julia
using JuMP
using AmplNLWriter

m = Model(solver=IpoptNLSolver())
@variable(m, x)
@NLconstraint(m, x <= sum{0, j = []})
@objective(m, Max, x)
solve(m)
```

fixes things there. Just looking at the other one now

JackDunnNZ:
The error with the Convex example comes because the constraint `x^2 <= 5` gets transformed into a set of constraints that includes `0 <= 1.0`, giving:

```
ERROR: MethodError: `process_expression!` has no method matching process_expression!(::Int64, ::Dict{Int64,Float64}, ::Array{Symbol,1})
Closest candidates are:
  process_expression!(::Expr, ::Dict{Int64,Float64}, ::Array{Symbol,1})
```

because `0` isn't an `Expr`. If I add the definition to process this special case:

```julia
function process_expression!(nonlin_expr::Real, lin_expr, varlinearities)
    0, nonlin_expr, :Lin
end
```
the .nl file becomes malformed because there's a constraint with no variables involved, and no solver will open it. 

@mlubin Why is it giving this empty constraint? And where should we head it off and delete it?

mlubin:
``0 <= 1.0`` is a valid constraint

JackDunnNZ:
Yes, but not in an .nl file. I'm happy to deal with it, just wanted to make
sure you didn't want to handle it further up

mlubin:
I think AmplNLWriter should be cleaning this one up. JuMP isn't in the business of preprocessing out constraints.

JackDunnNZ:
Will do

JackDunnNZ:
Great, thanks for this too!

PR #825, opened 2016-08-06
Add support for non-constant exponents in LPQP problems

### LINKED ISSUE #822: Evaluation of an Exponent Expression
```
using JuMP

m = Model()

@variable(m, 1 <= x <= 5)

costs = [10, 1, 0.1]

# gives: ERROR: LoadError: Only exponents of 2 are currently supported. 
# Are you trying to build a nonlinear problem? Make sure you use @NLconstraint/@NLobjective.
@objective(m, Min, sum{ c*x^(p-1), (p,c) in enumerate(costs) } )

solve(m)
```

The issue does not occur when using @NLobjective.

joehuchette:
Close #822

### DIFF
--- a/src/parseExpr_staged.jl
+++ b/src/parseExpr_staged.jl
@@ -656,13 +656,25 @@ function parseExpr(x, aff::Symbol, lcoeffs::Vector, rcoeffs::Vector, newaff::Sym
                 return newaff, blk
             end
         elseif x.head == :call && x.args[1] == :^ && is_complex_expr(x.args[2])
-            x.args[3] == 2 || error("Only exponents of 2 are currently supported")
-            blk = Expr(:block)
-            s = gensym()
-            newaff_, parsed = parseExprToplevel(x.args[2], s)
-            push!(blk.args, :($s = 0.0; $parsed))
-            push!(blk.args, :($newaff = $aff + $(Expr(:call,:*,lcoeffs...,newaff_,newaff_,rcoeffs...))))
-            return newaff, blk
+            if x.args[3] == 2
+                blk = Expr(:block)
+                s = gensym()
+                newaff_, parsed = parseExprToplevel(x.args[2], s)
+                push!(blk.args, :($s = 0.0; $parsed))
+                push!(blk.args, :($newaff = $aff + $(Expr(:call,:*,lcoeffs...,newaff_,newaff_,rcoeffs...))))
+                return newaff, blk
+            elseif x.args[3] == 1
+                return parseExpr(x.args[2], aff, lcoeffs, rcoeffs)
+            elseif x.args[3] == 0
+                return parseExpr(1, aff, lcoeffs, rcoeffs)
+            else
+                blk = Expr(:block)
+                s = gensym()
+                newaff_, parsed = parseExprToplevel(x.args[2], s)
+                push!(blk.args, :($s = 0.0; $parsed))
+                push!(blk.args, :($newaff = $aff + $(Expr(:call,:*,lcoeffs...,Expr(:call,:^,newaff_,esc(x.args[3])),rcoeffs...))))
+                return newaff, blk
+            end
         elseif x.head == :call && x.args[1] == :/
             @assert length(x.args) == 3
             numerator = x.args[2]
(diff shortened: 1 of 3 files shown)


src/parseExpr_staged.jl
```diff
@@ -656,7 +656,6 @@ function parseExpr(x, aff::Symbol, lcoeffs::Vector, rcoeffs::Vector, newaff::Sym
                 return newaff, blk
             end
         elseif x.head == :call && x.args[1] == :^ && is_complex_expr(x.args[2])
-            x.args[3] == 2 || error("Only exponents of 2 are currently supported")
```
mlubin: It's incorrect to remove this. The code below assumes the power is two.

joehuchette: Ah, true.

src/parseExpr_staged.jl
```diff
@@ -655,13 +655,22 @@ function parseExpr(x, aff::Symbol, lcoeffs::Vector, rcoeffs::Vector, newaff::Sym
...
             push!(blk.args, :($newaff = $aff + $(Expr(:call,:*,lcoeffs...,newaff_,newaff_,rcoeffs...))))
             return newaff, blk
+        elseif x.head == :call && x.args[1] == :^ && is_complex_expr(x.args[2]) && x.args[3] == 1
+            blk = Expr(:block)
+            s = gensym()
+            newaff_, parsed = parseExprToplevel(x.args[2], s)
+            push!(blk.args, :($s = 0.0; $parsed))
+            push!(blk.args, :($newaff = $aff + $(Expr(:call,:*,lcoeffs...,newaff_,rcoeffs...))))
+            return newaff, blk
+        elseif x.head == :call && x.args[1] == :^ && is_complex_expr(x.args[2]) && x.args[3] == 0
```
mlubin: This misses the case of a non-constant exponent and complex subexpression. I think the following code will crash: ``sum(x[i] for i in 1:10)^n``

mlubin: But it never worked before, so it's still ok to move ahead with this.

joehuchette: The error message isn't very good, though:
```
julia> @constraint(m, sum{x, i in 1:3}^(foo()) == 1)
ERROR: UndefVarError: i not defined
```

mlubin: Yep. Lazy feature...

src/parseExpr_staged.jl
```diff
@@ -655,14 +655,22 @@ function parseExpr(x, aff::Symbol, lcoeffs::Vector, rcoeffs::Vector, newaff::Sym
...
-        elseif x.head == :call && x.args[1] == :^ && is_complex_expr(x.args[2])
-            x.args[3] == 2 || error("Only exponents of 2 are currently supported")
+        elseif x.head == :call && x.args[1] == :^ && is_complex_expr(x.args[2]) && x.args[3] == 2
             blk = Expr(:block)
             s = gensym()
             newaff_, parsed = parseExprToplevel(x.args[2], s)
             push!(blk.args, :($s = 0.0; $parsed))
             push!(blk.args, :($newaff = $aff + $(Expr(:call,:*,lcoeffs...,newaff_,newaff_,rcoeffs...))))
             return newaff, blk
+        elseif x.head == :call && x.args[1] == :^ && is_complex_expr(x.args[2]) && x.args[3] == 1
```
mlubin: How about just ``return parseExpr(x.args[2], ...)`` here?

src/parseExpr_staged.jl
mlubin: Also this block can be ``return parseExpr(1, ...)``

src/parseExpr_staged.jl
```diff
@@ -656,13 +656,25 @@ function parseExpr(x, aff::Symbol, lcoeffs::Vector, rcoeffs::Vector, newaff::Sym
...
+            elseif x.args[3] == 1
+                return parseExpr(x.args[2], aff, lcoeffs, rcoeffs)
+            elseif x.args[3] == 0
+                return parseExpr(1, aff, lcoeffs, rcoeffs)
+            else
+                blk = Expr(:block)
+                s = gensym()
+                newaff_, parsed = parseExprToplevel(x.args[2], s)
+                push!(blk.args, :($s = 0.0; $parsed))
+                push!(blk.args, :($newaff = $aff + $(Expr(:call,:*,lcoeffs...,Expr(:call,:^,newaff_,esc(x.args[3])),rcoeffs...))))
```
mlubin: Good fix

mlubin:
LGTM

PR #834, opened 2016-08-20
Fix #830

### LINKED ISSUE #830: Internal Model not unloaded when SDP constraint added
In lines [468 -> 470 of macros.jl](https://github.com/JuliaOpt/JuMP.jl/blob/8b3074af0c51cdb12c9411e9a71175614bfe2866/src/macros.jl#L468-L470) (in macro `SDconstraint`), it is written

        c = SDConstraint($newaff)
        push!($(m).sdpconstr, c)
        c

while I would have expected something like

        addconstraint($m, SDConstraint($newaff))

since this is what is done, e.g., in the macro `constraint`. Indeed, those two blocks of code do not have the same behaviors as the second one also does `m.internalModelLoaded = false` inside the `addconstraint` function which seems important since one constraint has been added. Am I getting it wrong ?

blegat:
Fixes #830 and adds a test.

I couldn't test that the model was not resolved correctly since in `solvers.jl`, there is:

        # Obtain a fresh MPB model for the solver
        # If the problem is conic, we rebuild the problem from
        # scratch every time
        m.internalModel = MathProgBase.ConicModel(m.solver)

so the fact that `m.internalModelLoaded` is not set to false when adding a new SDconstraint has no effect. Therefore  in the tests I just test that `internalModelLoaded` is set to false;

An important effect of the fixed `@SDconstraint` though is that it returns now a `JuMP.ConstraintRef{JuMP.Model,JuMP.SDConstraint}` instead of a `JuMP.SDConstraint`. This is also tested in the test.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -465,9 +465,7 @@ macro SDconstraint(m, x)
     assert_validmodel(m, quote
         q = zero(AffExpr)
         $parsecode
-        c = SDConstraint($newaff)
-        push!($(m).sdpconstr, c)
-        c
+        addconstraint($m, SDConstraint($newaff))
     end)
 end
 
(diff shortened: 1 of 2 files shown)


joehuchette:
Thanks for the find and for the fix!

PR #835, opened 2016-08-21
Make it possible to write macros in extension

blegat:
I am writing a [JuMP extension for Sum of Squares Programming](https://github.com/blegat/SumOfSquares.jl).
All the JuMP specific code is in [the jump.jl file](https://github.com/blegat/SumOfSquares.jl/blob/master/src/jump.jl).
It adds the macros `@SOSconstraint` and `@SOSvariable` that create SOS constraint and SOS variable respectively.

They make use of the `assert_validmodel` and `parseExpr`functions.
However, since it is not called from inside the `JuMP` module, julia complains that it does not know `validmodel` and `addtoexpr_reorder`.
Adding the `JuMP.` prefix as in this pull request fixes the problem and it can also be useful for other extensions.
Of course, if you do not like this change, another solution would be to move [the jump.jl file](https://github.com/blegat/SumOfSquares.jl/blob/master/src/jump.jl) code inside the `JuMP` package.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -199,7 +199,7 @@ validmodel(m, name) = error("Expected $name to be a JuMP model, but it has type 
 function assert_validmodel(m, macrocode)
     # assumes m is already escaped
     quote
-        validmodel($m, $(quot(m.args[1])))
+        JuMP.validmodel($m, $(quot(m.args[1])))
         $macrocode
     end
 end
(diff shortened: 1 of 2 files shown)


IainNZ:
This might be a good idea, but, for the record, doing the following does work for extensions defining their own macros: https://github.com/IainNZ/JuMPeR.jl/blob/master/src/robustmacro.jl#L18-L21

blegat:
Thanks @IainNZ , this is exactly what I need :)
In view of your simple solution do not think that merging this pull request would have any benefit.

IainNZ:
It might still be worth a rethink of how this works - always nice to make it easier to create extensions.

mlubin:
@blegat, looking forward to the extension!
We will certainly take suggestions on how to make extensions better. Right now there's no documentation and we make no guarantee that we won't break things across JuMP releases. That should change at some point.

Issue #646, opened 2016-01-16
keys(::JuMPDict) returns an iterator containing tuples


zsunberg:
Is it the correct behavior for `keys(::JuMPDict)` to return a `KeyIterator` for a dictionary with tuples as keys? This seems strange when the keys for the `JumpDict` are not tuples.

In particular, if I have a variable `u` and a `JumpDict` from solving another problem stored in `initial.u`, I can't do the following

```
for key in keys(u)
      setValue(u[key], initial.u[key])
end
```

mlubin:
Well, there has to be an inconsistency somewhere because if a JuMPDict is indexed over multiple dimensions then the only thing that makes sense to return is a tuple. You can get the code working by writing:
```
for key in keys(u)
      setValue(u[key...], initial.u[key...])
end
```

zsunberg:
Hmm... yeah I see what you mean. It still seems a little weird that they objects you get when iterating over keys() are not actually valid keys. Would it make sense to add the following additional version of getindex:
```
Base.getindex(d::JuMPDict, t::Tuple) = d.tupledict[t]
```
? (This has the side effect of making `myjumpdict[1,2]` equivalent to `myjumpdict[(1,2)]`). If you think that's an ok idea I can submit a pull request.

mlubin:
I think that would break the case when the index set is a set of tuples. The internal key would be ``(t,)`` (where ``t`` is the element of the index set), but that method would try to look up ``t``.

zsunberg:
Would it work if the index set *only* contains tuples? i.e. would it break only if the index set contains both tuple and non tuple items? (I didn't dig deep enough to answer this question myself.) If that is the case, are there even any foreseeable cases when the index set would contain both tuples and non-tuples?

I don't see a great way to fix this without changing the internal representation, and I don't have any good alternative internal representations to suggest. Feel free to close this issue if you don't plan on fixing it.

mlubin:
> If that is the case, are there even any foreseeable cases when the index set would contain both tuples and non-tuples?

It's foreseeable that if we break this, someone will complain :). We're open to further discussion and PRs, but progress here likely won't come from me since it's pretty confusing for me to wrap my head around all of the corner cases.

mlubin:
Maybe the solution here is to use something like ``JuMP.keys`` so it's clear that this doesn't behave as you would expect from the base method?

joehuchette:
@zsunberg I think the case we would need to worry about is something like
[...]

PR #836, opened 2016-08-21
Fixing KeyIterator used for JuMPArray to work also when index sets are non indexable.

IssamT:
This just a beginning to solving issues raised #646 

I think it is better to not merge it yet. Maybe just give some feedback especially I m pretty sure the iterator methods could be implemented better.

The new `KeyIterator` works fine as long as the **index sets** are iterable (including Iterators and `Set`s). They **do not need to be indexable anymore**

### DIFF
--- a/src/JuMPArray.jl
+++ b/src/JuMPArray.jl
@@ -35,12 +35,25 @@ end
 end
 
 Base.getindex(d::JuMPArray, ::Colon) = d.innerArray[:]
+
+immutable JuMPKey
+    jk::Tuple
+end
+
+Base.start(x::JuMPKey) = Base.start(x.jk)
+Base.next(x::JuMPKey, i) = Base.next(x.jk, i)
+Base.done(x::JuMPKey, i) = Base.done(x.jk, i)
+Base.length(x::JuMPKey) = Base.length(x.jk)
+
 @generated function Base.getindex{T,N,NT<:NTuple}(d::JuMPArray{T,N,NT}, idx...)
     if N != length(idx)
         error("Indexed into a JuMPArray with $(length(idx)) indices (expected $N indices)")
     end
     Expr(:call, :getindex, :(d.innerArray), _to_cartesian(d,NT,idx)...)
 end
+
+Base.getindex(d::JuMPArray, x::JuMPKey) = Base.getindex(d, x.jk...)
+
 
 @generated function Base.setindex!{T,N,NT<:NTuple}(d::JuMPArray{T,N,NT}, v::T, idx...)
     if N != length(idx)
--- a/src/JuMPArray.jl
+++ b/src/JuMPArray.jl
@@ -36,8 +36,8 @@ end
 
 Base.getindex(d::JuMPArray, ::Colon) = d.innerArray[:]
 
-immutable JuMPKey
-    jk::Tuple
+immutable JuMPKey{T<:Tuple}
+    jk::T
 end
 
 Base.start(x::JuMPKey) = Base.start(x.jk)
@@ -53,7 +53,6 @@ Base.length(x::JuMPKey) = Base.length(x.jk)
 end
 
 Base.getindex(d::JuMPArray, x::JuMPKey) = Base.getindex(d, x.jk...)
-
 
 @generated function Base.setindex!{T,N,NT<:NTuple}(d::JuMPArray{T,N,NT}, v::T, idx...)
     if N != length(idx)
--- a/src/JuMPArray.jl
+++ b/src/JuMPArray.jl
@@ -36,23 +36,12 @@ end
 
 Base.getindex(d::JuMPArray, ::Colon) = d.innerArray[:]
 
-immutable JuMPKey{T<:Tuple}
-    jk::T
-end
-
-Base.start(x::JuMPKey) = Base.start(x.jk)
-Base.next(x::JuMPKey, i) = Base.next(x.jk, i)
-Base.done(x::JuMPKey, i) = Base.done(x.jk, i)
-Base.length(x::JuMPKey) = Base.length(x.jk)
-
 @generated function Base.getindex{T,N,NT<:NTuple}(d::JuMPArray{T,N,NT}, idx...)
     if N != length(idx)
         error("Indexed into a JuMPArray with $(length(idx)) indices (expected $N indices)")
     end
     Expr(:call, :getindex, :(d.innerArray), _to_cartesian(d,NT,idx)...)
 end
-
-Base.getindex(d::JuMPArray, x::JuMPKey) = Base.getindex(d, x.jk...)
 
 @generated function Base.setindex!{T,N,NT<:NTuple}(d::JuMPArray{T,N,NT}, v::T, idx...)
     if N != length(idx)
--- a/src/JuMPContainer.jl
+++ b/src/JuMPContainer.jl
@@ -186,57 +186,56 @@ Base.length(it::ValueIterator)  = length(it.x)
 type KeyIterator{JA<:JuMPArray}
     x::JA
     dim::Int
-    state::Tuple
-    done::Bool
+    next_k_cache::Array{Any,1}
     function KeyIterator(d)
         n = ndims(d.innerArray)
-        new(d, n, (), false)
+        new(d, n, Array(Any, n+1))
     end
 end
 
 KeyIterator{JA}(d::JA) = KeyIterator{JA}(d)
 
-function Base.start(it::KeyIterator)
-    it.state = tuple()
-    for i in 1:it.dim
-        it.state = tuple(it.state..., start(it.x.indexsets[i]))
+@generated function _start{T,N,NT}(x::JuMPArray{T,N,NT})
+    quote
+        $(Expr(:tuple, 0, [:(start(x.indexsets[$i])) for i in 1:N]...))
     end
-    it.state
+end
+
+Base.start(it::KeyIterator) = _start(it.x)
+
+@generated function _next{T,N,NT}(x::JuMPArray{T,N,NT}, k)
+    quote
+        $(Expr(:tuple, [:(next(x.indexsets[$i], k[$i+1])[1]) for i in 1:N]...))
+    end
 end
 
 function Base.next(it::KeyIterator, k)
-    cartesian_key = ()
-    for i in 1:it.dim
-        cartesian_key = tuple(cartesian_key..., next(it.x.indexsets[i], k[i])[1])
-    end
-    #@show cartesian_key
-    next_k = ()
+    cartesian_key = _next(it.x, k)
     pos = -1
     for i in 1:it.dim
-        if(!done(it.x.indexsets[i], next(it.x.indexsets[i], k[i])[2] ) )
+        if(!done(it.x.indexsets[i], next(it.x.indexsets[i], k[i+1])[2] ) )
             pos = i
             break
         end
     end
-    #@show pos
-    if pos == -1
-        it.done = true
-        return cartesian_key, nothing
+    if pos == - 1
+        return cartesian_key, (1)
     end
+    it.next_k_cache[1] = 0
     for i in 1:it.dim
         if i < pos
-            next_k = tuple(next_k..., start(it.x.indexsets[i]) )
+            it.next_k_cache[i+1] = start(it.x.indexsets[i])
         elseif i == pos
-            next_k = tuple(next_k..., next(it.x.indexsets[i], k[i])[2])
+            it.next_k_cache[i+1] = next(it.x.indexsets[i], k[i+1])[2]
         else
-            next_k = tuple(next_k..., k[i])
+            it.next_k_cache[i+1] = k[i+1]
         end
     end
-    cartesian_key, next_k
+    cartesian_key, tuple(it.next_k_cache...)
 end
 
 function Base.done(it::KeyIterator, k)
-    return it.done
+    return k[1] == 1
 end
 
 Base.length(it::KeyIterator)  = length(it.x.innerArray)
--- a/src/JuMPContainer.jl
+++ b/src/JuMPContainer.jl
@@ -189,54 +189,54 @@ type KeyIterator{JA<:JuMPArray}
     state::Tuple
     done::Bool
     function KeyIterator(d)
-      t = tuple()
-      n = ndims(d.innerArray)
-      for i in 1:n
-        t = tuple(t..., start(d.indexsets[i]))
-      end
-      new(d, n, t, false)
+        t = tuple()
+        n = ndims(d.innerArray)
+        for i in 1:n
+            t = tuple(t..., start(d.indexsets[i]))
+        end
+        new(d, n, t, false)
     end
 end
 
 KeyIterator{JA}(d::JA) = KeyIterator{JA}(d)
 
 function Base.start(it::KeyIterator)
-  it.state
+    it.state
 end
 
 function Base.next(it::KeyIterator, k)
-  cartesian_key = ()
-  for i in 1:it.dim
-    cartesian_key = tuple(cartesian_key..., next(it.x.indexsets[i], k[i])[1])
-  end
-  #@show cartesian_key
-  next_k = ()
-  pos = -1
-  for i in 1:it.dim
-    if(!done(it.x.indexsets[i], next(it.x.indexsets[i], k[i])[2] ) )
-      pos = i
-      break
+    cartesian_key = ()
+    for i in 1:it.dim
+        cartesian_key = tuple(cartesian_key..., next(it.x.indexsets[i], k[i])[1])
     end
-  end
-  #@show pos
-  if pos == -1
-    it.done = true
-    return cartesian_key, nothing
-  end
-  for i in 1:it.dim
-    if i < pos
-      next_k = tuple(next_k..., start(it.x.indexsets[i]) )
-    elseif i == pos
-      next_k = tuple(next_k..., next(it.x.indexsets[i], k[i])[2])
-    else
-      next_k = tuple(next_k..., k[i])
+    #@show cartesian_key
+    next_k = ()
+    pos = -1
+    for i in 1:it.dim
+        if(!done(it.x.indexsets[i], next(it.x.indexsets[i], k[i])[2] ) )
+            pos = i
+            break
+        end
     end
-  end
-  cartesian_key, next_k
+    #@show pos
+    if pos == -1
+        it.done = true
+        return cartesian_key, nothing
+    end
+    for i in 1:it.dim
+        if i < pos
+            next_k = tuple(next_k..., start(it.x.indexsets[i]) )
+        elseif i == pos
+            next_k = tuple(next_k..., next(it.x.indexsets[i], k[i])[2])
+        else
+            next_k = tuple(next_k..., k[i])
+        end
+    end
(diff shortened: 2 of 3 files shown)


src/JuMPContainer.jl
```diff
@@ -183,13 +185,58 @@ Base.length(it::ValueIterator)  = length(it.x)
...
+        t = tuple(t..., start(d.indexsets[i]))
+      end
+      new(d, n, t, false)
+    end
+end
+
+KeyIterator{JA}(d::JA) = KeyIterator{JA}(d)
+
+function Base.start(it::KeyIterator)
+  it.state
```
mlubin: Please follow the 4-space convention used in JuMP

mlubin:
I haven't dug into the code but it seems reasonable. @joehuchette, do we use iteration in any performance-critical places?

IssamT:
An alternative solution would be to keep the pervious iterator and apply `collect()` around the index sets provided by the user to `@variable` macro when they are not of type `Array`. But this will use more memory (for the additional storage of the Arrays that are duplicate to the initial index sets)

IssamT:
I would like to do some benchmarks to compare (the current KeyIterator, the one I posted and the improvement I am trying using @generated functions that I finally had the time read about in the manual. 

**My question** is are we interested more by the performance with Julia 0.4 or the dev version 0.5?

joehuchette:
We shouldn't use iteration in performance-critical code, and if we do, we probably shouldn't. I would be interested to see the performance here; based on my reading, it doesn't look like like the new implementation for ``next`` is type-stable (though maybe that doesn't matter).

>> My question is are we interested more by the performance with Julia 0.4 or the dev version 0.5?
I'd focus on 0.5, since the release is imminent, and we won't have a new JuMP release targeted at only 0.4.

IssamT:
> The new KeyIterator works fine as long as the index sets are iterable (including Iterators and Sets). They do not need to be indexable anymore

This was not completely true. it works fine only for iterables that implement the iteration interface in a "clean" way. for more details see http://stackoverflow.com/questions/39181313/julia-iteration-start-next-done-side-effects/

Edge Iterator in LightGraphs for instance did not have a "clean" implementation but we fixed it in https://github.com/JuliaGraphs/LightGraphs.jl/pull/423

In the last version I have just committed:
- there are less allocation thanks to caching and @generated functions
- the functions `start` `next` `done` do not modify arguments (except the cache...)

On this small example:
```

using MathProgBase
using JuMP

set = 2:101
set2 = 2001:2100

m = Model()

@variable(m, x[set, set2], Bin)
@objective(m , Max, sum{sum{x[e,p], e in set}, p in set2})
solve(m)
sol = getvalue(x)

for i in 1:20
    @time begin
        for i in keys(sol)
        end
    end
end
```
  
The results with the old version of iterators that only worked for indexable collections

 ```
 0.067887 seconds (56.75 k allocations: 1.056 MB)
  0.001908 seconds (29.49 k allocations: 347.586 KB)
  0.001749 seconds (29.49 k allocations: 347.586 KB)
  0.001712 seconds (29.49 k allocations: 347.586 KB)
  0.002250 seconds (29.49 k allocations: 347.586 KB)
  0.002008 seconds (29.49 k allocations: 347.586 KB)
  0.001950 seconds (29.49 k allocations: 347.586 KB)
  0.001922 seconds (29.49 k allocations: 347.586 KB)
  0.001932 seconds (29.49 k allocations: 347.586 KB)
  0.001864 seconds (29.49 k allocations: 347.586 KB)
  0.002175 seconds (29.49 k allocations: 347.586 KB)
  0.001853 seconds (29.49 k allocations: 347.586 KB)
  0.001862 seconds (29.49 k allocations: 347.586 KB)
  0.001881 seconds (29.49 k allocations: 347.586 KB)
  0.001840 seconds (29.49 k allocations: 347.586 KB)
  0.002117 seconds (29.49 k allocations: 347.586 KB)
  0.001855 seconds (29.49 k allocations: 347.586 KB)
  0.001837 seconds (29.49 k allocations: 347.586 KB)
  0.001881 seconds (29.49 k allocations: 347.586 KB)
  0.001855 seconds (29.49 k allocations: 347.586 KB)
```

The results with the last version I submitted.

 ```
 0.147226 seconds (106.76 k allocations: 2.177 MB)
  0.009874 seconds (60.00 k allocations: 937.559 KB)
  0.009324 seconds (60.00 k allocations: 937.559 KB)
  0.009328 seconds (60.00 k allocations: 937.559 KB)
  0.008945 seconds (60.00 k allocations: 937.559 KB)
  0.009665 seconds (60.00 k allocations: 937.559 KB)
  0.008988 seconds (60.00 k allocations: 937.559 KB)
  0.008875 seconds (60.00 k allocations: 937.559 KB)
  0.009551 seconds (60.00 k allocations: 937.559 KB)
  0.008905 seconds (60.00 k allocations: 937.559 KB)
  0.009292 seconds (60.00 k allocations: 937.559 KB)
  0.009465 seconds (60.00 k allocations: 937.559 KB)
  0.009953 seconds (60.00 k allocations: 937.559 KB)
  0.008890 seconds (60.00 k allocations: 937.559 KB)
  0.009219 seconds (60.00 k allocations: 937.559 KB)
  0.009539 seconds (60.00 k allocations: 937.559 KB)
  0.009209 seconds (60.00 k allocations: 937.559 KB)
  0.009049 seconds (60.00 k allocations: 937.559 KB)
  0.010312 seconds (60.00 k allocations: 937.559 KB)
  0.008850 seconds (60.00 k allocations: 937.559 KB)
```

As you can see there are still more allocations unfortunately...

IssamT:
I have added the tupple wrapper `JuMPKey` as it was suggested in #646 

Now it is used for keys(::JuMPArray) 

This wrapper is immutable so it doesn t impact the performance and the test above gives almost same result:
```

  0.143752 seconds (106.61 k allocations: 2.143 MB)
  0.008796 seconds (60.00 k allocations: 898.504 KB)
  0.008244 seconds (60.00 k allocations: 898.504 KB)
  0.008289 seconds (60.00 k allocations: 898.504 KB)
  0.008173 seconds (60.00 k allocations: 898.504 KB)
  0.008945 seconds (60.00 k allocations: 898.504 KB)
  0.008194 seconds (60.00 k allocations: 898.504 KB)
  0.008156 seconds (60.00 k allocations: 898.504 KB)
  0.008159 seconds (60.00 k allocations: 898.504 KB)
  0.008860 seconds (60.00 k allocations: 898.504 KB)
  0.008102 seconds (60.00 k allocations: 898.504 KB)
  0.008212 seconds (60.00 k allocations: 898.504 KB)
  0.008869 seconds (60.00 k allocations: 898.504 KB)
  0.008222 seconds (60.00 k allocations: 898.504 KB)
  0.008131 seconds (60.00 k allocations: 898.504 KB)
  0.008160 seconds (60.00 k allocations: 898.504 KB)
  0.008885 seconds (60.00 k allocations: 898.504 KB)
  0.008156 seconds (60.00 k allocations: 898.504 KB)
  0.008147 seconds (60.00 k allocations: 898.504 KB)
  0.008772 seconds (60.00 k allocations: 898.504 KB)

```

Hence the following two loops are now equivalent for `JuMPArray`. I think it s better to keep both for compatibility with any previous code using the old way.

```
sol = getvalue(x)

for i in keys(sol)
    @show sol[i]
end

for i in keys(sol)
    @show sol[i...]
end
```

If you are ok with these modifications I have done on `JuMPArray` (Support of index sets that are non indexable, Support of "intuitive" use of key), I can apply the same on `JuMPDict`

IssamT:
@joehuchette, @mlubin, if you prefer to keep the old good performance for indexable collections we can specialize the iteration process. But I'm not sure if we can check efficiently that a type has the function getindex()

http://stackoverflow.com/questions/39410299/check-if-a-type-implements-an-interface-in-julia

Any ideas?

src/JuMPArray.jl
```diff
@@ -35,13 +35,26 @@ end
 end
 
 Base.getindex(d::JuMPArray, ::Colon) = d.innerArray[:]
+
+immutable JuMPKey
+    jk::Tuple
```
joehuchette: Does performance improve if this is instead
```jl
immutable JuMPKey{T<:Tuple}
    jk::T
end
```
?

IssamT: No it doesn't @joehuchette . The wrapper is the second modification I have made and it is the first one that had a negative impact on the performance, namely using iterators instead of indexes. Maybe I confused you, by not opening another PR. I thought it is related ...

Should I open a new PR for the JuMPKey wrapper?

```

  0.143376 seconds (110.83 k allocations: 2.241 MB)
  0.009154 seconds (60.00 k allocations: 937.559 KB)
  0.009054 seconds (60.00 k allocations: 937.559 KB)
  0.008394 seconds (60.00 k allocations: 937.559 KB)
  0.008086 seconds (60.00 k allocations: 937.559 KB)
  0.009080 seconds (60.00 k allocations: 937.559 KB)
  0.008189 seconds (60.00 k allocations: 937.559 KB)
  0.008068 seconds (60.00 k allocations: 937.559 KB)
  0.008026 seconds (60.00 k allocations: 937.559 KB)
  0.009072 seconds (60.00 k allocations: 937.559 KB)
  0.008065 seconds (60.00 k allocations: 937.559 KB)
  0.008093 seconds (60.00 k allocations: 937.559 KB)
  0.009081 seconds (60.00 k allocations: 937.559 KB)
  0.021784 seconds (60.00 k allocations: 937.559 KB)
  0.016168 seconds (60.00 k allocations: 937.559 KB)
  0.014673 seconds (60.00 k allocations: 937.559 KB)
  0.012351 seconds (60.00 k allocations: 937.559 KB)
  0.011812 seconds (60.00 k allocations: 937.559 KB)
  0.010432 seconds (60.00 k allocations: 937.559 KB)
  0.010156 seconds (60.00 k allocations: 937.559 KB)
```

joehuchette:
Unfortunately I don't think there is a way to check if an iterator supports indexing besides using something like ``applicable`` at run-time.

@IssamT could you provide a (practical) example where this behavior is a usability improvement? I'm a bit concerned about performance regressions here.

IssamT:
@joehuchette, thanks for the followup. 

I don't know if you saw the last commit (of yesterday), but right now there is no regression. the old `KeyIterator` is used in the case where indexsets are indexable and the new one is only used if there is at least one indexset that is not indexable. 

Let me summarize for you the current state:

**1)**  I  reproduce the same performance on the benchmark above, as before my patch, in the case where indexsets are indexable (the only case for which `KeyIterator` was working before).

**2)**  The new specialized `KeyIterator` handles non indexable indexsets. In particular it handles indexsets that are iterators. An example of such indexset is `edges(::Graph)` the function of LightGraphs that returns an iterator over edges. Another example is when the indexes are stored in a `Set`. This example was crashing before and now works fine,

set = Set()
push!(set,2)
push!(set,5)
push!(set,2)

set2 = Set()
push!(set,11)
push!(set,12)

m = Model()
@variable(m, x[set, set2], Bin)
@objective(m , Max, sum{sum{x[e,p], e in set}, p in set2})
solve(m)
sol = getvalue(x)

for i in keys(sol)
    @show sol[i...]
end

**3)** This is somehow a related but separate modification (As I have mentioned in a previous comment I don't know if I should put it in a separate PR). Having `JuMPKey` as a sepcial wrapper for the tuples used as keys in `JuMPArray` allows the intuitive use of keys without the ambiguity you were afraid of in #646. Hence this code still works fine

```
for i in keys(sol)
    @show sol[i...]
end
```

but now can also be replaced with the more intuitive version:

```
for i in keys(sol)
    @show sol[i]
end
```

joehuchette:
Ahh, ok, thanks for the detailed explanation. This is really nice, especially point 3. I agree that might be better suited as a separate PR, if it's not too difficult to extricate it. I'll have a deeper look over the code in the next few days.

src/JuMPContainer.jl
```diff
@@ -183,13 +185,86 @@ Base.length(it::ValueIterator)  = length(it.x)
...
     x::JA
+    dim::Int
+    next_k_cache::Array{Any,1}
+    function KeyIterator(d)
+        n = ndims(d.innerArray)
+        new(d, n, Array(Any, n+1))
+    end
+end
+
+KeyIterator{JA}(d::JA) = KeyIterator{JA}(d)
```
joehuchette: Is this method necessary with the inner constructor above?

src/JuMPContainer.jl
```diff
@@ -183,13 +185,86 @@ Base.length(it::ValueIterator)  = length(it.x)
...
+
+immutable IndexableIndexsets end
+immutable NotIndexableIndexsets end
```
joehuchette: Instead of using types here, a simple true/false should suffice. Using ``method_exists`` below means ``indexability`` will lead to dynamic function calls anyway.

src/JuMPContainer.jl
```diff
@@ -183,13 +185,86 @@ Base.length(it::ValueIterator)  = length(it.x)
...
+    quote
+        $(Expr(:tuple, [:(next(x.indexsets[$i], k[$i+1])[1]) for i in 1:N]...))
+    end
+end
+
+function Base.next(it::KeyIterator, k::Tuple)
+    cartesian_key = JuMPKey(_next(it.x, k))
+    pos = -1
+    for i in 1:it.dim
+        if(!done(it.x.indexsets[i], next(it.x.indexsets[i], k[i+1])[2] ) )
```
joehuchette: Style point: it's preferred to use whitespace, not parentheses, in if statements:
```jl
if !done(it.x.indexsets[i], next(it.x.indexsets[i], k[i+1])[2])
```

src/JuMPContainer.jl
```diff
@@ -183,13 +185,86 @@ Base.length(it::ValueIterator)  = length(it.x)
...
+        elseif i == pos
+            it.next_k_cache[i+1] = next(it.x.indexsets[i], k[i+1])[2]
+        else
+            it.next_k_cache[i+1] = k[i+1]
+        end
+    end
+    cartesian_key, tuple(it.next_k_cache...)
+end
+
+function Base.done(it::KeyIterator, k::Tuple)
```
joehuchette: This can just be
```jl
Base.done(it::KeyIterator, k::Tuple) = (k[1] == 1)
```

src/JuMPContainer.jl
```diff
@@ -183,13 +185,86 @@ Base.length(it::ValueIterator)  = length(it.x)
...
+    cartesian_key = JuMPKey(_next(it.x, k))
+    pos = -1
+    for i in 1:it.dim
+        if(!done(it.x.indexsets[i], next(it.x.indexsets[i], k[i+1])[2] ) )
+            pos = i
+            break
+        end
+    end
+    if pos == - 1
+        return cartesian_key, (1,1) #The second 1 is needed to ambiguity with ints
```
joehuchette: Could you explain this further please?

src/JuMPContainer.jl
IssamT: deleting line 196 gives this error

ERROR: LoadError: MethodError: Cannot `convert` an object of type JuMP.JuMPArray{Float64,2,Tuple{UnitRange{Int32},UnitRange{Int32}}} to an object of type JuMP.KeyIterator{JA<:JuMP.JuMPArray}
This may have arisen from a call to the constructor JuMP.KeyIterator{JA<:JuMP.JuMPArray}(...),

src/JuMPContainer.jl
IssamT: I think that traits are cleaner for dispatch. I agree that now it isn't impacting performance because the dispatch of the function `next` is done using the type of k that is either a `Tuple` for the new iterator or an `Integer` for the old one. But if we manage to change the new iterator to use an `Integer` state as well, traits will be needed. I don't mind deleting them right now.

src/JuMPContainer.jl
IssamT: Sorry about that. I will fix it in my next commit.

src/JuMPContainer.jl
IssamT: The first element of the tuple is used for indicating that we are done. The second 1 here is added because otherwise the dispatch of the function `next` fails and the function `next` with `Integer` state is called instead of the one with `Tuple` state

src/JuMPContainer.jl
IssamT: Ok. I will fix that as well on my next commit.

src/JuMPContainer.jl
joehuchette: Hmm, strange. OK, thanks.

src/JuMPContainer.jl
joehuchette: Personally I would prefer booleans, as this code is already a bit difficult for me to parse. We can always add complexity later on as-needed

src/JuMPContainer.jl
joehuchette: No worries, just a minor point

src/JuMPContainer.jl
joehuchette: Won't this mess with the type stability of the function? Could there a cleaner way to implement this without using a sentinel value return type?

IssamT: You re right. It s not very clean. The following modification doesn't change the performance (I tested it with several containers as index sets) but I will still do it to keep the exact return type everywhere

```
-        return cartesian_key, (1,1) #The second 1 is needed to ambiguity with ints
+        it.next_k_cache[1] = 1
+        return cartesian_key, tuple(it.next_k_cache...)
```

IssamT:
@joehuchette Is this ready for the merge? or would you like to add something more ?

joehuchette:
I'll take another look over it in the next few days, thanks for adding tests! Any chance you could run your performance benchmarks again on the final version?

IssamT:
Here are the results when using the final version.

  0.137093 seconds (78.45 k allocations: 1.692 MB)
  0.001770 seconds (29.49 k allocations: 347.711 KB)
  0.001667 seconds (29.49 k allocations: 347.711 KB)
  0.001740 seconds (29.49 k allocations: 347.711 KB)
  0.001732 seconds (29.49 k allocations: 347.711 KB)
  0.002207 seconds (29.49 k allocations: 347.711 KB)
  0.001791 seconds (29.49 k allocations: 347.711 KB)
  0.001795 seconds (29.49 k allocations: 347.711 KB)
  0.001719 seconds (29.49 k allocations: 347.711 KB)
  0.001709 seconds (29.49 k allocations: 347.711 KB)
  0.001722 seconds (29.49 k allocations: 347.711 KB)
  0.001717 seconds (29.49 k allocations: 347.711 KB)
  0.001746 seconds (29.49 k allocations: 347.711 KB)
  0.001784 seconds (29.49 k allocations: 347.711 KB)
  0.003164 seconds (29.49 k allocations: 347.711 KB)
  0.003190 seconds (29.49 k allocations: 347.711 KB)
  0.003259 seconds (29.49 k allocations: 347.711 KB)
  0.003265 seconds (29.49 k allocations: 347.711 KB)
  0.002996 seconds (29.49 k allocations: 347.711 KB)
  0.002934 seconds (29.49 k allocations: 347.711 KB)

joehuchette:
This is great, thanks for pushing it through, @IssamT! Do open the PR with the JuMPKey changes when you get a chance.

IssamT:
You're welcome. My thanks to you for the followup

Issue #846, opened 2016-09-06
Make writeLP use proper variable names


leethargo:
I was debugging some models in JuMP and found it annoying that they are printed as `VAR*` and not given the names to JuMP. There's a method `getname` that seems to work and I wonder whether `writeLP` should get a keywords arguments (eg, `realnames=false`) to make it use proper variable names.

Of course there could be problems with non-ascii variable names, or arrays of variables (maybe `x[1]` is an invalid variable name in lp format?), but then there could be workarounds or just an error message.

The documentation says something about `setname` being disabled for performance reasons, but anyway this behavior would be optional and non-default.

If you think this is viable and useful, I'd give it a try.

joehuchette:
This seems like a good idea for the default behavior; my only concerns are the cases you bring up. I would rewrite ``x[1,2,3]`` to ``x_1_2_3``. A PR would be very welcome!

mlubin:
Yeah, I think it's okay to make this the default behavior at this point. The performance concerns are not too relevant anymore.

mlubin:
Closed by #848

PR #848, opened 2016-09-09
use given names in writeLP

leethargo:
Fix issue #846.

Does not deal with non-ascii at all. Also, if no name is given, it will come out as `col_%d`.

### DIFF
--- a/src/writers.jl
+++ b/src/writers.jl
@@ -187,8 +187,8 @@ function varname_given(m::Model, col::Integer)
     name
 end
 
-function writeLP(m::Model, fname::AbstractString; generic=true)
-    varname = generic ? varname_generic : varname_given
+function writeLP(m::Model, fname::AbstractString; genericnames=true)
+    varname = genericnames ? varname_generic : varname_given
 
     f = open(fname, "w")
 
--- a/src/writers.jl
+++ b/src/writers.jl
@@ -175,7 +175,20 @@ end
 # LP File Writer
 # We use the formatting defined at:
 #   http://lpsolve.sourceforge.net/5.0/CPLEX-format.htm
-function writeLP(m::Model, fname::AbstractString)
+
+varname_generic(m::Model, col::Integer) = "VAR$(col)"
+
+function varname_given(m::Model, col::Integer)
+    # TODO: deal with non-ascii characters?
+    name = getname(m, col)
+    for (pat, sub) in [("[", "_"), ("]", ""), (",", "_")]
+        name = replace(name, pat, sub)
+    end
+    name
+end
+
+function writeLP(m::Model, fname::AbstractString; generic=true)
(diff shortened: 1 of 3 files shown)


leethargo:
Maybe the selection of method (`varname_generic` vs `varname_given`) could be done with multiple dispatch instead?

src/writers.jl
```diff
@@ -175,7 +175,20 @@ end
 # LP File Writer
 # We use the formatting defined at:
 #   http://lpsolve.sourceforge.net/5.0/CPLEX-format.htm
-function writeLP(m::Model, fname::AbstractString)
+
+varname_generic(m::Model, col::Integer) = "VAR$(col)"
+
+function varname_given(m::Model, col::Integer)
+    # TODO: deal with non-ascii characters?
+    name = getname(m, col)
```
mlubin: Maybe run ``isascii`` here and throw an error if it returns false?

leethargo: I kind of want it to not fail, because I use greek letters a lot. So even if solvers can't read those `lp` files, I would use them for debugging.

mlubin:
@leethargo, could you run some before/after and ``generic=true``/``generic=false`` tests with https://github.com/JuliaOpt/JuMP.jl/blob/master/test/perf/speed.jl? Would be helpful to get a sense of the performance differences.

leethargo:
Sure, here you go:
```
# before (2555955)
PMEDIAN BUILD MIN=0.514896891  MED=0.697827519
PMEDIAN WRITE MIN=2.546088368  MED=2.695454883
CONT5 BUILD   MIN=0.269833307  MED=0.336173836
CONT5 WRITE   MIN=2.12491588   MED=2.162008624

# after (497c517) with default setting (generic=true)
PMEDIAN BUILD MIN=0.523728515  MED=0.750038129
PMEDIAN WRITE MIN=3.655520066  MED=4.036045283
CONT5 BUILD   MIN=0.283510199  MED=0.354755032
CONT5 WRITE   MIN=3.153419501  MED=3.253564387

# after (497c517~dirty) with new setting (generic=false)
PMEDIAN BUILD MIN=0.521917695  MED=0.751309123
PMEDIAN WRITE MIN=10.61211384  MED=10.872961377
CONT5 BUILD   MIN=0.361047545  MED=0.370199847
CONT5 WRITE   MIN=8.245381025  MED=8.462890679
```

Hopefully I did everything right. For the last test, I changed all calls to `writeLP` to use `generic=false` and created a commit to force recompilation.

mlubin:
I'm okay with that slowdown. I would rename ``generic`` to ``genericnames`` for clarity.

mlubin:
Now just needs updates to the docs and good to go.

leethargo:
Done, but did not test building the docs.

PR #850, opened 2016-09-13
Implement NLopt parameter change to avoid hanging during test

colbec:
NLopt parameter xtol_abs set to nothing can cause hanging in reformulation test.
Avoid hanging by specifying an arbitrary value of 1.0.
There appears to be a critical point somewhere below 0.34 for this parameter where hanging can occur.
Ref: https://github.com/JuliaOpt/JuMP.jl/issues/849

### DIFF
--- a/test/nonlinear.jl
+++ b/test/nonlinear.jl
@@ -430,6 +430,13 @@ end; end; end
 facts("[nonlinear] Test entropy maximization (reformulation)") do
 for nlp_solver in convex_nlp_solvers
 context("With solver $(typeof(nlp_solver))") do
+    # the reformulation test can cause hanging in the NLopt test
+    # when xtol_abs is nothing (as per solver.jl included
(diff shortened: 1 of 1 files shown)


mlubin:
Could you set the parameter [here](https://github.com/JuliaOpt/JuMP.jl/blob/616deee398ef80d4c0f6e218d1e20bd82998f864/test/solvers.jl#L117) instead?

colbec:
I can certainly do so, however my impression is that this alternate location resets the parameter for all tests related to NLopt; since only the reformulation test seems to be affected by this is this not informative, ie will this not affect the other tests?

mlubin:
Ideally we should have a single set of parameters for which the solver passes tests. If the point is just to disable this test for NLopt then we should do that instead of having ugly code to switch options that trigger a failure anyway.

mlubin:
Closing this as stale. Will still consider valid fixes to get NLopt passing.

PR #860, opened 2016-10-04
explicit warning when multiple variables and constraints are given the same name

mlubin:
As suggested at https://github.com/JuliaOpt/JuMP.jl/issues/692#issuecomment-251197721, I think this is a good step to push people towards anonymous variables and reduce potential confusion. Users will now get a warning on ``@variable(m, x[1]); @variable(m, x[2])``.

CC @odow @yeesian @JackDunnNZ @davidanthoff @chriscoey @ccoffrin

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -429,7 +429,7 @@ function getvalue(arr::Array{Variable})
             end
         end
     end
-    # Copy printing data from @defVar for Array{Variable} to corresponding Array{Float64} of values
+    # Copy printing data from @variable for Array{Variable} to corresponding Array{Float64} of values
     if registered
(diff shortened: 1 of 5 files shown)


mlubin:
@joehuchette @IainNZ any objections?

joehuchette:
Nope, this is great

joehuchette:
I guess it's not possible to test that the warning gets triggered?

mlubin:
It is currently covered by the tests but not sure how to explicitly test for it.

ccoffrin:
Last time I looked into it (v0.4.6), there was no way to catch a warning to generate a unit test for it.

PR #863, opened 2016-10-07
fix #861

### LINKED ISSUE #861: [Regression?] Operator .^ in objective functions no more working in Julia 0.5
When trying some of my code on the new Julia 0.5.0, I am getting error messages which I did not have with 0.4.6. With both, I'm using JuMP 0.14.1 (I've run `Pkg.update()` on both today). 

This is the incriminate code: 

```julia
m = Model()
@variable(m, xs[1:5] >= 0)
@objective(m, Min, sum(xs .^ 2))
```

Julia 0.4.6 has no problem with it, but 0.5.0 shouts: 

```
ERROR: MethodError: no method matching .^(::JuMP.Variable, ::Int64)
Closest candidates are:
  .^(::Range{T}, ::Number) at range.jl:813
  .^(::Number, ::Number) at operators.jl:151
  .^(::BitArray{N}, ::Integer) at bitarray.jl:1195
  ...
 in collect(::Base.Generator{Array{JuMP.Variable,1},Base.##164#166{Int64}}) at .\array.jl:307
 in .^(::Array{JuMP.Variable,1}, ::Int64) at .\arraymath.jl:79
```

Is this something about that "fused broadcast syntax" that came with 0.5?

mlubin:
CC @dourouc05

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -124,6 +124,7 @@ function (^)(lhs::Union{Variable,AffExpr}, rhs::Integer)
     end
 end
 (^)(lhs::Union{Variable,AffExpr}, rhs::Number) = error("Only exponents of 0, 1, or 2 are currently supported. Are you trying to build a nonlinear problem? Make sure you use @NLconstraint/@NLobjective.")
+(.^)(lhs::Union{Variable,AffExpr}, rhs::Number) = lhs^rhs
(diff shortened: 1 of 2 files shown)


dourouc05:
It seems to solve my problems, thank you!

PR #865, opened 2016-10-08
remove old deprecations

mlubin:
@tkelman, could you confirm that this will not lead to loss of life or property?

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -25,10 +25,6 @@ import Compat: UTF8String, view
 
 @compat import Base.show
 
-function __init__()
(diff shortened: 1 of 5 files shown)


tkelman:
How long have these had the abundance-of-caution error for now?

tkelman:
Actually when are you planning on dropping support for Julia 0.4? It might make a little bit of sense to have the first tagged version coincide for these two things.

mlubin:
> How long have these had the abundance-of-caution error for now?

7+ months

> Actually when are you planning on dropping support for Julia 0.4? It might make a little bit of sense to have the first tagged version coincide for these two things.

That's exactly the plan.

PR #868, opened 2016-10-11
[WIP] Add support for duals with SDP

blegat:
This pull request adds the support for using `getdual` in a conic model when SDP constraints are present (I need it for [this issue](https://github.com/blegat/SumOfSquares.jl/issues/1) of `SumOfSquares.jl`).
Please feel free to criticize anything I might be done in a better way :)

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -88,6 +88,11 @@ type Model <: AbstractModel
     linconstrDuals::Vector{Float64}
     conicconstrDuals::Vector{Float64}
     constrDualMap::Vector{Vector{Int}}
+    # Vector of the same length as sdpconstr.
+    # sdpconstrSym[c] is the list of pairs (i,j), i > j
+    # such that a symmetry-enforcing constraint has been created
+    # between sdpconstr[c].terms[i,j] and sdpconstr[c].terms[j,i]
+    sdpconstrSym::Vector{Vector{Tuple{Int,Int}}}
     # internal solver model object
     internalModel
     # Solver+option object from MathProgBase
@@ -157,7 +162,8 @@ function Model(;solver=UnsetSolver(), simplify_nonlinear_expressions::Bool=false
           Float64[],                   # redCosts
           Float64[],                   # linconstrDuals
           Float64[],                   # conicconstrDuals
-          Any[],                       # constrDualsMap
+          Vector{Int}[],               # constrDualMap
+          Vector{Tuple{Int,Int}}[],    # sdpconstrSym
           nothing,                     # internalModel
           solver,                      # solver
           false,                       # internalModelLoaded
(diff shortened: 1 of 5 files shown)


mlubin:
This is great! I'll comment more inline.
Looks like there's a couple small merge conflicts due to https://github.com/JuliaOpt/JuMP.jl/pull/866

src/JuMP.jl
```diff
@@ -93,6 +93,7 @@ type Model <: AbstractModel
     linconstrDuals::Vector{Float64}
     conicconstrDuals::Vector{Float64}
     constrDualMap::Vector{Vector{Int}}
+    sdpconstrSym::Vector{Vector{Tuple{Int,Int}}}
```
mlubin: Could you document what this field means?

src/JuMP.jl
```diff
@@ -579,17 +582,70 @@ function getNumSOCRows(m::Model)
     return numSOCRows
 end
 
-function getdual(c::ConstraintRef{Model,SOCConstraint})
-    numBndRows = getNumBndRows(c.m)
-    numSOCRows = getNumSOCRows(c.m)
-    if length(c.m.conicconstrDuals) != (MathProgBase.numlinconstr(c.m) + numBndRows + numSOCRows)
-        error("Dual solution not available. Check that the model was properly solved and no integer variables are present.")
+# Returns the number of semidefinite constraints
```
mlubin: Small terminology point, but I'd think of ``X \in SDP`` as a single semidefinite constraint. SDP rows is okay, just responding to the comment.

src/JuMP.jl
```diff
@@ -579,17 +582,70 @@ function getNumSOCRows(m::Model)
...
-        c.m.constrDualMap[MathProgBase.numlinconstr(c.m) + numBndRows + c.idx]]
+    return numSDPRows
 end
 
+# Returns the number of symmetry-enforcing constraints for SDP constraints
+function getNumSymRows(m::Model)
+    sum(map(length, m.sdpconstrSym))
+end
 
+function getconicdualaux(m::Model, idx, issdp)
```
mlubin: This method could use a descriptive comment

src/JuMP.jl
```diff
@@ -579,17 +582,70 @@ function getNumSOCRows(m::Model)
...
+    idx = 0
+    for i in 1:n
+        for j in i:n
+            idx += 1
+            X[j,i] = X[i,j] = dual[idx] / sqrt(2)
+        end
+    end
+    @assert length(symdual) == length(c.m.sdpconstrSym[c.idx])
+    idx = 0
+    for (i,j) in c.m.sdpconstrSym[c.idx]
```
mlubin: Could you sketch out the math a bit or provide a reference for why this is the correct way to compute a dual involving symmetry constraints?

mlubin:
Does this work at all for SDP variables?

src/JuMP.jl
blegat: You are right, the comment is confusing, I have changed it

src/JuMP.jl
blegat: I have added a comment

src/JuMP.jl
blegat: This is a bit tricky to explain without going into the details and I have not found a reference (classical reference only consider the symmetric case).
Therefore I have added a detailed explanation of why it is the correct way to do so

blegat:
No I have not touched at SDP variable, only constraints

mlubin:
> No I have not touched at SDP variable, only constraints

So what happens if someone does:
```jl
@variable(m, X[1:N,1:N], SDP)
...
solve(m)
getdual(X)
```

blegat:
I have just implemented `getdual` for SDP variable in bf2b120 :)

src/JuMP.jl
```diff
@@ -441,6 +441,63 @@ function getdual(v::Variable)
...
+                    sdpvar = false
+                end
+            end
+        end
+        if sdpvar && !((:SDP, start:arr[n,n].col) in m.varCones)
+            sdpvar = false
+        end
+    end
+    if !sdpvar
+        error("getdual on a matrix that is not an SDP variable is not supported")
```
mlubin: Except it is supported, you're allowed to define a matrix of variables, use them in an LP, and then ask for the reduced costs.

mlubin: See the method defined [here](https://github.com/JuliaOpt/JuMP.jl/blob/66379c4cd9e58eae30ce307fa9618aa9711ab452/src/JuMPContainer.jl#L101). What you may want to do is look into merging this logic into ``fillConicRedCosts``. ``getdual`` is already defined to read from the ``redCosts`` vector.

blegat: Then it seems that my approach is completely wrong.
In fact it already worked without my new function :-P

mlubin: Are you sure? Check that the off-diagonals are correct also.

mlubin: D'accord, bonne nuit.

blegat: Merci :D

blegat:
Normally, I have finished getdual for both SDP variable and SDP constraints. I have removed your scaling by `-1` in `fillConicRedCosts` (but not in `fillConicDuals`). I do not really understand why we need this scaling and I had to remove it to make getdual work for SDP variables.

mlubin:
The reason we do the rescaling is for consistency with LP duality. If you add an inactive SOC or SDP constraint to an LP, you shouldn't have the signs flip on you. See the test [here](https://github.com/JuliaOpt/JuMP.jl/blob/66379c4cd9e58eae30ce307fa9618aa9711ab452/test/socduals.jl#L73). I'm open to discussion on how to make this more coherent, maybe unflip the SDP and SOC duals and leave the signs flipped on the linear constraints?

src/macros.jl
```diff
@@ -422,7 +422,7 @@ macro SDconstraint(m, x)
     m = esc(m)
 
     if isa(x, Symbol)
-        error("in @SDConstraint: Incomplete constraint specification $x. Are you missing a comparison (<= or >=)?")
+        error("in @SDconstraint: Incomplete constraint specification $x. Are you missing a comparison (<= or >=)?")
```
mlubin: Good catch

blegat:
I do not dispute the fact that we need the `-1` scaling for `fillConicDuals`. By the way, if I remove it and run the tests, the whole world collapses.
But if I change it for `fillConicRedCosts`, nothing changes. This is because the redcosts are always 0 in every tests...
I have added a test (see 683457d) in `sdp.jl` (called "[sdp] Test nonzero dual for a scalar variable with sdp solver") modeling an LP using an SDP constraint and I have compared it with its implementation using only linear constraints. We need no `-1` scaling to get consistency with LP (adding the scaling creates the inconsistency).

mlubin:
I'm satisfied. @joehuchette, any feedback?

joehuchette:
Probably nothing major, but I'll take a look later this afternoon.

joehuchette:
This is fantastic, very thorough. Good to go for me, besides some very minor notes on some of the comments.

mlubin:
@joehuchette, what are the minor notes?

joehuchette:
Whoops, guess I never actually pressed the button to submit the review...

src/JuMP.jl
```diff
@@ -88,6 +88,11 @@ type Model <: AbstractModel
     linconstrDuals::Vector{Float64}
     conicconstrDuals::Vector{Float64}
     constrDualMap::Vector{Vector{Int}}
+    # Vector of the same length as sdpconstr
+    # sdpconstrSym[i] is the list of pairs (i,j), i > j
```
blegat: Yes indeed, the index should the constraint index, I will change that tomorrow :)

blegat: Fixed :)

mlubin:
Thanks, @blegat!

PR #875, opened 2016-10-23
Improves Xpress documentation and add it to tests

joaquimg:
I just committed to METADATA Xpress v0.1.0 which passes all tests from JuMP.

Therefore, we have it working for SOCP and QCP.


doc/installation.rst
```diff
@@ -129,7 +129,10 @@ ECOS can be used by JuMP to solve LPs and SOCPs. ECOS does not support general q
 FICO Xpress
 +++++++++++
 
-Requires a license. The interface is experimental and has not been widely tested. Callbacks are not yet supported. SOCPs have not been tested.
+Requires a working installation of Xpress with an active license (it is possible to get free license for academic use). Supports SOCP and "SOS" constraints. The interface is experimental, but it does pass all JuMP and MathProgBase tests. Callbacks are not yet supported.
```
mlubin: "Free for academic use" seems misleading, it's not nearly as easy to obtain as Gurobi, for example. Maybe we should just link to http://subscribe.fico.com/Academic-Partner-Program and let people figure it out.

joaquimg: I agree, indeed it not easy to get the license like for Gurobi.
Just committed new suggestion, what do you think?

mlubin:
Ignoring the nightly failures. Thanks!

PR #879, opened 2016-10-28
call registercallbacks also for NLPs

### LINKED ISSUE #878: Support Callbacks with nonlinear models
Currently, JuMP only registers callbacks for models that are neither nonlinear nor conic, that is, LinearQuadratic in MPB speak.

See the discussion in [SCIP.jl/#26](https://github.com/SCIP-Interfaces/SCIP.jl/issues/26) about how this is limiting from the point of view of solvers such as SCIP that support both nonconvex MINLP models and various callbacks.

Also somewhat related: #779.

leethargo:
Should fix #878.


mlubin:
This needs corresponding documentation updates in MathProgBase and JuMP itself.

PR #888, opened 2016-11-09
Avoid test failure if SCS is not loaded

blegat:


blegat:
Julia nightly is making my tests fail on Windows 32 bits and Mac OS 64 bits :(

PR #891, opened 2016-11-10
Fix #889

### LINKED ISSUE #889: `getvalue` method for sparse matrices of variables
There currently isn't one.  Here's one I wrote, I'm not sure if it's sufficiently general to get added to JuMP
```julia
function getvalue{Ti <: Integer}(x::SparseMatrixCSC{JuMP.Variable, Ti})
    x_val = spzeros(size(x)...)
    rows = rowvals(x)
    vars = nonzeros(x)
    m, n = size(x)
    for j ∈ 1:n
        for idx ∈ nzrange(x, j)
            var = vars[idx]
            # this is the row number
            i = rows[idx]
            x_val[i, j] = getvalue(var)
        end
    end
    x_val
end

```

blegat:


mlubin:
Thanks for the PR! Should we be doing this though or should we move towards the built-in broadcast syntax?

blegat:
The issue with the broadcast syntax is that it does not preserve the sparsity
```
julia> f(x) = x*x

julia> X = sparse([1, 3], [2, 3], [3, 9])
3×3 sparse matrix with 2 Int64 nonzero entries:
	[1, 2]  =  3
	[3, 3]  =  9

julia> f.(X)
3×3 Array{Int64,2}:
 0  9   0
 0  0   0
 0  0  81

julia> broadcast(f, X)
3×3 Array{Int64,2}:
 0  9   0
 0  0   0
 0  0  81
```

mlubin:
Is there an upstream discussion on this?

blegat:
The thing is that `broadcast` needs to be sure that the function maps zero to zero.
Currently there is the `_broadcast_unary_nz2nz_z2z` where we ensure that nonzero are mapped to nonzero and zero are mapped to zeros so that it can just apply the function to every nonzero entry.
There is also this pull request: https://github.com/JuliaLang/julia/pull/19239 that checks when you call broadcast whether the function maps zero to zero.

Do you think we should only support `getvalue` of objects created by JuMP (like `@variable [1:n,1:m]`) and ask the user to use `getvalue.` when he has created a container of JuMP expressions since it seems that in Julia 0.6, broadcast on sparse matrices will return sparse matrices ?

mlubin:
Yes, I'd say we shouldn't implement any magic for types that aren't directly created by JuMP. JuMP will never directly create a sparse matrix of variables. Mapping zero to zero is nonsensical in this case though, hope the broadcasting works when you have nonnumeric entries.

blegat:
If the JuliaLang pull request is merged as is, it will fail since it will do
```
julia> getvalue(typeof(zero(JuMP.Variable))) == zero(Variable)
ERROR: MethodError: no method matching getvalue(::Type{JuMP.GenericAffExpr{Float64,JuMP.Variable}})
Closest candidates are:
  getvalue(::JuMP.NonlinearExpression) at /home/blegat/.julia/v0.5/JuMP/src/nlp.jl:1236
  getvalue(::JuMP.NonlinearParameter) at /home/blegat/.julia/v0.5/JuMP/src/nlp.jl:39
  getvalue(::JuMP.GenericQuadExpr{Float64,JuMP.Variable}) at /home/blegat/.julia/v0.5/JuMP/src/quadexpr.jl:92
  ...
```

mlubin:
Well that's an issue. Should we let them know?

blegat:
We could also define getvalue for AffineExpr

mlubin:
IIRC that should already be defined.

blegat:
Oh I didn't write it correctly... What is going to be called is
```julia
julia> getvalue(zero(Variable)) == zero(Variable)
false
julia> typeof(zero(Variable))
JuMP.GenericAffExpr{Float64,JuMP.Variable}
julia> typeof(getvalue(zero(Variable)))
Float64
```
The problem is that there is no method `==` between an `AffExpr` and something else. If we add
```julia
(==)(lhs::AffExpr,rhs) = (sum(abs(lhs.coeffs)) == 0) && (lhs.constant == rhs)
(==)(lhs,rhs::AffExpr) = rhs == lhs
```
then we get
```julia
julia> getvalue(zero(Variable)) == zero(Variable)
true
```

joehuchette:
Any reason not to merge this?

mlubin:
I remove my objection. If you update the PR and get tests passing this is okay to merge. We'll revisit the use of broadcast syntax for everything in JuMP at a later point.

src/JuMPContainer.jl
```diff
@@ -98,8 +98,8 @@ pushmeta!(x::JuMPContainer, sym::Symbol, val) = (x.meta[sym] = val)
 getmeta(x::JuMPContainer, sym::Symbol) = x.meta[sym]
 
 # duck typing approach -- if eltype(innerArray) doesn't support accessor, will fail
-for accessor in (:getdual, :getlowerbound, :getupperbound)
-    @eval $accessor(x::Union{JuMPContainer,Array}) = map($accessor,x)
+for accessor in (:getdual, :getlowerbound, :getupperbound, :getvalue)
```
mlubin: Isn't ``getvalue(::JuMPContainer)`` already explicitly defined somewhere else? There's a reason it's not on this list.

blegat: That is an excellent question. I have updated the PR with a proposition to unify these accessors on `JuMPContainer/AbstractArray`

joehuchette:
Seems reasonable to me.

joehuchette:
Seems reasonable to me, but Travis has some pretty crazy-looking errors.

blegat:
Wouldn't it be more consistent with `getvalue` if `getdual` outputs a warning and a vector of NaN instead of throwing an error ?

joehuchette:
I would think so, yes. 






On Wed, Dec 14, 2016 at 12:47 PM -0500, "Benoît Legat" <notifications@github.com> wrote:












Wouldn't it be more consistent with getvalue if getdual outputs a warning and a vector of NaN instead of throwing an error ?

mlubin:
@blegat, if you tidy this up this week we can get it in for 0.15.

blegat:
@mlubin @joehuchette The PR is ready for review, the tests are passing for Julia v0.5

joehuchette:
Thanks, @blegat!

PR #895, opened 2016-11-17
Remove zero elements in IndexedVector

blegat:
It often happens that when taking the sum/different of linear expressions, terms cancel. While leaving zeros, e.g. in the conic sparse matrix A, may not be that bad, it still makes solvers throw warning (e.g. Mosek) which can scare the user (at least it scared me).
The `rmz!` function implemented in this PR should not occur any performance overhead at all while removing those slighly annoying issues. What do you think ?

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -25,6 +25,21 @@ function addelt!{T}(v::IndexedVector{T},i::Integer,val::T)
     return nothing
 end
 
+function rmz!{T}(v::IndexedVector{T})
+    i = 1
+    while i <= v.nnz
+        j = v.nzidx[i]
+        if v.elts[j] == zero(T)
+            v.empty[j] = true
+            # If i == v.nnz then this has no effect but it would be inefficient to branch
+            v.nzidx[i] = v.nzidx[v.nnz]
+            v.nnz -= 1
+        else
+            i += 1
+        end
+    end
+end
+
 function Base.empty!{T}(v::IndexedVector{T})
     elts = v.elts
     nzidx = v.nzidx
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -25,6 +25,21 @@ function addelt!{T}(v::IndexedVector{T},i::Integer,val::T)
...
 end
 
+function rmz!{T}(v::IndexedVector{T})
+    i = 1
+    while i <= v.nnz
+        j = v.nzidx[i]
+        if v.elts[j] == zero(T)
+            v.empty[j] = true
+            # If i == v.nnz then this has no effect but it would be inefficient to branch
+            v.nzidx[i] = v.nzidx[v.nnz]
```
mlubin: Clever

mlubin:
Could you do a quick before/after comparison on test/perf/speed2.jl to double check that there are no performance regressions?

blegat:
There does not seem to be any regression. See below:
* on master (3f2570afcfe3fa0dd413449939da0881655795be):
```
PMEDIAN BUILD MIN=0.228448775  MED=0.501220294
PMEDIAN INTRN MIN=0.212174892  MED=0.404674772
CONT5 BUILD   MIN=0.205204792  MED=0.39549537
CONT5 INTRN   MIN=0.29409365  MED=0.323526576
```
* on nzclean (d1be44c1d5c991739b6f2a8c602e88e1ebe8ab85):
```
PMEDIAN BUILD MIN=0.205619716  MED=0.422099735
PMEDIAN INTRN MIN=0.183113881  MED=0.36273885
CONT5 BUILD   MIN=0.21183477  MED=0.340012733
CONT5 INTRN   MIN=0.294772246  MED=0.311564861
```

PR #896, opened 2016-11-22
kill curly braces

mlubin:
This reimplements generator parsing to be entirely consistent with the usage in Julia. We can now parse
```jl
sum(x[i,j] for i in 1:N for j in 1:i)
```
and nested conditions like
```jl
sum(x[i,j] for i in 1:N if iseven(i) for j in 1:i if isodd(j))
```
Notably this breaks the behavior that I hacked together:
```jl
sum(x[i,j] for i in 1:N, j in 1:i)
```
which works currently in JuMP but doesn't work in plain Julia. I referenced this inconsistency in the [post](https://groups.google.com/d/msg/julia-opt/vUK1NHEHqfk/WD-6lSbMCAAJ) announcing the syntax change. I don't think we have a choice but to be consistent with Julia.

Another change is that I dumped ``norm2`` and friends for ``norm`` directly, e.g.,
```jl
norm(x[i] for i in 1:2) # 2-norm by default
norm((x[i] for i in 1:2), 1) # 1-norm with second argument
```
This is also consistent with Julia (https://github.com/JuliaOpt/JuMP.jl/issues/892) but the extra pair of parentheses needed is unfortunate. CC @IainNZ who's a user of non-L2-norms.

TODO:
- [x] Deprecation warnings for curly syntax
- [x] Figure out what to do about curlies with no index sets (``sum{x[i]; foo}``)
- [x] Update docs, examples, and tests

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -8,6 +8,55 @@ using Base.Meta
 issum(s::Symbol) = (s == :sum) || (s == :∑) || (s == :Σ)
 isprod(s::Symbol) = (s == :prod) || (s == :∏)
 
+function curly_to_generator(x)
+    # we have a filter condition
+    @assert isexpr(x,:curly)
+    if isexpr(x.args[2],:parameters)
+        cond = x.args[2].args[1]
+        body = x.args[3]
+        for i in length(x.args):-1:4
+            if i == length(x.args)
+                body = Expr(:generator,body,Expr(:filter,cond,x.args[i]))
+            else
+                body = Expr(:generator,body,x.args[i])
+            end
+        end
+    else
+        cond = nothing
+        body = x.args[2]
+        for i in length(x.args):-1:3
+            body = Expr(:generator,body,x.args[i])
+        end
+    end
+    if isexpr(body.args[1],:generator)
+        body = Expr(:flatten,body)
+    end
+    name = x.args[1]
+    if name == :norm2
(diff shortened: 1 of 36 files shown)


mlubin:
Also notable is that I removed the step that preallocates linear expressions anticipating the length of a summation, and this seemed give uniform improvements or no change at all on the benchmarks.

mlubin:
Actually I was mistaken, the new parsing changes won't break anything at all. Because of the way we parse things, ``sum(x[i,j] for i in 1:N, j in 1:i)`` will work fine inside JuMP code. People are free to use either form at this point.

mlubin:
Basically ready to go on my end, will merge soon. Parentheses are beautiful.

src/macros.jl
```diff
@@ -8,11 +8,61 @@ using Base.Meta
...
+                body = Expr(:generator,body,Expr(:filter,cond,x.args[i]))
+            else
+                body = Expr(:generator,body,x.args[i])
+            end
+        end
+    else
+        cond = nothing
+        body = x.args[2]
+        if length(x.args) == 2 # no iteration set!
+            push!(x.args,:(_ in 1))
```
joehuchette: ?

mlubin: For ``sum{foo;cond}``we will suggest ``sum(foo for _ in 1; cond)``. Not sure what else to do here.

PR #899, opened 2016-11-23
add test for #877

### LINKED ISSUE #877: constraints being ignored/removed
I am building some fairly complicated convex quadratic constraints from data where sometimes the coefficient of the quadratic term is 0.0 (or sometimes very small).  It looks like if such a constraint is constructed, then the entire constraint ignored, even if there is other information in the constraint. 
I expect this is done to improve numerically stability, but is this overall desired behavior? Would better behavior be to just drop the close to zero atoms from the constraint, rather than the whole constraint? 

Here is some simple example code that reproduces what I see

```

using MathProgBase
using JuMP
using Gurobi
using CPLEX

gurobi_solver = GurobiSolver()
cplex_solver = CplexSolver()

# The constraint appears to be ignored because there is a 0 coefficient in the x^2 term (Gurobi)
m = Model(solver=gurobi_solver)
@variable(m, 0 <= v <= 2)
@variable(m, 1 <= x <= 5)
c = @constraint(m, v >= 0.0 * x^2 + x)
println(c)
@objective(m, Min, v)
status = solve(m)
println("Objective value: ", getobjectivevalue(m))

# The constraint appears to be ignored because there is a 0 coefficient in the x^2 term (cplex)
m = Model(solver=cplex_solver)
@variable(m, 0 <= v <= 2)
@variable(m, 1 <= x <= 5)
@constraint(m, v >= 0.0 * x^2 + x)
@objective(m, Min, v)
status = solve(m)
println("Objective value: ", getobjectivevalue(m))

# If I drop the ) * x^2 term, everything is ok
m = Model(solver=gurobi_solver)
@variable(m, 0 <= v <= 2)
@variable(m, 1 <= x <= 5)
@constraint(m, v >= x)
@objective(m, Min, v)
status = solve(m)
println("Objective value: ", getobjectivevalue(m))

# Similarly, if I make the coefficient of the x^2 term non zero, everything is ok
m = Model(solver=gurobi_solver)
@variable(m, 0 <= v <= 2)
@variable(m, 1 <= x <= 5)
c = @constraint(m, v >= 0.01 * x^2 + x)
println(c)
@objective(m, Min, v)
status = solve(m)
println("Objective value: ", getobjectivevalue(m))



```

mlubin:
Closes #877

### DIFF
--- a/test/model.jl
+++ b/test/model.jl
@@ -1014,3 +1014,15 @@ facts("[model] .^ broadcasting") do
     @variable(m, x[1:2])
     @fact (x.^2)[1] --> x[1]^2
 end
+
+facts("[model] Quadratic constraints with zero coefficients") do
+for solver in quad_solvers
+context("With solver $(typeof(solver))") do
+    m = Model(solver=solver)
+    @variable(m, 0 <= v <= 2)
+    @variable(m, 1 <= x <= 5)
+    @constraint(m, v >= 0.0 * x^2 + x)
+    @objective(m, Min, v)
+    @fact solve(m) --> :Optimal
+    @fact getvalue(v) --> roughly(1.0, TOL)
+end; end; end


mlubin:
If anything these tests have been catching issues in CoinOptServices: https://github.com/JuliaOpt/CoinOptServices.jl/issues/26

PR #906, opened 2016-12-03
Fix error for invalid lb/ub in range constraint

blegat:


mlubin:
Looks like some other tests need to be corrected:
```
Check @constraint basics: Test Failed
  Expression: @constraint m x <= t <= y
    Expected: UndefVarError
      Thrown: ErrorException
```

blegat:
@mlubin Fixed. It is now failing on Julia nightly just like master is also failing (any idea why ?)

mlubin:
I haven't looked into the errors on nightly yet. Occasionally they go away on their own.

Issue #600, opened 2015-10-18
Avoid generic fallback for Ac_mul_B


mlubin:
Writing `X'*y` can lead to performance warnings. Looks like it's using the generic fallback.

```
julia> m = Model();

julia> @defVar(m, X[1:3,1:3]);

julia> @defVar(m, y[1:3]);

julia> @which X*y
*[...] at /home/mlubin/.julia/v0.4/JuMP/src/operators.jl:390

julia> @which X'*y
Ac_mul_B{T,S}(A::Union{DenseArray{T,2},SubArray{T,2,A<:DenseArray{T,N},I<:Tuple{Vararg{Union{Colon,Int64,Range{Int64}}}},LD}}, x::Union{DenseArray{S,1},SubArray{S,1,A<:DenseArray{T,N},I<:Tuple{Vararg{Union{Colon,Int64,Range{Int64}}}},LD}}) at linalg/matmul.jl:119
```

joehuchette:
I really didn't want to get in the business of overloading ``A_mul_B`` and friends, but maybe it's inevitable.

mlubin:
There are likely still performance issues in the overloads, but this is stale because all of the linear algebra has been rewritten.

Issue #907, opened 2016-12-05
addition operator warning during matrix operations


ExpandingMan:
Normal matrix multiplication of a sparse matrix of variables applied to a vector
```julia
X*v
```
generates the following warning when `X` is sufficiently large (in my case about 5*10^6 non-zero entries) 
```
WARNING: The addition operator has been used on JuMP expressions a large number of times. This warning is safe to ignore but may indicate that model generation is slower than necessary. For performance reasons, you should not add expressions in a loop. Instead of x += y, use append!(x,y) to modify x in place. If y is a single variable, you may also use push!(x, coef, y) in place of x += coef*y.
```
This seems non-ideal.  Since `*` of a matrix of JuMP variables is a method of JuMP, shouldn't it just use the most efficient method available?

I have tested this on a dense matrix with approximately the same number of (non-zero) elements, and it is quite fast and does not generate a warning, so this appears to only be an issue with dense matrices (unless it somehow has to do with the matrix size).

mlubin:
We would welcome a contribution of an efficient specialized implementation. By default we fallback on Julia's generic implementations which are not efficient for working with JuMP expression objects.

ExpandingMan:
I see.  I've been looking into this and it actually seems to be rather a huge amount of work, which is probably what it hasn't been done (there are other problems with sparse matrices as well, #887 has been rather a big problem for me).

I'm not sure I have the expertise to do a really perfect job on this, but I'm going to experiment with it today, as all these have been ongoing issues for me.  Will post here again if I make progress.

ExpandingMan:
Ok, I'm definitely learning a lesson about not putting the *complete* code I used to get the error or warning in the issue reports.

I could have sworn that I was getting this from matrix multiplication alone, but for some reason today I can only reproduce it within a macro.

```julia
using JuMP

const m = 100
const n = 100

const M = 60000
const N = 60000

model = Model()

X = spzeros(Variable, M, N)
i, j = 1, 1
while (i + m - 1 ≤ M) && (j + n - 1 ≤ N)
    X[i:(i+m-1), j:(j+n-1)] = @variable(model, [1:m, 1:n])
    i += m
    j += n
end

u = rand(N)
v = rand(N)

@constraint(model, X*u .≥ X*v)
```
The last line produces the addition warning.  The macro expansion is the following
```
quote  # /home/user/.julia/v0.5/JuMP/src/macros.jl, line 243:
[...]

PR #908, opened 2016-12-05
Fix #887

### LINKED ISSUE #887: MethodError when matrix-multiplying transpose of sparse matrix
I get the following error when attempting to matrix-multiply a constant (in this case integer) vector by the transpose of a sparse matrix of variables i.e.
```julia
x'*v
```
where `v` is a constant, `Int64`-valued vector.

```
ERROR: MethodError: no method matching Ac_mul_B!(::JuMP.GenericAffExpr{Float64,JuMP.Variable}, ::SparseMatrixCSC{JuMP.Variable,Int64}, ::Array{Int64,1}, ::JuMP.GenericAffExpr{Float64,JuMP.Variable}, ::Array{JuMP.GenericAffExpr{Float64,JuMP.Variable},1})
Closest candidates are:
  Ac_mul_B!(::Number, ::SparseMatrixCSC{Tv,Ti<:Integer}, ::Union{Base.ReshapedArray{T,1,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},Base.ReshapedArray{T,2,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray{T,1},DenseArray{T,2},SubArray{T,1,A<:Union{Base.ReshapedArray{T,N,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N}},L},SubArray{T,2,A<:Union{Base.ReshapedArray{T,N,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N}},L}}, ::Number, ::Union{Base.ReshapedArray{T,1,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},Base.ReshapedArray{T,2,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray{T,1},DenseArray{T,2},SubArray{T,1,A<:Union{Base.ReshapedArray{T,N,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N}},L},SubArray{T,2,A<:Union{Base.ReshapedArray{T,N,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N}},L}}) at sparse/linalg.jl:51
  Ac_mul_B!(::Union{Base.ReshapedArray{T,1,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},Base.ReshapedArray{T,2,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray{T,1},DenseArray{T,2},SubArray{T,1,A<:Union{Base.ReshapedArray{T,N,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N}},L},SubArray{T,2,A<:Union{Base.ReshapedArray{T,N,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N}},L}}, ::SparseMatrixCSC{Tv,Ti<:Integer}, ::Union{Base.ReshapedArray{T,1,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},Base.ReshapedArray{T,2,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray{T,1},DenseArray{T,2},SubArray{T,1,A<:Union{Base.ReshapedArray{T,N,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N}},L},SubArray{T,2,A<:Union{Base.ReshapedArray{T,N,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N}},L}}) at sparse/linalg.jl:97
  Ac_mul_B!(::AbstractArray{T,1}, ::Union{AbstractArray{T,1},AbstractArray{T,2}}, ::AbstractArray{T,1}) at linalg/matmul.jl:123
  ...
 in Ac_mul_B(::SparseMatrixCSC{JuMP.Variable,Int64}, ::Array{Int64,1}) at ./sparse/linalg.jl:85
```

This seems very strange to me because `x*w` works perfectly fine.

A workaround is available by doing 
```julia
(v'*x)'
```

ExpandingMan:
It appears that the definitions for `*` in lines 406, 407 and 408 were missing the type `Vector` in the unions of their first arguments, even though it seems that the methods for this are fully implemented.

Edit: This is a very wrong statement.  See below.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -358,10 +358,10 @@ Base.diagm(x::Vector{Variable}) = diagm(convert(Vector{AffExpr}, x))
 function _multiply!{T<:JuMPTypes}(ret::Array{T}, lhs::Array, rhs::Array)
     m, n = size(lhs,1), size(lhs,2)
     r, s = size(rhs,1), size(rhs,2)
-    for i in 1:m, j in 1:s
+    for i ∈ 1:m, j ∈ 1:s
         q = ret[i,j]
         _sizehint_expr!(q, n)
-        for k in 1:n
+        for k ∈ 1:n
             tmp = convert(T, lhs[i,k]*rhs[k,j])
             append!(q, tmp)
         end
@@ -385,11 +385,11 @@ function _multiplyt!{T<:JuMPTypes}(ret::Array{T}, lhs::Array, rhs::Array)
 end
 
 function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::SparseMatrixCSC, rhs::Array)
-    nzv = lhs.nzval
-    rv  = lhs.rowval
-    for col in 1:lhs.n
-        for k in 1:size(ret, 2)
-            for j in lhs.colptr[col]:(lhs.colptr[col+1]-1)
+    nzv = nonzeros(lhs)
+    rv  = rowvals(lhs)
+    for col ∈ 1:lhs.n
+        for k ∈ 1:size(ret, 2)
+            for j ∈ lhs.colptr[col]:(lhs.colptr[col+1]-1)
                 append!(ret[rv[j], k], nzv[j] * rhs[col,k])
             end
         end
@@ -403,14 +403,14 @@ function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lh
 end
 
 function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC)
-    rowval = rhs.rowval
-    nzval  = rhs.nzval
+    rowval = rowvals(rhs)
+    nzval  = nonzeros(rhs)
     for multivec_row in 1:size(lhs,1)
-        for col in 1:rhs.n
+        for col ∈ 1:rhs.n
             idxset = rhs.colptr[col]:(rhs.colptr[col+1]-1)
             q = ret[multivec_row, col]
             _sizehint_expr!(q, length(idxset))
-            for k in idxset
+            for k ∈ idxset
                 append!(q, lhs[multivec_row, rowval[k]] * nzval[k])
             end
         end
@@ -420,8 +420,8 @@ end
 
 # this computes lhs.'*rhs and places it in ret
 function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC)
-    rowval = rhs.rowval
-    nzval  = rhs.nzval
+    rowval = rowvals(rhs)
+    nzval  = nonzeros(rhs)
     for multivec_row ∈ 1:size(lhs,2) # transpose
         for col ∈ 1:rhs.n
             idxset = rhs.colptr[col]:(rhs.colptr[col+1]-1)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -451,11 +451,11 @@ import Base.Ac_mul_B
 # these methods are called when one does A.'*v or A'*v respectively 
 At_mul_B{T<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix, Vector, SparseMatrixCSC}) = _matmult(A, x)
 At_mul_B{T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R}, Vector{R}, SparseMatrixCSC{R}}) = _matmult(A, x)
-At_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, X)
+At_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, x)
 # these methods are the same as above since complex numbers are not implemented in JuMP
 Ac_mul_B{T<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix, Vector, SparseMatrixCSC}) = _matmult(A, x)
 Ac_mul_B{T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R}, Vector{R}, SparseMatrixCSC{R}}) = _matmult(A, x)
-Ac_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, X)
+Ac_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, x)
 
 function _matmul(A, x)
     m, n = size(A,1), size(A,2)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -399,7 +399,7 @@ end
 
 # this computes lhs.'*rhs and places it in ret
 function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::SparseMatrixCSC, rhs::Array)
-    _multiply!(ret, transpose(lhs), rhs) # for sparse matrices this should be cheap
+    _multiply!(ret, transpose(lhs), rhs) # TODO fully implement
 end
 
 function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -403,9 +403,9 @@ _multiply!{T<:JuMPTypes}(ret::AbstractArray{T}, lhs::SparseMatrixCSC, rhs::Spars
 
 _multiply!(ret, lhs, rhs) = A_mul_B!(ret, lhs, ret)
 
-(*){T<:JuMPTypes}(             A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix,   Vector,   SparseMatrixCSC})    = _matmul(A, x)
-(*){T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R},Vector{R},SparseMatrixCSC{R}}) = _matmul(A, x)
-(*){T<:JuMPTypes}(             A::Union{Matrix,   SparseMatrixCSC},    x::Union{Matrix{T},Vector{T},SparseMatrixCSC{T}}) = _matmul(A, x)
+(*){T<:JuMPTypes}(             A::Union{Matrix{T},Vector{T},SparseMatrixCSC{T}}, x::Union{Matrix,   Vector,   SparseMatrixCSC})    = _matmul(A, x)
+(*){T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},Vector{T},SparseMatrixCSC{T}}, x::Union{Matrix{R},Vector{R},SparseMatrixCSC{R}}) = _matmul(A, x)
+(*){T<:JuMPTypes}(             A::Union{Matrix,   Vector,   SparseMatrixCSC},    x::Union{Matrix{T},Vector{T},SparseMatrixCSC{T}}) = _matmul(A, x)
 
 function _matmul(A, x)
     m, n = size(A,1), size(A,2)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -448,11 +448,11 @@ _multiply!(ret, lhs, rhs) = A_mul_B!(ret, lhs, ret)
 
 import Base.At_mul_B
 import Base.Ac_mul_B
-# confusingly, these methods are called when one does A'*v or A.'*v respectively
+# these methods are called when one does A.'*v or A'*v respectively
 At_mul_B{T<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix, Vector, SparseMatrixCSC}) = _matmult(A, x)
 At_mul_B{T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R}, Vector{R}, SparseMatrixCSC{R}}) = _matmult(A, x)
 At_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, X)
-# these methods are the same as a bove since complex numbers are not implemented in JuMP
+# these methods are the same as above since complex numbers are not implemented in JuMP
 Ac_mul_B{T<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix, Vector, SparseMatrixCSC}) = _matmult(A, x)
 Ac_mul_B{T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R}, Vector{R}, SparseMatrixCSC{R}}) = _matmult(A, x)
 Ac_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, X)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -369,7 +369,7 @@ function _multiply!{T<:JuMPTypes}(ret::Array{T}, lhs::Array, rhs::Array)
     ret
 end
 
-# this for transposed multiplication
+# this computes lhs.'*rhs and places it in ret
 function _multiplyt!{T<:JuMPTypes}(ret::Array{T}, lhs::Array, rhs::Array)
     m, n = size(lhs,2), size(lhs,1) # transpose
     r, s = size(rhs,1), size(rhs,2)
@@ -397,7 +397,7 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
     ret
 end
 
-# this for transposed multiplication
+# this computes lhs.'*rhs and places it in ret
 function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::SparseMatrixCSC, rhs::Array)
     _multiply!(ret, transpose(lhs), rhs) # for sparse matrices this should be cheap
 end
@@ -418,7 +418,7 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
     ret
 end
 
-# for transposed multiplication
+# this computes lhs.'*rhs and places it in ret
 function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC)
     rowval = rhs.rowval
     nzval  = rhs.nzval
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -389,7 +389,7 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
     rv  = rowvals(lhs)
     for col ∈ 1:lhs.n
         for k ∈ 1:size(ret, 2)
-            for j ∈ lhs.colptr[col]:(lhs.colptr[col+1]-1)
+            for j ∈ nzrange(lhs, col)
                 append!(ret[rv[j], k], nzv[j] * rhs[col,k])
             end
         end
@@ -407,7 +407,7 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
     nzval  = nonzeros(rhs)
     for multivec_row in 1:size(lhs,1)
         for col ∈ 1:rhs.n
-            idxset = rhs.colptr[col]:(rhs.colptr[col+1]-1)
+            idxset = nzrange(rhs, col)
             q = ret[multivec_row, col]
             _sizehint_expr!(q, length(idxset))
             for k ∈ idxset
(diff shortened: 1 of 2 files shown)


mlubin:
Could you add unit test for the failure in #887?

ExpandingMan:
Sure.  Bear with me a bit, I'm actually pretty new to contributing to existing projects that are not my own, still figuring out the process...

ExpandingMan:
Where's the appropriate place to create the unit test?  In the test directory of my fork?  Do you guys have any kind of standardized way of writing them?

mlubin:
I would add the relevant tests under this block: https://github.com/JuliaOpt/JuMP.jl/blob/b1a7bda717a0d8fae5a9f8169cf3495f8bfb6fa5/test/operator.jl#L521
If you push to the ``patch-1`` branch on your fork the changes will show up under this PR.

ExpandingMan:
Ok, I added a few more lines to the unit test.  There are several errors, but all of them existed before my new change, they just weren't caught in the unit tests.  Multiplying sparse matrices of floats by sparse matrices of `Variable` still isn't implemented.

The really weird error is that `X'*A` isn't working (throws "No method matching `Ac_mul_B!`), where `X` is a sparse matrix of variables but `A` is a sparse matrix of floats.  This weird because `A'*X` and `X*A` work fine.  I'm wondering if there would be anything wrong with defining the operators such that `X'*A` is defined as `(A'*X)'`.

Anyway, I haven't found any errors with defining multiplication of sparse matrices of variables by vectors, so I think this pull is ok...

test/operator.jl
```diff
@@ -586,6 +590,25 @@ const sub2 = JuMP.repl[:sub2]
...
+                            X11   X23  0
+                            0     2X23 0])
+        @test vec_eq(X'*A, [2X11  X11  0
+                            0     0    0
+                            X23   2X23 X23])
+        @test vec_eq(X*A, X*B)
+        @test vec_eq(A*X, B*X)
+        @test vec_eq(A*X', B*X')
+        # TODO this is known to fail currently
+        @test vec_eq(X'*A, X'*B)
```
mlubin: Use ``@test_broken`` if it's known to fail

ExpandingMan: Done.  There is an error in the concatenation section, but I don't think it's new.

mlubin:
Tests are failing: https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/181483226#L319

ExpandingMan:
This was not the issue. 

I had forgotten that, confusingly, Julia makes transposed vectors `Matrix`, which is why `Vector` was not defined in those expressions.  Still looking into it now...

ExpandingMan:
Wow... after a great deal of confusion, I have tracked down the issue, and it is *really* strange.  See the following code
```julia
import Base.*

function (*)(A::SparseMatrixCSC{Int32}, B::Vector) 
    println("being called!")
end

A = spzeros(Int64, 3, 3)
A[1,1] = 2
A[3,3] = 4

B = spzeros(Int32, 3, 3)
B[1,1] = 2
B[3,3] = 4

v = rand(0:99, 3)

A*v == A'*v  # evaluates to true
B*v == B'*v  # evaluates to false, rhs never gets overloaded!
typeof(B) == typeof(B') # even though this evaluates to true!

transpose(B)*v # however this uses the new method defined above
```
Somehow, even though the multiplication of a vector by a transposed sparse matrix involves *exactly* the same types as the multiplication of a vector by a non-transposed sparse matrix, a completely different method gets called!  

The reason this happens is because if you do `A'*v` it **does not call `*`**, no, no, no... instead it calls `Ac_mul_B`, a completely different function.  Apparently unlike ordinary functions, operator symbols like `*` can act as aliases for multiple different functions depending on the context.

Part of the problem is that I've been using the adjoint instead of `transpose`, but since that is a perfectly legitimate thing to do, it should still be fixed (besides, nobody wants to write `transpose(A)`).

I should be able to fix this now by creating a new `Ac_mul_B` method.  Give me a bit to test it...

blegat:
That is indeed surprising, I guess this is to avoid applying the transpose operator.
Note that `A'` is the conjugate transpose (`ctranspose`, `Ac_mul_B`) while `A.'` is the transpose (`transpose`, `At_mul_B`) so it might also make sense to define `At_mul_B` which should be equivalent to `Ac_mul_B` as long as no complex numbers are involved.

ExpandingMan:
Yes, but the very strange thing is that `transpose(A)*v` seems to call `*` (still investigating).  I didn't realize until now that `A.'` gave the transpose.  I think since JuMP doesn't have complex numbers implemented it makes sense to define `Ac_mul_B`.  Besides, it is frequently the case that the transpose operator on real matrices generalizes to the adjoint operator on complex matrices.

I might raise an issue about this in Julia main, at the very least it should be more well documented.

ExpandingMan:
Ok... I have now definitely fixed the real problem.  I had to make the following additions to operators.jl
* There are now `At_mul_B` and `Ac_mul_B` methods for each of the `*` matrix multiplication methods.
* Those call a new method `_matmult` which is equivalent to `_matmul` but for the first matrix transposed.
* That in turn calls the new methods `_return_arrayt` and `_multiplyt!`.
* `_return_arrayt` gets the proper return array for multiplication with the first matrix transposed.
* The `_multiplyt!` methods are the equivalent to `_multiply!` methods except that the first matrix is transposed.  They should be as efficient as `_multiply!` (i.e. `transpose` is only called for a `SparseMatrixCSC`).
* The sparse-matrix, sparse-matrix `_multiplyt!` implementation is every bit as bad as the one the existing `_multiply!` one, at some point we need to add efficient sparse-matrix multiplication.

There are now two broken unit tests: but I have confirmed that they were broken prior to my changes.  Multiplying a transposed sparse matrix of variables by another matrix still doesn't work.  I believe this can be fixed by adding `A_mul_Bc` and `A_mul_Bt` methods, but I didn't want to do that yet as there is still a lot of work to be done on sparse matrices in general.

There is still #907, which I don't completely understand.  All matrix multiplications call `append!` so that particular warning should not pop up.  We should also try to understand why it is so damn slow.

src/operators.jl
```diff
@@ -398,15 +418,45 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
...
 
 _multiply!(ret, lhs, rhs) = A_mul_B!(ret, lhs, ret)
 
 (*){T<:JuMPTypes}(             A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix,   Vector,   SparseMatrixCSC})    = _matmul(A, x)
 (*){T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R},Vector{R},SparseMatrixCSC{R}}) = _matmul(A, x)
 (*){T<:JuMPTypes}(             A::Union{Matrix,   SparseMatrixCSC},    x::Union{Matrix{T},Vector{T},SparseMatrixCSC{T}}) = _matmul(A, x)
 
+import Base.At_mul_B
+import Base.Ac_mul_B
+# confusingly, these methods are called when one does A'*v or A.'*v respectively
```
mlubin: Let's drop "confusingly" here and keep the comment strictly factual.

src/operators.jl
```diff
@@ -398,15 +418,45 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
...
+At_mul_B{T<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix, Vector, SparseMatrixCSC}) = _matmult(A, x)
+At_mul_B{T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R}, Vector{R}, SparseMatrixCSC{R}}) = _matmult(A, x)
+At_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, X)
+# these methods are the same as a bove since complex numbers are not implemented in JuMP
```
mlubin: "a bove" should be "above"

test/operator.jl
```diff
@@ -586,6 +590,31 @@ const sub2 = JuMP.repl[:sub2]
...
+                            0     2X23 0])
+        @test vec_eq(X'*A, [2X11  X11  0
+                            0     0    0
+                            X23   2X23 X23])
+        @test vec_eq(X.'*A, [2X11 X11  0 
+                             0    0    0
+                             X23  2X23 X23])
+        @test vec_eq(X.'*A, X'*A)
+        @test vec_eq(X*A, X*B)
+        @test_broken vec_eq(A*X, B*X)
```
mlubin: Can you comment these out for now? They're causing travis to report failures.

mlubin:
@joehuchette, could you review?

mlubin:
@ExpandingMan, your commits are not properly linked to your github account.

ExpandingMan:
Hmm... I have no idea why that would be.  Any idea how I fix that?

mlubin:
https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

ExpandingMan:
Made a trivial commit (deleted space in comment) to fix user email.

mlubin:
You can rebase the commits to clean them all up

ExpandingMan:
I can rebase the commits locally, but for some reason I am unable to push them...

blegat:
You need to do `git push -f ...` when you rebase. This is because in addition to adding new commits you also modify commits already pushed to you need to force the push hence the `-f`.

ExpandingMan:
I did that... how do I change only the user name when calling `git rebase`?  I am brought to the script but I can only do things like squash or reword, I think my change wasn't doing anything.  Is there a command for changing the user?  

I apologize for my complete inexperience with this.

blegat:
You will find the answer [here](http://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit). No problem ;)

ExpandingMan:
Ok, got it, thanks for your help.

src/operators.jl
```diff
@@ -398,15 +418,45 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
...
 (*){T<:JuMPTypes}(             A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix,   Vector,   SparseMatrixCSC})    = _matmul(A, x)
 (*){T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R},Vector{R},SparseMatrixCSC{R}}) = _matmul(A, x)
 (*){T<:JuMPTypes}(             A::Union{Matrix,   SparseMatrixCSC},    x::Union{Matrix{T},Vector{T},SparseMatrixCSC{T}}) = _matmul(A, x)
 
+import Base.At_mul_B
+import Base.Ac_mul_B
+# these methods are called when one does A.'*v or A'*v respectively 
+At_mul_B{T<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix, Vector, SparseMatrixCSC}) = _matmult(A, x)
+At_mul_B{T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R}, Vector{R}, SparseMatrixCSC{R}}) = _matmult(A, x)
+At_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, X)
```
joehuchette: ``_matmult(A, x)``

src/operators.jl
```diff
@@ -398,15 +418,45 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
...
+# these methods are the same as above since complex numbers are not implemented in JuMP
+Ac_mul_B{T<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix, Vector, SparseMatrixCSC}) = _matmult(A, x)
+Ac_mul_B{T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix{R}, Vector{R}, SparseMatrixCSC{R}}) = _matmult(A, x)
+Ac_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC}, x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, X)
```
joehuchette: ``_matmult(A, x)``

src/operators.jl
ExpandingMan: not sure what you're asking here...

ExpandingMan:
Currently working on adding tests that should cover the rest of the functions.

src/operators.jl
joehuchette: The argument to the function is ``x`` (lowercase), while the argument you pass to ``_matmult`` is ``X`` (uppercase).

mlubin:
The new code should be 100% covered by unit tests to catch issues like this.

On Dec 7, 2016 10:30, "ExpandingMan" <notifications@github.com> wrote:

*@ExpandingMan* commented on this pull request.
------------------------------

In src/operators.jl <https://github.com/JuliaOpt/JuMP.jl/pull/908>:

>
 _multiply!(ret, lhs, rhs) = A_mul_B!(ret, lhs, ret)

 (*){T<:JuMPTypes}(
A::Union{Matrix{T},SparseMatrixCSC{T}}, x::Union{Matrix,   Vector,
SparseMatrixCSC})    = _matmul(A, x)
 (*){T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}},
x::Union{Matrix{R},Vector{R},SparseMatrixCSC{R}}) = _matmul(A, x)
 (*){T<:JuMPTypes}(             A::Union{Matrix,   SparseMatrixCSC},
 x::Union{Matrix{T},Vector{T},SparseMatrixCSC{T}}) = _matmul(A, x)

+import Base.At_mul_B
+import Base.Ac_mul_B
+# these methods are called when one does A.'*v or A'*v respectively
+At_mul_B{T<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}},
x::Union{Matrix, Vector, SparseMatrixCSC}) = _matmult(A, x)
+At_mul_B{T<:JuMPTypes,R<:JuMPTypes}(A::Union{Matrix{T},SparseMatrixCSC{T}},
x::Union{Matrix{R}, Vector{R}, SparseMatrixCSC{R}}) = _matmult(A, x)
+At_mul_B{T<:JuMPTypes}(A::Union{Matrix,SparseMatrixCSC},
x::Union{Matrix{T}, Vector{T}, SparseMatrixCSC{T}}) = _matmult(A, X)

Ah, good catch, thanks!

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/JuliaOpt/JuMP.jl/pull/908>, or mute the thread

ExpandingMan:
How do I trigger the codecov to run again?

joehuchette:
Let's see if that works. Otherwise, you can run the [code coverage locally](https://github.com/JuliaCI/Coverage.jl).

ExpandingMan:
Ok, looking good, all of my code seems to be covered and tests are passing.  I'm still a bit confused about how I publish the coverage report though... It seems very strange to me that one of us can't just trigger it to run again.

src/operators.jl
```diff
@@ -369,6 +369,21 @@ function _multiply!{T<:JuMPTypes}(ret::Array{T}, lhs::Array, rhs::Array)
     ret
 end
 
+# this for transposed multiplication
```
mlubin: Please expand on the comment. Does this compute ``lhs'*rhs``?

src/operators.jl
```diff
@@ -382,6 +397,11 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
     ret
 end
 
+# this computes lhs.'*rhs and places it in ret
+function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::SparseMatrixCSC, rhs::Array)
+    _multiply!(ret, transpose(lhs), rhs) # for sparse matrices this should be cheap
```
mlubin: I'm not sure if i'd call transposing a sparse matrix cheap, it actually could be relatively slow. Maybe leave a smarter implementation as a TODO?

ExpandingMan: Funny you should mention that, I was literally just this second about to make a post about it.  Part of the reason I didn't actually test how expensive this was is because sparse matrix multiplication is currently broken (i.e. way slower than it should be), so this will all have to change anyway.  See issue #907 .

ExpandingMan:
Hello all.  I'm pretty sure my code for `_multiplyt!` on line 422 which is based on `_multiply!` on line 405 is correct, but it uses some really bizarre variable names (`multivec_row`?) which right now is making me paranoid.  If somebody can look at it and check me carefully, I would be appreciated, although it does seem to be producing the correct result.

test/operator.jl
```diff
@@ -586,6 +599,42 @@ const sub2 = JuMP.repl[:sub2]
...
+        @test vec_eq(Y'*A, Y.'*A)
+        @test vec_eq(A*Y', A*Y.')
+        @test vec_eq(Z'*A, Z.'*A)
+        @test vec_eq(Xd'*Y, Xd.'*Y)
+        @test vec_eq(Y'*Xd, Y.'*Xd)
+        @test vec_eq(Xd'*Xd, Xd.'*Xd)
+        # @test_broken vec_eq(A*X, B*X)
+        # @test_broken vec_eq(A*X', B*X')
+        @test vec_eq(X'*A, X'*B)
+        # @test_broken(X'*X, X.'*X) # sparse quadratic known to be broken
```
mlubin: Why is this known to be broken? If it's known could you write down a reference?

ExpandingMan: It is part of a plethora of existing problems with sparse-matrix multiplication, which really needs to be reimplemented (see #907, which still has me confused).  I haven't made an issue for this particular problem.  Give me a few minutes to go back to master and check the error it throws, and I'll create a new issue for it.

Sorry my comments aren't better, I'm a physicist rather than a software engineer by training, so even this is one of the most formal code-writing processes I've ever been through!  I'll try to keep things a little clearer.

mlubin: Don't worry, you're doing a good job. None of us are software engineers :)

src/operators.jl
```diff
@@ -398,15 +418,45 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
     ret
 end
 
+# this computes lhs.'*rhs and places it in ret
+function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC)
+    rowval = rhs.rowval
+    nzval  = rhs.nzval
+    for multivec_row ∈ 1:size(lhs,2) # transpose
+        for col ∈ 1:rhs.n
+            idxset = rhs.colptr[col]:(rhs.colptr[col+1]-1)
```
mlubin: It's better to use ``nzrange`` here

src/operators.jl
```diff
@@ -398,15 +418,45 @@ function _multiply!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs
     ret
 end
 
+# this computes lhs.'*rhs and places it in ret
+function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC)
+    rowval = rhs.rowval
+    nzval  = rhs.nzval
```
mlubin: and ``nonzeros(rhs)``

src/operators.jl
mlubin: If you could update ``_multiply!`` above for consistency then I think we're all set

src/operators.jl
```diff
@@ -358,55 +358,105 @@ Base.diagm(x::Vector{Variable}) = diagm(convert(Vector{AffExpr}, x))
...
     ret
 end
 
+# this computes lhs.'*rhs and places it in ret
+function _multiplyt!{T<:Union{GenericAffExpr,GenericQuadExpr}}(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC)
+    rowval = rowvals(rhs)
+    nzval  = nonzeros(rhs)
+    for multivec_row ∈ 1:size(lhs,2) # transpose
+        for col ∈ 1:rhs.n
+            idxset = rhs.colptr[col]:(rhs.colptr[col+1]-1)
```
mlubin: ``nzrange(rhs,col)`` should replace ``rhs.colptr[col]:(rhs.colptr[col+1]-1)``

mlubin:
@ExpandingMan @joehuchette, does this PR also fix https://github.com/JuliaOpt/JuMP.jl/issues/600?

joehuchette:
It should. 






On Thu, Dec 8, 2016 at 10:20 PM -0500, "Miles Lubin" <notifications@github.com> wrote:












@ExpandingMan @joehuchette, does this PR also fix #600?

ExpandingMan:
Yes, it fixes #600, but bear in mind, there are a lot of methods that ultimately need to get overloaded.
```julia
*         # "A_mul_B", done
At_mul_B  # done in this patch
A_mul_Bt
At_mul_Bt

Ac_mul_B  # done in this patch
A_mul_Bc
Ac_mul_Bc
```
But really, since there is no implementation of complex numbers in JuMP, it's only two more because the adjoint is equivalent to the transpose for all real matrices.

Again, I didn't bother to fix the others yet because there is still something quite wrong with the sparse matrix multiplication ( #907), so I didn't want to go too crazy until that is completely figured out.

It would be really, really nice if someone would write a package with macros for creating all this "boiler-plate" linear algebra code for custom types.  It's a little complicated, but I definitely think it's doable.  I might look into something like that once I figure out why the sparse matrix multiplication is slow.

joehuchette:
This looks great, and ready to go from my end. Thanks for your hard work, @ExpandingMan!

ExpandingMan:
Awesome!  Thanks all, glad to have contributed.

PR #910, opened 2016-12-06
preallocate reusable GradientConfig for AD of user-defined function

jrevels:
Note that this will always use `Float64` for the element type unless you give it a different type - I added this capability to the function signature but haven't touched any callers. Let me know if I should run any perf test here.

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -1288,8 +1288,9 @@ function MathProgBase.eval_grad_f(d::UserFunctionEvaluator,grad,x)
     nothing
 end
 
-function UserAutoDiffEvaluator(dimension::Integer, f::Function)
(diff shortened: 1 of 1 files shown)


jrevels:
Looks like one of the Mac builds timed out? Looking through the build history, it seems like that might not be too uncommon for JuMP?

mlubin:
Nightly is a mess, I've only been looking at 0.5 recently.

jrevels:
If it's helpful, you could [do something like this](https://github.com/JuliaDiff/ReverseDiff.jl/commit/31dd04d8998781c485b749d78a8b2061cb690e74) in your Travis settings. It'll still attempt to run that specific build, but it won't count against the overall CI pass/fail.

PR #920, opened 2016-12-12
Fix 907

### LINKED ISSUE #907: addition operator warning during matrix operations

ExpandingMan:
This fixes #907 in what I think is the least disruptive way possible.  This also fixes warnings arising from expressions like `X*u + y` that appear as arguments to constraint macros.  Unit tests have been added to cover the 3 new lines of code.

In my opinion we should also reconsider whether the warning
```
WARNING: The addition operator has been used on JuMP expressions a large number of times. This warning is safe to ignore but may indicate that model generation is slower than necessary. For performance reasons, you should not add expressions in a loop. Instead of x += y, use append!(x,y) to modify x in place. If y is a single variable, you may also use push!(x, coef, y) in place of x += coef*y.
```
should be removed altogether (since there isn't much of a performance detriment, since all users should use `+` 100% of the time and the only alternative `append!` is actually a slightly different use case).  However, it's hard to get the warning to appear outside of macros (for reasons I don't entirely understand) so with this fix it's unlikely to ever occur unless the user does something strange.  Alternatively, the default behavior of `+` can be changed.

### DIFF
--- a/src/parseExpr_staged.jl
+++ b/src/parseExpr_staged.jl
@@ -268,7 +268,11 @@ for T1 in (GenericAffExpr,GenericQuadExpr), T2 in (Number,Variable,GenericAffExp
     @eval addtoexpr(::$T1, ::_NLExpr, ::$T2) = _nlexprerr()
 end
 
-addtoexpr(ex, c, x) = ex + c*x
+addtoexpr{T<:GenericAffExpr}(ex::Array{T}, c::AbstractArray, x::AbstractArray) = append!.(ex, c*x)
+addtoexpr{T<:GenericAffExpr}(ex::Array{T}, c::AbstractArray, x::Number) = append!.(ex, c*x)
+addtoexpr{T<:GenericAffExpr}(ex::Array{T}, c::Number, x::AbstractArray) = append!.(ex, c*x)
+
+addtoexpr(ex, c, x) = ex + c*x  
 
 @generated addtoexpr_reorder(ex, arg) = :(addtoexpr(ex, 1.0, arg))
 
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -688,12 +688,12 @@ const sub2 = JuMP.repl[:sub2]
         u = [2, 3]
         v = [4, 5]
         expr_base = zero(JuMP.AffExpr)
-        @constraint(m, x*u .≤ x*v)
-        @test string(m.linconstr[1]) == "-2 x[1,1] - 2 x[1,2] ≤ 0"
-        @test string(m.linconstr[2]) == "-2 x[2,1] - 2 x[2,2] ≤ 0"
-        @constraint(m, x*u + y .≤ v)
-        @test string(m.linconstr[3]) == "2 x[1,1] + 3 x[1,2] + y[1] ≤ 4"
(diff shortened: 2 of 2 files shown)


joehuchette:
This seems reasonable to me. Out of curiosity, how are you verifying the new methods are being used in your tests?

The warning only appears after a global counter of arithmetic operations passes some (relatively large) [threshold](https://github.com/JuliaOpt/JuMP.jl/blob/3dd548ef34cee125c668425f188b86fb677f83af/src/JuMP.jl#L816-L818). This is probably why you only see the warning in the macros, where it's easy to exceed this threshold without knowing, because of slow fallbacks to ``+`` like you address here. The warning was added to nudge those using ``+`` in a hot loop to use the much more efficient ``append!`` only in this case. We can add new methods like this to avoid spurious warnings, but I think we'd like to keep the warning around for those cases for which it's intended.

ExpandingMan:
I verified the code coverage the old-fashioned way; I first declared those functions with a `println` in them so I knew when they were called.  I'm sure once the code-coverage report is generated it will confirm this.

Loops are sort of an anathema to what I'd think of as the design philosophy of something like JuMP anyway ("do everything pseudo-algebraically"), so I have no strong feelings about the warning as long as it is not popping up in any normal expressions, which I believe it no longer is.

joehuchette:
> I'm sure once the code-coverage report is generated it will confirm this.

Good point, forgot about this.

> Loops are sort of an anathema to what I'd think of as the design philosophy of something like JuMP anyway ("do everything pseudo-algebraically")

I tend to agree, but people tend to write code this way (and it's even useful every once in a while), hence the warning :)

ExpandingMan:
I'm quite confused as to why the code coverage report is showing 6 new lines with 3 missed.  I've clearly only added 3 lines (as can be seen in the Github diff) and they are all shown as covered in codecov.

ExpandingMan:
Ok, now I really don't understand why the windows builds are failing.  Perhaps it's something unrelated in the current master that fails.

joehuchette:
Appveyor failure looks unrelated (something with Ipopt).

joehuchette:
Looks good to me. @mlubin?

mlubin:
Looks good, thanks for the minimal fix!

joehuchette:
Thanks, @ExpandingMan!

ExpandingMan:
You're quite welcome.  I'll just be glad once all of the linear algebra feels completely tight and fixed :)

PR #921, opened 2016-12-13
Add support for singleton anonymous at-expression

joehuchette:
@mlubin was there any reason we didn't have this before?


mlubin:
Yes, because we were deprecating the old syntax. Now this is safe to add.

joehuchette:
K, that's what I thought.

PR #924, opened 2016-12-19
switch from throw(CallbackAbort()) to return CallbackAbort

mlubin:
Pajarito will be using ``CallbackAbort()`` as part of the normal flow of the algorithm. ``CallbackAbort()`` has been segfaulting on OS X with Gurobi for a long time (https://github.com/JuliaOpt/Gurobi.jl/issues/47). Doesn't look like any fix for Julia is incoming, so the workaround is to just not throw exceptions in callbacks.

CC @chriscoey @JackDunnNZ

### DIFF
--- a/doc/callbacks.rst
+++ b/doc/callbacks.rst
@@ -423,8 +423,8 @@ This code can also be found in ``/JuMP/examples/simplelazy2.jl``.
 Exiting a callback early
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-If you need to exit the optimization process earlier than a solver otherwise would, it is possible to throw a ``CallbackAbort`` exception in callback code::
+If you need to exit the optimization process earlier than a solver otherwise would, it is possible to return ``JuMP.StopTheSolver`` from the callback code::
 
-    throw(CallbackAbort())
+    return JuMP.StopTheSolver
 
 This will trigger the solver to exit immediately and return a ``:UserLimit`` status.
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -21,6 +21,8 @@ type InfoCallback <: JuMPCallback
     when::Symbol
 end
 
+type StopTheSolver end
+
 type CallbackAbort <: Exception end
 export CallbackAbort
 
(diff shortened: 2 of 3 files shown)


chriscoey:
Yes, the segfaulting has made it tough to do debugging especially.

mlubin:
Will merge today barring objections from @joehuchette

joehuchette:
Why make it an exception if we're not going to throw it? Might as well just be symbol at this point.

mlubin:
A bit of type safety never hurt. We can remove the inheritance from ``Exception`` after the deprecation period.

joehuchette:
> A bit of type safety never hurt.
I'm not sure what exactly it's buying us here. I'd be more convinced if this the change was to return a new object, rather than this weird expection -> deprecated exception -> custom type route.

mlubin:
What new object would you like? What would it be called besides ``CallbackAbort``?

joehuchette:
:CallbackAbort?



		_____________________________

mlubin:
Symbols are more error prone than custom types. I imagine tracking down ``:CallbackAbrot`` wouldn't be fun. What we can do is make things simpler and have ``return CallbackAbort`` instead of ``return CallbackAbort()``.

mlubin:
Made the change from ``return CallbackAbort()`` to ``return CallbackAbort``.

joehuchette:
I don't get the aversion to symbols; we use them in our API elsewhere. If you return ``:CallbackAbrot`` from a callback, we can have a handler that prints the value, and then you just search in your code. It seems much more straightforward than telling users to return the type (not instance!) of an exception...

mlubin:
> I don't get the aversion to symbols; we use them in our API elsewhere.

We should be moving away from symbols when reasonable. In particular whenever we redo the MPB return statuses we won't be using symbols.

> we can have a handler

What does "we can have" mean? That we're recommending someone should go in and modify the JuMP source code as part of debugging their model or that I need to write extra handling code in this PR to check for typos? Neither seems appropriate when the alternative is just letting the compiler do the job of catching these errors.

> It seems much more straightforward than telling users to return the type (not instance!) of an exception...

I don't see what's not straightforward about telling people to write ``return CallbackAbort``. That it's a type that happens to inherit from Exception is irrelevant as far as the complexity that the user sees. It could also just be an enum or named constant.

mlubin:
@joehuchette, @chriscoey and I need this to move forward. Okay to merge?

joehuchette:
> It could also just be an enum or named constant.

Why not just use an enum or named constant then? 

I don't really care that ``CallbackAbort`` is subtyping Expection, I'm just not sure why we're deprecating throwing _instances_ of a type, but still keeping around the type for some reason. Change the code to use a named constant, enum, singleton Exception type named ``AbortCallback`` or something, I just don't understand this half deprecation.

mlubin:
@joehuchette, the half deprecation is just because the name ``CallbackAbort`` already seemed fine, but I can call it ``JuMP.ExitFromCallback`` if we want to change things up.

mlubin:
Here's another attempt. Now it's ``return JuMP.StopTheSolver``. I would like to release 0.15 ASAP, so any bikeshedding needs to happen soon.

joehuchette:
Hah, like the name. Looks good to me.

Issue #919, opened 2016-12-12
'fixate' function for convenience


martinbiel:
In some problem formulations, for example MPC problems, it would be convenient to be able to specify the state variables as 
`@variable(m, x[1:5])`
and then fix `x[1,1]` to some given initial value. Currently, it is achievable through `setlowerbound(x[1,1],x_init)` and `setupperbound(x[1,1],x_init)`, but I also found that by allowing `:Fixed` in `setcategory` I was able to do
```
@variable(m, x[1:5])
setcategory(x[1,1],:Fixed)
setvalue(x[1,1],1)
```
which on `print` yields
```
Subject to
 x[1] = 1
 x[2] free
 x[3] free
 x[4] free
 x[5] free
```
Would it be possible to allow for this behavior for convenience and readability? for example through some `fixate` function:
```
function fixate(v::Variable, val::Number)
    v.m.colVal[v.col] = val
    v.m.colCat[v.col] = :Fixed
    v.m.colLower[v.col] = val
    v.m.colUpper[v.col] = val
end
```

joehuchette:
I'm not opposed to this, though I'm not crazy about the name. @mlubin?

mlubin:
I'm open to a PR. Maybe ``JuMP.fix``?

blegat:
Would it make more sense as a macro or as a function ?

mlubin:
I don't see much value in the additional complexity of using a macro.

joaquimg:
Has anybody done something? If not, I volunteer... the function was even written in this post...

joehuchette:
Go for it!






On Tue, Dec 20, 2016 at 12:11 PM -0500, "Joaquim Dias Garcia" <notifications@github.com> wrote:












Has anybody done something? If not, I volunteer... the function was even written in this post...

blegat:
@joaquimg I was planning to do it but I didn't find the time :( Thanks !

mlubin:
Fixed by https://github.com/JuliaOpt/JuMP.jl/pull/929. Thanks @joaquimg!

PR #929, opened 2016-12-20
Add fix function for #919

joaquimg:
fix candidate for #919

### DIFF
--- a/doc/refvariable.rst
+++ b/doc/refvariable.rst
@@ -220,8 +220,7 @@ Two possible uses for fixed variables are:
 
 2. For solving a sequence of problems with varying parameters.
    One may call ``JuMP.fix(x, val)``
-   to change the value to which the variable is fixed. For LPs
+   to change the value of a fixed variable or to fix a 
+   previously unfixed variable. For LPs
    in particular, most solvers are able to efficiently hot-start when
    solving the resulting modified problem.
-   
-Note that ``setvalue(x, val)`` will not set initial value for fixed variables, it will only set a warmstart for non-fixed variables. One can fix existing, non-fixed variables with ``JuMP.fix(x, val)`` which sets both bounds and initial value of ``x`` to ``val``.
--- a/doc/refvariable.rst
+++ b/doc/refvariable.rst
@@ -184,7 +184,7 @@ Methods
 * ``getvalue(x)`` - Get the value of this variable in the solution. If ``x`` is a single variable, this will simply return a number.
   If ``x`` is indexable then it will return an indexable dictionary of values. When the model is unbounded, ``getvalue`` will
   instead return the corresponding components of an unbounded ray, if available from the solver.
-* ``setvalue(x,v)`` - Provide an initial value ``v`` for this variable that can be used by supporting MILP solvers. If ``v`` is ``NaN``, the solver may attempt to fill in this value to construct a feasible solution.
+* ``setvalue(x,v)`` - Provide an initial value ``v`` for this variable that can be used by supporting MILP solvers. If ``v`` is ``NaN``, the solver may attempt to fill in this value to construct a feasible solution.  ``setvalue`` is not supported for fixed variables, their initial value is set with ``JuMP.fix(x,v)`` together with their bounds.
 * ``getdual(x)`` - Get the reduced cost of this variable in the solution. Similar behavior to ``getvalue`` for indexable variables.
 
 .. note::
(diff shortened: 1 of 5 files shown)


src/JuMP.jl
```diff
@@ -391,6 +391,14 @@ function setvalue(v::Variable, val::Number)
     end
 end
 
+# Fix a variable that was not previously fixed
+function fix(v::Variable, val::Number)
+    v.m.colVal[v.col] = val
```
mlubin: Could ``setvalue`` and ``setcategory`` instead of the internal methods?

mlubin:
Needs tests and docs.

joaquimg:
This one runs smoothly:
```function fix(v::Variable, val::Number)
    v.m.colCat[v.col] = :Fixed
    setvalue(v, val)
end```
While this one does not work:
```function fix(v::Variable, val::Number)
    setcategory(v, :Fixed)
    setvalue(v, val)
end```
Due to:
```
const var_cats = [:Cont, :Int, :Bin, :SemiCont, :SemiInt]
function setcategory(v::Variable, cat::Symbol)
    cat in var_cats || error("Unrecognized variable category $cat. Should be one of:\n    $var_cats")
    v.m.colCat[v.col] = cat
end
```
Clearly ``:Fixed`` is not allowed...

**Comment by mlubin** [comment]
Okay, you could argue that using ``setcategory`` is not appropriate for fixed variables because you should also update the bounds at the same time to maintain consistency. I wonder if we should remove the special behavior from ``setvalue`` and just use ``fix`` to update the values of fixed variables.

**Comment by joaquimg** [comment]
I think the current behavior is ambiguous and makes a little confusion, although I added the docs.

I would prefer ``setvalue`` only for warmstarts and ``fix`` for fixed values (allowing the user to fix previously non-fixed variables)

**Comment by joaquimg** [comment]
Lets go for  

> remove the special behavior from `setvalue` and just use `fix` to update the values of fixed variables.

?

**Comment by mlubin** [comment]
Yes. Calling ``setvalue`` on a fixed variable should (for now) throw an error since this is a breaking change and we don't want code silently changing behavior.

File: src/JuMP.jl [hunk 2]
```diff
@@ -386,15 +386,16 @@ getupperbound(v::Variable) = v.m.colUpper[v.col]
 function setvalue(v::Variable, val::Number)
     v.m.colVal[v.col] = val
     if v.m.colCat[v.col] == :Fixed
-        v.m.colLower[v.col] = val
-        v.m.colUpper[v.col] = val
+        error("setvalue use for setting bounds of fixed variables was deprecated in favour of JuMP.fix(v::Variable, val::Number), which changes bounds of fixed variables and fixes previously not fixed variables ")
```
mlubin: Deprecated would mean that we still support it but throw a warning. This isn't the case. ``s/favour/favor/``

joaquimg: what do I say?
`setvalue use for setting bounds of fixed variables was made obsolete in favor of JuMP.fix(v::Variable, val::Number), which changes bounds of fixed variables and fixes previously not fixed variables`

mlubin: ``setvalue for fixed variables is no longer supported. Use JuMP.fix instead.``

doc/refvariable.rst
```diff
@@ -219,7 +219,9 @@ Two possible uses for fixed variables are:
    the sensitivity of the objective with respect to the fixed value may be queried with ``getdual(x)``.
 
 2. For solving a sequence of problems with varying parameters.
-   One may call ``setvalue(x, val)``
+   One may call ``JuMP.fix(x, val)``
    to change the value to which the variable is fixed. For LPs
    in particular, most solvers are able to efficiently hot-start when
    solving the resulting modified problem.
+   
+Note that ``setvalue(x, val)`` will not set initial value for fixed variables, it will only set a warmstart for non-fixed variables. One can fix existing, non-fixed variables with ``JuMP.fix(x, val)`` which sets both bounds and initial value of ``x`` to ``val``.
```
mlubin: I don't think this paragraph is needed

doc/refvariable.rst
```diff
@@ -219,7 +219,9 @@ Two possible uses for fixed variables are:
    the sensitivity of the objective with respect to the fixed value may be queried with ``getdual(x)``.
 
 2. For solving a sequence of problems with varying parameters.
-   One may call ``setvalue(x, val)``
+   One may call ``JuMP.fix(x, val)``
```
mlubin: One may call ``JuMP.fix(x,val)`` to change the value the which the variable is fixed or to fix a previously unfixed variable.

mlubin:
Almost ready to merge, but it's currently failing tests:
```
Test Summary:                                 | Pass  Error  Total
  Variables                                   |   85      1     86
  constructors                                |   11            11
  get and set bounds                          |   11      1     12
```

mlubin:
There are still failing tests

joaquimg:
I will look into that!

joaquimg:
I had no errors in Variables, also I could not notice errors in Xpress and GLPK

mlubin:
See https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/190050309#L409

joaquimg:
Now its passing all tests!

PR #930, opened 2016-12-25
model-level scope for user-defined functions

### LINKED ISSUE #821: Make function registration local
Registering user-defined functions appears to be a global action.  This can lead to some undesirable outcomes.  For example:

```
function foo( A::Float64 )
  f(x) = exp( A * x )
  f1(x) = A * exp( A * x )
  f2(x) = A * A * exp( A * x )
      # Function to register
  JuMP.register(:f,1,f1,f2)
      # Registration
  ####
    # Optimization problem here using f
    # Leads to some return statement
  ####
end

# Session code
use1 = foo(2.0)
    # Works fine
use2 = foo(3.0)
    # Gives error "Operator f has already been defined"
use3 = foo(2.0)
    # Also gives the error(!)
```

Here I want the symbol `:f` to register a locally-defined JuMP function.  But it appears to be global _and_ immutable.  So I'd like to request the following features:
1. If possible, make function registration local.
2. If this isn't possible, add methods to: a) check that a function is registered, and b) de-register a previously-defined function.

The example above seems like a pretty important one.  For starters, it is a way to use one function (here `foo`) to solve similar problems with differently parameterized objective functions.  In addition, the current set-up leads to an error when calling the same function twice (see `use1` and `use3` above) with no apparent way to check for this (aside from a `try-catch` block).

Hopefully this isn't too hard to add, and it seems like the payoff would be pretty large.  Thanks!

PS.  This is based on my similar stackoverflow question here: http://stackoverflow.com/questions/38774351/how-to-check-if-a-user-defined-function-is-already-registered-in-julia-jump

mlubin:
``JuMP.register`` now takes the model as the first parameter, and registrations apply only to a given model. This is a breaking change.
Closes #821

CC @tkelman

### DIFF
--- a/doc/nlp.rst
+++ b/doc/nlp.rst
@@ -146,22 +146,22 @@ JuMP uses `ForwardDiff.jl <https://github.com/JuliaDiff/ForwardDiff.jl>`_ to per
     mysquare(x) = x^2
     myf(x,y) = (x-1)^2+(y-2)^2
 
-    JuMP.register(:myf, 2, myf, autodiff=true)
-    JuMP.register(:mysquare, 1, mysquare, autodiff=true)
-
     m = Model()
+
+    JuMP.register(m, :myf, 2, myf, autodiff=true)
+    JuMP.register(m, :mysquare, 1, mysquare, autodiff=true)
 
     @variable(m, x[1:2] >= 0.5)
     @NLobjective(m, Min, myf(x[1],mysquare(x[2])))
 
-The above code creates a JuMP model with the objective function ``(x[1]-1)^2 + (x[2]^2-2)^2``. The first argument to ``JuMP.register`` is a Julia symbol object which registers the name of the user-defined function in JuMP expressions; the JuMP name need not be the same as the name of the corresponding Julia method. The second argument specifies how many arguments the function takes. The third argument is the name of the Julia method which computes the function, and ``autodiff=true`` instructs JuMP to compute exact gradients automatically.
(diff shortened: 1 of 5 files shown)


mlubin:
CC @squipbar @evanfields @davidanthoff

doc/nlp.rst
```diff
@@ -146,22 +146,22 @@ JuMP uses `ForwardDiff.jl <https://github.com/JuliaDiff/ForwardDiff.jl>`_ to per
...
 
-The above code creates a JuMP model with the objective function ``(x[1]-1)^2 + (x[2]^2-2)^2``. The first argument to ``JuMP.register`` is a Julia symbol object which registers the name of the user-defined function in JuMP expressions; the JuMP name need not be the same as the name of the corresponding Julia method. The second argument specifies how many arguments the function takes. The third argument is the name of the Julia method which computes the function, and ``autodiff=true`` instructs JuMP to compute exact gradients automatically.
+The above code creates a JuMP model with the objective function ``(x[1]-1)^2 + (x[2]^2-2)^2``. The first argument to ``JuMP.register`` the model for which the functions are registered. The second argument is a Julia symbol object which wserves as the name of the user-defined function in JuMP expressions; the JuMP name need not be the same as the name of the corresponding Julia method. The third argument specifies how many arguments the function takes. The fourth argument is the name of the Julia method which computes the function, and ``autodiff=true`` instructs JuMP to compute exact gradients automatically.
```
evanfields: nit: typo in description of second argument; `wserves` -> `serves`

evanfields:
Glad to see this change - thanks!

PR #931, opened 2017-01-02
enable printing for nonlinear expressions

mlubin:
Mostly working, but IJulia output isn't right and needs more tests.
Slightly awkward since we don't have names for subexpressions or parameters, so they're printed as ``subexpression[i]`` or ``parameter[i]``.
Closes https://github.com/JuliaOpt/JuMP.jl/issues/201


mlubin:
IJulia printouts are correct now, ready to go from my end.

PR #938, opened 2017-01-14
save optimization data for ``stat !=: Optimal`` nlp problems 

huckl3b3rry87:
added functionality to save design variables, objective function values and constraint values for a nlp when the `stat != :Optimal`. This very useful when defining tolerances for the feasibility of the model. This works on the cases tested for both IPOPT and KNITRO. These tools both pass back the data even if the result is not feasible and studying this data can be very useful

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -1208,24 +1208,29 @@ function solvenlp(m::Model, traits; suppress_warnings=false)
     MathProgBase.optimize!(m.internalModel)
     stat = MathProgBase.status(m.internalModel)
 
-    if stat != :Infeasible && stat != :Unbounded
-        m.objVal = MathProgBase.getobjval(m.internalModel)
-        m.colVal = MathProgBase.getsolution(m.internalModel)
+    if stat == :Infeasible || stat == :Unbounded
+        suppress_warnings || warn("Saving objective value and design variable data even though, status: $stat")
     end
+    m.objVal = MathProgBase.getobjval(m.internalModel)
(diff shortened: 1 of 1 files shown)


mlubin:
This really calls for a way to query solution status via MathProgBase. CC @ccoffrin

huckl3b3rry87:
So, I guess that this can be closed because now JuMP passes back the solution even if it is not optimal?

I do have another issue related to this though, when it is solving my screen looks like this:
```
WARNING: Dual solution not available. Check that the model was properly solved and no integer variables are present.
WARNING: Dual solution not available. Check that the model was properly solved and no integer variables are present.
WARNING: Dual solution not available. Check that the model was properly solved and no integer variables are present.
WARNING: Dual solution not available. Check that the model was properly solved and no integer variables are present.
WARNING: Dual solution not available. Check that the model was properly solved and no integer variables are present.
```
etc....
Can we suppress these warnings somehow?  I noticed there has been discussion related to this here https://github.com/JuliaOpt/JuMP.jl/issues/91
and
here https://github.com/JuliaOpt/JuMP.jl/issues/325

mlubin:
You could just not call ``getdual`` if you know there's no dual solution to get.

huckl3b3rry87:
@mlubin oh, I forgot that I was calling that. Thanks!

mlubin:
This will be addressed by https://github.com/JuliaOpt/MathProgBase.jl/issues/164 and the corresponding changes to JuMP

Issue #940, opened 2017-01-20
Dot operator of vectors of AffExpr should avoid warning (like sum)?


cossio:
For example, a constraint of the form;

```julia
JuMP.@constraint(model, dot(u,v) == 0)
```
is very slow if `u,v` are large, and it generates a warning:

> WARNING: The addition operator has been used on JuMP expressions a large number of times. This warning is safe to ignore but may indicate that model generation is slower than necessary. For performance reasons, you should not add expressions in a loop. Instead of x += y, use append!(x,y) to modify x in place. If y is a single variable, you may also use push!(x, coef, y) in place of x += coef*y.

However, using `sum` is very fast and doesn't generate a warning:

```julia
JuMP.@constraint(model, sum(u[k] * v[k] for k = 1:N) == 0)
```

When `u,v` are vectors of `AffExpr` or vectors of JuMP variables, maybe `dot(u,v)` should automatically fall back on `sum(u[k] * v[k] for k = 1:N)`. If this is correct, I could try to implement it myself... although I could use some guidance on the codebase.

mlubin:
The fix would be to add a specialized method for ``dot`` [here](https://github.com/JuliaOpt/JuMP.jl/blob/dd4f5ea2e4cb9b2e142a0dc6162299b7ce5cb333/src/operators.jl#L306). The contribution is welcome, and should be a one-liner.

cossio:
@mlubin Please, review it.

cossio:
Here is a minimum example that raises the warning:

```
using JuMP
m = Model()
@variable(m, x[1:1_000_000]);
@constraint(m, dot(x, 1 : 1_000_000) <= 1);
```

joehuchette:
Great, thanks. For this particular case, I believe a fix will be to change ``Array`` to ``AbstractArray`` and ``Vector`` to ``AbstractVector`` here: https://github.com/JuliaOpt/JuMP.jl/blob/d725975b571309ffaff840ca9709e739287c8d41/src/operators.jl#L299-L322

PR #941, opened 2017-01-20
Update operators.jl (_dot of vectors calls sum)

cossio:
Trying to fix:

https://github.com/JuliaOpt/JuMP.jl/issues/940#issuecomment-274198879

Now `_dot(::Vector, ::Vector)` calls `sum`, which already as an efficient implementation.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -314,7 +314,10 @@ vecdot{T,S<:JuMPTypes,N}(lhs::Array{T,N},rhs::Array{S,N}) = _dot(lhs,rhs)
 
 function _dot{T,S}(lhs::Vector{T}, rhs::Vector{S})
     length(lhs) == length(rhs) || error("Incompatible dimensions")
-    sum(lhs[i] * rhs[i] for i = 1:length(lhs))
+    res = AffExpr()
+    for i = 1 : length(lhs)
+        append!(res, lhs[i] * rhs[i])
+    end
 end
     
 function _dot{T,S}(lhs::Array{T}, rhs::Array{S})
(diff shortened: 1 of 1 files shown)


joehuchette:
I don't believe ``sum`` is specialized to avoid temporary allocations (and therefore also the warning message). Our macros (e.g. ``@constraint``) rewrite summations to avoid temporaries. Maybe try writing out the ``for`` loop explicitly and use ``append!``?

mlubin:
``sum`` outside of a macro does not have an efficient implementation.

mlubin:
Also, this implementation is way too general. It overrides Julia's implementation of ``dot`` even when no JuMP-related types are involved.

cossio:
@mlubin The method is called `_dot`, there is no conflict with Julia's `dot`.

@joehuchette I wrote the explicit loop.

joehuchette:
This is now a more specific method of https://github.com/cossio/JuMP.jl/blob/47aa22928ef5450014c9677930837187a94354f1/src/operators.jl#L323-L330. What is the case that currently gives a warning that you are trying to fix?

cossio:
@joehuchette Trying to fix https://github.com/JuliaOpt/JuMP.jl/issues/940.

joehuchette:
Can you post a short code snippet that reproduces #940? I can't with something like:
```jl
using JuMP
m = Model()
@variable(m, x[1:1_000_000]);
dot(x,x);
```

cossio:
@joehuchette I added a minimum example to https://github.com/JuliaOpt/JuMP.jl/issues/940. Reproducing here:

```julia
using JuMP
m = Model()
@variable(m, x[1:1_000_000]);
@constraint(m, dot(x, 1 : 1_000_000) <= 1);
```

cossio:
Following your suggestion @joehuchette. Replaced with https://github.com/JuliaOpt/JuMP.jl/pull/943

PR #942, opened 2017-01-24
Splits prepProblemBounds into prepAffObjective and prepConstrBounds

blegat:


src/solvers.jl
```diff
@@ -503,8 +504,7 @@ function addSOS(m::Model)
 end
 
 # Returns coefficients for the affine part of the objective and the
```
mlubin: s/and the//

mlubin:
LGTM

PR #944, opened 2017-01-24
enable warmstarts for continuous problems through the LPQP pathway

mlubin:
https://discourse.julialang.org/t/create-quadratic-objective-will-objective-and-nlobjective-lead-to-different-solution-using-ipopt/1666

Requires https://github.com/JuliaOpt/MathProgBase.jl/commit/00c9247e9b840ef5ee7eac91f94aafa5d4ab03bb to be tagged before we can test properly

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -1254,13 +1254,7 @@ function _buildInternalModel_nlp(m::Model, traits)
         end
     end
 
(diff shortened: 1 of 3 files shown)


mlubin:
Rebased and added a test. Ready to go once travis approves.

PR #943, opened 2017-01-24
Update dot product of JuMP containers

cossio:
Fixing https://github.com/JuliaOpt/JuMP.jl/issues/940.
Updated version of previous pull: https://github.com/JuliaOpt/JuMP.jl/pull/941.

PR #947, opened 2017-01-27
Added tests for pull #943

cossio:
https://github.com/JuliaOpt/JuMP.jl/pull/943
https://github.com/JuliaOpt/JuMP.jl/issues/940

### DIFF
--- a/test/operator.jl
+++ b/test/operator.jl
@@ -499,7 +499,7 @@ const sub2 = JuMP.repl[:sub2]
         # https://github.com/JuliaOpt/JuMP.jl/pull/943
         pull943 = Model()
         @variable(pull943, x[1 : 10^6]);
-        setvalue.(x, rand(10^6))
+        setvalue.(x, 1 : 10^6)
         @expression(pull943, testsum, sum(x[i] * i for i = 1 : 10^6))
         @expression(pull943, testdot1, dot(x, 1 : 10^6))
         @expression(pull943, testdot2, dot(1 : 10^6, x))
--- a/test/operator.jl
+++ b/test/operator.jl
@@ -495,6 +495,17 @@ const sub2 = JuMP.repl[:sub2]
         anys[1] = 10
         anys[2] = 20 + x
         @test dot(floats, anys) == 10 + 40 + 2x
+            
+        # https://github.com/JuliaOpt/JuMP.jl/pull/943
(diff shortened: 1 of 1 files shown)


test/operator.jl
```diff
@@ -495,6 +495,17 @@ const sub2 = JuMP.repl[:sub2]
         anys[1] = 10
         anys[2] = 20 + x
         @test dot(floats, anys) == 10 + 40 + 2x
+            
+        # https://github.com/JuliaOpt/JuMP.jl/pull/943
+        pull943 = Model()
+        @variable(pull943, x[1 : 10^6]);
+        setvalue.(x, rand(n))
```
joehuchette: What is ``n``?

cossio: Sorry, that's 10^6. Fixed it.

test/operator.jl
```diff
@@ -495,6 +495,17 @@ const sub2 = JuMP.repl[:sub2]
         anys[1] = 10
         anys[2] = 20 + x
         @test dot(floats, anys) == 10 + 40 + 2x
+            
+        # https://github.com/JuliaOpt/JuMP.jl/pull/943
+        pull943 = Model()
+        @variable(pull943, x[1 : 10^6]);
+        setvalue.(x, rand(10^6))
+        setvalue.(y, 1 : 10^6)
```
joehuchette: What is ``y``? This part is failing Travis.

cossio: Sorry again. I copy and pasted from a version that had other var names.
Fixed it.

cossio:
@joehuchette  Do you think having random data in the test is a good idea? It's been worrying me. I could replace that with some fixed data.

mlubin:
> Do you think having random data in the test is a good idea?

No, it's not a good idea.

mlubin:
There's no need to test with ``10^6`` variables, three would be just fine.

joehuchette:
That wouldn't trigger the warning in the case of a regression, though.

mlubin:
Ah, right

cossio:
@mlubin Using fixed data now.

joehuchette:
Thanks, @cossio.

In the future, it would be preferable to attach the tests to the PR which lead to them. Learning to use git is a good investment of time :)

PR #952, opened 2017-01-31
Adding Model Getters for SOS, SD, and NL Constraint Counts

ccoffrin:
A hack at adding this without understanding the best way to do it.  Suggestions welcome.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -194,7 +194,7 @@ numsdconstr(m::Model) = length(m.sdpconstr)
 numnlconstr(m::Model) = m.nlpdata !== nothing ? length(m.nlpdata.nlconstr) : 0
 
 function MathProgBase.numconstr(m::Model)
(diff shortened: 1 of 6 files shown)


mlubin:
Seems reasonable to me. Could you update the corresponding documentation in ``doc/refmodel.rst``?

src/JuMP.jl
```diff
@@ -188,8 +188,13 @@ end
 MathProgBase.numvar(m::Model) = m.numCols
 MathProgBase.numlinconstr(m::Model) = length(m.linconstr)
 MathProgBase.numquadconstr(m::Model) = length(m.quadconstr)
+# Note SOC constraints are captured in numquadconstr
```
mlubin: SOC constraints are only captured in ``numquadconstr`` if they're specified as quadratic constraints. If they're specified as SOC constraints using the ``norm`` syntax then they will not be counted there.

ccoffrin: Ok, I'll fix that too.

ccoffrin:
I think both comments are addressed now.

mlubin:
Thanks!

PR #970, opened 2017-02-21
remove default solver functionality (#965)

### LINKED ISSUE #965: bikeshedding default solver behavior
Ok, time to figure out the user-facing side of default solvers. As of https://github.com/JuliaOpt/JuMP.jl/pull/959, we have the following significant changes:
1. An informational message is printed whenever no solver is satisfied but one happens to be available.
2. On Julia 0.6, users have to explicitly load solver packages to make them available.
3. On Julia 0.5, there are no breaking changes yet, but we now print a warning when we have to explicitly load a solver in a way that will break on 0.6.

The question is, how do we present example code now?

```jl
using JuMP

m = Model()
@variable(m, 0 <= x <= 2)
@variable(m, 0 <= y <= 30)
@objective(m, Max, 5x + 3y)
@constraint(m, 1x + 5y <= 3.0)
status = solve(m)
```
will no longer work. Some options are:

**Option 1**:
```jl
using JuMP, Clp

m = Model()
@variable(m, 0 <= x <= 2)
@variable(m, 0 <= y <= 30)
@objective(m, Max, 5x + 3y)
@constraint(m, 1x + 5y <= 3.0)
status = solve(m)
# prints a message on solve that clp was chosen by default
```

**Option 2**:
```jl
using JuMP, Clp

m = Model(solver = ClpSolver())
@variable(m, 0 <= x <= 2)
@variable(m, 0 <= y <= 30)
@objective(m, Max, 5x + 3y)
@constraint(m, 1x + 5y <= 3.0)
status = solve(m)
# no messages printed by JuMP
```

This is the most similar to "production" code but feels a bit ugly to me. It's potentially annoying that you'd have to install Clp or modify the code (digging through it to find all uses of ``ClpSolver()``) to run examples.

**Option 3**:
```jl
using JuMP

JuMP.loaddefaultsolvers() # Loads available solvers

m = Model()
@variable(m, 0 <= x <= 2)
@variable(m, 0 <= y <= 30)
@objective(m, Max, 5x + 3y)
@constraint(m, 1x + 5y <= 3.0)
status = solve(m)
# If an LP solver is installed, prints a message on solve that one was chosen by default. If not, prints an error with suggested solvers.
```

Should we even have a ``loaddefaultsolvers`` method?

Relevant prior discussion at https://github.com/JuliaOpt/MathProgBase.jl/pull/150.

CC @odow @chriscoey @blegat @chkwon

mlubin:
Closes #965.

### DIFF
--- a/src/solvers.jl
+++ b/src/solvers.jl
@@ -38,103 +38,50 @@ function ProblemTraits(m::Model; relaxation=false)
     ProblemTraits(int, !(qp|qc|nlp|soc|sdp|sos), qp, qc, nlp, soc, sdp, sos, soc|sdp)
 end
 
-# these are the default solver packages in order of decreasing precedence
-# for each supported problem class.
+const suggestedsolvers = Dict("LP" => [(:Clp,:ClpSolver),
+                                       (:GLPKMathProgInterface,:GLPKSolverLP),
+                                       (:Gurobi,:GurobiSolver),
+                                       (:CPLEX,:CplexSolver),
+                                       (:Mosek,:MosekSolver),
+                                       (:Xpress,:XpressSolver)],
+                              "MIP" => [(:Cbc,:CbcSolver),
+                                        (:GLPKMathProgInterface,:GLPKSolverMIP),
+                                        (:Gurobi,:GurobiSolver),
+                                        (:CPLEX,:CplexSolver),
+                                        (:Mosek,:MosekSolver),
+                                        (:Xpress,:XpressSolver)],
+                              "QP" => [(:Gurobi,:GurobiSolver),
+                                       (:CPLEX,:CplexSolver),
+                                       (:Mosek,:MosekSolver),
+                                       (:Ipopt,:IpoptSolver),
+                                       (:Xpress,:XpressSolver)],
+                              "SDP" => [(:Mosek,:MosekSolver),
+                                       (:SCS,:SCSSolver)],
+                              "NLP" => [(:Ipopt,:IpoptSolver),
+                                        (:KNITRO,:KnitroSolver),
+                                        (:Mosek,:MosekSolver)],
+                              "Conic" => [(:ECOS,:ECOSSolver),
+                                         (:SCS,:SCSSolver),
+                                         (:Mosek,:MosekSolver)])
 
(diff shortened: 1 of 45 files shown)


src/solvers.jl
```diff
@@ -392,10 +339,8 @@ end
 # Converts the JuMP Model into a MathProgBase model based on the
 # traits of the model
 function build(m::Model; suppress_warnings=false, relaxation=false, traits=ProblemTraits(m,relaxation=relaxation))
-    # Set solver based on the model's traits if it hasn't provided
     if isa(m.solver, UnsetSolver)
-        m.solver = default_solver(traits)
-        suppress_warnings || println("JuMP is using $(m.solver) as the solver. Use the \"solver=\" keyword to Model() or setsolver() to change the solver or solver options.")
+        m.solver = no_solver_error(traits)
```
joehuchette: Why the assignment?

src/solvers.jl
```diff
@@ -38,103 +38,50 @@ function ProblemTraits(m::Model; relaxation=false)
...
+
+function no_solver_error(traits::ProblemTraits)
+
+    # This is pretty coarse and misses out on MIConic, MIQP, etc.
     if traits.nlp
-        defaultNLPsolver()
+        class = "NLP"
     elseif traits.int || traits.sos
-        defaultMIPsolver()
+        class = "MIP"
```
joehuchette: Since the issues with the "coarseness" of this function seem to come up with discrete problems, might be good to split this logic out of this loop. Could add a helper function like 
```jl
function supports_MIP(x::Tuple{Symbol,Symbol})
    x[1] in (:Cbc, :Gurobi, :CPLEX, :Knitro, ...)
end
```

src/solvers.jl
mlubin: Oops

src/solvers.jl
mlubin: I'll leave refining the categories for a future PR, just trying to get things working using the lists of solvers in their existing categories.

mlubin: (These are just the solvers that were already chosen as defaults, so we're not introducing anything new here.)

mlubin:
Examples and docs updated, all set on my end.

PR #975, opened 2017-02-22
Fix equality symbol for windows in tests

blegat:


mlubin:
Good catch. I've been ignoring AppVeyor since Ipopt has been broken. We won't get a check mark until that's fixed.

PR #980, opened 2017-02-24
Make things work for general `AbstractArray`s

tkoolen:
A lot of functions currently require `Array` types, while they would work just fine for any other subtype of `AbstractArray`. This PR simply replaces most `::Array`s in function signatures by `::AbstractArray` (and likewise for `Vector` and `Matrix`).

I defined `norm` only for `AbstractVector` and `AbstractMatrix`, not for general `AbstractArray`s, because Base only defines it for `AbstractVector` and `AbstractMatrix`, and defining it for `AbstractArray` results in an ambiguity error.

Please be extra careful reviewing my changes to `_multiply!` and `_multiplyt!` in operators.jl.

This came up while I was playing around with storing JuMP `Variable`s in `AxisArray`s for a trajectory optimization style problem, because I had trouble keeping track of the order of the dimensions of the variable arrays (this is proving to be quite a nice combination of tools by the way).


tkoolen:
cc: @rdeits

src/JuMP.jl
```diff
@@ -909,9 +909,9 @@ include("print.jl")
 # Deprecations
 include("deprecated.jl")
 
-getvalue{T<:JuMPTypes}(arr::Array{T}) = map(getvalue, arr)
+getvalue{T<:JuMPTypes}(arr::AbstractArray{T}) = map(getvalue, arr)
```
mlubin: Not needed, use ``getvalue.(arr)`` instead.

src/JuMP.jl
```diff
@@ -909,9 +909,9 @@ include("print.jl")
...
 
-function setvalue{T<:AbstractJuMPScalar}(set::Array{T}, val::Array)
+function setvalue{T<:AbstractJuMPScalar}(set::AbstractArray{T}, val::AbstractArray)
```
mlubin: Also not needed, use ``setvalue.(set,val)``

src/JuMPArray.jl
```diff
@@ -10,7 +10,7 @@ immutable JuMPArray{T,N,NT} <: JuMPContainer{T,N}
     meta::Dict{Symbol,Any}
 end
 
-@generated function JuMPArray{T,N}(innerArray::Array{T,N}, indexsets::NTuple{N,Any})
+@generated function JuMPArray{T,N}(innerArray::AbstractArray{T,N}, indexsets::NTuple{N,Any})
```
mlubin: I don't think this is needed

src/affexpr.jl
```diff
@@ -181,10 +181,10 @@ function addconstraint(m::Model, c::LinearConstraint)
     end
     return LinConstrRef(m,length(m.linconstr))
 end
-addconstraint(m::Model, c::Array{LinearConstraint}) =
+addconstraint(m::Model, c::AbstractArray{LinearConstraint}) =
     error("The operators <=, >=, and == can only be used to specify scalar constraints. If you are trying to add a vectorized constraint, use the element-wise dot comparison operators (.<=, .>=, or .==) instead")
 
-function addVectorizedConstraint(m::Model, v::Array{LinearConstraint})
+function addVectorizedConstraint(m::Model, v::AbstractArray{LinearConstraint})
```
mlubin: This is an internal function, it's unclear how it could be called with an ``AbstractArray``

src/callbacks.jl
```diff
@@ -59,7 +59,7 @@ function addinfocallback(m::Model, f::Function; when::Symbol = _unspecifiedstate
     push!(m.callbacks, InfoCallback(f, when))
 end
 
-function lazycallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vector{LazyCallback})
+function lazycallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::AbstractVector{LazyCallback})
```
mlubin: Internal, won't be called with ``AbstractVector``

src/callbacks.jl
```diff
@@ -92,7 +92,7 @@ function lazycallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vecto
     :Continue
 end
 
-function attach_callbacks(m::Model, cbs::Vector{LazyCallback})
+function attach_callbacks(m::Model, cbs::AbstractVector{LazyCallback})
```
mlubin: Internal, won't be called with ``AbstractVector``

src/callbacks.jl
```diff
@@ -101,7 +101,7 @@ function attach_callbacks(m::Model, cbs::Vector{LazyCallback})
     end
 end
 
-function cutcallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vector{CutCallback})
+function cutcallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::AbstractVector{CutCallback})
```
mlubin: Internal, won't be called with ``AbstractVector``

src/callbacks.jl
```diff
@@ -119,7 +119,7 @@ function cutcallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vector
     :Continue
 end
 
-function attach_callbacks(m::Model, cbs::Vector{CutCallback})
+function attach_callbacks(m::Model, cbs::AbstractVector{CutCallback})
```
mlubin: Internal, won't be called with ``AbstractVector``

src/callbacks.jl
```diff
@@ -129,7 +129,7 @@ function attach_callbacks(m::Model, cbs::Vector{CutCallback})
 end
 
 
-function heurcallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vector{HeuristicCallback})
+function heurcallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::AbstractVector{HeuristicCallback})
```
mlubin: Internal, won't be called with ``AbstractVector``

src/callbacks.jl
```diff
@@ -147,7 +147,7 @@ function heurcallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vecto
     :Continue
 end
 
-function attach_callbacks(m::Model, cbs::Vector{HeuristicCallback})
+function attach_callbacks(m::Model, cbs::AbstractVector{HeuristicCallback})
```
mlubin: Internal, won't be called with ``AbstractVector``

src/callbacks.jl
```diff
@@ -156,7 +156,7 @@ function attach_callbacks(m::Model, cbs::Vector{HeuristicCallback})
     end
 end
 
-function infocallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::Vector{InfoCallback})
+function infocallback(d::MathProgBase.MathProgCallbackData, m::Model, cbs::AbstractVector{InfoCallback})
```
mlubin: Internal, won't be called with ``AbstractVector``

src/nlp.jl
```diff
@@ -169,7 +169,7 @@ type NLPEvaluator <: MathProgBase.AbstractNLPEvaluator
     end
 end
 
-function simplify_expression(nd::Vector{NodeData}, const_values, subexpression_linearity, fixed_variables, parameter_values, x_values, subexpression_values)
+function simplify_expression(nd::AbstractVector{NodeData}, const_values, subexpression_linearity, fixed_variables, parameter_values, x_values, subexpression_values)
```
mlubin: None of the changes in this file are needed

src/operators.jl
```diff
@@ -370,7 +370,7 @@ function _multiply!{T<:JuMPTypes}(ret::Array{T}, lhs::Array, rhs::Array)
 end
 
 # this computes lhs.'*rhs and places it in ret
-function _multiplyt!{T<:JuMPTypes}(ret::Array{T}, lhs::Array, rhs::Array)
+function _multiplyt!{T<:JuMPTypes}(ret::AbstractArray{T}, lhs::AbstractArray, rhs::AbstractArray)
```
mlubin: These methods are designed for dense matrices and shouldn't be widened

src/sos.jl
```diff
@@ -25,7 +25,7 @@ Base.copy(sos::SOSConstraint, new_model::Model) =
 
 # Given a vector of affine expressions, extract a vector of the single
 # variable in each expression and a vector of their coefficients
-function constructSOS(m::Model, coll::Vector{AffExpr})
+function constructSOS(m::Model, coll::AbstractVector{AffExpr})
```
mlubin: These changes don't seem relevant

mlubin:
I'm willing to consider this use case, but the changes here are way too broad and touch a number of internal functions that would never be called with ``AbstractArray``. Could you try to make the changes minimal with respect to what you need, and also add unit tests covering them?

tkoolen:
Sure. Thanks for the quick review. I'd like to argue that `AbstractArray` should always be the default (unless things are specialized for dense arrays, like in `_multiply!`, of course), and restricting to `Array` is kind of arbitrary, even in internal code that is currently not being called with other types. But I'll try to make the changes minimal and add tests.

src/JuMP.jl
tkoolen: So delete the method?

src/quadexpr.jl
```diff
@@ -95,7 +95,7 @@ function getvalue(a::QuadExpr)
     end
     return ret
 end
-getvalue(arr::Array{QuadExpr}) = map(getvalue, arr)
+getvalue(arr::AbstractArray{QuadExpr}) = map(getvalue, arr)
```
mlubin: Also ``getvalue.(arr)`` here

src/JuMP.jl
mlubin: It's been previously discussed. The conclusion for now is to allow ``getvalue`` on objects that JuMP creates, and JuMP can create ``Array{Variable}``, e.g., PSD matrices.

rdeits:
To motivate why this would be nice to have, here are some cool things that @tkoolen and I have been doing w.r.t. AxisArrays of JuMP variables: https://gist.github.com/rdeits/3371af92c9c4aa61353f38c9a91c02e9

mlubin:
AxisArrays look like a potential replacement for JuMPArrays.

mlubin:
> restricting to Array is kind of arbitrary, even in internal code that is currently not being called with other types.

Personally I don't see much value in putting ``AbstractArray`` over no type restriction at all if we're already going to loosen the type beyond what's actually used and has been tested.

rdeits:
AxisArrays do seem like a great replacement for the categorical indexing in JuMPArray. I think they don't cover the ragged and non-1-based indexing that JuMPArray supports, though.

tkoolen:
I've addressed your comments. If there are more places where you want things to be more specific, let me know.

I limited the test to cases that resulted in `MethodError`s before this PR. There's quite a few cases that 'work' before this PR, but don't use the (I guess optimized) JuMP code, even though it would be applicable. I'm not sure how to test for these. For example,
```julia
m = Model()
x = view(@variable(m, [1: 3]), :)
A = rand(3, 3)

@show @which sum(x)
@show @which diagm(x)
@show @which norm(x)
@show @which A * x
@show @which x + 1
@show @which x * 1
```
prints
```
@which(sum(x)) = sum(a) at reduce.jl:229
@which(diagm(x)) = diagm{T}(v::AbstractArray{T,1}) at linalg/dense.jl:121
@which(norm(x)) = norm(x::AbstractArray{T<:Any,1}) at linalg/generic.jl:201
@which(A * x) = *{T<:Union{Complex{Float32},Complex{Float64},Float32,Float64},S}(A::Union{Base.ReshapedArray{T,2,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N<:Any}}},DenseArray{T,2},SubArray{T,2,A<:Union{Base.ReshapedArray{T<:Any,N<:Any,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N<:Any}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N<:Any}},L<:Any}}, x::Union{Base.ReshapedArray{S,1,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N<:Any}}},DenseArray{S,1},SubArray{S,1,A<:Union{Base.ReshapedArray{T<:Any,N<:Any,A<:DenseArray,MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N<:Any}}},DenseArray},I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex,Colon,Int64,Range{Int64}},N<:Any}},L<:Any}}) at linalg/matmul.jl:79
@which(x + 1) = +(A::AbstractArray, x::Number) at arraymath.jl:94
@which(x * 1) = *(A::AbstractArray, B::Number) at abstractarraymath.jl:97
```
while
```julia
m = Model()
x = @variable(m, [1: 3])
A = rand(3, 3)

@show @which sum(x)
@show @which diagm(x)
@show @which norm(x)
@show @which A * x
@show @which x + 1
@show @which x * 1
```
prints
```
@which(sum(x)) = sum(j::Array{JuMP.Variable,N<:Any}) at /Users/twan/code/julia/MixedIntegerExperiments/v0.5/JuMP/src/operators.jl:279
@which(diagm(x)) = diagm(x::Array{JuMP.Variable,1}) at /Users/twan/code/julia/MixedIntegerExperiments/v0.5/JuMP/src/operators.jl:351
@which(norm(x)) = norm{V<:JuMP.AbstractJuMPScalar}(x::Array{V,N<:Any}) at /Users/twan/code/julia/MixedIntegerExperiments/v0.5/JuMP/src/norms.jl:44
@which(A * x) = *{T<:Union{JuMP.AbstractJuMPScalar,JuMP.GenericNormExpr{2,Float64,JuMP.Variable},JuMP.GenericNorm{P,Float64,JuMP.Variable},JuMP.NonlinearExpression}}(A::Union{Array{T<:Any,2},SparseMatrixCSC}, x::Union{Array{T,1},Array{T,2},SparseMatrixCSC{T,Ti<:Integer}}) at /Users/twan/code/julia/MixedIntegerExperiments/v0.5/JuMP/src/operators.jl:447
@which(x + 1) = +{T<:Union{JuMP.AbstractJuMPScalar,JuMP.GenericNormExpr{2,Float64,JuMP.Variable},JuMP.GenericNorm{P,Float64,JuMP.Variable},JuMP.NonlinearExpression}}(lhs::Union{Array{T,N<:Any},SparseMatrixCSC{T,Ti<:Integer}}, rhs::Number) at /Users/twan/code/julia/MixedIntegerExperiments/v0.5/JuMP/src/operators.jl:507
@which(x * 1) = *{T<:Union{JuMP.AbstractJuMPScalar,JuMP.GenericNormExpr{2,Float64,JuMP.Variable},JuMP.GenericNorm{P,Float64,JuMP.Variable},JuMP.NonlinearExpression}}(lhs::Array{T,N<:Any}, rhs::Number) at /Users/twan/code/julia/MixedIntegerExperiments/v0.5/JuMP/src/operators.jl:538
```

(note: this PR in its current form doesn't address all of these).

> Personally I don't see much value in putting `AbstractArray` over no type restriction at all if we're already going to loosen the type beyond what's actually used and has been tested.

The main value I see is that it conveys the intention of the code. Moreover, you still get to put constraints on the dimension and element type, and you need it to avoid ambiguities. However, widening signatures to `::AbstractArray` also makes it so that if things don't work for a type that satisfies the requirements of being a (reasonable) `AbstractArray` subtype, then users get to complain, so I understand your hesitancy.

A (possibly overly) cheap response to your argument is that for some methods, you're not currently testing all possible values of `N` or `T` for `Array{T, N}` either. You couldn't possibly test all possible `AbstractArray` subtypes, but in my opinion, that doesn't mean that using `::Array` instead of `::AbstractArray` is necessarily the right thing to do.

src/operators.jl
```diff
@@ -348,14 +348,14 @@ function Base.issymmetric{T<:JuMPTypes}(x::Matrix{T})
...
-Base.diagm(x::Vector{Variable}) = diagm(convert(Vector{AffExpr}, x))
+Base.diagm(x::AbstractVector{Variable}) = diagm(convert(Vector{AffExpr}, x))
 
 ###############
 # The _multiply!(buf,y,z) adds the results of y*z into the buffer buf. No bounds/size
 # checks are performed; it is expected that the caller has done this, has ensured
 # that the eltype of buf is appropriate, and has zeroed the elements of buf (if desired).
 
-function _multiply!{T<:JuMPTypes}(ret::Array{T}, lhs::Array, rhs::Array)
+function _multiply!{T<:JuMPTypes}(ret::AbstractArray{T}, lhs::Array, rhs::Array)
```
mlubin: @joehuchette should comment but I don't think this method was designed for ``AbstractArray``s. The built-in fallback should work

src/operators.jl
```diff
@@ -571,7 +571,7 @@ end; end
 
 # The following are primarily there for internal use in the macro code for @constraint
 for op in [:(+), :(-)]; @eval begin
-    function $op(lhs::Array{Variable},rhs::Array{Variable})
+    function $op(lhs::AbstractArray{Variable},rhs::AbstractArray{Variable})
         (sz = size(lhs)) == size(rhs) || error("Incompatible sizes for $op: $sz $op $(size(rhs))")
         ret = Array{AffExpr}(sz)
         for I in eachindex(ret)
```
mlubin: I'm concerned about the validity of using ``eachindex`` from ``ret`` to index into ``lhs`` when these could be different types.

src/print.jl
```diff
@@ -44,7 +44,7 @@ end
 
 # TODO: get rid of this! This is only a helper, and should be Base.values
 # (and probably live there, as well)
-_values(x::Array) = x
+_values(x::AbstractArray) = x
```
mlubin: Where is this used and why does it need to change?

src/print.jl
```diff
@@ -103,7 +103,7 @@ math(s,mathmode) = mathmode ? s : "\$\$ $s \$\$"
 # helper to look up corresponding JuMPContainerData
 printdata(v::JuMPContainer) = _getmodel(v).varData[v]
 getname(x::JuMPContainer) = hasmeta(x, :model) ? printdata(x).name : "__anon__"
-function printdata(v::Array{Variable})
+function printdata(v::AbstractArray{Variable})
```
mlubin: Are you redefining printing for ``AbstractArray{Variable}``? If not, this doesn't need to be touched

src/quadexpr.jl
```diff
@@ -149,10 +149,10 @@ function addconstraint(m::Model, c::QuadConstraint)
     end
     return ConstraintRef{Model,QuadConstraint}(m,length(m.quadconstr))
 end
-addconstraint(m::Model, c::Array{QuadConstraint}) =
+addconstraint(m::Model, c::AbstractArray{QuadConstraint}) =
     error("Vectorized constraint added without elementwise comparisons. Try using one of (.<=,.>=,.==).")
 
-function addVectorizedConstraint(m::Model, v::Array{QuadConstraint})
+function addVectorizedConstraint(m::Model, v::AbstractArray{QuadConstraint})
```
mlubin: This change is not covered by tests.

src/sos.jl
```diff
@@ -43,7 +43,7 @@ end
 
 addSOS1(m::Model, coll) = addSOS1(m, convert(Vector{AffExpr}, coll))
 
-function addSOS1(m::Model, coll::Vector{AffExpr})
+function addSOS1(m::Model, coll::AbstractVector{AffExpr})
```
mlubin: Not covered by tests. This method calls ``constructSOS`` which is defined for ``Vector{AffExpr}``.

mlubin: Also doesn't seem necessary for your use case

src/JuMP.jl
```diff
@@ -529,7 +529,7 @@ type SDConstraint <: AbstractConstraint
 end
 
 # Special-case X ≥ 0, which is often convenient
-function SDConstraint(lhs::Matrix, rhs::Number)
+function SDConstraint(lhs::AbstractMatrix, rhs::Number)
```
mlubin: This change is not covered by tests.

mlubin:
Also if you're claiming to accept ``AbstractArray``s then you should also guarantee that the code does not produce incorrect results for [arrays with custom indices](http://docs.julialang.org/en/stable/devdocs/offset-arrays/#devdocs-offsetarrays) (errors are okay though).

tkoolen:
I'm working on addressing your comments.

I was wondering about the operator definitions starting at
https://github.com/JuliaOpt/JuMP.jl/blob/5735b54e98cad041af3606b2b1dbccc0c573b440/src/operators.jl#L498.

Currently, `+`, `-`, `*`, and `/` work just fine for combinations of `AbstractArray{JuMPType}`s and `Number`s, as the generic methods in Base are called. The definitions in JuMP don't appear to be doing anything special compared to the generic methods, so should I just remove these?

mlubin:
@tkoolen, at one point in time (more than a year ago) I believe they were needed because the built-in definitions followed different promotion rules that didn't work for us because, e.g., ``AffExpr*AffExpr`` is a ``QuadExpr``. This hasn't been looked at recently.

mlubin:
@rdeits, so all index sets must be ordered, I presume? JuMPArrays don't make that assumption but it's reasonable.

src/operators.jl
joehuchette: Yes, I don't think this part of the code will work with generic abstract arrays as-is.

tkoolen:
About the operator definitions: simply removing them results in `StackOverflowError`s with e.g. `*(A::Number, B::AbstractArray)` in base and the .* overload in `operators.jl`. The dotop overloads should probably be removed, but I didn't want to go further down the rabbit hole in this PR.

mlubin:
Not a problem if you don't want to touch them in this PR

rdeits:
I think so. For example, you *can* create an Axis with a Set of categorical values (instead of a vector), but indexing it won't work: 

```julia
julia> a = AxisArray(rand(1:10, 3, 2), Axis{:time}(1:3), Axis{:side}(Set([:left, :right])))
2-dimensional AxisArray{Int64,2,...} with axes:
    :time, 1:3
    :side, Set(Symbol[:left,:right])
And data, a 3×2 Array{Int64,2}:
 2  9
 5  6
 9  5

julia> a[1, :left]
ERROR: elementwise indexing is not supported for axes of type Set{Symbol}
 in getindex(::AxisArrays.AxisArray{Int64,2,Array{Int64,2},Tuple{AxisArrays.Axis{:time,UnitRange{Int64}},AxisArrays.Axis{:side,Set{Symbol}}}}, ::Int64, ::Symbol) at /Users/rdeits/.julia/v0.5/AxisArrays/src/indexing.jl:84
```

Fortunately, it gives a pretty reasonable error message when you try.

src/JuMP.jl
tkoolen: Fixed.

src/operators.jl
tkoolen: Oops, forgot this one.

src/quadexpr.jl
tkoolen: I wasn't able to construct a test case involving nonstandard arrays of `QuadConstraint` that didn't already pass or call this unexported function directly, so I changed it back to `Array`.

tkoolen:
> Also if you're claiming to accept AbstractArrays then you should also guarantee that the code does not produce incorrect results for arrays with custom indices (errors are okay though).

I could change the tests to use [OffsetArrays](https://github.com/alsam/OffsetArrays.jl), but I do foresee problems on 0.5, where [support for such arrays is only partial at the moment](https://github.com/alsam/OffsetArrays.jl#special-note-for-julia-05).

src/operators.jl
tkoolen: Fixed (here and in a number of other locations).

mlubin:
> I could change the tests to use OffsetArrays, but I do foresee problems on 0.5, where support for such arrays is only partial at the moment.

It's the opposite. Any code in JuMP that that's calling ``size`` on (now) an AbstractArray with custom indices is probably incorrect. This will throw an error on 0.5 which is fine. On 0.6 it may silently do the wrong thing which is not okay.

mlubin:
One option is to follow the suggestion in the docs:

> Note that if you don’t want to be bothered supporting arrays with non-1 indexing, you can add the following line:
```jl
@assert all(x->isa(x, Base.OneTo), indices(A))
```
> at the top of any function.

tkoolen:
I'll push some more changes tonight.

tkoolen:
New round of changes, please take a look.

src/operators.jl
```diff
@@ -277,7 +277,8 @@ Base.sum(j::JuMPDict)  = sum(values(j.tupledict))
 Base.sum(j::JuMPArray{Variable}) = AffExpr(vec(j.innerArray), ones(length(j.innerArray)), 0.0)
 Base.sum(j::JuMPDict{Variable})  = AffExpr(collect(values(j.tupledict)), ones(length(j.tupledict)), 0.0)
 Base.sum(j::Array{Variable}) = AffExpr(vec(j), ones(length(j)), 0.0)
-function Base.sum{T<:GenericAffExpr}(affs::Array{T})
+Base.sum(j::AbstractArray{Variable}) = sum([j...]) # to handle non-one-indexed arrays.
```
mlubin: Unless things have changed recently, splatting large arrays can cause big performance issues when the dimension is large, e.g., into the thousands. ``collect(j)`` might be better here

src/operators.jl
```diff
@@ -569,18 +570,6 @@ end; end
 (/){T<:JuMPTypes}(lhs::SparseMatrixCSC{T}, rhs::Number) =
     SparseMatrixCSC(lhs.m, lhs.n, copy(lhs.colptr), copy(lhs.rowval), lhs.nzval ./ rhs)
 
-# The following are primarily there for internal use in the macro code for @constraint
-for op in [:(+), :(-)]; @eval begin
```
mlubin: Are these methods now covered by built-ins in Julia?

src/operators.jl
tkoolen: `collect(::OffsetArray)` returns a copy of the `OffsetArray` (with the same indices). I found it surprisingly hard to convert from an `OffsetArray` to a regular `Array`. Maybe I'm missing something.

rdeits: How about [v[i] for i in eachindex(v)] ?

src/operators.jl
tkoolen: Yes. https://github.com/JuliaLang/julia/blob/bdef3b57d34ec90781d30f62adf73d048e4a3090/base/arraymath.jl#L36 on current master, https://github.com/JuliaLang/julia/blob/v0.5.0/base/arraymath.jl#L49 on 0.5.

src/operators.jl
tkoolen: D'oh. Yeah, that's better.

src/operators.jl
mlubin: Ok, thanks for checking

mlubin:
Getting closer. Please correct the travis failures as well. (Appveyor failures are expected.)

tkoolen:
More fixes.

Two random comments on getting things to work for OffsetArrays, for future reference:
* the multi-argument `eachindex` is pretty useless when dealing with `OffsetArray`s. See https://github.com/alsam/OffsetArrays.jl/issues/24.
* even in 0.6, `diagm` doesn't work for `OffsetArrays`. See https://github.com/alsam/OffsetArrays.jl/issues/25.

src/affexpr.jl
```diff
@@ -181,13 +181,15 @@ function addconstraint(m::Model, c::LinearConstraint)
...
 end
-addconstraint(m::Model, c::Array{LinearConstraint}) =
+addconstraint(m::Model, c::AbstractArray{LinearConstraint}) =
     error("The operators <=, >=, and == can only be used to specify scalar constraints. If you are trying to add a vectorized constraint, use the element-wise dot comparison operators (.<=, .>=, or .==) instead")
 
-function addVectorizedConstraint(m::Model, v::Array{LinearConstraint})
-    ret = Array{LinConstrRef}(size(v))
-    for I in eachindex(v)
-        ret[I] = addconstraint(m, v[I])
+function addVectorizedConstraint(m::Model, v::AbstractArray{LinearConstraint})
```
joehuchette: I'm somewhat uncertain if this method should be changed. Is there a compelling reason to pass a sparse matrix of linear constraints, for example?

tkoolen:
Travis failures with no output seem to be because Amazon S3 is having issues: https://www.traviscistatus.com/incidents/hmwq9yy5dh9d.

mlubin:
I'll restart the builds when travis stabilizes a bit.

src/operators.jl
```diff
@@ -277,7 +277,8 @@ Base.sum(j::JuMPDict)  = sum(values(j.tupledict))
 Base.sum(j::JuMPArray{Variable}) = AffExpr(vec(j.innerArray), ones(length(j.innerArray)), 0.0)
 Base.sum(j::JuMPDict{Variable})  = AffExpr(collect(values(j.tupledict)), ones(length(j.tupledict)), 0.0)
 Base.sum(j::Array{Variable}) = AffExpr(vec(j), ones(length(j)), 0.0)
-function Base.sum{T<:GenericAffExpr}(affs::Array{T})
+Base.sum(j::AbstractArray{Variable}) = sum([j[i] for i in eachindex(j)]) # to handle non-one-indexed arrays.
```
joehuchette: As long as ``vec(::AbstractArray{Variable})`` will reliably work, I think the definition [above](https://github.com/JuliaOpt/JuMP.jl/pull/980/files#diff-188d49b6a9fe32e3c270a948c830cbcfR279) should work fine.

joehuchette:
Modulo my iffy-ness about changing ``addVectorizedConstraint``, this PR looks good to me.

src/affexpr.jl
tkoolen: Without changing this method, the following fails:
```julia
m = Model()
v = @variable(m, [1:3])
x = OffsetArray(v, -3)
@constraint(m, x .== 0)
```
because there is no `Convert` method that constructs an `Array` from an `OffsetArray` (a conscious design decision in OffsetArrays).

mlubin: This doesn't seem like the right fix. I think ``constructconstraint!(x::Array, sense::Symbol) = map(c->constructconstraint!(c,sense), x)`` should be changed to return a flat ``Vector`` of constraints. Then there's no need to touch ``addVectorizedConstraint``.

joehuchette: Since this also doesn't work:
```jl
t = OffsetArray(rand(3), -3)
t .== 0
```
I'm not sure this is desirable behavior.

src/operators.jl
tkoolen: I tried that first, but `AffExpr` is a type alias of `GenericAffExpr{Float64,Variable}`, so that directly calls the inner constructor of `AffExpr`, which expects precisely a `Vector{Variable}`. I tried adding an inner constructor to `GenericAffExpr` that takes `AbstractArray`s (+ additional outer constructors), but ran into some more issues that I don't remember anymore. Basically, it was going to be a pretty big change, whereas `sum` seemed to be the only current case where handling `AbstractVector`s was needed when constructing `AffExpr`s.

joehuchette: Cool, thanks for the explanation.

src/affexpr.jl
tkoolen: OK, will change. @joehuchette, that seems like a bug in Base.

tkoolen: Done.

src/macros.jl
```diff
@@ -305,20 +305,21 @@ function constructconstraint!(normexpr::SOCExpr, sense::Symbol)
 end
 
 constructconstraint!(x::Array, sense::Symbol) = map(c->constructconstraint!(c,sense), x)
+constructconstraint!(x::AbstractArray, sense::Symbol) = constructconstraint!([x[i] for i in eachindex(x)], sense::Symbol)
```
joehuchette: the second ``::Symbol`` shouldn't be necessary

joehuchette:
Looks good to merge to me. @mlubin?

mlubin:
I'm having trouble viewing the status of travis.

mlubin:
@tkoolen, thanks for addressing all of our concerns. Looking forward to seeing the uses of ``AxisArray``s.

tkoolen:
Awesome, thanks guys!

joehuchette:
Thanks for this, @tkoolen!

PR #987, opened 2017-03-07
Don't print that annoying 'free' for free variables

joehuchette:
This always annoyed me


mlubin:
Ok with me but doesn't pass tests.

joehuchette:
Travis failure is something with Ipopt on OSX.

PR #992, opened 2017-03-10
Add AppVeyor's badges

blegat:


mlubin:
We're avoided adding these because appveyor more often than not is failing for a reason that's not an issue with JuMP.

blegat:
Makes sense, we see it on the PR anyway

PR #1009, opened 2017-04-25
[WIP] add support for general conic constraints

chriscoey:
technically could work for general sets too, if you overload the interpretation of MPB conic model format


chriscoey:
duals are not set up

chriscoey:
related to https://github.com/JuliaOpt/MathProgBase.jl/issues/168, except naming is specific to cones. want more general sets, so this will have to be reworked

PR #1012, opened 2017-05-02
Refactor conicdata

blegat:
Related to #1009 
Refactor `conicdata` and add the `checkowned` option. Even if the `checkowned` option will always be `true` in JuMP, extensions such as StructJuMP may want to set it to `false` so that the constraint matrix only contains the columns of the variables of the model given as `m` but doesn't error if it sees variables from an other model.
Note that the code of [this file](https://github.com/StructJuMP/StructJuMP.jl/blob/616588ae3ce6b3f564fdb1ecfbff51bedae2cce4/src/BendersBridge.jl) and of [this file](https://github.com/blegat/StochasticDualDynamicProgramming.jl/blob/13befeb2840ce2619f19cb52068063b77c13b5e9/src/interface.jl) can now be greatly trimmed down thanks to this.

### DIFF
--- a/src/solvers.jl
+++ b/src/solvers.jl
@@ -573,17 +573,18 @@ function vartypes_without_fixed(m::Model)
     return colCats
 end
 
-function collect_expr!(m, tmprow, terms::AffExpr)
+function collect_expr!(m, tmprow, terms::AffExpr, checkowned=true)
     empty!(tmprow)
     assert_isfinite(terms)
     coeffs = terms.coeffs
     vars = terms.vars
     # collect duplicates
     for ind in 1:length(coeffs)
-        if vars[ind].m !== m
+        if is(vars[ind].m, m)
+            addelt!(tmprow, vars[ind].col, coeffs[ind])
+        elseif checkowned
             error("Variable not owned by model present in constraints")
         end
-        addelt!(tmprow,vars[ind].col, coeffs[ind])
     end
     rmz!(tmprow)
     tmprow
@@ -614,12 +615,9 @@ function offdiagsdpvars(m::Model)
     return offdiagvars
 end
 
-function conicdata(m::Model)
-    var_cones = Any[cone for cone in m.varCones]
-    con_cones = Any[]
-    nnz = 0
-
+function getSDPRowsInfo(m::Model)
     # find starting column indices for sdp matrices
+    nnz = 0
     numSDPRows = 0
     numSymRows = 0
     for c in m.sdpconstr
@@ -635,9 +633,10 @@ function conicdata(m::Model)
             numSymRows += convert(Int, n*(n-1)/2)
         end
     end
+    numSDPRows, numSymRows, nnz
+end
 
-    linconstr = m.linconstr::Vector{LinearConstraint}
-    numLinRows = length(linconstr)
+function var_range2cone!(var_cones, m::Model)
     numBounds = 0
     nonNeg  = Int[]
     nonPos  = Int[]
@@ -646,7 +645,7 @@ function conicdata(m::Model)
     for i in 1:m.numCols
         seen = false
         lb, ub = m.colLower[i], m.colUpper[i]
-        for (_,cone) in m.varCones
+        for (_, cone) in m.varCones
             if i in cone
                 seen = true
                 @assert lb == -Inf && ub == Inf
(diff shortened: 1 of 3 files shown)


src/solvers.jl
```diff
@@ -573,17 +573,18 @@ function vartypes_without_fixed(m::Model)
     return colCats
 end
 
-function collect_expr!(m, tmprow, terms::AffExpr)
+function collect_expr!(m, tmprow, terms::AffExpr, checkowned=true)
```
mlubin: Could you document what is supposed to happen when ``checkowned=false``? Why is it useful to have an option to ignore terms in the ``AffExpr``?

mlubin: I'd also rename ``checkowned`` to ``ignore_not_owned`` or something similar. ``checkowned=false`` implies to me that terms not owned will be added, not ignored.

mlubin:
Could you add in-line comments about the design of this (extensible) interface and how people are supposed to use it?

src/solvers.jl
blegat: True, that's a better naming.
When you have a constraint with variables from two models, calling `fillContrLHS!(I1, J1, V2, ..., m1, false)` and then `fillContrLHS!(I2, J2, V2, ..., m2, false)` will build two separate matrices. That is useful in StructJuMP for example. I will add some comments :)

blegat:
I was also thinking. Wouldn't it be more readable if we split all SOC related function in a file `socp.jl` and all SDP related function in a file `sdp.jl` ?

mlubin:
> I was also thinking. Wouldn't it be more readable if we split all SOC related function in a file socp.jl and all SDP related function in a file sdp.jl ?

We could at least move the generation code for all cones into ``cones.jl`` analogous to ``nlp.jl``. Not sure if each cone needs a separate file.

blegat:
I would also put the `getvalue`, `getdual`, ... that is why I was thinking about separate files. And it kind of show what needs to be implemented for each cone and that the code of different cones are uncorrelated.

mlubin:
That makes sense also. How does this reorganization fit in with general cone constraints (https://github.com/JuliaOpt/JuMP.jl/pull/1009)?

src/solvers.jl
```diff
@@ -573,17 +573,18 @@ function vartypes_without_fixed(m::Model)
...
     empty!(tmprow)
     assert_isfinite(terms)
     coeffs = terms.coeffs
     vars = terms.vars
     # collect duplicates
     for ind in 1:length(coeffs)
-        if vars[ind].m !== m
+        if is(vars[ind].m, m)
```
joehuchette: Trivial, but I find ``vars[ind].m === m`` clearer

src/solvers.jl
```diff
@@ -614,12 +615,9 @@ function offdiagsdpvars(m::Model)
     return offdiagvars
 end
 
-function conicdata(m::Model)
-    var_cones = Any[cone for cone in m.varCones]
-    con_cones = Any[]
-    nnz = 0
-
+function getSDPRowsInfo(m::Model)
```
joehuchette: We've been trying to move away from CamelCase if we can avoid it :)

src/solvers.jl
```diff
@@ -635,9 +633,10 @@ function conicdata(m::Model)
             numSymRows += convert(Int, n*(n-1)/2)
         end
     end
+    numSDPRows, numSymRows, nnz
+end
 
-    linconstr = m.linconstr::Vector{LinearConstraint}
-    numLinRows = length(linconstr)
+function var_range2cone!(var_cones, m::Model)
```
joehuchette: ``var_range_to_cone!``?

joehuchette: or ``variable_range_to_cone!``. No need to short names here.

src/solvers.jl
```diff
@@ -685,101 +684,28 @@ function conicdata(m::Model)
...
-    sizehint!(I, nnz)
-    sizehint!(J, nnz)
-    sizehint!(V, nnz)
-
-    # Fill it up
-    nnz = 0
-    tmprow = IndexedVector(Float64,m.numCols)
-    tmpelts = tmprow.elts
-    tmpnzidx = tmprow.nzidx
+function fillConstrRHS!(b, con_cones, c, linconstr::Vector{LinearConstraint})
```
joehuchette: Same re. CamelCase

src/solvers.jl
```diff
@@ -872,37 +844,138 @@ function conicdata(m::Model)
...
-    # Objective coefficients and columns of A matrix are
-    # rescaled for SDP variables
-    offdiagvars = offdiagsdpvars(m)
-    f[offdiagvars] /= sqrt(2)
-    for k in 1:length(J)
-        if offdiagvars[J[k]]
-            V[k] /= sqrt(2)
-        end
-    end
+    rescaleSDcols!(f, J, V, m)
```
joehuchette: Is this the first step in your master plan to getting rid of the ``sqrt(2)`` scaling? :)

blegat:
Yep, in #1009 we could do something like `c, d = fillConstr!(I, J, V, b, con_cones, tmprow, constr_dual_map, c, d, m.conicconstr, m)`. However, it would call the same function for all conicconstr which we should dispatch differently for each constr.
So in `fillConstr!` we may do something like
```julia
for con in constrs
    fillConstr!(..., con.terms, Val{con.cone}())
end
```
and add a `fillConstr!` method for each cone. The `Val` is just there to make it dispatch on the symbol but if we use objects instead of symbol it is not needed.

src/solvers.jl
blegat: Do you prefer nothing or underscore ? `getSDrowsinfo` or `get_SD_rows_info` ?

src/solvers.jl
blegat: Currently it just seems like moving code out of the `conicdata` but indeed it is actually part of the pan to remove the `sqrt(2)` scaling which is itself part of the plan to remove the dependency of JuMP on `Float64` ! :-P

src/solvers.jl
joehuchette: I'm fine with the first

blegat:
@mlubin @joehuchette your comments were addressed in db6fce9 :)

src/solvers.jl
```diff
@@ -573,17 +573,21 @@ function vartypes_without_fixed(m::Model)
...
     empty!(tmprow)
     assert_isfinite(terms)
     coeffs = terms.coeffs
     vars = terms.vars
     # collect duplicates
     for ind in 1:length(coeffs)
-        if vars[ind].m !== m
+        if vars[ind].m === m
+            addelt!(tmprow, vars[ind].col, coeffs[ind])
+        elseif checkowned
```
mlubin: ``checkowned`` not defined. You should make sure this branch is covered by a test.

blegat: Looking at codecov, it seems it wasn't covered by the tests. I have added a test in c715a24

src/solvers.jl
```diff
@@ -791,49 +725,108 @@ function conicdata(m::Model)
...
+        end
+        push!(con_cones, (:SOC, soc_start:c))
+    end
+    c
+end
+
+# Represents the constraints `constr` as
+# `b - Ax ∈ K`,
+# writes the matrix `A` of this representation in the sparse format using `I`, `J` and `V`,
+# starting at row index `c+1`, and specifies the indices of each constraint in `constr_dual_map`,
```
mlubin: I find this explanation of ``constr_dual_map`` confusing, and also the name could use some work. It's the map from the constraint index to the rows in the matrix, right? Why does ``c != d``?

blegat: I agree the name can be improved (a quick git blame show that the name was coined by @emreyamangil).
`c` is the current row in the matrix while `d` is the current index in the vector of constraints stored in the JuMP model. So for example a SOC constraint increments `d` only once while it intcrements `c` `n+1` times.

mlubin: ``constr_to_row``?

blegat: Ok for `constr_to_row`, I renamed it in 8dda2cd

mlubin:
Looks pretty good to me now

blegat:
@joehuchette Indeed, splitting things can go in another PR.

chriscoey:
Good stuff

PR #1022, opened 2017-05-11
Fix #1021

### LINKED ISSUE #1021: Inconsistency between getobjbound() and getobjectivevalue()
Small example:

```julia
using JuMP, Gurobi
m = Model(solver=GurobiSolver())
@variable(m, 0 <= x <= 5)
@variable(m, y, Bin)
@constraint(m, x >= y)
@objective(m, Min, x - 5)

solve(m)

println("objective = ",getobjectivevalue(m))
println("objective bound = ",getobjbound(m))
```

Output:
```julia
julia> include("test.jl")
Optimize a model with 1 rows, 2 columns and 2 nonzeros
Variable types: 1 continuous, 1 integer (1 binary)
Coefficient statistics:
  Matrix range     [1e+00, 1e+00]
  Objective range  [1e+00, 1e+00]
  Bounds range     [1e+00, 5e+00]
  RHS range        [0e+00, 0e+00]
Found heuristic solution: objective 0

Explored 0 nodes (0 simplex iterations) in 0.00 seconds
Thread count was 1 (of 8 available processors)

Solution count 1: 0 
Pool objective bound 0

Optimal solution found (tolerance 1.00e-04)
Best objective 0.000000000000e+00, best bound 0.000000000000e+00, gap -
objective = -5.0
objective bound = 0.0
```

Gurobi (or JuMP?) is ignoring the constant (-5) term in the objective. The constant is added back in for getobjectivevalue() but is not for getobjbound().

I tested this with Julia 0.5.0 (Mac), MathProgBase 0.6.4, and JuMP 0.16.2.

blegat:


mlubin:
This is not correct because the user may set a new objective function with a different constant after solving.

mlubin:
The right fix is probably to store a new ``objbound`` field inside the JuMP model and update it after the solve, as we do for the objective value.

src/JuMP.jl
```diff
@@ -215,6 +217,16 @@ for f in MathProgBase.SolverInterface.methods_by_tag[:rewrap]
     end
     eval(Expr(:export,f))
 end
+import MathProgBase.getobjbound
+function getobjbound(m::Model)
```
mlubin: It's inconsistent to confuse the solver's ``getobjbound`` and JuMP's ``getobjbound`` here when we have separate names for ``MathProgBase.getobjval`` and ``JuMP.getobjectivevalue``. We should leave ``getobjbound`` alone and implement a new ``JuMP.getobjectivebound``.

src/solvers.jl
```diff
@@ -252,6 +254,7 @@ function solve(m::Model; suppress_warnings=false,
                 colVal[offdiagvars] /= sqrt(2)
             end
             # Don't corrupt the answers if one of the above two calls fails
+            m.objBound = objBound
```
mlubin: Also need to set though the ``Nonlinear`` solver pathway.

src/solvers.jl
```diff
@@ -244,6 +245,7 @@ function solve(m::Model; suppress_warnings=false,
     # limit or tolerance is set (:UserLimit)
     if !(stat == :Infeasible || stat == :Unbounded)
         try
+            objBound = MathProgBase.getobjbound(m.internalModel) + m.obj.aff.constant
```
mlubin: I think this means that we'll silently fail to set the objective value and solution if the solver doesn't implement ``getobjbound``. Make sure you test this on a solver that doesn't implement ``getobjbound`` like Ipopt. Either we need have a special check to see if the solver implements it or you'll have the pleasure of going through every solver package to implement ``getobjbound``.

blegat: Can't we just return `NaN` if the solver does not implements it ?

mlubin: We can, but it would be nice to start documenting these implicit behaviors in MathProgBase.

doc/refmodel.rst
```diff
@@ -66,6 +66,7 @@ Methods
 * ``getobjectivesense(m::Model)`` - returns objective sense, either ``:Min`` or ``:Max``.
 * ``setobjectivesense(m::Model, newSense::Symbol)`` - sets the objective sense (``newSense`` is either ``:Min`` or ``:Max``).
 * ``getobjectivevalue(m::Model)`` - returns objective value after a call to ``solve``.
+* ``getobjectivevalue(m::Model)`` - returns the best known bound on the optimal objective value after a call to ``solve``.
```
mlubin: typo

src/JuMP.jl
```diff
@@ -216,6 +218,15 @@ for f in MathProgBase.SolverInterface.methods_by_tag[:rewrap]
     eval(Expr(:export,f))
 end
 
+function getobjectivebound(m::Model)
+    # check internal model exists
+    if !m.internalModelLoaded
```
mlubin: This check isn't needed. Just implement a method consistent with ``getobjectivevalue``.

PR #1023, opened 2017-05-12
Fix 'Variable not owned' for conic

blegat:


src/solvers.jl
```diff
@@ -585,7 +585,7 @@ function collect_expr!(m::Model, tmprow::IndexedVector, terms::AffExpr, ignore_n
     for ind in 1:length(coeffs)
         if vars[ind].m === m
             addelt!(tmprow, vars[ind].col, coeffs[ind])
-        elseif ignore_not_owned
+        elseif !ignore_not_owned
```
mlubin: why does this chance only affect conic?

test/model.jl
```diff
@@ -1001,10 +1001,12 @@ end
...
-        M = Model()
-        @variable(m, x)
-        @constraint(M, x == 1)
-        @test_throws ErrorException solve(M)
+        if !isempty(soc_solvers)
+            m = Model()
+            M = Model(solver=first(soc_solvers))
+            @variable(m, x[1:3])
+            @constraint(M, norm(x[1:2]) <= x[3])
+            @test_throws ErrorException solve(M)
```
mlubin: There are lots of errors that can happen on a call to solve. It would be nice to test the specific error message.

src/solvers.jl
blegat: Yes because it is the only one using `collect_expr!`

test/model.jl
```diff
@@ -1001,10 +1001,12 @@ end
     end
 
     @testset "Variable not owned by the model" begin
-        m = Model()
-        M = Model()
-        @variable(m, x)
-        @constraint(M, x == 1)
-        @test_throws ErrorException solve(M)
+        if !isempty(soc_solvers)
```
mlubin: Could we test this in both the LP and conic contexts?

PR #1026, opened 2017-05-13
Apply lb/ub check for bin vars specified by category kw

blegat:


mlubin:
Eh, this won't catch the case
```julia
cat = :Bin
@variable(m, 2 <= x <= 3, category=cat)
```
and it will incorrectly throw an error on:
```julia
l = 0
u = 1
@variable(m, l <= x <= u, category=:Bin)
```
We might need to rethink whether we should be checking bounds at all in the macro.

blegat:
Indeex it might be more consistent not to throw anything and just le the
solver returns the :Infeasible status.

Le 13 mai 2017 21:40, "Miles Lubin" <notifications@github.com> a écrit :

Eh, this won't catch the case

cat = :Bin@variable(m, 2 <= x <= 3, category=cat)

and it will incorrectly throw an error on:

l = 0
u = 1@variable(m, l <= x <= u, category=:Bin)

We might need to rethink whether we should be checking bounds at all in the
macro.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/JuliaOpt/JuMP.jl/pull/1026#issuecomment-301270373>, or mute
the thread

mlubin:
> it might be more consistent not to throw anything and just le the
solver returns the :Infeasible status.

At some point a long time ago we considered that and decided against it, but I'm willing to try that and see what breaks. I think some solvers don't behave in the way that you would expect, but in any case it would be better to do this consistency check at build time and not inside the macros.

blegat:
Here is what happens:

| Solver | Result |
| ------------- | ------------- |
| CPLEX, Cbc |  `:Infeasible` with the warning "Infeasibility ray (Farkas proof) not available" |
| Mosek | `:Unknown` because of [this line](https://github.com/JuliaOpt/Mosek.jl/blob/15f31e98c5023fa016e9ab92b40a379593054449/src/MosekSolverInterface.jl#L32) |
| GLPK | `:Error` with the printing "glp_simplex: column 1: lb = 2, ub = 1; incorrect bounds", we might do something similar to https://github.com/JuliaOpt/GLPKMathProgInterface.jl/pull/34 |

joehuchette:
Agreed that this check would be better at solve time

mlubin:
See the tests failing. I think we should still set the lower bound to 0 and upper bound to 1 if the user hasn't specified other bounds.

blegat:
The question is: should we do it in the variable constructor, e.g.
```julia
lower = max(lower, 0)
upper = min(upper, 1)
@assert lower == 0
@assert upper == 1
```

mlubin:
Yes, the constructor is a reasonable place.

mlubin:
Seems ok now but please uncomment all of the tests :)

PR #1028, opened 2017-05-16
WIP: Overload getindex to allow variables to be accessed with []

odow:
@reganbaucke proposed the following as a nicer way to associate variables with models.

It's particularly attractive if you are copying models around and losing track of which scope `x` is in.

### DIFF
--- a/doc/refmodel.rst
+++ b/doc/refmodel.rst
@@ -57,7 +57,8 @@ Methods
 * ``solve(m::Model; suppress_warnings=false, relaxation=false)`` - solves the model using the selected solver (or a default for the problem class), and takes two optional arguments that are disabled by default. Setting ``suppress_warnings`` to ``true`` will suppress all JuMP-specific output (e.g. warnings about infeasibility and lack of dual information) but will not suppress solver output (which should be done by passing options to the solver). Setting ``relaxation=true`` solves the standard continuous relaxation for the model: that is, integrality is dropped, special ordered set constraints are not enforced, and semi-continuous and semi-integer variables with bounds ``[l,u]`` are replaced with bounds ``[min(l,0),max(u,0)]``.
 * ``JuMP.build(m::Model)`` - builds the model in memory at the MathProgBase level without optimizing.
 * ``setsolver(m::Model,s::AbstractMathProgSolver)`` - changes the solver which will be used for the next call to ``solve()``, discarding the current internal model if present.
-* ``getindex(m::Model,name::Symbol)`` - returns the variable, or group of variables, or constraint, or group of constraints, of the given name which were added to the model.
+* ``getindex(m::Model,name::Symbol)`` - returns the variable, or group of variables, or constraint, or group of constraints, of the given name which were added to the model. This errors if multiple variables or constraints share the same name.
+* ``setindex!(m::Model, value, name::Symbol)`` - stores the object ``value`` in the model ``m`` so that it can be accessed via ``getindex``.
 
 **Objective**
 
--- a/doc/refmodel.rst
+++ b/doc/refmodel.rst
@@ -57,8 +57,7 @@ Methods
 * ``solve(m::Model; suppress_warnings=false, relaxation=false)`` - solves the model using the selected solver (or a default for the problem class), and takes two optional arguments that are disabled by default. Setting ``suppress_warnings`` to ``true`` will suppress all JuMP-specific output (e.g. warnings about infeasibility and lack of dual information) but will not suppress solver output (which should be done by passing options to the solver). Setting ``relaxation=true`` solves the standard continuous relaxation for the model: that is, integrality is dropped, special ordered set constraints are not enforced, and semi-continuous and semi-integer variables with bounds ``[l,u]`` are replaced with bounds ``[min(l,0),max(u,0)]``.
 * ``JuMP.build(m::Model)`` - builds the model in memory at the MathProgBase level without optimizing.
 * ``setsolver(m::Model,s::AbstractMathProgSolver)`` - changes the solver which will be used for the next call to ``solve()``, discarding the current internal model if present.
-* ``getvariable(m::Model,name::Symbol)`` - returns the variable or group of variables of the given name which were added to the model with ``@variable``. Throws an error if multiple variables were created with the same name.
-* ``getconstraint(m::Model,name::Symbol)`` - returns the constraint or group of constraint of the given name which were added to the model with ``@constraint`` or ``@NLconstraint``. Throws an error if multiple constraints were created with the same name.
+* ``getindex(m::Model,name::Symbol)`` - returns the variable, or group of variables, or constraint, or group of constraints, of the given name which were added to the model.
 
 **Objective**
 
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -835,6 +835,9 @@ function getvariable(m::Model, varname::Symbol)
     end
 end
 
+# allow easy accessing of JuMP Variables
+Base.getindex(m::JuMP.Model, varname::Symbol) = getvariable(m, varname)
+
 function getconstraint(m::Model, conname::Symbol)
     if !haskey(m.conDict, conname)
         error("No constraint with name $conname")
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -823,6 +823,12 @@ function Base.getindex(m::JuMP.Model, name::Symbol)
         return m.objDict[name]
     end
 end
+function Base.setindex!(m::JuMP.Model, value, name::Symbol)
+    if haskey(m.objDict, name)
+        error("There is already an object in the model with the name $name")
+    end
+    m.objDict[name] = value
+end
 
 # usage warnings
 function mapcontainer_warn(f, x::JuMPContainer, var_or_expr)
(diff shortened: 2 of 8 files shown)


mlubin:
Mmmm... someone's going to come along and ask why this only works for variables and not constraints. This seems like a reasonable direction but we need a more well thought out concept of what's in the "scope" of a model and how ``getvariable`` and ``getconstraint`` relate.

odow:
@mlubin 

Four cases:

1. `name` is a variable but not a constraint in the model. `m[:name]` returns the variable
2. `name` is a constraint but not a variable in the model. `m[:name]` returns the constraint
3. `name` is a variable and a constraint in the model. `m[:name]` errors. Use `getvariable` or `getconstraint` instead
4. `name` is neither a variable nor a constraint in the model. `m[:name]` errors.

Multpiple constraints or variables with the same name are not accessible by their name to avoid ambiguity. Use anonymous variables instead.

mlubin:
Why not just get rid of (deprecate) ``getvariable`` and ``getconstraint`` and have a single dictionary of things that you can stuff into a JuMP model?
Then
```julia
@variable(m, x[1:N])
```
becomes a shorthand for
```julia
m[:x] = @variable(m, [1:N])
```
which seems nice and easy to explain.

odow:
So maybe something like this? I changed the duplicate warning to an error. There is no reason why people shouldn't just use anonymous variables.

mlubin:
> I changed the duplicate warning to an error. There is no reason why people shouldn't just use anonymous variables.

That's fine. The warning was there for deprecation purposes.

mlubin:
I like how this looks. Any reason not to implement ``setindex``?

odow:
You mean so people can do
```julia
m[:x] = @variable(m)
```

Seems reasonable.

mlubin:
Yes. People will misguidedly end up using this syntax to store cat pictures and databases inside the model, but I don't see a reason to get in the way of that.

joehuchette:
I like this, but I think the deprecated ``getvariable`` and ``getconstraint`` will have different behavior in cases like:
```jl
using JuMP
m = Model()
@variable(m, x)
c = getconstraint(m, :x)
```

mlubin:
@joehuchette, so? The main point is to not break existing code that works.

joehuchette:
This existing code breaks:
```jl
using JuMP
m = Model()
@variable(m, x)
@constraint(m, x, x == 1)
```

joehuchette:
My point being that mashing two dictionaries into one can lead to weird conflicts, especially when you're using keys programmatically.

odow:
We could change the error back to a warning and say 

```julia
if haskey(m.objDict, name)
    warn("Overwriting the object $name stored in the model. Consider using anonymous variables and constraints instead")
end
```

odow:
Although I think `@constraint(m, x, x==1)` is far too confusing and should be forbidden syntax.

mlubin:
I'm ok with breaking code that has a variable and a constraint with the same name. Throwing a warning in this case isn't the right answer because we can't guarantee that the behavior will remain the same as before.

joehuchette:
> I'm ok with breaking code that has a variable and a constraint with the same name.
I think this is a lot more common pattern than you're giving credit for (only so many 1-letter variable names...).

mlubin:
> I think this is a lot more common pattern than you're giving credit for (only so many 1-letter variable names...).

It doesn't bother me if it's common since I think this is a change in the right direction for the long term. If we want to maintain separate variables and constraints dictionaries for deprecation purposes that would be ok but not necessary in my view.

joehuchette:
> It doesn't bother me if it's common since I think this is a change in the right direction for the long term.

It's maybe not good practice, but I think name conflicts are inevitable when multiple people/packages are working on the same model, and that there's not much we can do to dissuade people from that.

> If we want to maintain separate variables and constraints dictionaries

I think this is probably the way to go.

mlubin:
> > If we want to maintain separate variables and constraints dictionaries

> I think this is probably the way to go.

This means that we can't throw an error when there's a conflicting name in the new ``objDict``, otherwise there's no point in maintaining separate dictionaries for legacy reasons.

joehuchette:
The behavior of ``[]`` and ``getvariable`` don't have to coincide, I just don't think breaking existing code is a good idea.

odow:
There have been plenty of other times that JuMP changes broke existing code. The solution is pretty simple for current users who might use the functionality
```julia
m=Model()
@variable(m, x)

# change
@constraint(m, x, x<=1)
# to
x = @constraint(m, x<=1)
```

Long-term, I think this is simpler, there is less confusion around scoping, and it means (for better or worse), that people can store data in `m[key]`.

joehuchette:
I agree that we should push users to make this change. Also, to be clear, I'm more concerned about cases like:

```jl
m = Model()
@variable(m, x[1:1_000])
@variable(m, c) # cost, or something
add_other_constraints!(m)

# deep in someone elses code/package...
function add_other_constraints!(model)
    ...
    @constraint(model, c[1:100], ...)
    ...
end
```

I agree that ``@constraint(m, x, x<=1)`` is pathological.

mlubin:
The ``# deep in someone elses code/package...`` code should absolutely not be using named variables.

mlubin:
CC @ccoffrin

joehuchette:
That's not spelled out anywhere in our documentation AFAIK. It's also perfectly valid code, and people are gonna write it.

mlubin:
@joehuchette, if we turn duplicate names into a warning and throw an error only when a duplicately named object is accessed, then it seems like it would avoid breaking most of the casual use cases. It would only break (gracefully, with a useful error) on your example above if someone tried ``getconstraint(model,:c)``.

joehuchette:
What about your concern from before? I think that's better, but I think keeping the old implementation with two dictionaries during the deprecation period is preferable.

src/JuMP.jl
```diff
@@ -804,45 +797,37 @@ end
...
 
-function getconstraint(m::Model, conname::Symbol)
-    if !haskey(m.conDict, conname)
-        error("No constraint with name $conname")
-    elseif m.conDict[conname] === nothing
-        error("Multiple constraints with name $conname")
+# allow easy accessing of JuMP Variables and Constraints
+function Base.getindex(m::JuMP.Model, name::Symbol)
+    if !haskey(m.objDict, name)
+        error("No object with name $name")
```
mlubin: It would be consistent to throw a ``KeyError`` here

mlubin:
> What about your concern from before?

Which concern? There is no possible way to avoid breaking your example if we throw an error on duplicate names.

joehuchette:
Yeah, I guess we could potentially make it throw an error upon trying to use ``[]`` to retrieve a duplicate name, but that's not worth the trouble. The warning is more appealing to me now.

mlubin:
What I'm proposing is a warning when creating duplicate names and an error when retrieving.

joehuchette:
In that case, the error might as well be restricted to ``[]``.

mlubin:
I will personally mail a JuliaOpt sticker to whoever gets hit by an error from
```julia
m = Model()
@variable(m, x[1:1_000])
@variable(m, c) # cost, or something
add_other_constraints!(m)

# deep in someone elses code/package...
function add_other_constraints!(model)
    ...
    @constraint(model, c[1:100], ...)
    ...
end
getconstraint(m, :c)
```

joehuchette:
Famous last words!

mlubin:
It's on the record now :)

ccoffrin:
I am curious about the comment,
```
The # deep in someone else's code/package... code should absolutely not be using named variables.
```
Why not?  And what are the alternatives?

One of the software design challenges I am facing is how two more-or-less independent model building packages can be composed, esp. avoiding variable name clashes in a shared JuMP model.

blegat:
People might also like to use named variables because that gives pretty printing.
```julia
m = Model()
@variable(m, [1:2])
print(m)
```
gives
```
Min 0
Subject to
 __anon__[i] ∀ i ∈ {1,2} -> not as nice as x[i]
```

mlubin:
> One of the software design challenges I am facing is how two more-or-less independent model building packages can be composed, esp. avoiding variable name clashes in a shared JuMP model.

Each model building package should maintain its own "namespace" by storing extra data in the model's extension dictionary or some other mechanism.

> People might also like to use named variables because that gives pretty printing.

We have a separate mechanism to set names with the ``basename`` keyword.

joehuchette:
@ccoffrin one other motivation is that the anonymous syntax is hopefully clearer as to what's going on:  ``x = @variable(m)`` is very clearly an assignment to ``x``, which can be used programmatically like any other variable you might have in your julia code.

joehuchette:
This looks good to me now, thanks @odow!

doc/refmodel.rst
```diff
@@ -57,8 +57,7 @@ Methods
...
 * ``JuMP.build(m::Model)`` - builds the model in memory at the MathProgBase level without optimizing.
 * ``setsolver(m::Model,s::AbstractMathProgSolver)`` - changes the solver which will be used for the next call to ``solve()``, discarding the current internal model if present.
-* ``getvariable(m::Model,name::Symbol)`` - returns the variable or group of variables of the given name which were added to the model with ``@variable``. Throws an error if multiple variables were created with the same name.
-* ``getconstraint(m::Model,name::Symbol)`` - returns the constraint or group of constraint of the given name which were added to the model with ``@constraint`` or ``@NLconstraint``. Throws an error if multiple constraints were created with the same name.
+* ``getindex(m::Model,name::Symbol)`` - returns the variable, or group of variables, or constraint, or group of constraints, of the given name which were added to the model.
```
mlubin: document ``setindex!`` as well?

odow:
@mlubin you'd better get your stickers ready...

Issue #692, opened 2016-03-05
syntax brainstorming


mlubin:
Addressing the issue raised in https://github.com/JuliaOpt/JuMP.jl/pull/688, the current syntax of `@defVar`, despite the name, doesn't make it clear that all we're doing is defining a julia variable. Maybe it's time to brainstorm a new approach:

Both examples below are valid julia syntax:
Idea 1:

``` julia
@variables m begin
    x = (0 <= Variable[i=1:n](Int, start=5) <= i)
    ...
end
```

The trouble here is that it looks like you're assigning x to refer to the constraints enforcing the bounds instead of the variables.

Idea 2:

``` julia
# variant 1
@variables m begin
    x = Variable[i=1:n](Int, start=5,lowerbound=0,upperbound=i)
    ...
end
# variant 2
@variables m begin
    x[i=1:n] = Variable(Int, start=5,lowerbound=0,upperbound=i)
    ...
end
```

We lose some of the magic of using comparison operators, but maybe this is much clearer. You can always use explicit constraints instead if you prefer the mathy syntax. Of the two variants I prefer the first because it's more clear that we're just assigning a collection of variables to `x` whereas the second variant might lead you to think that you can append indices to `x` later on.

CC @StefanKarpinski @carnaval @yeesian @JackDunnNZ @davidanthoff @juan-pablo-vielma @chriscoey

davidanthoff:
I think I like idea 2 variant 1 best. Idea 1 seems too busy, and the syntax too weird, to me.

If you go with the ``x = Variable`` approach, it might be worth thinking about renaming the macro to something like ``@jump``, and then also allow lines that read like ``y = Parameter()``, and maybe even constraints like that? It seems weird to have to use a macro called variable, and then to specify again later that I am adding variables...

mlubin:
Yes, a single ``@jump`` macro is a natural conclusion from idea 2.

chriscoey:
I agree with David on idea 2 variant 1.

I also never liked the LB <= var[...] <= UB syntax in defvar - it messes up
vertical alignment (which helps when quickly glancing over a model) and
doesn't feel like a natural way to code.

joehuchette:
I prefer idea 2 variant 1 the most of the options presented, but still find it less satisfying than the current syntax. We lose the bounds for syntax which I like very much, and although the assignment is a bit clearer with the ``=``, it's still unlike any other assignment you would be able to write in julia (without macros).

[...]

PR #1029, opened 2017-05-16
Add the constructvariable!/variabletype functions

blegat:
This allows extensions to use the @variable macro to create other types
of variables. For example, in PolyJuMP, we could add the syntax
```julia
@variable(m, p, Poly(X))
@variable(m, p >= 0, Poly(X))
# or
@variable(m, p, :Poly, monomials=X)
@variable(m, p >= 0, :Poly, monomials=X)
```
where X is the vector of monomials

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -815,18 +815,18 @@ esc_nonconstant(x) = esc(x)
 variabletype(m::Model) = Variable
 # Returns a new variable belonging to the model `m`. Additional positional arguments can be used to dispatch the call to a different method.
 # The return type should only depends on the positional arguments for `variabletype` to make sense.
-function constructvariable!(m::Model; _error::Function=error, lowerbound::Number=-Inf, upperbound::Number=Inf, category::Symbol=:Default, objective::Number=0., inconstraints::Vector=[], coefficients::Vector{Float64}=Float64[], basename::AbstractString="", start::Number=NaN, extra_kwargs...)
+function constructvariable!(m::Model, _error::Function, lowerbound::Number, upperbound::Number, category::Symbol, objective::Number, inconstraints::Vector, coefficients::Vector{Float64}, basename::AbstractString, start::Number; extra_kwargs...)
     for (kwarg, _) in extra_kwargs
         _error("Unrecognized keyword argument $kwarg")
     end
-    if category == :Default
-        category = :Cont
+    Variable(m, lowerbound, upperbound, category == :Default ? :Cont : category, objective, inconstraints, coefficients, basename, start)
+end
+
+function constructvariable!(m::Model, _error::Function, lowerbound::Number, upperbound::Number, category::Symbol, basename::AbstractString, start::Number; extra_kwargs...)
+    for (kwarg, _) in extra_kwargs
+        _error("Unrecognized keyword argument $kwarg")
     end
-    if objective == 0 && isempty(inconstraints)
-        Variable(m, lowerbound, upperbound, category, basename, start)
-    else
-        Variable(m, lowerbound, upperbound, category, objective, inconstraints, coefficients, basename, start)
-    end
+    Variable(m, lowerbound, upperbound, category == :Default ? :Cont : category, basename, start)
 end
 
 const EMPTYSTRING = ""
@@ -1020,7 +1020,7 @@ macro variable(args...)
         _error("Can only create one variable at a time when adding to existing constraints.")
 
         return assert_validmodel(m, quote
-            $variable = constructvariable!($m, $(extra...); _error=$_error, lowerbound=$lb, upperbound=$ub, category=$t, objective=$obj, inconstraints=$inconstraints, coefficients=$coefficients, basename=string($quotvarname), start=$value, $(extra_kwargs...))
+            $variable = constructvariable!($m, $(extra...), $_error, $lb, $ub, $t, $obj, $inconstraints, $coefficients, string($quotvarname), $value; $(extra_kwargs...))
             $(anonvar ? variable : :($escvarname = $variable))
         end)
     end
@@ -1028,7 +1028,7 @@ macro variable(args...)
     if isa(var,Symbol)
         # Easy case - a single variable
         sdp && _error("Cannot add a semidefinite scalar variable")
-        code = :($variable = constructvariable!($m, $(extra...); _error=$_error, lowerbound=$lb, upperbound=$ub, category=$t, basename=string($quotvarname), start=$value, $(extra_kwargs...)))
+        code = :($variable = constructvariable!($m, $(extra...), $_error, $lb, $ub, $t, string($quotvarname), $value); $(extra_kwargs...))
         if !anonvar
             code = quote
                 $code
@@ -1046,7 +1046,7 @@ macro variable(args...)
     clear_dependencies(i) = (isdependent(idxvars,idxsets[i],i) ? () : idxsets[i])
 
     # Code to be used to create each variable of the container.
-    code = :( $(refcall) = constructvariable!($m, $(extra...); _error=$_error, lowerbound=$lb, upperbound=$ub, category=$t, basename=EMPTYSTRING, start=$value, $(extra_kwargs...)) )
+    code = :( $(refcall) = constructvariable!($m, $(extra...), $_error, $lb, $ub, $t, EMPTYSTRING, $value; $(extra_kwargs...)) )
     # Determine the return type of constructvariable!. This is needed to create the container holding them.
     vartype = :( variabletype($m, $(extra...)) )
 
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -822,6 +822,7 @@ function constructvariable!(m::Model, _error::Function, lowerbound::Number, uppe
     Variable(m, lowerbound, upperbound, category == :Default ? :Cont : category, objective, inconstraints, coefficients, basename, start)
 end
 
+variabletype(m::Model, _error::Function, lowerbound::Number, upperbound::Number, category::Symbol, basename::AbstractString, start::Number) = Variable
 function constructvariable!(m::Model, _error::Function, lowerbound::Number, upperbound::Number, category::Symbol, basename::AbstractString, start::Number; extra_kwargs...)
     for (kwarg, _) in extra_kwargs
         _error("Unrecognized keyword argument $kwarg")
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -812,7 +812,7 @@ esc_nonconstant(x::Expr) = isexpr(x,:quote) ? x : esc(x)
 esc_nonconstant(x) = esc(x)
 
 # Returns the type of what `constructvariable!` would return with the same positional arguments.
-variabletype(m::Model) = Variable
+variabletype(m::Model, _error::Function, lowerbound::Number, upperbound::Number, category::Symbol, objective::Number, inconstraints::Vector, coefficients::Vector{Float64}, basename::AbstractString, start::Number) = Variable
 # Returns a new variable belonging to the model `m`. Additional positional arguments can be used to dispatch the call to a different method.
 # The return type should only depends on the positional arguments for `variabletype` to make sense.
 function constructvariable!(m::Model, _error::Function, lowerbound::Number, upperbound::Number, category::Symbol, objective::Number, inconstraints::Vector, coefficients::Vector{Float64}, basename::AbstractString, start::Number; extra_kwargs...)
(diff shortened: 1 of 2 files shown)


joehuchette:
Cool. I like the first syntax FWIW. Could also be ``@variable(m, p, Poly(monomials=X))``.

mlubin:
It's hard to keep track of what the ``@variable`` macro does at this point. I think we need to fully document what syntax the ``@variable`` macro supports before we can decide about extending it. Sorry to ask you to do more work, @blegat, but as we push to 1.0 we need to make sure that currently messy parts of the code like ``@variable`` become more well thought out, consistent, and easy to explain.

odow:
I think this needs to be done in conjunction with https://github.com/JuliaOpt/JuMP.jl/issues/692

```julia
@variable(m, x[i,j] in Int for i in I, j in J if cond(i,j))
@variable(m, x[i,j] in [l, u] for i in I, j in J if cond(i,j))
@variable(m, x[i,j] in {0} | [l, u] for i in I, j in J if cond(i,j))
@variable(m, x[i,j] >= 0 in Poly(X) for i in I, j in J if cond(i,j))
```

mlubin:
#692 is going to need a discussion at the developers meetup, and I'm not super convinced about it.
I'm open to non-breaking or incremental changes to the current syntax, but it needs to be well thought out and documented.

blegat:
I have added a developer documentation of `@variable` in 22bc8fe. I agree that transforming `Bin`, `Int`, ... before passing them to `constructvariable!` is a bit hacky since we do not apply the same treatments to the other elements of `extra`. As a non-breaking change, we could define the types `Bin`, `SemiCont`, ... however this may also look strange...

@odow I also like the `in` syntax but I have tried to make this PR as small as possible since it is already an arguable change. I was thinking to add the second argument of `in` at the beginning of `extra` so that `@variable(m, x[i,j] in Int)` would be equivalent to `@variable(m, x[i,j], Int)` but that would belong to a different PR.

src/macros.jl
```diff
@@ -811,11 +811,59 @@ esc_nonconstant(x::Number) = x
...
+#   additional information that will be passed as keyword arguments. The latter can always be given directly as keyword arguments to the macro.
+# * The `SDP` and `Symmetric` positional arguments in `extra` will not be passed to `constructvariable!`. Instead,
+#    * the `Symmetric` argument will check that the container is symmetric and only allocate one variable for each pair of non-diagonal entries.
+#    * the `SDP` argument will do the same as `Symmetric` but in addition it will specify that the variables created belongs to the SDP cone in the `varCones` field of the model.
+#   Moreover, if a category is passed in `extra` not as a symbol (e.g. `Bin` instead of `:Bin`), it will be transformed to a symbol before being
+#   passed to `constructvariable!`.
+# * The keyword arguments start, objective, inconstraints, coefficients, basename, lowerbound, upperbound, category may not be passed as is to
+#   `constructvariable!` since they may be altered by the parsing of `expr` and we may need to pass it pointwise if it is a container since
+#   `constructvariable!` is called separately for each variable of the container.
```
mlubin: Could you give a trivial, e.g. ``@variable(m, x >= 0)`` and nontrivial example, e.g., ``@variable(m, x[1:N,1:N], SDP, Poly())`` of how the macro creates calls to ``constructvariable!``?

src/macros.jl
```diff
@@ -811,11 +811,59 @@ esc_nonconstant(x::Number) = x
...
+    cat = t.parameters[1]
+    if category != :Default && category != cat
+        _error("A variable cannot be both of category $cat and $category. Please specify only one category.")
+    end
+    constructvariable!(m; category=cat, extra_kwargs...)
+end
+
+variabletype(m::Model, t::Symbol) = variabletype(m, Val{t})
+function constructvariable!(m::Model, t::Symbol; extra_kwargs...)
+    constructvariable!(m, Val{t}; extra_kwargs...)
```
mlubin: I'm slightly worried about the overhead in dynamic dispatch here if we're creating 1,000,000 ``:Cont`` variables in a loop. Are there any performance regressions?

blegat: Yes there might be since the method called cannot be determined only be the type of the arguments

mlubin: What about creating the ``Val{:Cont}`` (etc) types inside the macro when the variable category is known?

blegat: Then we might as well modify `category` directly. The whole point of this method is to support specifying a `Symbol` in `extra` but I start to think that this is not a good idea. In PolyJuMP for example since I will create the `Poly` type I might just as well ask to give `Poly` instead of `:Poly`.

mlubin: Yeah, I don't see a need to support symbols outside of the keyword arguments. Either ``@variable(m, x, Int)`` or ``@variable(m, x, category=:Int)``.

src/macros.jl
```diff
@@ -856,8 +842,7 @@ variable_error(args, str) = error("In @variable($(join(args,","))): ", str)
 # * The `SDP` and `Symmetric` positional arguments in `extra` will not be passed to `constructvariable!`. Instead,
 #    * the `Symmetric` argument will check that the container is symmetric and only allocate one variable for each pair of non-diagonal entries.
 #    * the `SDP` argument will do the same as `Symmetric` but in addition it will specify that the variables created belongs to the SDP cone in the `varCones` field of the model.
-#   Moreover, if a category is passed in `extra` not as a symbol (e.g. `Bin` instead of `:Bin`), it will be transformed to a symbol before being
-#   passed to `constructvariable!`.
+#   Moreover, if a category is passed in `extra`, it will be passed using the `category` keyword to `constructvariable!`.
```
mlubin: Could you list the "special" categories that we recognize?

src/macros.jl
```diff
@@ -1041,10 +1035,11 @@ macro variable(args...)
     code = :( $(refcall) = constructvariable!($m, $(extra...); _error=$_error, lowerbound=$lb, upperbound=$ub, category=$t, basename=EMPTYSTRING, start=$value, $(extra_kwargs...)) )
 
     # Determine the return type of constructvariable!. This is needed to create the container holding them
-    # we could try to do Core.Inference.return_type(constructvariable!, tuple(Model, $(typeof.(extra)...)))
-    # if isdefined(Core, :Inference)
-    # however, since there could be symbols in extra, which are only thansformed with Val after, the type could be not inferrable
-    vartype = :( variabletype($m, $(extra...)) )
+    if isdefined(Core, :Inference)
```
mlubin: Is this kosher?

blegat:
The `variabletype` function seems a bit useless now. The only thing that makes it still be useful is the fact that it seems that `isdefined(Core, :Inference)` may be false (see [here](https://github.com/JuliaLang/julia/blob/a0ada1dc46414acefac4d64546bac7ca7434043c/base/promotion.jl#L309-L313)). However, I don't understand it which case this might happen

mlubin:
I'd rather have ``variabletype`` and avoid using ``Core.Inference.return_type`` if we can.

mlubin:
Looks pretty good to me, thanks for addressing all the comments. Could you check for performance regressions by running ``test/perf/speed2.jl``?

blegat:
I had a big 0.5x performance penalty because I was passing `lowerbound`, ... as keyword arguments to `constructvariable!`. It is now passed as positional arguments in 2e3f32c579ae4b07fe57747c0f973eed6c2825e0.
I now get a performance speedup for CONT5 that I don't understand thanks to the `; extra_kwargs...` at the end of `constructvariable!`. See the details below
```
➜  perf git:(master) julia5 speed2.jl
PMEDIAN BUILD MIN=0.175563795  MED=0.320741196
PMEDIAN INTRN MIN=0.229743075  MED=0.428653894
CONT5 BUILD   MIN=0.267186116  MED=0.416271085 <- ~25, 40
CONT5 INTRN   MIN=0.116656921  MED=0.241355246
➜  perf git:(master) git co constructvariable\! 
➜  perf git:(constructvariable!) julia5 speed2.jl
PMEDIAN BUILD MIN=0.215696743  MED=0.326776765
PMEDIAN INTRN MIN=0.210215336  MED=0.388464331
CONT5 BUILD   MIN=0.160209637  MED=0.314444853 <- ~15, 30
CONT5 INTRN   MIN=0.118863281  MED=0.254929506
➜  perf git:(constructvariable!) julia5 speed2.jl
PMEDIAN BUILD MIN=0.191695599  MED=0.355724032
PMEDIAN INTRN MIN=0.225102732  MED=0.402638121
CONT5 BUILD   MIN=0.15938412   MED=0.294249865 <- ~15, 30
CONT5 INTRN   MIN=0.116651164  MED=0.240908563
➜  perf git:(constructvariable!) julia5 speed2.jl
PMEDIAN BUILD MIN=0.19490058  MED=0.337533164
PMEDIAN INTRN MIN=0.222318405  MED=0.399466155
CONT5 BUILD   MIN=0.16060928   MED=0.29709146  <- ~15, 30
CONT5 INTRN   MIN=0.117465759  MED=0.238921681
➜  perf git:(constructvariable!) git co master
➜  perf git:(master) julia5 speed2.jl
PMEDIAN BUILD MIN=0.170941645  MED=0.322015502
PMEDIAN INTRN MIN=0.214928813  MED=0.428653469
CONT5 BUILD   MIN=0.266489441  MED=0.392994993 <- ~25, 40
CONT5 INTRN   MIN=0.116713644  MED=0.236784862
➜  perf git:(master) julia5 speed2.jl
PMEDIAN BUILD MIN=0.171263875  MED=0.327381283
PMEDIAN INTRN MIN=0.217781721  MED=0.433130278
CONT5 BUILD   MIN=0.264515939  MED=0.428355359 <- ~25, 40
CONT5 INTRN   MIN=0.117319895  MED=0.238396184
➜  perf git:(master) julia5 speed2.jl
PMEDIAN BUILD MIN=0.166273391  MED=0.316517186
PMEDIAN INTRN MIN=0.221250783  MED=0.427581392
CONT5 BUILD   MIN=0.26114252  MED=0.388637578  <- ~25, 40
CONT5 INTRN   MIN=0.117485551  MED=0.237800865
➜  perf git:(master!) git co constructvariable\!
(remove the for loops for the errors of extra_kwargs)
➜  perf git:(constructvariable!) julia5 speed2.jl          
PMEDIAN BUILD MIN=0.214137355  MED=0.318337662
PMEDIAN INTRN MIN=0.211794955  MED=0.393985586
CONT5 BUILD   MIN=0.15971617   MED=0.312552981 <- ~15, 30
CONT5 INTRN   MIN=0.118461951  MED=0.257029013
(remove the "; extra_kwargs...")
➜  perf git:(constructvariable!) julia5 speed2.jl
PMEDIAN BUILD MIN=0.179196647  MED=0.341086195
PMEDIAN INTRN MIN=0.219295261  MED=0.449476719
CONT5 BUILD   MIN=0.272981907  MED=0.411483141 <- ~25, 40
CONT5 INTRN   MIN=0.119036939  MED=0.248688892
(add everything back)
➜  perf git:(constructvariable!) julia5 speed2.jl
PMEDIAN BUILD MIN=0.21665448  MED=0.32489813
PMEDIAN INTRN MIN=0.214153409  MED=0.393124851
CONT5 BUILD   MIN=0.160253171  MED=0.317724199 <- ~15, 30
CONT5 INTRN   MIN=0.11881246  MED=0.255349231
```

mlubin:
Does the new change to positional arguments mean that anyone who implements ``constructvariable!`` has to take arguments ``objective::Number``,  ``inconstraints::Vector`` and , ``coefficients::Vector{Float64}``? That seems pretty messy since these arguments are very specific to LinearQuadratic models.

blegat:
No you can implement the second version without these 3 arguments.

mlubin:
Is that explained somewhere in the comments?

blegat:
Now it is :)

mlubin:
Seems good now, but tests are failing.

blegat:
I use the `isdependent` function to determine if `sym` depends on the `idxvars`. If this is not the case, I fall back to the old behavior. Otherwise I do the nested loop to execute `sym` with values for all of the `idxvars`. Once it has been executed back I break all the nested loops. By default the type is `Any` so
```julia
@variable x[i=1:0] == i
```
will be a container of `Any`.
However
```julia
@variable x[i=1:0] == 0
```
will be a container of `Variable` since it will detect that `sym` does not depends on `i`.

mlubin:
Why would the type of the variable depend on the value of the bounds?

blegat:
For example for polynomials, for `p(x)`, we create one Variable per coefficient of and it creates a `Polynomial{JuMP.Variable}` but for `p(x) >= 0`, we create one SDP matrix of variables so it returns a `MatPolynomial{JuMP.Variable}`.

mlubin:
@blegat, that syntax isn't very consistent with how we use ``@variable`` now. Could you use a tag like ``SDP`` for this?

blegat:
Sorry I meant `p` and `p >= 0` not `p(x)` and `p(x) >= 0`

mlubin:
Still ``p >= a`` returning a different type when ``a = 1`` and ``a = 0`` is not consistent with what we do. The types shouldn't depend on data.

blegat:
I could do
```julia
@variable(m, p, Poly(X))
@variable(m, p, SOSPoly(x)) equivalent to @variable(m, p >= 0, SOSPoly(x)) -> instead of @variable(m, p >= 0, Poly(X))
```
What do you prefer then:
* Remove `getloopedtype` and do `variabletype(m, extra...)` instead of `variabletype(m, extra..., lowerbound, upperbound, basename, category, start)`
* Allow `variable type to depend on the type of `lowerbound, upperbound, basename, category, start` but not on its value.

mlubin:
The first option is much simpler and preferable unless there's a good reason for the second option.

blegat:
I have implemented the first option in 088c84b :)

mlubin:
To make sure I'm understanding things, you will implement implement ``variabletype(::Model,::Poly)`` and ``variabletype(::Model,::SOSPoly)`` right?

blegat:
Yes exactly :) (although technically it will be `Poly{false}` and `Poly{true}` and `SOSPoly` will be an alias created by SumOfSquares)

mlubin:
Ok, makes sense. It would be nice if we removed ``SDP`` and ``Symmetric`` as special cases, but that can be future work.

mlubin:
Final request before merging. Could you add a test that implements a trivial extension and calls, e.g., ``@variable(m, x, MyType)``?

mlubin:
I'll merge this tomorrow if there are no further objections. @joehuchette

joehuchette:
Can we hold off merging for a day or two, until I can review in detail?

mlubin:
@joehuchette, sure

src/macros.jl
```diff
@@ -811,11 +813,60 @@ esc_nonconstant(x::Number) = x
...
+    end
+    Variable(m, lowerbound, upperbound, category == :Default ? :Cont : category, basename, start)
+end
+
 const EMPTYSTRING = ""
 
 variable_error(args, str) = error("In @variable($(join(args,","))): ", str)
 
+# @variable(m, expr, extra...; kwargs...)
+# where `extra` is a list of extra positional arguments and `extra_kwargs` is a list of keyword arguments.
```
joehuchette: ``extra_kwargs`` --> ``kwargs``

PR #1038, opened 2017-05-31
Add SDP cone and use it in SDconstraint

blegat:
Related to https://github.com/JuliaOpt/PolyJuMP.jl/issues/10

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -29,6 +29,8 @@ export
     LinearConstraint, QuadConstraint, SDConstraint, SOCConstraint,
     NonlinearConstraint,
     ConstraintRef,
+# Cones
+    PSDCone,
 # Functions
     # Model related
     getobjectivebound, getobjectivevalue, getobjective,
@@ -530,6 +532,9 @@ include("norms.jl")
 ##########################################################################
 # SOSConstraint  (special ordered set constraints)
 include("sos.jl")
+
+# Used in @constraint m X in SDP
+immutable PSDCone end
 
 ##########################################################################
 # SDConstraint is a (dual) semidefinite constraint of the form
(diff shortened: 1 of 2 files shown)


joehuchette:
Cool, thanks for this. I think it's potentially confusing to allow both ``x in SDP`` and ``x in SDP()``; we should choose one. We could also have something like
```jl
immutable PSDCone end
const SDP = PSDCone()
```

blegat:
I like your idea ! We should say that the cone should be an instantiated type and with your trick we get the convenience of not having to write the parentheses

mlubin:
It would be nice to have a discussion first on how we're going to change MathProgBase format to handle these cone objects so that we can avoid the situation where the JuMP uses different objects than MathProgBase and needs to do a translation.

blegat:
One worry I have with MathProgBase is that if `SOCone` and `PSDCone` are different types then the eltype of `[(1:3, SOCone()), (4:7, PSDCone())]` will not be concrete.

joehuchette:
I'm not convinced that how we represent cones at the modeling layers needs to match with what's going on at the MPB-level. We should discuss this further at the meet-up, though.

@blegat I don't think those types of constructions are used anywhere internally where incomplete type information would lead to a performance bottleneck, but I could be wrong.

mlubin:
The only potential performance issue I see is if you have an LP with a million rows and decide to represent each one with a separate cone. Doesn't seem like a big deal to me.

If you don't mind breaking things in the future when we figure out a plan with MPB, then ok. The only thing I won't allow is something called the SDP cone because that's definitely going to change to PSD cone.

blegat:
I don't mind if this breaks in the future, I just need to be able to use `@SDconstraint` for SOS-matrices otherwise I cannot deprecate `@polyconstraint` :)

mlubin:
I'd be ok merging this if you just remove ``const SDP = PSDCone()`` and use ``PSDCone()`` for the time being.

blegat:
Thanks ! I have remove the `SDP` constant. No objection left ? :)

PR #1039, opened 2017-05-31
Add constructvariablecontainer! function

blegat:
Depends on #1038 
This could help creating different containers like `AxisArray`s #980

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -833,6 +833,66 @@ function constructvariable!(m::Model, _error::Function, lowerbound::Number, uppe
     end
     Variable(m, lowerbound, upperbound, category == :Default ? :Cont : category, basename, start)
 end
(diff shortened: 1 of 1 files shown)


chriscoey:
we are doing some housekeeping, so bumping this. anything we need to discuss?

blegat:
@chriscoey Yes, I have just rebased it now that #1038 has been merged and I'd like to know what you think of it :)

mlubin:
There's no documentation, what does this change?

blegat:
Yes documentation should be added. It basically calls `constructvariablecontainer` with the extra arguments. This allows to do the special treatment of `SDP` and `Symmetric` not so special. The issue is that this not done in evaluation of the code so we still have `:SDP`  and `:Symmetric` as symbols. Therefore we should evaluate the part of extra that is used for building the container but not evalueate the part that is used to build each part yet... This still needs some thought but comments are welcome :)

mlubin:
Looks like this is stale, can we close?

blegat:
Yes, we can try to make the change later

PR #1049, opened 2017-06-10
drop support for Julia 0.5

mlubin:


joehuchette:
Maybe we should turn nightly on Travis off...

mlubin:
@jrevels could help us update ForwardDiff now

PR #1058, opened 2017-06-15
Initial Migration to Docstrings

ccoffrin:
My best effort to migrate the documentation to use Docstrings.  I focused on just moving the existing text around, and did not consider any organization or polishing yet.  

Things I could use help with,
- finding the remaining functions mentioned in refmodel.md, refvariable.md
- it is not clear how to best document the macro pluralization feature in refexpr.md 
- If macros don't have types, does a type signature in their documentation make sense?
- it looks like latex is not rendering inside doc strings (e.g.  refexpr.md, `linearterms(aff::GenericAffExpr)`) may need to apply the recommended fix

Some improvements to consider during this sprint,
- Incorporating doc strings highlights the mix of manual and library docs, I am in favor of breaking these into two sections at the top level
- Is there a resonable way to incorporate the epic nodes in JuMP's code (For example, "Let S₊ be the cone of symmetric semidefinite matrices in" note in JuMP.jl) into docstrings that get into the developer docs?

### DIFF
--- a/docs/src/refexpr.md
+++ b/docs/src/refexpr.md
@@ -27,28 +27,21 @@ Methods
 ```@docs
 addconstraint
 @constraint
+@constraints
 @expression
 @SDconstraint
 addSOS1
 addSOS2
 @LinearConstraint
+@LinearConstraints
 @QuadConstraint
+@QuadConstraints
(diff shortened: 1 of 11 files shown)


ccoffrin:
@mlubin I believe all notable issues are now resolved.

PR #1081, opened 2017-07-23
initial transition to MOI

mlubin:
I moved the old tests to `test/old`, and we'll have to manually update them one by one. The best way to get a sense of the changes is to look at the tests that are passing now ([1](https://github.com/JuliaOpt/JuMP.jl/blob/jump/moi/test/variable.jl) [2](https://github.com/JuliaOpt/JuMP.jl/blob/jump/moi/test/objective.jl) [3](https://github.com/JuliaOpt/JuMP.jl/blob/jump/moi/test/constraint.jl)).

Big changes so far:
- All constraint data is stored in the `Instance`. The JuMP model is just responsible for keeping references.
- Variable bounds (and fixed) are considered constraints that may or may not be present, so `haslowerbound` is a valid question to ask.
- Got rid of variable categories. We have two special kinds of constraints for integer and binary restrictions. ``setinteger`` and ``setbinary`` turn these constraints on, ``unsetinteger`` and ``unsetbinary`` turn these constraints off. Removed `category` kwarg to `@variable` and replaced with `binary=true/false` and `integer=true/false`. `SemiInt` etc maybe don't need special syntax, could be imposed as variablewise constraints.
- The concept of a variable value is gone. We have separate objects for the start and the result vector.
- Old deprecation warnings are now errors (curlies, multiple variables with the same name, etc.)
- Container revamp (#1102)

I'm not putting any effort into deprecations at the moment. That should probably be a separate effort after we have things up and running. All of the new methods could use a round of bikeshedding.

At the moment we can fully represent an (MI)LP instance, so I'd like to try setting up the solver infrastructure next and see if we can solve a 2-dimensional LP using CSDP.


src/macros.jl
```diff
@@ -279,26 +278,40 @@ end
...
+function constructconstraint!(aff::AffExpr, set::MOI.GreaterThan)
+    offset = aff.constant
+    aff.constant = 0.0
+    return LinearConstraint(aff, MOI.GreaterThan(set.lower-offset))
+end
+function constructconstraint!(aff::AffExpr, set::MOI.EqualTo)
+    offset = aff.constant
+    aff.constant = 0.0
+    return LinearConstraint(aff, MOI.EqualTo(set.value-offset))
+end
```
blegat: You can replace the three functions with
```julia
function constructconstraint!{S<:Union{MOI.EqualTo, MOI.GreaterThan, MOI.LessThan}(aff::AffExpr, set::S)
    offset = aff.constant
    aff.constant = 0.0
    return LinearConstraint(aff, S(MOIU.getconstant(set)-offset))
end
```

src/macros.jl
```diff
@@ -1075,51 +1096,37 @@ macro variable(args...)
...
-        !isa(var,Symbol) &&
-        _error("Can only create one variable at a time when adding to existing constraints.")
-
-        variablecall = :( constructvariable!($m, $(extra...), $_error, $lb, $ub, $t, $obj, $inconstraints, $coefficients, string($quotvarname), $value) )
-        addkwargs!(variablecall, extra_kwargs)
-        return assert_validmodel(m, quote
-            $variable = $variablecall
-            $(anonvar ? variable : :($escvarname = $variable))
-        end)
-    end
```
blegat: Do we drop column generation in JuMP in addition to dropping it in MOI ?

src/macros.jl
mlubin: Nice

src/macros.jl
mlubin: We can have a specific function to help with column generation, but given that the information is no longer needed at the time of variable construction, I don't see a reason to keep this awkward syntax in the `@variable` macro.

src/JuMP.jl
```diff
@@ -719,22 +542,31 @@ end
 ##########################################################################
 # ConstraintRef
 # Reference to a constraint for retrieving solution info
-immutable ConstraintRef{M<:AbstractModel,T<:AbstractConstraint}
+# struct ConstraintRef{M<:AbstractModel,C <: MOICON}
+struct ConstraintRef{M<:AbstractModel,C} # TODO temporary
```
blegat: Why is it temporary ?
I liked the `ConstraintRef{M<:AbstractModel, T<:AbstractConstraint}` design. This allows a constraint to be mapped to several MOICON. For example a SOS constraint is mapped to a VectorOfVariables/PSD constraint and a VectorAffineFunction/Zeros constraint.
With `ConstraintRef{M<:AbstractModel, T<:MOICON}`, we will be forced to define the Sum of squares cone as an `MOI.AbstractSet` and ask for every solver to support a bridge for it. The bridges cannot be implemented outside the solver (e.g. by defininte `addconstraint(m, f, s::SumOfSquaresCone)` since the solver needs to be aware the the initial constraint to implement attributes such as `ListOfConstraints`, `NumberOfConstraints`, `ConstraintFunction`, ...

mlubin: Ok, let's drop this restriction.

src/JuMP.jl
```diff
@@ -56,51 +61,82 @@ export
...
+mutable struct Model <: AbstractModel
+
+    instance::JuMPInstance{Float64}
+    # special variablewise properties that we keep track of:
+    # lower bound, upper bound, fixed, integrality, binary
+    variabletolowerbound::Dict{MOIVAR,LBREF}
+    variabletoupperbound::Dict{MOIVAR,UBREF}
+    variabletofix::Dict{MOIVAR,FIXREF}
+    variabletointegrality::Dict{MOIVAR,INTREF}
+    variabletozeroone::Dict{MOIVAR,BINREF}
```
blegat: Why don't we define `getindex(m::VariableToValueMap, ::MOIVAR)` and use `VariableToValueMap` here ?

mlubin: Why have an extra level of wrapping?

src/macros.jl
blegat: Ok, it would make the `@variable` macro clearer indeed.

src/JuMP.jl
blegat: I don't really understand why some are `VariableToValueMap` and some are `Dict{MOIVAR, ...}`. Internally it is just the same thing, and when you use e.g. `variabletolowerbound`, you do `v.m.variabletolowerbound[instancereference(v)]` while if it was a `VariableToValueMap`, you could just do `v.m.variabletolowerbound[v]`.

mlubin: `VariableToValueMap` is meant to be user facing and corresponds to MOI attributes. We will return them to users e.g., for multiple solutions, and users can use them to provide attributes like a starting point or starting basis.
`variabletoupperbound` is an internal data structure that will never be visible to users. I guess it would be okay to make it a `VariableToValueMap` but to me it confuses two different concepts.

blegat: Oh ok, makes sense now

src/JuMP.jl
```diff
@@ -542,8 +542,7 @@ end
 ##########################################################################
 # ConstraintRef
 # Reference to a constraint for retrieving solution info
-# struct ConstraintRef{M<:AbstractModel,C <: MOICON}
-struct ConstraintRef{M<:AbstractModel,C} # TODO temporary
+struct ConstraintRef{M<:AbstractModel,C}
```
blegat: Thanks !

mlubin:
I removed all automatic broadcasting for the new ``getvalue`` equivalents. Let's see how far we can get by using the broadcast syntax. JuMPContainers are not yet compatible with the broadcast syntax though.

PR #1085, opened 2017-07-27
fix for callback macros on 0.6

mlubin:
callback.jl was failing tests


mlubin:
I'm going to leave this on the release-0.18 branch only to avoid unnecessary conflicts between master and jump/moi.

PR #1089, opened 2017-08-01
Add support for SOC constraints and SDP variable constraints

blegat:


src/macros.jl
```diff
@@ -252,18 +252,12 @@ end
 #     LinearConstraint(aff, lb-offset, ub-offset)
 # end
 
-constructconstraint!(quad::QuadExpr, sense::Symbol) = QuadConstraint(quad, sense)
+constructconstraint!(quad::QuadExpr, set::Symbol) = QuadConstraint(quad, sense)
```
joehuchette: ``sense`` is still on the RHS

src/norms.jl
```diff
@@ -100,34 +100,58 @@ GenericSOCExpr{C,V} = GenericNormExpr{2,C,V}
...
-#     addconstraint(m::Model, c::SOCConstraint)
-#
-# Add a SOC constraint to `Model m`.
-# """
-# function addconstraint(m::Model, c::SOCConstraint)
-#     push!(m.socconstr,c)
-#     m.internalModelLoaded = false
-#     ConstraintRef{Model,SOCConstraint}(m,length(m.socconstr))
-# end
+getvalue(n::GenericNormExpr) = n.coeff * getvalue(n.norm) + getvalue(n.aff)
```
mlubin: ``getvalue`` doesn't exist anymore

src/norms.jl
```diff
@@ -100,34 +100,58 @@ GenericSOCExpr{C,V} = GenericNormExpr{2,C,V}
...
+
+##########################################################################
+# GenericSOCConstraint
+# Second-order cone constraint of form
+# α||Ax-b||₂ + cᵀx + d ≤ 0
+type GenericSOCConstraint{T<:GenericSOCExpr} <: AbstractConstraint
```
mlubin: update to ``mutable struct`` or ``struct``

src/norms.jl
```diff
@@ -100,34 +100,58 @@ GenericSOCExpr{C,V} = GenericNormExpr{2,C,V}
...
-#     push!(m.socconstr,c)
-#     m.internalModelLoaded = false
-#     ConstraintRef{Model,SOCConstraint}(m,length(m.socconstr))
-# end
+getvalue(n::GenericNormExpr) = n.coeff * getvalue(n.norm) + getvalue(n.aff)
+
+##########################################################################
+# GenericSOCConstraint
+# Second-order cone constraint of form
+# α||Ax-b||₂ + cᵀx + d ≤ 0
```
mlubin: I would question if we should use this data structure anymore. To be more consistent with MOI, a SOC constraint should just be a vector of affine expressions in the same order as the cone. This new structure could then be generalized to represent any VectorAffineFunction-in-set constraint.

src/sd.jl
```diff
@@ -0,0 +1,26 @@
+export PSDCone
+# Used in @constraint m X in SDP
+struct PSDCone end
+
+struct SDVariableConstraint <: AbstractConstraint
+    Q::Matrix{JuMP.Variable}
+end
+
+# Used by the @variable macro but can also be used through the @constraint macro
+function constructconstraint!(Q::Matrix{JuMP.Variable}, PSDCone)
```
mlubin: ``::PSDCone``?

src/sd.jl
```diff
@@ -0,0 +1,26 @@
...
+    @assert issymmetric(Q) # TODO it could be nonsymmetric if used through the @constraint macro
```
mlubin: We're not currently constructing variablewise constraints via the `@constraint` macro. This needs a larger discussion if we're going to allow this syntax.

src/sd.jl
```diff
@@ -0,0 +1,26 @@
+export PSDCone
+# Used in @constraint m X in SDP
```
mlubin: `in SDP` doesn't exist

src/variables.jl
```diff
@@ -351,24 +351,24 @@ resultvalue(v::Variable) = variableresult(v.m)[v]
...
-#     for I in eachindex(arr)
-#         ret[I] = resultvalue(arr[I])
-#     end
-#     # Copy printing data from @variable for Array{Variable} to corresponding Array{Float64} of values
-#     if registered
-#         m.varData[ret] = m.varData[arr]
-#     end
-#     ret
-# end
+function resultvalue(arr::Array{Variable})
```
mlubin: Why uncomment this? You can use `resultvalue.(arr)`.

mlubin:
Good start. We need some more discussion on syntax and data structures before this can go through, see some of the comments above. I also won't allow ``@variable(m, ..., SDP)``. We can rename it ``@variable(m, ..., PSD)`` for now, but we'll eventually want to discuss more.

src/macros.jl
blegat: Oops, I didn't want to change this method, I am not using it so it is for a further PR

src/JuMP.jl
```diff
@@ -565,6 +565,22 @@ end
 # GenericRangeConstraint, LinearConstraint
 include("affexpr.jl")
 
+struct VectorAffineConstraint{S <: MOI.AbstractVectorSet} <: AbstractConstraint
+    func::MOI.VectorAffineFunction{Float64}
```
mlubin: Having MOI references here doesn't make too much sense to me. This is a JuMP-level constraint which should have JuMP variables in it. ``Vector{AffExpr}`` could work. It should be consistent with [`LinearConstraint`](https://github.com/JuliaOpt/JuMP.jl/blob/eea38231faaf8ace1e1914d8235759fedc6f336b/src/affexpr.jl#L186). (Maybe `LinearConstraint` is a confusing name since we allow arbitrary sets.)

blegat:
Any more objection or good to merge ?

mlubin:
Will do another full review later today

src/affexpr.jl
```diff
@@ -134,6 +134,35 @@ function AffExpr(m::Model, f::MOI.ScalarAffineFunction)
     return AffExpr(Variable.(m,f.variables), f.coefficients, f.constant)
 end
 
+"""
+    _fillvaf!(outputindex, variables, coefficients, offset::Int, oi::Int, coeff, aff::AffExpr)
```
joehuchette: Extra ``coeff`` in argument list

src/macros.jl
```diff
@@ -937,7 +931,7 @@ macro variable(args...)
         anon_singleton = true
     else
         x = shift!(extra)
-        if x in [:Int,:Bin,:SDP]
+        if x in [:Int,:Bin,:PSD]
```
joehuchette: @mlubin do we want deprecations for ``SDP`` here and throughout?

mlubin: I've been ignoring deprecations so far, there are too many changes without 1-1 replacements and we still have unsettled syntax questions. I think we'll have to do a deprecation pass later on. We can write the upgrading guide at the same time.

mlubin:
I approve modulo Joey's comment on the documentation typo.

PR #1093, opened 2017-08-23
remove norm expression type and norm() syntax, test SOC and RotatedSOC

mlubin:
This needs a bit of discussion, but I propose to remove the Norm expression type which was pretty ad-hoc and as far as I'm aware only really used in its general form by JuMPeR. Also remove the syntax
```julia
@constraint(m, norm(x) <= t)
```
in favor of
```julia
@constraint(m, [t, x...] in MOI.SecondOrderCone(n))
```

We can have a discussion of special syntax for common conic constraints, but my feeling is that the best place for ``norm()`` syntax is at the level of DCP. As is, the syntax is quite deceptive because one would think you could write ``norm(x) + norm(y) <= t`` which you can't in JuMP but could in DCP. The 300-line reduction isn't bad either.

@blegat, see the commented RotatedSOC test. Looks like there's a bug in SOI.


chriscoey:
Yes. If JuMP is not the place for DCP, then it's not the place for norm.

blegat:
The `norm(x) + norm(y) <= t` example is rather convincing. It is going to break people's code very badly if we release a first version without DCP though :-P

@mlubin Currently, changing parameters in the constructor do not work with CSDP (see [here](https://github.com/JuliaOpt/CSDP.jl/issues/15)). You need to put [this file](https://github.com/JuliaOpt/CSDP.jl/blob/master/test/param.csdp) in the `JuMP/test` folder.

mlubin:
> It is going to break people's code very badly

It possible to write a deprecation handler for this like we did with the curly syntax.

mlubin:
A deprecation is actually a bit tricky because it could depend on the runtime values of constants, e.g., we wouldn't want to transform ``c*norm(x) <= t`` unless `c >= 0`. I'll add a stub error message that can be improved later.

PR #1096, opened 2017-08-30
simplify printing

mlubin:
Fix printing of AffExpr
remove special cache for JuMPContainer{Variable}
Disable special printing for JuMPContainer until we decide what to do with it.

We can put back in a mechanism for avoiding generating strings names for all variables if dictated by benchmarking, but I think it's easier to remove the old code for now.


src/macros.jl
```diff
@@ -1014,7 +1015,7 @@ macro variable(args...)
             hasstart = true
             value = esc(ex.args[2])
         elseif kwarg == :basename
-            quotvarname = esc(ex.args[2])
+            basename = esc(ex.args[2])
```
blegat: This sets basename to a `Symbol` but above when we do `basename = string(getname(var))`, it is a string and `constructvariable!` creates a string. Why is this working ? :-P

mlubin: `esc` applied to a constant has no effect, but `ex.args[2]` is not always a constant. If there are objects spliced into the string like `"var-$i"` then we get an expression object that needs to be escaped.

PR #1103, opened 2017-09-18
Unexported solvers

joaquimg:
Add support to unexported `XpressSolver` now `Xpress.XpressSolver`

Is it what you meant [here](https://github.com/JuliaOpt/MathOptInterface.jl/issues/70#issuecomment-330079082) @mlubin ?

Should I do something elsewhere?

PR #1104, opened 2017-09-19
update xpress docs

joaquimg:
goes with #1103


mlubin:
Change looks good but the version number is incorrect, should be 0.18 docs which we link to from the README.

joaquimg:
Ok, only changing in 0.18 is ok?

mlubin:
Yes, we don't update documentation for anything but the latest release (and the development version).

joaquimg:
closed in favor of #1105

PR #1108, opened 2017-09-23
JuMP/MOI: Basic support for quadratic expressions and constraints

mlubin:
~Quadratic objectives don't yet work because of https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/issues/12~

- Renamed `LinearConstraint` to `AffExprConstraint`
- Similarly, `QuadExprConstraint`

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -87,9 +87,17 @@ if VERSION < v"0.6.0-dev.1632" # julia PR #17623
     @eval (.^)(lhs::Union{Variable,AffExpr}, rhs::Number) = lhs^rhs
 end
 # AffExpr--Variable
-(+){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::V) = (+)(rhs,lhs)
+(+){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::V) = GenericAffExpr{C,V}(vcat(lhs.vars,rhs),vcat(lhs.coeffs,one(C)), lhs.constant)
 (-){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::V) = GenericAffExpr{C,V}(vcat(lhs.vars,rhs),vcat(lhs.coeffs,-one(C)),lhs.constant)
-(*)(lhs::AffExpr, rhs::Variable) = (*)(rhs,lhs)
+# Don't fall back on Variable*AffExpr to preserve lhs/rhs consistency (appears in printing)
+function (*)(lhs::AffExpr, rhs::Variable)
+    n = length(lhs.vars)
+    if lhs.constant != 0.
+        ret = QuadExpr(copy(lhs.vars),[rhs for i=1:n],copy(lhs.coeffs),AffExpr([rhs], [lhs.constant], 0.))
+    else
+        ret = QuadExpr(copy(lhs.vars),[rhs for i=1:n],copy(lhs.coeffs),zero(AffExpr))
+    end
+end
 (/)(lhs::AffExpr, rhs::Variable) = error("Cannot divide affine expression by a variable")
 # AffExpr--AffExpr
 (+){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::GenericAffExpr{C,V}) = (operator_warn(lhs,rhs); GenericAffExpr(vcat(lhs.vars,rhs.vars),vcat(lhs.coeffs, rhs.coeffs),lhs.constant+rhs.constant))
--- a/src/quadexpr.jl
+++ b/src/quadexpr.jl
@@ -11,8 +11,7 @@
 # Defines all types relating to expressions with a quadratic and affine part
 # - GenericQuadExpr             ∑qᵢⱼ xᵢⱼ  +  ∑ aᵢ xᵢ  +  c
 #   - QuadExpr                  Alias for (Float64, Variable)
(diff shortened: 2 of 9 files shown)


mlubin:
This is updated and ready to go.

src/operators.jl
```diff
@@ -87,9 +87,17 @@ if VERSION < v"0.6.0-dev.1632" # julia PR #17623
...
 end
 # AffExpr--Variable
-(+){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::V) = (+)(rhs,lhs)
+(+){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::V) = GenericAffExpr{C,V}(vcat(lhs.vars,rhs),vcat(lhs.coeffs,one(C)), lhs.constant)
 (-){C,V<:JuMPTypes}(lhs::GenericAffExpr{C,V}, rhs::V) = GenericAffExpr{C,V}(vcat(lhs.vars,rhs),vcat(lhs.coeffs,-one(C)),lhs.constant)
-(*)(lhs::AffExpr, rhs::Variable) = (*)(rhs,lhs)
+# Don't fall back on AffExpr*Variable to preserve lhs/rhs consistency (appears in printing)
+function (*)(lhs::AffExpr, rhs::Variable)
+    n = length(lhs.vars)
+    if lhs.constant != 0.
```
blegat: Could we do simply `lhs.constant * rhs` here and move the `if` to `(*)(lhs::Number, rhs::Variable)` ?

src/quadexpr.jl
```diff
@@ -61,25 +60,59 @@ function Base.isequal{T,S}(q::GenericQuadExpr{T,S},other::GenericQuadExpr{T,S})
...
+    return MOI.ScalarQuadraticFunction(instancereference.(q.aff.vars), q.aff.coeffs, instancereference.(q.qvars1), instancereference.(q.qvars2), q.qcoeffs, q.aff.constant)
+end
+
+function QuadExpr(m::Model, f::MOI.ScalarQuadraticFunction)
+    return QuadExpr(Variable.(m,f.quadratic_rowvariables), Variable.(m, f.quadratic_colvariables), f.quadratic_coefficients, AffExpr(Variable.(m,f.affine_variables), f.affine_coefficients, f.constant))
+end
+
+function setobjective(m::Model, sense::Symbol, a::QuadExpr)
+    @assert !m.solverinstanceattached # TODO
+    if sense == :Min
```
blegat: Why don't we do
```julia
function setobjective(m::Model, sense::Symbol, a::AbstractJuMPScalar)
    ...
    MOI.set!(m.instance, MOI.ObjectiveFunction(), JuMPScalar2MOIfunction(a))
    nothing
end
```
Of course the name `JuMPScalar2MOIfunction` should be replaced by something better.

src/quadexpr.jl
```diff
@@ -61,25 +60,59 @@ function Base.isequal{T,S}(q::GenericQuadExpr{T,S},other::GenericQuadExpr{T,S})
...
-#             MathProgBase.setquadobjterms!(m.internalModel, Cint[v.col for v in m.obj.qvars1], Cint[v.col for v in m.obj.qvars2], m.obj.qcoeffs)
-#         else
-#             isa(m.internalModel, MathProgBase.AbstractLinearQuadraticModel) && Base.warn_once("Solver does not support adding a quadratic objective to an existing model. JuMP's internal model will be discarded.")
-#             m.internalModelLoaded = false
-#         end
-#     end
-#     setobjectivesense(m, sense)
-# end
+function MOI.ScalarQuadraticFunction(q::QuadExpr)
+    return MOI.ScalarQuadraticFunction(instancereference.(q.aff.vars), q.aff.coeffs, instancereference.(q.qvars1), instancereference.(q.qvars2), q.qcoeffs, q.aff.constant)
```
joehuchette: What is ``instancereference`` here?

joehuchette: ohhhh is this a vectorized constructor? It looked like a crazy field access...

src/quadexpr.jl
```diff
@@ -105,67 +138,27 @@ end
...
-const QuadConstraint = GenericQuadConstraint{QuadExpr}
+"""
+    addconstraint(m::Model, c::QuadExprConstraint)
 
-function Base.copy(c::QuadConstraint, new_model::Model)
-    return QuadConstraint(copy(c.terms, new_model), c.sense)
+Add a `QuadExpr` constraint to `Model m`.
+"""
+function addconstraint(m::Model, c::QuadExprConstraint)
+    @assert !m.solverinstanceattached # TODO
```
joehuchette: This will be addressed when problem modification support is added?

src/quadexpr.jl
mlubin: Yes, welcome to the future

src/quadexpr.jl
mlubin: Definitely

src/operators.jl
mlubin: How? What is `(*)(lhs::Number, rhs::Variable)` supposed to return if `lhs = 0`? An `AffExpr` with zero terms?

mlubin: Ok, `zero(AffExpr)` makes sense and is arguably simpler. However, I don't want to make arbitrary tweaks in performance critical code without profiling. This new method just copies Variable*AffExpr.

src/quadexpr.jl
mlubin: Makes sense, but I'll leave it for a future refactor.

Issue #1110, opened 2017-10-03
try ... catch statements in solve kill performance


odow:
To start, much of this will be addressed with the MOI re-write. However, until this is implemented there are some serious performance penalties.

Stochastic Dual Dynamic Programming presents a difficult challenge to solvers as it requires iteratively solving hundreds of thousands of LP's. It comes up in a few issues https://github.com/JuliaOpt/JuMP.jl/issues/969 https://github.com/JuliaOpt/MathProgBase.jl/issues/139 discussing various improvements that could be made to JuMP. 

I have a real world example (using Clp) where a _single_ line in JuMP takes up **30%** of the total program (not even just JuMP solves) execution. And it's the [`try catch` for the objective bound](https://github.com/JuliaOpt/JuMP.jl/blob/master/src/solvers.jl#L259) (which Clp doesn't have implemented so it throws a method error).

<s>Defining an appropriate method
```julia
MathProgBase.getobjbound(m) = NaN
```
solves this problem and gives a large speed-up.</s>

The fix above works for Clp, but not for solvers like Gurobi. The real problem is the solvers through an exception when the objective bound is queried of an LP.

### Action

Should MathProgBase methods have default fallbacks that return `NaN` for solvers that don't implement them?

In addition, should solvers return `NaN` rather than throw errors if the value is unavailable?

This will tidy up code that looks like
https://github.com/JuliaOpt/JuMP.jl/blob/master/src/solvers.jl#L202-L206

@joaquimg @blegat @leclere @frapac @thuener

mlubin:
This is a tricky issue since we might unexpectedly break user's code by making a small change. Given that MOI is in the pipeline and will resolve this, I'd rather avoid annoying users with additional breaking changes before that. I would recommend making the minimal change needed for your application, possibly on a branch.

blegat:
Why don't we define `getobjbound` in Clp ? Isn't this supposed to be the value of the dual objective ?

odow:
It's more a problem with the try catch design. Gurobi has the same problem

blegat:
What does Gurobi throws ?

odow:
Gurobi only returns the objective bound for MIP models (see https://www.gurobi.com/documentation/7.5/refman/objbound.html). For LP's it throws an attribute error saying that it can access the attribute.

blegat:
[...]

PR #1112, opened 2017-10-04
Update solvers.jl

odow:
A temporary fix for #1110 

Ref https://github.com/JuliaOpt/JuMP.jl/pull/1111#issuecomment-334034169


mlubin:
@joehuchette, what does BARON do here? Does it give bounds for nonconvex continuous problems?

joehuchette:
BARON tries to prove optimality on continuous nonconvex problems, so it should give a bound

mlubin:
@odow ok to close this?

PR #1117, opened 2017-10-20
writeLP now includes SOS1 and SOS2

leethargo:
addresses #867

Issue #867, opened 2016-10-11
writeLP does not include SOS constraints


leethargo:
But solvers such as Gurobi or SCIP include them in the `.lp` output.

See also [SCIP implementation](http://scip.zib.de/doc/html_devel/reader__lp_8c_source.php#l02818).

mlubin:
PRs welcome.

mlubin:
Closed by https://github.com/JuliaOpt/JuMP.jl/pull/1118 for JuMP/MPB. Will also need to be implemented for whatever we do to write LP files under JuMP/MOI. CC @odow 

Closing for now.

PR #1118, opened 2017-10-23
writeLP now includes SOS1 and SOS2

leethargo:
solves #867, follow-up from #1117

### DIFF
--- a/src/writers.jl
+++ b/src/writers.jl
@@ -280,6 +280,26 @@ function writeLP(m::Model, fname::AbstractString; genericnames=true)
             constrcount += 2
         end
     end
+    # SOS constraints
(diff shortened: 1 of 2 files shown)


leethargo:
In the CI run for #1117, there was some error about IPOPT. I guess that's unrelated, right?
So it's enough to have the Linux run on Julia 0.6 succeed?

mlubin:
That's an interesting failure, is Ipopt broken on OS X? I don't have a mac to test. @joehuchette

joehuchette:
I'm also seeing this Ipopt build failure, ref. https://github.com/JuliaOpt/Ipopt.jl/issues/61#issuecomment-339160928

PR #1119, opened 2017-11-06
Enable solvehook

blegat:
Needed it to start playing with SumOfSquares/PolyJuMP on top of JuMP/MOI.


mlubin:
Tests?

blegat:
Added :)

mlubin:
The hooks might need to change, but this is ok for now.

blegat:
Yes, this needs to be vastly improved :)

PR #1120, opened 2017-11-08
fix local lazy or cut constraints broken due to a keyword symbol issue

madanim:
This is a very small fix because keyword arguments are not filtered correctly for lazy/cut callbacks in callbacks.jl since the change of kw symbol, see Julia PR #19868. I have taken the fix e.g. in macro.jl for the function addkwargs!

(Otherwise, for now, lazy or use cuts cannot be specified as local, as the "localcut" keyword  argument is not taken into account.)

Best regards,

Mehdi


mlubin:
Please reopen this as a PR against the release-0.18 branch. `master` is closed for PRs until #1081 is merged. 

Note that the question of how to handle callbacks in JuMP 0.19 is still under discussion, and they will look very different to the current form. See the recent discussion in the [JuMP-dev](https://gitter.im/JuliaOpt/JuMP-dev) channel.

madanim:
Thanks, Miles. I'm changing now the target branch to release-0.18, and thank you also for pointing out the ongoing discussion regarding callbacks in the future

madanim:
Will submit the PR against the release-0.18 branch ...

madanim:
I have just seen that the branch release-0.18 is indeed already OK for Julia 0.6 (with dropped support for Julia 0.5) so there is finally no PR to submit. I was using a JuliaPro distribution and JuMP 0.17.1 ... Thanks for your quick reply and I'll be pleased to contribute in the future.

PR #1122, opened 2017-11-15
SDP constraint

### LINKED ISSUE #1050: syntax support for encoding symmetry
The fix to https://github.com/JuliaOpt/JuMP.jl/issues/972 was a hack that uses a numerical tolerance. It would be nice to have a syntax of some form for encoding symmetry so that it doesn't need to be detected numerically later on.
@chriscoey

blegat:
Add support for SDP constraint defined with the `@SDconstraint` macro.
What is new is that in the reference returned `cref`, `cref.instanceref` is not the reference to the PSD constraint. Indeed we also need to keep a reference to the symmetry constraints to compute the dual.
Therefore, `cref.instanceref` is an `SDSymInstanceRef`. If the matrix has type `Symmetric` then we can set `cref.instanceref` to the reference to the PSD constraint of the instance. We do not do this if the matrix is symmetric but not of type `Symmetric` since that would not be type stable since the return type would depend on the value of the elements of the matrix.
That way, if the user wants all its constraints reference to have the type `ConstraintRef{JuMP.Model, MOI.ConstraintReference}`, he can use the `Base.Symmetric` type to achieve this.

Closes #1050

### DIFF
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -63,7 +63,7 @@ end
 struct SDSymInstanceRef{T}
     sd::MOICON{MOI.VectorAffineFunction{T}, MOI.PositiveSemidefiniteConeTriangle}
     # symmetry enforcing constraints, it is null if the matrix is symmetric
-    symref::Nullable{MOICON{MOI.VectorAffineFunction{T}, MOI.Zeros}}
+    symref::Union{MOICON{MOI.VectorAffineFunction{T}, MOI.Zeros}, Void}
     # if k in symidx then invtrimap(k) has a symmetry enforcing constraint in symref
     symidx::IntSet
 end
@@ -99,10 +99,10 @@ function _constrain_symmetry(m::Model, c::SDConstraint{T}, sdref) where T
         push!(symaff, aff)
     end
     if isempty(symaff)
-        symref = Nullable{MOICON{MOI.VectorAffineFunction{Float64}, MOI.Zeros}}()
+        symref = nothing
     else
         symcon = VectorAffExprConstraint(symaff, MOI.Zeros(length(symaff)))
-        symref = Nullable{MOICON{MOI.VectorAffineFunction{Float64}, MOI.Zeros}}(addconstraint(m, symcon).instanceref)
+        symref = addconstraint(m, symcon).instanceref
     end
     SDSymInstanceRef(sdref, symref, symidx)
 end
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -71,20 +71,7 @@ end
 constructconstraint!(x::AbstractMatrix, ::PSDCone) = SDConstraint(x)
 
 _constrain_symmetry(m::Model, c::SDConstraint{T, MT}, sdref) where {T, MT<:Symmetric} = sdref
-function fill_expr!(m::Model, expr::Dict{Variable, T}, terms, coeff=one(T)) where T
-    assert_isfinite(terms)
-    coeffs = terms.coeffs
-    vars = terms.vars
-    # collect duplicates
-    for ind in eachindex(coeffs)
-        if vars[ind].m === m
-            var = vars[ind]
-            expr[var] = get(expr, var, zero(T)) + coeff * coeffs[ind]
-        else
-            throw(VariableNotOwnedError("constraints"))
-        end
-    end
-end
+
 function _constrain_symmetry(m::Model, c::SDConstraint{T}, sdref) where T
     expr = Dict{Variable, Float64}()
     symaff = AffExpr[]
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -25,3 +25,113 @@ function addconstraint(m::Model, c::SDVariableConstraint)
     cref = MOI.addconstraint!(m.instance, MOI.VectorOfVariables([instancereference(c.Q[i, j]) for j in 1:n for i in 1:j]), MOI.PositiveSemidefiniteConeTriangle(n))
     return ConstraintRef(m, cref)
 end
+
+##########################################################################
+# SDConstraint is a (dual) semidefinite constraint of the form
+# ∑ cᵢ Xᵢ ⪰ D, where D is a n×n data matrix, cᵢ are scalars,
+# and Xᵢ are n×n symmetric variable matrices. The inequality
+# is taken w.r.t. the psd partial order.
+struct SDConstraint{T<:AbstractJuMPScalar, MT<:AbstractMatrix{T}} <: AbstractConstraint
+    terms::MT
+end
+SDConstraint(terms::MT) where {T<:AbstractJuMPScalar, MT<:AbstractMatrix{T}} = SDConstraint{T, MT}(terms)
+
+"""
+    trimap(i, j)
+
+Returns the vectorized index `k` corresponding to the matrix indices `(i, j)`.
+"""
+function trimap(i::Integer, j::Integer)
+    if i < j
+        trimap(j, i)
+    else
+        div((i-1)*i, 2) + j
+    end
+end
+
+"""
+    invtrimap(i, j)
+
+Returns the matrix indices `(i, j)` (`i <= j`) corresponding to the vectorized index `k`.
+"""
+function invtrimap(k::Integer)
+    i = isqrt(2k)
+    j = k - div((i-1)*i, 2)
+    i, j
+end
+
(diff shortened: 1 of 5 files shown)


src/sd.jl
```diff
@@ -25,3 +25,113 @@ function addconstraint(m::Model, c::SDVariableConstraint)
...
+    # symmetry enforcing constraints, it is null if the matrix is symmetric
+    symref::Nullable{MOICON{MOI.VectorAffineFunction{T}, MOI.Zeros}}
+    # if k in symidx then invtrimap(k) has a symmetry enforcing constraint in symref
+    symidx::IntSet
+end
+
+constructconstraint!(x::AbstractMatrix, ::PSDCone) = SDConstraint(x)
+
+_constrain_symmetry(m::Model, c::SDConstraint{T, MT}, sdref) where {T, MT<:Symmetric} = sdref
+function fill_expr!(m::Model, expr::Dict{Variable, T}, terms, coeff=1) where T
```
joehuchette: Maybe ``coeff=one(T)``?

blegat: Good idea, done in 02ed6f8

test/sdp.jl
```diff
@@ -64,6 +67,141 @@
         @test eigmin(Xv) > -1e-6
     end
 
+#    @testset "Simple SDP" begin
```
joehuchette: Why keep this around?

blegat: I have copy pasted from old/sdp.jl and started to make some changes but it is not working yet. Since changes have already been made I didn't want to loose them so I kept it commented :)

joehuchette: Haha ok, carry on

mlubin:
This fixes https://github.com/JuliaOpt/JuMP.jl/issues/1050, correct?

test/sdp.jl
mlubin: FYI the plan is to remove most of the JuMP tests that involve solving a model and checking for the answer. Instead we'll check separately that JuMP generates the model correctly and that we talk to a fake solver correctly. The JuMP tests shouldn't be used for testing the solvers.

src/sd.jl
```diff
@@ -25,3 +25,113 @@ function addconstraint(m::Model, c::SDVariableConstraint)
     cref = MOI.addconstraint!(m.instance, MOI.VectorOfVariables([instancereference(c.Q[i, j]) for j in 1:n for i in 1:j]), MOI.PositiveSemidefiniteConeTriangle(n))
     return ConstraintRef(m, cref)
 end
+
+##########################################################################
+# SDConstraint is a (dual) semidefinite constraint of the form
+# ∑ cᵢ Xᵢ ⪰ D, where D is a n×n data matrix, cᵢ are scalars,
+# and Xᵢ are n×n symmetric variable matrices. The inequality
+# is taken w.r.t. the psd partial order.
+struct SDConstraint{T<:AbstractJuMPScalar, MT<:AbstractMatrix{T}} <: AbstractConstraint
```
mlubin: Can we just reuse `VectorAffExprConstraint` here?

src/sd.jl
```diff
@@ -25,3 +25,113 @@ function addconstraint(m::Model, c::SDVariableConstraint)
...
+function invtrimap(k::Integer)
+    i = isqrt(2k)
+    j = k - div((i-1)*i, 2)
+    i, j
+end
+
+struct SDSymInstanceRef{T}
+    sd::MOICON{MOI.VectorAffineFunction{T}, MOI.PositiveSemidefiniteConeTriangle}
+    # symmetry enforcing constraints, it is null if the matrix is symmetric
+    symref::Nullable{MOICON{MOI.VectorAffineFunction{T}, MOI.Zeros}}
```
mlubin: `Nullable` seems to be on its way out (https://github.com/JuliaLang/julia/pull/23642). Can we avoid using it?

src/sd.jl
```diff
@@ -25,3 +25,113 @@ function addconstraint(m::Model, c::SDVariableConstraint)
...
+    # if k in symidx then invtrimap(k) has a symmetry enforcing constraint in symref
+    symidx::IntSet
+end
+
+constructconstraint!(x::AbstractMatrix, ::PSDCone) = SDConstraint(x)
+
+_constrain_symmetry(m::Model, c::SDConstraint{T, MT}, sdref) where {T, MT<:Symmetric} = sdref
+function fill_expr!(m::Model, expr::Dict{Variable, T}, terms, coeff=one(T)) where T
```
mlubin: This is general code, shouldn't go into `sd.jl`.

test/sdp.jl
```diff
@@ -1,4 +1,7 @@
 @testset "Semidefinite Programming" begin
+
```
mlubin: Please add tests to `constraint.jl` that test model generation. The proposed representation of SD constraints is not obviously compatible with the existing `constraintobject` interface, which suggests a design issue. Maybe we need an MOI set which is a full square matrix that includes imposition of equality constraints if necessary.

src/sd.jl
blegat: Not really since here we have a matrix of `AffExpr` and not a vector of `AffExpr`. Moreover `VectorAffExprConstraint` assumes that the vector will be passed as is to the instance while here we need to only take the upper triangular part and check symmetry.

blegat:
Yes, it closes #1050 , I have added it in the head comment.

src/sd.jl
blegat: Fixed in 13f18916eef6c775d2ccfc13e86213c0eb37c625

src/sd.jl
blegat: Fixed in e5f66d8005708953140b9ede853822d4654825fc

test/sdp.jl
blegat: Fixed in 1649b42e1c22944ebef3ee75dbcc1bbc4410a185.

blegat:
I agree that it would be cleaner and more consistent to create an MOI set for Matrix that we constraint to be both symmetric and PSD.

Ideally, we should have a macro `wrapinstance` that does something like
```julia
@wrapinstance MosekInstance NewSet wrapper
```
that wraps an instance into a structure containing both the instance and the constraints of NewSet in an MOIU instance and use the function `wrapper` to translate these constraints into other constraints for the solver instance.
This would allow the solver wrapper to easily add support for other constraints out of the box if they can be translated without needing to write all the code.

Once this macro exists, we can create this set constraining symmetry and PSD, write the `wrapper` function which will be similar to what we have in this PR and all solvers will be able to add the support to that easily.

The macro would also allow user to add support for custom new set or functions to a solver.

mlubin:
Until we have an appropriate MOI representation for the nonsymmetric PSD constraint, it shouldn't be in JuMP. I'm going to fight here to keep confusing concepts like a constraint reference that refers to multiple MOI constraints out of the code, because they'll make things more difficult later on.

I would consider a PR that only supports the symmetric case, since that should be cleaner.

blegat:
I'm ok with that. I have update the PR, it now simply uses `VectorAffExprConstraint`.

test/constraint.jl
```diff
@@ -46,4 +46,19 @@
         # @test JuMP.isequal_canonical(c.func, -1 + x^2)
         # @test c.set == MOI.SecondOrderCone(1)
     end
+
+    @testset "Nonsensical SDPs" begin
```
mlubin: Could you also test `JuMP.constraintobject` as in the examples above?

blegat:
I have implemented constraint object for `Vector{AffExpr}`. To something clean, I use `map` but this needs https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/29 to be merged to work.

src/sd.jl
```diff
@@ -25,3 +25,11 @@ function addconstraint(m::Model, c::SDVariableConstraint)
     cref = MOI.addconstraint!(m.instance, MOI.VectorOfVariables([instancereference(c.Q[i, j]) for j in 1:n for i in 1:j]), MOI.PositiveSemidefiniteConeTriangle(n))
     return ConstraintRef(m, cref)
 end
+
+function constructconstraint!(x::AbstractMatrix, ::PSDCone)
+    n = Base.LinAlg.checksquare(x)
+    @assert issymmetric(x)
```
mlubin: Could you add a comment here discussing the plans for handling nonsymmetric matrices?

PR #1124, opened 2017-11-16
Implements resultvalue for affine expression

blegat:


src/affexpr.jl
```diff
@@ -125,6 +125,20 @@ function assert_isfinite(a::AffExpr)
     end
 end
 
+"""
+    resultvalue(v::AffExpr)
+
+Evaluate an `AffExpr` given the result returned by a solver.
+Replaces `getvalue` for most use cases.
+"""
+function resultvalue(a::AffExpr)
```
mlubin: Could you write a generic version that works for any ``VariableToValueMap`` and then have this as a single-line special case?

src/affexpr.jl
```diff
@@ -31,6 +31,21 @@ Base.zero(a::GenericAffExpr) = zero(typeof(a))
 Base.one( a::GenericAffExpr) =  one(typeof(a))
 Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.vars),copy(a.coeffs),copy(a.constant))
 
+"""
+    value(a::GenericAffExpr, map::Function)
+
+Evaluate `a` given the value `map(v)` for each variable `v`.
+"""
+function value(a::GenericAffExpr{T, V}, map::Function) where {T, V}
```
mlubin: Why not use a `VariableToValueMap` instead of a generic function? No need to infer the result type of `map`.

blegat: You cannot get the `VariableToValueMap` because you do not have a reference to the `model`.

mlubin: I'm confused. I mean define ``function value(a::GenericAffExpr{T, V}, map::VariableToValueMap{S}) where {T, V, S}``.

mlubin: Oh, this is too generic with respect to the variable type. Leave this for now, but we're going to change this later when we have a coherent story for dealing with the different result types.

PR #1129, opened 2017-11-21
Add support for constraint dual

### LINKED ISSUE #1127: resultdual for JuMP/MOI

blegat:
Closes #1127

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -100,7 +100,8 @@ mutable struct Model <: AbstractModel
     # obj#::QuadExpr
     # objSense::Symbol
 
-    # We use the constraint reference value to have a concrete type
+    # Mapping from the constraint reference in `instance`
+    # and the constraint reference in `solverinstance`
     constrainttosolverconstraint::Dict{MOICON,MOICON}
 
     # linconstr#::Vector{LinearConstraint}
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -46,7 +46,7 @@ export
     setlowerbound, setupperbound,
     #getlowerbound, getupperbound,
     #getvalue, setvalue,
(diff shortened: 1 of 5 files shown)


src/JuMP.jl
```diff
@@ -100,6 +100,9 @@ mutable struct Model <: AbstractModel
     # obj#::QuadExpr
     # objSense::Symbol
 
+    # We use the constraint reference value to have a concrete type
+    constrainttosolverconstraint::Dict{UInt64,UInt64}
```
mlubin: 1) I don't like how this is untyped.
2) I think it's incorrect, because constraint reference values are allowed to overlap if they're of different types.

blegat: I knew you were not going to like it :D

src/JuMP.jl
```diff
@@ -100,6 +100,9 @@ mutable struct Model <: AbstractModel
     # obj#::QuadExpr
     # objSense::Symbol
 
+    # We use the constraint reference value to have a concrete type
```
mlubin: Comment should be updated.

src/JuMP.jl
```diff
@@ -100,6 +100,9 @@ mutable struct Model <: AbstractModel
...
+    constrainttosolverconstraint::Dict{MOICON,MOICON}
```
mlubin: Doesn't this field need to be initialized in the Model() constructor? Why doesn't this code crash?

blegat: I am doing the same than what you did with `variabletosolvervariable`. It is initialized in every call of `attach`. If the variable is accessed before `attach` is ever called, there will be an `Undef...` error but this does not happen since the dictionary only have sense once a solver is attached.

PR #1134, opened 2017-11-25
Bump MPB requirement

mlubin:
to fix failures on nightly


mlubin:
Still failing, needs more investigation

PR #1138, opened 2017-11-26
use tuple recursion to fix type-stability of JuMPArray getindex()

rdeits:
I noticed a TODO in 9466fe9bc15caae3684f980b34d52a545a6c1d12 and figured I'd try to fix it. Rather than using an `@generated` function (which is generally not recommended), I managed to solve it with the lisp-y tuple recursion approach. With this change, `getindex()` on a JuMPArray with two axes goes from 1.8 us down to 60 ns on my machine (30X improvement). 

Note that to do this I had to put the type of the lookup dicts into the JuMPArray type. This is (essentially) what AxisArrays does, too, so I think it's reasonable.

### DIFF
--- a/src/JuMPArray.jl
+++ b/src/JuMPArray.jl
@@ -10,44 +10,71 @@
 # https://github.com/JuliaArrays/AxisArrays.jl/issues/84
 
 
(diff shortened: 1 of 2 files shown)


mlubin:
Awesome, looks great!

PR #1140, opened 2017-12-01
Update macro helper documentation

### LINKED ISSUE #827: update inline comments for buildrefsets and getloopedcode
The comments for these internal methods are out of date and I'm trying to update JuMPChance for 0.14. @IainNZ, I'm assuming JuMPeR is broken also.

joehuchette:
Closes #827

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -14,18 +14,6 @@ end
 
 include("parseexpr.jl")
 
-###############################################################################
-# buildrefsets
-# Unexported. Takes as input an object representing a name, associated index
-# sets, and conditions on those sets, for example
-# buildrefsets(:(x[i=1:3,[:red,:blue]],k=S; i+k <= 6))
-# Used internally in macros to build containers and constraints. Returns
-#       refcall:  Expr to reference a particular element, e.g. :(x[i,j,k])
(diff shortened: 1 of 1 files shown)


src/macros.jl
```diff
@@ -65,24 +53,33 @@ function buildrefsets(expr::Expr, cname)
...
+
+"""
+    JuMP.buildrefsets(expr::Expr)
+
+Helper function for macros to construct container objects. Takes an Expr that specifies the container, e.g. `:(x[i=1:3,[:red,:blue]],k=S; i+k <= 6)`, and returns:
+
+    1. `refcall`: Expr to reference a particular element in the container, e.g. :(x[i,red,s])
+    2. `idxvars`: Names for the index variables, e.g. [:i, gensym(), :k]
+    3. `idxsets`: Sets used for indexing, e.g. [1:3, [:red,:blue], S]
+    4. `condition`: Expr containing any conditional imposed on indexing, or :() if none is present
```
blegat: Shouldn't we also use backsticks for the exemples. So "e.g. `:(x[i,red,s])`" instead of "e.g. :(x[i,red,s])"

joehuchette: Yep

PR #1141, opened 2017-12-01
Specify nonlinear solver

cossio:
The first example in the docs for Nonlinear Modeling does not work if you don't specify a solver.


mlubin:
Thanks!

PR #1146, opened 2017-12-05
fix appveyor

blegat:


blegat:
Windows 32 bits is still not working because of CSDP but that dependency will be removed with dummy solvers.

mlubin:
Are there allowed failures? Would be nice to have a green checkmark.

PR #1148, opened 2017-12-10
Fix vectorized constraints

blegat:


src/macros.jl
```diff
@@ -391,7 +391,7 @@ macro constraint(args...)
             lhs = :($(x.args[2]) - $(x.args[3]))
             newaff, parsecode = parseExprToplevel(lhs, :q)
             if vectorized
-                constraintcall = :(addconstraint.($m, constructconstraint!($newaff,$set)))
+                constraintcall = :(addconstraint.($m, constructconstraint!.($newaff,$set)))
```
mlubin: There's a subtle point here that the simple comparisons would never involve multidimensional sets, right?

joehuchette: Simple comparisons meaning non-vectorized comparisons?

mlubin: Simple comparison in the sense of the comment above: ``# Simple comparison - move everything to the LHS``

blegat: Yes, currently `@constraint m x .== y` returns a multidimensional array of the same shape of `x` and `y` with constraint references for the set `MOI.EqualTo`. We could instead return a single constraint with `MOI.Zeros` but then we need to transform `x-y` to a `Vector` first.
What do you prefer ?

mlubin: I prefer the current behavior, but it deserves some in-line comments.

src/macros.jl
```diff
@@ -283,6 +283,7 @@ end
 #constructconstraint!(x::Array, sense::Symbol) = map(c->constructconstraint!(c,sense), x)
 #constructconstraint!(x::AbstractArray, sense::Symbol) = constructconstraint!([x[i] for i in eachindex(x)], sense)
 
+constructconstraint!(x::AbstractArray, set::MOI.AbstractScalarSet) = error("Dot operators should be used to compare array expressions.")
```
mlubin: - This could be more helpful, e.g., ``"Unexpected vector in scalar constraint. Did you mean to use the dot comparison operators like .==, .<=, and .>= instead?"``

- Does this also trigger for `[x,y] in LessThan(1.0)`? If so, the error message is not helpful at all.

blegat: With your suggestion of error message, it is clear that it is just a tip for common mistake so the case `[x,y] in LessThan(1.0)` is not an issue.

PR #1158, opened 2018-01-13
RFC: solver-free tests

### LINKED ISSUE #1137: testing JuMP without solvers
I'd like to make JuMP almost or entirely testable without any external solvers. The JuMP/MPB tests morphed into tests for solvers more so than tests for JuMP itself (https://github.com/JuliaOpt/JuMP.jl/issues/901) and also became really slow (https://github.com/JuliaOpt/JuMP.jl/issues/971) and fragile (https://github.com/JuliaOpt/JuMP.jl/issues/849). 

So we need to decide what exactly to test. I'd propose the following:
- Test that JuMP generates MOI instances correctly (https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/22 will be useful).
- Test that JuMP communicates with a solver correctly. For this we need a mock solver. This work needs to happen in conjunction with or after https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/issues/32.

mlubin:
I've been wanting to do this since September. We're living in the future.

~Before merging I'd like to implement duals in the mock solver and convert the soc and sdp tests (removing many that are no longer needed).~

Closes #1137

@odow @blegat


blegat:
It is the end of JuMP tests running for 30 minutes on Travis \o/

test/generate_and_solve.jl
```diff
@@ -157,7 +157,55 @@
...
+    @testset "SDP" begin
+        m = Model()
+        # TODO: PSD variable construction needs to be redone to make it possible
+        # to access the corresponding MOI constraint.
+        @variable(m, x[1:2,1:2], Symmetric) #, PSD)
+        setname(x[1,1], "x11")
+        setname(x[1,2], "x12")
+        setname(x[2,2], "x22")
+        @objective(m, Max, trace(x))
+        conpsd = @SDconstraint(m, x ⪰ [1.0 0.0; 0.0 1.0])
```
blegat: You should either use Min, or use ⪯ otherwise, the problem is unbounded.

test/generate_and_solve.jl
```diff
@@ -157,7 +157,55 @@
...
+
+        mocksolver = MOIU.MockSolverInstance(JuMP.JuMPInstance{Float64}())
+        JuMP.attach(m, mocksolver)
+
+        MOI.set!(mocksolver, MOI.TerminationStatus(), MOI.Success)
+        MOI.set!(mocksolver, MOI.ResultCount(), 1)
+        MOI.set!(mocksolver, MOI.PrimalStatus(), MOI.FeasiblePoint)
+        MOI.set!(mocksolver, MOI.DualStatus(), MOI.FeasiblePoint)
+        MOI.set!(mocksolver, MOI.VariablePrimal(), JuMP.solverinstanceindex(x[1,1]), 1.0)
+        MOI.set!(mocksolver, MOI.VariablePrimal(), JuMP.solverinstanceindex(x[1,2]), 2.0)
```
blegat: The optimal solution is 0 in both case mentioned in the above comment

test/generate_and_solve.jl
```diff
@@ -157,7 +157,55 @@
...
+        MOI.set!(mocksolver, MOI.VariablePrimal(), JuMP.solverinstanceindex(x[2,2]), 1.0)
+        MOI.set!(mocksolver, MOI.ConstraintDual(), JuMP.solverinstanceindex(conpsd), [1.0,2.0,3.0])
```
blegat: The optimal solution is [1, 0, 1] in both cases

mlubin:
I added a note that we're not attempting to use correct solutions at this point. It's sufficient to check that the input to the mock solver matches what comes out from JuMP.

PR #1162, opened 2018-01-22
direct solver and InstanceManager backends

### LINKED ISSUE #606: disable constraints
I would be nice to have a macro @removeconstraint() working also on a subset of the cns previously defined (?)
so that something like that works

@addConstraint(m, xyconstr[i=1:3,j=6:-2:2], x[i] - y[j] == 1)

@removeconstraint(m, xyconstr[i=1,j=6:-2:4])

Fabrizio

### LINKED ISSUE #1069: expose the concept of JuMP model being attached to a MathProgBase model
@sebmart made a useful suggestion that we should make it more explicit when a JuMP model is attached to a MPB model, and give the user the option to ignore, warn, or error on the case where we cannot pass through a JuMP model modification to the MPB model and need to toss it and regenerate from scratch.

mlubin:
The most interesting lines are probably https://github.com/JuliaOpt/JuMP.jl/compare/ml/instancemanager?expand=1#diff-6371f737b4b9a6f386c9816f184bfd72R177.

Oh BTW, we can delete constraints now (https://github.com/JuliaOpt/JuMP.jl/compare/ml/instancemanager?expand=1#diff-6371f737b4b9a6f386c9816f184bfd72R533),

There's *a lot* of API bikeshedding that needs to happen, but I'm okay merging and doing that on `master`.

Closes #1069
Closes #606

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -77,12 +77,18 @@ const MOIBIN = MOICON{MOI.SingleVariable,MOI.ZeroOne}
 @MOIU.instance JuMPInstance (ZeroOne, Integer) (EqualTo, GreaterThan, LessThan, Interval) (Zeros, Nonnegatives, Nonpositives, SecondOrderCone, RotatedSecondOrderCone, GeometricMeanCone, PositiveSemidefiniteConeTriangle, PositiveSemidefiniteConeSquare, RootDetConeTriangle, RootDetConeSquare, LogDetConeTriangle, LogDetConeSquare) () (SingleVariable,) (ScalarAffineFunction,ScalarQuadraticFunction) (VectorOfVariables,) (VectorAffineFunction,)
 
 ###############################################################################
-# Model class
-# Keeps track of all model and column info
+# Model
+
+# Model has three modes:
+# 1) Automatic: moibackend field holds an InstanceManager in Automatic mode.
+# 2) Manual: moibackend field holds an InstanceManager in Manual mode.
+# 3) Direct: moibackend field holds an AbstractSolverInstance. No extra copy of the instance is stored. The moibackend must support adddconstraint! etc.
+# Methods to interact with the InstanceManager are defined in solverinterface.jl.
+@enum ModelMode Automatic Manual Direct
+
 abstract type AbstractModel end
 mutable struct Model <: AbstractModel
 
-    instance::JuMPInstance{Float64}
     # special variablewise properties that we keep track of:
     # lower bound, upper bound, fixed, integrality, binary
     variabletolowerbound::Dict{MOIVAR,MOILB}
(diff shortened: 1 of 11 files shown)


mlubin:
I've updated this, ready for comments. If people are happy with the overall design (3 modes) then we can merge and refine the API on `master` as we've been doing for MOI.

src/JuMP.jl
```diff
@@ -169,23 +162,29 @@ mutable struct Model <: AbstractModel
...
         m.variabletozeroone = Dict{MOIVAR,MOIBIN}()
-        m.variablestart = VariableToValueMap{Float64}(m)
-        m.variableresult = VariableToValueMap{Float64}(m)
         m.customnames = Variable[]
         m.objbound = 0.0
         m.objval = 0.0
-        m.solverinstance = nothing
-        m.solverinstanceattached = false
+        if mode == Automatic
+            @assert solver === nothing
```
blegat: Why not do `resetsolver!` in case it is not `nothing` ?

src/JuMP.jl
```diff
@@ -544,17 +570,52 @@ Use `hasresultdual` to check if a result exists before asking for values.
...
+"""
+    canget(m::JuMP.Model, attr::MathOptInterface.AbstractInstanceAttribute)::Bool
 
-setname(cr::ConstraintRef{Model,<:MOICON}, s::String) = MOI.set!(cr.m.instance, MOI.ConstraintName(), instanceindex(cr), s)
+Return `true` if one may query the attribute `attr` from the model's MOI backend.
+false if not.
+"""
+MOI.canget(m::Model, attr::MOI.AbstractInstanceAttribute) = MOI.canget(m.moibackend, attr)
+MOI.canget(m::Model, attr::MOI.AbstractVariableAttribute, ::Type{Variable}) = MOI.canget(m.moibackend, attr, MOIVAR)
+MOI.canget(m::Model, attr::MOI.AbstractConstraintAttribute, ::Type{ConstraintRef{Model,T}}) where {T <: MOICON} = MOI.canget(m.moibackend, attr, T)
```
blegat: This a nice extension of `canget`!

src/solverinterface.jl
```diff
@@ -113,16 +39,7 @@ function solve(m::Model;
...
-    @assert m.solverinstanceattached
-    MOI.optimize!(m.solverinstance)
-
-    empty!(m.variableresult)
-    if MOI.canget(m.solverinstance, MOI.VariablePrimal(), MOI.VariableIndex)
-        for vindex in keys(m.variabletosolvervariable)
-            m.variableresult[Variable(m,vindex)] = MOI.get(m.solverinstance, MOI.VariablePrimal(), m.variabletosolvervariable[vindex])
-        end
-    end
+    MOI.optimize!(m.moibackend)
```
blegat: This function has become quite simple :D

src/JuMP.jl
mlubin: Done

PR #1195, opened 2018-03-08
Add support for VectorOfVariables in PSD constraints

### LINKED ISSUE #1161: access MOI constraint from PSD variable
https://github.com/JuliaOpt/JuMP.jl/blob/54933b0959a5b22edd5007b536d0553c96c47f3c/test/generate_and_solve.jl#L182-L184

blegat:
This allows to creates VectorOfVariables in PSDCTriangle constraints by doing
```julia
@variable m x[1:2,1:2] Symmetric
varpsd = @constraint m x in PSDCone()
```
or VectorOfVariables in PSDCSquares constraints by doing
```julia
@variable m x[1:2,1:2]
varpsd = @constraint m x in PSDCone()
```
To achieve this in a type stable way, `@variable` now returns a `Symmetric` matrix when building a symmetric matrix (i.e. when `Symmetric` or `PSD` are set).

This breaks the `*`, `+`, `-` operators defined between abstract matrix and scalar since this tries to use `setindex!` on `Symmetric` matrix.
It seems to me that these should be removed and replaced with the broadcast syntax.
I had to do some changes in `parseexpr.jl` to make sure we use the broadcast syntax internally.

Closes #1161


src/operators.jl
```diff
@@ -421,67 +421,70 @@ function _fillwithzeros(arr::AbstractArray{T}) where T
...
-    end
-    function $op(lhs::AbstractArray{S},rhs::T) where {T<:JuMPTypes,S}
-        ret = similar(lhs, typeof($op(zero(S), rhs)))
-        for I in eachindex(ret)
-            ret[I] = $op(lhs[I], rhs)
-        end
-        ret
-    end
-end; end
+# .+ and .- should be used instead
```
mlubin: How hard is this to address? I don't like the block of commented code, and it's not great either that removing this code didn't break any tests.

blegat: > How hard is this to address?

I don't mean to say that this code should be replaced by something that use broadcase, I mean to say that this code can be removed as the user shouldn't do `+(scalar, matrix)` but rather `.+(scalar, matrix)`.

> I don't like the block of commented code, and it's not great either that removing this code didn't break any tests.

It broke some tests, that's why I needed to add changes to `parseexpr.jl` but I agree that the coverage of `parseexpr.jl` (currently 37.19%) should be improved before we release v0.19.

mlubin: If this code is no longer needed, then just remove it.

src/parseexpr.jl
```diff
@@ -272,7 +272,7 @@ addtoexpr(ex::AbstractArray{T}, c::AbstractArray, x::AbstractArray) where {T<:Ge
 addtoexpr(ex::AbstractArray{T}, c::AbstractArray, x::Number) where {T<:GenericAffExpr} = append!.(ex, c*x)
 addtoexpr(ex::AbstractArray{T}, c::Number, x::AbstractArray) where {T<:GenericAffExpr} = append!.(ex, c*x)
 
-addtoexpr(ex, c, x) = ex + c*x
+addtoexpr(ex, c, x) = ex .+ c*x
```
mlubin: Why should we be doing this by default instead of asking the user to use `.+` instead of `+` when appropriate?

blegat: If I understand it correctly this is used in `parseExpr` by JuMP for constraints, e.g. `@SDconstraint Q >= eye(2)` is transformed into `addtoexpr(addtoexpr(zero(AffExpr), 1.0, Q), -1.0, eye(2))` so this is not due to a `+` in the user code

mlubin: Let's fix that then

blegat: What should be done instead ?

mlubin: In some cases I already switched from `zero(AffExpr)` to `Val{false}()` as the default "zero" value: https://github.com/JuliaOpt/JuMP.jl/blob/110fbb64693f3f51f990cc43b0a34863c25609d0/src/macros.jl#L768

blegat: Indeed, I have been able to remove my changes to parseexpr by doing that :)

Issue #1208, opened 2018-03-22
Solve with relaxation=true re-initializes Gurobi solver environment


vtjeng:
```julia
using JuMP
using Gurobi

m = Model(solver = GurobiSolver(OutputFlag=0))
@variable(m, 0 <= x[i=1:10] <= i)

function upperbound_mip(x)
    @objective(x.m, Max, x)
    solve(x.m)
    return getobjectivevalue(x.m)
end

function upperbound_lp(x)
    @objective(x.m, Max, x)
    solve(x.m, relaxation=true)
    return getobjectivevalue(x.m)
end
```

```julia
>>> map(upperbound_mip, x)

Academic license - for non-commercial use only
10-element Array{Float64,1}:
  1.0
  2.0
  3.0
  4.0
  5.0
  6.0
  7.0
  8.0
  9.0
 10.0

>>> map(upperbound_lp, x)

Academic license - for non-commercial use only
Academic license - for non-commercial use only
Academic license - for non-commercial use only
Academic license - for non-commercial use only
Academic license - for non-commercial use only
Academic license - for non-commercial use only
Academic license - for non-commercial use only
Academic license - for non-commercial use only
Academic license - for non-commercial use only
10-element Array{Float64,1}:
  1.0
  2.0
  3.0
  4.0
  5.0
  6.0
  7.0
  8.0
  9.0
 10.0
```

When calling `upperbound_mip` (with `relaxation=false`), we create a single Gurobi environment which is reused (as reflected in the single `Academic license message`); on ther other hand, it seems like we create multiple Gurobi environments. This makes solves with `relaxation=true` take more time than necessary (and means that aggregate solve times can be slower with `relaxation=true` if individual solves are very quick!)

This is with `julia 0.6.2`, `Gurobi.jl 0.3.3`, `JuMP 0.18.0`, and `Gurobi 7.5.2`. I did not previously observe the issue with `Gurobi 7.0.x`, but cannot confirm that it did not exist.

(I don't have any binary variables in this toy model, but I do have them in the model that I'm actually trying to solve).

odow:
See this section in the README: https://github.com/JuliaOpt/Gurobi.jl#reusing-the-same-gurobi-environment-for-multiple-solves

```julia
julia> using JuMP, Gurobi

julia> env = Gurobi.Env()
Academic license - for non-commercial use only
Gurobi.Env(Ptr{Void} @0x0000000000193ca0)

[...]

PR #1209, opened 2018-03-23
[release-0.18] Don't wipe internal model when relaxation=true

joehuchette:
ref #1208

### DIFF
--- a/src/solvers.jl
+++ b/src/solvers.jl
@@ -284,9 +284,6 @@ function solve(m::Model; suppress_warnings=false,
         end
     end
 
-    # don't keep relaxed model in memory
-    relaxation && (m.internalModelLoaded = false)
-
     # Return the solve status
     stat
 end


mlubin:
I did a quick check and didn't find a test that runs `solve(m, relaxation=true)` and then `solve(m, relaxation=false)` and checks that the integer constraints are respected.

odow:
There is a test here 

https://github.com/JuliaOpt/JuMP.jl/blob/aa22198d8448a3cad69b42772bb09f8a61b9ec70/test/model.jl#L857-L862

mlubin:
Ah, good find. That's tested with:
https://github.com/JuliaOpt/JuMP.jl/blob/aa22198d8448a3cad69b42772bb09f8a61b9ec70/test/solvers.jl#L64-L67
so there's no coverage on travis. Could we check that this works with Cbc and GLPK (without asking for duals)?

odow:
Just tested with Cbc, Gurobi, and CPLEX. All seems to work.

mlubin:
And GLPK?

PR #1213, opened 2018-03-30
Fix container printing tests for 32 bits

blegat:


mlubin:
See the failing tests

mlubin:
If this passes, could you also remove the allowed failure on appveyor julia 0.6 32-bit? That's the only CI coverage we have for 32-bit.

PR #1216, opened 2018-04-05
Fix #1193

### LINKED ISSUE #1193: Nested tuple destructuring in macros
```jl
m = Model()
@variable(m, x)
d = Dict((1,2) => 3)
@constraint(m, x == sum(i+j+k for ((i,j),k) in d))
```
```
ERROR: AssertionError: x.head == :escape
```
whereas
```
d = Dict((1,2) => 3)
sum(i+j+k for ((i,j),k) in d)
```
```
6
```

joehuchette:
This is against the release branch; not sure how best to open it against master due to the changes in tests.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -206,7 +206,7 @@ localvar(x::Symbol) = _localvar(x)
 localvar(x::Expr) = Expr(:block, _localvar(x)...)
 _localvar(x::Symbol) = :(local $(esc(x)))
 function _localvar(x::Expr)
-    @assert x.head == :escape
+    @assert x.head in (:escape,:tuple)
     args = Any[]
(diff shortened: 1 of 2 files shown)


mlubin:
These tests haven't been ported over to `master` yet. I'll take a look.

mlubin:
Note we can remove all this `localvar` junk in Julia 0.7.

[NEWS.md](https://github.com/JuliaLang/julia/blob/master/NEWS.md):
> In for i = ..., if a local variable i already existed it would be overwritten during the loop. This behavior is deprecated, and in the future for loop variables will always be new variables local to the loop (#22314). The old behavior of overwriting an existing variable is available via for outer i = ....

PR #1236, opened 2018-04-17
Add sum constraint test

blegat:


test/constraint.jl
```diff
@@ -200,4 +200,55 @@
         @test macroexpand(:(@variable(m, -rand(5,5) <= nonsymmetric[1:5,1:5] <= rand(5,5), Symmetric))).head == :error
     end
 
+    @testset "[macros] sum(generator)" begin
+        m = Model()
+        @variable(m, x[1:3,1:3])
+        @variable(m, y)
+        C = [1 2 3; 4 5 6; 7 8 9]
+
+        cref = @constraint(m, sum( C[i,j]*x[i,j] for i in 1:2, j = 2:3 ) <= 1)
```
mlubin: It would be more direct to use `@expression` here so you don't need to re-extract the `AffExpr` from the constraint. These tests may have been written originally before `@expression` existed.

test/constraint.jl
```diff
@@ -200,4 +200,55 @@
...
+        c = JuMP.constraintobject(cref, AffExpr, MOI.EqualTo)
+        @test JuMP.isequal_canonical(c.func, sum( 0*x[i,1] for i=1:3))
+        @test c.set == MOI.EqualTo(0.0)
+
+        cref = @constraint(m, sum( 0*x[i,1] + y for i=1:3) == 0)
+        c = JuMP.constraintobject(cref, AffExpr, MOI.EqualTo)
+        @test JuMP.isequal_canonical(c.func, sum( 0*x[i,1] + y for i=1:3))
+        @test c.set == MOI.EqualTo(0.0)
+
+        #@test isexpr(macroexpand(:(@constraint(m, sum( 0*x[i,1] + y for i=1:3 for j in 1:3) == 0))),:error) == true
```
mlubin: Either remove or add a comment explaining why the test is commented out.

blegat: It was already commented out in `old/macro.jl`

test/constraint.jl
```diff
@@ -200,4 +200,31 @@
         @test macroexpand(:(@variable(m, -rand(5,5) <= nonsymmetric[1:5,1:5] <= rand(5,5), Symmetric))).head == :error
     end
 
+    @testset "[macros] sum(generator)" begin
+        m = Model()
+        @variable(m, x[1:3,1:3])
+        @variable(m, y)
+        C = [1 2 3; 4 5 6; 7 8 9]
+
+        @test JuMP.isequal_canonical(@expression(m, sum( C[i,j]*x[i,j] for i in 1:2, j = 2:3 )),
```
mlubin: Optional, or add a TODO: this calls for a small macro that lets you avoid writing the expression twice.

blegat: Good idea

PR #1240, opened 2018-04-18
Add more constraint tests

blegat:
Due to the use of `Val{false}`, `aff` can now be a variable or a number in range constraints hence we also need to define `coeftype` for these.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -268,17 +268,12 @@ const ScalarPolyhedralSets = Union{MOI.LessThan,MOI.GreaterThan,MOI.EqualTo,MOI.
 constructconstraint!(v::Variable, set::MOI.AbstractScalarSet) = SingleVariableConstraint(v, set)
 constructconstraint!(v::Vector{Variable}, set::MOI.AbstractVectorSet) = VectorOfVariablesConstraint(v, set)
 
+constructconstraint!(α::Number, set::MOI.AbstractScalarSet) = constructconstraint!(convert(AffExpr, α), set)
 function constructconstraint!(aff::AffExpr, set::S) where S <: Union{MOI.LessThan,MOI.GreaterThan,MOI.EqualTo}
     offset = aff.constant
     aff.constant = 0.0
     return AffExprConstraint(aff, S(MOIU.getconstant(set)-offset))
 end
-
-# function constructconstraint!(aff::AffExpr, lb, ub)
-#     offset = aff.constant
-#     aff.constant = 0.0
-#     AffExprConstraint(aff, lb-offset, ub-offset)
-# end
 
 constructconstraint!(x::AbstractArray, set::MOI.AbstractScalarSet) = error("Unexpected vector in scalar constraint. Did you mean to use the dot comparison operators like .==, .<=, and .>= instead?")
 constructconstraint!(x::Vector{AffExpr}, set::MOI.AbstractVectorSet) = VectorAffExprConstraint(x, set)
@@ -309,17 +304,19 @@ end
 # end
 
 # three-argument constructconstraint! is used for two-sided constraints.
+constructconstraint!(v::Variable, lb::Real, ub::Real) = SingleVariableConstraint(v, MOI.Interval(lb, ub))
+
 function constructconstraint!(aff::AffExpr, lb::Real, ub::Real)
     offset = aff.constant
     aff.constant = 0.0
     AffExprConstraint(aff,MOI.Interval(lb-offset,ub-offset))
 end
 
-# constructconstraint!(aff::Variable, lb::Real, ub::Real) = constructconstraint!(convert(AffExpr,v),lb,ub)
-
 constructconstraint!(q::QuadExpr, lb, ub) = error("Two-sided quadratic constraints not supported. (Try @NLconstraint instead.)")
 
 constraint_error(args, str) = error("In @constraint($(join(args,","))): ", str)
+
+coeftype(::T) where T<:Number = T
 
 # TODO: update 3-argument @constraint macro to pass through names like @variable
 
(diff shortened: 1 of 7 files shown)


src/macros.jl
```diff
@@ -309,18 +304,20 @@ end
 # end
 
 # three-argument constructconstraint! is used for two-sided constraints.
+constructconstraint!(v::Variable, lb::Real, ub::Real) = SingleVariableConstraint(v, MOI.Interval(lb, ub))
```
mlubin: What happens when `lb` and `ub` are integers, and the solver only supports `Interval{Float64}`?

src/variables.jl
```diff
@@ -11,6 +11,7 @@ struct Variable <: AbstractJuMPScalar
     m::Model
     index::MOIVAR
 end
+coeftype(::Variable) = Float64
```
mlubin: This is a sketchy helper function for the macros. I think it should be defined in `macros.jl`.

src/macros.jl
```diff
@@ -448,12 +445,12 @@ macro constraint(args...)
                 try
```
mlubin: This whole `coeftype` thing doesn't seem well thought-out. Maybe we should get rid of it.

src/macros.jl
blegat: Currently this doesn't happen because they are initialized as floating point number
https://github.com/JuliaOpt/JuMP.jl/blob/e6c9f401b63afa630a0af720b6ed507bc0b97dc9/src/macros.jl#L434-L437
I'd say that it is the responsability of the caller of `constructconstraint!` to use relevant types for `lb` and `ub`.

blegat:
I have replaced the `lbval`/`ubval` error with an error at the `constructconstraint!` level and added the `_error` function to the `constructconstraint!` function to have clearer errors (it also improve the error for range quadratic constraints).
The new error is now
```julia
julia> using JuMP
julia> m = Model()
julia> @variable(m, x)
julia> @variable(m, y)
julia> t = 10.0
julia> @constraint(m, x <= t <= y)
ERROR: In @constraint(m,x <= t <= y): Expected x to be a number.
Stacktrace:
 [1] constraint_error(::Array{Any,1}, ::String) at /home/blegat/.julia/v0.6/JuMP/src/macros.jl:323
 [2] constructconstraint!(::JuMP.#_error#78, ::Float64, ::JuMP.GenericAffExpr{Float64,JuMP.Variable}, ::JuMP.GenericAffExpr{Float64,JuMP.Variable}) at /home/blegat/.julia/v0.6/JuMP/src/macros.jl:319
 [3] macro expansion at ./REPL.jl:97 [inlined]
 [4] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73
```

src/macros.jl
```diff
@@ -265,25 +265,20 @@ function sense_to_set(sense::Symbol)
...
 
-# function constructconstraint!(aff::AffExpr, lb, ub)
-#     offset = aff.constant
-#     aff.constant = 0.0
-#     AffExprConstraint(aff, lb-offset, ub-offset)
-# end
-
-constructconstraint!(x::AbstractArray, set::MOI.AbstractScalarSet) = error("Unexpected vector in scalar constraint. Did you mean to use the dot comparison operators like .==, .<=, and .>= instead?")
-constructconstraint!(x::Vector{AffExpr}, set::MOI.AbstractVectorSet) = VectorAffExprConstraint(x, set)
+constructconstraint!(_error::Function, x::AbstractArray, set::MOI.AbstractScalarSet) = error("Unexpected vector in scalar constraint. Did you mean to use the dot comparison operators like .==, .<=, and .>= instead?")
```
mlubin: error on the RHS should be `_error`

src/macros.jl
```diff
@@ -309,17 +304,24 @@ end
...
+constructconstraint!(_error::Function, v::Variable, lb::Real, ub::Real) = SingleVariableConstraint(v, MOI.Interval(lb, ub))
+
+function constructconstraint!(_error::Function, aff::AffExpr, lb::Real, ub::Real)
     offset = aff.constant
     aff.constant = 0.0
     AffExprConstraint(aff,MOI.Interval(lb-offset,ub-offset))
 end
 
-# constructconstraint!(aff::Variable, lb::Real, ub::Real) = constructconstraint!(convert(AffExpr,v),lb,ub)
+constructconstraint!(_error::Function, q::QuadExpr, lb, ub) = error("Two-sided quadratic constraints not supported. (Try @NLconstraint instead.)")
```
mlubin: same here

src/macros.jl
```diff
@@ -309,17 +304,24 @@ end
...
 
-constructconstraint!(q::QuadExpr, lb, ub) = error("Two-sided quadratic constraints not supported. (Try @NLconstraint instead.)")
+function constructconstraint!(_error::Function, expr, lb, ub)
+    @assert !(lb isa Number) || !(ub isa Number)
```
mlubin: What's the point of this assert?

src/macros.jl
```diff
@@ -309,17 +304,24 @@ end
...
+    lb isa Number || _error(string("Expected $lb to be a number."))
+    ub isa Number || _error(string("Expected $ub to be a number."))
+end
 
-constraint_error(args, str) = error("In @constraint($(join(args,","))): ", str)
+coeftype(::Variable) = Float64
```
mlubin: Is coeftype still needed for anything?

src/macros.jl
blegat: It could happen that a call is dispatched to this method with `lb` and `ub` that are not `Number` if there is not `constructconstraint!` method defined for the type of `expr`. In that case, no error will be thrown that the `@constraint` will silently do nothing. We want to avoid that

PR #1247, opened 2018-04-26
Add addvariable function

### LINKED ISSUE #1172: constructconstraint! is not consistent with constructvariable!
The former creates a constraint that's *not* attached to the model, while the latter creates a variable that *is* attached to the model. Probably one should be renamed. This is also related to the inconsistency between Variable and ConstraintRef:
https://github.com/JuliaOpt/JuMP.jl/blob/2f6d6e570a37922f76606c99694a2b0d35b5629b/src/JuMP.jl#L268-L269

blegat:
Closes #1172

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -849,7 +849,8 @@ esc_nonconstant(x::Number) = x
 esc_nonconstant(x::Expr) = isexpr(x,:quote) ? x : esc(x)
 esc_nonconstant(x) = esc(x)
 
-# Returns the type of what `buildvariable` would return with these starting positional arguments.
+# Returns the type of what `addvariable(::Model, buildvariable(...))` would return where `...` represents the positional arguments.
+# E.g.: @variable m [1:3] foo will allocate an vector of element type `variabletype(foo())`
 variabletype(m::Model) = VariableRef
 # Returns a new variable belonging to the model `m`. Additional positional arguments can be used to dispatch the call to a different method.
 # The return type should only depends on the positional arguments for `variabletype` to make sense.
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -849,51 +849,15 @@ esc_nonconstant(x::Number) = x
 esc_nonconstant(x::Expr) = isexpr(x,:quote) ? x : esc(x)
 esc_nonconstant(x) = esc(x)
 
-mutable struct VariableInfo{S, T, U, V}
-    haslb::Bool
-    lowerbound::S
-    hasub::Bool
-    upperbound::T
-    hasfix::Bool
-    fixedvalue::U
-    hasstart::Bool
-    start::V
-    binary::Bool
-    integer::Bool
-    name::String
-end
-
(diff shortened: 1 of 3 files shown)


src/variables.jl
```diff
@@ -3,6 +3,24 @@
...
+    upperbound::T
+    hasfix::Bool
+    fixedvalue::U
+    hasstart::Bool
+    start::V
+    binary::Bool
+    integer::Bool
+end
+
+struct ScalarVariable{S, T, U, V}
```
mlubin: `JuMP.Variable` needs to be renamed to `JuMP.VariableRef` before we can have an object called `ScalarVariable` with this meaning.

mlubin: (If you do that, make it a separate PR and then update this one afterwards.)

blegat: I have rebase the PR on master, it is now ready for review :)

src/macros.jl
```diff
@@ -849,51 +849,15 @@ esc_nonconstant(x::Number) = x
...
-    fixedvalue::U
-    hasstart::Bool
-    start::V
-    binary::Bool
-    integer::Bool
-    name::String
-end
-
-# Returns the type of what `constructvariable!` would return with these starting positional arguments.
+# Returns the type of what `buildvariable` would return with these starting positional arguments.
```
mlubin: Does this still make sense? The positional arguments were replaced by the `VariableInfo` struct.

src/variables.jl
```diff
@@ -3,6 +3,25 @@
...
```
mlubin: Is `ScalarVariable` supposed to inherit from `AbstractVariable`?

src/variables.jl
```diff
@@ -389,3 +408,35 @@ resultvalue(v::VariableRef) = MOI.get(v.m, MOI.VariablePrimal(), v)
 hasresultvalues(m::Model) = MOI.canget(m, MOI.VariablePrimal(), VariableRef)
 
 @Base.deprecate setvalue(v::VariableRef, val::Number) setstart(v, val)
+
+"""
+    addvariable(m::Model, v::AbstractVariable, name::String="")
```
mlubin: The signature in the documentation should match the method you're defining. (`ScalarVariable` vs. `AbstractVariable`)

src/macros.jl
```diff
@@ -849,51 +849,17 @@ esc_nonconstant(x::Number) = x
...
-    hasstart::Bool
-    start::V
-    binary::Bool
-    integer::Bool
-    name::String
-end
-
-# Returns the type of what `constructvariable!` would return with these starting positional arguments.
+# Returns the type of what `addvariable(::Model, buildvariable(...))` would return where `...` represents the positional arguments.
+# Example: `@variable m [1:3] foo` will allocate an vector of element type `variabletype(foo())`
```
mlubin: `variabletype(m, foo)`? (missing `m` and I think the parenthesis are incorrect)

PR #1249, opened 2018-04-26
Overload operators with Base. prefix instead of importall

blegat:
This removes quite a lot of warnings on v0.7


blegat:
No that was quick. I don't understand the coverage bump by the way but it's nice ^^

PR #1251, opened 2018-04-27
Add Base. prefix for vecdot instead of importing it

blegat:


src/operators.jl
```diff
@@ -235,10 +233,9 @@ Base.dot(lhs::AbstractVector{T},rhs::AbstractVector{S}) where {T<:JuMPTypes,S<:J
 Base.dot(lhs::AbstractVector{T},rhs::AbstractVector{S}) where {T<:JuMPTypes,S} = _dot(lhs,rhs)
 Base.dot(lhs::AbstractVector{T},rhs::AbstractVector{S}) where {T,S<:JuMPTypes} = _dot(lhs,rhs)
 
-# TODO: qualify Base.vecdot once v0.3 support is dropped
```
mlubin: Yeah it's safe to do this now :|

PR #1256, opened 2018-04-30
Fix tests of macro errors in Julia v0.7

blegat:
The macroexpand way to test if a macro throws an error during pre-compilation does not work on Julia v0.7 because of https://github.com/JuliaLang/julia/pull/23533. One solution is suggested [here](https://discourse.julialang.org/t/test-throws-with-macros-after-pr-23533/5878)

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -20,12 +20,8 @@ const MOIU = MOI.Utilities
 
 # Test that the macro call `m` throws an error exception during pre-compilation
 macro test_macro_throws(m)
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -18,6 +18,12 @@ const MOI = MathOptInterface
 const MOIT = MOI.Test
 const MOIU = MOI.Utilities
 
+# Test that the macro call `m` throws an error exception during pre-compilation
+macro test_macro_throws(m)
```
mlubin: To be consistent with `@test_throws`, this could be `@test_macro_throws(errortype, ...)` or `@test_macro_throws_error`.

PR #1260, opened 2018-05-01
Introduce AbstractVariableRef

blegat:
JuMP extensions may define new variable types returned by `addvariable` that should also behave like classical variables for standard arithmetic operations (e.g. as discussed with @mlubin on Gitter, StructJuMP may create its own variable type not tied to MOI).
This PR creates the `AbstractVariableRef` for holding variables as returned by `addvariable`/`@variable`. It also generalize some operations defined in `operators.jl` to work on any `AbstractVariableRef`

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -25,7 +25,7 @@ end
 """
     AbstractVariableRef
 
-Variable returned by [`addvariable`](@ref). Affine (resp. quadratic) operations with variables of type `V<:AbstractVariableRef` and coefficients of type `T` creates a `GenericAffExpr{T,V}` (resp. `GenericQuadExpr{T,V}`).
+Variable returned by [`addvariable`](@ref). Affine (resp. quadratic) operations with variables of type `V<:AbstractVariableRef` and coefficients of type `T` create a `GenericAffExpr{T,V}` (resp. `GenericQuadExpr{T,V}`).
 """
 abstract type AbstractVariableRef <: AbstractJuMPScalar end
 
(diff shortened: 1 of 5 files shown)


mlubin:
`src/operators.jl` doesn't have enough test coverage for me to trust these changes

mlubin:
The diff coverage is 84.61%. It should really be 100% (bugs in operators are not nice).

src/variables.jl
```diff
@@ -22,10 +22,21 @@ struct ScalarVariable{S, T, U, V} <: AbstractVariable
...
-# Holds a reference to the model and the corresponding MOI.VariableIndex.
-struct VariableRef <: AbstractJuMPScalar
+"""
+    AbstractVariableRef
+
+Variable returned by [`addvariable`](@ref). Affine (resp. quadratic) operations with variables of type `V<:AbstractVariableRef` and coefficients of type `T` create a `GenericAffExpr{T,V}` (resp. `GenericQuadExpr{T,V}`).
+"""
+abstract type AbstractVariableRef <: AbstractJuMPScalar end
+
+variablereftype(v::AbstractVariableRef) = typeof(v)
```
mlubin: What's the point of this method?

mlubin:
Merging this so that I can rebase an incoming PR on top of it. The question about `variablereftype` is still unresolved.

src/variables.jl
blegat: GenericAffExpr and GenericQuadExpr both work on arbitrary coefficient type and AbstractVariableRef type.
To write code that works for any `VariableRef`, it is often useful to get the variable ref type of an object that can be either an `AbstractVariableRef` or a `GenericAffExpr` or a `GenericQuadExpr`, e.g.
https://github.com/JuliaOpt/JuMP.jl/blob/151c75fba8e7a2f8b36e96aa29909441b06e9207/src/operators.jl#L73-L83

PR #1278, opened 2018-05-08
Remove unary multiplication

blegat:
These lines are not covered by the tests and not covered by the tests of JuMP v0.18 either.
Moreover, it does not seem to work
```julia
julia> *1
ERROR: syntax: "*" is not a unary operator
julia> m = Model()
julia> @variable m x
julia> *x
ERROR: syntax: "*" is not a unary operator
```


mlubin:
The way to call it is `*(x)`.

mlubin:
At one point this was needed for the macros. The macros don't have full coverage, so even if tests pass it's unclear if it's safe to remove.

blegat:
I have readded the first line as it is tested in https://github.com/JuliaOpt/JuMP.jl/pull/1288. However the last two can be removed since they are covered by the first one (i.e. `AffExpr, QuadExpr <: AbstractJuMPScalar`).

mlubin:
I still see the line with `Base.:*(lhs::AbstractJuMPScalar) = lhs` being deleted. Is that intended?

blegat:
No, it's not intended. I must have forgot to pull before rebasing

PR #1279, opened 2018-05-08
Add concatenation tests

blegat:


test/operator.jl
```diff
@@ -529,4 +529,47 @@ using Compat.Test
             @test all(c -> c.set.upper == 3, c7)
         end
     end
+
+    @testset "JuMPArray concatenation" begin
```
mlubin: Are there actually any `JuMPArray`s here? Looks like plain `Array`s to me. The tests name and comments seem out of date. You could then also question why we're testing `vcat` and `hcat` for plain julia `Array`s.

blegat: > Are there actually any JuMPArrays here?

No, it's two Vector and two Matrix

> You could then also question why we're testing vcat and hcat for plain julia Arrays.

This doubt is supported by the Codecov Report

mlubin: Let's drop these tests then

PR #1296, opened 2018-05-10
Replace vec_eq and add vector creation tests

blegat:


test/operator.jl
```diff
@@ -285,35 +285,36 @@ Base.transpose(t::MySumType) = MySumType(t.a)
...
-
-    function vec_eq(x::Array{QuadExpr}, y::Array{QuadExpr})
-        size(x) == size(y) || return false
-        for i in 1:length(x)
-            JuMP.isequal_canonical(x[i], y[i]) || return false
+    @testset "Vectorized operations" begin
+        @testset "Array creation" begin
+            m = Model()
+            @variable(m, x)
+            # @inferred requires a call expression so [x, 1] does not work but Base.vect(x, 1) works
```
mlubin: Why do we need to be testing `@inferred` of `[x, 1]` anyway? Maybe `test_expression_with_string` shouldn't call `inferred`.

blegat: This was aiming to cover the `promote_rule` calls of operators.jl but it didn't work apparently. We could simply call `promote_type` in the tests but I was looking for something less direct

mlubin: It's usually better to directly test what you're intending to test.

PR #1301, opened 2018-05-13
Refactor AffExpr and QuadExpr to use OrderedDict for internal storage.

mlubin:
This means that duplicates are no longer possible. I left zeros in there for now. We might decide to always remove them. I also expect some bikeshedding on the API (`push!`/`append!` are weird names now).

See https://github.com/JuliaOpt/MathOptInterface.jl/issues/234#issuecomment-366868878 for initial benchmarks. I haven't benchmarked the new setup but paid attention to implementing things efficiently. It's not obvious how to do a fair comparison given that `master` doesn't currently remove duplicates, and too much has changed since 0.18.

We can probably close https://github.com/JuliaOpt/MathOptInterface.jl/issues/234 by saying that solvers don't need to accept duplicates.

I'll track down any coverage gaps if the diff is not 100% covered.

@rdeits @vtjeng @ExpandingMan

### DIFF
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -128,46 +128,50 @@ Base.done( lti::LinearTermIterator, state::Int) = done(lti.aff.terms, state)
 Base.next( lti::LinearTermIterator, state::Int) = reorder_iterator(next(lti.aff.terms, state)...)
 Base.length(lti::LinearTermIterator) = length(lti.aff.terms)
 
-# TODO: Consider renaming. push! is an unusual name to update a term in a
-# dictionary.
 """
-    Base.push!{C,V}(aff::GenericAffExpr{C,V}, new_coef::C, new_var::V)
+    add_to_expression!(expression, terms...)
 
-An efficient way to add to an affine expression in place. For example, to add
-`5x` to an existing expression `aff`, use `push!(aff, 5.0, x)`. This is
-*significantly* more efficient than `aff += 5.0*x`.
+Updates `expression` *in place* to `expression + (*)(terms...)`. This is
+typically much more efficient than `expression += (*)(terms...)`. For example,
+`add_to_expression!(expression, a, b)` produces the same result as `expression
++= a*b`, and `add_to_expression!(expression, a)` produces the same result as
+`expression += a`.
+
+Only a few methods are defined, mostly for internal use, and only for the cases
+when (1) they can be implemented efficiently and (2) `expression` is capable of
+storing the result. For example, `add_to_expression!(::AffExpr, ::VariableRef,
+::VariableRef)` is not defined because a `GenericAffExpr` cannot store the
+product of two variables.
 """
-function Base.push!(aff::GenericAffExpr{C,V}, new_coef::C, new_var::V) where {C,V}
+function add_to_expression! end
+
+# TODO: add deprecations for Base.push! and Base.append!
+
+function add_to_expression!(aff::GenericAffExpr{C,V}, new_coef::C, new_var::V) where {C,V}
     add_or_set!(aff.terms, new_var, new_coef)
     aff
 end
 
-# TODO: Consider renaming. append! is an unusual name to add two expressions.
-"""
-    Base.append!{C,V}(aff::GenericAffExpr{C,V}, other::GenericAffExpr{C,V})
+function add_to_expression!(aff::GenericAffExpr{C,V}, new_var::V) where {C,V}
+    add_or_set!(aff.terms, new_var, one(C))
+    aff
+end
 
-Efficiently append the terms of an affine expression to an existing affine
-expression. For example, given `aff = 5.0*x` and `other = 7.0*y + 3.0*z`,
-we can add to `aff` in place by using `append!(aff, other)`.This results in
-`aff` equal to `5x + 7y + 3z`. `append!` is *significantly* more efficient than
-`aff += other`.
-"""
-function Base.append!(aff::GenericAffExpr{C,V}, other::GenericAffExpr{C,V}) where {C,V}
+function add_to_expression!(aff::GenericAffExpr{C,V}, other::GenericAffExpr{C,V}) where {C,V}
     merge!(+, aff.terms, other.terms)
     aff.constant += other.constant
     aff
 end
-# For consistency, allow appending constants and individual variables
-Base.append!(aff::GenericAffExpr{C,C}, other::C) where {C} = error() # for ambiguity
-function Base.append!(aff::GenericAffExpr{C,V}, other::C) where {C,V}
+# TODO: do we need this?
+#add_to_expression!(aff::GenericAffExpr{C,C}, other::C) where {C} = error() # for ambiguity
+function add_to_expression!(aff::GenericAffExpr{C,V}, other::C) where {C,V}
     aff.constant += other
     aff
 end
-function Base.append!(aff::GenericAffExpr{C,V}, other::Real) where {C,V}
+function add_to_expression!(aff::GenericAffExpr{C,V}, other::Real) where {C,V}
     aff.constant += other
     aff
 end
-Base.append!(aff::GenericAffExpr{C,V}, other::V) where {C,V} = push!(aff,one(C),other)
 
 function Base.isequal(aff::GenericAffExpr{C,V},other::GenericAffExpr{C,V}) where {C,V}
     return isequal(aff.constant, other.constant) && isequal(aff.terms, other.terms)
@@ -235,7 +239,7 @@ end
 function AffExpr(m::Model, f::MOI.ScalarAffineFunction)
     aff = AffExpr()
     for i in 1:length(f.variables)
-        push!(aff, f.coefficients[i], VariableRef(m,f.variables[i]))
+        add_to_expression!(aff, f.coefficients[i], VariableRef(m,f.variables[i]))
     end
     aff.constant = f.constant
     return aff
@@ -301,7 +305,7 @@ end
 function Base.copy(a::GenericAffExpr, new_model::Model)
     result = zero(a)
     for (coef, var) in linearterms(a)
-        push!(result, coef, copy(v, new_model))
+        add_to_expression!(result, coef, copy(v, new_model))
     end
     result.constant = a.constant
     return result
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -79,7 +79,7 @@ Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.constant), copy(a.terms))
 
 GenericAffExpr{C, V}() where {C, V} = zero(GenericAffExpr{C, V})
 
-function map_coefficients!(a::GenericAffExpr, f::Function)
+function map_coefficients_inplace!(f::Function, a::GenericAffExpr)
     # The iterator remains valid if existing elements are updated.
     for (coef, var) in linearterms(a)
         a.terms[var] = f(coef)
@@ -88,8 +88,8 @@ function map_coefficients!(a::GenericAffExpr, f::Function)
     return a
 end
 
-function map_coefficients(a::GenericAffExpr, f::Function)
-    return map_coefficients!(copy(a), f)
+function map_coefficients(f::Function, a::GenericAffExpr)
+    return map_coefficients_inplace!(f, copy(a))
 end
 
 Base.sizehint!(a::GenericAffExpr, n::Int) = sizehint!(a.terms, n)
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -15,23 +15,83 @@
 # Operator overloads in src/operators.jl
 #############################################################################
 
+# Utilities for OrderedDict
+function add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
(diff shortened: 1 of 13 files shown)


vtjeng:
Thanks @mlubin! (For the problems that I've been working with, simplifying the `AffExpr`s has in some instances been taking the majority of the time, so this would be a great help).

@rdeits - this means that simplification will no longer be needed in the next release of `ConditionalJuMP` that builds on this version of `JuMP`, right?

rdeits:
Yup, although that might just happen when I rewrite it for MathOptInterface.

Anyway, this is great!

src/affexpr.jl
```diff
@@ -15,23 +15,83 @@
 # Operator overloads in src/operators.jl
 #############################################################################
 
+# Utilities for OrderedDict
+function add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
+    # TODO: This unnecessarily requires two lookups for k.
+    # TODO: Decide if we want to drop zeros here.
```
blegat: I am in favor of removing the zeros here

src/affexpr.jl
```diff
@@ -15,23 +15,83 @@
...
+    terms::OrderedDict{VarType,CoefType}
 end
 
 variablereftype(::GenericAffExpr{C, V}) where {C, V} = V
 
-Base.iszero(a::GenericAffExpr) = isempty(a.vars) && iszero(a.constant)
-Base.zero(::Type{GenericAffExpr{C,V}}) where {C,V} = GenericAffExpr{C,V}(V[],C[],zero(C))
-Base.one(::Type{GenericAffExpr{C,V}}) where { C,V} = GenericAffExpr{C,V}(V[],C[], one(C))
+function GenericAffExpr(constant::V, kv::AbstractArray{Pair{K,V}}) where {K,V}
+    result = GenericAffExpr{V,K}(constant, new_ordered_dict(K, V, kv))
```
blegat: Why doing `result =` and not just `return` ?

src/affexpr.jl
```diff
@@ -100,51 +170,39 @@ end
...
-Base.convert(::Type{GenericAffExpr{T,V}}, v::V)    where {T,V} = GenericAffExpr([v], [one(T)], zero(T))
-Base.convert(::Type{GenericAffExpr{T,V}}, v::Real) where {T,V} = GenericAffExpr(VariableRef[], T[], T(v))
+# Check if two AffExprs are equal after dropping zeros and disregarding the
+# order. Mostly useful for testing.
+function isequal_canonical(aff::GenericAffExpr{C,V}, other::GenericAffExpr{C,V}) where {C,V}
+    aff_nozeros = dropzeros(aff)
+    other_nozeros = dropzeros(other)
+    # Note: This depends on equality of OrderedDicts ignoring order.
+    # This is the current behavior, but it seems questionable.
+    return isequal(aff_nozeros.terms, other_nozeros.terms) && aff.constant == other.constant
```
blegat: Why not call `isequal(aff_nozeros, other_nozeros)` ?

src/parseexpr.jl
```diff
@@ -22,36 +22,34 @@ function parseIdxSet(arg::Expr)
...
     # It's only safe to mutate the first argument.
-    if c == 0
-        T(ex)
+    if iszero(c)
+        convert(T, ex)
     else
-        x = copy(x)
-        scale!(x.coeffs, c)
-        x.constant *= c
+        x = map_coefficients(x, coef -> c*coef)
```
blegat: This is equivalent but less readable than `x = c * x`

src/affexpr.jl
```diff
@@ -42,48 +102,58 @@ function value(a::GenericAffExpr{T, V}, map::Function) where {T, V}
...
-function Base.push!(aff::GenericAffExpr{C,V}, new_coeff::C, new_var::V) where {C,V}
-    push!(aff.coeffs, new_coeff)
-    push!(aff.vars, new_var)
+function Base.push!(aff::GenericAffExpr{C,V}, new_coef::C, new_var::V) where {C,V}
+    add_or_set!(aff.terms, new_var, new_coef)
     aff
 end
 
-# Add an affine expression to an existing affine expression
+# TODO: Consider renaming. append! is an unusual name to add two expressions.
```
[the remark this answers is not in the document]
blegat: Or `addterm!` and `addterms!` which is like `MOI.addvariable!` and `MOI.addvariables!`

mlubin:
I refactored `push!`, `append!`, and `addtoexpr` into:

```julia
help?> JuMP.add_to_expression!
  add_to_expression!(expression, terms...)

  Updates expression in place to expression + (*)(terms...). This is typically much more efficient
  than expression += (*)(terms...). For example, add_to_expression!(expression, a, b) produces the
  same result as expression += a*b, and add_to_expression!(expression, a) produces the same result as
  expression += a.

  Only a few methods are defined, mostly for internal use, and only for the cases when (1) they can be
  implemented efficiently and (2) expression is capable of storing the result. For example,
  add_to_expression!(::AffExpr, ::VariableRef, ::VariableRef) is not defined because a GenericAffExpr
  cannot store the product of two variables.
```
and
```julia
help?> JuMP.destructive_add!
  destructive_add!(expression, terms...)

  Returns the result of expression + (*)(terms...) and possibly destroys expression (i.e., reusing its
  storage for the result, if possible). For example result = destructive_add!(expression, 2, x) sets
  result to the value of expression + 2*x. expression should not be referred to after this call.

  Note: Discarding the result of destructive_add! suggests an improper usage.
```

Big win for clarity.

I'll merge this after getting the diff coverage sufficiently close to 100%.

src/affexpr.jl
mlubin: This could have performance implications that I don't understand well. I'd want to check that the implementation handles deletion and re-adding efficiently or that benchmarks don't show a regression if we delete zeros.

mlubin:
I'm happy with the coverage now.
@blegat @joehuchette please take a look.

src/affexpr.jl
```diff
@@ -15,23 +15,84 @@
...
+
+
 #############################################################################
 # GenericAffExpr
 # ∑ aᵢ xᵢ  +  c
 mutable struct GenericAffExpr{CoefType,VarType} <: AbstractJuMPScalar
-    vars::Vector{VarType}
-    coeffs::Vector{CoefType}
     constant::CoefType
+    terms::OrderedDict{VarType,CoefType}
```
blegat: Having `terms` as the first field would be more consistent with `MOI.ScalarAffineFunction`

src/affexpr.jl
```diff
@@ -15,23 +15,84 @@
...
+Base.zero(::Type{GenericAffExpr{C,V}}) where {C,V} = GenericAffExpr{C,V}(zero(C), OrderedDict{V,C}())
+Base.one(::Type{GenericAffExpr{C,V}}) where {C,V}  = GenericAffExpr{C,V}(one(C), OrderedDict{V,C}())
 Base.zero(a::GenericAffExpr) = zero(typeof(a))
 Base.one( a::GenericAffExpr) =  one(typeof(a))
-Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.vars),copy(a.coeffs),copy(a.constant))
+Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.constant), copy(a.terms))
+
+GenericAffExpr{C, V}() where {C, V} = zero(GenericAffExpr{C, V})
+
+function map_coefficients!(a::GenericAffExpr, f::Function)
```
blegat: It would be more consistent with `Base.map` to have `f` as a first parameter

src/operators.jl
```diff
@@ -18,63 +18,107 @@
...
+end
+Base.:*(lhs::Number, rhs::GenericQuadExpr) = map_coefficients(rhs, c -> lhs * c)
 
 # AbstractVariableRef (or, AbstractJuMPScalar)
+# TODO: What is the role of AbstractJuMPScalar??
 Base.:+(lhs::AbstractJuMPScalar) = lhs
-Base.:-(lhs::AbstractVariableRef) = GenericAffExpr([lhs],[-1.0],0.0)
-Base.:*(lhs::AbstractJuMPScalar) = lhs # make this more generic so extensions don't have to define unary multiplication for our macros
+Base.:-(lhs::AbstractVariableRef) = GenericAffExpr(0.0, lhs => -1.0)
+Base.:*(lhs::AbstractVariableRef) = lhs # make this more generic so extensions don't have to define unary multiplication for our macros
```
blegat: By replacing `AbstractJuMPScalar` with `AbstractVariableRef`, it won't work anymore for `AffExpr` and `QuadExpr`. This should be covered by the tests :/

src/operators.jl
```diff
@@ -83,120 +127,152 @@ function Base.:^(lhs::Union{AbstractVariableRef,GenericAffExpr}, rhs::Integer)
 end
 Base.:^(lhs::Union{AbstractVariableRef,GenericAffExpr}, rhs::Number) = error("Only exponents of 0, 1, or 2 are currently supported. Are you trying to build a nonlinear problem? Make sure you use @NLconstraint/@NLobjective.")
 # GenericAffExpr--AbstractVariableRef
-Base.:+(lhs::GenericAffExpr{C,V}, rhs::V) where {C,V<:JuMPTypes} = GenericAffExpr{C,V}(vcat(lhs.vars,rhs),vcat(lhs.coeffs,one(C)), lhs.constant)
-Base.:-(lhs::GenericAffExpr{C,V}, rhs::V) where {C,V<:JuMPTypes} = GenericAffExpr{C,V}(vcat(lhs.vars,rhs),vcat(lhs.coeffs,-one(C)),lhs.constant)
-# Don't fall back on AbstractVariableRef*GenericAffExpr to preserve lhs/rhs consistency (appears in printing)
-function Base.:*(lhs::GenericAffExpr{T,V}, rhs::V) where {T,V}
-    n = length(lhs.vars)
+function Base.:+(lhs::GenericAffExpr{C,V}, rhs::V) where {C, V <: AbstractVariableRef}
```
blegat: I don't think we need `V <: AbstractVariableRef` since this is already implied by the fact that `V` is the second parameter of `GenericAffExpr{C,V}`

src/operators.jl
```diff
@@ -83,120 +127,152 @@ function Base.:^(lhs::Union{AbstractVariableRef,GenericAffExpr}, rhs::Integer)
...
 # GenericQuadExpr
 Base.:+(lhs::GenericQuadExpr) = lhs
-Base.:-(lhs::GenericQuadExpr{T}) where T = zero(T)-lhs
+Base.:-(lhs::GenericQuadExpr) = map_coefficients(lhs, -)
 # GenericQuadExpr--Number
 Base.:+(lhs::GenericQuadExpr, rhs::Number) = (+)(+rhs,lhs)
 Base.:-(lhs::GenericQuadExpr, rhs::Number) = (+)(-rhs,lhs)
 Base.:*(lhs::GenericQuadExpr, rhs::Number) = (*)(rhs,lhs)
-Base.:/(lhs::GenericQuadExpr, rhs::Number) = (*)(inv(rhs),lhs)
+Base.:/(lhs::GenericQuadExpr, rhs::Number) = (*)(1.0/rhs,lhs)
```
blegat: Why replacing `inv` ? Isn't it more specific ?

src/operators.jl
```diff
@@ -210,12 +286,12 @@ _sizehint_expr!(q, n) = nothing
 #############################################################################
 
 # TODO: specialize sum for Dict and JuMPArray of JuMP objects?
-Base.sum(j::Array{<:AbstractVariableRef}) = GenericAffExpr(vec(j), ones(length(j)), 0.0)
+Base.sum(vars::Array{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])
```
blegat: Isn't more inference-friendly to use `map` instead of a generator ? I always try to avoid them because `eltype(v => -1.0 for v in vars)` is `Any`

src/affexpr.jl
mlubin: I'm aware of the inconsistency with `MOI.ScalarAffineFunction`, but I like the `AffExpr(constant, pairs...)` constructor, and this ordering is consistent with that.

src/affexpr.jl
mlubin: Good point, done. I renamed `map_coefficients!` to `map_coefficients_inplace!`, because `map!` is not an in-place operation (it only stores the output in a given container).

src/operators.jl
mlubin: Don't remember why I changed this.

src/operators.jl
mlubin: We specify `V <: AbstractVariableRef` in lots of places in the overloads for `AbstractVariableRef` above, why shouldn't it be here?

src/operators.jl
mlubin: Done and added tests.

src/operators.jl
mlubin: I added a test with `@inferred` and this passes.

src/operators.jl
blegat: As a second thought, it's probably better to leave it. We might change the method at some point and it might be needed then so it's safer to be explicit

PR #1302, opened 2018-05-15
Test and fix JuMP extensibility

blegat:
<s>There a still a few tests failing which is why there is WIP in the title.</s>
This PR adds a `JuMPExtension` module showing an example of JuMP extension (@joehuchette this is close this the new design that StructJuMP could have).
It uses the `constraints` tests to test that the JuMPExtension works.
A few fixes are scattered thoughout the code to make the extension possible.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -422,6 +422,7 @@ function registercon(m::AbstractModel, conname::Symbol, value)
 end
 registercon(m::AbstractModel, conname, value) = error("Invalid constraint name $conname")
 
+# This function needs to be implemented by all `AbstractModel`s
 object_dictionary(m::Model) = m.objdict
 
 function registerobject(m::AbstractModel, name::Symbol, value, errorstring::String)
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -412,40 +412,44 @@ include("quadexpr.jl")
 include("sd.jl")
 
 # handle dictionary of variables
-function registervar(m::Model, varname::Symbol, value)
+function registervar(m::AbstractModel, varname::Symbol, value)
     registerobject(m, varname, value, "A variable or constraint named $varname is already attached to this model. If creating variables programmatically, use the anonymous variable syntax x = @variable(m, [1:N], ...).")
 end
-registervar(m::Model, varname, value) = error("Invalid variable name $varname")
+registervar(m::AbstractModel, varname, value) = error("Invalid variable name $varname")
 
-function registercon(m::Model, conname::Symbol, value)
+function registercon(m::AbstractModel, conname::Symbol, value)
     registerobject(m, conname, value, "A variable or constraint named $conname is already attached to this model. If creating constraints programmatically, use the anonymous constraint syntax con = @constraint(m, ...).")
 end
-registercon(m::Model, conname, value) = error("Invalid constraint name $conname")
+registercon(m::AbstractModel, conname, value) = error("Invalid constraint name $conname")
 
-function registerobject(m::Model, name::Symbol, value, errorstring::String)
-    if haskey(m.objdict, name)
+object_dictionary(m::Model) = m.objdict
+
+function registerobject(m::AbstractModel, name::Symbol, value, errorstring::String)
+    objdict = object_dictionary(m)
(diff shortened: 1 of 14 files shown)


src/JuMP.jl
```diff
@@ -412,40 +412,44 @@ include("quadexpr.jl")
...
-function registercon(m::Model, conname::Symbol, value)
+function registercon(m::AbstractModel, conname::Symbol, value)
     registerobject(m, conname, value, "A variable or constraint named $conname is already attached to this model. If creating constraints programmatically, use the anonymous constraint syntax con = @constraint(m, ...).")
 end
-registercon(m::Model, conname, value) = error("Invalid constraint name $conname")
+registercon(m::AbstractModel, conname, value) = error("Invalid constraint name $conname")
 
-function registerobject(m::Model, name::Symbol, value, errorstring::String)
-    if haskey(m.objdict, name)
+object_dictionary(m::Model) = m.objdict
```
mlubin: Add a comment that this function needs to be implemented by all `AbstractModels`

src/macros.jl
```diff
@@ -376,6 +376,8 @@ end
 
 # TODO: update 3-argument @constraint macro to pass through names like @variable
 
+constrainttype(m::Model) = ConstraintRef{typeof(m)}
```
mlubin: Add a comment that this function needs to be implemented by all AbstractModels

src/print.jl
```diff
@@ -110,36 +110,33 @@ end
 #------------------------------------------------------------------------
 ## VariableRef
 #------------------------------------------------------------------------
-Base.show(io::IO, v::VariableRef) = print(io, var_str(REPLMode,v))
-Base.show(io::IO, ::MIME"text/latex", v::VariableRef) =
+Base.show(io::IO, v::AbstractVariableRef) = print(io, var_str(REPLMode,v))
```
mlubin: Why do we want extensions to implement `var_str` instead of `Base.show` for their variable type?

test/constraint.jl
```diff
@@ -1,4 +1,8 @@
-@testset "Constraints" begin
+# The parameter names correspond to the JuMP types so that the test can serve as examples that can be copy pasted
+function constraints_test(Model::Type{<:JuMP.AbstractModel}, VariableRef::Type{<:JuMP.AbstractVariableRef})
+    AffExpr = JuMP.GenericAffExpr{Float64, VariableRef}
```
mlubin: Just use `JuMP.AffExpr` and `JuMP.QuadExpr`?

test/operator.jl
```diff
@@ -21,7 +21,10 @@ Base.transpose(t::MySumType) = MySumType(t.a)
 *(t1::S, t2::MyType{T}) where {S, T} = MyType(t1*t2.a)
 *(t1::MyType{S}, t2::MyType{T}) where {S, T} = MyType(t1.a*t2.a)
 
-@testset "Operator" begin
+function operators_test(Model::Type{<:JuMP.AbstractModel}, VariableRef::Type{<:JuMP.AbstractVariableRef})
+    AffExpr = JuMP.GenericAffExpr{Float64, VariableRef}
```
mlubin: Same comment here

test/constraint.jl
blegat: The names of the function correspond to the name of the JuMP types so that examples can be copy pasted but they may not be. So `VariableRef` may not be equal to `JuMP.VariableRef`.

src/print.jl
blegat: They only need to implement `JuMP.name`

mlubin: Document please

test/constraint.jl
mlubin: That's confusing, use variable names that don't conflict with existing type names, e.g., `ModelType`, `VariableRefType`, `AffExprType`.

src/variables.jl
```diff
@@ -106,7 +106,7 @@ function VariableRef(m::Model)
 end
 
 # Name setter/getters
-
+# These functions need to be implemented by all `AbstractModel`s
```
mlubin: Unclear, does this mean for all `AbstractVariableRef`s?

PR #1305, opened 2018-05-22
Switch to using an MOI release

mlubin:
In preparation for breaking changes in MOI


blegat:
Looking at travis log, we probably need to revert https://github.com/JuliaOpt/JuMP.jl/pull/1270/

mlubin:
Let's hold off on this for a bit until we need to make the switch then.

PR #1309, opened 2018-05-26
Fix operator_warn

blegat:
A bug was introduced in `operator_warn` in https://github.com/JuliaOpt/JuMP.jl/pull/1301 which shows the importance of code coverage.
I have added a test that covers this part of the code and a check that it is actually covers in addition to the trivial bug fix.


mlubin:
Good catch!

Issue #1189, opened 2018-02-07
Keyword arguments not supported by SDconstraints


blegat:
Extra keyword arguments to `@constraint` as passed to `addconstraint` but it is not the case for `@SDconstraint`.
See https://discourse.julialang.org/t/sumofsquares-jump-implementing-semidefinite-constraint-with-specified-domain/8830
cc @rdbrackston

PR #1314, opened 2018-06-02
Share code between constraint and SDconstraint

blegat:
This PR creates a `constraintlike` function (I am open to a change in the name) that contains basically all the code for the `@constraint` macro that can be shared with `@SDconstraint` (i.e. all of it except the parse function).
With this change, the `@constraint` and `@SDconstraint` macros contains a single line calling `constraintlike`.

This therefore adds all the features of `@constraint` that were missing for `@SDconstraint` (e.g. constraint name, group of constraints, support for passing keyword arguments (hence fixing #1189)).

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1189

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -519,9 +519,9 @@ end
 function parseSDconstraint(_error::Function, sense::Symbol, lhs, rhs)
     # Simple comparison - move everything to the LHS
     aff = :()
-    if sense == :⪰ || sense == :(≥)
+    if sense == :⪰ || sense == :(≥) || sense == :(>=)
         aff = :($lhs - $rhs)
-    elseif sense == :⪯ || sense == :(≤)
+    elseif sense == :⪯ || sense == :(≤) || sense == :(<=)
         aff = :($rhs - $lhs)
     else
         _error("Invalid sense $sense in SDP constraint")
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -478,7 +478,7 @@ Add a group of constraints described by the expression `expr` parametrized by
 The expression `expr` can either be
 
 * of the form `func in set` constraining the function `func` to belong to the
-  set `set`, e.g. `@constraint(m, [1, x-1, y-2] in MOI.SecondOrderCone)`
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -465,45 +459,58 @@ macro constraint(args...)
...
+add groups of linear or quadratic constraints.
+
+"""
+macro constraint(args...)
+    # JuMP accepts constraint syntax of the form @constraint(m, foo in bar).
+    # This will be rewritten to a call to buildconstraint(_error,foo, bar). To
+    # extend JuMP to accept set-based constraints of this form, it is necessary
+    # to add the corresponding methods to buildconstraint. Note that this
+    # will likely mean that bar will be some custom type, rather than e.g. a
+    # Symbol, since we will likely want to dispatch on the type of the set
```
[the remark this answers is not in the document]
blegat: I have updated it to also talk about the type of `func` (for PolyJuMP, it is the type of `func` which is used for dispatch, the set is simply `MOI.GreaterThan{Float64}`). One could also have a set different from MOI to be used for dispatch, e.g. the `PSDCone` set.

src/macros.jl
```diff
@@ -355,6 +355,7 @@ end
 #     end
 # end
 
+# TODO replace these with buildconstraint(_error, fun, ::Interval) for more consistency, quad expres in Interval should now be supported with MOI anyway
```
mlubin: s/expres/exprs/

src/macros.jl
```diff
@@ -465,45 +462,98 @@ macro constraint(args...)
...
+
+    @constraint(m::Model, ref[i=..., j=..., ...], expr)
+
+Add a group of constraints described by the expression `expr` parametrized by
+`i`, `j`, ...
+
+The expression `expr` can either be
+
+* of the form `func in set` constraining the function `func` to belong to the
+  set `set`, e.g. `@constraint(m, [1, x-1, y-2] in MOI.SecondOrderCone)`
```
mlubin: Currently this needs to be `MOI.SecondOrderCone(3)`, right?

blegat: Ups, yes indeed !

PR #1315, opened 2018-06-03
Refactor @variable to have a common return

blegat:
This slightly refactor the `@variable` macro so that all three branches have the same `return` statement instead of three different ones. The distinctions between their returned code can be factored out into `loopedcode` and `finalvariable`.
This makes it clearer what are the distinctions between the branches and gets us closed to having a `parsevariable` function.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1113,76 +1113,78 @@ macro variable(args...)
         buildcall = :( buildvariable($_error, $info, $(extra...)) )
         addkwargs!(buildcall, extra_kwargs)
         variablecall = :( addvariable($m, $buildcall, $basename) )
-        code = :($variable = $variablecall)
-        if !anonvar
-            code = quote
-                $code
-                registervar($m, $quotvarname, $variable)
-                $escvarname = $variable
-            end
-        end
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -1113,76 +1113,78 @@ macro variable(args...)
...
-            rng = _rng.args[1] # undo escaping
-            if VERSION >= v"0.7-"
-                (isexpr(rng,:call) && length(rng.args) == 3 && rng.args[1] == :(:) && rng.args[2] == 1) ||
-                    _error("Index sets for SDP variables must be ranges of the form 1:N")
-            else
-                (isexpr(rng,:(:)) && rng.args[1] == 1 && length(rng.args) == 2) ||
-                    _error("Index sets for SDP variables must be ranges of the form 1:N")
+    return assert_validmodel(m, quote
+        $loopedcode
+        $(if anonvar
```
mlubin: It would be clearer to lift this `if` to right before the return statement.

blegat:
I have added a test to test that we return `finalvariable` in both cases (anonymous and non-anonymous).

PR #1318, opened 2018-06-03
Refactor return statement of constraint_macro

blegat:
Mirror the change made in https://github.com/JuliaOpt/JuMP.jl/pull/1315/commits/365d4915cebf507a48d8e556b15c3e854434f71b for the `constraint_macro`

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -446,19 +446,28 @@ function constraint_macro(args, macro_name::Symbol, parsefun::Function)
         $parsecode
         $(refcall) = $constraintcall
     end
+
     # Determine the return type of addconstraint. This is needed for JuMP extensions for which this is different than ConstraintRef
     contype = :( constrainttype($m) )
+    creationcode = getloopedcode(variable, code, condition, idxvars, idxsets, contype, requestedcontainer)
+
+    if anonvar
+        # Anonymous constraint, no need to register it to a name
+        # nor to assign it to a variable in the user scope.
+        # We simply return the constraint reference
+        assignmentcode = variable
+    else
+        # We register the constraint reference to its name and
+        # we assign it to a variable of this name
+        assignmentcode = quote
+            registervar($m, $quotvarname, $variable)
+            $escvarname = $variable
+        end
+    end
 
     return assert_validmodel(m, quote
-        $(getloopedcode(variable, code, condition, idxvars, idxsets, contype, requestedcontainer))
-        $(if anonvar
-            variable
-        else
-            quote
-                registercon($m, $quotvarname, $variable)
-                $escvarname = $variable
-            end
-        end)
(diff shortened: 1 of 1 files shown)


src/macros.jl
```diff
@@ -446,19 +446,28 @@ function constraint_macro(args, macro_name::Symbol, parsefun::Function)
...
+    if anonvar
+        # Anonymous constraint, no need to register it to a name
+        # nor to assign it to a variable in the user scope.
+        # We simply return the constraint reference
+        assignmentcode = variable
+    else
+        # We register the constraint reference to its name and
+        # we assign it to a variable of this name
+        assignmentcode = quote
+            registervar($m, $quotvarname, $variable)
```
mlubin: `registercon`?

src/macros.jl
```diff
@@ -446,19 +446,28 @@ function constraint_macro(args, macro_name::Symbol, parsefun::Function)
         $parsecode
         $(refcall) = $constraintcall
     end
+
     # Determine the return type of addconstraint. This is needed for JuMP extensions for which this is different than ConstraintRef
     contype = :( constrainttype($m) )
+    creationcode = getloopedcode(variable, code, condition, idxvars, idxsets, contype, requestedcontainer)
+
+    if anonvar
+        # Anonymous constraint, no need to register it to a name
```
mlubin: Clarify that "name" here refers to the model-level dictionary and not the string name.

src/macros.jl
```diff
@@ -446,19 +446,28 @@ function constraint_macro(args, macro_name::Symbol, parsefun::Function)
...
+        # nor to assign it to a variable in the user scope.
+        # We simply return the constraint reference
+        assignmentcode = variable
+    else
+        # We register the constraint reference to its name and
+        # we assign it to a variable of this name
```
mlubin: to a variable in the local scope

src/macros.jl
blegat: Wow, weird that it passes the tests

blegat: Ah only the error message changes so it makes sense.

src/macros.jl
```diff
@@ -446,19 +446,28 @@ function constraint_macro(args, macro_name::Symbol, parsefun::Function)
...
     contype = :( constrainttype($m) )
+    creationcode = getloopedcode(variable, code, condition, idxvars, idxsets, contype, requestedcontainer)
+
+    if anonvar
+        # Anonymous constraint, no need to register it in the model-level
+        # dictionary nor to assign it to a variable in the user scope.
+        # We simply return the constraint reference
+        assignmentcode = variable
+    else
+        # We register the constraint reference to its name and
```
mlubin: We register the constraint reference to its name in the model-level dictionary

PR #1322, opened 2018-06-04
Create VariableInfo with keywords

blegat:
This PR creates a `VariableInfo` with the keyword arguments of the `@variable` macro which is then modified by the expression (e.g. `0 <= x <= 1` modifies lower and upper bounds).
This `VariableInfo` may contain expressions in its used during parsing.
The advantage of this approach is that we only have one variable `infoexpr` instead of `haslb`, `lb`, ... hence it will make easier to create `parsevariable` as we will only have to pass one variable.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1008,15 +1008,15 @@ macro variable(args...)
             # ub >= x >= lb
             x.args[4] == :>= || x.args[4] == :≥ || _error("Invalid variable bounds")
             var = x.args[3]
-            setlowerbound(infoexpr, esc_nonconstant(x.args[5]), _error)
-            setupperbound(infoexpr, esc_nonconstant(x.args[1]), _error)
+            setlowerbound_or_error(infoexpr, esc_nonconstant(x.args[5]), _error)
+            setupperbound_or_error(infoexpr, esc_nonconstant(x.args[1]), _error)
         elseif x.args[2] == :<= || x.args[2] == :≤
             # lb <= x <= u
             var = x.args[3]
             (x.args[4] != :<= && x.args[4] != :≤) &&
                 _error("Expected <= operator after variable name.")
-            setlowerbound(infoexpr, esc_nonconstant(x.args[1]), _error)
-            setupperbound(infoexpr, esc_nonconstant(x.args[5]), _error)
+            setlowerbound_or_error(infoexpr, esc_nonconstant(x.args[1]), _error)
+            setupperbound_or_error(infoexpr, esc_nonconstant(x.args[5]), _error)
         else
             _error("Use the form lb <= ... <= ub.")
         end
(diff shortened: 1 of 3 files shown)


src/variables.jl
```diff
@@ -14,10 +14,52 @@ mutable struct VariableInfo{S, T, U, V}
...
-    integer::Bool
+    binary::BT
+    integer::IT
 end
 
+function setlowerbound(info::VariableInfo, lower, _error::Function)
+    info.haslb && _error("Cannot specify variable lowerbound twice")
+    info.haslb = true
+    info.lowerbound = lower
+    # lowerbound might be of different type than info.lowerbound
```
mlubin: This is pretty awkward. Why not use a struct with `Any` members for the processing that happens at macro time, and then turn it into a strongly-typed struct for the processing that happens at runtime?

blegat: Good idea!

src/variables.jl
```diff
@@ -54,11 +51,24 @@ end
 function keywordify(kw::Expr)
     (kw.args[1], esc_nonconstant(kw.args[2]))
 end
-function VariableInfo(; lowerbound=NaN, upperbound=NaN, start=NaN, binary=false, integer=false)
-    VariableInfo(lowerbound !== NaN, lowerbound, upperbound !== NaN, upperbound, false, NaN, start !== NaN, start, binary, integer)
+function VariableInfoExpr(; lowerbound=NaN, upperbound=NaN, start=NaN, binary=false, integer=false)
+    VariableInfoExpr(lowerbound !== NaN, lowerbound, upperbound !== NaN, upperbound, false, NaN, start !== NaN, start, binary, integer)
```
mlubin: Use `!isnan` instead of `!== NaN`.

src/macros.jl
```diff
@@ -1032,23 +1027,24 @@ macro variable(args...)
             var = x.args[2]
             var isa Number && _error("Variable declaration of the form `$(x.args[3]) $(x.args[1]) $var` is not supported. Use `$(x.args[3]) <= $var` instead.")
             @assert length(x.args) == 3
-            lb = esc_nonconstant(x.args[3])
-            haslb = true
-            ub = Inf
+            setlowerbound(infoexpr, esc_nonconstant(x.args[3]), _error)
+            if !infoexpr.hasub
+                infoexpr.upperbound = Inf # Change NaN -> Inf which has same type
```
mlubin: Is this still needed? (Same below)

src/variables.jl
```diff
@@ -18,6 +68,8 @@ mutable struct VariableInfo{S, T, U, V}
     integer::Bool
 end
 
+evalcode(info::VariableInfoExpr) = :(VariableInfo($(info.haslb), $(info.lowerbound), $(info.hasub), $(info.upperbound), $(info.hasfix), $(info.fixedvalue), $(info.hasstart), $(info.start), $(info.binary), $(info.integer)))
```
mlubin: The name `evalcode` suggests to me that it calls `eval`. Maybe `constructor_expr`?

src/variables.jl
```diff
@@ -5,6 +5,56 @@
...
+    upperbound::Any
+    hasfix::Bool
+    fixedvalue::Any
+    hasstart::Bool
+    start::Any
+    binary::Any
+    integer::Any
+end
+
+function setlowerbound(info::VariableInfoExpr, lower, _error::Function)
```
mlubin: `setlowerbound_or_error` might be a clearer name. setters don't usually fail like this.

src/variables.jl
blegat: `isnan(::Expr)` is not defined, I have added a comment

src/macros.jl
blegat: I don't think so, let's remove it

blegat: You meant the comment or the whole three lines ?

blegat: I removed the lines, the `upperbound` value does not matter since `hasub` is false.

src/macros.jl
```diff
@@ -994,34 +994,29 @@ macro variable(args...)
...
         explicit_comparison = true
-        hasub = true
-        haslb = true
         if x.args[2] == :>= || x.args[2] == :≥
             # ub >= x >= lb
             x.args[4] == :>= || x.args[4] == :≥ || _error("Invalid variable bounds")
             var = x.args[3]
-            lb = esc_nonconstant(x.args[5])
-            ub = esc_nonconstant(x.args[1])
+            setlowerbound_or_error(infoexpr, esc_nonconstant(x.args[5]), _error)
```
mlubin: The `parseXXXconstraint` functions take `_error` as the first argument. Any reason not to be consistent?

PR #1326, opened 2018-06-05
Fix mixing vectorized constraints with indices

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1310


test/constraint.jl
```diff
@@ -145,6 +145,18 @@ function constraints_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType
         end
     end
 
+    @testset "Broadcasted constraint with indices" begin
+        m = ModelType()
+        @variable m x[1:2]
+        @constraint m cref1[i=2:4] x .== [i, i+1]
+        @test cref1 isa JuMP.JuMPArray{AbstractArray{ConstraintRefType}}
```
mlubin: Ugh, this is a terrible type to create. The issue is related to https://github.com/JuliaOpt/JuMP.jl/issues/525. At least add a TODO mentioning that we don't like this.

PR #1338, opened 2018-06-10
Update issue templates

### LINKED ISSUE #1337: new setup for issue templates
`.github/ISSUE_TEMPLATE.md` is [deprecated](https://help.github.com/articles/manually-creating-a-single-issue-template-for-your-repository/). Consider switching to https://help.github.com/articles/creating-issue-templates-for-your-repository/.

odow:
<s>Ref</s> Closes #1337 

Just had a play following this instructions at https://help.github.com/articles/about-issue-and-pull-request-templates/
<s>We probably need to make a feature request template as well?</s> Done


mlubin:
Yes, we should have a feature request template as well if we make this change. Do we want to add an actual usable template for the bug reports in addition to the warnings? (I'm not sure it's needed.)

odow:
I've added a template since it is too easy to skim or delete the template text without reading.

mlubin:
Adding extra text could make it even more tempting to skim. I'm hesitant about the checklist. It's a bit patronizing and not 100% applicable. There are rare cases where a minimum working example is too hard to make but we might still want to know that someone is experiencing an issue. There are also not always multiple solvers available for a particular problem class. It would make more sense to formulate these as debugging tips rather than a checklist.

odow:
I agree with all those points. Maybe we should just try this for a while and see how it goes.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,15 @@
+---
+name: Bug report
+about: Help us track down bugs in JuMP
+
+---
+
+Welcome to JuMP! The JuMP developers use the GitHub issue tracker for bug reports and feature requests only. Please read the following before posting a new issue:
```
mlubin: Given the split between bug report and feature request, maybe:

"Welcome to JuMP! Please read the following before posting a new bug report:"

odow:
Deleted.

PR #1344, opened 2018-06-14
Fix operator_warn for extensions

blegat:


src/JuMP.jl
```diff
@@ -467,14 +467,19 @@ function Base.setindex!(m::JuMP.Model, value, name::Symbol)
 end
 
 # usage warnings
+# If extensions do not overwrite this, the warning won't be shown
+function _operator_warn(::AbstractVariableRef) end
```
mlubin: This design needs more thought and documentation if we're going to make it part of the extension API. The starting underscore should be avoided for extension APIs. When exactly is this function called, what should an overwrite look like, etc...

blegat: Done :)

src/JuMP.jl
```diff
@@ -466,15 +466,30 @@ function Base.setindex!(m::JuMP.Model, value, name::Symbol)
     m.objdict[name] = value
 end
 
-# usage warnings
+"""
+    operator_warn_with_variable(vref::AbstractVariableRef)
+
+Everytime two expression are summed not using `desctructive_add!` and one of
```
mlubin: s/expression/expressions/

src/JuMP.jl
```diff
@@ -466,15 +466,30 @@ function Base.setindex!(m::JuMP.Model, value, name::Symbol)
     m.objdict[name] = value
 end
 
-# usage warnings
+"""
+    operator_warn_with_variable(vref::AbstractVariableRef)
```
mlubin: Why not call this on one of the expressions directly instead of one of the variables? It seems unusual to pick out the variable just to dispatch to the right model.

test/operator.jl
```diff
@@ -648,13 +648,19 @@ function operators_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
     end
 
     @testset "operator_warn" begin
-        m = Model()
+        m = ModelType()
         @variable m x[1:51]
-        @test m.operator_counter == 0
+        # JuMPExtension does not have the operator_counter but this test
+        # check that it does not make it fail
```
mlubin: Awkward english. Just say that "JuMPExtension does not have the operator_counter field."?

src/JuMP.jl
```diff
@@ -466,15 +474,28 @@ function Base.setindex!(m::JuMP.Model, value, name::Symbol)
     m.objdict[name] = value
 end
 
-# usage warnings
+"""
+    operator_warn(m::AbstractModel)
+
+Everytime two expressions are summed not using `desctructive_add!` and one of
```
mlubin: Typo: `desctructive_add!`

src/JuMP.jl
```diff
@@ -250,6 +250,14 @@ abstract type AbstractConstraint end
 # In JuMP, used only for VariableRef. Useful primarily for extensions
 abstract type AbstractJuMPScalar end
 
+
+"""
+    model(s::AbstractJuMPScalar)
+
+Return the model owning the scalar `s`.
```
mlubin: How about `owner_model`? `model` looks like it could create a new `Model`.

src/JuMP.jl
```diff
@@ -466,15 +474,28 @@ function Base.setindex!(m::JuMP.Model, value, name::Symbol)
...
 end
 
-# usage warnings
+"""
+    operator_warn(m::AbstractModel)
+
+Everytime two expressions are summed not using `desctructive_add!` and one of
+the two expressions have more than 50 terms, this function is called on the model.
+
+## Notes for extension
```
mlubin: Nit: for extensions

src/JuMP.jl
```diff
@@ -466,15 +474,28 @@ function Base.setindex!(m::JuMP.Model, value, name::Symbol)
...
+
+By default it does nothing so every new model type must implement this
```
mlubin: "it does nothing" -> "this method does nothing"

PR #1345, opened 2018-06-14
Test extensibility of more tests

blegat:


test/JuMPExtension.jl
```diff
@@ -49,10 +49,51 @@ function MOI.delete!(m::MyModel, vref::MyVariableRef)
     delete!(m.varnames, vref.idx)
 end
 MOI.isvalid(m::MyModel, vref::MyVariableRef) = vref.idx in keys(m.variables)
+JuMP.numvar(m::MyModel) = length(m.variables)
```
mlubin: Why do we have a method called `numvar`? This needs to change to `num_variables` or something similar.

src/nlp.jl
```diff
@@ -739,7 +739,6 @@ function hessian_slice(d, ex, x, H, scale, nzcount, recovery_tmp_storage,::Type{
     num_products = size(R,2) # number of hessian-vector products
     num_chunks = div(num_products, CHUNK)
     @assert size(R,1) == length(local_to_global_idx)
-    num_variables = length(x)
```
blegat: This variable was not used

src/nlp.jl
```diff
@@ -296,23 +296,23 @@ function MOI.initialize!(d::NLPEvaluator, requested_features::Vector{Symbol})
         return
     end
 
-    num_variables = numvar(d.m)
+    nvariables = num_variables(d.m)
```
mlubin: How about `num_variables_`. We shouldn't inconsistently use `n` and `num_` to mean the same thing.

PR #1346, opened 2018-06-16
very minor 0.7 fixes

mlubin:
I stopped for the evening after getting to https://docs.julialang.org/en/latest/manual/interfaces/#man-interfaces-broadcasting-1. We have to implement this new broadcasting interface for JuMPArray.

### DIFF
--- a/src/containers.jl
+++ b/src/containers.jl
@@ -6,7 +6,7 @@
 include("JuMPArray.jl")
 
 validarrayindexset(::Base.OneTo) = true
-validarrayindexset(r::UnitRange) = start(r) == 1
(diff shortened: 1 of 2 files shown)


blegat:
@mlubin What's the status of https://github.com/JuliaArrays/AxisArrays.jl/issues/84 ? Would it be easier to use AxisArrays ? Of course, it might take too much time if we want broadcasting to work on JuMPArray for JuMP-dev

mlubin:
I'm not aware of any actual progress on JuliaArrays/AxisArrays.jl#84. However, https://github.com/JuliaArrays/AxisArrays.jl/pull/131 might be useful as an example for implementing the broadcast interface.

PR #1348, opened 2018-06-17
Parametrize JuMP model in optimizer type

blegat:
The optimizer type allows zero overhead on the JuMP side.
The variable type therefore now need to be parametrized on the model type so that its field are concretely typed which induces many changes of this PR.

### Benchmarking results
This use the benchmarking file in `test/perf/backend_overhead.jl`.
Before this change:
```
v0.6
14.526 ns (0 allocations: 0 bytes)
43.865 ns (5 allocations: 96 bytes)
v0.7
19.497 ns (0 allocations: 0 bytes)
35.113 ns (3 allocations: 64 bytes)
```
After this change:
```
v0.6
1.832 ns (0 allocations: 0 bytes)
24.567 ns (3 allocations: 64 bytes)
v0.7
1.824 ns (0 allocations: 0 bytes)
11.284 ns (1 allocation: 32 bytes)
```

Related to https://github.com/JuliaOpt/MathOptInterface.jl/issues/321

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -170,10 +170,10 @@ function aff_str(mode, a::GenericAffExpr{C, V}, show_constant=true) where {C, V}
     end
 end
 # Precompile for faster boot times
-Base.precompile(aff_str, (Type{JuMP.REPLMode}, AffExpr, Bool))
-Base.precompile(aff_str, (Type{JuMP.IJuliaMode}, AffExpr, Bool))
-Base.precompile(aff_str, (Type{JuMP.REPLMode}, AffExpr))
-Base.precompile(aff_str, (Type{JuMP.IJuliaMode}, AffExpr))
+Base.precompile(aff_str, (Type{JuMP.REPLMode}, GenericAffExpr{Float64, VariableRef{Model{NonDirectBackendType}}}, Bool))
+Base.precompile(aff_str, (Type{JuMP.IJuliaMode}, GenericAffExpr{Float64, VariableRef{Model{NonDirectBackendType}}}, Bool))
+Base.precompile(aff_str, (Type{JuMP.REPLMode}, GenericAffExpr{Float64, VariableRef{Model{NonDirectBackendType}}}))
+Base.precompile(aff_str, (Type{JuMP.IJuliaMode}, GenericAffExpr{Float64, VariableRef{Model{NonDirectBackendType}}}))
 
 
 #------------------------------------------------------------------------
--- a/src/print.jl
+++ b/src/print.jl
@@ -130,17 +130,7 @@ function var_str(::Type{IJuliaMode}, v::AbstractVariableRef; mathmode=true)
         return math("noname", mathmode)
     end
 end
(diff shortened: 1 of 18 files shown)


test/print.jl
```diff
@@ -116,10 +116,10 @@ end
 
         v = [x,y,x]
         A = [x y; y x]
-        io_test(REPLMode,   v, "JuMP.VariableRef[x, y, x]")
+        io_test(REPLMode,   v, "JuMP.VariableRef{JuMP.Model{MathOptInterface.Utilities.CachingOptimizer{Union{MathOptInterface.AbstractOptimizer, Void},MathOptInterface.Utilities.UniversalFallback{JuMP.JuMPMOIModel{Float64}}}}}[x, y, x]")
```
mlubin: ``JuMP.VariableRef{JuMP.Model{MathOptInterface.Utilities.CachingOptimizer{Union{MathOptInterface.AbstractOptimizer, Void},MathOptInterface.Utilities.UniversalFallback{JuMP.JuMPMOIModel{Float64}}}}}`` is not ok for user-facing printouts.

blegat: This is fixed

src/print.jl
```diff
@@ -130,7 +130,17 @@ function var_str(::Type{IJuliaMode}, v::AbstractVariableRef; mathmode=true)
         return math("noname", mathmode)
     end
 end
-
+# We want arrays of variables to be printed with `JuMP.VariableRef` as eltype
```
mlubin: Eh, this is another chunk of code that will break between julia versions. Weird printing will also confuse developers when they try to create arrays to store `VariableRef`s.

blegat: Yes, but it improves printing :-P Ok I'll revert this ^^

blegat:
The overhead of having an untyped backend is now 1/3 for both time and space in `@variable(model)`.
With Julia v1.0.2:
```julia
julia> using JuMP

julia> const model = Model()
A JuMP Model

julia> using BenchmarkTools

julia> @btime @variable(model)
  33.454 ns (2 allocations: 48 bytes)
noname
```
By annotating the type of the backend in https://github.com/JuliaOpt/JuMP.jl/blob/master/src/variables.jl#L206:
```julia
julia> using JuMP

julia> const model = Model()
A JuMP Model

julia> using BenchmarkTools

julia> @btime @variable(model)
  18.558 ns (1 allocation: 32 bytes)
noname
```

mlubin:
I was looking into this a bit last night. I couldn't figure out why exactly Julia is allocating (I couldn't make a small example of an allocation when dispatching on an object whose type is unknown). Maybe there's a trick to fix it.

Another option without parameterizing the model is to do the work in batch and call `add_variables` instead of `add_variable` when we want to create 1M variables.

blegat:
The batch idea could help. For the 32 remaining bytes, it is due to the fact that VariableRef is not `isbits` because it holds the models as a field and the model is not isbits

odow:
What is the status of this?

odow:
Closing as stale. There are open PRs exploring alternatives: #2520 and #2610

PR #1351, opened 2018-06-19
RFC: modifications

odow:
I have updated this to the new JuMP syntax.

### TODO
- [ ] docs

<s>We discussed having a proof of concept for JuMP-dev of constraint modification.

Is this the right way to do it? I haven't delved into the new JuMP code at all so have no idea if this is the "correct" way to implement it or where it should live etc.

At the very least, it works!*

*Requires my PRs to LQOI, GLPK and MOI</s>

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -210,29 +210,6 @@ function add_constraint(m::Model, c::AbstractConstraint, name::String="")
 end
 
 """
-    set_rhs(constraint, value)
-
-Set the right-hand side of a constraint. Supported constraints are
- - ScalarAffineFunction -in- LessThan
- - ScalarAffineFunction -in- GreaterThan
- - ScalarAffineFunction -in- EqualTo.
-
-Interval sets are *not* supported.
-
-Note that prior to this step, JuMP will move all variable terms onto the
-left-hand side, and all constant terms onto the right-hand side. For example,
-given a constraint `2x + 1 <= 2`, `JuMP.set_rhs(c, 3)` will create the
-constraint `2x <= 3`, not `2x + 1 <= 3`.
-"""
-function set_rhs(constraint::ConstraintRef{Model, MOICON{F, SetType}},
-                 value) where {F <: MOI.ScalarAffineFunction,
-                               SetType <: Union{MOI.LessThan, MOI.GreaterThan,
-                                                MOI.EqualTo}}
-    MOI.set(constraint.m.moi_backend, MOI.ConstraintSet(), index(constraint),
-            SetType(value))
-end
-
-"""
     set_coefficient(constraint::ConstraintRef, variable::VariableRef, value)
 
 Set the coefficient of `variable` in the constraint `constraint` to `value`.
@@ -242,7 +219,7 @@ same variable. For example, given a constraint `2x + 3x <= 2`,
 `JuMP.set_coefficient(c, x, 4)` will create the constraint `4x <= 2`.
 
 
-```jldoctest
+```jldoctest; setup = :(using JuMP)
 model = Model()
 @variable(model, x)
 @constraint(model, con, 2x + 3x <= 2)
@@ -251,7 +228,7 @@ con
 
 # output
 
-con : 4x <= 2.0
+con : 4 x <= 2.0
 ```
 """
 function set_coefficient(constraint::ConstraintRef{Model, MOICON{F, S}},
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -221,7 +221,7 @@ Interval sets are *not* supported.
 
 Note that prior to this step, JuMP will move all variable terms onto the
 left-hand side, and all constant terms onto the right-hand side. For example,
-given a constraint `2x + 1 <= 2`, `JuMP.set_constant(c, 3)` will create the
+given a constraint `2x + 1 <= 2`, `JuMP.set_rhs(c, 3)` will create the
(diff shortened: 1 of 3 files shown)


examples/basic.jl
```diff
@@ -16,20 +16,34 @@
...
 @variable(m, 0 <= y <= 30)
 
 @objective(m, Max, 5x + 3y)
-@constraint(m, 1x + 5y <= 3.0)
+c = @constraint(m, 1x + 5y <= 3.0)
+
+const MOI = MathOptInterface
+function setrhs(c, rhs)
+    moi = c.m.moibackend
+    MOI.set!(moi, MOI.ConstraintSet(), c.index, MOI.set_type(c.index)(rhs))
```
blegat: It is cleaner to do `index(c)` rather than `c.index`

blegat:
> have no idea if this is the "correct" way to implement it

Looks like the correct way to me :)

> where it should live

I'd say we define a `setconstant` function in JuMP.
* For `ScalarAffineFunction`-in-* and `VectorAffineFunction`-in-*, it modifies the constant of the function
* For *-in-`GreaterThan`, *-in-`LessThan` and *-in-`EqualTo`, it modifies the constant in the set
* For `ScalarAffineFunction`-in-`GreaterThan`, ... there is an ambiguity which needs to be resolved by modifying the constant in the set.

It would probably be simpler if we always stored the constant in the function for `GreaterThan`, .. sets except for `SingleVariable`, that way we would have had the simpler rule
* For `ScalarAffineFunction`-in-* and `VectorAffineFunction`-in-*, it modifies the constant of the function
* For `SingleVariable`-in-`GreaterThan`, `SingleVariable`-in-`LessThan` and `SingleVariable`-in-`EqualTo`, it modifies the constant in the set

By the way if we didn't have `SingleVariable` we wouldn't have needed a field in these sets and could have simply defined `GreaterThanZero`, ...

examples/basic.jl
mlubin: The JuMP call should look at least like:
```
MOI.set!(m, MOI.ConstraintSet(), c, MOI.LessThan(rhs))
```
Users shouldn't need to extract the moibackend or the MOI index (even though it's easy).

mlubin: `MOI.set_type(c.index)(rhs)` is confusing for an example. Just use a concrete type.

odow:
Oops. I managed to drop the old commit instead of reverting it. It was totally superseded by this anyway.

odow:
I need this so I can go around changing right-hand sides and coefficients in the constraint matrix.

I'm not sure what the best approach is. We either provide JuMP-specific functions (as done here), or somehow make the MOI modification layer accessible.

src/constraints.jl
```diff
@@ -208,3 +208,43 @@ function add_constraint(m::Model, c::AbstractConstraint, name::String="")
...
+Set the right-hand side of a constraint. Supported constraints are
+ - ScalarAffineFunction -in- LessThan
+ - ScalarAffineFunction -in- GreaterThan
+ - ScalarAffineFunction -in- EqualTo.
+
+Interval sets are *not* supported.
+
+Note that prior to this step, JuMP will move all variable terms onto the
+left-hand side, and all constant terms onto the right-hand side. For example,
+given a constraint `2x + 1 <= 2`, `JuMP.set_constant(c, 3)` will create the
```
blegat: Is is called `set_constant` or `set_rhs` ?

src/constraints.jl
```diff
@@ -208,3 +208,43 @@ function add_constraint(m::Model, c::AbstractConstraint, name::String="")
...
+constraint `2x <= 3`, not `2x + 1 <= 3`.
```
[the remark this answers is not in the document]
blegat: It seems safer to define `add_constant` which is mathematically defined as adding a constant in the function side but which adds the opposite value in the rhs for scalar constraint.
`set_rhs` is ambiguous for Interval and does not work for vector constraints.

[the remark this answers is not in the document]
odow: Revisiting https://github.com/JuliaOpt/JuMP.jl/issues/470 Maybe the simplest thing is to add an extra variable and then use `JuMP.fix(x, value)`. It is a whole lot cleaner with less issues. In which case, this is just a documentation issue.

The issue with `add_to_rhs` is that it makes sense, but how many people would actually use it?

src/constraints.jl
odow: Oops, I changed back and forth a few times and missed this one.

src/constraints.jl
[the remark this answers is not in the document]
mlubin: If you're not excited about `add_to_rhs` we can drop RHS modifications from this PR. `set_coefficient` looks fine.

src/constraints.jl
```diff
@@ -208,3 +208,57 @@ function add_constraint(m::Model, c::AbstractConstraint, name::String="")
...
+    set_coefficient(constraint::ConstraintRef, variable::VariableRef, value)
+
+Set the coefficient of `variable` in the constraint `constraint` to `value`.
+
+Note that prior to this step, JuMP will aggregate multiple terms containing the
+same variable. For example, given a constraint `2x + 3x <= 2`,
+`JuMP.set_coefficient(c, x, 4)` will create the constraint `4x <= 2`.
+
+
+```jldoctest
```
mlubin: Could you drop `set_coefficient` somewhere in the docs so that the doctest gets run?

PR #1352, opened 2018-06-19
Fix printing with Unitful coefficient

blegat:
This is needed for the printing in https://github.com/JuliaOpt/JuMP.jl/issues/1350 to actually work :-P


mlubin:
Could you add a test for this? It's not immediately clear what the issue was.

PR #1354, opened 2018-06-20
Add variabletype kwarg and document @variable

### LINKED ISSUE #1100: Anonymous scalar variable with positional argument
Currently, it is not possible to use anonymous scalar variable and using positional argument at the same time.
This is an issue for PolyJuMP since you cannot create an anonymous polynomial, e.g.
```julia
@variable m Poly(X)
```
does not work.

What do you think of detecting `_` as an anonymous variable so that
```julia
@variable m _ Poly(X)
```
would be considered as an anonymous variable just like
```julia
@variable m [1:2] Poly(X)
```
creates an vector of anonymous variable.

### LINKED ISSUE #1353: Add a docstring to `@variable` and include in the doc

blegat:
The keyword argument name is `variabletype` and not `type` as suggested in https://github.com/JuliaOpt/JuMP.jl/issues/1100#issuecomment-394090272 because `type` is reserved by Julia.
The examples include an example showing how to construct a `JuMPArray` so it should help with https://github.com/JuliaOpt/JuMP.jl/issues/1303
Closes #1353 
Closes #1100


src/macros.jl
```diff
@@ -1005,34 +1005,151 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
-#   for i in 1:N
-#       for j in 1:N
-#           x[i,j] = x[j,i] = addvariable(m, buildvariable(msg -> error("In @variable(m, x[1:N,1:N], Symmetric, Poly(X)): ", msg), VariableInfo(false, NaN, false, NaN, false, NaN, false, NaN, false, false), Poly(X)), "")
-#       end
-#   end
-#   ```
+"""
+    @variable(m; kwargs...)
+
+Add an anonymous variable described by the keyword arguments `kwargs` that
```
mlubin: This should reference or explain what "anonymous variable" means.

src/macros.jl
```diff
@@ -1005,34 +1005,151 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+belongs to the model `m`. Note that if the `basename` argument is specified
+to `varname`, the variable created is not anonymous but has the name `varname`.
```
mlubin: Confusion: I understand a variable to be anonymous even if it has a `VariableName` property set. I'd say that anonymous means that it's not named in the model dictionary.

src/macros.jl
```diff
@@ -1005,34 +1005,151 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+  `axes`;
+* of the form `[axes]` creating an `Array` or `JuMPArray` of anonymous scalar
+  real variables. Note that if the `basename` argument is specified to
+  `varname`, the variables created are note anonymous but have the names
+  `varname[...]` for each indices `...` of the axes `axes` as in the previous
+  case.
+
+The expression `axes` can either be
+
+* of the form `1:s1,1:s2,...,1:sn` creating an `n`-dimensional `Array` of size
```
mlubin: This misleadingly implies that you can't mix the different axis types.

src/macros.jl
```diff
@@ -1005,34 +1005,151 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+
+The expression `varexpr` can either be
+
+* of the form `varname` creating a scalar real variable of name `varname`;
+* of the form `varname[axes]` creating an `Array` or `JuMPArray` of scalar
+  real variables of names `varname[...]` for each indices `...` of the axes
+  `axes`;
+* of the form `[axes]` creating an `Array` or `JuMPArray` of anonymous scalar
+  real variables. Note that if the `basename` argument is specified to
+  `varname`, the variables created are note anonymous but have the names
```
mlubin: Same comment about the definition of anonymous

src/macros.jl
```diff
@@ -1005,34 +1005,151 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+"""
+    @variable(m; kwargs...)
+
+Add an anonymous variable described by the keyword arguments `kwargs` that
+belongs to the model `m`. Note that if the `basename` argument is specified
+to `varname`, the variable created is not anonymous but has the name `varname`.
+
+    @variable(m, expr, args...; kwargs...)
+
+Add a variable described by the expression `expr`, the positional arguments
```
mlubin: Add a variable to the model `m`

src/macros.jl
blegat: What do you mean ? All the axis have type `Base.OneTo` here, haven't they ?

mlubin: I mean, it's easy to infer from this list that you shouldn't write `[i=1:N, foo_set]`.

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
-#       for j in 1:N
-#           x[i,j] = x[j,i] = addvariable(m, buildvariable(msg -> error("In @variable(m, x[1:N,1:N], Symmetric, Poly(X)): ", msg), VariableInfo(false, NaN, false, NaN, false, NaN, false, NaN, false, false), Poly(X)), "")
-#       end
-#   end
-#   ```
+"""
+    @variable(m; kwargs...)
+
+Add an *anonymous* (see the "Variable naming" section below) variable to the
+model `m` described by the keyword arguments `kwargs` that belongs.
```
odow: Why does the sentence end in "that belongs". It's not clear if the model `m` is described by the keyword arguments of whether the variable is. I suggest:

```
Add an *anonymous* (see the "Variable naming" section below) variable 
described by the keyword arguments `kwargs` to the model `m`.
```

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+
+    @variable(m, expr, args...; kwargs...)
+
+Add a variable to the model `m` described by the expression `expr`, the
+positional arguments `args` and the keyword arguments `kwargs` that belongs.
```
odow: "that belongs" again. Maybe I'm miss-reading this?

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+The expression `expr` can either be (note that in the following the symbol
+`<=` can be used instead of `≤` and the symbol `>=`can be used instead of
+`≥`)
+
+* of the form `varexpr` creating variables described by `varexpr`;
+* of the form `varexpr ≤ ub` (resp. `varexpr ≥ lb`) creating variables described by
+  `varexpr` with upper bounds given by `ub` (resp. lower bounds given by `lb`);
+* of the form `varexpr == value` creating variables described by `varexpr` with
+  fixed values given by `value`;
```
odow: s/`value`;/`value`; or

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+* of the form `lb ≤ varexpr ≤ ub` or `ub ≥ varexpr ≥ lb` creating variables
+  described by `varexpr` with lower bounds given by `lb` and upper bounds given
+  by `ub`.
+
+The expression `varexpr` can either be
+
+* of the form `varname` creating a scalar real variable of name `varname`;
+* of the form `varname[idxs1,idxs2,...,idxn]` creating an `n`-dimensional
+  `Array` or `JuMPArray` of scalar real variables.
```
odow: Nit: Here and throughout. I'm not sure what our style policy on lists is, but it either needs to be full sentences ending in ".", or fragments separated by ";", with the penultimate item having "; and" or "; or", and the final item ending in ".". At the very least, let's not mix styles in the same list.

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+
+## Note for extending the variable macro
+
+The single scalar variable or each scalar variable of the container are created
+using `addvariable(m, buildvariable(_error, info, extra_args...;
+extra_kwargs...))` where
+
+* `m` is the model passed to the `@variable` macro;
+* `_error` is an error function showing the `@variable` call in addition to the
+  error message given as argument;
```
odow: Do we need to detail the arguments to this error function?

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+* `info` is the `VariableInfo` struct containing the information gathered in
+  `expr`, the recognized keyword arguments (except `basename` and
+  `variabletype`) and the recognized positional arguments (except `Symmetric`
+  and `PSD`);
+* `extra_args` are the unrecognized positional arguments of `args` plus the
+  value of the `variabletype` keyword argument if present. The `variabletype`
+  keyword argument allows to pass a positional argument to `buildvariable`
```
odow: Nit: allows who? Perhaps: "keyword argument allows the user to pass a position argument to?

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+  without the need to give a positional argument to `@variable`. This allows to
```
odow: Ditto allows

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+  have a positional argument in the `buildvariable` call when using the
+  anonymous single variable syntax `@variable(m; kwargs...)`;
+* `extra_kwargs` are the unrecognized keyword argument of `kwargs`.
+
+## Examples:
```
odow: Nit: none of the other headings end in ":"

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+
+Each expression `idxsi` can either be
+
+* of the form `idxset` specifying that the `i`th index set of the container
+  is `idxset`.
+* of the form `idxname=idxset` specifying that the `i`th index set of the
+  container is `idxset` and allowing values used in `expr` and keyword
+  arguments to be expressions depending on the `idxname`.
+
+If each `idxset` is of the form `1:s`, the container creating is an `Array`.
```
odow: created instead of creating?

src/macros.jl
```diff
@@ -1005,34 +1005,178 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+
+* `basename`: Sets the base name used to generate variable names. It
+  corresponds to the variable name for scalar variable, otherwise, the
+  variable names are `basename[...]` for each indices `...` of the axes `axes`
+* `lowerbound`: Sets the value of the variable lower bound.
+* `upperbound`: Sets the value of the variable upper bound.
+* `start`: Sets the variable starting value used as initial guess in optimization.
+* `binary`: Sets whether the variable is binary or not.
+* `integer`: Sets whether the variable is integer or not.
+* `variabletype`: See the "Note for extending the variable macro " section below.
```
odow: space after "macro"

src/macros.jl
blegat: What do you mean ? Is it detailed to much or does it need to be more detailed ?

odow: Needs to be more. I don't know how many arguments the function `_error` should have. Should I write `_error() = "Bad thing"` or `_error(msg) = msg` or what?

mlubin:
This is too much for a docstring. Discussions on
- the different types of JuMP containers, and the syntax to create them
- the "name" of a variable and how JuMP has a model-level scope

should just be sections in the manual that we can refer to from the docstring. These discussions apply to almost all macros, not just `@variable`.

chriscoey:
sorry if this was already hashed out, but is there a better name than "variabletype", like say "domain"?

blegat:
I like "domain" !

blegat:
In fact it might be confusing because we already have the domain keyword in PolyJuMP for the `@constraint` macro with a different meaning.

chriscoey:
Hmm true. Could there be a better name than "domain" in the polynomial non-negativity context?

blegat:
I don't really see a better name for domain for polynomial constraints. It might also cause compatibility issues to change it.
We could also consider category, class, kind instead of variabletype. What do you think ?

chriscoey:
I'm ok with keeping variabletype, but I think it would be worth having short aliases for the long keyword names, like lb for lowerbound, ub for upperbound, and vt for variabletype.

blegat:
Good to go ?

docs/src/containers.md
```diff
@@ -0,0 +1,36 @@
...
+JuMP.generatecontainer
+```
+
+Containers in macro
+-------------------
+
+In the [`@variable`](@ref) (resp. [`@constraint`](@ref)) macro, containers of
+variables (resp. constraints) can be created the following syntax
+
+* `name[idxs1,idxs2,...,idxn]` creating an `n`-dimensional container of name
```
mlubin: `index_set_1, index_set_2, ..., index_set_n`

docs/src/containers.md
```diff
@@ -0,0 +1,36 @@
...
+  `name`; or
+* `[idxs1,idxs2,...,idxn]` creating an *anonymous* (see [Names](@ref))
+  `n`-dimensional container.
+
+Each expression `idxsi` can either be
```
mlubin: `index_set_i`

src/macros.jl
```diff
@@ -1005,34 +1005,128 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
-#   ```
-#   x = Matrix{...}(N, N)
-#   for i in 1:N
-#       for j in 1:N
-#           x[i,j] = x[j,i] = addvariable(m, buildvariable(msg -> error("In @variable(m, x[1:N,1:N], Symmetric, Poly(X)): ", msg), VariableInfo(false, NaN, false, NaN, false, NaN, false, NaN, false, false), Poly(X)), "")
-#       end
-#   end
-#   ```
+"""
+    @variable(m; kwargs...)
```
mlubin: Nit: use `model` instead of `m`

src/macros.jl
```diff
@@ -1005,34 +1005,128 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+
+Add an *anonymous* (see [Names](@ref)) variable to the model `m`
```
mlubin: Mention that the anonymous variable is also returned.

src/macros.jl
mlubin: Does the macro call work correctly if you put a semicolon before the keyword arguments? If not, using a semicolon here might be confusing.

src/macros.jl
```diff
@@ -1005,34 +1005,128 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+The recognized positional arguments in `args` are the following:
+
+* `Bin`: Sets the variable to be binary, i.e. either 0 or 1.
+* `Int`: Sets the variable to be integer, i.e. one of ..., -2, -1, 0, 1, 2, ...
+* `Symmetric`: Only available when creating a square matrix of variables, i.e.
+  when `varexpr` is of the form `varname[1:n,1:n]` or `varname[i=1:n,j=1:n]`.
+  It creates a symmetric matrix of variable, that is, it only creates a
+  new variable for `varname[i,j]` with `i ≤ j` and sets `varname[j,i]` to the
+  same variable as `varname[i,j]`.
+* `PSD`: Same as `Symmetric` but also constrains the matrix to be positive
```
mlubin: Nit: "Same as Symmetric" is confusing. You can say that it is both `Symmetric` and constrained to be PSD.

src/macros.jl
```diff
@@ -1005,34 +1005,128 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, lvalue, lsi
...
+The following are equivalent ways of creating a variable `x` of name `x` with
+lowerbound 0:
+```julia
+# Specify everything in `expr`
+@variable(m, x >= 0)
+# Specify the lower bound using a keyword argument
+@variable(m, x, lowerbound=0)
+# Specify everything in `kwargs`
+x = @variable(m, basename="x", lowerbound=0)
+# Without the `@variable` macro
```
mlubin: Maybe put the simple examples that are relevant to most users first (before "Notes for extending ...") and move the examples for "Without the `@variable` macro" under the "Notes for extending ...". This is very advanced and may confuse users just looking for examples of how to use the macro.

mlubin:
`variabletype` is fine with me. :-1: for short and arguably less readable aliases. `lb` and `ub` could be ok, but `vt` is meaningless.

PR #1362, opened 2018-06-27
 Apply fullbridgeoptimizer in Automatic and Manual modes

blegat:


blegat:
Any objection to merge ?

mlubin:
Could we allow a verbose mode for the fullbridgeoptimizer where it prints out what bridges it decided to use? The current code is very magical and may be hard for users to debug.

blegat:
Where would that be printed ? As a warning whenever a constraint is bridged ?

mlubin:
Not a warning, but an option for debugging output like:
```
SplitIntervalBridge used XX times
GeoMeanBridge used XX times
RSOCBridge used XX times
...
```
It could be a keyword argument on the JuMP model.

blegat:
I like the idea it would require changes in MOI.Bridges to be made first. Could we pustpone that improvement to a later PR ?

src/JuMP.jl
```diff
@@ -161,7 +161,8 @@ mutable struct Model <: AbstractModel
             m.moibackend = backend
         else
             @assert mode != Direct
-            m.moibackend = MOIU.CachingOptimizer(MOIU.UniversalFallback(JuMPMOIModel{Float64}()), mode == Automatic ? MOIU.Automatic : MOIU.Manual)
+            cached_optimizer = MOIU.CachingOptimizer(MOIU.UniversalFallback(JuMPMOIModel{Float64}()), mode == Automatic ? MOIU.Automatic : MOIU.Manual)
```
mlubin: Nit: Why is this called `cached_optimizer` instead of `caching_optimizer`?

src/JuMP.jl
```diff
@@ -178,12 +179,16 @@ mutable struct Model <: AbstractModel
     end
 end
 
+# In Automatic and Manual mode, `m.moibackend` is a `LazyBridgeOptimizer` and
+# the `CachingOptimizer` is stored in the `model` field
+caching_optimizer(m::Model) = m.moibackend.model
```
mlubin: This method should assert Automatic or Manual mode. It's invalid and potentially dangerous in direct mode.

mlubin:
We at least need a debugging option to disable the bridges then.

src/JuMP.jl
```diff
@@ -142,48 +142,68 @@ mutable struct Model <: AbstractModel
     # dictionary keyed on an extension-specific symbol
     ext::Dict{Symbol,Any}
     # Default constructor
-    function Model(; mode::ModelMode=Automatic, backend=nothing, optimizer=nothing)
-        m = new()
+    function Model(; mode::ModelMode=Automatic, backend=nothing, optimizer=nothing, bridge=true)
```
mlubin: `bridge_constraints` is a clearer name

PR #1366, opened 2018-06-29
Fix helpful error

### LINKED ISSUE #1365: bad error message in variable
```julia
julia> m = Model()
A JuMP Model

julia> @variable(m, 0 <= x)
ERROR: UndefVarError: S not defined
```

blegat:
Closes #1365

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -974,7 +974,7 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, sense::Symb
     # an helpful error in the the case were we can easily determine that the
     # user placed the variable in the rhs, i.e. the case where the rhs is a
     # constant number.
-    var isa Number && _error("Variable declaration of the form `$var $S $value` is not supported. Use `$value $(reverse_sense(sense)) $var` instead.")
+    var isa Number && _error("Variable declaration of the form `$var $sense $value` is not supported. Use `$value $(reverse_sense(Val(sense))) $var` instead.")
     parse_one_operator_variable(_error, infoexpr, Val(sense), esc_nonconstant(value))
     var
 end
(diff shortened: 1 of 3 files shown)


odow:
Should we throw a proper exception type to distinguish the helpful error message from the un-helpful "S not defined"? Because this wasn't caught in tests as it still throws an error.

mlubin:
Can't we just make this work? 
```julia
if var isa Number && !(value isa Number)
     parse_one_operator_variable(_error, infoexpr, Val(reverse_sense(sense)), esc_nonconstant(var))
    return value
end
```

blegat:
I don't understand why it wasn't called since we check that its an `ErrorException` and `S not defined` should be an `UndefVarError`.

odow:
The question is whether it is more or less confusing for users. Then you have to say why
```julia
@variable(m, 0 <= x)
```
works but
```julia
a = 0
@variable(m, a <= x)
```
I can see it leading to more confusion than "it never works"

blegat:
> Can't we just make this work?

We should at least throw a warning because it's going to fail badly when it's not constant.
Or we add in the FAQ the meaning of the weird error that is got. I am fine with both.

mlubin:
The confusion always seems to be with
```julia
@variable(m, 0 <= x)
```
I've yet to see a complaint about
```julia
a = 0
@variable(m, a <= x)
```

odow:
Okay lets change it.

blegat:
I have changed it. It now works instead of erroring :)

src/macros.jl
```diff
@@ -965,18 +965,22 @@ function parsevariable(_error::Function, infoexpr::VariableInfoExpr, sense::Symb
     # Variable declaration of the form: var sense value
```
mlubin: This comment is out of date, so are the argument names.

blegat: I explains where the function arguments come from. We should probably rename to lhs rhs though

blegat: I have refactored it in two methods

test/variable.jl
```diff
@@ -34,26 +34,35 @@ function variables_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
...
-            @test !JuMP.haslowerbound(ubonly)
-            @test JuMP.hasupperbound(ubonly)
-            @test JuMP.upperbound(ubonly) == 1.0
-            @test !JuMP.isfixed(ubonly)
-            @test !JuMP.isbinary(ubonly)
-            @test JuMP.isinteger(ubonly)
-            @test isequal(mcon[:ubonly],ubonly)
+            @variable(mcon, ubonly1 <= 1, Int)
+            @test isequal(mcon[:ubonly1],ubonly1)
+            @variable(mcon, 1 >= ubonly2, Int)
```
mlubin: Make a separate test for this.

PR #1367, opened 2018-06-29
[release-0.18] update for Julia v0.7

joehuchette:
This drops support for v0.6 for the moment

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,6 +3,4 @@ MathProgBase 0.6 0.8
 ReverseDiffSparse 0.8 0.9
 ForwardDiff 0.5 0.8
 Calculus
-LinearAlgebra
-SparseArrays
-Printf
+Compat 0.49
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,8 @@
-julia 0.6
+julia 0.7
 MathProgBase 0.6 0.8
 ReverseDiffSparse 0.8 0.9
 ForwardDiff 0.5 0.8
 Calculus
+LinearAlgebra
+SparseArrays
+Printf
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.7
+julia 0.6
 MathProgBase 0.6 0.8
 ReverseDiffSparse 0.8 0.9
 ForwardDiff 0.5 0.8
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,8 @@ environment:
   matrix:
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-beta-win32.exe"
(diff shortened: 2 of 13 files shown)


REQUIRE
```diff
@@ -1,5 +1,8 @@
-julia 0.6
+julia 0.7
 MathProgBase 0.6 0.8
 ReverseDiffSparse 0.8 0.9
 ForwardDiff 0.5 0.8
 Calculus
+LinearAlgebra
```
mlubin: I don't think these are needed, they're baked into Julia.

mlubin:
You can just put `Compat` in front of the packages to make this work on 0.6.

REQUIRE
joehuchette: I think you're right

blegat: You need to do `using Compat.LinearAlgebra` for it to work on both v0.6 and v0.7

src/JuMPContainer.jl
```diff
@@ -211,9 +211,9 @@ mutable struct KeyIterator{JA<:JuMPArray}
     x::JA
     dim::Int
     next_k_cache::Array{Any,1}
-    function (::Type{KeyIterator{JA}}){JA}(d)
+    function (::Type{KeyIterator{JA}})(d) where JA
```
blegat: It should just be `function KeyIterator{JA}(d) where JA` if I'm not mistaken

joehuchette:
I believe it's working on both v0.6 and v0.7 now

src/solvers.jl
```diff
@@ -250,6 +250,8 @@ function solve(m::Model; suppress_warnings=false,
             # Do a separate try since getobjval could work while getobjbound does not and vice versa
             objBound = MathProgBase.getobjbound(m.internalModel) + m.obj.aff.constant
             m.objBound = objBound
+        catch
+            nothing
```
mlubin: Why is this needed?

joehuchette: On v0.7 each try block needs either a ``catch`` or a ``finally`` condition

mlubin:
Could you switch nightly to 0.7 on travis/appveyor and remove it as an allowed failure? There are still some errors.

appveyor.yml
```diff
@@ -2,6 +2,8 @@ environment:
   matrix:
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-beta-win32.exe"
```
mlubin: Is there a `0.7-latest`?

mlubin:
I don't see any regressions on 0.6, so we could merge and continue 0.7 work in separate PRs.

joehuchette:
Feel free to merge if you don't want updating to block on this PR; I won't get the chance to return to this for a couple days.

ufechner7:
Would be nice if this could be merged even if it not the final fix for 0.7 yet. I think we need an iterative approach to make the packages work.

mlubin:
I agree, but I won't merge until I get a green light from travis on 0.6. This is blocked by (at least) https://github.com/JuliaLang/METADATA.jl/pull/15933.

IainNZ:
METADATA PR in, can re-run CI?

odow:
Not sure how to restart Appveyor? Do I need some extra permissions or something?

blegat:
@odow I have restarted 30 minutes ago

PR #1369, opened 2018-07-08
WIP: style guide

odow:
Refer to the [notes from our discussion in Bordeaux](https://docs.google.com/document/d/1knl5tOg4WW6xqy2TRgAcWoy9lb4fqggm-qh2XEk8q9Q/edit?usp=sharing).

Edit: @mohamed82008 asked (https://github.com/JuliaOpt/MathOptInterface.jl/pull/420#issuecomment-403357788) about the reasoning behind the removal of ! in function names. I thought I would outline our reasoning here.

First, we are aware that this goes against the Julia style guide. In `Base`, bangs are used to denote functions which modify their arguments. However, it is not apparent which arguments are modified. In most cases, it is the first, but not always. How do you distinguish between functions which modify some of their arguments but not all? Since the docstring will always have to mention which arguments are modified, we made the decision to remove bangs and force the user to read the documentation. (There were also some members pushing for this change due to aesthetic reasons.)

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -4,8 +4,43 @@ Style guide and design principles
 Style guide
 -----------
 
-TODO: A style guide for JuMP, JuMP models and surrounding Julia code.
-Formatting, naming, use of macros, comments, TODOs, docstrings, etc.
+### JuMP Source Code
+
+ - 80 character line limit. (Except for long URLs in comments.)
+ - 4 spaces for indentation.
+- Use `lowercase_with_underscore` for variable and function names.
+- Types and Modules should be in `CamelCase`.
+- Use `UPPERCASE` for constants.
+ - No one-letter variable names. This includes using `m = Model()`. The only
+exceptions are for indices in loops such as `for i in 1:N`.
+ - Start internal functions (i.e., those not meant to be used by a user) with an
+ underscore, e.g., `_add_constraint`.
+ - Use three backticks for name of internal function docstrings.
+ - Variables should never start with an underscore.
+ - To instantiate a vector, use `T[]` instead of `Vector{T}()`.
+ - Use proper punctuation and full sentences in comments.
+ - Use `for i in 1:N` instead of `for i = 1:N`.
+ - Never use trailing periods for floating point, e.g., `0.`.
+
+### Writing a JuMP Model
+
+These style points address the convention for how users should write JuMP
+models.
+
+ - Never use `m` for the name of a model.
+ - If the model fits on one screen, short variable names are okay (e.g.,
+`@variable(model, x)`).
+ - If the model is to large to fit on one screen, its construction should be
+ decomposed into functions with long variable names.
+
+#### Inside JuMP model macros
+
+ - Never use `1.0` when `1` is okay.
+ - Use the order `coefficient * variable` instead of `variable * coefficient`.
+ - Use parenthesis in macros
+ - Prefer blocks of `@variables` over lots of `@variable` lines. (The same goes
+for the other plural versions of the macros.)
+
 
 Design principles
 -----------------
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -1,16 +1,25 @@
 Style guide and design principles
 =================================
 
+This document outlines JuMP's style guide. It also contains a set or principles
+for users writing JuMP models.
+
 Style guide
 -----------
+
+!!! info
+    For developers: if you modify JuMP's source code, please fix the style of
+    the surrounding code while you are there.
 
 ### JuMP Source Code
 
  - 80 character line limit. (Except for long URLs in comments.)
  - 4 spaces for indentation.
-- Use `lowercase_with_underscore` for variable and function names.
-- Types and Modules should be in `CamelCase`.
-- Use `UPPERCASE` for constants.
+ - Use `lowercase_with_underscore` for variable and function names.
+ - Types and Modules should be in `CamelCase`.
+ - Use `UPPERCASE` for constants.
+ - No bangs (!) in function names. Use docstrings to indicate what is modified
+and variable lifetime issues.
  - No one-letter variable names. This includes using `m = Model()`. The only
 exceptions are for indices in loops such as `for i in 1:N`.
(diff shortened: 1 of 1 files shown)


docs/src/style.md
```diff
@@ -4,8 +4,43 @@ Style guide and design principles
 Style guide
 -----------
 
-TODO: A style guide for JuMP, JuMP models and surrounding Julia code.
-Formatting, naming, use of macros, comments, TODOs, docstrings, etc.
+### JuMP Source Code
+
+ - 80 character line limit. (Except for long URLs in comments.)
+ - 4 spaces for indentation.
+- Use `lowercase_with_underscore` for variable and function names.
```
blegat: Missing a starting space for this item and the two following

docs/src/style.md
```diff
@@ -1,15 +1,78 @@
...
+ - Start internal functions (i.e., those not meant to be used by a user) with an
+ underscore, e.g., `_add_constraint`.
+ - Use three backticks for name of internal function docstrings.
+ - Variables should never start with an underscore.
+ - To instantiate a vector, use `T[]` instead of `Vector{T}()`.
+ - Use proper punctuation and full sentences in comments.
+ - Use `for i in 1:N` instead of `for i = 1:N`.
+ - Never use trailing periods for floating point, e.g., `0.`.
+ - Never more than one consecutive blank line inside a function.
+ - Always use return from a function unless it is a one-line function.
```
blegat: return -> `return`

IssamT:
I am fine with not using bangs. But we need some naming convention for functions that do exactly the same thing, except that one modifies the 'main' object of the function while the other one returns the new object that has the modification. No matter how detailed the docstrings are, those two fonctions still need to have different names.

docs/src/style.md
```diff
@@ -1,15 +1,78 @@
...
+ - Use `lowercase_with_underscore` for variable and function names.
+ - Types and Modules should be in `CamelCase`.
+ - Use `UPPERCASE` for constants.
+ - No bangs (!) in function names. Use docstrings to indicate what is modified
+and variable lifetime issues.
+ - No one-letter variable names. This includes using `m = Model()`. The only
+exceptions are for indices in loops such as `for i in 1:N`.
+ - Start internal functions (i.e., those not meant to be used by a user) with an
+ underscore, e.g., `_add_constraint`.
+ - Use three backticks for name of internal function docstrings.
```
IainNZ: I find this sentence really hard to parse, or know what its referring to.

mlubin:
All of the style guide points should include a justification. (People don't like being told what to do without knowing why.) Some points might also need a clarification as to what they're talking about, e.g., as @IainNZ commented. For example, see this discussion about generators in python: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#27-comprehensions--generator-expressions. 

It might be easier to set up the structure for the style guide and then move the points over from the doc in smaller PRs.

docs/src/style.md
```diff
@@ -1,15 +1,78 @@
...
 
-TODO: A style guide for JuMP, JuMP models and surrounding Julia code.
-Formatting, naming, use of macros, comments, TODOs, docstrings, etc.
+!!! info
+    For developers: if you modify JuMP's source code, please fix the style of
+    the surrounding code while you are there.
+
+### JuMP Source Code
+
+ - 80 character line limit. (Except for long URLs in comments.)
```
chriscoey: I'm REALLY not a fan of character line limits. most modern code editors automatically wrap at an appropriate length given the specific user's display size. I think line limits paradoxically make code harder to read, and they make it more work to update (eg if you change a function name or a comment you often have to reformat the remaining lines to even the lengths out).

docs/src/style.md
```diff
@@ -1,15 +1,78 @@
...
+ underscore, e.g., `_add_constraint`.
+ - Use three backticks for name of internal function docstrings.
+ - Variables should never start with an underscore.
+ - To instantiate a vector, use `T[]` instead of `Vector{T}()`.
+ - Use proper punctuation and full sentences in comments.
+ - Use `for i in 1:N` instead of `for i = 1:N`.
+ - Never use trailing periods for floating point, e.g., `0.`.
+ - Never more than one consecutive blank line inside a function.
+ - Always use return from a function unless it is a one-line function.
+ - Do not split a one-line function over lines. Only if less than 80 characters.
```
chriscoey: see my comment above - I want to remove the 80 characters limit

docs/src/style.md
```diff
@@ -1,15 +1,78 @@
...
+ - Type arguments to user-facing functions so that users do not get internal
+errors when passing wrong types.
+ - Imports: use `import` instead of `using`. Importing specific things from a
+ module is fine, stdlib imports first, alphabetical order after that.
```
rdeits: `using Module: foo` is strictly safer than `import Module: foo`, as it doesn't allow `Module.foo` to be overloaded. Do you really want to prohibit all usages of `using Module: foo`?

docs/src/style.md
```diff
@@ -1,15 +1,78 @@
...
+    the surrounding code while you are there.
+
+### JuMP Source Code
+
+ - 80 character line limit. (Except for long URLs in comments.)
+ - 4 spaces for indentation.
+ - Use `lowercase_with_underscore` for variable and function names.
+ - Types and Modules should be in `CamelCase`.
+ - Use `UPPERCASE` for constants.
+ - No bangs (!) in function names. Use docstrings to indicate what is modified
```
rdeits: I definitely disagree with this. This seems unnecessarily confusing and differs from essentially every single Julia package, and Julia itself. Do we really want to be the one package that disobeys an established convention forever?

docs/src/style.md
chriscoey: I do however, like [semantic linebreaking](http://rhodesmill.org/brandon/2012/one-sentence-per-line/). it is particularly convenient with git diffs too.

could we recommend semantic linebreaking instead of a hard 80 char rule?

mlubin:
I'm closing this in favor of #1371 and subsequent PRs where we'll have a better space to discuss the more controversial points.

PR #1370, opened 2018-07-10
added example adapted from MOSEK manual

ulfworsoe:
NOTE: This only work with MOSEK 9.0 (since it requires power cones)

### DIFF
--- a/examples/lowner_john.jl
+++ b/examples/lowner_john.jl
@@ -220,8 +220,8 @@ if Pkg.installed("Plots") != nothing && Pkg.installed("GR") != nothing
                 t -> Poinv[2,:]' * ( [cos(t),sin(t)] + c ),
                 0,2*pi)
     
-    println("Plot done.")
-    sleep(20)
+    println("Plot done. Press enter to continue.")
+    readline()
 else
     println("Install Plots package to plot solution")    
 end
--- a/examples/lowner_john.jl
+++ b/examples/lowner_john.jl
@@ -1,0 +1,227 @@
+# 
+# Löwner-John ellipsoid example
+#
+# Compute the inner and outer ellipsoids of a polytope.
+#
+# Adapter from example in the MOSEK distro.
+#
+# The inner ellipsoidal approximation to a polytope
+#
+#    S = { x \in R^n | Ax < b }.
+#
+# maximizes the volume of the inscribed ellipsoid,
+#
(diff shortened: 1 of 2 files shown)


examples/lowner_john.jl
```diff
@@ -0,0 +1,227 @@
...
+    @variable(model, C[1:n,1:n], PSD)
+    @variable(model, d[1:n])
+
+    # [ b-A*d ; A*C ] in QCone
+
+    for i in 1:m
+        @constraint(model, [ b[i] - (A[i,:]' * d) ; (A[i,:]'*C)'] in MOI.SecondOrderCone(n+1))
+    end
+
+    det_root(model,C,t)
```
blegat: It should work with the LogDetConeTriangle once https://github.com/JuliaOpt/JuMP.jl/pull/1362 is merged

ulfworsoe: That sounds great! It would simplify it a bit. 

Also, it hardcodes the Mosek solver and uses `mode=JuMP.Direct`, which I'm not sure is so pretty. Is there a way to let JuMP choose the solver based on the model type?

blegat: No, automatically choosing the solver was removed during the update to Julia v0.6.
See https://github.com/JuliaOpt/MathProgBase.jl/pull/151
The new approach is to let the user choose the solver and then try to automatically bridge the constraints to match the types of constraints supported by the solvers instead of automatically selecting a solver that supports the types of the constraints created by the user.

ulfworsoe: Probably a more solid approach. Is there any consensus on what examples should do?

blegat: I don't think so. Currently the examples pick a solver.
What I usually do with Jupyter is creating different cells. Each cell sets `optimizer` to a different value and the user can run the cell he prefers in order to choose the solver.
Cc @joaquimg

mlubin:
@blegat @ulfworsoe Any plans to tidy this up for merging?

blegat:
I felt it was blocked by https://github.com/JuliaOpt/MathOptInterface.jl/issues/406 
The idea is to show how to the choice of solver and root-det vs log-det affects how the problem is reformulated by the bridges.
@ulfworsoe what do you think ?

odow:
Close as stale? There is now better support for integrating examples into the documentation, but we can't use Mosek.

ulfworsoe:
I'll close it. I may revisit this in the coming year.

PR #1383, opened 2018-07-28
Clean up JuMP.Model.

IainNZ:
Remove unused fields, style guide, initial attempts to document from
an outsiders perspective - very much WIP.

Also an attempt to start matching TODOs with issues/people.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -88,97 +88,83 @@ const MOIBIN = MOICON{MOI.SingleVariable,MOI.ZeroOne}
 @enum ModelMode Automatic Manual Direct
 
 abstract type AbstractModel end
+# All `AbstractModels` must define `num_variables`.
+
+"""
+    Model
+
+A mathematical model of an optimization problem.
+"""
 mutable struct Model <: AbstractModel
 
-    # special variablewise properties that we keep track of:
+    # Special variablewise properties that we keep track of:
     # lower bound, upper bound, fixed, integrality, binary
-    variabletolowerbound::Dict{MOIVAR,MOILB}
-    variabletoupperbound::Dict{MOIVAR,MOIUB}
-    variabletofix::Dict{MOIVAR,MOIFIX}
-    variabletointegrality::Dict{MOIVAR,MOIINT}
-    variabletozeroone::Dict{MOIVAR,MOIBIN}
+    variabletolowerbound::Dict{MOIVAR, MOILB}
+    variabletoupperbound::Dict{MOIVAR, MOIUB}
+    variabletofix::Dict{MOIVAR, MOIFIX}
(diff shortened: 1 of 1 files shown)


IainNZ:
Need to rebase on top of #1362

IainNZ:
Rebase done.

src/JuMP.jl
```diff
@@ -77,108 +77,104 @@ const MOIBIN = MOICON{MOI.SingleVariable,MOI.ZeroOne}
 
 @MOIU.model JuMPMOIModel (ZeroOne, Integer) (EqualTo, GreaterThan, LessThan, Interval) (Zeros, Nonnegatives, Nonpositives, SecondOrderCone, RotatedSecondOrderCone, GeometricMeanCone, PositiveSemidefiniteConeTriangle, PositiveSemidefiniteConeSquare, RootDetConeTriangle, RootDetConeSquare, LogDetConeTriangle, LogDetConeSquare) () (SingleVariable,) (ScalarAffineFunction,ScalarQuadraticFunction) (VectorOfVariables,) (VectorAffineFunction,)
 
-###############################################################################
-# Model
+################################################################################
+# Model.
+# TODO(IainNZ): Clarify the three keyword arguments. `mode` is explained,
```
mlubin: This is being refactored in https://github.com/JuliaOpt/JuMP.jl/pull/1384

src/JuMP.jl
```diff
@@ -210,38 +211,44 @@ function mode(model::Model)
...
 
 """
-    objectivevalue(m::Model)
+    objectivevalue(model::Model)
 
-Return the objective value after a call to `solve`.
+Return the objective value after a call to `optimize(model)`.
 """
-objectivevalue(m::Model) = MOI.get(m, MOI.ObjectiveValue())
+objectivevalue(model::Model) = MOI.get(model, MOI.ObjectiveValue())
```
mlubin: We haven't yet moved over the naming discussion to the style guide, but what do you think about `objective_value` etc for these?

IainNZ: I'm a fan of underscores (another difference from Base, but w/e...), so sgtm

src/JuMP.jl
IainNZ: Should hold off merging this for now or quickly get it in and build from it?

src/JuMP.jl
mlubin: Let's hold off on renaming until we resolve the style question, otherwise it might just be unnecessary churn.

src/JuMP.jl
mlubin: Could you revert the docstring for `Model` and the TODO right here that's going to be out of date? Everything else seems fine.

IainNZ:
Done

PR #1384, opened 2018-07-28
Factory

blegat:
New constructor interface as discussed [here](https://docs.google.com/document/d/1knl5tOg4WW6xqy2TRgAcWoy9lb4fqggm-qh2XEk8q9Q/).
Factories allow to create new optimizers on demand which would allow implementing copy (https://github.com/JuliaOpt/JuMP.jl/issues/1381, https://github.com/JuliaOpt/JuMP.jl/issues/1300) without the need for any change in MOI like https://github.com/JuliaOpt/MathOptInterface.jl/pull/387.

### DIFF
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -38,11 +38,11 @@ Direct.
 In Automatic and Manual modes, two MOI layers are automatically applied to the
 optimizer:
 
-* `CachingOptimizer`: it maintain a cache of the model so that when the
-  the optimizer does not support an incremental change to the model, the
-  optimizer's internal model can be discarded and restored from the cache just
-  before optimization. The `CachingOptimizer` has two different modes: Automatic
-  and Manual corresponding to the two JuMP modes with the same names.
+* `CachingOptimizer`: maintains a cache of the model so that when the optimizer
+  does not support an incremental change to the model, the optimizer's internal
+  model can be discarded and restored from the cache just before optimization.
+  The `CachingOptimizer` has two different modes: Automatic and Manual
+  corresponding to the two JuMP modes with the same names.
 * `LazyBridgeOptimizer`: when a constraint added is not supported by the
   optimizer, it tries transform the constraint into an equivalent form,
   possibly adding new variables and constraints that are supported by the
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -68,7 +68,8 @@ JuMP.setoptimizer
 
 New JuMP models are created using the [`Model`](@ref) constructor:
 ```@docs
-Model
+Model()
+Model(::JuMP.Factory)
 ```
 
 ## Direct mode
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -56,16 +56,16 @@ optimizer:
 See the [MOI documentation](http://www.juliaopt.org/MathOptInterface.jl/stable/)
 for more details on these two MOI layers.
 
-To create a fresh new JuMP model, JuMP needs to create a new empty optimizer
-instance. New optimizer instances can be obtained using an
+To attach an optimizer to a JuMP model, JuMP needs to create a new empty
+optimizer instance. New optimizer instances can be obtained using an
 `OptimizerFactory` that can be created using the [`with_optimizer`](@ref)
 function:
 ```@docs
 with_optimizer
 ```
 
-The optimizer factory can be set to the JuMP model in the
-[`JuMP.optimize`](@ref) function:
+The factory can be provided either at model construction time or at
+[`JuMP.optimize`](@ref) time:
 ```@docs
 JuMP.optimize
 ```
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -2,10 +2,11 @@ Interacting with solvers
 ========================
 
 A JuMP model keeps a [MathOptInterface (MOI)](https://github.com/JuliaOpt/MathOptInterface.jl)
-backend internally that stores the optimization problem and acts as the
-optimization solver (the backend can also not support optimization, e.g. it can
-simply store the model in a file). JuMP can be viewed as a lightweight
-user-friendly layer on top of the MOI backend:
+*backend* internally that stores the optimization problem and acts as the
+optimization solver. We call it an MOI *backend* and not optimizer as it can
+also be a wrapper around an optimization file format such as MPS that writes
+the JuMP model in a file. JuMP can be viewed as a lightweight user-friendly
+layer on top of the MOI backend:
 
 * JuMP does not maintain any copy of the model outside this MOI backend.
 * JuMP variable (resp. constraint) references are simple structures containing
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -1,8 +1,7 @@
 Interacting with solvers
 ========================
 
-A JuMP model keeps a
-[MathOptInterface (MOI)](https://github.com/JuliaOpt/MathOptInterface.jl)
+A JuMP model keeps a [MathOptInterface (MOI)](https://github.com/JuliaOpt/MathOptInterface.jl)
 backend internally that stores the optimization problem and act as the
 optimization solver (the backend can also not support optimization, e.g. it can
 simply store the model in a file). JuMP can be viewed as a lightweight
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -2,11 +2,11 @@ Interacting with solvers
 ========================
 
 A JuMP model keeps a [MathOptInterface (MOI)](https://github.com/JuliaOpt/MathOptInterface.jl)
-*backend* internally that stores the optimization problem and acts as the
-optimization solver. We call it an MOI *backend* and not optimizer as it can
-also be a wrapper around an optimization file format such as MPS that writes
-the JuMP model in a file. JuMP can be viewed as a lightweight user-friendly
-layer on top of the MOI backend:
+*backend* of type `MOI.ModelLike` internally that stores the optimization
+problem and acts as the optimization solver. We call it an MOI *backend* and
+not optimizer as it can also be a wrapper around an optimization file format
+such as MPS that writes the JuMP model in a file. JuMP can be viewed as a
+lightweight user-friendly layer on top of the MOI backend:
 
 * JuMP does not maintain any copy of the model outside this MOI backend.
 * JuMP variable (resp. constraint) references are simple structures containing
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -1,6 +1,28 @@
 Interacting with solvers
 ========================
 
+
+```@docs
+with_optimizer
+```
+
+The solvers can be set with
+```@docs
+JuMP.setoptimizer
+```
+
+```@docs
+Model
+```
+
+## Direct mode
+
+For advanced users, the model can be created without using a caching nor a
+bridge optimizer using the [`JuMP.direct_model`](@ref) function:
+```@docs
+JuMP.direct_model
+```
+
 TODO: Describe the connection between JuMP and solvers. Automatic vs. Manual
 mode. CachingOptimizer. How to set/change solvers. How to set parameters (solver
 specific and generic). Status codes. Accessing the result.
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -1,10 +1,12 @@
 Interacting with solvers
 ========================
 
-A JuMP model stores a
+A JuMP model keeps a
 [MathOptInterface (MOI)](https://github.com/JuliaOpt/MathOptInterface.jl)
-backend internally that contains the optimization solver. The JuMP layer on top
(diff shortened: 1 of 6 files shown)


test/nlp_solver.jl
```diff
@@ -40,7 +38,7 @@ new_optimizer() = IpoptOptimizer(print_level=0)
         #     1 <= x1, x2, x3, x4 <= 5
         # Start at (1,5,5,1)
         # End at (1.000..., 4.743..., 3.821..., 1.379...)
-        m = Model(optimizer=new_optimizer())
+        m = Model(with_optimizer(IpoptOptimize, print_level=0))
```
mlubin: Typo: IpoptOptimizer. This file isn't run under CI because it depends on Ipopt. Please test manually.

IainNZ:
The word `backend` doesn't seem to appear in MathOptInterface.jl at all - does it have a formal definition?

docs/src/solvers.md
```diff
@@ -1,7 +1,84 @@
 Interacting with solvers
 ========================
 
-TODO: Describe the connection between JuMP and solvers. Automatic vs. Manual
-mode. CachingOptimizer. How to set/change solvers. How to set parameters (solver
+A JuMP model stores a
+[MathOptInterface (MOI)](https://github.com/JuliaOpt/MathOptInterface.jl)
+backend internally that contains the optimization solver. The JuMP layer on top
```
mlubin: "contains" -> "acts as"?

mlubin: The backend might also not support `optimize!`.

mlubin: "JuMP can be viewed as a lightweight user-friendly layer on top of the MOI backend.

docs/src/solvers.md
```diff
@@ -1,7 +1,84 @@
...
+  both a reference to the JuMP model and the MOI index of the variable (resp.
+  constraint).
+* JuMP gives the constraints to the MOI backend in the form provided by the user
+  without doing any automatic reformulation.
+* variables additions, constraints additions/modifications and objective
+  modifications are directly applied to the MOI backend thus expecting the
+  backend to support such modifications.
+
+While this allows JuMP API to to be a thin wrapper on top of the solver API,
+as mentionned in the last point above, this seems rather demanding on the
```
mlubin: Spelling: "mentioned"

docs/src/solvers.md
```diff
@@ -1,7 +1,84 @@
...
+JuMP models can be created in three different modes: Automatic, Manual and
+Direct.
+
+## Automatic and Manual modes
+
+In Automatic and Manual modes, two MOI layers are automatically applied to the
+optimizer:
+
+* `CachingOptimizer`: it maintain a cache of the model so that when the
+  the optimizer does not support a modification of the model, the optimizer's
```
mlubin: "incremental change" seems more appropriate than "modification" (since it could include adding constraints)

src/JuMP.jl
```diff
@@ -77,6 +78,43 @@ const MOIBIN = MOICON{MOI.SingleVariable,MOI.ZeroOne}
 
 @MOIU.model JuMPMOIModel (ZeroOne, Integer) (EqualTo, GreaterThan, LessThan, Interval) (Zeros, Nonnegatives, Nonpositives, SecondOrderCone, RotatedSecondOrderCone, GeometricMeanCone, PositiveSemidefiniteConeTriangle, PositiveSemidefiniteConeSquare, RootDetConeTriangle, RootDetConeSquare, LogDetConeTriangle, LogDetConeSquare) () (SingleVariable,) (ScalarAffineFunction,ScalarQuadraticFunction) (VectorOfVariables,) (VectorAffineFunction,)
 
+struct Factory
```
mlubin: `OptimizerFactory` maybe?
Add a comment describing the purpose. It's a user-friendly closure.

src/JuMP.jl
```diff
@@ -124,42 +168,18 @@ mutable struct Model <: AbstractModel
...
     ext::Dict{Symbol, Any}
 
-    # Default constructor.
-    function Model(;
-            mode::ModelMode=Automatic,
-            backend=nothing,
-            optimizer=nothing,
-            bridge_constraints=true)
+    # Inner constructor
+    function Model(factory::Union{Nothing, Factory}, moibackend::MOI.ModelLike)
```
mlubin: Does this still need to be an inner constructor?

src/optimizerinterface.jl
```diff
@@ -3,6 +3,27 @@
...
+
+## Examples
+
+The following sets the optimizer of `model` to be
+`IpoptOptimizer(print_level=0)`:
+```julia
+setoptimizer(model, with_optimizer(IpoptOptimizer, print_level=0))
+```
+"""
+function setoptimizer(model::Model, factory::Factory)
```
mlubin: We really need to decide on a convention for underscores. `with_optimizer` and `setoptimizer` are inconsistent.

src/JuMP.jl
```diff
@@ -107,6 +145,12 @@ mutable struct Model <: AbstractModel
...
 
+    # Factory used to create a new optimizer, it is kept as it might be needed
+    # again if the user requests a copy of the model using `Base.copy`.
+    # In Manual and Automatic mode: Factory used to create the optimizer or
+    #                               Nothing if it has not already been set
+    # In Direct mode: Nothing
+    factory::Union{Nothing, Factory}
     # In Manual and Automatic modes, LazyBridgeOptimizer{CachingOptimizer}.
     # In Direct mode, will hold an AbstractOptimizer.
     moibackend::MOI.AbstractOptimizer
```
mlubin: Could this be an `MOI.ModelLike` to support non-optimizer backends?

mlubin:
"backend" isn't an MOI concept. It's usage needs to be defined in the docs. It's called a backend instead of the optimizer because:

- In Manual and Automatic modes, it's not the same optimizer as what the user provided.
- I imagine that it could not support `optimize!`, and just be a model container, e.g., backed by an MPS file for example.

src/JuMP.jl
blegat: Yes, I don't remember which PR removed that possibility. The model of the CachingOptimizer can be a `MOI.ModelLike` and `CachingOptimizer` will still be an `AbstractOptimizer` so this just removes the possibility to use a non-optimizer in direct mode. Do we want to allow non-optimizer in direct mode ?

src/optimizerinterface.jl
blegat: I agree with using underscores, it's a typo :)

src/JuMP.jl
```diff
@@ -123,50 +174,108 @@ mutable struct Model <: AbstractModel
...
+                 Dict{Symbol, Any}(),
+                 0,
+                 Dict{Symbol, Any}())
+end
+
+"""
+    Model(; caching_mode::MOIU.CachingOptimizerMode=MOIU.Automatic,
+            bridge_constraints::Bool=true)
+
+Return a new JuMP model without any optimizer storing the model in a cache.
```
mlubin: "without any optimizer storing the model in a cache" or just "without any optimizer"?

src/JuMP.jl
```diff
@@ -123,50 +174,108 @@ mutable struct Model <: AbstractModel
...
+    return model
+end
+
+"""
+    direct_model(backend::MOI.ModelLike)
+
+Return a new JuMP model using `backend` to store the model and solve it. As
+opposed to the [`Model`](@ref) constructor, no cache of the model is stored
+outside of `backend` and no bridges are automatically applied to `backend`.
+The absence of cache reduces the memory footprint but it is importnat to bear
```
mlubin: Typo: "important"

src/JuMP.jl
```diff
@@ -123,50 +174,108 @@ mutable struct Model <: AbstractModel
...
+in mind the following implications of creating models using this *direct* mode:
+
+* When `backend` does not support an operation such as adding
+  variables/constraints after solver or modifying constraints, an error is
+  thrown. With models created using the [`Model`](@ref) constructor, such
+  situations can be dealt with modifying the cache only and copying the model
```
mlubin: "such situations can be dealt with modifying the cache only and copying the model cache once `JuMP.optimize` is called." -> "such situations are dealt with by storing the modifications in a cache and loading them into the optimizer when `JuMP.optimize` is called".

src/JuMP.jl
```diff
@@ -123,50 +174,108 @@ mutable struct Model <: AbstractModel
...
+  cache once `JuMP.optimize` is called.
+* When `backend` does not support a constraint type, the constraint is not
```
mlubin: More clearly, "no constraint bridging is supported by default".

src/JuMP.jl
```diff
@@ -123,50 +174,108 @@ mutable struct Model <: AbstractModel
...
+  automatically bridged to constraints supported by `backend`.
+* The optimizer used cannot be changed. With models created using the
```
mlubin: The explanation past "The optimizer used cannot be changed" is too verbose. I would just say "The optimizer cannot be changed after the Model is constructed."

src/JuMP.jl
```diff
@@ -77,6 +78,60 @@ const MOIBIN = MOICON{MOI.SingleVariable,MOI.ZeroOne}
...
+created with [`with_optimizer`](@ref) and new models are created from the
+factory with [`create_model`](@ref).
+
+## Examples
+
+The following construct a factory and then use it to create two independent
+`IpoptOptimizer`s:
+```julia
+factory = with_optimizer(IpoptOptimizer, print_level=0)
+optimizer1 = JuMP.create_model(factory)
```
mlubin: I'd expect `JuMP.create_model` to create a JuMP model. This is very confusing. What about overloading the call operator on the factory? Like `factory()` to create an optimizer.

src/optimizerinterface.jl
```diff
@@ -30,15 +30,30 @@ function MOIU.attachoptimizer!(model::Model)
 end
 
 
-function optimize(model::Model;
-                ignore_optimize_hook=(model.optimizehook===nothing))
+"""
+    function optimize(model::Model,
+                      factory::Union{Nothing, OptimizerFactory} = nothing;
```
mlubin: It's better to call it `optimizer_factory` (in code) and "optimizer factory" (in text). `factory` is too generic.

test/generate_and_solve.jl
```diff
@@ -39,22 +39,19 @@
...
-        MOI.set!(mocksolver, MOI.PrimalStatus(), MOI.FeasiblePoint)
-        MOI.set!(mocksolver, MOI.DualStatus(), MOI.FeasiblePoint)
-        MOI.set!(mocksolver, MOI.VariablePrimal(), JuMP.optimizerindex(x), 1.0)
-        MOI.set!(mocksolver, MOI.VariablePrimal(), JuMP.optimizerindex(y), 0.0)
-        MOI.set!(mocksolver, MOI.ConstraintDual(), JuMP.optimizerindex(c), -1.0)
-        MOI.set!(mocksolver, MOI.ConstraintDual(), JuMP.optimizerindex(JuMP.UpperBoundRef(x)), 0.0)
-        MOI.set!(mocksolver, MOI.ConstraintDual(), JuMP.optimizerindex(JuMP.LowerBoundRef(y)), 1.0)
-
-        JuMP.optimize(m)
+        JuMP.optimize(m, with_optimizer(MOIU.MockOptimizer, JuMP.JuMPMOIModel{Float64}(), evalobjective=false))
```
mlubin: I like how this looks.

docs/src/solvers.md
```diff
@@ -1,7 +1,89 @@
...
+function:
+```@docs
+with_optimizer
+```
+
+The optimizer factory can be set to the JuMP model in the
+[`JuMP.optimize`](@ref) function:
+```@docs
+JuMP.optimize
+```
```
mlubin: We're completely missing a discussion about how to control the caching optimizer states. This can be done later, but we should have a TODO.

src/optimizerinterface.jl
```diff
@@ -49,6 +49,12 @@ function optimize(model::Model,
     end
 
     if optimizer_factory !== nothing
+        if mode(model) == Direct
+            error("An optimizer factory cannot be provided at the `optimize` call in Direct mode")
```
mlubin: Nit: period at the end of the error (as below).

test/generate_and_solve.jl
```diff
@@ -90,6 +90,7 @@
         MOI.set!(mockoptimizer, MOI.ConstraintDual(), JuMP.optimizerindex(JuMP.UpperBoundRef(x)), 0.0)
         MOI.set!(mockoptimizer, MOI.ConstraintDual(), JuMP.optimizerindex(JuMP.LowerBoundRef(y)), 1.0)
 
+        @test_throws ErrorException JuMP.optimize(m, with_optimizer(MOIU.MockOptimizer, JuMP.JuMPMOIModel{Float64}()))
```
mlubin: It's better to add a separate small test for this (also below).

Issue #957, opened 2017-02-02
Base.show(model) calling print on Jupyter


bbrunaud:
Hello

Whenever a model is defined in Jupyter, the output is the full printout of the model, which can be annoying (many many lines of output). Is this the intended behavior?.

regards

Braulio

mlubin:
If something is annoying, we're likely to accept a PR fixing it.

blegat:
Jupyter is calling something like `show` on the result of the command. The fact that the Model prints itself when told to do so seems reasonable.
In any case, one can use `;` at the end of expression to avoid 
@bbrunaud Would you say that the printout of the model is too verbose ? Or that some function returns the model, triggering a full printout while they could return `nothing` ? Doing `m = Model()` displays the empty model, which is not a huge output.

odow:
Ref #1180

odow:
This is going to take some refactoring, because the issue is that we have a mix of `print` and `show`, a summary, a plain text model, and a LaTeX model.

In the REPL, `Julia> model` calls
https://github.com/jump-dev/JuMP.jl/blob/a171d750d0103304f2dbf1916202a69ab24bd5a9/src/print.jl#L152-L161

but in IJulia/Documenter, ending a cell with `model` calls
https://github.com/jump-dev/JuMP.jl/blob/a171d750d0103304f2dbf1916202a69ab24bd5a9/src/print.jl#L195-L197

but calling `print(model)` calls:
https://github.com/jump-dev/JuMP.jl/blob/a171d750d0103304f2dbf1916202a69ab24bd5a9/src/print.jl#L192-L194

To bring some sanity, I'd propose

* `Julia> model` or ending a cell with `model` prints the summary
* `print(model)` prints the model with all constraints in plain-text
* `print(model; latex = true)` prints the latex version of the model
  * In IJulia and documenter, this pretty prints in MathJax.
  * In the REPL, this prints a string that can be copy-pasted into a tex file. 

The only breaking change for this from a user-perspective is that models in IJulia would show a summary by default, and you'd have to use `print(model; latex = true)` to get the latex version.

### Proof-of-concept

```Julia
struct Model end

function _print_summary(io::IO, model::Model)
    print(io, "model summary")
end

function _print_model(io::IO, model::Model)
    print(io, "full model")
end

function _print_latex(io::IO, model::Model)
    print(io, "\$\$\\min Model + latex\$\$")
end


Base.show(io::IO, model::Model) = _print_summary(io, model)

[...]

PR #1389, opened 2018-07-29
First pass of printing for MOI constraints.

### LINKED ISSUE #1180: Printing the JuMP/MOI model
... is not yet implemented.
https://github.com/JuliaOpt/JuMP.jl/blob/1e881e6d23b1bf85cc26912a59561e1735f67524/src/print.jl#L106-L108

mlubin:
Feels like I've untangled the printing rat's nest. JuMP `ConstraintRef` objects now print out the underlying constraint. There's obvious room for improvement, especially in how constraint sets print out (like, print fancy versions in IJulia). See the tests for example of what the print-out looks like. I also print the constraint name if it's non-empty.

Does not yet resolve the issue #1180 because we need to decide what model printing should look like. Based on #957 we might want to print a summary by default and have another method to get the extensive-form print-out of the model.

@IainNZ, care to review since you did most of the original printing work?

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -28,97 +28,136 @@ abstract type REPLMode <: PrintMode end
 abstract type IJuliaMode <: PrintMode end
 
 # Whether something is zero or not for the purposes of printing it
-# oneunit is useful e.g. if coef is a Unitful quantity
-iszeroforprinting(coef) = abs(coef) < 1e-10 * oneunit(coef)
-# Whether something is one or not for the purposes of printing it
-isoneforprinting(coef) = iszeroforprinting(abs(coef) - oneunit(coef))
-str_sign(coef) = coef < zero(coef) ? " - " : " + "
-
-# List of indices available for variable printing
-const DIMS = ["i","j","k","l","m","n"]
+# oneunit is useful e.g. if coef is a Unitful quantity.
+is_zero_for_printing(coef) = abs(coef) < 1e-10 * oneunit(coef)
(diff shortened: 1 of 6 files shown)


src/print.jl
```diff
@@ -28,182 +28,223 @@ abstract type REPLMode <: PrintMode end
...
+end
+
+wrap_in_math_mode(str) = "\$\$ $str \$\$"
+wrap_in_inline_math_mode(str) = "\$ $str \$"
 
 #------------------------------------------------------------------------
 ## Model
 #------------------------------------------------------------------------
-function Base.show(io::IO, m::Model) # TODO temporary
+function Base.show(io::IO, model::Model) # TODO temporary
```
IainNZ: Nit: attach the TODO to the issue?

src/print.jl
```diff
@@ -28,182 +28,223 @@ abstract type REPLMode <: PrintMode end
...
         return "noname"
     end
 end
-function var_str(::Type{IJuliaMode}, v::AbstractVariableRef; mathmode=true)
+function var_string(::Type{IJuliaMode}, v::AbstractVariableRef)
     var_name = name(v)
     if !isempty(var_name)
         # TODO: This is wrong if variable name constains extra "]"
-        return math(replace(replace(var_name,"[","_{",1),"]","}"), mathmode)
+        return replace(replace(var_name,"[","_{",1),"]","}")
```
IainNZ: Style: spaces between arguments.

PR #1395, opened 2018-08-04
Add shape concept

blegat:
Every constraint should be mapped into a `Vector` in `MOI.AbstractVectorSet` to be given to MOI and then MOI returns a vector for `ConstraintPrimal` and `ConstraintDual`. However, when the user do
```julia
@constraint some_object in some_set
```
and `some_object` is not a vector, it is rather confusing to get a vector back as `ConstraintPrimal` and `ConstraintDual`. This is currently the case when constraining a matrix to be PSD. It used to return a matrix in JuMP v0.18 so this is currently blocking for JuMP v0.19.

One solution is to keep track of the original shape in an `AbstractShape` object and reshape it in `resultdual` and `resultvalue` (which is currently missing: https://github.com/JuliaOpt/JuMP.jl/issues/1394).
This solution has been implemented in this PR.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -370,7 +370,7 @@ Base.isempty(::AbstractJuMPScalar) = false
 """
     AbstractShape
 
-Abstract vectorizable shape. Given a vectorized form of an object of shape
+Abstract vectorizable shape. Given a flat vector form of an object of shape
 `shape`, the original object can be obtained by [`reshape`](@ref).
 """
 abstract type AbstractShape end
@@ -425,10 +425,10 @@ Return an object in it original shape `shape` given its vectorized form
 
 ## Examples
 
-Given a [`SymmetricMatrix`](@ref) of vectorized form `[1, 2, 3]`, the following
-code retrieve the matrix `Symmetric(Matrix[1 2; 2 3])`:
+Given a [`SymmetricMatrixShape`](@ref) of vectorized form `[1, 2, 3]`, the
+following code retrieve the matrix `Symmetric(Matrix[1 2; 2 3])`:
 ```julia
-reshape([1, 2, 3], SymmetricMatrix(2))
+reshape([1, 2, 3], SymmetricMatrixShape(2))
 ```
 """
 function reshape end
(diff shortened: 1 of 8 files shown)


src/JuMP.jl
```diff
@@ -397,7 +487,11 @@ false if not.
 """
 MOI.canget(m::Model, attr::MOI.AbstractModelAttribute) = MOI.canget(m.moibackend, attr)
 MOI.canget(m::Model, attr::MOI.AbstractVariableAttribute, ::Type{VariableRef}) = MOI.canget(m.moibackend, attr, MOIVAR)
-MOI.canget(m::Model, attr::MOI.AbstractConstraintAttribute, ::Type{ConstraintRef{Model,T}}) where {T <: MOICON} = MOI.canget(m.moibackend, attr, T)
+function MOI.canget(model::Model, attr::MOI.AbstractConstraintAttribute,
+                    ::Type{ConstraintRef{Model, T, Shape}}) where {T <: MOICON,
```
mlubin: Can't you drop the shape parameter in this case since it's not referred to?

src/sd.jl
```diff
@@ -1,19 +1,62 @@
 # Used in @constraint model x in PSDCone
 struct PSDCone end
 
+"""
+    SymmetricMatrix
```
mlubin: SymmetricMatrixShape?

src/sd.jl
```diff
@@ -1,19 +1,62 @@
...
+
+Symmetric square matrix of `side_dimension` rows and columns. The vectorized
```
mlubin: Shape object for a symmetric ...

src/sd.jl
```diff
@@ -1,19 +1,62 @@
...
+        for i in 1:j
+            k += 1
+            matrix[j, i] = matrix[i, j] = vectorized_form[k]
+        end
+    end
+    return Symmetric(matrix)
+end
+
+"""
+    SquareMatrix
```
mlubin: SquareMatrixShape

src/JuMP.jl
```diff
@@ -278,13 +278,103 @@ Base.next(x::AbstractJuMPScalar, state) = (x, true)
 Base.done(::AbstractJuMPScalar, state) = state
 Base.isempty(::AbstractJuMPScalar) = false
 
+"""
+    AbstractShape
+
+Abstract vectorizable shape. Given a vectorized form of an object of shape
```
mlubin: "vectorized" is used in lots of contexts. Maybe "flat vector form" would be less ambiguous?

mlubin:
Note CI failures

blegat:
Resolved :)

blegat:
Then it has VectorShape. If the user enter the constraint in plain vector form he gets the result in plain vector form

mlubin:
Sounds good. Could you add a test for that case? I didn't see one.

PR #1396, opened 2018-08-04
Add with_optimizer shortcut for Module.Optimizer

blegat:
For solvers following the guideline: https://github.com/JuliaOpt/MathOptInterface.jl/issues/425#issuecomment-409707912, this PR allows to write `with_optimizer(CSDP)` instead of `with_optimizer(CSDP.Optimizer)`

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -123,6 +123,15 @@ function with_optimizer(constructor::Type, args...; kwargs...)
     return OptimizerFactory(constructor, args, kwargs)
 end
 
+"""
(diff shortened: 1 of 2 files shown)


mlubin:
This has big implications for how solver wrappers should be structured which I'm not sure we fully understand yet, especially with the changes to how repositories can be structured in Julia 0.7. It's a nice feature but I'd rather not introduce one more thing that's likely to change after 0.19 stabilizes.

mlubin:
Closing since this is unlikely to be merged as is.

PR #1399, opened 2018-08-04
fixes to pass tests on 0.7

mlubin:
Tests are passing. The only significant test that I had to disable was sparse matrix-sparse matrix multiplication involving JuMP types. We can fix that in a follow-up PR (that whole chunk of operator code needs to be rewritten.)
A lot of the changes are also relevant for `release-0.18`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,10 @@ os:
 julia:
     - 0.6
     - 0.7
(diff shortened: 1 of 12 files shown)


mlubin:
CI is green. :)
Waiting for a review before merging.

PR #1404, opened 2018-08-05
[release-0.18] more fixes for 0.7

ExpandingMan:
Hello all.  I've continued the process of getting 0.18 up and running on 0.7.  As this will be a long process, I'm hoping that we can do this with lots of small PR's like this one, not worrying much about whether we are passing tests on 0.7 until things are a little further along.  (When I updated DataFrames.jl for 0.7 I did one gigantic PR, and it was not a good way of doing it.)

The biggest changes I've made so far are probably implementing `iterate` methods and constructors (as now `convert` falls back to constructors and not vice versa).

Many of the errors are macro errors, and as the macro code is incredibly dense these will be the hardest to work on, so I've avoided fixing macros so far.

I have a question concerning the desired print behavior, so that perhaps I can make another fix before this gets merged.  Looking at the test set starting [here](https://github.com/JuliaOpt/JuMP.jl/blob/release-0.18/test/print.jl#L179) I cannot understand why `empty_set` and `empty_free` would be expected to print differently.  They are both variables being initialized with empty arrays, both of those macros create `JuMPContainer` objects with very similar type signatures exactly as I would expect them to.  The code in `print.jl` prints `empty_free` as `empty_free (no indices)` and, having poured through that code, it seems like the intended behavior, so I don't understand why that test is that way.  Note that [this line](https://github.com/JuliaOpt/JuMP.jl/blob/release-0.18/src/print.jl#L411) causes all `JuMPContainer`s to print with the name they were initialized with, as I would expect.  Perhaps this is an error in the macro parsing and `empty_free` was supposed to be an `Array{Variable}`?

Anyway, any suggestions for simply fixing this PR up a bit more before moving onto the next iteration would be appreciated.  Thanks!

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
-MathProgBase 0.6
-ReverseDiffSparse 0.8
-ForwardDiff 0.5
+MathProgBase 0.6 0.8
+ReverseDiffSparse 0.8 0.9
+ForwardDiff 0.5 0.8
 Calculus
-Compat 0.49
+Compat 1.0.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
-MathProgBase 0.6 0.8
-ReverseDiffSparse 0.8 0.9
-ForwardDiff 0.5 0.8
+MathProgBase 0.6
+ReverseDiffSparse 0.8
+ForwardDiff 0.5
 Calculus
 Compat 0.49
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -336,8 +336,14 @@ abstract type AbstractConstraint end
 # In JuMP, used only for Variable. Useful primarily for extensions
 abstract type AbstractJuMPScalar end
 
-Base.iterate(x::AbstractJuMPScalar) = (x, true)
-Base.iterate(x::AbstractJuMPScalar, state) = nothing
+if VERSION < v"0.7-"
+    Base.start(::AbstractJuMPScalar) = false
+    Base.next(x::AbstractJuMPScalar, state) = (x, true)
+    Base.done(::AbstractJuMPScalar, state) = state
+else
+    Base.iterate(x::AbstractJuMPScalar) = (x, true)
+    Base.iterate(x::AbstractJuMPScalar, state) = nothing
+end
 Base.isempty(::AbstractJuMPScalar) = false
 
 #############################################################################
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -7,9 +7,7 @@
 # An algebraic modeling language for Julia
 # See http://github.com/JuliaOpt/JuMP.jl
 #############################################################################
-
-__precompile__()
-
+VERSION < v"0.7.0-beta2.199" && __precompile__()
 module JuMP
 
 import Base.map
@@ -338,9 +336,8 @@ abstract type AbstractConstraint end
 # In JuMP, used only for Variable. Useful primarily for extensions
 abstract type AbstractJuMPScalar end
 
-Base.start(::AbstractJuMPScalar) = false
-Base.next(x::AbstractJuMPScalar, state) = (x, true)
-Base.done(::AbstractJuMPScalar, state) = state
+Base.iterate(x::AbstractJuMPScalar) = (x, true)
+Base.iterate(x::AbstractJuMPScalar, state) = nothing
 Base.isempty(::AbstractJuMPScalar) = false
 
 #############################################################################
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -41,18 +41,24 @@ end
 
 linearterms(aff::GenericAffExpr) = LinearTermIterator(aff)
 
-function Base.iterate(lti::LinearTermIterator)
-    if length(lti.aff.vars) ≥ 1
-        ((lti.aff.coeffs[1], lti.aff.vars[1]), 2)
-    else
-        nothing
+if VERSION < v"0.7-"
+    Base.start(lti::LinearTermIterator) = 1
+    Base.done( lti::LinearTermIterator, state::Int) = state > length(lti.aff.vars)
+    Base.next( lti::LinearTermIterator, state::Int) = ((lti.aff.coeffs[state], lti.aff.vars[state]), state+1)
+else
+    function Base.iterate(lti::LinearTermIterator)
+        if length(lti.aff.vars) ≥ 1
+            ((lti.aff.coeffs[1], lti.aff.vars[1]), 2)
+        else
+            nothing
+        end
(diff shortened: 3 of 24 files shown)


src/JuMP.jl
```diff
@@ -338,9 +336,8 @@ abstract type AbstractConstraint end
 # In JuMP, used only for Variable. Useful primarily for extensions
 abstract type AbstractJuMPScalar end
 
-Base.start(::AbstractJuMPScalar) = false
-Base.next(x::AbstractJuMPScalar, state) = (x, true)
-Base.done(::AbstractJuMPScalar, state) = state
+Base.iterate(x::AbstractJuMPScalar) = (x, true)
```
blegat: We should do `if` `else` with Julia version and keep start/next/done on v0.6

blegat: I didn't know there was a `Compat.iterate`, could you point me to where it's defined ?

REQUIRE
```diff
@@ -1,6 +1,6 @@
 julia 0.6
-MathProgBase 0.6 0.8
-ReverseDiffSparse 0.8 0.9
-ForwardDiff 0.5 0.8
+MathProgBase 0.6
```
blegat: Why removing the upper bounds ?

src/affexpr.jl
```diff
@@ -83,10 +102,12 @@ end
 # Alias for (Float64, Variable), the specific GenericAffExpr used by JuMP
 const AffExpr = GenericAffExpr{Float64,Variable}
 AffExpr() = zero(AffExpr)
+AffExpr(v::Variable) = AffExpr([v], [1.], 0.)
```
blegat: Why is it needed to define these constructors ?

test/callback.jl
```diff
@@ -10,8 +10,7 @@
 # test/callback.jl
 # Testing callbacks
 #############################################################################
-using JuMP, MathProgBase
-using Base.Test
+using MathProgBase
```
blegat: Having `using JuMP` allows to run the file separately. Just replace `Base.Test` by `Compat.Test`. It's better not to mix this PR with other changes

test/hockschittkowski/runhs.jl
```diff
@@ -22,7 +22,7 @@ if Pkg.installed("Ipopt") == nothing
     println("Cannot run NLP tests because Ipopt is not installed.")
     exit(1)
 end
-using Ipopt, JuMP, Base.Test
+using Ipopt, JuMP, Test
```
blegat: Compat.Test

REQUIRE
ExpandingMan: Sorry, I forgot about this, this may be a potential problem.  In some of these cases the upper bounds are not good for 0.7.  We might have to go back and make 0.7 compatibility updates to the earlier releases of some of these packages.  I'll look into this when I get a chance.

test/macros.jl
```diff
@@ -1,9 +1,6 @@
 # test/macros.jl
 # Testing macros work correctly
 #############################################################################
-using JuMP
-using Base.Test
```
blegat: Same here, don't remove `using JuMP`, just replace `Base.Test` by `Compat.Test`

REQUIRE
blegat: If we make a new version of these packages that work with v0.7, we will update the upper bounds in JuMP

src/affexpr.jl
ExpandingMan: This particular one is a little weird, I'm not 100% sure why this one was throwing the warning.  It must have something to do with the alias.  Of course, the reason the constructors were added in general is because `convert` now falls back to constructors, and not vice versa.

REQUIRE
ExpandingMan: Yeah, sorry, I took those out for testing and forgot to restore them.  Will do so now.

src/print.jl
```diff
@@ -396,7 +396,7 @@ exprstr(n::Norm) = norm_str(REPLMode, n)
 #------------------------------------------------------------------------
 ## JuMPContainer{Variable}
 #------------------------------------------------------------------------
-Base.show(io::IO, j::Union{JuMPContainer{Variable}, Array{Variable}}) = print(io, cont_str(REPLMode,j))
```
ExpandingMan: This issue was very weird, I don't understand how this was not causing a stack overflow in 0.6 as it does in 0.7.  Will revisit after looking at 0.6 tests.

src/affexpr.jl
ExpandingMan: Apparently the constructor is used, because I'm getting the warning, but it's very difficult to track down exactly what is causing it.

blegat: Ok then it's good :)

mlubin:
@ExpandingMan please see the discussion at https://discourse.julialang.org/t/jump-and-julia-0-7. 

> Looking at the test set starting here I cannot understand why empty_set and empty_free would be expected to print differently.

`empty_free` is an `Array{Variable}` while `empty_set` is a `JuMPArray{Variable}`, so they don't necessarily print the same way. 0.18 has a pretty complex printing system with lazy name generation that has been greatly simplified on 0.19.

ExpandingMan:
Ok, that's pretty much what I thought.

ExpandingMan:
Just so I don't forget myself, the reason for the stack overflow in this is that I replaced the `Base.showarray` function with some version of `Base.show`.  The `Base.showarray` function is completely gone in 0.7, so I'm really not sure what to replace it with.

ExpandingMan:
Ok, passing on 0.6, still tons of problems on 0.7 but it's an improvement.

If you guys approve I suggest merging this so we can do more improvements in other PR's.

One step at a time...

src/print.jl
```diff
@@ -436,15 +443,15 @@ function cont_str(mode, j, sym::PrintSymbols)
...
+    idxvars = Array{String}(undef, num_dims)
     dimidx = 1
     for i in 1:num_dims
         if data.indexexprs[i].idxvar == nothing
             while DIMS[dimidx] in locvars
                 dimidx += 1
             end
             if dimidx > length(DIMS)
-                error("Unexpectedly ran out of indices")
+                throw(ErrorException("Unexpectedly ran out of indices"))
```
mlubin: I'd prefer `Compat.@error` here.

ExpandingMan: That doesn't actually throw an exception.  At least it shouldn't, because `@error` in 0.7 doesn't throw an exception.

ExpandingMan: (Note that `error` in 0.6 *did* throw an exception.)

mlubin: Wait, `error` is not deprecated in 0.7. Why does this need to change?

ExpandingMan: oh wow... not sure why I thought it was then.  Ok changing...

PR #1406, opened 2018-08-05
Clean up operator_warn.

IainNZ:
Rephrase docstring.
Update warning message to not refer to removed push!/append!.
Style guidification.
Inline the (affexpr,affexpr) method as its only called from that
one place, and there doesn't seem any intention to add more.
It just adds a level of indirection that isn't well documented.
Remove strange fallback.
Remove comment in affexpr about deprecation - it won't happen.

### DIFF
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -146,6 +146,8 @@ product of two variables.
 """
 function add_to_expression! end
 
+# TODO: add deprecations for Base.push! and Base.append!
+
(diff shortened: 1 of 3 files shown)


src/affexpr.jl
```diff
@@ -146,8 +146,6 @@ product of two variables.
 """
 function add_to_expression! end
 
-# TODO: add deprecations for Base.push! and Base.append!
```
mlubin: This was a recent comment, shouldn't be removed yet.

IainNZ: Fixed.

PR #1407, opened 2018-08-05
Remove some code for weird arrays.

IainNZ:
Move `one_indexed` next to only place that uses it.


IainNZ:
Breadcrumb trail to #980 that added this stuff.

PR #1410, opened 2018-08-07
added 0.7 iterators

ExpandingMan:
I've added iterators using `Base.iterate` for 0.7.  Only 3 objects (`LinearTermIterator`, `QuadTermIterator` and `AbstractJuMPScalar`) require manual iterator implementations.  There are a number of different ways of going about this, I've chosen to do it by defining more methods for the existing helper function `reorder_iterator`.  Note that if this is *not* done than each of the `iterate` functions will need to de-structure output.  This method also eliminates splatting (which was probably ok here).  I'm not necessarily claiming it was the best way of doing it, but it seemed like a good approach to me.

I've also taken the liberty to make a few very simple fixes to 0.7 deprecation warnings.

I've yet to run the full tests locally, so lets see how this goes.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -8,7 +8,7 @@
 # See http://github.com/JuliaOpt/JuMP.jl
 #############################################################################
 
-__precompile__()
+VERSION < v"0.7.0-beta2.199" && __precompile__()
 
 module JuMP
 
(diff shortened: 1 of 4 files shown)


mlubin:
Thanks!

PR #1411, opened 2018-08-07
RFC: begin writing some docs

odow:


docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+To illustrate these three types of variables, consider the following JuMP code
+(the full syntax is explained below):
+```julia
+model = Model()
+@variable(model, x[1:2])
+```
+This code does three things:
+1. it tells `model` to add two _optimization_ variables
+2. it creates two _JuMP_ variables that act as references to those
+optimization variables
```
blegat: should be indented by 3 spaces to align with `it`

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+`MOI.VariableIndex`, and also contains a reference to the JuMP model.)
+
+To illustrate these three types of variables, consider the following JuMP code
+(the full syntax is explained below):
+```julia
+model = Model()
+@variable(model, x[1:2])
+```
+This code does three things:
+1. it tells `model` to add two _optimization_ variables
```
blegat: doesn't it need a comma at the end (not sure) ?

blegat: Maybe say "of names `x[1]` and `x[2]` ?

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+Therefore, we have:
+```julia
+julia> x
+2-element Array{JuMP.VariableRef, 1}
+ x[1]
+ x[2]
+```
+
+To reduce confusion, we will attempt, where possible, to always refer to
+variables with their corresponding prefix.
```
blegat: I like this disambiguation section !

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+JuMP variables can have attributes, such as names or an initial primal start
+value. We illustrate the name attribute in the following example:
+```julia
+model = Model()
+@variable(model, x, basename="decision variable")
+```
+This code does four things:
+1. it tells `model` to add one _optimization_ variable
+2. it creates one _JuMP_ variable that acts as a reference to that
+optimization variables
```
blegat: indent

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+3. it binds the JuMP variable to the name `x`
+4. it tells JuMP that the _name_ attribute of this JuMP variable is "decision
```
blegat: I thought when reading the previous section that the optimization variable was the MOI variable. in that case, `basename` is the name of the optimization variable isn't it ?

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+variable". JuMP uses the value of `basename` when it has to print the variable
```
blegat: indent

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+For example:
+```julia
+julia> x
+decision variable
+
+julia> typeof(x)
+JuMP.VariableRef
+```
+
+Because `x` is a JuMP variable, I can bind it to a different value. For example,
```
blegat: You meant "Julia" instead of "JuMP" ?

docs/src/variables.md
blegat: Maybe not as names are discussed below...

mlubin:
I'll do a pass through this tomorrow.

docs/src/variables.md
mlubin: More clearly: it adds two optimization variables to `model`

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+value. We illustrate the name attribute in the following example:
+```julia
+model = Model()
+@variable(model, x, basename="decision variable")
+```
+This code does four things:
+1. it tells `model` to add one _optimization_ variable
+2. it creates one _JuMP_ variable that acts as a reference to that
+optimization variables
+3. it binds the JuMP variable to the name `x`
```
mlubin: To the Julia variable `x`?

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+4. it tells JuMP that the _name_ attribute of this JuMP variable is "decision
+variable". JuMP uses the value of `basename` when it has to print the variable
+as a string.
+
+For example:
+```julia
+julia> x
+decision variable
```
mlubin: Seeing this makes me think that we should print variables like `A JuMP VariableRef with name: ...".

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+
+julia> typeof(x)
```
mlubin: This should be formatted as a doc test (you can set it up so that it verifies that the REPL output matches the doc).

docs/src/variables.md
```diff
@@ -1,13 +1,97 @@
...
+function add_component_to_model(model::JuMP.Model)
+    x = model[:x]
+    # ... code that uses x
+end
+function build_model()
+    model = Model()
+    @variable(model, x)
+    add_component_to_model(model)
+end
+```
```
mlubin: Maybe add a TODO to describe looking up JuMP variables by string names once we have a nice syntax for it.

odow:
Okay this is now RFC. I've tried pretty hard to distinguish between optimization variables, JuMP variables (which are `VariableRef`'s), and Julia variables. I know @jrevels likes to call Julia variables bindings (which I kinda like), but I don't know if thats yet-another-piece-of-terminology for new users to learn.

docs/src/variables.md
```diff
@@ -1,23 +1,478 @@
...
+    When creating a variable with only a lower-bound or an upper-bound, the name
+    must appear on the left-hand side. Putting the name on the right-hand side
+    (e.g., `@variable(model, 0 <= x)`) will result in an error.
+
+    **Extra for experts:** the reason for this is that at compile time, JuMP
+    does not type and value information. Therefore, the case `@variable(model,
+    a <= b)` is ambiguous as JuMP cannot infer whether `a` is a constant and
+    `b` is the intended variable name, or vice-versa.
+
+We can query whether an optimization variable has a lower- or upper-bound via
```
mlubin: I'm not sure if the discussion about attributes falls entirely under the scope of the `@variable` macro. However, there's a conflict between presenting the common snippets of how to get things done and presenting the complete API for handling attributes (e.g., setting bounds without `@variable`), so maybe we can push this off for now.

docs/src/variables.md
```diff
@@ -1,23 +1,478 @@
...
+c
+
+julia> @variable(model, 2 <= d <= 3)
+d
+```
+
+!!! note
+    When creating a variable with only a lower-bound or an upper-bound, the name
+    must appear on the left-hand side. Putting the name on the right-hand side
+    (e.g., `@variable(model, 0 <= x)`) will result in an error.
```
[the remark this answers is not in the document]
odow: Woah. The future is here.

docs/src/variables.md
odow: > setting bounds without @variable

I also don't know how to do this, so that might be a reason it's not in these docs... :)

mlubin: See methods around https://github.com/JuliaOpt/JuMP.jl/blob/133df20e9e5b6fcfd166ac300c9f62110ce4917e/src/variables.jl#L269 :)

PR #1420, opened 2018-08-12
Allow Function and UnionAll in with_optimizer

blegat:
This is needed by CSDP as CSDP.Optimizer is a function, it is not a type as the type returned is
```julia
julia> typeof(CSDP.Optimizer())
SemidefiniteOptInterface.SOItoMOIBridge{Float64,CSDP.SDOptimizer}
```
Accepting UnionAll is also needed in case the optimizer type is parametrized

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -83,7 +83,7 @@ struct OptimizerFactory
     # * `Function`: a function, or
     # * `DataType`: a type, or
     # * `UnionAll`: a type with missing parameters.
-    constructor::Union{Function, DataType, UnionAll}
+    constructor
     args::Tuple
     kwargs # type changes from Julia v0.6 to v0.7 so we leave it untyped for now
 end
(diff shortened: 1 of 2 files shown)


src/JuMP.jl
```diff
@@ -102,7 +102,8 @@ the constructor call `IpoptOptimizer(print_level=0)`:
 with_optimizer(IpoptOptimizer, print_level=0)
 ```
 """
-function with_optimizer(constructor::Type, args...; kwargs...)
+function with_optimizer(constructor::Union{Function, DataType, UnionAll},
```
mlubin: The signature should match the docstring. Either drop the types from the `constructor` argument or add them to the docstring. I don't think the types add much since the compiler can't specialize, but we do have to be aware of postponing a `MethodError` that may happen in `(optimizer_factory::OptimizerFactory)()`. Maybe an explicit `hasmethod` in `(optimizer_factory::OptimizerFactory)()` with a nice error message would address that?

PR #1421, opened 2018-08-14
[release-0.18] more updates for 1.0

ExpandingMan:
At the moment tests are failing again on 0.6, I'm still working on this.

In the meantime, perhaps somebody can help me understand the logic behind what type of object should be returned by expressions such as `@variable(m, x[1:N])`.  I'm having a problem right now where 1.0 returns a `JuMPContainer` (seems like the proper behavior?) and 0.6 returns an `Array{Variable}`.  The function `gendict` is getting called in 1.0 to place a constructor in the returned expression for a `JuMPContainer`, but in 0.6 the function `gendict` never gets called, and I'm trying to figure out why.  Thanks!

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
 MathProgBase 0.6 0.8
-ReverseDiffSparse 0.8 0.9
-ForwardDiff 0.5 0.8
+ReverseDiffSparse 0.8
+ForwardDiff 0.7
 Calculus
 Compat 1.0.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
 MathProgBase 0.6 0.8
-ReverseDiffSparse 0.8
-ForwardDiff 0.7
+ReverseDiffSparse 0.8 0.9
+ForwardDiff 0.5 0.8
 Calculus
 Compat 1.0.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
 MathProgBase 0.6 0.8
 ReverseDiffSparse 0.8 0.9
-ForwardDiff 0.5 0.8
+ForwardDiff 0.5 0.9
 Calculus
 Compat 1.0.0
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -29,6 +29,7 @@ if VERSION < v"0.7-"
     const SparseArrays = Compat.SparseArrays
     const Printf = Compat.Printf
     const Sys = Compat.Sys
+    using Compat: rmul!
 end
 
 export
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -31,10 +31,7 @@ if VERSION < v"0.7-"
     const Sys = Compat.Sys
     using Compat: rmul!
     # because iteration of CartesianIndex is broken in 0.6
-    Base.Tuple(idx::CartesianIndex{1}) = (idx[1],)
-    Base.Tuple(idx::CartesianIndex{2}) = (idx[1], idx[2])
-    Base.Tuple(idx::CartesianIndex{3}) = (idx[1], idx[2], idx[3])
-    Base.Tuple(idx::CartesianIndex{4}) = (idx[1], idx[2], idx[3], idx[4])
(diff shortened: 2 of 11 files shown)


blegat:
It should be `Array{Variable}`. We detect that it can be an array as in `1:N`, we can see at parsing time in the macro that the range starts at 1 so we can use an Array

ExpandingMan:
Ok, so the logic is use an `Array` as long as the indexing scheme matches that of an `Array`?

ExpandingMan:
(sorry, those previous tests ran with most commented out, that's why the coverage report looks whacky)

Ok, big progress I think.  You can now compile JuMP 0.18 successfully on 1.0 (might depend on some other packages being updated).  Additionally, some of the most baffling errors, including an incorrect output of the `@variable` macro on 1.0, has been fixed.

ExpandingMan:
Wow actually looking at the 0.7 tests we may not be all that far off.

Please merge this if you approve and I will continue working on another separate PR.

REQUIRE
```diff
@@ -1,6 +1,6 @@
 julia 0.6
 MathProgBase 0.6 0.8
-ReverseDiffSparse 0.8 0.9
```
mlubin: Please keep the upper bound. I'll tag the next ReverseDiffSparse release as 0.8.x.

REQUIRE
```diff
@@ -1,6 +1,6 @@
...
-ForwardDiff 0.5 0.8
```
mlubin: Same, please keep the upper bound (0.9 would be fine since 0.8 works). ForwardDiff may break its API between releases.

src/JuMP.jl
```diff
@@ -29,6 +29,12 @@ if VERSION < v"0.7-"
     const SparseArrays = Compat.SparseArrays
     const Printf = Compat.Printf
     const Sys = Compat.Sys
+    using Compat: rmul!
+    # because iteration of CartesianIndex is broken in 0.6
+    Base.Tuple(idx::CartesianIndex{1}) = (idx[1],)
+    Base.Tuple(idx::CartesianIndex{2}) = (idx[1], idx[2])
+    Base.Tuple(idx::CartesianIndex{3}) = (idx[1], idx[2], idx[3])
+    Base.Tuple(idx::CartesianIndex{4}) = (idx[1], idx[2], idx[3], idx[4])
```
mlubin: Where is this used and why do the method definitions stop at 4?

ExpandingMan: This is because `ind2sub` is deprecated and the iteration protocol of `CartesianIndex` is broken in 0.6 because of iteration protocol change.

Sorry, I had a better plan for this but forgot to put it in.  One sec...

ExpandingMan:
Ah, that would make sense.  That's probably the problem I was having that prompted me to change it in the first place.

src/JuMP.jl
```diff
@@ -29,6 +29,9 @@ if VERSION < v"0.7-"
     const SparseArrays = Compat.SparseArrays
     const Printf = Compat.Printf
     const Sys = Compat.Sys
+    using Compat: rmul!
+    # because iteration of CartesianIndex is broken in 0.6
+    Base.Tuple(idx::CartesianIndex{N}) where {N} = tuple((idx[i] for i ∈ 1:N)...)
```
mlubin: This is type piracy. It's better to define a function with a new name rather than overloading base methods that weren't meant to be overloaded.

Also, where is this used? If it's for every `getindex` with `JuMPArray` the splatting could have serious performance implications.

ExpandingMan: This is used in `KeyIterator` and in printing (do a search for `CartesianIndices`).  As I understand it the splatting only needs to compile once for each *rank* of array it is called on, so it shouldn't be to bad.  This is the replacement for `ind2sub`.

Ok, I was trying very hard to get rid of as much 0.6 code as possible as it'll probably be ancient history in a matter of weeks, but if you prefer I can define a new function instead.

mlubin: > Ok, I was trying very hard to get rid of as much 0.6 code as possible as it'll probably be ancient history in a matter of weeks, but if you prefer I can define a new function instead.

I'm reviewing this as code that we'll be responsible for supporting and responding to issues about. Any ugly hacks need a good justification, and in this case the cleaner version is easy to do.

PR #1423, opened 2018-08-16
[release-0.18] Fix _ind2sub

blegat:
cc @ExpandingMan


ExpandingMan:
Thanks.

Sorry about that, those names are very easy to get confused.  We had gone through several iterations of what to put there, and by the time we decided to use `_ind2sub` I just typed in the wrong type.

blegat:
@ExpandingMan no problem, thanks for all the fixes !

PR #1425, opened 2018-08-17
Fix addition of matrices

blegat:
This PR adds a few tests that cover in particular the bug of https://github.com/JuliaOpt/JuMP.jl/issues/1422. There was also a more subtle bug that was caught with these tests: taking the copy of the matrix do not copy its elements. Taking a deepcopy will copy the model so we don't want to replace copy with deepcopy. The simple fix is to broadcast copy.

Related to https://github.com/JuliaOpt/JuMP.jl/issues/1422 (does not close it as it also need to be fixed on JuMP v0.18)

### DIFF
--- a/src/parseexpr.jl
+++ b/src/parseexpr.jl
@@ -146,8 +146,8 @@ function destructive_add!(aff::GenericAffExpr{C,V}, c::GenericQuadExpr{C,V}, x::
         GenericQuadExpr{C,V}(aff)
     else
         result = c*x
-        add_to_expression!(result.aff, aff)
-        result
+        add_to_expression!(result, aff)
+        return result
     end
 end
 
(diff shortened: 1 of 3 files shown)


src/parseexpr.jl
```diff
@@ -261,14 +261,35 @@ function destructive_add!(ex::AbstractArray{T}, c::Number, x::AbstractArray) whe
...
+end
+function destructive_add!(ex::AbstractArray{<:GenericAffExpr}, c::Number,
+                          x::AbstractArray{<:GenericQuadExpr})
+    result = c*x
+    add_to_expression!.(result, ex)
+    return result
+end
+
+
 destructive_add!(ex, c, x) = ex + c*x
```
mlubin: I don't understand the 5 different fallbacks right here. What is going on and why are tuples needed?

blegat: These were added in https://github.com/JuliaOpt/JuMP.jl/pull/1416 by @mohamed82008 
I suspect it is for ones using the `(...,)` syntax is to make a scalar object that does not implement `Base.broadcastable` broadcastable but a simpler fix might be to define `broadcastable` for `JuMPScalar` elements

mohamed82008: True.

blegat: I added an issue: https://github.com/JuliaOpt/JuMP.jl/issues/1430 as it might need some bikeshedding which will be clearer if done in a separate PR

PR #1428, opened 2018-08-19
Fix index set parsing in JuMP macros on Julia 1.0.

IainNZ:
This is a change introduced between Julia v0.7 and v1.0.

### DIFF
--- a/src/parseexpr.jl
+++ b/src/parseexpr.jl
@@ -4,7 +4,11 @@
 #  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 function tryParseIdxSet(arg::Expr)
-    if arg.head === :(=)
+    if arg.head === :kw  # Julia 1.0 onwards.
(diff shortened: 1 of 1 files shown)


src/parseexpr.jl
```diff
@@ -4,7 +4,11 @@
 #  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 function tryParseIdxSet(arg::Expr)
-    if arg.head === :(=)
+    if arg.head === :kw  # Julia 1.0 onwards.
```
mlubin: It's better to use explicit VERSION checks so that they can be removed semi-automatically in the future.

IainNZ: Should the version check be added for the old behaviour of the new behaviour?

mlubin: I don't think we have a convention, just as long as it's clear which code blocks can be removed when.

PR #1431, opened 2018-08-19
update style recommendations for bangs

mlubin:
Some proposed updates based on the discussion in https://github.com/JuliaOpt/MathOptInterface.jl/issues/475. 

@IainNZ @matbesancon @juan-pablo-vielma @rdeits @IssamT

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -115,7 +115,7 @@ modifies its arguments. We recommend to:
 - Omit `!` when the name itself makes it clear that modification is taking
   place, e.g., `add_constraint` and `set_name`. We depart from the Julia style
   guide because `!` does not provide a reader with any additional information
-  in this case, and adherence to the this convention is not uniform even in base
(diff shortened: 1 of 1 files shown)


matbesancon:
Nice addition to the explanation. This was not in the initial discussion, but I would add putting the mutated argument first, like `do_magic!(model, x, y)`. I found it a lot clearer in other packages to be able to identify it that way

IainNZ:
lgtm

mlubin:
> `!` is used to help readability, it is not an obligation and should not be used when redundant

To clarify, we should try to reduce scope for personal preferences when applying the rules the style guide. There's still a question of if a name clearly implies modification, but we should we should either omit `!` or use `!` based on the answer to that question.

mlubin:
@matbesancon I agree the ordering point is important. I believe that's addressed by the link to the corresponding discussion in the Julia style guide.

PR #1432, opened 2018-08-21
[release-0.18] light at the end of the tunnel

ExpandingMan:
Hello all.  I apologize for the gargantuan size of this PR, but I was on a role and it got away from me.

I've made a large number of fixes to get this running on 1.0.  All non-solver-dependent tests now pass on 1.0 with two exceptions
- There as a block (in print.jl) that I took out completely, because it was extremely pedantic and had been taken out in master (even though the functionality still exists in master).
- There seem to be various errors with OffsetArrays.  According to the stack traces these seem to be entirely contained to OffsetArrays.  For now I have commented out a few (I think 3) tests that use OffsetArrays.

Most of the fixes here involve macros or operators.  I have taken as much code from @mlubin 's fixes to master as possible.  Fortunately this was quite a lot.

Obviously the operator code is now a mess and needs major refectoring (for arrays, scalars should be fine).  I'm pretty sure what's here now should be about as good as what's in master.  I now have lots of good ideas for an operator PR to master, so I hope to get to that relatively soon.

Ok, now the bad news.  I absolutely *for the life of me* cannot figure out the error I'm seeing in 0.6:
```
ERROR: LoadError: generated function body is not pure. this likely means it contains a closure or comprehension.
Stacktrace:
 [1] macro expansion at /home/expandingman/.julia/v0.6/JuMP/src/macros.jl:217 [inlined]
 [2] anonymous at ./<missing>:?
 [3] include_from_node1(::String) at ./loading.jl:576
 [4] include(::String) at ./sysimg.jl:14
while loading /home/expandingman/src/test1.jl, in expression starting on line 11
```
This seems to occur whenever there are ranges present in variable definitions such as `@variable(m, x[-5:5])`.  I have actually gone in and compared the macro output to the macro output in the original 0.18 line by line and **they are identical**, so I'm really stumped.  See this example
```julia
quote  # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 284:
    (JuMP.validmodel)(m, :m) # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 285:
    begin  # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 1208:
        begin  # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 231:
            kouprey = (JuMP.JuMPArray)((JuMP.Array){(JuMP.variabletype)(m)}(JuMP.undef, ((JuMP.length)(-5:5),)...), (-5:5,)) # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 232:
            begin  # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 214:
                let  # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 215:
                    begin 
                        local finch
                    end # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 216:
                    for finch = -5:5 # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 217:
                        begin 
                            (JuMP.coloncheck)(finch)
                            kouprey[finch] = (JuMP.constructvariable!)(m, JuMP._error, 0, 10, :Default, JuMP.EMPTYSTRING, NaN)
                        end
                    end
                end
            end # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 233:
            JuMP.nothing
        end # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 1209:
        kouprey isa JuMP.JuMPContainer && (JuMP.pushmeta!)(kouprey, :model, m) # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 1210:
        (JuMP.push!)(m.dictList, kouprey) # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 1211:
        !false && (JuMP.registervar)(m, :x, kouprey) # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 1212:
        (JuMP.storecontainerdata)(m, kouprey, :x, (-5:5,), JuMP.IndexPair[JuMP.IndexPair(nothing, :(-5:5))], $(Expr(:copyast, :($(QuoteNode(:(()))))))) # /home/expandingman/.julia/v0.6/JuMP/src/macros.jl, line 1215:
        x = kouprey
    end
end
```
There don't appear to be any closures or comprehensions that I can see.  Help with this would be hugely appreciated because at this point I'm completely stumped.

Obviously this is very large and will take a long time to review.  As you make your review, please keep in mind the above error and let me know if you have any ideas.

The good news is JuMP 0.18 is now usable in 1.0! (in principle)

### DIFF
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -28,6 +28,12 @@ GenericAffExpr(a::GenericAffExpr) = GenericAffExpr(a.vars, a.coeffs, a.constant)
 
 coeftype(::GenericAffExpr{C,V}) where {C,V} = C
 
+# variables must be ∈ ℝ but coeffs can be ∈ ℂ
+Compat.adjoint(a::GenericAffExpr{<:Real}) = a
+function Compat.adjoint(a::GenericAffExpr)
+    GenericAffExpr(a.vars, [z' for z ∈ a.coeffs], a.constant')
+end
+
 Base.zero(::Type{GenericAffExpr{C,V}}) where {C,V} = GenericAffExpr{C,V}(V[],C[],zero(C))
 Base.one(::Type{GenericAffExpr{C,V}}) where { C,V} = GenericAffExpr{C,V}(V[],C[], one(C))
 Base.zero(a::GenericAffExpr) = zero(typeof(a))
@@ -55,7 +61,7 @@ else
     end
     function Base.iterate(lti::LinearTermIterator, state::Int)
         if state ≤ length(lti.aff.vars)
-            ((lti.affs.coeffs[state], lti.affs.vars[state]), state+1)
+            ((lti.aff.coeffs[state], lti.aff.vars[state]), state+1)
         else
             nothing
         end
@@ -210,7 +216,7 @@ addconstraint(m::Model, c::Array{LinearConstraint}) =
     error("The operators <=, >=, and == can only be used to specify scalar constraints. If you are trying to add a vectorized constraint, use the element-wise dot comparison operators (.<=, .>=, or .==) instead")
 
 function addVectorizedConstraint(m::Model, v::Array{LinearConstraint})
-    ret = Array{LinConstrRef}(size(v))
+    ret = Array{LinConstrRef}(undef, size(v))
     for I in eachindex(v)
         ret[I] = addconstraint(m, v[I])
     end
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -200,7 +200,7 @@ function addconstraint(m::Model, c::LinearConstraint)
             indices, coeffs = merge_duplicates(Cint, c.terms, m.indexedVector, m)
             MathProgBase.addconstr!(m.internalModel,indices,coeffs,c.lb,c.ub)
         else
-            Base.warn_once("Solver does not appear to support adding constraints to an existing model. JuMP's internal model will be discarded.")
+            warn_once("Solver does not appear to support adding constraints to an existing model. JuMP's internal model will be discarded.")
             m.internalModelLoaded = false
         end
     end
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -40,6 +40,8 @@ Base.zero(a::GenericAffExpr) = zero(typeof(a))
 Base.one( a::GenericAffExpr) =  one(typeof(a))
 Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.vars),copy(a.coeffs),copy(a.constant))
 
+Base.convert(::Type{GenericAffExpr{C,V}}, a::GenericAffExpr{C,V}) where {C,V} = a
+
 # Old iterator protocol - iterates over tuples (aᵢ,xᵢ)
 struct LinearTermIterator{GAE<:GenericAffExpr}
     aff::GAE
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -152,12 +152,12 @@ mutable struct NLPEvaluator <: MathProgBase.AbstractNLPEvaluator
                 has_user_mv_operator |= ReverseDiffSparse.has_user_multivariate_operators(nlconstr.terms.nd)
             end
             d.disable_2ndorder = has_user_mv_operator
-            d.user_output_buffer = Array{Float64}(m.nlpdata.largest_user_input_dimension)
-            d.jac_storage = Array{Float64}(max(numVar,m.nlpdata.largest_user_input_dimension))
+            d.user_output_buffer = Array{Float64}(undef, m.nlpdata.largest_user_input_dimension)
+            d.jac_storage = Array{Float64}(undef, max(numVar,m.nlpdata.largest_user_input_dimension))
         else
             d.disable_2ndorder = false
-            d.user_output_buffer = Array{Float64}(0)
-            d.jac_storage = Array{Float64}(numVar)
+            d.user_output_buffer = Array{Float64}(undef, 0)
+            d.jac_storage = Array{Float64}(undef, numVar)
         end
 
         d.eval_f_timer = 0
@@ -207,7 +207,7 @@ function FunctionStorage(nd::Vector{NodeData}, const_values,numVar, coloring_sto
     else
         hess_I = hess_J = Int[]
         rinfo = Coloring.RecoveryInfo()
-        seed_matrix = Array{Float64}(0,0)
+        seed_matrix = Array{Float64}(undef, 0,0)
         linearity = [NONLINEAR]
     end
 
@@ -223,7 +223,7 @@ function SubexpressionStorage(nd::Vector{NodeData}, const_values,numVar, fixed_v
     reverse_storage = zeros(length(nd))
     linearity = classify_linearity(nd, adj, subexpression_linearity, fixed_variables)
 
-    empty_arr = Array{Float64}(0)
+    empty_arr = Array{Float64}(undef, 0)
 
     return SubexpressionStorage(nd, adj, const_values, forward_storage, partials_storage, reverse_storage, empty_arr, empty_arr, empty_arr, linearity[1])
 
@@ -256,7 +256,7 @@ function MathProgBase.initialize(d::NLPEvaluator, requested_features::Vector{Sym
 
     d.parameter_values = nldata.nlparamvalues
 
-    tic()
+    #tic()
 
     d.linobj = prepAffObjective(d.m)
     linrowlb, linrowub = prepConstrBounds(d.m)
@@ -268,8 +268,8 @@ function MathProgBase.initialize(d::NLPEvaluator, requested_features::Vector{Sym
 
     d.has_nlobj = isa(nldata.nlobj, NonlinearExprData)
     max_expr_length = 0
-    main_expressions = Array{Vector{NodeData}}(0)
-    subexpr = Array{Vector{NodeData}}(0)
+    main_expressions = Array{Vector{NodeData}}(undef, 0)
+    subexpr = Array{Vector{NodeData}}(undef, 0)
     for nlexpr in nldata.nlexpr
         push!(subexpr, nlexpr.nd)
     end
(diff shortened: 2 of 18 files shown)


ExpandingMan:
Ok, I have fixed an absolutely bizarre `@generated` function related bug on 0.6, think this is it.

All that remains is to get the solver-dependent tests running, but with any luck that should mostly just be a matter of getting those packages updated and tagged.

Once we go through review and make changes, if you approve let's merge and finish solvers on a separate PR.

src/affexpr.jl
```diff
@@ -28,12 +28,20 @@ GenericAffExpr(a::GenericAffExpr) = GenericAffExpr(a.vars, a.coeffs, a.constant)
 
 coeftype(::GenericAffExpr{C,V}) where {C,V} = C
 
+# variables must be ∈ ℝ but coeffs can be ∈ ℂ
+Compat.adjoint(a::GenericAffExpr{<:Real}) = a
+function Compat.adjoint(a::GenericAffExpr)
+    GenericAffExpr(a.vars, [z' for z ∈ a.coeffs], a.constant')
```
blegat: Could you use `adjoint. (a.coeffs)` and `adjoint (constant)` ? It is more readable IMO as it does not require to remember that ' is not transpose but adjoint

src/quadexpr.jl
```diff
@@ -30,13 +30,21 @@ GenericQuadExpr(q::GenericQuadExpr) = GenericQuadExpr(q.qvars1, q.qvars2, q.qcoe
 
 coeftype(::GenericQuadExpr{C,V}) where {C,V} = C
 
+# variables are ∈ ℝ but coeffs can be ∈ ℂ
+Compat.adjoint(q::GenericQuadExpr{<:Real}) = q
+function Compat.adjoint(q::GenericQuadExpr)
+    GenericQuadExpr(q.qvars1, q.qvars2, [z' for z ∈ q.qcoeffs], aff')
```
blegat: Same comment here

src/affexpr.jl
ExpandingMan: That would create an `Adjoint{Vector}` rather than a `Vector`, so no, you can't do that.

ExpandingMan: Actually, yes you can do it with the dot operator, didn't notice your dot.  Ok, I'll do that.

src/operators.jl
```diff
@@ -322,6 +309,25 @@ function _dot(lhs::AbstractArray{T}, rhs::AbstractArray{S}) where {T,S}
     ret
 end
 
+# this is a generic fallback from stdlib LinearAlgebra that's needed for tests to pass on 0.6
+# for example, dot(::Array{Float64,3}, ::Array{Int,3}) is not defined in 0.6
+if VERSION < v"0.7-"
+    function Base.dot(x::AbstractArray, y::AbstractArray)
```
blegat: This is type piracy if we define a method for a Base function, it should be with custom types. Can't we implement this in a custom function like `_dot` ?

ExpandingMan: Sorry, I think I just started getting too desparate to get 0.6 working.

The proper solution would probably just be to ensure in the tests that this the arrays are promoted properly first.  I'm not entirely sure why this wasn't causing a problem in the original 0.18.  Let me look into it a bit.  I don't think it would be unreasonable to change the test to ensure that promotion occurs, what do you think?

ExpandingMan: Looking this I am more and more confident that the original 0.18 must have been commiting type piracy.  Regardless, I think the fixes to the tests will be pretty simple so we'll all be much happier with that.

ExpandingMan:
Ok, sorry that was really silly of me, should be better now.

src/operators.jl
blegat: Which line of the tests is making it fail ?

src/operators.jl
```diff
@@ -440,64 +442,36 @@ function _multiplyt!(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC) where T<:
 end
 
 
-# TODO: implement sparse * sparse code as in base/sparse/linalg.jl (spmatmul)
-_multiply!(ret::AbstractArray{T}, lhs::SparseMatrixCSC, rhs::SparseMatrixCSC) where {T<:JuMPTypes} = _multiply!(ret, lhs, full(rhs))
-_multiplyt!(ret::AbstractArray{T}, lhs::SparseMatrixCSC, rhs::SparseMatrixCSC) where {T<:JuMPTypes} = _multiplyt!(ret, lhs, full(rhs))
+function Base.Matrix(S::SparseMatrixCSC{V}) where V<:Variable
```
blegat: It is needed because of this bug right: https://github.com/JuliaLang/julia/issues/27015 ?

ExpandingMan: I believe that's one of the reasons.  Also remember there's no `dense` in 1.0, so you need a safe way of converting to a dense matrix.

ExpandingMan:
By the way, as you review please bear in mind `src/operators.jl` in the current master.  I realize I changed the 0.18 operators extensively, but I tried hard to stick as close to 0.19 as possible (and stole the code in most cases).

src/operators.jl
blegat: It was added in master to fix https://github.com/JuliaOpt/JuMP.jl/issues/1275. and there is a comment `See https://github.com/JuliaLang/julia/issues/27015` explaining why it is needed on master above the method. I think we should add it here too

test/operator.jl
```diff
@@ -628,9 +634,9 @@ Base.transpose(t::MySumType) = MySumType(t.a)
         @test vec_eq(@JuMP.Expression(A*x/2), A*x/2)
         @test vec_eq(X*v,  [4X11; 6X23; 0])
         @test vec_eq(v'*X,  [4X11  0   5X23])
-        @test vec_eq(v.'*X, [4X11  0   5X23])
+        @test vec_eq(transpose(v)*X, [4X11  0   5X23])
```
blegat: Isn't `.'` equivalent to `transpose` ?

blegat: It seems it has been removed, never mind :)

src/nlp.jl
```diff
@@ -530,7 +530,7 @@ function MathProgBase.eval_grad_f(d::NLPEvaluator, g, x)
 end
 
 function MathProgBase.eval_g(d::NLPEvaluator, g, x)
-    tic()
+    VERSION < v"0.7-" && tic()
```
mlubin: Why are the `tic()`s version protected but not the `toq()`s? Does this code work?

ExpandingMan: I think the `toq`s are in the other packages no?

The easiest thing to do will be to disable them in those packages also in 0.7, the only other option is to re-implement the feature ourselves.

mlubin: What other packages? I'm confused because both `tic` and `toq` are deprecated, and it's an error to run `toq` without a corresponding `tic`.

ExpandingMan:
Hm, something is wrong.  It's not supposed to print that warning more than once.  The new logging functionality has a max warnings feature, and I'm using it.  A bug in the logger maybe?  I'll look into it.

For what it's worth, getting this working would not even have been particularly difficult if it were not for the fact that we had to continue to support 0.6.  It's extremely unfortunate that we were in the position of having to support it when no features will be added for it.  I know the situation JuMP was in here was just a lot of bad luck in timing, but I hope they drop the METADATA description where you can't change language versions in minor package version numbers.  It really caused a lot of grief here (whether or not I was masochistic to take this on in the first place).

src/nlp.jl
ExpandingMan: Hm, I think for some reason I though they were in the solvers or `MathProgBase`.  I'm not sure how it is that they are not being called in tests but `tic` is.  I'll look into it.

IainNZ:
https://github.com/JuliaLang/julia/issues/28786 This maybe?

ExpandingMan:
Yup, probably.

src/nlp.jl
ExpandingMan: This is a bit tricky.  I think the only good option here is to re-implement `tic` `toq` in 1.0.  I don't think we can take this out completely as it looks like some of the `MathProgBase` logic may depend on it.  Errors in tests were probably burried in the 1.0 solver errors.

mlubin:
Huh? There's no logic in MPB that depends on tic/toc. They're used only as
timers within the scope of a single function.

src/nlp.jl
```diff
@@ -125,12 +125,12 @@ mutable struct NLPEvaluator <: MathProgBase.AbstractNLPEvaluator
...
     max_chunk::Int # chunk size for which we've allocated storage
-    # timers
-    eval_f_timer::Float64
-    eval_g_timer::Float64
-    eval_grad_f_timer::Float64
-    eval_jac_g_timer::Float64
-    eval_hesslag_timer::Float64
+    # init flags
+    eval_f_init::Bool
+    eval_g_init::Bool
```
mlubin: Are any of these flags used except for eval_f_init? Please remove them if not.

ExpandingMan: Apparently they all get set, but whether they actually get used for anything I don't know (probably not within JuMP).  Perhaps best to just leave it as this is only 0.18.

ExpandingMan: And when I say "they all get set" I mean there are other functions that set them, not just when they are initialized in the constructors.

ExpandingMan:
Ok, I have:
- Replaced the timers with simple flags (on both 0.6 and 1.0).
- Temporarily commented out the addition warning.  There don't seem to be any good solutions to this while the Julia logger seems to have a bug.  If you want I can put a global `const WARN_FLAGS = Dict(:warning_name=>f::Bool)` to emulate the warn once behavior as a temporary solution.

mlubin:
Please re-enable the warning on 0.6 so we're not introducing a regression. The warning is pretty useful, and there will be people using the 0.18 branch on 0.6 for quite a while.

After that looks good to merge from my end.

mlubin:
We'll do a round of solver tagging soon (after MOI 0.6 is released). Right now it will just create more of a mess of incompatible versions.

PR #1436, opened 2018-08-23
RFC: more docs

odow:
Continuation of my doc-writing quest in order to learn the new JuMP syntax.

The majority of the quickstart is borrowed from https://github.com/JuliaOpt/JuMP.jl/blob/v0.18.2/doc/example.rst

<s>The most contentious change is that I added GLPK to travis so that we have a solver for the doc tests.</s> We now use a mock optimizer.

### DIFF
--- a/docs/src/quickstart.md
+++ b/docs/src/quickstart.md
@@ -39,9 +39,6 @@ DocTestSetup = quote
                                  JuMP.JuMPMOIModel{Float64}(),
                                  evalobjective=false))
 end
-# v0.6 prepends the module name to printed type information, whereas v0.7 does
-# not.
-DocTestFilters = [r"JuMP\.", r"MathOptInterface\."]
 ```
 !!! note
     Your model doesn't have to be called `model` - it's just a name.
--- a/docs/src/quickstart.md
+++ b/docs/src/quickstart.md
@@ -4,24 +4,24 @@ Quick Start Guide
 This quick start guide will introduce the main concepts of JuMP. If you are
 familiar with another modeling language embedded in a high-level language such
 as PuLP (Python) or a solver-specific interface you will find most of this
-familiar, with the exception of *macros*. A deep understanding of macros is not
-essential, but if you would like to know more please see the
-[Julia documentation](http://docs.julialang.org/en/latest/manual/metaprogramming/).
-If you are coming from an AMPL or similar background, you may find some of the
-concepts novel but the general appearance will still be familiar.
+familiar. If you are coming from an AMPL or similar background, you may find
+some of the concepts novel but the general appearance will still be familiar.
 
-There are more complex examples in the [`JuMP/examples/` folder](https://github.com/JuliaOpt/JuMP.jl/tree/master/examples).
+The example in this guide is deliberately kept simple. There are more complex
+examples in the [`JuMP/examples/` folder](https://github.com/JuliaOpt/JuMP.jl/tree/master/examples).
 
 Once JuMP is installed, to use JuMP in your programs, you just need to say:
 ```jldoctest quickstart_example
 julia> using JuMP
 ```
 
-We also need to include a Julia package which provides an appropriate solver. In
-this case, we will use GLPK:
+You also need to include a Julia package which provides an appropriate solver.
+One such solver is `GLPK.Optimizer`, which is provided by the
+[GLPK.jl package](https://github.com/JuliaOpt/GLPK.jl).
 ```julia
 julia> using GLPK
 ```
+See [Installation Guide](@ref) for a list of other solvers you can use.
 
 Models are created with the `Model()` function. The `with_optimizer` syntax is
 used to specify the optimizer to be used:
@@ -39,12 +39,12 @@ DocTestSetup = quote
                                  JuMP.JuMPMOIModel{Float64}(),
                                  evalobjective=false))
 end
-# v0.6 prepends JuMP. to printed type information, whereas v0.7 does not.
-DocTestFilters = r"JuMP\."
+# v0.6 prepends the module name to printed type information, whereas v0.7 does
+# not.
+DocTestFilters = [r"JuMP\.", r"MathOptInterface\."]
 ```
 !!! note
-    Your model doesn't have to be called `model` - it's just a name. However,
-    the JuMP style guide prefers `model`.
+    Your model doesn't have to be called `model` - it's just a name.
 
 There are a few options for defining a variable, depending on whether you want
 to have lower bounds, upper bounds, both bounds, or even no bounds. The
@@ -69,9 +69,9 @@ objective we are setting! The objective sense, `Max` or `Min`, should be
 provided as the second argument. Note also that we don't have a multiplication
 `*` symbol between `5` and our variable `x` - Julia is smart enough to not need
 it! Feel free to stick with `*` if it makes you feel more comfortable, as we
-have done with `3*y`:
+have done with `3 * y`:
 ```jldoctest quickstart_example
-julia> @objective(model, Max, 5x + 3*y)
+julia> @objective(model, Max, 5x + 3 * y)
 ```
 
 Adding constraints is a lot like setting the objective. Here we create a
@@ -118,7 +118,7 @@ Success::MathOptInterface.TerminationStatusCode = 0
 ```
 In this case, `GLPK` returned `Success`. This does not mean that it has found
 the optimal solution. Instead, it indicates that GLPK has finished running and
-did not encounter any errors or termination limits.
+did not encounter any errors or user-provided termination limits.
 
 ```@meta
 DocTestSetup = nothing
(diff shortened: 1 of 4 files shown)


blegat:
The output of the doctests on Travis seem to indicate failures:
```julia
=====[Test Error]==============================
> Location: src/quickstart.md:63-65
> Code block:
```jldoctest quickstart_example
julia> con = @constraint(model, 1x + 5y <= 3)
x + 5 y <= 3.0
```
> Subexpression:
    con = @constraint(model, 1x + 5y <= 3)
> Output Diff (REQUIRES COLOR):
x + 5 y <= ≤ 3.0
=====[End Error]===============================
=====[Test Error]==============================
> Location: src/quickstart.md:128-139
> Code block:
```jldoctest quickstart_example
julia> x_upper = JuMP.UpperBoundRef(x)
x <= 2.0
julia> JuMP.resultdual(x_upper)
-4.4
julia> y_lower = JuMP.LowerBoundRef(y)
y >= 0.0
julia> JuMP.resultdual(y_lower)
0.0
```
> Subexpression:
    x_upper = JuMP.UpperBoundRef(x)
> Output Diff (REQUIRES COLOR):
x <= ≤ 2.0
=====[End Error]===============================
=====[Test Error]==============================
> Location: src/quickstart.md:128-139
> Code block:
```jldoctest quickstart_example
julia> x_upper = JuMP.UpperBoundRef(x)
x <= 2.0
julia> JuMP.resultdual(x_upper)
-4.4
julia> y_lower = JuMP.LowerBoundRef(y)
y >= 0.0
julia> JuMP.resultdual(y_lower)
0.0
```
> Subexpression:
    y_lower = JuMP.LowerBoundRef(y)
> Output Diff (REQUIRES COLOR):
y >= ≥ 0.0
=====[End Error]===============================
```

Wouldn't it be preferable to put the `using GLPK` outside the `jldoctest` and build a MockOptimizer instead in `setup` and feed it with the expected outputs ?

**Comment by odow** [comment]
Urgh, this is some platform issue. For me, the docs build locally on v0.6 without errors.

File: docs/src/variables.md [hunk 1]
```diff
@@ -130,9 +135,9 @@ In the above examples, `x_free` represents an unbounded optimization variable,
     `@variable(model, a <= x)`) will result in an error.
```
blegat: `a=1` above should be `a = 1` according to  the style guide

docs/src/quickstart.md
```diff
@@ -1,4 +1,182 @@
 Quick Start Guide
 =================
 
-TODO: Quick example of solving an LP and getting the solution back.
+This quick start guide will introduce the main concepts of JuMP. If you are
+familiar with another modeling language embedded in a high-level language such
+as PuLP (Python) or a solver-specific interface you will find most of this
+familiar, with the exception of *macros*. A deep understanding of macros is not
+essential, but if you would like to know more please see the
```
blegat: Reading this, it seems to me that it would be a good idea to first understanding macros while this really  not a good idea to learn metaprogramming to use JuMP. This might discourage many users. I don't know if essential is the weak enough, maybe "needed" or "required" ? Not sure :)

docs/src/quickstart.md
```diff
@@ -1,4 +1,182 @@
...
+    model = Model(with_optimizer(MOI.Utilities.MockOptimizer,
+                                 JuMP.JuMPMOIModel{Float64}(),
+                                 evalobjective=false))
+end
+# v0.6 prepends JuMP. to printed type information, whereas v0.7 does not.
+DocTestFilters = r"JuMP\."
+```
+!!! note
+    Your model doesn't have to be called `model` - it's just a name. However,
+    the JuMP style guide prefers `model`.
```
blegat: Maybe add add href to the style guide here

docs/src/quickstart.md
```diff
@@ -1,4 +1,182 @@
...
+    const MOI = JuMP.MathOptInterface
+    model = Model(with_optimizer(MOI.Utilities.MockOptimizer,
+                                 JuMP.JuMPMOIModel{Float64}(),
+                                 evalobjective=false))
+end
+# v0.6 prepends JuMP. to printed type information, whereas v0.7 does not.
+DocTestFilters = r"JuMP\."
+```
+!!! note
+    Your model doesn't have to be called `model` - it's just a name. However,
```
blegat: doesn't -> does not

docs/src/quickstart.md
```diff
@@ -1,4 +1,182 @@
...
+julia> @variable(model, 0 <= y <= 30)
+y
+```
+See the [Variables](@ref) section for more information on creating variables.
+
+```@meta
+DocTestSetup = nothing
+```
+
+Next we'll set our objective. Note again the `model`, so we know which model's
```
blegat: we'll -> we will

docs/src/quickstart.md
blegat: Not sure if we want to allow shortening or not in fact. If you think it is more friendly then I am ok to leave it :)

docs/src/quickstart.md
odow: So I started reading the Google documentation style guide. They recommend using contractions: https://developers.google.com/style/contractions

blegat: Ok let's follow Google's style guideline for contraction

docs/src/variables.md
```diff
@@ -470,7 +510,28 @@ Dict{Symbol,Symmetric{JuMP.VariableRef,Array{JuMP.VariableRef,2}}} with 2 entrie
...
-**TODO(@odow):** explain how to delete variables.
+Finally, JuMP supports the deletion of optimization variables. However, before
+we explain how to do so, it is useful to add a short alias for
+`MathOptInterface`:
+```jldoctest variables_delete; setup=:(model=Model())
+julia> const MOI = JuMP.MathOptInterface
+MathOptInterface
+```
+
+To delete variables, we can use the `MOI.delete!` method. We can also check
```
mlubin: Let's implement `JuMP.delete!` and `JuMP.is_valid` instead of documenting the MOI way.

odow:
Doctests pass on v0.6. Annoyingly, they fail on v0.7 because it does not print the module that defines the type. For example, v0.6 will print `JuMP.VariableRef` whereas v0.7 prints `VariableRef`. However, Documenter only has the ability to filter output that appears in both the docs and actual output.

mlubin:
Could we run the doctests on only one version of Julia? Might as well
switch to 1.0 unless there's something blocking it.

docs/src/quickstart.md
```diff
@@ -1,4 +1,180 @@
...
+objective we are setting! The objective sense, `Max` or `Min`, should be
+provided as the second argument. Note also that we don't have a multiplication
+`*` symbol between `5` and our variable `x` - Julia is smart enough to not need
+it! Feel free to stick with `*` if it makes you feel more comfortable, as we
+have done with `3 * y`:
+```jldoctest quickstart_example
+julia> @objective(model, Max, 5x + 3 * y)
```
mlubin: Maybe point out that this example is intentionally inconsistent, but it's good practice to use one way or the other consistently in your code.

docs/src/quickstart.md
```diff
@@ -1,4 +1,180 @@
...
+```
+
+Adding constraints is a lot like setting the objective. Here we create a
+less-than-or-equal-to constraint using `<=`, but we can also create equality
+constraints using `==` and greater-than-or-equal-to constraints with `>=`:
+```jldoctest quickstart_example; filter=r"≤|<="
+julia> con = @constraint(model, 1x + 5y <= 3)
```
mlubin: Why use the anonymous constraint syntax here and the named variable syntax above? (I'd recommend using the named constraint syntax for consistency.)

docs/src/quickstart.md
```diff
@@ -1,4 +1,180 @@
...
+optimizer stopped. This can be for a number of reasons. First, the solver might
+have found the optimal solution, or proved that the problem is infeasible.
+However, it might also have run into numerical difficulties, or terminated due
+to a setting such as a time limit. We can ask the solver why it stopped using
+the `JuMP.termination_status` function:
+```jldoctest quickstart_example
+julia> JuMP.termination_status(model)
+Success::TerminationStatusCode = 0
+```
+In this case, `GLPK` returned `Success`. This does not mean that it has found
```
mlubin: Emphasis on *not*

docs/src/variables.md
```diff
@@ -44,6 +44,10 @@ This code does three things:
 To reduce confusion, we will attempt, where possible, to always refer to
 variables with their corresponding prefix.
 
+!!! warn
+    If you create two JuMP variables with the same name, an error will be
```
mlubin: More directly: "Creating two JuMP variables with the same name results in an error at runtime."

docs/src/variables.md
```diff
@@ -65,8 +69,8 @@ julia> y
 decision variable
 ```
 
-Because `y` is a Julia variable, I can bind it to a different value. For example,
-if I go:
+Because `y` is a Julia variable, we can bind it to a different value. For
+example, if we go:
```
mlubin: "if we go" sounds a bit too informal and might be confusing for non-native speakers. "if we write" instead?

odow:
> Sorry for the slow review... was on vacation.

We're not paying you enough to need to apologize...

PR #1439, opened 2018-08-25
Implement model copy

blegat:
This PR introduces a `JuMP.copy` function that copies the model. Note that it returns both a JuMP model and a reference map so it is not consistent with `Base.copy` which is the why it is a different function.
I needed to prefix all use of `Base.copy` in the JuMP code by `Base` otherwise it would refer to `JuMP.copy`.

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1381
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1300

### DIFF
--- a/src/copy.jl
+++ b/src/copy.jl
@@ -70,7 +70,9 @@ to a given `model`'s reference.
 
 Model copy is not supported in Direct mode, i.e. when a model is constructed
 using the [`direct_model`](@ref) constructor instead of the [`Model`](@ref)
-constructor.
+constructor. Moreover, independently on whether an optimizer was provided at
+model construction, the new model will have no optimizer, i.e., an optimizer
+will have to be provided to the new model in the [`optimize!`](@ref) call.
 
 ## Examples
 
@@ -151,7 +153,9 @@ and its copy.
 
 Model copy is not supported in Direct mode, i.e. when a model is constructed
 using the [`direct_model`](@ref) constructor instead of the [`Model`](@ref)
-constructor.
+constructor. Moreover, independently on whether an optimizer was provided at
+model construction, the new model will have no optimizer, i.e., an optimizer
+will have to be provided to the new model in the [`optimize!`](@ref) call.
 
 ## Examples
 
--- a/src/copy.jl
+++ b/src/copy.jl
@@ -13,20 +13,22 @@ extension storing data in the `ext` field.
 function copy_extension_data end
 
 """
-    copy_variablewise_constraints(dest::Dict{MOIVAR,
-                                             MOICON{MOI.SingleVariable, S}},
-                                  src::Dict{MOIVAR,
-                                            MOICON{MOI.SingleVariable, S}},
-                                  index_map) where S
+    copy_single_variable_constraints(dest::Dict{MOI.VariableIndex,
+                                                MOICON{MOI.SingleVariable, S}},
+                                     src::Dict{MOI.VariableIndex,
+                                               MOICON{MOI.SingleVariable, S}},
+                                     index_map) where S
 
-Copy the variablewise constraint indices of `src` into `dest` mapping variable
-and constraint indices using `index_map`.
+Copy the single variable constraint indices of `src` into `dest` mapping
+variable and constraint indices using `index_map`.
 """
-function copy_variablewise_constraints(dest::Dict{MOIVAR,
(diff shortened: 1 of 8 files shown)


src/copy.jl
```diff
@@ -0,0 +1,128 @@
...
+                                  model.variabletofix, index_map)
+    copy_variablewise_constraints(new_model.variabletointegrality,
+                                  model.variabletointegrality, index_map)
+    copy_variablewise_constraints(new_model.variabletozeroone,
+                                  model.variabletozeroone, index_map)
+
+    new_model.optimizehook = model.optimizehook
+
+    # TODO copy NLP data
+    @assert model.nlpdata === nothing
```
mlubin: Prefer a useful error message over `@assert`.

src/copy.jl
```diff
@@ -0,0 +1,128 @@
...
+
+In the following example, a model `model` is constructed with a variable `x` and
+a constraint `cref`. It is then copied into a model `new_model` with the new
+references assigned to `x_new` and `cref_new`.
+```julia
+model = Model()
+@variable(model, x)
+@constraint(model, cref, x == 2)
+
+new_model, reference_map = JuMP.copy(model)
```
mlubin: What do you think about not returning a `reference_map` and instead telling users to look up the variables by name or symbol? It would be nice to be able to match the signature of `Base.copy` to avoid prefixing all of its uses.  Or maybe have `Base.copy` return only the model and `JuMP.copy_model` return both the model and the reference_map?

blegat: So if the user use `Base.copy` then he is forced to use the `objdict` and if he uses `JuMP.copy_model` then he can use both the reference map and the `objdict`. Is that what you suggest ?

mlubin: Yes (except you can use both lookup by symbol and lookup by name) in either case.

src/copy.jl
```diff
@@ -0,0 +1,172 @@
...
+    copy_variablewise_constraints(dest::Dict{MOIVAR,
+                                             MOICON{MOI.SingleVariable, S}},
+                                  src::Dict{MOIVAR,
+                                            MOICON{MOI.SingleVariable, S}},
+                                  index_map) where S
+
+Copy the variablewise constraint indices of `src` into `dest` mapping variable
+and constraint indices using `index_map`.
+"""
+function copy_variablewise_constraints(dest::Dict{MOIVAR,
```
mlubin: `copy_single_variable_constraints`? "variablewise" isn't a concept in MOI anymore.

src/copy.jl
```diff
@@ -0,0 +1,172 @@
...
+    copy_extension_data(data, new_model::AbstractModel, model::AbstractModel)
+
+Return a copy of the extension data `data` of the model `model` to the extension
+data of the new model `new_model`. A method should be added for any JuMP
+extension storing data in the `ext` field.
+"""
+function copy_extension_data end
+
+"""
+    copy_variablewise_constraints(dest::Dict{MOIVAR,
```
mlubin: There's enough room for `MOI.VariableIndex`. The abbreviation to `MOIVAR` doesn't help readability.

src/copy.jl
```diff
@@ -0,0 +1,172 @@
...
+"""
+    copy_model(model::Model)
+
+Return a copy of the model `model` and a [`ReferenceMap`](@ref) that can be used
+to obtain the variable and constraint reference of the new model corresponding
+to a given `model`'s reference.
+
+## Note
+
+Model copy is not supported in Direct mode, i.e. when a model is constructed
```
mlubin: Does it also work in automatic or manual mode when a solver is attached?

src/copy.jl
```diff
@@ -0,0 +1,172 @@
...
+using the [`direct_model`](@ref) constructor instead of the [`Model`](@ref)
+constructor.
```
mlubin: Add a reference to the `Base.copy` version also.

src/macros.jl
```diff
@@ -15,7 +15,7 @@ end
 include("parseexpr.jl")
 
 function buildrefsets(expr::Expr, cname)
-    c = copy(expr)
+    c = Base.copy(expr)
```
mlubin: Are these changes still needed since we're not defining `JuMP.copy` anymore?

PR #1441, opened 2018-08-26
Move isequal_canonical definition out of test, fix TODO.

IainNZ:
This is seemingly only used by test/operator.jl, but I think the TODO is
correct and moving this with other methods for AbstractJS is better.


src/JuMP.jl
```diff
@@ -372,6 +372,12 @@ else
 end
 Base.isempty(::AbstractJuMPScalar) = false
 
+# Check if two arrays of AbstractJuMPScalars are equal. Useful for testing.
+function isequal_canonical(x::AbstractArray{<:JuMP.AbstractJuMPScalar},
+                           y::AbstractArray{<:JuMP.AbstractJuMPScalar})
+    size(x) == size(y) && all(JuMP.isequal_canonical.(x, y))
```
mlubin: Style nit: use `return`.

PR #1442, opened 2018-08-26
rename user-facing functions for style compliance

### LINKED ISSUE #1382: `numnlconstr` is a bad name.
Rename it.

mlubin:
Mostly focused on user-facing functions to break the important parts first. Includes `optimize` -> `optimize!` for consistency with MOI.

Closes #1382


blegat:
I agree with `optimize` -> `optimize!`. The fact that it does not remove the solution but modifies its internals to contain the solution also apply here

PR #1443, opened 2018-08-27
Speed up test/operator.jl, fix bug on Julia 1.0 with Base.warn_one.

IainNZ:
Operator warning used Base.warn_once, which doesn't exist on Julia 1.0.
Test was unnecessarily expensive to basically just check one function wasn't
called too many times

Comment out all other includes in test/runtests.jl, and in test/operator.jl
comment out extension test. Saving is probably 20 seconds in full test.

```
Before
real	2m6.533s
user	2m5.312s
sys	0m1.349s

After
real	1m55.578s
user	1m54.548s
sys	0m1.176s
```

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -724,14 +724,16 @@ function operator_warn(::AbstractModel) end
 function operator_warn(model::Model)
     model.operator_counter += 1
     if model.operator_counter > 20000
-        Base.warn_once(
+        Compat.@warn(
             "The addition operator has been used on JuMP expressions a large " *
             "number of times. This warning is safe to ignore but may " *
             "indicate that model generation is slower than necessary. For " *
             "performance reasons, you should not add expressions in a loop. " *
             "Instead of x += y, use add_to_expression!(x,y) to modify x in " *
             "place. If y is a single variable, you may also use " *
-            "add_to_expression!(x, coef, y) for x += coef*y.")
+            "add_to_expression!(x, coef, y) for x += coef*y.", maxlog=1)
+        # NOTE: On Julia 1.0 (at least), maxlog=1 does not work correctly.
+        # See https://github.com/JuliaLang/julia/issues/28786.
     end
 end
 
--- a/test/operator.jl
+++ b/test/operator.jl
@@ -319,21 +319,24 @@ function operators_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
             end
 
             if ModelType <: Model
-                # Only `Model` guaranteed to have `operator_counter`, so
+                # Only `Model` is guaranteed to have `operator_counter`, so
                 # only test for that case.
-                @testset "JuMP PR #943" begin
-                    pull943 = ModelType()
-                    @test pull943.operator_counter == 0
-                    @variable(pull943, x[1:100])
+                @testset "dot not using inefficient addition" begin
+                    # Check that dot is not falling back to default, inefficient
+                    # addition (JuMP PR #943).
+                    model = ModelType()
+                    @test model.operator_counter == 0
+                    @variable(model, x[1:100])
                     JuMP.set_start_value.(x, 1:100)
-                    @expression(pull943, testsum, sum(x[i] * i for i in 1:100))
-                    @expression(pull943, testdot1, dot(x, 1:100))
-                    @expression(pull943, testdot2, dot(1:100, x))
-                    @test pull943.operator_counter == 0
-                    testadd = testdot1 + testdot2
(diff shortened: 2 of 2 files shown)


src/JuMP.jl
```diff
@@ -702,15 +712,15 @@ a warning is generated once.
...
+    if model.operator_counter > OPERATOR_WARN_LIMIT
+        Compat.@warn(
             "The addition operator has been used on JuMP expressions a large " *
             "number of times. This warning is safe to ignore but may " *
             "indicate that model generation is slower than necessary. For " *
             "performance reasons, you should not add expressions in a loop. " *
             "Instead of x += y, use add_to_expression!(x,y) to modify x in " *
             "place. If y is a single variable, you may also use " *
-            "add_to_expression!(x, coef, y) for x += coef*y.")
+            "add_to_expression!(x, coef, y) for x += coef*y.", maxlog=1)
```
mlubin: Maybe add a comment linking to https://github.com/JuliaLang/julia/issues/28786?

src/JuMP.jl
```diff
@@ -702,15 +712,15 @@ a warning is generated once.
 function operator_warn(::AbstractModel) end
 function operator_warn(model::Model)
     model.operator_counter += 1
-    if model.operator_counter > 20000
-        Base.warn_once(
+    if model.operator_counter > OPERATOR_WARN_LIMIT
```
mlubin: This introduces a type inference issue. Maybe `OPERATOR_WARN_LIMIT::Int` here and `function set_operator_warn_limit(new_limit::Int)` above?

IainNZ: Is it because OPERATOR_WARN_LIMIT is a non-const module level global?
It only works on JuMP.Model-s anyway - could make a Model field?

IainNZ: Would have the benefit of making clear the global-ish behaviour that this has already.

IainNZ: In fact, later in this file (test/operators.jl), something similar to that happens!
I'm going to integrate the princple of this change into that, and drop the global.

src/JuMP.jl
IainNZ: Done.

IainNZ:
Simplified the change a lot, ready for review (as long as tests pass on CI!)

test/operator.jl
```diff
@@ -318,15 +318,23 @@ function operators_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
...
-                pull943 = ModelType()
-                @variable(pull943, x[1 : 10^6]);
-                JuMP.set_start_value.(x, 1 : 10^6)
-                @expression(pull943, testsum, sum(x[i] * i for i = 1 : 10^6))
-                @expression(pull943, testdot1, dot(x, 1 : 10^6))
-                @expression(pull943, testdot2, dot(1 : 10^6, x))
-                @test JuMP.value(testsum, JuMP.start_value) ≈ JuMP.value(testdot1, JuMP.start_value)
-                @test JuMP.value(testsum, JuMP.start_value) ≈ JuMP.value(testdot2, JuMP.start_value)
+            if ModelType <: Model
+                # Only `Model` guaranteed to have `operator_counter`, so
```
mlubin: Nit (complete sentences): Only `Model` is guaranteed ...

test/operator.jl
```diff
@@ -318,15 +318,23 @@ function operators_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
...
+                # only test for that case.
+                @testset "JuMP PR #943" begin
```
mlubin: Consider changing the name of the testset to something that doesn't require looking up a PR. If the PR number is really needed it could be a comment instead.

test/operator.jl
```diff
@@ -318,15 +318,23 @@ function operators_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
...
+                    pull943 = ModelType()
```
mlubin: `model` seems like it would be a better name than `pull943`.

test/operator.jl
```diff
@@ -318,15 +318,23 @@ function operators_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
...
+                    @test pull943.operator_counter == 0
+                    @variable(pull943, x[1:100])
+                    JuMP.set_start_value.(x, 1:100)
+                    @expression(pull943, testsum, sum(x[i] * i for i in 1:100))
```
mlubin: Style: `test_sum`, `test_dot_1` etc

IainNZ:
Updated

PR #1445, opened 2018-08-27
Expose JuMP.delete and JuMP.is_valid methods

odow:
As [suggested by Miles](https://github.com/JuliaOpt/JuMP.jl/pull/1436#discussion_r212835763). 

I have prefixed `JuMP.delete` in the code to disambiguate `Base.delete!` as it was slightly confusing distinguishing between `delete` and `delete!`. 

Maybe this is a good candidate to break from the style guide and extend a Base method?

<s>There will be some conflicts with https://github.com/JuliaOpt/JuMP.jl/pull/1442</s>

### DIFF
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -448,10 +448,10 @@ julia> x = @variable(model, [i=1:2], basename="x", lower_bound=i, integer=true)
 
 ## User-defined containers
 
-Finally, in the section [Variable containers](@ref), we explained how JuMP
-supports the efficient creation of collections of JuMP variables in three types
-of containers. However, users are also free to create collections of JuMP
-variables in their own datastructures. For example, the following code creates a
+In the section [Variable containers](@ref), we explained how JuMP supports the
+efficient creation of collections of JuMP variables in three types of
+containers. However, users are also free to create collections of JuMP variables
+in their own datastructures. For example, the following code creates a
 dictionary with symmetric matrices as the values:
 ```jldoctest; setup=:(model=Model())
 julia> variables = Dict{Symbol, Symmetric{JuMP.VariableRef,
@@ -470,9 +470,9 @@ Dict{Symbol,Symmetric{JuMP.VariableRef,Array{JuMP.VariableRef,2}}} with 2 entrie
 
 ## Deleting variables
 
-Finally, JuMP supports the deletion of optimization variables. To delete
-variables, we can use the `JuMP.delete` method. We can also check whether `x`
-is a valid JuMP variable in `model` using the `JuMP.is_valid` method:
+JuMP supports the deletion of optimization variables. To delete variables, we
+can use the `JuMP.delete` method. We can also check whether `x` is a valid JuMP
+variable in `model` using the `JuMP.is_valid` method:
 ```jldoctest variables_delete
 julia> @variable(model, x)
 x
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -476,9 +476,12 @@ is a valid JuMP variable in `model` using the `JuMP.is_valid` method:
 ```jldoctest variables_delete
 julia> @variable(model, x)
 x
+
 julia> JuMP.is_valid(model, x)
 true
+
 julia> JuMP.delete(model, x)
+
 julia> JuMP.is_valid(model, x)
 false
 ```
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -470,7 +470,18 @@ Dict{Symbol,Symmetric{JuMP.VariableRef,Array{JuMP.VariableRef,2}}} with 2 entrie
 
 ## Deleting variables
 
-**TODO(@odow):** explain how to delete variables.
+Finally, JuMP supports the deletion of optimization variables. To delete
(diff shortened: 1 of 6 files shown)


docs/src/variables.md
```diff
@@ -470,7 +470,21 @@ Dict{Symbol,Symmetric{JuMP.VariableRef,Array{JuMP.VariableRef,2}}} with 2 entrie
 
 ## Deleting variables
 
-**TODO(@odow):** explain how to delete variables.
+Finally, JuMP supports the deletion of optimization variables. To delete
+variables, we can use the `JuMP.delete` method. We can also check whether `x`
```
mlubin: What happens when the variable that's deleted already appears in constraints?

src/JuMP.jl
```diff
@@ -481,13 +481,26 @@ if VERSION >= v"0.7-"
...
-function MOI.delete!(m::Model, cr::ConstraintRef{Model})
-    @assert m === cr.m
-    MOI.delete!(m.moi_backend, index(cr))
+"""
+    delete(model::Model, constraint_ref::ConstraintRef)
+
+Delete the constraint associated with `constraint_ref` from the model `model`.
+"""
+function delete(model::Model, constraint_ref::ConstraintRef{Model})
+    # TODO: should model be a parameter in the constraint_ref?
```
mlubin: `Model` is already a parameter in the constraint ref. Not sure what this TODO means, maybe just delete?

src/variables.jl
```diff
@@ -111,12 +111,25 @@ end
...
-function MOI.delete!(m::Model, v::VariableRef)
-    @assert m === v.m
-    MOI.delete!(m.moi_backend, v.index)
+"""
+    delete(model::Model, variable_ref::VariableRef)
+
+Delete the variable associated with `variable_ref` from the model `model`.
+"""
+function delete(model::Model, variable_ref::VariableRef)
+    @assert model === variable_ref.m
```
mlubin: Prefer error messages over asserts for validating user-provided arguments.

src/variables.jl
```diff
@@ -300,10 +313,10 @@ end
...
 Delete the lower bound constraint of a variable.
 """
-function delete_lower_bound(v::VariableRef)
-    MOI.delete!(v.m, LowerBoundRef(v))
-    delete!(v.m.variable_to_lower_bound, index(v))
-    nothing
+function delete_lower_bound(variable_ref::VariableRef)
+    JuMP.delete(variable_ref.m, LowerBoundRef(variable_ref))
+    delete!(variable_ref.m.variable_to_lower_bound, index(variable_ref))
+    return nothing
```
mlubin: Nit: `return` instead of `return nothing` (also below).

mlubin:
I see the issue with `JuMP.delete` vs. `Base.delete!`. 
The docstring for `Base.delete!` says:
```
delete!(collection, key)

  Delete the mapping for the given key in a collection, and return the collection.
```
Can we plausibly say that this applies to deleting a variable from a JuMP model? Also `and return the collection` means that we should return the model?

docs/src/variables.md
odow: Urgh. Solver-specific? At least in direct mode. What does the caching optimizer do? @blegat?

odow:
> Can we plausibly say that this applies to deleting a variable from a JuMP model?

Yes, but for consistency, I would prefer that the `JuMP.` prefix is for every user-facing JuMP function. It is a simple rule to remember.

blegat:
The decision we take for JuMP should be the same than for MOI. I don't see any different between the JuMP model/references and MOI model/indices cases. Except that we care more about user-friendliness with JuMP so that might be an argument for `Base.delete!`

odow:
Okay, this diff coverage is now 100% so we're good to merge from my point of view.

mlubin:
We're missing a style guide point on when it's appropriate to add methods to existing functions. I would say that it should be one concept per function, and you shouldn't just add a method because of the name. One way to check this is "can the new method be described by the existing docstring?"

Deleting a variable from a model is arguably not the same as removing an item from a collection (models aren't iterable), so that's an argument for not using `Base.delete!`. I agree we should be consistent between MOI and JuMP now that we have style guidelines.

test/JuMPExtension.jl
```diff
@@ -51,11 +51,15 @@ function JuMP.add_variable(m::MyModel, v::JuMP.AbstractVariable, name::String=""
     JuMP.set_name(vref, name)
     vref
 end
-function MOI.delete!(m::MyModel, vref::MyVariableRef)
-    delete!(m.variables, vref.idx)
-    delete!(m.varnames, vref.idx)
+function JuMP.delete(model::MyModel, variable_ref::MyVariableRef)
+    @assert JuMP.is_valid(model, variable_ref)
```
mlubin: Nit: Error message instead of `assert`.

odow: I think I just made it an assert because it is a test for an extension. Should this be a style thing to never use asserts and always have an informative error message?

blegat: using errors in JuMPExtension is maybe overkill

mlubin: Oops, I got confused. It's fine to use this in a test.

PR #1449, opened 2018-08-30
Make test files standalone, refactor to improve isolation and time, part 1/N

IainNZ:
Goal: Make it so during developing/testing locally it is easy to test what you need, and if you need everything, do it fast.

My assault on test time has two prongs:
* Make the tests themselves faster.
* Make it so the tests can be run in parallel.

Along the way, improve hygiene by making functions for each test set.

Longer term, I'd like to make a new testing library (💯 test libraries)!

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -22,30 +22,11 @@ const MOI = MathOptInterface
 const MOIT = MOI.Test
 const MOIU = MOI.Utilities
 
-macro test_expression(expr)
-    esc(quote
-            @test JuMP.isequal_canonical(@expression(m, $expr), $expr)
-    end)
-end
-
-macro test_expression_with_string(expr, str)
-    esc(quote
-            @test string(@inferred $expr) == $str
-            @test_expression $expr
-    end)
-end
-
-# Test that the macro call `m` throws an error exception during pre-compilation
-macro test_macro_throws(errortype, m)
-    # See https://discourse.julialang.org/t/test-throws-with-macros-after-pr-23533/5878
-    :(@test_throws $errortype try @eval $m catch err; throw(err.error) end)
-end
-
+include("utilities.jl")
 include("JuMPExtension.jl")
 
 include("derivatives.jl")
 include("derivatives_coloring.jl")
-
 include("containers.jl")
 include("model.jl")
 include("variable.jl")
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -56,20 +37,4 @@ include("nlp.jl")
...
-#include("callback.jl");     length( lazy_solvers) == 0 && warn("Callback tests not run!")
-#include("qcqpmodel.jl");    length( quad_solvers) == 0 && warn("Quadratic tests not run!")
-#include("nonlinear.jl");    length(  nlp_solvers) == 0 && warn("Nonlinear tests not run!")
-#                            length(minlp_solvers) == 0 && warn("Mixed-integer Nonlinear tests not run!")
-#include("sdp.jl");          length(  sdp_solvers) == 0 && warn("Semidefinite tests not run!")
-#include("socduals.jl");     length(conic_solvers_with_duals) == 0 && warn("Conic solvers with duals tests not run!")
-
-# hygiene.jl should be run separately
-# hockschittkowski/runhs.jl has additional nonlinear tests
+include("macros.jl")
```
blegat: Which text editor still does not end its files by `\n` :-P

IainNZ: Hah I'm not sure sure how that happened!

test/variable.jl
```diff
@@ -10,425 +10,513 @@
...
 using Compat.Test
 
-function variables_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Type{<:JuMP.AbstractVariableRef})
-    AffExprType = JuMP.GenericAffExpr{Float64, VariableRefType}
-
-    @testset "Constructors" begin
-        mcon = ModelType()
+include("utilities.jl")
+@static if !(:JuMPExtension in names(Main))
+    include("JuMPExtension.jl")
```
mlubin: This is a bit sketchy. I think the root of the issue is trying to make one file both includable and executable at the same time?

IainNZ: Not doing this is fine too (its warning-only), it just seemed like the most non-hacky way to check if a module is already defined. Its basically an include guard, right down to the compile-time-ness.

IainNZ: I'm going to stage new PRs on top of this branch so some time to think about it if you are unsure.

mlubin: Maybe I misunderstood, is the purpose of this include guard just to make it explicit that `variable.jl` depends on `JuMPExtension.jl` or so that the file can be run standalone?

If it's the former, that's not really a pattern that exists in Julia right now. Typically you'd import a module instead?

If it's the latter, how do we test that these test files like `variable.jl` run as standaone files? That's not covered by the regular tests... Wouldn't it be nicer to declare the tests in a file and then have a main `runtests.jl` that decides which subset of tests to run? I guess that leads to your new testing framework :)

IainNZ: The purpose is so that variable.jl can be run alone AND as include-from-runtests.jl

To run alone, each file needs to import this JumpExt code.

As runtests.jl will include every .jl file, we'll be running this importing code many times.

I'm not sure how `import` works for modules that are just defined locally in the same directory, and whether calling `import` loads the module multiple times - I can try that.

Running subsets is the goal, but yeah, I think the first step in that direction is just getting the ability to a least run a file by itself / multiple files in parallel.

mlubin: Ok, so how do we test that the tests can run alone?

IainNZ: `julia1 --project=. test/variable.jl` works (if you have `Project.toml` created)

IainNZ: Oh you want to unit test that tests can be run standalone?
I don't know how to do that for cheap. Certainly possible on Travis but expensive.

IainNZ: Is that important? `runtests.jl` will be the "canonical" way, this to be was for developing JuMP and to be a secondary way.

mlubin: > Oh you want to unit test that tests can be run standalone?

Yes, so that this functionality doesn't rot. If we don't test it, we at least need something like a README.md under test/ describing the structure.

IainNZ: Can I at least get it working first before getting to that? That was going to be last step.

mlubin: Yes :)

PR #1456, opened 2018-09-05
clean up some linear algebra primitives for JuMP eltypes

jrevels:
Did this while working on some of the TODOs floating about this code (e.g. properly intercepting sparse matmul on Julia >v0.7). I can just continue that work within this PR, or open a different PR on top of this one later - whatever's easiest for review.

Let's see if tests pass with what I currently have...

EDIT: Tests now pass locally

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -689,12 +689,12 @@ function Base.:-(x::AbstractArray{T}) where T<:JuMPTypes
     ret
 end
 
-# TODO This will interact poorly with other packages that overload broadcast on Julia >v0.7;
-# we should be using the new broadcast interface instead.
-for (op, opsymbol) in [(+, :+), (-, :-), (*, :*), (/, :/)]
-    @eval begin
-        Base.broadcast(::typeof($op), x::Number, y::JuMPTypes) = $opsymbol(x, y)
-        Base.broadcast(::typeof($op), x::JuMPTypes, y::Number) = $opsymbol(x, y)
+if VERSION < v"0.7-"
+    for (op, opsymbol) in [(+, :+), (-, :-), (*, :*), (/, :/)]
+        @eval begin
+            Base.broadcast(::typeof($op), x::Number, y::JuMPTypes) = $opsymbol(x, y)
+            Base.broadcast(::typeof($op), x::JuMPTypes, y::Number) = $opsymbol(x, y)
+        end
     end
 end
 
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -363,201 +363,163 @@ function Compat.LinearAlgebra.diagm(x::AbstractVector{<:AbstractVariableRef})
     diagm(0=>copyto!(similar(x, GenericAffExpr{Float64,eltype(x)}), x))
 end
 
-###############
-# The _multiply!(buf,y,z) adds the results of y*z into the buffer buf. No bounds/size
-# checks are performed; it is expected that the caller has done this, has ensured
-# that the eltype of buf is appropriate, and has zeroed the elements of buf (if desired).
+###############################################################################
+# Matrix/Vector Arithmetic with JuMP eltypes
+###############################################################################
 
-function _multiply!(ret::Array{T}, lhs::Array, rhs::Array) where T<:JuMPTypes
-    m, n = size(lhs,1), size(lhs,2)
-    r, s = size(rhs,1), size(rhs,2)
-    for i ∈ 1:m, j ∈ 1:s
-        q = ret[i,j]
-        _sizehint_expr!(q, n)
-        for k ∈ 1:n
-            tmp = convert(T, lhs[i,k]*rhs[k,j])
-            add_to_expression!(q, tmp)
-        end
+###############################################################################
+# convenience/utility definitions
+
+const GenericAffOrQuadExpr = Union{GenericAffExpr,GenericQuadExpr}
+
+densify_with_jump_eltype(x::AbstractMatrix) = convert(Matrix, x)
+
+function densify_with_jump_eltype(x::SparseMatrixCSC{V}) where V<:AbstractVariableRef
+    return convert(Matrix{GenericAffExpr{Float64,V}}, x)
+end
+
+# see https://github.com/JuliaLang/julia/pull/18218
+_A_mul_B_eltype(::Type{R}, ::Type{S}) where {R,S} = typeof(one(R) * one(S) + one(R) * one(S))
+
+_A_mul_B_ret_dims(A::AbstractMatrix, B::AbstractVector) = (size(A, 1),)
+_A_mul_B_ret_dims(A::AbstractMatrix, B::AbstractMatrix) = (size(A, 1), size(B, 2))
+
+# don't do size checks here; defer that to `_A_mul_B(A, B)`
+function _A_mul_B_ret(A, B, dims...)
+    T = _A_mul_B_eltype(eltype(A), eltype(B))
+    ret = Array{T}(undef, _A_mul_B_ret_dims(A, B))
+    return fillwithzeros!(ret, T)
+end
+
+if VERSION < v"0.7-"
+    _At_mul_B_ret_dims(A::AbstractMatrix, B::AbstractVector) = (size(A, 2),)
+    _At_mul_B_ret_dims(A::AbstractMatrix, B::AbstractMatrix) = (size(A, 2), size(B, 2))
+    function _At_mul_B_ret(A, B, dims...)
+        T = _A_mul_B_eltype(eltype(A), eltype(B))
+        ret = Array{T}(undef, _At_mul_B_ret_dims(A, B))
+        return fillwithzeros!(ret, T)
     end
-    ret
 end
 
-# this computes lhs.'*rhs and places it in ret
-function _multiplyt!(ret::Array{T}, lhs::Array, rhs::Array) where T<:JuMPTypes
-    m, n = size(lhs,2), size(lhs,1) # transpose
-    r, s = size(rhs,1), size(rhs,2)
-    for i ∈ 1:m, j ∈ 1:s
-        q = ret[i,j]
-        _sizehint_expr!(q, n)
-        for k ∈ 1:n
-            tmp = convert(T, lhs[k,i]*rhs[k,j]) # transpose
-            add_to_expression!(q, tmp)
-        end
-    end
-    ret
-end
-
-function _multiply!(ret::Array{T}, lhs::SparseMatrixCSC, rhs::Array) where T<:Union{GenericAffExpr,GenericQuadExpr}
-    nzv = nonzeros(lhs)
-    rv  = rowvals(lhs)
-    for col ∈ 1:lhs.n
-        for k ∈ 1:size(ret, 2)
-            for j ∈ nzrange(lhs, col)
-                add_to_expression!(ret[rv[j], k], nzv[j] * rhs[col,k])
-            end
-        end
-    end
-    ret
-end
-
-# this computes lhs.'*rhs and places it in ret
-function _multiplyt!(ret::Array{T}, lhs::SparseMatrixCSC, rhs::Array) where T<:Union{GenericAffExpr,GenericQuadExpr}
-    _multiply!(ret, transpose(lhs), rhs) # TODO fully implement
-end
-
-function _multiply!(ret::Array{T}, lhs::Matrix, rhs::SparseMatrixCSC) where T<:Union{GenericAffExpr,GenericQuadExpr}
-    rowval = rowvals(rhs)
-    nzval  = nonzeros(rhs)
-    for multivec_row in 1:size(lhs,1)
-        for col ∈ 1:rhs.n
-            idxset = nzrange(rhs, col)
-            q = ret[multivec_row, col]
-            _sizehint_expr!(q, length(idxset))
-            for k ∈ idxset
-                add_to_expression!(q, lhs[multivec_row, rowval[k]] * nzval[k])
-            end
-        end
-    end
-    ret
-end
-
-# this computes lhs.'*rhs and places it in ret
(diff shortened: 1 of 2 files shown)


test/operator.jl
```diff
@@ -547,9 +547,9 @@ function operators_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
             @test_throws ErrorException A./y
             @test_throws ErrorException B./y
 
-            @test JuMP.isequal_canonical((2 .* x) ./ 3, Matrix((2 .* y) ./ 3))
-            @test JuMP.isequal_canonical(2 .* (x ./ 3), Matrix(2 * (y ./ 3)))
-            @test JuMP.isequal_canonical((x[1,1],) .* A, Matrix((x[1,1],) .* B))
+            @test JuMP.isequal_canonical((2 .* x) ./ 3, JuMP.densify_with_jump_eltype((2 .* y) ./ 3))
```
mlubin: It's better to structure the tests so that they test only one thing. That may be happening here, but it's hard to tell from reading the test. Is it a test of `JuMP._densify_with_jump_eltype` or of operators on sparse objects?

src/operators.jl
```diff
@@ -583,113 +545,123 @@ if VERSION >= v"0.7-"
...
 function Base.:-(x::AbstractArray{T}) where T<:JuMPTypes
     ret = similar(x, typeof(-one(T)))
     for I in eachindex(ret)
         ret[I] = -x[I]
     end
     ret
 end
-Base.:*(x::AbstractArray{T}) where {T<:JuMPTypes} = x
+
+# TODO This will interact poorly with other packages that overload broadcast on Julia >v0.7;
```
mlubin: What breaks if we don't define this code on 0.7+? It's not clear why we need to override this.

src/operators.jl
```diff
@@ -583,113 +545,123 @@ if VERSION >= v"0.7-"
...
 end
 
+Base.:*(A::Number, B::SparseMatrixCSC{T}) where {T<:JuMPTypes} = SparseMatrixCSC(B.m, B.n, copy(B.colptr), copy(B.rowval), A .* B.nzval)
+Base.:*(A::JuMPTypes, B::SparseMatrixCSC) = SparseMatrixCSC(B.m, B.n, copy(B.colptr), copy(B.rowval), A .* B.nzval)
+Base.:*(A::SparseMatrixCSC{T}, B::Number) where {T<:JuMPTypes} = SparseMatrixCSC(A.m, A.n, copy(A.colptr), copy(A.rowval), A.nzval .* B)
+Base.:*(A::SparseMatrixCSC, B::JuMPTypes) = SparseMatrixCSC(A.m, A.n, copy(A.colptr), copy(A.rowval), A.nzval .* B)
+
+Base.:*(x::AbstractArray{T}) where {T<:JuMPTypes} = x
+
+Base.:/(A::SparseMatrixCSC{T}, B::Number) where {T<:JuMPTypes} = SparseMatrixCSC(A.m, A.n, copy(A.colptr), copy(A.rowval), A.nzval ./ B)
```
mlubin: Nit: Use one-line functions only when the function fits on one line (i.e., 80 characters).

src/operators.jl
```diff
@@ -583,113 +545,123 @@ if VERSION >= v"0.7-"
...
-    @eval begin
-        Base.broadcast(::typeof($op),lhs::Number,rhs::JuMPTypes) = $opsymbol(lhs,rhs)
-        Base.broadcast(::typeof($op),lhs::JuMPTypes,rhs::Number) = $opsymbol(lhs,rhs)
-    end
+    Base.:*(A::Transpose{<:JuMPTypes,<:SparseMatrixCSC}, B::Matrix) = _At_mul_B(parent(A), B)
+    Base.:*(A::Transpose{<:Any,<:SparseMatrixCSC}, B::Matrix{<:JuMPTypes}) = _At_mul_B(parent(A), B)
+    Base.:*(A::Transpose{<:JuMPTypes,<:SparseMatrixCSC}, B::Matrix{<:JuMPTypes}) = _At_mul_B(parent(A), B)
 end
 
+Base.:*(A::Number, B::SparseMatrixCSC{T}) where {T<:JuMPTypes} = SparseMatrixCSC(B.m, B.n, copy(B.colptr), copy(B.rowval), A .* B.nzval)
```
mlubin: Why aren't these covered by the built-ins? Because `JuMPType`s aren't `Number`s? This merits a comment.

src/operators.jl
```diff
@@ -363,201 +363,163 @@ function Compat.LinearAlgebra.diagm(x::AbstractVector{<:AbstractVariableRef})
...
-            tmp = convert(T, lhs[k,i]*rhs[k,j]) # transpose
+function fillwithzeros!(A, ::Type{T}) where {T}
+    for I in eachindex(A)
+        A[I] = zero(T)
+    end
+    return A
+end
+
+###############################################################################
+# `_A_mul_B!(ret, A, B)` adds the result of `A*B` into the buffer `ret`. We roll our own
```
mlubin: Does the `A_mul_B` naming scheme still exist in Julia after 0.7? I thought it went away. If that's the case, we should be moving away from it in favor of something more style compliant like `mat_times_mat`.

src/operators.jl
```diff
@@ -363,201 +363,163 @@ function Compat.LinearAlgebra.diagm(x::AbstractVector{<:AbstractVariableRef})
...
+# matmul here (instead of using Julia's generic fallbacks) because doing so allows us to
+# accumulate the expressions for the inner loops in-place. Additionally, Julia's generic
+# fallbacks can be finnicky when your array elements aren't `<:Number`.
+#
+# No bounds/size checks are performed; it is expected that the caller has done this, has
+# ensured that the eltype of `ret` is appropriate, and has zeroed the elements of `ret` (if
+# desired).
+
+function _A_mul_B!(ret::AbstractArray{T}, A, B) where {T<:JuMPTypes}
```
mlubin: Nit: Spaces between binary operators, so `T <: JuMPTypes`.

jrevels:
On the style nits - sorry, I didn't see http://www.juliaopt.org/JuMP.jl/latest/style.html until just now :) I'll get those fixes in soon.

src/operators.jl
```diff
@@ -363,201 +363,163 @@ function Compat.LinearAlgebra.diagm(x::AbstractVector{<:AbstractVariableRef})
...
+###############################################################################
+
+###############################################################################
+# convenience/utility definitions
+
+const GenericAffOrQuadExpr = Union{GenericAffExpr,GenericQuadExpr}
+
+densify_with_jump_eltype(x::AbstractMatrix) = convert(Matrix, x)
+
+function densify_with_jump_eltype(x::SparseMatrixCSC{V}) where V<:AbstractVariableRef
```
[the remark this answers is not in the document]
jrevels: Just because (AFAICT, maybe I missed something) that was what the code was doing before, when it was pirate-overloading `Matrix`. I could change to `AbstractMatrix{<:AbstractVariableRef}` if you want.

src/operators.jl
jrevels: > Does the A_mul_B naming scheme still exist in Julia after 0.7?

No longer in Julia 0.7, but it's an appropriate and familiar naming scheme for this pattern which JuMP itself is still using. We can change the name if you want, of course.

We can more easily be rid of this pattern once we can drop Julia 0.6. As it is now, it's a nice cross-version (i.e. `Transpose`/`Adjoint`-agnostic) way to separate out the compute operations from the interception overloads.

src/operators.jl
jrevels: That's a good question. This code was here before I got here, though :p I'll try to read deeper and figure out an answer

src/operators.jl
jrevels: good point, I'll try wrapping this in a VERSION conditional

test/operator.jl
jrevels: I guess the old test was technically testing both, though I'm not sure of the intention since I didn't write it (just renamed the old pirated constructor). I can factor these out.

src/operators.jl
jrevels: Hmmm. No spaces surrounding `<:` seems to be the norm in the rest of the codebase. Seems best to change them all at once or not all to avoid inconsistent style - should I do that in this PR or leave until later?

jrevels:
Bump. This is ready (pending responses to my previous questions from the initial review).

src/operators.jl
mlubin: We're fixing the style issues as we touch lines of code. It's pretty painful to update the whole code base at once, but you're welcome to if you'd like.

jrevels: This seems like an easy one, I'll just go for it

src/operators.jl
```diff
@@ -337,18 +337,18 @@ end
...
 Base.promote_rule(::Type{<:GenericAffExpr{S, V}}, ::Type{<:GenericQuadExpr{T, V}}) where {S, T, V} = GenericQuadExpr{promote_type(S, T), V}
-Base.promote_rule(::Type{GenericQuadExpr{S, V}}, R::Type{<:Real}                 ) where {S, V}    = GenericQuadExpr{promote_type(S, R), V}
+Base.promote_rule(::Type{GenericQuadExpr{S, V}}, R::Type{<:Real}) where {S, V} = GenericQuadExpr{promote_type(S, R), V}
 
 Base.transpose(x::AbstractJuMPScalar) = x
 
 # Can remove the following code once == overloading is removed
 
-function Compat.LinearAlgebra.issymmetric(x::Matrix{T}) where T<:JuMPTypes
+function Compat.LinearAlgebra.issymmetric(x::Matrix{T}) where {T <: JuMPTypes}
```
blegat: Why use `{` and `}` when it is not a one line function ?

src/operators.jl
```diff
@@ -360,204 +360,168 @@ end
...
-# this computes lhs.'*rhs and places it in ret
-function _multiplyt!(ret::Array{T}, lhs::Array, rhs::Array) where T<:JuMPTypes
-    m, n = size(lhs,2), size(lhs,1) # transpose
-    r, s = size(rhs,1), size(rhs,2)
-    for i ∈ 1:m, j ∈ 1:s
-        q = ret[i,j]
-        _sizehint_expr!(q, n)
-        for k ∈ 1:n
-            tmp = convert(T, lhs[k,i]*rhs[k,j]) # transpose
+function fill_with_zeros!(A, ::Type{T}) where {T}
```
blegat: Isn't `A` an AbstractArray{T} ? Why is the second argument needed ?

src/operators.jl
```diff
@@ -360,204 +360,168 @@ end
...
+# See https://github.com/JuliaLang/julia/pull/18218.
+_A_mul_B_eltype(::Type{R}, ::Type{S}) where {R, S} = typeof(one(R) * one(S) + one(R) * one(S))
+
+_A_mul_B_ret_dims(A::AbstractMatrix, B::AbstractVector) = (size(A, 1),)
+_A_mul_B_ret_dims(A::AbstractMatrix, B::AbstractMatrix) = (size(A, 1), size(B, 2))
+
+# Don't do size checks here; defer that to `_A_mul_B(A, B)`.
+function _A_mul_B_ret(A, B, dims...)
+    T = _A_mul_B_eltype(eltype(A), eltype(B))
+    ret = Array{T}(undef, _A_mul_B_ret_dims(A, B))
```
blegat: Why not do `zeros(T, ...)` instead of `Array...` followed by `fill_with_zeros!` ?

src/operators.jl
jrevels: It's bad practice to dispatch on element type in cases like these; the eltype of `A` and `T` only need be related by a subtype relation, not equality.

src/operators.jl
jrevels: I just carried this pattern over from the old code. It does leave a nice path to generalize to non-`Array` `ret`s later, but that's probably premature. 

I could replace it with `zeros` if we want to make that change.

src/operators.jl
jrevels: stylistic consistency (this is also more readable IMO, but that's subjective)

There probably should just be a rule in the style guide about the desired style here

src/operators.jl
mlubin: `zeros` is broken for JuMP types because `zero` returns a mutable value.

src/operators.jl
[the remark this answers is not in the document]
mlubin: >  "code block comment separator" format for JuMP? 

No, not really. I thought about it briefly but don't have a clear answer. Maybe more consistent use of empty lines would make it easier to separate sections without large horizontal comment bars.

test/operator.jl
```diff
@@ -547,9 +547,12 @@ function operators_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::
             @test_throws ErrorException A./y
             @test_throws ErrorException B./y
 
-            @test JuMP.isequal_canonical((2 .* x) ./ 3, Matrix((2 .* y) ./ 3))
-            @test JuMP.isequal_canonical(2 .* (x ./ 3), Matrix(2 * (y ./ 3)))
-            @test JuMP.isequal_canonical((x[1,1],) .* A, Matrix((x[1,1],) .* B))
+            z = JuMP.densify_with_jump_eltype((2 .* y) ./ 3)
```
mlubin: Using an internal JuMP function to construct the expected answer seems unnecessary and makes it harder to verify that the test is correct.

mlubin: These tests are really poorly structured but I'll let this go for now. Unit tests should test only one thing and not refer to variables like `y` that are defined far away.

jrevels: Yeah - before this PR, these tests' reliance on internal constructors was obscured by the fact that the constructors were pirated overloads of `Matrix`. Now that it's easily visible where this is happening, I've added a TODO to refactor the tests

PR #1472, opened 2018-09-18
Refactor constraints

blegat:
This PR

* moves constraints-related code to a constraints.jl file, moving a lot of code out of JuMP.jl;
* merges SingleVariableConstraint, AffExprConstraint and QuadExprConstraint into ScalarConstraint (same for the vector version) removing a lot of duplicated code both from the definition of the constraint and from the build_constraint function calling them.

As a by-product, it improves support for constraints for which the function is a quadratic expression since the code is now shared with affine expression and variable expression.

### DIFF
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -112,6 +112,11 @@ function value(a::GenericAffExpr{T, V}, map::Function) where {T, V}
     ret
 end
 
+"""
+    constant(aff::GenericAffExpr{C, V})::C
+
+Return the constant of the affine expression.
+"""
 constant(aff::GenericAffExpr) = aff.constant
 
 # Iterator protocol - iterates over tuples (aᵢ,xᵢ)
--- a/src/affexpr.jl
+++ b/src/affexpr.jl
@@ -112,6 +112,8 @@ function value(a::GenericAffExpr{T, V}, map::Function) where {T, V}
     ret
 end
 
+constant(aff::GenericAffExpr) = aff.constant
+
 # Iterator protocol - iterates over tuples (aᵢ,xᵢ)
 struct LinearTermIterator{GAE<:GenericAffExpr}
     aff::GAE
@@ -243,6 +245,7 @@ function MOI.ScalarAffineFunction(a::AffExpr)
     terms = map(t -> MOI.ScalarAffineTerm(t[1], index(t[2])), linear_terms(a))
     return MOI.ScalarAffineFunction(terms, a.constant)
 end
+moi_function(a::GenericAffExpr) = MOI.ScalarAffineFunction(a)
 
 function AffExpr(m::Model, f::MOI.ScalarAffineFunction)
     aff = AffExpr()
@@ -251,6 +254,12 @@ function AffExpr(m::Model, f::MOI.ScalarAffineFunction)
     end
     aff.constant = f.constant
     return aff
+end
+function jump_function(model::AbstractModel, f::MOI.ScalarAffineFunction)
+    return AffExpr(model, f)
+end
+function jump_function(model::AbstractModel, f::MOI.VectorAffineFunction)
+    return map(f -> AffExpr(model, f), MOIU.eachscalar(f))
 end
 
 """
@@ -279,6 +288,7 @@ function MOI.VectorAffineFunction(affs::Vector{AffExpr})
     end
     MOI.VectorAffineFunction(terms, constant)
 end
+moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
 
 function set_objective(m::Model, sense::Symbol, a::AffExpr)
     if sense == :Min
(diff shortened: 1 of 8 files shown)


src/constraints.jl
```diff
@@ -0,0 +1,207 @@
+#############################################################################
```
IainNZ: Can drop this line now, and the standard copyright header.

src/constraints.jl
```diff
@@ -0,0 +1,207 @@
...
+# AbstractShape
```
IainNZ: I think these lines are redundant with the docstring.

src/constraints.jl
```diff
@@ -0,0 +1,207 @@
...
+
+## Examples
+
+Given a [`SymmetricMatrixShape`](@ref) of vectorized form `[1, 2, 3]`, the
+following code retrieve the matrix `Symmetric(Matrix[1 2; 2 3])`:
+```julia
+reshape([1, 2, 3], SymmetricMatrixShape(2))
+```
+"""
+function reshape end
```
IainNZ: Maybe stupid question: Given there is a Base.reshape, does this squash the Base definition in the scope of JuMP from this point on? What does it do to the docstrings?

src/constraints.jl
```diff
@@ -0,0 +1,207 @@
...
+Vector for which the vectorized form corresponds exactly to the vector given.
+"""
+struct VectorShape <: AbstractShape end
+reshape(vectorized_form, ::VectorShape) = vectorized_form
+
+##########################################################################
+# Constraint Reference
+# Holds the index of a constraint in a Model.
+# TODO: Rename "m" field (breaks style guidelines).
+struct ConstraintRef{M <: AbstractModel, C, Shape <: AbstractShape}
```
IainNZ: Style: are we putting spaces around `<:`?

blegat: I thought so as it is a binary operator but we might make an exception for this one

src/constraints.jl
blegat: It defines a JuMP.reshape which is a different function form `Base.reshape` so `reshape` in the JuMP scope and one need to explicitely write `Base.reshape`

src/affexpr.jl
```diff
@@ -112,6 +112,8 @@ function value(a::GenericAffExpr{T, V}, map::Function) where {T, V}
     ret
 end
 
+constant(aff::GenericAffExpr) = aff.constant
```
mlubin: Is this meant to be a public API or internal only? If it's the former it should have a doc string, if it's the latter then it could use a scarier name like `extract_constant` and comment that it's meant for internal use.

src/constraints.jl
```diff
@@ -0,0 +1,207 @@
...
+"""
+    reshape(vectorized_form::Vector, shape::AbstractShape)
+
+Return an object in it original shape `shape` given its vectorized form
+`vectorized_form`.
+
+## Examples
+
+Given a [`SymmetricMatrixShape`](@ref) of vectorized form `[1, 2, 3]`, the
+following code retrieve the matrix `Symmetric(Matrix[1 2; 2 3])`:
```
mlubin: Grammar: `retrieves`. Consider `returns` instead though.

src/constraints.jl
```diff
@@ -0,0 +1,207 @@
...
+dual_shape(shape::PolynomialShape) = MomentsShape(shape.monomials)
+dual_shape(shape::MomentsShape) = PolynomialShape(shape.monomials)
+```
+"""
+dual_shape(shape::AbstractShape) = shape
+
+"""
+    reshape(vectorized_form::Vector, shape::AbstractShape)
+
+Return an object in it original shape `shape` given its vectorized form
```
mlubin: Typo: `in its original shape`.

src/constraints.jl
```diff
@@ -0,0 +1,207 @@
...
+function constraint_object(ref::ConstraintRef{Model, MOICON{FuncType, SetType}}) where
+        {FuncType <: MOI.AbstractVectorFunction, SetType <: MOI.AbstractVectorSet}
+    model = ref.m
+    f = MOI.get(model, MOI.ConstraintFunction(), ref)::FuncType
+    s = MOI.get(model, MOI.ConstraintSet(), ref)::SetType
+    return VectorConstraint(jump_function(model, f), s, ref.shape)
+end
+
+# TODO: Find somewhere to put this error message.
+#add_constraint(m::Model, c::Array{AffExprConstraint}) =
```
mlubin: Do you know what the intention of this code is? I don't know what error case it handles and why `add_constraint` would say something about macro syntax.

src/macros.jl
```diff
@@ -347,28 +347,21 @@ function parse_constraint(_error::Function, args...)
...
-    return QuadExprConstraint(quad, S(MOIU.getconstant(set)-offset))
+function build_constraint(_error::Function,
+                          expr::Union{GenericAffExpr, GenericQuadExpr},
+                          set::MOI.AbstractScalarSet)
+    offset = constant(expr)
+    add_to_expression!(expr, -offset)
+    return ScalarConstraint(expr, MOIU.shift_constant(set, -offset))
 end
-#build_constraint(x::Vector{<:GenericQuadExpr}, set::MOI.AbstractVectorSet) = VectorQuadExprConstraint(x, set)
+build_constraint(_error::Function, α::Number, set::MOI.AbstractScalarSet) = build_constraint(_error, convert(AffExpr, α), set)
```
mlubin: Style nit: Split across lines. Same right below.

src/sd.jl
```diff
@@ -41,43 +41,31 @@ function reshape(vectorized_form::Vector{T}, shape::SquareMatrixShape) where T
                         shape.side_dimension)
 end
 
-# Used by the @variable macro. It can also be used with the @constraint macro,
-# this allows to get the constraint reference, e.g.
-# @variable model x[1:2,1:2] Symmetric # x is Symmetric{VariableRef,Matrix{VariableRef}}
-# varpsd = @constraint model x in PSDCone()
+# * If `eltype(Q)` is `AbstractVariableRef`:
```
mlubin: The formatting of this comment makes it a bit hard to follow. If you're writing pseudo-markdown it may be better to use a docstring instead. Same below.

src/constraints.jl
blegat: We can remove it, it is done in
https://github.com/JuliaOpt/JuMP.jl/blob/672de56de2e7462dfe454a8b1b716eca40b74f19/src/macros.jl#L362

PR #1475, opened 2018-09-19
[release-0.18] Fix isdefined for Julia v1.0

blegat:


ExpandingMan:
Looks like the build process for Ipopt is segfaulting, any idea why that might happen?

Anyway, can't this be simplified a bit?  I think `isdefined(@__MODULE__, :var)` works on all versions, no?  (I actually can't test at the moment as I'm on a new machine and haven't installed 0.6.)

mlubin:
What makes you say that the build for Ipopt is segfaulting? Looks like julia itself is segfaulting: https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/430436305#L7006.

ExpandingMan:
Sorry you're right.  That's scarier.

ExpandingMan:
By the way, I just looked through the documentation.  Looks like `@__MODULE__` is undefined in Julia 0.6 but you can use `Compat.@__MODULE__`.

PR #1481, opened 2018-09-22
update NLP docs for MOI

### LINKED ISSUE #1479: NonlinearParameters still use setvalue and getvalue
Maybe these should stay as is (with renaming for style) because parameters don't have start or result values?

mlubin:
This updates the docs to describe how NLP works with JuMP/MOI. They're pretty rough but now they're at least correct and useful.

Closes #1479 

@ccoffrin @dpo @abelsiqueira

### DIFF
--- a/docs/src/nlp.md
+++ b/docs/src/nlp.md
@@ -1,249 +1,391 @@
 Nonlinear Modeling
 ==================
 
-TODO: This is out of date and has not yet been updated for MOI.
-
-JuMP has support for general smooth nonlinear (convex and nonconvex) optimization problems. JuMP is able to provide exact, sparse second-order derivatives to solvers. This information can improve solver accuracy and performance.
-
-Nonlinear objectives and constraints are specified by using the `@NLobjective` and `@NLconstraint` macros. The familiar `sum()` syntax is supported within these macros, as well as `prod()` which analogously represents the product of the terms within. Note that the `@objective` and `@constraint` macros (and corresponding functions) do *not* currently support nonlinear expressions. However, a model can contain a mix of linear, quadratic, and nonlinear constraints or objective functions. Starting points may be provided by using the `start` keyword argument to `@variable`. If a starting value is not provided for a variable, it will be set to the projection of zero onto the interval defined by the variable bounds. For nonconvex problems, the returned solution is only guaranteed to be locally optimal. Convexity detection is not currently provided.
-
-For example, we can solve the classical Rosenbrock problem (with a twist) as follows:
-
-```julia
-using JuMP
-m = Model()
-@variable(m, x, start = 0.0)
-@variable(m, y, start = 0.0)
-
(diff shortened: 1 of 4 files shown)


docs/src/nlp.md
```diff
@@ -1,249 +1,391 @@
...
-m = Model()
-@variable(m, x, start = 0.0)
-@variable(m, y, start = 0.0)
+using Ipopt
+model = Model(with_optimizer(Ipopt.Optimizer))
+@variable(model, x, start = 0.0)
+@variable(model, y, start = 0.0)
 
-@NLobjective(m, Min, (1-x)^2 + 100(y-x^2)^2)
+@NLobjective(model, Min, (1 - x) ^ 2 + 100 * (y - x ^ 2) ^ 2)
```
blegat: Spacing with `^` seems odd. Doesn't the `100(` still work or do we need `100 * (` ?

src/macros.jl
```diff
@@ -1508,7 +1511,40 @@ macro NLexpression(args...)
     end)
 end
 
-# syntax is @NLparameter(m, p[i=1] == 2i)
+"""
+    @NLparameter(model, param == value)
+
+Creates and returns a nonlinear parameter `param` attached to the model `model`
+with initial value set to `value`. Nonlinear paramters may be used only in
```
blegat: "paramters" -> "parameters"

src/macros.jl
```diff
@@ -1508,7 +1511,40 @@ macro NLexpression(args...)
     end)
 end
 
-# syntax is @NLparameter(m, p[i=1] == 2i)
+"""
+    @NLparameter(model, param == value)
+
+Creates and returns a nonlinear parameter `param` attached to the model `model`
```
blegat: We usually use `Return` in docstrings so for consistency, it should be `Create and return`

src/macros.jl
```diff
@@ -1508,7 +1511,40 @@ macro NLexpression(args...)
...
+@NLparameter(model, x == 10)
+JuMP.value(x)
+
+# output
+10.0
+```
+
+    @NLparameter(model, param_collection[...] == value_expr)
+
+Creates and returns a collection of nonlinear parameters `param_collection`
```
blegat: Same here

src/macros.jl
mlubin: Done

src/macros.jl
mlubin: Done

docs/src/nlp.md
blegat: > Maybe, or maybe we're not used to the binary operator convention. Should be raised as a style question.

here the binary operator is special since it is using as a replacement for putting it in exponent so we expect it to be close, smaller and slightly moved up. We cannot really make it smaller and move it up but at least we want it to be close

For the `100(`, I also though that since it is in the doc we might want to show up that in Julia we are not forced to use `*` in some situations

PR #1483, opened 2018-09-22
Rename `m` field of VariableRef to `model`.

mlubin:
For style compliance. Also switch to `owner_model` method instead of
accessing the field directly.

The definition of VariableRef is now:
```julia
struct VariableRef <: AbstractVariableRef
    model::Model
    index::MOI.VariableIndex
end
```
which is much nicer IMO than
```julia
struct VariableRef <: AbstractVariableRef
    m::Model
    index::MOIVAR
end
```
We should do the same for `ConstraintRef` after this goes through.

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -102,7 +102,7 @@ end
 
 # owner_model should be implemented by all `AbstractVariableRef`.
 """
-    owner_model(v::VariableRef)
+    owner_model(v::AbstractVariableRef)
 
 Returns the model to which `v` belongs.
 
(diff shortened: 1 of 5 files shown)


src/variables.jl
```diff
@@ -96,15 +96,29 @@ variable_ref_type(v::AbstractVariableRef) = typeof(v)
...
-    index::MOIVAR
+    model::Model
+    index::MOI.VariableIndex
 end
 
-# Should be implemented by all `AbstractVariableRef`
-owner_model(v::VariableRef) = v.m
+# owner_model should be implemented by all `AbstractVariableRef`.
+"""
+    owner_model(v::VariableRef)
```
blegat: This could be part of the AbstractVariableRef API

mlubin: What does that mean in practice? Moving this docstring to `function owner_model end`?

blegat: Yes, and using `::AbstractVariableRef`

mlubin: Done

PR #1491, opened 2018-09-23
update install instructions for MOI

mlubin:
Still a bit messy but at least it's more useful than before. I simplified the solver table with a (structured) free text column instead of a column for a fixed set of problem classes.

Closes #1103

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -1,63 +1,99 @@
 Installation Guide
 ==================
 
-DRAFT: Install Julia. Install JuMP.
+JuMP is a package for [Julia](https://julialang.org). To use JuMP, first
+[download and install](https://julialang.org/downloads/) Julia or open up
+a remote notebook at [JuliaBox](https://www.juliabox.com/) or similar services.
+
+JuMP is compatible with both Julia 1.0 and 0.6. The compatibility with 0.6 is
+intended to facilitate upgrading from previous JuMP and Julia releases.
+The following instructions assume Julia 1.0.
+
+From Julia, JuMP is installed by using the built-in package manager:
+```julia
+import Pkg
+Pkg.add("JuMP")
+```
+
+!!! note
+    The installation instructions above assume that JuMP 0.19 has already been
+    released. Until that time, see the JuMP
+    [README](https://github.com/JuliaOpt/JuMP.jl/blob/master/README.md) for
+    instructions on installing a development release that's compatible with
+    Julia 1.0.
+
 
 Getting Solvers
 ---------------
 
-TODO: Not yet updated for MOI.
+JuMP depends on solvers to solve optimization problems. Most solvers are not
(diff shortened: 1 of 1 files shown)


docs/src/installation.md
```diff
@@ -1,63 +1,99 @@
...
 -   MINLP = Mixed-integer nonlinear programming
 -   SDP = Semidefinite programming
+-   MISDP = Mixed-integer semidefinite programming
 
-\* requires CoinOptServices installed, see below.
-
-To install Gurobi, for example, and use it with a JuMP model `m`, run:
+To install Gurobi, for example, and use it with a JuMP model `model`, run:
 
 ```julia
```
blegat: Missing `import Pkg`

mlubin: Done

docs/src/installation.md
```diff
@@ -1,63 +1,100 @@
...
+| [BARON](http://archimedes.cheme.cmu.edu/?q=baron>)                             | [BARON.jl](https://github.com/joehuchette/BARON.jl) | Comm.   | MINLP                       |
+| [Cbc](https://projects.coin-or.org/Cbc)                                        | [Cbc.jl](https://github.com/JuliaOpt/Cbc.jl)        | EPL     | MILP                        |
+| [Clp](https://projects.coin-or.org/Clp)                                        | [Clp.jl](https://github.com/JuliaOpt/Clp.jl)        | EPL     | LP                          |
+| [CPLEX](http://www-01.ibm.com/software/commerce/optimization/cplex-optimizer/) | [CPLEX.jl](https://github.com/JuliaOpt/CPLEX.jl)    | Comm.   | LP, MILP, SOCP, MISOCP      |
+| [CSDP](https://projects.coin-or.org/Csdp/)                                     | [CSDP.jl](https://github.com/JuliaOpt/CSDP.jl)      | EPL     | SDP                         |
+| [ECOS](https://github.com/ifa-ethz/ecos)                                       | [ECOS.jl](https://github.com/JuliaOpt/ECOS.jl)      | GPL     | SOCP                        |
```
blegat: What is the reasoning behind the "LP". ECOS and CSDP can also do LP even if they might not be competitive at it. Does "LP" means "has a simplex solver" ? Or do we mean "should consider using this solver for a pure LP problem" ? Here the heading says "supports".

mlubin: LP should be listed if the solver supports it. This isn't the right place to make judgments on whether the solver is competitive.

blegat: Which solver does not support LP then ?

mlubin: Fixed. The only solver that I didn't say LP for was Cbc because it doesn't return duals so it can't prove that it solved the LP :)

blegat: It makes sense to only call LP for solvers with duals

PR #1492, opened 2018-09-24
[release-0.18] method_exists -> hasmethod

blegat:
Comment out all solvers in REQUIRE. Let's see if a bottom up method works better than a top down.
Let's get green status and uncomment solvers in PRs once they are ready.
It might get easier than having red status and trying to make it green.

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1486

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,11 @@ os:
 julia:
     - 0.6
     - 0.7
+    - 1.0
 matrix:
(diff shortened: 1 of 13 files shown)


blegat:
@ExpandingMan the segfault seems to go away when we disable all solvers even without the ForwardDiff bound

PR #1493, opened 2018-09-24
[release-0.18] Fix undefined reference with sparse matrices

blegat:


ExpandingMan:
I think that ideally we're supposed to use all functions from the `SparseArrays` API such as `findnz`, but I suppose that may not be possible here.  I feel a bit like the fact that we had this problem in the first place is probably an indication that we were using the wrong functions.

blegat:
@ExpandingMan I switching to simply using map

PR #1497, opened 2018-09-26
Fix hygiene

blegat:
The same fix have been applied in https://github.com/JuliaOpt/JuMP.jl/pull/1495/commits/755f58eb2f7b4930bab20c63c166fe195ba5b7c9 for JuMP v0.18
Changes made on this PR will be backported to JuMP v0.18 so review/discuss this change on this PR, not on the JuMP v0.18 PR.

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1496


test/hygiene.jl
```diff
@@ -5,7 +5,6 @@
 
 # hygiene.jl
 # Make sure that our macros have good hygiene
-using Base.Test
 
 module M
```
mlubin: I though the issue occurs when the expressions are at the global scope. Is this testing what we intend when the code is inside a model?

test/runtests.jl
```diff
@@ -37,4 +37,5 @@ include("nlp.jl")
 include("generate_and_solve.jl")
 include("print.jl")
 include("operator.jl")
-include("macros.jl")
\ No newline at end of file
+include("macros.jl")
+include("hygiene.jl")
```
mlubin: The reason hygiene was run separately from the rest of the tests was to make sure it worked when JuMP hadn't already been loaded using `using`. Can we emulate that somehow?

test/hygiene.jl
blegat: According to [Julia docs](https://docs.julialang.org/en/v1/manual/variables-and-scoping/), global scopes are on modules and REPL

mlubin: Ok, LGTM. Could you add a TODO to add actual test statements below? :)

test/runtests.jl
blegat: Maybe with another job like @odow did in https://github.com/JuliaOpt/JuMP.jl/pull/1485 ?

test/hygiene.jl
blegat: What should the TODO say ? That it needs to be in global scope and we use a module for that ?

PR #1517, opened 2018-10-03
Fix hygiene in objective macro

blegat:
Fixes the issue reported in https://discourse.julialang.org/t/undefvarerror-in-jump-constraint/15701/3


mlubin:
Needs a test.

Can we make sure all macros work instead of fixing them one by one?

Does this affect 0.18.3?

blegat:
> Can we make sure all macros work instead of fixing them one by one?

Yes, I am working on a better fix that defines two new functions that should be called at the end of every macros. The fix in getloopedcode wasn't sufficient

> Does this affect 0.18.3?

Yes, we will need to backport this fix

src/macros.jl
```diff
@@ -278,6 +265,49 @@ function assert_validmodel(m, macrocode)
     end
 end
 
+"""
+    macro_return(model, code, variable)
+
+Return a code that runs `code` in a local scope which returns the value of `variable`.
```
mlubin: Nit: Return a block of code that 1. runs the code block `code` in a local scope and 2. returns the value of a local variable named `variable`. `variable` must reference a variable defined by `code`.

src/macros.jl
```diff
@@ -278,6 +265,49 @@ function assert_validmodel(m, macrocode)
...
+    end
+end
+
+"""
+    macro_assign_and_return(model, code, variable, registerfun::Function,
+                            quotvarname, escvarname, final_variable=variable)
+
+Return runs `code` in a local scope which returns the value of `variable`
+and then assign `final_variable` to `escvarname`.
+If `registerfun` is given, it registers the value returned by the local scope
```
mlubin: `registerfun` is a function passed in. How do we know that it "registers the value returned by the local scope"?
The docstring should either provide the expected signature or call a concrete function.

src/macros.jl
```diff
@@ -278,6 +265,49 @@ function assert_validmodel(m, macrocode)
...
+                                 registerfun::Union{Nothing, Function}=nothing,
+                                 model=nothing,
+                                 quotvarname=nothing)
+    macro_code = macro_return(code, variable)
+    return quote
+        $variable = $macro_code
+        $(if registerfun !== nothing
+              :($registerfun($model, $quotvarname, $variable))
+          end)
+        # escvarname should be set in the scope calling the macr
```
mlubin: Typo: "macro"

src/macros.jl
```diff
@@ -278,6 +265,49 @@ function assert_validmodel(m, macrocode)
...
+        end
+    end
+end
+
+"""
+    macro_assign_and_return(model, code, variable, registerfun::Function,
+                            quotvarname, escvarname, final_variable=variable)
+
+Return runs `code` in a local scope which returns the value of `variable`
+and then assign `final_variable` to `escvarname`.
```
mlubin: Could we take the `varname` and `quotvarname` as a single symbol and escape/quote it as needed?

blegat: Indeed, I was able to get rid of `quotvarname` and `escvarname`. It makes things more readable to isolate this technical details in macro_assign_and_return :)

src/macros.jl
```diff
@@ -278,6 +265,49 @@ function assert_validmodel(m, macrocode)
...
+            $code
+            $variable
+        end
+    end
+end
+
+"""
+    macro_assign_and_return(code, variable, name;
+                            final_variable=variable,
+                            registerfun::Union{Nothing, Function}=nothing,
```
mlubin: Style nit: `register_fun`

test/Untitled.ipynb
```diff
@@ -0,0 +1,222 @@
+{
```
mlubin: Hmm?

carlocab:
Hi there, I'm the one who reported the issue at [the discourse page]( https://discourse.julialang.org/t/undefvarerror-in-jump-constraint/15701/3). I can see that this PR has been merged into master, but I'm getting the exact same problems as reported on discourse when I use the master branch.

PR #1526, opened 2018-10-06
Refactor objective sense

### LINKED ISSUE #1459: Add set_objective_sense to JuMP master
Migrate `setobjectivesense` to `set_objective_sense` for the next release of JuMP.

mlubin:
Use MOI.OptimizationSense instead of symbols at the JuMP level. (Enums are better than symbols when applicable.)
Add `set_objective_sense`.
Closes #1459 

@ccoffrin

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -888,14 +888,19 @@ macro objective(m, args...)
         error("in @objective: needs three arguments: model, objective sense (Max or Min) and expression.")
     end
     sense, x = args
-    if sense == :Min || sense == :Max
-        sense = Expr(:quote,sense)
+    if sense == :Min
+        sense = MOI.MinSense
+    elseif sense == :Max
+        sense = MOI.MaxSense
+    else
+        # Refers to a variable that holds the sense.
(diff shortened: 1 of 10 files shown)


src/macros.jl
```diff
@@ -1429,13 +1434,18 @@ end
...
 # TODO: Add a docstring.
 macro NLobjective(m, sense, x)
-    if sense == :Min || sense == :Max
-        sense = Expr(:quote,sense)
+    if sense == :Min
+        sense = MOI.MinSense
+    elseif sense == :Max
+        sense = MOI.MaxSense
+    else
+        # Refers to a variable that holds the sense.
```
blegat: It might improve readability to create a funtion like `sense_from_symbol` doing the if-then-else and used in both macros

mlubin: I added TODOs. IMO just moving this if/else into a method would be confusing because of the logic in
```julia
# Refers to a variable that holds the sense.
sense = esc(sense)
```
Needs a bigger refactor to properly address.

PR #1527, opened 2018-10-07
performance improvement in name generation

mlubin:
https://github.com/JuliaLang/julia/issues/29550

Thanks @IainNZ


KristofferC:
Are the elements of `idxvars` always integers?

mlubin:
@KristofferC no, they're not always integers. They could be any item used as a key in a dictionary. (To be precise, `idxvars` is an array of symbols. At runtime, the values will refer to some object.)

PR #1533, opened 2018-10-09
register expressions in the model scope

### LINKED ISSUE #1109: Accessing the expression
I create a JuMP model within a function and return the model. Within the model, expressions are created using the ``@expression`` macro. Where in the model can I find a dictionary of created expressions?

Thanks

mlubin:
Closes #1109

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -286,26 +286,22 @@ end
 """
     macro_assign_and_return(code, variable, name;
                             final_variable=variable,
-                            register::Bool=false,
-                            model=nothing)
+                            model_for_registering=nothing)
 
 Return runs `code` in a local scope which returns the value of `variable`
 and then assign `final_variable` to `name`.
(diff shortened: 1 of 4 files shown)


src/macros.jl
```diff
@@ -286,22 +286,26 @@ end
...
 Return runs `code` in a local scope which returns the value of `variable`
 and then assign `final_variable` to `name`.
-If `register_fun` is given, `register_fun(model, name, variable)` is called.
+If `register` is true is given, `register_object(model, name, variable)` is
+called in the generated code.
 """
 function macro_assign_and_return(code, variable, name;
                                  final_variable=variable,
-                                 register_fun::Union{Nothing, Function}=nothing,
+                                 register::Bool=false,
```
blegat: Maybe we can remove the `register` boolean argument and branch on whether `model` is `nothing` or not ?
This function does not work if `register` is `true` and `model` is nothing anyway so it might be cleaner

PR #1536, opened 2018-10-10
[release-0.18] Reorders function declarations in JuMPArray.jl

### LINKED ISSUE #1531: Reorder function declarations in JuMPArray.jl to improve compatibility with PackageCompiler
As described in [this thread](https://discourse.julialang.org/t/change-to-improve-jump-compatibility-with-packagecompiler/15944) on Discourse, I've found the order of function declarations in JuMPArray.jl (release 0.18) creates an incompatibility with PackageCompiler. The problem is there are @generated functions (Base.getindex and Base.setindex!) that reference a regular function (_to_cartesian) before it is declared. This causes an error when using PackageCompiler.

Moving the declaration statements to the end of the JuMPArray.jl file fixes the problem.

jwveysey:
Reorders function declarations in JuMPArray.jl to improve compatibility with PackageCompiler. Adds tests of compiling JuMP into a new system image using PackageCompiler.

Closes #1531.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -7,3 +7,4 @@ Ipopt
 ECOS
 SCS
 OffsetArrays 0.2.13
+PackageCompiler
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -7,4 +7,3 @@ Ipopt
 ECOS
 SCS
 OffsetArrays 0.2.13
-PackageCompiler
--- a/src/JuMPArray.jl
+++ b/src/JuMPArray.jl
@@ -34,22 +34,6 @@ end
     :(JuMPArray(innerArray, indexsets, $dicttuple, Dict{Symbol,Any}()))
 end
 
-Base.getindex(d::JuMPArray, ::Colon) = d.innerArray[:]
-
-@generated function Base.getindex(d::JuMPArray{T,N,NT}, idx...) where {T,N,NT}
-    if N != length(idx)
-        error("Indexed into a JuMPArray with $(length(idx)) indices (expected $N indices)")
-    end
-    Expr(:call, :getindex, :(d.innerArray), _to_cartesian(d,NT,idx)...)
-end
-
-@generated function Base.setindex!(d::JuMPArray{T,N,NT}, v, idx...) where {T,N,NT}
-    if N != length(idx)
-        error("Indexed into a JuMPArray with $(length(idx)) indices (expected $N indices)")
-    end
-    Expr(:call, :setindex!, :(d.innerArray), :v, _to_cartesian(d,NT,idx)...)
-end
-
 # Note that this is needed because automatic broadcasting of addition operators is deprecated in 1.0.
 # The explicitly broadcasted version ought to work in 0.6, but causes a mysterious generated function
 # related error because of the presence of this function in `@generated Base.setindex!`.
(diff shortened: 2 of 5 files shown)


jwveysey:
Hi, all - here's my attempt at a pull request to address issue #1531. Comments appreciated. Mostly what's new here are some tests of compiling JuMP with PackageCompiler. These take a while to execute due to the overhead of running PackageCompiler and creating a new system image. I'm not sure there's any way around that if you want to test the compilation _per se_.

As you may know, PackageCompiler requires a set of "snooping" commands that exercise the package that's being compiled. The JIT compilation that results from these commands is inspected to create a set of instructions for the ahead-of-time compilation. I chose snooping commands that create and solve LP and MILP models (borrowing from existing tests) since that's how I've been using PackageCompiler in reality. The commands could be extended to cover other use cases if desired.

blegat:
You need to add PackageCompiler in `test/REQUIRE`

jwveysey:
Quite right – sorry about that! I just made the necessary change.

IainNZ:
How long does the new test take to run?

blegat:
~ 30 min

mlubin:
Seems to be failing on 0.6 and 1.0?

ianfiske:
Had a look at the 1.0 failure.  The error is `UndefVarError: serialize not defined`.  @jwveysey You just need to put `using Serialization` in your test.

jwveysey:
Thanks for tracking it down, ianfiske.  I believe I've also identified the problem in v. 0.6.  I'm going to try to fix both today.

blegat:
Looking at the log we have: https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/440289533#L3809-L3811
Why is it complaining about undefined variable "MathOptInterface" ? Why is it using MathOptInterface on JuMP release-0.18 branch ?

jwveysey:
I'm not sure about that. The code I added doesn't call MathOptInterface directly, so I assume it's being referenced somewhere in JuMP itself.

blegat:
It is used by some solvers. Aren't these warning complaining about undefined variables an issue ?

jwveysey:
Having just taken a closer look at PackageCompiler's code, I would say not.  In this case, it's PackageCompiler working as designed.  I believe what's happening is the following:

1) PackageCompiler uses the snooping commands (the script in snoop_commands.jl) to run JuMP through some use cases and to create a precompilation file.  This file contains a set of precompile() statements based on the use cases.  Many of these statements reference JuMP, and some reference other packages involved in the use cases.  

2) At the beginning of the precompilation file, PackageCompiler inserts commands to import JuMP and any packages referenced in JuMP.  This behavior is not user-modifiable.

3) PackageCompiler builds a new system image using the precompilation file (among other resources).  The build operation happens in a separate instance of Julia, so the operative namespace contains only the packages in 2).

4) Every precompile() statement is wrapped in a try-catch block, however, so statements referencing packages not included in the current namespace are skipped.  These include statements that refer to other packaged involved in the use cases but not imported in JuMP itself.

The end result is probably a system image that's not quite as efficient for the test cases as it could be.  But it's still much better than the default system image.  And in any event, the exclusion of packages referenced in the use cases but not in the package being compiled is an inherent limitation of PackageCompiler.

jwveysey:
Noted, thanks. I've just implemented the requested change. The compile test script (compilation.jl) has been removed from runtests.jl and can be run separately as desired.

jwveysey:
@mlubin, please let me know if anything else is needed to accept this pull request. Thanks!

test/REQUIRE
```diff
@@ -7,3 +7,4 @@ Ipopt
 ECOS
 SCS
 OffsetArrays 0.2.13
+PackageCompiler
```
odow: Since this isn't getting run in the automated tests, we probably don't need this in REQUIRE?

PR #1537, opened 2018-10-10
Mention broadcasting syntax

mlubin:


blegat:
> What do we recommends for `Dict`s?

There is `Dict(idx => JuMP.result_value(var) for (idx, var) in d)` but it is annoying that `JuMP.result_value.(d)` does not work regardless of the container

mlubin:
Maybe we have to go back to supporting `JuMP.result_value(::JuMPContainer)` for consistency with dicts.

blegat:
> Maybe we have to go back to supporting JuMP.result_value(::JuMPContainer) for consistency with dicts.

Yes, we can just have
```julia
result_value(x::AbstractArray{<:AbstractJuMPScalar})= result_value.(x)
```
and similar for JuMPArray, Dict{<:Any, <:AbstractJuMPScalar} and for constraints.

mlubin:
@blegat and I discussed this today. There are a few different proposals.

1. Implement `JuMP.value(dict::Dict) = map_values(value, dict)`, `JuMP.value(arr::Union{Array, JuMPArray}) = value.(arr)` and similarly for *every* function on VariableRef and ConstraintRef like `upper_bound`, `lower_bound`, `dual`, .... We would not recommend broadcasting syntax for users who don't know what container type they're using. `JuMP.value(x)` would "just work" for all container types.

2. Make dictionary construction in `@variable` visually distinct, e.g., forcing users to write `@variable(model, x[i=1:N, j=i:N], container=Dict)`. Don't implement `JuMP.value(::Dict)`. Ask users to write `map_values(JuMP.value, x)` for dictionaries of JuMP variables.

3. Revive `JuMPDict` and implement broadcasting support so that `JuMP.value.(jump_dict)` works. Broadcasting support for SparseArrays ([code](https://github.com/JuliaLang/julia/blob/master/stdlib/SparseArrays/src/higherorderfns.jl)) could be a useful starting point.

Option 1 is a bit tedious to implement but not so bad. IMO option 3 is a substantial amount of work but @blegat is already experimenting :)

@chkwon, do you have an option on the usability implications of the three proposals?

Note:

`map_values` is defined as:
```julia
function map_values(f, dict::Dict)
    return Dict(key => f(value) for (key, value) in dict)
end
```
This can be defined in JuMP for convenience.

joaquimg:
1 already looks like a never ending pain, there will always be some functionality missing.

About 3, What if JuMPDict simply wrapped Dict to have some extra functionality?
Is that @blegat ‘s plan?

mlubin:
Updated. This should be good to go now.

chkwon:
Sorry for catching up late.  Option 3 looks good to me.  Option 2 can be quite challenging to beginners.

PR #1543, opened 2018-10-11
Helpful user message for old objective syntax

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1541

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -896,21 +896,16 @@ function moi_sense(_error::Function, sense)
         # TODO: Better document this behavior
         expr = esc(sense)
     end
-    return :(throw_error_for_symbol_sense($_error, $expr))
+    return :(throw_error_for_invalid_sense($_error, $expr))
 end
 
-# TODO remove for JuMP v0.20
-function throw_error_for_symbol_sense(_error::Function,
-                                      sense::Symbol)
-    _error("The `@objective(model, :Min, ...)` and",
-           " `@objective(model, :Max, ...)` are no longer available in JuMP",
-           " 0.19 and later. Use `@objective(model, Min, ...)`," *
-           " `@objective(model, Max, ...)`,",
-           " `@objective(model, MOI.MinSense, ...)` or " *
-           " `@objective(model, MOI.MaxSense, ...)` instead.")
+function throw_error_for_invalid_sense(_error::Function,
+                                       sense)
+    _error("Unexpected sense `$value`. The sense must be an",
+           " `MOI.OptimizatonSense`, `Min` or `Max`.")
 end
-function throw_error_for_symbol_sense(_error::Function,
-                                      sense::MOI.OptimizationSense)
+function throw_error_for_invalid_sense(_error::Function,
+                                       sense::MOI.OptimizationSense)
     return sense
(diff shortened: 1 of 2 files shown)


mlubin:
Why do we need to provide two ways to write the same thing?

blegat:
It is not mandatory to support `@objective(model, :Min, x+1)` but we need to support `@objective(mode, sense, x+1)` where `sense` is a variable whose value is `:Min`.
If we want to support the second one, the first one comes for free.

mlubin:
> @objective(mode, sense, x+1) where sense is a variable whose value is :Min.

No, this shouldn't work anymore after #1562. Sense symbols are no longer recognized. You can use MOI sense values instead.

blegat:
> Sense symbols are no longer recognized

Shouldn't we add an entry in NEWS.md about that ?

mlubin:
Yep, done (#1545).

blegat:
@mlubin I have replace the support of `:Min` and `:Max` by helpful errors in case symbols are used.

src/macros.jl
```diff
@@ -875,32 +875,64 @@ Constructs a vector of `LinearConstraint` objects. Similar to `@LinearConstraint
 Constructs a vector of `QuadConstraint` objects. Similar to `@QuadConstraint`, except it accepts multiple constraints as input as long as they are separated by newlines.
 """ :(@QuadConstraints)
 
+"""
+    moi_sense(_error::Function, sense)
+
+Return an expression whose value is an `MOI.OptimizationSense` corresponding
+to `sense`. Sense is either the symbol `:Min` or `:Max`, corresponding
+respectively to `MOI.MinSense` and `MOI.MaxSense` or it is another symbol,
+which should be the name of a variable or expression whose value is `:Min`,
```
mlubin: This discussion about `:Min` and `:Max` is out of date now, right?

src/macros.jl
```diff
@@ -875,32 +875,64 @@ Constructs a vector of `LinearConstraint` objects. Similar to `@LinearConstraint
...
+`:Max` or an `MOI.OptimizationSense`.
+In the last case, the expression throws an error using the `_error`
+function in case the value is a symbol which is not `:Min` nor `:Max`.
```
mlubin: Out of date?

src/macros.jl
```diff
@@ -875,32 +875,64 @@ Constructs a vector of `LinearConstraint` objects. Similar to `@LinearConstraint
...
+        expr = MOI.MaxSense
+    else
+        # Refers to a variable that holds the sense.
+        # TODO: Better document this behavior
+        expr = esc(sense)
+    end
+    return :(throw_error_for_symbol_sense($_error, $expr))
+end
+
+# TODO remove for JuMP v0.20
```
mlubin: Actually something like this should stay, because of the `MethodError` principle. What about updating the error message so that it just says what's wrong instead of referring to old JuMP? For example, `"Unexpected sense $value. The sense must be an `OptimizatonSense` or \"Min\" or \"Max\"."` (Also remove the ::Symbol, we want an error if someone gives a sense of `10`.)

PR #1545, opened 2018-10-11
add senses to list of breaking changes

mlubin:


NEWS.md
```diff
@@ -64,6 +64,9 @@ Breaking changes:
   forms and second-order cone constraints. Support for these
   constraint classes depends on the solver.
 
+- The symbols `:Min` and `:Max` are no longer used as optimization senses.
```
blegat: This might easily be misinterpreted as saying that `@objective(mode, Min, x+1)` is not supported.

PR #1549, opened 2018-10-14
Export MOI shortcut

blegat:
This PR exports `MOI` from JuMP so that users of JuMP do not have to copy paste
```julia
using MathOptInterface
const MOI = MathOptInterface
```
everywhere

This was [suggested](https://discourse.julialang.org/t/first-alpha-release-of-jump-0-19-jump-mathoptinterface/16099/5) by @odow

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -18,6 +18,7 @@ using Compat.SparseArrays
 
 import MathOptInterface
 const MOI = MathOptInterface
+export MOI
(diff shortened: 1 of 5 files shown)


mlubin:
This doesn't seem right. Having JuMP export a mysterious object named "MOI" doesn't help with readability of the code. Scipy doesn't export `np` as numpy. We should consider re-exporting the important enums from MOI like optimization sense and statuses.

PR #1560, opened 2018-10-25
📝 Improve doc of optimize!

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1559

### DIFF
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -51,8 +51,6 @@ model = Model()
 # ...fill model with variables, constraints and objectives...
 # Solve the model with GLPK
 JuMP.optimize!(model, with_optimizer(GLPK.Optimizer))
-# Solve the model with Clp
-JuMP.optimize!(model, with_optimizer(Clp.Optimizer))
 ```
 """
 function optimize!(model::Model,
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -27,12 +27,24 @@ end
 
 
 """
-    function optimize!(model::Model,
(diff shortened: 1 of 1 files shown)


src/optimizer_interface.jl
```diff
@@ -27,12 +27,24 @@ end
...
+              optimizer_factory::Union{Nothing, OptimizerFactory}=nothing;
+              ignore_optimize_hook=(model.optimize_hook === nothing))
 
-Optimize the model. If `optimizer_factory` is not `nothing`, it first set the
-optimizer to a new one created using the optimizer factory.
+Optimize the model. If `optimizer_factory` is not `nothing`, it first sets the
+optimizer to a new one created using the optimizer factory. The factory can be
+created using the [`with_optimizer`](@ref) function.
+
+## Examples
```
mlubin: I'd suggest two separate examples, one where the optimizer is provided with `Model()` and another when it's provided with `optimize!`.

src/optimizer_interface.jl
```diff
@@ -27,12 +27,33 @@ end
...
+# Solve the model with GLPK
+JuMP.optimize!(model)
+```
+or in the `optimize!` call as follows:
+```julia
+model = Model()
+# ...fill model with variables, constraints and objectives...
+# Solve the model with GLPK
+JuMP.optimize!(model, with_optimizer(GLPK.Optimizer))
+# Solve the model with Clp
```
mlubin: Why is the same model solved twice?

blegat: You are right, it won't work, it will throw `"An optimizer factory cannot both be provided in the `Model` constructor and at the `optimize` call."` at the second call

PR #1568, opened 2018-10-27
Implement shadow_price

### LINKED ISSUE #1552: Implement shadow_price and mention dual sign changes in NEWS
https://discourse.julialang.org/t/result-dual-vs-getdual-signs-are-the-opposite/16355

mlubin:
Closes #1552 

@chkwon please check

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -245,9 +245,9 @@ end
 
 The change in the objective from an infinitesimal relaxation of the constraint.
 This value is computed from [`result_dual`](@ref) and can be queried only when
-`has_result_dual` is `true` and the objective sense is `MinSense` or `MaxSense`.
-For linear constraints, the shadow prices differ at most in sign from the
-`result_dual` value depending on the objective sense.
+`has_result_dual` is `true` and the objective sense is `MinSense` or `MaxSense`
+(not `FeasibilitySense`). For linear constraints, the shadow prices differ at
+most in sign from the `result_dual` value depending on the objective sense.
 
 ## Notes
 
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -239,3 +239,98 @@ function set_coefficient(constraint::ConstraintRef{Model, MOICON{F, S}},
         MOI.ScalarCoefficientChange(index(variable), convert(T, value)))
     return
 end
+
+"""
+    shadow_price(constraint::ConstraintRef)
+
+The change in the objective from an infinitesimal relaxation of the constraint.
+This value is computed from [`result_dual`](@ref) and can be queried only when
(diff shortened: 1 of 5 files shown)


src/constraints.jl
```diff
@@ -239,3 +239,98 @@ function set_coefficient(constraint::ConstraintRef{Model, MOICON{F, S}},
         MOI.ScalarCoefficientChange(index(variable), convert(T, value)))
     return
 end
+
+"""
+    shadow_price(constraint::ConstraintRef)
+
+The change in the objective from an infinitesimal relaxation of the constraint.
+This value is computed from [`result_dual`](@ref) and can be queried only when
+`has_result_dual` is `true` and the objective sense is `MinSense` or `MaxSense`.
```
blegat: Most user will not understand what the objective sense can be if it is not `MinSense` nor `MaxSense`.

test/constraint.jl
```diff
@@ -284,3 +284,86 @@ end
...
+    variables: x, y
+    maxobjective: 1.0*x
+    xub: x <= 2.0
+    ylb: y >= 0.0
+    c: x + y <= 1.0
+    """,
+    # Optimal duals
+    Dict("xub" => 0.0, "ylb" => 1.0, "c" => -1.0),
+    # Expected shadow prices
+    Dict("xub" => 0.0, "ylb" => 1.0, "c" => 1.0))
```
blegat: If we increase the constant of `ylb` by `ε` then, we have in `c` : `1.0 > x + y > x + ε` hence `x < 1 - ε`.
Hence I would expect a shadow price of `-1.0` for `ylb`.

src/constraints.jl
```diff
@@ -239,3 +239,98 @@ function set_coefficient(constraint::ConstraintRef{Model, MOICON{F, S}},
...
+          "of constraint.")
+end
+
+# Internal function.
+function shadow_price_less_than_(dual_value, sense::MOI.OptimizationSense)
+    # When minimizing, the shadow price is nonpositive and when maximizing the
+    # shadow price is nonnegative (because relaxing a constraint can only
+    # improve the objective). By MOI convention, a feasible dual on a LessThan
+    # set is nonpositive, so we flip the sign when maximizing.
+    if sense == MOI.MaxSense
```
blegat: Why not merge the two functions into one and do `sense == MOI.MaxSense ⊻ greater_than` ?
One might argue that it decreases readability but the fact that the comment in the other function references this one means that merging them might also increase readability.

mlubin: I understand the two cases separately. Also both functions are small enough to fit on a single screen together, so readability seems fine to me.

chkwon:
Thanks. Looks good to me.

PR #1570, opened 2018-10-28
Nice error for unsupported objective function

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1500

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -344,10 +344,27 @@ end
 """
     set_objective_sense(model::Model, sense::MathOptInterface.OptimizationSense)
 
-Sets the objective sense of the model to the given sense.
+Sets the objective sense of the model to the given sense. See
+[`set_objective_function`](@ref) to set the objective function.
 """
 function set_objective_sense(model::Model, sense::MOI.OptimizationSense)
     MOI.set(model, MOI.ObjectiveSense(), sense)
+end
+
+"""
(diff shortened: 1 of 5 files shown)


src/JuMP.jl
```diff
@@ -344,12 +344,29 @@ end
 """
     set_objective_sense(model::Model, sense::MathOptInterface.OptimizationSense)
 
-Sets the objective sense of the model to the given sense.
+Sets the objective sense of the model to the given sense. See
+[`set_objective_function`](@ref) to set the objective function.
```
mlubin: Also mention the `@objective` macro? `set_objective_function` isn't the typical way to set the objective function.

src/aff_expr.jl
```diff
@@ -297,9 +297,7 @@ moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
 
 function set_objective(model::Model, sense::MOI.OptimizationSense, a::AffExpr)
     set_objective_sense(model, sense)
-    MOI.set(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}(),
-            MOI.ScalarAffineFunction(a))
-    return nothing
```
mlubin: Keeping the explicit `return` is helpful for type inference because we don't know what `MOI.set` will return.

PR #1571, opened 2018-10-28
📝 Document symmetry handling for psd constraint

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1471

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -614,19 +614,70 @@ function parseSDconstraint(_error::Function, args...)
 end
 
 """
-    @SDconstraint(m::Model, expr)
+    @SDconstraint(model::Model, expr)
 
 Add a semidefinite constraint described by the expression `expr`.
 
-    @SDconstraint(m::Model, ref[i=..., j=..., ...], expr)
+    @SDconstraint(model::Model, ref[i=..., j=..., ...], expr)
 
 Add a group of semidefinite constraints described by the expression `expr`
 parametrized by `i`, `j`, ...
 
 The expression `expr` needs to be of the form `a sign b` where `sign` is `⪰`,
 `≥`, `>=`, `⪯`, `≤` or `<=` and `a` and `b` are `square` matrices. It
-constrains that `a - b` (or `b - a` if the sign is `⪯`, `≤` or `<=`) is
-positive semidefinite.
+constrains the matrix `x = a - b` (or `x = b - a` if the sign is `⪯`, `≤` or
+`<=`) to be symmetric and positive semidefinite.
+
+If `x` is already symmetric, use `@constraint(model, Symmetric(x) in PSDCone())`
+instead to remove the need to compare the corresponding off-diagonal entries
+and add equality constraints if they contain different expression.
+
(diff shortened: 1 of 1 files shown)


docs/src/constraints.md
```diff
@@ -15,6 +15,18 @@ nonlinear constraints.
 @SDconstraint
 ```
 
+## Sets
+
+As mentioned in the documentation of the [`@constraint`](@ref) and
+[`@SDconstraint`](@ref) macros, the following sets can be used to create
```
mlubin: In addition to any MOI set.

src/macros.jl
```diff
@@ -614,19 +617,53 @@ function parseSDconstraint(_error::Function, args...)
...
 
 The expression `expr` needs to be of the form `a sign b` where `sign` is `⪰`,
 `≥`, `>=`, `⪯`, `≤` or `<=` and `a` and `b` are `square` matrices. It
-constrains that `a - b` (or `b - a` if the sign is `⪯`, `≤` or `<=`) is
-positive semidefinite.
+constrains the matrix `x = a - b` (or `x = b - a` if the sign is `⪯`, `≤` or
+`<=`) to be symmetric and positive semidefinite.
+
+If `x` is already symmetric, use `@constraint(model, Symmetric(x) in PSDCone())`
+instead to remove the need to compare the corresponding off-diagonal entries
```
mlubin: "remove the need" is a bit confusing (why is it needed?). We can say that by default, we check numerical symmetry of the matrix `x`, and if symmetry is violated by some arbitrary amount, we add explicit equality constraints. You can use `Symmetric(x)` to skip these checks if you know the matrix must be symmetric.

src/sd.jl
```diff
@@ -1,4 +1,44 @@
-# Used in @constraint model x in PSDCone
+"""
+    PSDCone
+
+Positive semidefinite cone object that can be used to constrain a square matrix
+to be positive semidefinite in the [`@constraint`](@ref) macro. If the matrix
+has type `Symmetric` then the columns vectorization of its upper triangular
```
mlubin: The "columns vectorization" concept is used a lot in this docstring. Consider spending a sentence defining what it is.

src/sd.jl
```diff
@@ -1,4 +1,44 @@
...
+part is constrained to belong to the `MOI.PositiveSemidefiniteConeTriangle` set,
+otherwise it column vectorization is constrained to belong to the
```
mlubin: Typo: it -> its

docs/src/constraints.md
```diff
@@ -19,7 +19,7 @@ nonlinear constraints.
 
 As mentioned in the documentation of the [`@constraint`](@ref) and
 [`@SDconstraint`](@ref) macros, the following sets can be used to create
-constraints.
+constraints in addition to [any MOI set](http://www.juliaopt.org/MathOptInterface.jl/stable/apireference.html#Sets-1).
```
mlubin: It's better to use versioned doc links so they don't break later.

PR #1572, opened 2018-10-28
👷 Disable osx on Travis and 0.6/0.7 on AppVeyor

blegat:
See discussion on Gitter. osx does not provide much information as there is no binary dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: julia
 os:
     - linux
+#   - osx # Disable to speed up CI. JuMP has no binary dependency so the result
+          # with osx and linux should be similar.
 julia:
     - 0.6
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: julia
 os:
     - linux
-    - osx
```
mlubin: Better to comment this out and say that it's disabled because of long CI times.

appveyor.yml
```diff
@@ -1,7 +1,5 @@
 environment:
   matrix:
-  - julia_version: 0.6
```
mlubin: Same here. As long as we still support 0.6 and 0.7, add a comment explaining why they're disabled.

PR #1574, opened 2018-10-30
♻️ Refactor objective_function out of aff/quad/var

blegat:
Also add examples in the doc. Note that the current docs were incorrect as it errors if it is not *convertible*, not if is not *exactly this type*.
I think the behavior changed in MOI from the time the doc was written.

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1576


src/objective.jl
```diff
@@ -78,3 +78,48 @@ function set_objective(model::Model, sense::MOI.OptimizationSense,
...
+julia> JuMP.objective_function(model, JuMP.GenericAffExpr{Float64})
+2 x + 1
+
+julia> JuMP.objective_function(model, JuMP.GenericQuadExpr{Float64})
+2 x + 1
+
+julia> typeof(JuMP.objective_function(model, JuMP.GenericQuadExpr{Float64}))
+JuMP.GenericQuadExpr{Float64,VariableRef}
+```
+We see with the last two command that even if the objective function is affine,
```
mlubin: s/command/commands/

src/objective.jl
```diff
@@ -78,3 +78,48 @@ function set_objective(model::Model, sense::MOI.OptimizationSense,
...
+as it is convertible to a quadratic function, it can be queried as a quadratic
+function and the result is quadratic.
+
+However, it is not convertible to a variable.
+```
+julia> JuMP.objective_function(model, JuMP.VariableRef)
```
mlubin: Consider doctesting this block also.

src/objective.jl
```diff
@@ -78,3 +78,48 @@ function set_objective(model::Model, sense::MOI.OptimizationSense,
...
+julia> model = Model()
+A JuMP Model
+
+julia> @variable(model, x)
+x
+
+julia> @objective(model, Min, 2x + 1)
+2 x + 1
+
+julia> JuMP.objective_function(model, JuMP.GenericAffExpr{Float64})
```
mlubin: Most users are probably interested in `AffExpr` instead of `GenericAffExpr{Float64}`.

src/objective.jl
blegat: The issue is that the output shows there the package is located which varies depending on the user setup.

odow: You can use doctest filters to strip this out:
https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Filtering-Doctests-1

src/macros.jl
```diff
@@ -947,7 +947,55 @@ function throw_error_for_invalid_sense(_error::Function,
...
+julia> @objective(model, Max, 2x - 1)
+2 x - 1
+```
+
+To set a quadratic objective and set the objective sense programatically, do
+as follows:
+```jldoctest @objective
+julia> using MathOptInterface
+
+julia> const MOI = MathOptInterface
```
mlubin: Why spend an extra line importing MOI to use it once?

blegat: To show the user were our acronym comes from and how to get it working

blegat:
I have added `strict = true` so that doctest failure make travis fail

PR #1577, opened 2018-11-01
limited support for splatting in nonlinear expressions

### LINKED ISSUE #1115: Is it possible to support varargs functions in objective function definition, like `setNLobjective`?
Currently using varargs functions in `setNLobjective` doesn't work, so there are questions on StackOverflow like 
<https://stackoverflow.com/questions/44710900/juliajump-variable-number-of-arguments-to-function/44711305#44711305>
and 
<https://stackoverflow.com/questions/46632458/an-example-of-nonlinear-optimization-with-jump>

It is natural to solve the issues in these questions by things like `myf(x...)` and `rosen(x...)` as the objective, but this doesn't work.

Is it possible to support varargs functions in objective function definition? For example, we could define a function `varargsExpand!` like this:

```
function varargsExpand!(m :: JuMP.Model, expr :: Expr)
    if expr.head == :... && length(expr.args) == 1 &&
        haskey(m.objDict, expr.args[1]) && typeof(m.objDict[expr.args[1]]) == Vector{JuMP.Variable}
            return([m.objDict[expr.args[1]][i] for i in 1:length(m.objDict[expr.args[1]])])
    else
        expr.args = vcat([varargsExpand!(m, expr.args[i]) for i in 1:length(expr.args)]...)
        return([expr])
    end
end

function varargsExpand!(m :: JuMP.Model, x)
    [x]
end
````

And the function could be used like this:

```
function setNLobjective(m::Model, sense::Symbol, x)
    initNLP(m)
    setobjectivesense(m, sense)
    x = varargsExpand!(m, x)[1]
    ex = NonlinearExprData(m, x)
    m.nlpdata.nlobj = ex
    m.obj = zero(QuadExpr)
    m.internalModelLoaded = false
    return
end
```

This is not a sophisticated way, for example, it only supports things like `rosen(x...)` but doesn't support things like `rosen(x[1:10]...)`, but at least it is better than nothing, and it is really useful and could be improved.

### PR DESCRIPTION
**Pull request by mlubin**
It was easier to implement than keep explaining that JuMP wasn't designed for this.

Closes #1115
Also fixes pre-existing doctest failures.

### DIFF
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -38,7 +38,6 @@ JuMP.set_coefficient
 ## Duals
 
 ```@docs
-JuMP.has_result_dual
 JuMP.result_dual
 JuMP.shadow_price
 ```
--- a/NEWS.md
+++ b/NEWS.md
@@ -79,6 +79,9 @@ Breaking changes:
 
 New features:
 
+- Splatting (like `f(x...)`) is recognized in restricted settings in nonlinear
+  expressions.
+
 - Support for deleting constraints and variables.
 
 - The documentation has been completely rewritten using docstrings and
--- a/docs/src/nlp.md
+++ b/docs/src/nlp.md
@@ -58,7 +58,8 @@ Syntax notes
 The syntax accepted in nonlinear expressions is more restricted than the syntax
 for linear and quadratic expressions. We note some important points below.
 
-- All expressions must be simple scalar operations. You cannot use `dot`,
+- With the exception of the splatting syntax discussed below, all expressions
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

File: docs/src/constraints.md [hunk 1]
```diff
@@ -38,7 +38,6 @@ JuMP.set_coefficient
 ## Duals
 
 ```@docs
-JuMP.has_result_dual
```
blegat: Why removing it ?

blegat: I see, missing docstring ^^

PR #1581, opened 2018-11-04
Rename cref field m->model

blegat:


blegat:
Merged as it is required by https://github.com/JuliaOpt/JuMP.jl/pull/1583

PR #1584, opened 2018-11-05
Improve show extensibility for AbstractConstraint

blegat:
In printing, the function and set are accessed by the `.func` and `.set` field so it assumes that any `AbstractConstraint` has this field.
With this PR, an `AbstractConstraint` can either implement `jump_function` and `set_function` or even have custom printing by implementing `function_string` and `in_set_string`

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -322,7 +322,7 @@ Return a `String` representing the function of the constraint `constraint`
 using print mode `print_mode`.
 """
 function function_string(print_mode::PrintMode, constraint::AbstractConstraint)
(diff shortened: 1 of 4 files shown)


src/print.jl
```diff
@@ -272,13 +272,23 @@ end
 ## Constraints
 #------------------------------------------------------------------------
 
-function Base.show(io::IO, ref::ConstraintRef{Model})
+function Base.show(io::IO, ref::ConstraintRef)
```
mlubin: Please document the methods that a `ConstraintRef` needs to implement in order to avoid a `MethodError` on `show`.

PR #1587, opened 2018-11-07
Improve error message when querying a variable bound that does not exist

### LINKED ISSUE #1509: Better error message on missing variable bounds
In JuMP master this code,
```julia
using JuMP
m = Model()
x = @variable(m, [i in 1:3], lower_bound=0)
JuMP.lower_bound(x[1])
JuMP.upper_bound(x[1])
```
leads to the following AssertionError,
```
ERROR: AssertionError: has_upper_bound(v)
Stacktrace:
 [1] upper_bound_index(::JuMP.VariableRef) at /home/.julia/v0.6/JuMP/src/variables.jl:344
 [2] UpperBoundRef at /home/.julia/v0.6/JuMP/src/variables.jl:372 [inlined]
 [3] upper_bound(::JuMP.VariableRef) at /home/.julia/v0.6/JuMP/src/variables.jl:395
```

odow:
Closes #1509

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -257,7 +257,9 @@ function has_lower_bound(v::VariableRef)
 end
 
 function lower_bound_index(v::VariableRef)
-    @assert has_lower_bound(v) # TODO error message
+    if !has_lower_bound(v)
+        error("Variable $(v) does not have a lower bound.")
+    end
     return owner_model(v).variable_to_lower_bound[index(v)]
 end
 function set_lower_bound_index(v::VariableRef, cindex::MOILB)
@@ -320,7 +322,9 @@ function has_upper_bound(v::VariableRef)
 end
 
 function upper_bound_index(v::VariableRef)
-    @assert has_upper_bound(v) # TODO error message
+    if !has_upper_bound(v)
+        error("Variable $(v) does not have an upper bound.")
+    end
     return owner_model(v).variable_to_upper_bound[index(v)]
 end
 function set_upper_bound_index(v::VariableRef, cindex::MOIUB)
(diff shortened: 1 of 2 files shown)


src/variables.jl
```diff
@@ -257,7 +257,9 @@ function has_lower_bound(v::VariableRef)
 end
 
 function lower_bound_index(v::VariableRef)
-    @assert has_lower_bound(v) # TODO error message
+    if !has_lower_bound(v)
+        error("Variable $(v) does not have a lower bound.")
```
blegat: This might be misleading. Consider a polyhedral feasible set, if you eliminate all variables except `v` then it might be a set that is bounded from below.
Maybe `lower bound constraint` ?

odow: This error message will almost always be called when users run `JuMP.lower_bound(x)` if `x` is missing a lower bound.

There might be other constraints that enforce a lower bound on the variable, but that is a different kettle of fish.

blegat: > that is a different kettle of fish.

That is clear for us, but is that clear for a user that has just started using JuMP and has no idea what `JuMP.lower_bound` is ?

odow: http://www.juliaopt.org/JuMP.jl/latest/variables/#Variable-bounds-1

blegat: Who reads the doc if it is not thrown on your face :-P

blegat: Also a user that has done `@constraint model x >= 0` will not understand why there is no lower bound.

odow: This relates to https://github.com/JuliaOpt/JuMP.jl/issues/1586. I feel that if the user writes `@constraint`, we should add a constraint (e.g., a row) to the problem. 

Then we make "variable bound" mean a bound that was introduced in `@variable` or via `JuMP.set_lower_bound` and `JuMP.set_upper_bound`.

mlubin: I think the error message is sufficiently clear given that `has_lower_bound` is a documented concept in JuMP.

PR #1588, opened 2018-11-07
Implement value for GenericQuadExpr.

### LINKED ISSUE #1563: result_value for QuadExpr
https://github.com/JuliaOpt/JuMP.jl/blob/8c36bf2a850af6392a8bba3b8241d3577b69e38a/src/quad_expr.jl#L250

odow:
Closes #1563

### DIFF
--- a/src/quad_expr.jl
+++ b/src/quad_expr.jl
@@ -232,7 +232,7 @@ function Base.copy(q::GenericQuadExpr, new_model::Model)
                     copy(q.qcoeffs), copy(q.aff, new_model))
 end
 
-# Requires that value(::GenericAffExpr, foo) and foo(::VarType) is defined.
+# Requires that foo(::VarType) is defined.
 function value(ex::GenericQuadExpr{CoefType, VarType},
                foo::Function) where {CoefType, VarType}
     RetType = Base.promote_op(
--- a/src/quad_expr.jl
+++ b/src/quad_expr.jl
@@ -232,14 +232,15 @@ function Base.copy(q::GenericQuadExpr, new_model::Model)
                     copy(q.qcoeffs), copy(q.aff, new_model))
 end
 
-# Requires that foo(::VarType) is defined.
+# Requires that value_func(::VarType) is defined.
 function value(ex::GenericQuadExpr{CoefType, VarType},
-               foo::Function) where {CoefType, VarType}
+               value_func::Function) where {CoefType, VarType}
     RetType = Base.promote_op(
-        (ctype, vtype) -> ctype * foo(vtype) * foo(vtype), CoefType, VarType)
-    ret = convert(RetType, value(ex.aff, foo))
+        (ctype, vtype) -> ctype * value_func(vtype) * value_func(vtype),
+        CoefType, VarType)
+    ret = convert(RetType, value(ex.aff, value_func))
     for (vars, coef) in ex.terms
-        ret += coef * foo(vars.a) * foo(vars.b)
+        ret += coef * value_func(vars.a) * value_func(vars.b)
     end
     return ret
 end
--- a/src/quad_expr.jl
+++ b/src/quad_expr.jl
@@ -232,16 +232,12 @@ function Base.copy(q::GenericQuadExpr, new_model::Model)
                     copy(q.qcoeffs), copy(q.aff, new_model))
 end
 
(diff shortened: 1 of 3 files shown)


src/quad_expr.jl
```diff
@@ -232,4 +232,20 @@ function Base.copy(q::GenericQuadExpr, new_model::Model)
...
-# TODO: result_value for QuadExpr
+# Requires that foo(::GenericAffExpr) is defined.
+function value(ex::GenericQuadExpr{CoefType, VarType},
+               foo::Function) where {CoefType, VarType}
+    # The return type of appling map to ::VarType.
+    MapVarType = Base.promote_op(foo, VarType)
+    # Later, we're going to multiply two MapVarType together
+    MapVarType2 = Base.promote_op(*, MapVarType, MapVarType)
+    # We're also going to multiply a constant with ::MapVarType2
+    RetType = Base.promote_op(*, CoefType, MapVarType2)
```
blegat: I was once told that it is not safe to call `promote_op` on top of `promote_op`, see https://github.com/JuliaLang/julia/pull/18218#issuecomment-242331704.
So maybe something like `RetType = Base.promote_op((c, v) -> c * foo(v) * foo(v), CoefType, VarType)` is better.

src/quad_expr.jl
```diff
@@ -232,4 +232,20 @@ function Base.copy(q::GenericQuadExpr, new_model::Model)
...
+    ret = convert(RetType, foo(ex.aff))
```
blegat: `foo` gives the value of variables so it would make more sense to do `value(ex.aff, foo)`

src/quad_expr.jl
odow: I just followed
https://github.com/JuliaOpt/JuMP.jl/blob/6daac7ae3105d1c2ddcd44a1324acd3769fcef64/src/aff_expr.jl#L105-L113

blegat: I never understood the argument so I am not surprised to see that I am not following it :-P
In fact, it appears that using `promote_op` is just bad: https://github.com/JuliaLang/julia/pull/26344#issuecomment-371613750
So the best way is just to play with `typeof`, `zero`, `one` and `oneunit`. To choose whether to use `one` or `oneunit` you need to think what would happen if the type was frim Unitful. For choosing between `zero` and `one`, you have some freedom but you need to avoid dividing by zero.
Maybe the best way here is doing `typeof(oneunit(CoefType) * oneunit(MapVarType) * oneunit(MapVarType))`. We are kind of forced to use `promote_op` to get `MapVarType` since `one` ... would on `JuMP.VariableRef` would not really do what we want.

src/quad_expr.jl
```diff
@@ -232,4 +232,16 @@ function Base.copy(q::GenericQuadExpr, new_model::Model)
                     copy(q.qcoeffs), copy(q.aff, new_model))
 end
 
-# TODO: result_value for QuadExpr
+# Requires that value(::GenericAffExpr, foo) is defined.
```
blegat: Why not "requires that `foo(::JuMP.VariableRef)` is defined" ?

src/quad_expr.jl
```diff
@@ -232,7 +232,7 @@ function Base.copy(q::GenericQuadExpr, new_model::Model)
                     copy(q.qcoeffs), copy(q.aff, new_model))
 end
 
-# Requires that value(::GenericAffExpr, foo) is defined.
+# Requires that value(::GenericAffExpr, foo) and foo(::VarType) is defined.
```
blegat: Why is "`value(::GenericAffExpr, foo)` is defined" in the requirements ? It is defined and it will work if `foo(::VarType)` is defined

odow: Oh, that's what you meant. Good point.

src/quad_expr.jl
```diff
@@ -232,4 +232,16 @@ function Base.copy(q::GenericQuadExpr, new_model::Model)
                     copy(q.qcoeffs), copy(q.aff, new_model))
 end
 
-# TODO: result_value for QuadExpr
+# Requires that foo(::VarType) is defined.
```
mlubin: Please pick a more descriptive name than `foo` :)
Maybe `value_func`.

PR #1591, opened 2018-11-07
Add getter for moi_backend

### LINKED ISSUE #1284: Optimizer Getter Function
Consider adding something along the lines of `getoptimizer(model)` to coincide with `resetoptimizer!`.  Otherwise, JuMP users will need to know that the optimizer field is called `model.moibackend`.

odow:
Closes #1284 

- [x] Still needs docs

First commit adds `JuMP.backend(model)` as a function. 
Second commit replaces all calls to `model.moi_backend` with the function for consistency.

### DIFF
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -3,10 +3,11 @@ Interacting with solvers
 
 A JuMP model keeps a [MathOptInterface (MOI)](https://github.com/JuliaOpt/MathOptInterface.jl)
 *backend* of type `MOI.ModelLike` internally that stores the optimization
-problem and acts as the optimization solver. We call it an MOI *backend* and
-not optimizer as it can also be a wrapper around an optimization file format
-such as MPS that writes the JuMP model in a file. JuMP can be viewed as a
-lightweight user-friendly layer on top of the MOI backend:
+problem and acts as the optimization solver. We call it an MOI *backend* and not
+optimizer as it can also be a wrapper around an optimization file format such as
+MPS that writes the JuMP model in a file. From JuMP, the MathOptInterface
+backend can be accessed using the [@ref](`JuMP.backend`) function. JuMP can be
+viewed as a lightweight user-friendly layer on top of the MOI backend:
 
 * JuMP does not maintain any copy of the model outside this MOI backend.
 * JuMP variable (resp. constraint) references are simple structures containing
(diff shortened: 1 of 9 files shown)


src/JuMP.jl
```diff
@@ -269,30 +269,38 @@ if VERSION >= v"0.7-"
...
     else
         error("The function `caching_optimizer` cannot be called on a model " *
               "in `Direct` mode.")
     end
 end
 
+"""
+    backend(model::Model)
+
+Return the MathOptInterface model. This should only be used by advanced users
```
mlubin: Which MathOptInterface model? There are typically lots of them. The docstring to point to the documentation on manual/automatic/direct modes to explain what object you get when you ask for the backend.

odow: I made a stab, but honestly, there are so many layers I'm not 100% sure I got it right.

docs/src/solvers.md
```diff
@@ -3,10 +3,11 @@ Interacting with solvers
...
 A JuMP model keeps a [MathOptInterface (MOI)](https://github.com/JuliaOpt/MathOptInterface.jl)
 *backend* of type `MOI.ModelLike` internally that stores the optimization
-problem and acts as the optimization solver. We call it an MOI *backend* and
-not optimizer as it can also be a wrapper around an optimization file format
-such as MPS that writes the JuMP model in a file. JuMP can be viewed as a
-lightweight user-friendly layer on top of the MOI backend:
+problem and acts as the optimization solver. We call it an MOI *backend* and not
+optimizer as it can also be a wrapper around an optimization file format such as
+MPS that writes the JuMP model in a file. From JuMP, the MathOptInterface
+backend can be accessed using the [@ref](`JuMP.backend`) function. JuMP can be
```
blegat: Shouldn't it be the other way around ? e.g. [`JuMP.backend`](@ref) ?

src/JuMP.jl
```diff
@@ -269,30 +269,48 @@ if VERSION >= v"0.7-"
...
 end
 
+"""
+    backend(model::Model)
+
+Return the lower-level MathOptInterface model that sits underneath JuMP. This
+model depends on which operating mode JuMP is in (manual, automatic, or direct),
+and whether there are any bridges in the model.
+
+If JuMP is in direct mode (i.e., the model was created using `direct_model`),
```
blegat: Use `@ref` for `direct_model`

PR #1592, opened 2018-11-08
@constraint with <= ... <= promote to affexpr

### LINKED ISSUE #1586: It is confusing when JuMP modifies the variable bound as opposed to adding a constraint
Is this an intended part of the JuMP design?
```julia
julia> model = JuMP.direct_model(Gurobi.Optimizer())
A JuMP Model

julia> @variable(model, x >= 0)  # Sets the lower bound.
x

julia> JuMP.lower_bound(x)
0.0

julia> @constraint(model, x >= -1) # Adds a constraint.
x >= -1.0

julia> JuMP.lower_bound(x)
0.0

julia> @constraint(model, -2 <= x <= 1) # Sets both bounds.
x in [-2.0, 1.0]

julia> JuMP.lower_bound(x)
-2.0
```

blegat:
Closes #1586

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -435,11 +435,21 @@ end
 # end
 
 # three-argument build_constraint is used for two-sided constraints.
-build_constraint(_error::Function, func::AbstractJuMPScalar, lb::Real, ub::Real) = build_constraint(_error, func, MOI.Interval(lb, ub))
-# To create `MOI.SingleVariable`-in-`MOI.Interval`, use
-# `@constraint(model, var in MOI.Interval(lb, ub)`.
-# See https://github.com/JuliaOpt/JuMP.jl/issues/1586
-build_constraint(_error::Function, func::AbstractVariableRef, lb::Real, ub::Real) = build_constraint(_error, 1.0func, lb, ub)
+function build_constraint(_error::Function, func::AbstractJuMPScalar,
+                          lb::Real, ub::Real)
+    return build_constraint(_error, func, MOI.Interval(lb, ub))
+end
+
+# This method intercepts `@constraint(model, lb <= var <= ub)` and promotes
+# `var` to an `AffExpr` to form a `ScalarAffineFunction-in-Interval` instead of
+# `SingleVariable-in-Interval`. To create a
+# `MOI.SingleVariable`-in-`MOI.Interval`, use
+# `@constraint(model, var in MOI.Interval(lb, ub))`. We do this for consistency
+# with how one-sided (in)equality constraints are parsed.
+function build_constraint(_error::Function, func::AbstractVariableRef,
+                          lb::Real, ub::Real)
+    return build_constraint(_error, 1.0func, lb, ub)
+end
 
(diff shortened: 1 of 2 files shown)


odow:
Oops. The documentation failure is my fault.

src/macros.jl
```diff
@@ -436,6 +436,10 @@ end
 
 # three-argument build_constraint is used for two-sided constraints.
 build_constraint(_error::Function, func::AbstractJuMPScalar, lb::Real, ub::Real) = build_constraint(_error, func, MOI.Interval(lb, ub))
+# To create `MOI.SingleVariable`-in-`MOI.Interval`, use
+# `@constraint(model, var in MOI.Interval(lb, ub)`.
+# See https://github.com/JuliaOpt/JuMP.jl/issues/1586
```
odow: This comment could be expanded to say that it intercepts `@constraint(model, lb <= var <= ub)` and promotes `var` to an `AffExpr` to form a `ScalarAffineFunction-in-Interval` instead of `SingleVariable-in-Interval`.

The multiplication by `1.0` is a bit subtle at first glance.

src/macros.jl
```diff
@@ -436,6 +436,10 @@ end
 
 # three-argument build_constraint is used for two-sided constraints.
 build_constraint(_error::Function, func::AbstractJuMPScalar, lb::Real, ub::Real) = build_constraint(_error, func, MOI.Interval(lb, ub))
+# To create `MOI.SingleVariable`-in-`MOI.Interval`, use
+# `@constraint(model, var in MOI.Interval(lb, ub)`.
```
mlubin: Nit: Missing close parentheses

src/macros.jl
mlubin: Nit: We shouldn't be littering the code with issue links. In this case the conclusion of the issue can be summarized concisely (we do this for consistency with how one-sided (in)equality constraints are parsed).

src/macros.jl
```diff
@@ -436,6 +436,10 @@ end
...
+# See https://github.com/JuliaOpt/JuMP.jl/issues/1586
+build_constraint(_error::Function, func::AbstractVariableRef, lb::Real, ub::Real) = build_constraint(_error, 1.0func, lb, ub)
```
mlubin: Nit: line length

test/macros.jl
```diff
@@ -119,17 +119,20 @@ function macros_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Typ
 
         cref = @constraint(m, -1 <= x <= 1)
         c = JuMP.constraint_object(cref)
-        @test c.func == x
+        @test c.func isa JuMP.GenericAffExpr # see #1586
```
mlubin: Same comment about the issue links. It doesn't seem necessary to tag these lines. The point is not very subtle and should be explained sufficiently in the comment on `build_constraint`. If I can caricature this practice, imagine 
```
@test 1 + 1 == 2  # see #1234, #1356, #10657
```

PR #1597, opened 2018-11-10
RFC: docs, docs, docs, docs!

odow:
- [x] linear constraints
- [x] constraint containers
- [x] vectorized constraints
- [ ] gettting the MOI function/set
- [x] quadratic constraints
- [ ] general MOI function/set constraints
- [ ] PSD constraints
- [x] SOS1, SOS2
- [x] Variablewise ZeroOne, Integer, Semiinteger, etc
- [x] modification
- [x] deletion
- [x] linear duality
- [ ] more duality?
- [ ] attributes
 - ConstraintDualStart
 - ConstraintName
 - ConstraintSet
 - ConstraintFunction

### DIFF
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -1,5 +1,111 @@
-Constraints
-===========
+```@meta
+CurrentModule = JuMP
+DocTestSetup = quote
+    using JuMP
+end
+DocTestFilters = [r"≤|<=", r"≥|>=", r" ∈ | in "]
+```
+
+# Constraints
+
+This page explains how to write various types of constraints in JuMP. Before
+reading further, please make sure you are familiar with JuMP models, and JuMP
+[Variables](@ref).
+
+JuMP is based on the MathOptInterface API. Because of this, JuMP thinks of a
+constraint as a *function* belonging to a *set*. For example, instead of
+thinking  about a constraint $a^\top x \le b$ as a *less-than-or-equal-to*
+constraint, JuMP thinks about this as the *scalar affine* function $a^\top x$
+belonging to the *less-than* set $(-\infty, b]$. Thus, instead of a
+*less-than-or-equal-to* constraint, we consider this constraint to be a  *scalar
+affine -in- less than* constraint. More generally, we use the shorthand
+*function-in-set* to refer to constraints composed of different types of
+functions and sets. In the rest of this page, we will introduce the different
+types of functions and sets that JuMP knows about as needed. You can also more
+details about this *function-in-set* concept in the MathOptInterface
+documentation.
+
+## The `@constraint` macro
+
+Constraints are added to a JuMP model using the [`@constraint`](@ref) macro. It
+is similar to the [`@variable`](@ref) macro. Here is an example of how to add
+the constraint `2x <= 1` to a JuMP model:
+```jldoctest con1; setup = :(model = JuMP.Model(); @variable(model, x))
+julia> @constraint(model, con, 2x <= 1)
+con : 2 x <= 1.0
+```
+Wasn't that easy! Let's unpack what happened, because just like
+[`@variable`](@ref) there are a few subtle things going on.
+ 1. The mathematical constraint $2x \le 1$ was added to the model.
+ 2. A JuMP called `con` was created that acts as a reference to the
+    constraint.
+ 3. JuMP set the *name* of the constraint to `"con"`.
+
+Just like the Julia variables created in [`@variable`](@ref), `con` can be bound
+to a different value. For example:
+```jldoctest con1
+julia> con
+con : 2 x <= 1.0
+
+julia> con = 1
+1
+
+julia> con
+1
+```
+However, the reference can be retrieved by querying the model using the symbolic
+name:
+```jldoctest con1
+julia> con = model[:con]
+con : 2 x <= 1.0
+
+julia> con
+con : 2 x <= 1.0
+```
+Because the named variables and constraints are stored in the same namespace,
+creating a constraint with the same name as a variable or an existing constraint
+will result in an error. To overcome this limitation, it is possible to create
+anonymous constraints, just like it is possible to create
+[Anonymous JuMP variables](@ref). This is done by dropping the second argument
+to [`@constraint`](@ref):
+```jldoctest con1
+julia> con = @constraint(model, 2x <= 1)
+2 x <= 1.0
+```
+
+It is also possible use different comparison operators (e.g., `>=` and `==`) to
+create the following types of constraints:
+```jldoctest con1
+julia> @constraint(model, 2x >= 1)
+2 x >= 1.0
+
+julia> @constraint(model, 2x == 1)
+2 x == 1.0
+
+julia> @constraint(model, 1 <= 2x <= 3)
+2 x ∈ [1.0, 3.0]
+```
+
+## Constraint containers
+
+So far, we've adding constraints one-by-one. However, just like
+[Variable containers](@ref), JuMP provides a mechanism for building groups of
+constraints in *containers*. Three types of constraint containers are supported:
+`Array`s, `JuMPArray`s, and `Dict`ionaries. We explain each of these in the
+following.
+
+### Arrays
+
+```jldoctest constraint_arrays; setup=:(model=Model(); @variable(model, x))
+julia> @constraint(model, con[i = 1:2], i * x <= i + 1)
+2-element Array{ConstraintRef{Model,C,Shape} where Shape<:JuMP.AbstractShape where C,1}:
+ con[1] : x <= 2.0
+ con[2] : 2 x <= 3.0
+```
+
+### JuMPArrays
+
+### Dictionaries
 
 DRAFT: Describe how constraints are represented (link to MOI docs). Constraints
 are very similar to variables in (1) how names work (2) how attributes work, and
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -10,13 +10,14 @@ DocTestFilters = [r"≤|<=", r"≥|>=", r" ∈ | in "]
 
 This page explains how to write various types of constraints in JuMP. Before
 reading further, please make sure you are familiar with JuMP models, and JuMP
-[Variables](@ref).
+[Variables](@ref). If you want to add nonlinear constraints, read
+[Nonlinear Modeling](@ref) instead.
 
 JuMP is based on the MathOptInterface API. Because of this, JuMP thinks of a
 constraint as a *function* belonging to a *set*. For example, instead of
-thinking  about a constraint $a^\top x \le b$ as a *less-than-or-equal-to*
-constraint, JuMP thinks about this as the *scalar affine* function $a^\top x$
-belonging to the *less-than* set $(-\infty, b]$. Thus, instead of a
+thinking  about a constraint ``a^\top x \le b`` as a *less-than-or-equal-to*
+constraint, JuMP thinks about this as the *scalar affine* function ``a^\top x``
+belonging to the *less-than* set ``(-\infty, b]``. Thus, instead of a
 *less-than-or-equal-to* constraint, we consider this constraint to be a  *scalar
 affine -in- less than* constraint. More generally, we use the shorthand
 *function-in-set* to refer to constraints composed of different types of
@@ -29,14 +30,14 @@ documentation.
 
 Constraints are added to a JuMP model using the [`@constraint`](@ref) macro. It
 is similar to the [`@variable`](@ref) macro. Here is an example of how to add
-the constraint `2x <= 1` to a JuMP model:
+the constraint ``2x \le 1`` to a JuMP model:
 ```jldoctest con1; setup = :(model = JuMP.Model(); @variable(model, x))
 julia> @constraint(model, con, 2x <= 1)
 con : 2 x <= 1.0
 ```
 Wasn't that easy! Let's unpack what happened, because just like
 [`@variable`](@ref) there are a few subtle things going on.
- 1. The mathematical constraint $2x \le 1$ was added to the model.
+ 1. The mathematical constraint ``2x \le 1`` was added to the model.
  2. A JuMP called `con` was created that acts as a reference to the
     constraint.
  3. JuMP set the *name* of the constraint to `"con"`.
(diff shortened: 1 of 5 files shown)


docs/src/constraints.md
```diff
@@ -1,5 +1,111 @@
...
+[Variable containers](@ref), JuMP provides a mechanism for building groups of
+constraints in *containers*. Three types of constraint containers are supported:
+`Array`s, `JuMPArray`s, and `Dict`ionaries. We explain each of these in the
+following.
+
+### Arrays
+
+```jldoctest constraint_arrays; setup=:(model=Model(); @variable(model, x))
+julia> @constraint(model, con[i = 1:2], i * x <= i + 1)
+2-element Array{ConstraintRef{Model,C,Shape} where Shape<:JuMP.AbstractShape where C,1}:
```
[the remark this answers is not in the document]
mlubin: I'm not sure about overloading `show` for specializations of built-in types like Arrays. It might break expectations or make the types seem more magical than they are.

docs/src/constraints.md
```diff
@@ -1,5 +1,261 @@
...
+```jldoctest constraint_arrays
+julia> @constraint(model, [i = 1:2], i * x <= i + 1)
+2-element Array{ConstraintRef{Model,C,Shape} where Shape<:JuMP.AbstractShape where C,1}:
+ x <= 2.0
+ 2 x <= 3.0
+```
+
+Just like [`@variable`](@ref), JuMP will form an `Array` of constraints when it
+can determine at compile time that the indices are one-based integer ranges.
+Therefore `con[1:b]` will work, but `con[a:b]` will throw an error. If JuMP
```
mlubin: This will only throw an error if you request the container to be `Array`. Otherwise it will return a `JuMPArray`.

docs/src/constraints.md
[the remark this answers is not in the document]
mlubin: @blegat Do we know the shape at macro time so that we can fill in the `Shape` parameter?

docs/src/constraints.md
```diff
@@ -1,5 +1,261 @@
...
+```jldoctest constraint_jumparrays; setup=:(model=Model(); @variable(model, x))
+julia> @constraint(model, con[i = 1:2, j = 1:2; i != j], i * x <= j + 1)
+Dict{Any,ConstraintRef{Model,C,Shape} where Shape<:JuMP.AbstractShape where C} with 2 entries:
+  (1, 2) => con[1,2] : x <= 3.0
+  (2, 1) => con[2,1] : 2 x <= 2.0
+```
+
+### Forcing the container type
+
+When creating a container of constraints, JuMP will attempt to choose the
```
mlubin: We should only have this discussion once. Can we just say that it's analogous to http://www.juliaopt.org/JuMP.jl/dev/variables/#Forcing-the-container-type-1?

odow:
Phew! Okay I think I'm going to take a break here. This is a pretty substantial improvement compared to where we were a few hours ago. There are still a few items unchecked, but I might resume this tomorrow.

I also might leave the semidefinite stuff to @blegat / @chriscoey because I have no idea what to write.

docs/src/constraints.md
```diff
@@ -1,5 +1,510 @@
...
+reading further, please make sure you are familiar with JuMP models, and JuMP
+[Variables](@ref). If you want to add nonlinear constraints, read
+[Nonlinear Modeling](@ref) instead.
+
+JuMP is based on the MathOptInterface API. Because of this, JuMP thinks of a
+constraint as a *function* belonging to a *set*. For example, instead of
+thinking  about a constraint ``a^\top x \le b`` as a *less-than-or-equal-to*
+constraint, JuMP thinks about this as the *scalar affine* function ``a^\top x``
+belonging to the *less-than* set ``(-\infty, b]``. Thus, instead of a
+*less-than-or-equal-to* constraint, we consider this constraint to be a  *scalar
```
blegat: Two spaces before scalar

docs/src/constraints.md
```diff
@@ -1,5 +1,510 @@
...
+*function-in-set* to refer to constraints composed of different types of
+functions and sets. In the rest of this page, we will introduce the different
+types of functions and sets that JuMP knows about as needed. You can also more
+details about this *function-in-set* concept in the MathOptInterface
+documentation.
+
+!!! note
+    Throughout this page (and these docs), we use `MOI` as a shorthand for the
+    `MathOptInterface` module. This can be created by including the line `const
+    MOI = JuMP.MathOptInterface` after `using JuMP` in your code.
```
blegat: Why not `MOI = JuMP.MOI` ? People will prefer something shorter

odow: Updated. But I have a question. What happens if JuMP uses a different version of MOI to the one the user gets from `using MathOptInterface`? Can they interact?

mlubin: How is it possible for JuMP to use a different version of MOI than the one from `using MathOptInterface`?

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+
+julia> JuMP.optimize!(model)
+
+julia> JuMP.dual(con)
+-2.0
+```
+
+To help users who may be less familiar with conic duality, JuMP provides the
+[`JuMP.shadow_price`](@ref) function which returns a value that can be
+interpreted as the improvement in the objective given a 1-unit change in the
```
mlubin: an infinitesimal change (on the scale of one unit), not a 1-unit change

docs/src/constraints.md
odow: Hmm, for some reason I thought Pkg3 allowed JuMP to depend upon a different version of a package compared to the one I add. Never mind.

docs/src/constraints.md
mlubin: s/change/relaxation/

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+```jldoctest; setup=:(model=Model(); @variable(model, x))
+julia> @constraint(model, 2x + 1 <= 4x + 4)
+-2 x <= 3.0
+```
+
+## [Duality](@id constraint_duality)
+
+JuMP adopts the notion of duality from MathOptInterface. Roughly speaking, the
+dual variable associated with a constraint can be viewed as the decrease in the
+value of the objective given an infinitesimal relaxation of the constraint. For
```
mlubin: This relaxation logic corresponds to how I understand `shadow_price`. The only difference between this text and the definition of `shadow_price` is that "the decrease in the objective" becomes "the improvement in the objective". Was that intended?

The relaxation argument also doesn't extend to conic constraints (how do you relax membership in a cone?). The heuristic that we should explain is that a feasible dual on a `>=` constraint is nonnegative and a feasible dual on a `<=` constraint is nonpositive. That should be enough for readers familiar with LP to figure it out.

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+example, in a linear constraint, this relaxation is equivalent to `2x <= 1 + δ`
+and `2x >= 1 - δ`. If the constraint is an equality constraint, it depends on
+which direction is binding.
+
+Notably, this definition is independent of the objective sense. That is, the
+dual associated with a constraint is the same when the objective is a
+maximization and when it is a minimization. **This is different to traditional
```
mlubin: This is a bit fuzzy. When you flip the objective sense you usually get a different optimal primal/dual solution so the dual would definitely depend on the sense in that way. The difference from LP textbooks is that the feasible sign of a dual depends on the direction of the constraint and not on the objective sense.

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+duality in linear programming.**
+
+The dual value associated with a constraint can be accessed using the
```
mlubin: ... associated with a constraint in the most recent solution. Should mention `has_dual` to see if you can query the dual.

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+julia> @objective(model, Max, 2x)
+2 x
+
+julia> JuMP.optimize!(model)
+
+julia> JuMP.dual(con)
+-2.0
+```
+
+To help users who may be less familiar with conic duality, JuMP provides the
```
mlubin: Conic duality hasn't been mentioned anywhere up to this point (although it should be)

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+
+In the example above, `JuMP.dual(con)` returned `-2.0` regardless of the
+optimization sense. However, in the second case when the optimization sense is
+`Max`, [`JuMP.shadow_price`](@ref) returns:
+```jldoctest con_duality
+julia> JuMP.shadow_price(con)
+2.0
+```
+
+To query the dual variables associated a variable bounds, first obtain a
```
mlubin: Typo: "a variable bound"

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+```@meta
+DocTestSetup = quote
+    using JuMP
+    const MOI = JuMP.MathOptInterface
+end
+```
+
+## Constraint containers
+
+So far, we've adding constraints one-by-one. However, just like
```
mlubin: Missing word in "we've adding"

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+[Variable containers](@ref), JuMP provides a mechanism for building groups of
+constraints in *containers*. Three types of constraint containers are supported:
+`Array`s, `JuMPArray`s, and `Dict`ionaries. We explain each of these in the
```
mlubin: The emphasis seems a bit reversed. The point of this syntax is to let you add groups of constraints compactly, and as a side effect we return containers of references that let you index into the constraints. The constraints themselves aren't in containers; the references are.

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+following.
+
+### [Arrays](@id constraint_arrays)
+
+The syntax for adding `Array`s of JuMP constraints is very similar to the
```
mlubin: Related to the comment above, this isn't adding arrays of constraints. This is adding a block of constraints and returning an array of references.

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+ 6
+
+julia> @constraint(model, con, A * x .== b)
+2-element Array{ConstraintRef{Model,MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.EqualTo{Float64}},JuMP.ScalarShape},1}:
+ x[1] + 2 x[2] == 5.0
+ 3 x[1] + 4 x[2] == 6.0
+```
+
+!!! note
+    Make sure to use the broadcasting syntax `.==` (or `.>=` and `.<=`). If you
```
mlubin: Readers might not know what broadcasting means.

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+can instead construct a `VectorAffineFunction-in-Nonnegatives` constraint as
+follows:
+```jldoctest con_vector
+julia> @constraint(model, A * x - b in MOI.Nonnegatives(2))
+[x[1] + 2 x[2] - 5, 3 x[1] + 4 x[2] - 6] in MathOptInterface.Nonnegatives(2)
+```
+
+In addition to the `Nonnegatives` set, MathOptInterface defines a number of
+other vector-valued sets such as `Nonpositives`. See the [MathOptInterface
+documentation](http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#Sets-1)
```
mlubin: Link to versioned docs

docs/src/constraints.md
```diff
@@ -1,5 +1,514 @@
...
+In addition to the second order cone and rotated second order cone,
+MathOptInterface defines a number of other conic sets such as the exponential
+and power cones. See the [MathOptInterface documentation](http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#Sets-1)
+for more information.
+
+## Semidefinite constraints
+
+TODO: discuss [`@SDconstraint`] and [`PSDCone`].
+
+## Constraint modifications
```
mlubin: Consider subsections for different types of modifications

docs/src/variables.md
```diff
@@ -256,9 +256,11 @@ julia> JuMP.lower_bound.(x)
 
 JuMP will form an `Array` of JuMP variables when it can determine at compile
 time that the indices are one-based integer ranges. Therefore `x[1:b]` will
-work, but `x[a:b]` will throw an error.
+work, but `x[a:b]` will not. If JuMP cannot determine that the indices are
```
mlubin: This wording is still ambiguous. "works" in this sentence means provides sufficient information for JuMP to conclude that it's correct to create an `Array`.

Maybe we need a table with different examples of what index sets will default to which containers (and which are also possible with the `container=` keyword). (Possible TODO.)

docs/src/constraints.md
odow: I shouldn't write this sort of stuff on a Saturday night... My brain was fuzzy.

docs/src/constraints.md
```diff
@@ -1,5 +1,531 @@
...
+is similar to the [`@variable`](@ref) macro. Here is an example of how to add
+the constraint ``2x \le 1`` to a JuMP model:
+```jldoctest con1; setup = :(model = Model(); @variable(model, x))
+julia> @constraint(model, con, 2x <= 1)
+con : 2 x <= 1.0
+```
+Wasn't that easy! Let's unpack what happened, because just like
+[`@variable`](@ref) there are a few subtle things going on.
+ 1. The mathematical constraint ``2x \le 1`` was added to the model.
+ 2. A JuMP called `con` was created that acts as a reference to the
```
blegat: "A JuMP called" ->missing something between "JuMP" and "called"

docs/src/constraints.md
```diff
@@ -1,5 +1,531 @@
...
+```jldoctest constraint_arrays
+julia> @constraint(model, [i = 1:2], i * x <= i + 1)
+2-element Array{ConstraintRef{Model,C,Shape} where Shape<:JuMP.AbstractShape where C,1}:
+ x <= 2.0
+ 2 x <= 3.0
+```
+
+Just like [`@variable`](@ref), JuMP will form an `Array` of constraints when it
+can determine at compile time that the indices are one-based integer ranges.
+Therefore `con[1:b]` will create an `Array`, but `con[a:b]` will not. If JuMP
```
blegat: It is unclear why for `1:b`, we can see at compile time that it is one-based because it has the same type than `2:b`. Maybe "parsing time" instead of "compile time" ?

docs/src/constraints.md
```diff
@@ -1,5 +1,531 @@
...
+Dict{Any,ConstraintRef{Model,C,Shape} where Shape<:JuMP.AbstractShape where C} with 2 entries:
+  (1, 2) => con[1,2] : x <= 3.0
+  (2, 1) => con[2,1] : 2 x <= 2.0
+```
+
+### Forcing the container type
+
+When creating a container of constraints, JuMP will attempt to choose the
+tightest container type that can store the constraints. However, because this
+happens at compile time, it does not always make the best choice. Just like in
```
blegat: "parse time" or "parsing time"

docs/src/constraints.md
blegat: In https://docs.julialang.org/en/v1/manual/metaprogramming/#Hold-up:-why-macros?-1, they say

> Macros are necessary because they execute when code is parsed

so it seems that "parse" is accurate

odow:
Once https://github.com/JuliaOpt/JuMP.jl/pull/1616 is merged, this PR should be rebased on top and:
- [ ] change the syntax of `has_dual`
- [ ] update the `has_dual` reference in text
- [ ] add `has_dual` to the `@docs` block at the bottom.

docs/src/constraints.md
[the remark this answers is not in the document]
blegat: In most of the case yes. If someone does `@constraint(model, c[i=1:2], f[i] in s[i])` we cannot infer `C` nor `Shape` though. Even with `@constraint(model, f .== g)` we could say that the shape is `ScalarShape` and `C` is `MOI.ConstraintIndex{???, MOI.EqualTo{Float64})` so we are missing information. We might need to do some magic trick like what base does for determining the type of the output container when doing `[... for ... in ...]` or with broadcast and `map`. But that will be a compile time or run time, not macro time/parse time.

mlubin:
Is this ready for another review?

odow:
Docs are still failing because of https://github.com/JuliaOpt/JuMP.jl/issues/1618. We just need a new tag of MOI.

docs/src/quickstart.md
```diff
@@ -163,6 +163,10 @@ julia> JuMP.dual(con)
 -0.6
 ```
 
+!!! info
+    If you're confused about the sign associated with the dual value, read the
```
mlubin: Maybe `"See [constraint duality section](@ref constraint_duality) for a discussion of the convention that JuMP uses for signs of duals".`

It's not necessary to be confused in order to want to know about signs of duals.

odow:
Okay, let's try this; I've rebased onto the latest master and it is passing locally.

blegat:
Merging to be able to rebase https://github.com/JuliaOpt/JuMP.jl/pull/1657 on top (see https://github.com/JuliaOpt/JuMP.jl/pull/1657#discussion_r239977592)

PR #1604, opened 2018-11-11
Performance improvements in linear_terms and quad_terms

mlubin:
Ref https://discourse.julialang.org/t/allocations-when-manipulating-an-iterator/17405 and https://github.com/JuliaOpt/JuMP.jl/issues/1403#issuecomment-437630140.

I see a 5x speedup in `JuMP.moi_function(::AffExpr)` and a 3.3x speedup in `JuMP.moi_function(::QuadExpr)`. This removes 2 of 5 allocations per term in the minipmed example (https://github.com/JuliaOpt/JuMP.jl/issues/1403#issuecomment-437626804).
Before:
```
0.656219 seconds (5.00 M allocations: 568.758 MiB, 37.93% gc time)
```
After:
```
0.564028 seconds (3.00 M allocations: 507.723 MiB, 30.32% gc time)
```
There are still 3 more allocations per term to track down before we can (hopefully) get back to 0.18's performance.

### DIFF
--- a/src/quad_expr.jl
+++ b/src/quad_expr.jl
@@ -112,7 +112,7 @@ if VERSION < v"0.7-"
     Base.done(qti::QuadTermIterator, state::Int) = done(qti.quad.terms, state)
     Base.next(qti::QuadTermIterator, state::Int) = reorder_iterator(next(qti.quad.terms, state)...)
 else
-    function reorder_and_flatten(p::Pair{<:UnorderedPair,<:Any})
+    function reorder_and_flatten(p::Pair{<:UnorderedPair})
         return (p.second, p.first.a, p.first.b)
     end
     function Base.iterate(qti::QuadTermIterator)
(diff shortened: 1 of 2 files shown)


src/quad_expr.jl
```diff
@@ -112,15 +112,31 @@ if VERSION < v"0.7-"
     Base.done(qti::QuadTermIterator, state::Int) = done(qti.quad.terms, state)
     Base.next(qti::QuadTermIterator, state::Int) = reorder_iterator(next(qti.quad.terms, state)...)
 else
-    function reorder_iterator(t::Tuple{Pair{<:UnorderedPair,<:Any},Int})
-        ((first(t).second, first(t).first.a, first(t).first.b), last(t))
+    function reorder_and_flatten(p::Pair{<:UnorderedPair,<:Any})
```
blegat: `Pair{<:UnorderedPair,<:Any}` is equivalent to `Pair{<:UnorderedPair}` which is arguably simpler/more readable

PR #1606, opened 2018-11-11
Add example of a user-defined function in constraint

### LINKED ISSUE #1323: Docs: add example with user defined function in constraint
Leaving a breadcrumb to:
https://discourse.julialang.org/t/how-to-add-user-defined-nonlinear-functions-as-constraints-with-jump/11436

This would be good to address as it is unlikely to be the only time it gets asked.

odow:
Closes #1323

### DIFF
--- a/docs/src/nlp.md
+++ b/docs/src/nlp.md
@@ -1,5 +1,4 @@
-Nonlinear Modeling
-==================
+# Nonlinear Modeling
 
 ```@meta
 CurrentModule = JuMP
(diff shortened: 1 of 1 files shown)


mlubin:
The confusion in https://discourse.julialang.org/t/how-to-add-user-defined-nonlinear-functions-as-constraints-with-jump/11436 looks to be about functions with vector inputs. What's missing is an example of the new splatting syntax.

odow:
Updated.

PR #1616, opened 2018-11-13
Change has_dual to operate on the constraint object.

odow:
> Why does this method depend on the type of the constraint?

Answer: I have no idea. @mlubin introduced it in https://github.com/JuliaOpt/JuMP.jl/pull/1158/commits/16f867a5689ffba2b9ed6411bafb4254f042a051, but it doesn't seem needed so I reverted back to the object.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -428,18 +428,14 @@ end
 index(cr::ConstraintRef) = cr.index
 
 """
-    has_dual(con_ref::ConstraintRef)
+    has_duals(model::Model)
 
 Return true if the solver has a dual solution available to query, otherwise
 return false.
 
 See also [`dual`](@ref) and [`shadow_price`](@ref).
 """
-function has_dual(con_ref::ConstraintRef{Model, <:MOICON})
-    # TODO(odow): should we be more clever here? What about solvers that only
-    # support duals on a subset of their constraints?
-    return MOI.get(con_ref.model, MOI.DualStatus()) != MOI.NoSolution
-end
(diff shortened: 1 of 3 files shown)


src/JuMP.jl
```diff
@@ -427,11 +427,18 @@ end
...
-    MOI.get(model, MOI.DualStatus()) != MOI.NoSolution
+"""
+    has_dual(con_ref::ConstraintRef)
+
+Return true if the solver has a dual solution available to query, otherwise
+return false.
+
+See also [`dual`](@ref) and [`shadow_price`](@ref).
+"""
+function has_dual(con_ref::ConstraintRef{Model, <:MOICON})
```
mlubin: Since `DualStatus()` is a property of the model and not the ConstraintRef, shouldn't this be `has_dual(::Model)`? That also makes it less misleading to say that the model has duals when potentially not all constraints do.

Consider also renaming to `has_duals` if it becomes a property of the model.

blegat: Then if latter we support duals for some constraint but not all in MOI we might use `has_dual` for the constraint-wise function and `has_duals` will still be for the model.
`has_dual(::ConstraintRef)` made sense when we had `canget` but now it might be wrong because if some constraints do not have dual while the `DualStatus` is not `NoSolution`, the solver might error so we don't want to give constraint-wise guarantee.

odow:
We now implement `has_duals`. It turns out there was already `has_values`. They have v. similar definitions.

Issue #1618, opened 2018-11-13
Error message: xxx_status


odow:
Also `primal_status` etc.
```julia
model = Model()
JuMP.dual_status(model)

julia> JuMP.dual_status(model)
ERROR: MethodError: no method matching get(::Nothing, ::MathOptInterface.DualStatus)
Closest candidates are:
  get(::Model, ::MathOptInterface.AbstractModelAttribute) at C:\Users\Oscar\.julia\packages\JuMP\t3yHD\src\JuMP.jl:453
  get(::Ipopt.Optimizer, ::MathOptInterface.DualStatus) at C:\Users\Oscar\.julia\packages\Ipopt\45sRy\src\MOIWrapper.jl:783
  get(::MathOptInterface.Utilities.MockOptimizer, ::MathOptInterface.DualStatus) at C:\Users\Oscar\.julia\packages\MathOptInterface\QO3Mk\src\Utilities\mockoptimizer.jl:162
  ...
Stacktrace:
 [1] get(::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.UniversalFallback{JuMP.JuMPMOIModel{Float64}}}, ::MathOptInterface.DualStatus) at C:\Users\Oscar\.julia\packages\MathOptInterface\QO3Mk\src\Utilities\cachingoptimizer.jl:433
 [2] get(::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.UniversalFallback{JuMP.JuMPMOIModel{Float64}}},MathOptInterface.Bridges.AllBridgedConstraints{Float64}}, ::MathOptInterface.DualStatus) at C:\Users\Oscar\.julia\packages\MathOptInterface\QO3Mk\src\Bridges\bridgeoptimizer.jl:172
 [3] get(::Model, ::MathOptInterface.DualStatus) at C:\Users\Oscar\.julia\packages\JuMP\t3yHD\src\JuMP.jl:453
 [4] dual_status(::Model) at C:\Users\Oscar\.julia\packages\JuMP\t3yHD\src\JuMP.jl:341
 [5] top-level scope at none:0
```

odow:
[...]

PR #1619, opened 2018-11-13
Add docstrings for xxx_status.

odow:
In my quest to solve #1618, I found that these weren't documented.

The question remains how to provide nicer error message when the model hasn't been solved yet.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -336,6 +336,7 @@ end
 
 # TODO(IainNZ): Document these too.
 object_dictionary(model::Model) = model.obj_dict
+
 """
     termination_status(model::Model)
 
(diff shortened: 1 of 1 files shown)


src/JuMP.jl
```diff
@@ -336,9 +336,36 @@ end
 
 # TODO(IainNZ): Document these too.
 object_dictionary(model::Model) = model.obj_dict
-termination_status(model::Model) = MOI.get(model, MOI.TerminationStatus())
-primal_status(model::Model) = MOI.get(model, MOI.PrimalStatus())
-dual_status(model::Model) = MOI.get(model, MOI.DualStatus())
+"""
```
mlubin: Nit: missing new line before docstring

PR #1626, opened 2018-11-16
✨ Add support for solver at optimize with bridges

### LINKED ISSUE #1504: Better error for unsupported constraint when the optimizer is provided at `optimize!`
When the optimizer is given in the `Model` constructor, unsupported added constraint are either bridged or throw an helpful error:

https://github.com/JuliaOpt/JuMP.jl/blob/500945cac4949dfd9e30cb0ca68450f19f4ff34d/src/constraints.jl#L202

If it is given in the `optimize!` call, a cryptic error is thrown instead.
See https://github.com/JuliaOpt/SCS.jl/issues/112

blegat:
Before this PR, the MOI layers on top of the optimizer where:
Bridges -> CachingOptimizer
Now if the optimizer `supports_default_copy_to`, it is:
CachingOptimizer -> Bridges
otherwise, it is:
CachingOptimizer -> Bridges -> CachingOptimizer

When a model is created without any optimizer we have simply:
CachingOptimizer
and the bridges are added with the optimizer when it is added. This allows bridges to work when the optimizer is added after model creation.

Closes #1504

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -134,8 +134,8 @@ end
 # Model
 
 # Model has three modes:
-# 1) Automatic: moi_backend field holds a LazyBridgeOptimizer{CachingOptimizer} in Automatic mode.
-# 2) Manual: moi_backend field holds a LazyBridgeOptimizer{CachingOptimizer} in Manual mode.
+# 1) Automatic: moi_backend field holds a CachingOptimizer in Automatic mode.
+# 2) Manual: moi_backend field holds a CachingOptimizer in Manual mode.
 # 3) Direct: moi_backend field holds an AbstractOptimizer. No extra copy of the model is stored. The moi_backend must support add_constraint etc.
 # Methods to interact with the CachingOptimizer are defined in solverinterface.jl.
 @enum ModelMode Automatic Manual Direct
@@ -157,7 +157,7 @@ mutable struct Model <: AbstractModel
     variable_to_fix::Dict{MOIVAR, MOIFIX}
     variable_to_integrality::Dict{MOIVAR, MOIINT}
     variable_to_zero_one::Dict{MOIVAR, MOIBIN}
-    # In Manual and Automatic modes, LazyBridgeOptimizer{CachingOptimizer}.
+    # In Manual and Automatic modes, CachingOptimizer.
     # In Direct mode, will hold an AbstractOptimizer.
     moi_backend::MOI.AbstractOptimizer
     # Hook into a solve call...function of the form f(m::Model; kwargs...),
@@ -184,11 +184,10 @@ a cache. The mode of the `CachingOptimizer` storing this cache is
 `caching_mode`. The optimizer can be set later in the [`JuMP.optimize!`](@ref)
 call. If `bridge_constraints` is true, constraints that are not supported by the
 optimizer are automatically bridged to equivalent supported constraints when
-an appropriate is defined in the `MathOptInterface.Bridges` module or is
-defined in another module and is explicitely added.
+an appropriate transformation is defined in the `MathOptInterface.Bridges`
+module or is defined in another module and is explicitely added.
 """
 function Model(; caching_mode::MOIU.CachingOptimizerMode=MOIU.Automatic,
-                 bridge_constraints::Bool=true,
                  solver=nothing)
     if solver !== nothing
         error("The solver= keyword is no longer available in JuMP 0.19 and " *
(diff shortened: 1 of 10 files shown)


src/objective.jl
```diff
@@ -115,12 +115,12 @@ However, it is not convertible to a variable.
 julia> JuMP.objective_function(model, JuMP.VariableRef)
 ERROR: InexactError: convert(MathOptInterface.SingleVariable, MathOptInterface.ScalarAffineFunction{Float64}(MathOptInterface.ScalarAffineTerm{Float64}[ScalarAffineTerm{Float64}(2.0, VariableIndex(1))], 1.0))
 Stacktrace:
- [1] convert at /home/blegat/.julia/dev/MathOptInterface/src/functions.jl:393 [inlined]
- [2] get(::JuMP.JuMPMOIModel{Float64}, ::MathOptInterface.ObjectiveFunction{MathOptInterface.SingleVariable}) at /home/blegat/.julia/dev/MathOptInterface/src/Utilities/model.jl:259
- [3] get at /home/blegat/.julia/dev/MathOptInterface/src/Utilities/universalfallback.jl:105 [inlined]
- [4] get at /home/blegat/.julia/dev/MathOptInterface/src/Utilities/cachingoptimizer.jl:436 [inlined]
+ [1] convert at /home/blegat/.julia/dev/MathOptInterface/src/functions.jl:398 [inlined]
```
odow: Surely this fails because we're not filtering out the path as well? Or does that happen somewhere else?

blegat: We are filtering out this output so I was not required to update it.

odow: Oh, I missed the `.*` at the end of the regex. `r"^Stacktrace:.*"s` might be more clear?

blegat: If I'm not mistaken, `.*` doesn't catch newlines

odow: The modifier `s` on the end makes it. I guess that's just as opaque... nm.

src/JuMP.jl
```diff
@@ -308,16 +285,34 @@ backend(model::Model) = model.moi_backend
...
+    bridge_constraints(model::Model)
+
+Return a `Bool` indicating whether the model `model` is in manual or automatic
+mode, the optimizer is set and unsupported constraints are automatically bridged
+to equivalent supported constraints when an appropriate transformation is
+available.
+"""
+function bridge_constraints(model::Model)
+    caching_optimizer = backend(model)
+    if caching_optimizer isa MOIU.CachingOptimizer
```
mlubin: Nit: `if backend(model) isa MOIU.CachingOptimizer`. It's weird call something `caching_optimizer` if it might not be one.

src/JuMP.jl
```diff
@@ -308,16 +285,34 @@ backend(model::Model) = model.moi_backend
...
         return Automatic
     else
         return Manual
     end
 end
 
+"""
+    bridge_constraints(model::Model)
+
+Return a `Bool` indicating whether the model `model` is in manual or automatic
```
mlubin: The wording is a bit confusing. Maybe split into cases. When in manual or automatic mode, returns X. When in direct mode, returns false.

src/optimizer_interface.jl
```diff
@@ -3,28 +3,57 @@
...
 function MOIU.resetoptimizer!(model::Model)
-    @assert mode(model) != Direct
-    MOIU.resetoptimizer!(caching_optimizer(model))
+    error_if_direct_mode(model, :resetoptimizer!)
+    MOIU.resetoptimizer!(backend(model))
 end
 
 function MOIU.dropoptimizer!(model::Model)
+    error_if_direct_mode(model, :dropoptimizer!)
     @assert mode(model) != Direct
```
mlubin: The assert can be dropped now

src/optimizer_interface.jl
```diff
@@ -3,28 +3,57 @@
...
-    MOIU.dropoptimizer!(caching_optimizer(model))
+    MOIU.dropoptimizer!(backend(model))
 end
 
 function MOIU.attachoptimizer!(model::Model)
+    error_if_direct_mode(model, :attachoptimizer!)
     @assert mode(model) != Direct
```
mlubin: Same here

src/JuMP.jl
blegat: I renamed it `moi_backend`. The advantage of storing it in a local variable is that inference can use the fact that it is in the first clause of the `if` to infer that the variable is a CachingOptimizer and we loose a bit of the disadvantage that `moi_backend` is not concretely typed

PR #1630, opened 2018-11-17
Implement `@Constraint` macro for unattached constraints

### LINKED ISSUE #710: Its kind of weird we export `at-LinearConstraint` and `at-QuadConstraint`
Because we don't ask users to make that distinction when using `@addConstraint`. This fits in with the "great rename" plans to some extent I think.

mlubin:
Closes #710 (because it's not exported)

@odow

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -14,8 +14,7 @@ Building the documentation
 The documentation is built using the following command:
 
 ```julia
-julia make.jl
+julia --project=. make.jl
 ```
 
 The compiled documents can be viewed at `build/index.html`.
-
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -41,3 +41,11 @@ JuMP.set_coefficient
 JuMP.dual
 JuMP.shadow_price
 ```
+
+## Constructing constraints without adding them to the model
+
+For advanced use cases.
+
+```@docs
+JuMP.@Constraint
+```
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -39,8 +39,6 @@ export
     linear_terms,
 
 # Macros and support functions
-    @LinearConstraint, @LinearConstraints, @QuadConstraint, @QuadConstraints,
-    @SOCConstraint, @SOCConstraints,
     @expression, @expressions, @NLexpression, @NLexpressions,
     @variable, @variables, @constraint, @constraints,
     @NLconstraint, @NLconstraints,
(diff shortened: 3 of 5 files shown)


odow:
I really don't like the ambiguity between `@constraint(model, 2x<=1)` and `@Constraint(2x <= 1)`. I see it becoming a source of confusion for users, particularly if we're then going to advertise it through callbacks.

Possible ideas:
`@constraint_object(2x <= 1)`
`@standalone_constraint(2x <= 1)`
`@unattached_constraint(2x <= 1)`

mlubin:
Yes, the naming is tricky. I think the underlying issue is that `@constraint` should be `@add_constraint` if we're naming things after what they do, but we don't want to rename `@constraint`.

We could pull an `undef` and try `JuMP.@constraint(unattached, 2x <= 1)` :confused:.

The names you proposed are more visually distinct but still not *actually* consistent with `@constraint` from first principles.

blegat:
It is true that `@Constraint` can easily be confused with `@constraint`. There might be users loosing a lot of time debugging trying to understand why the solver ignore their constraint not realizing that they have written a capital `C`.
I like `JuMP.@constraint(JuMP.unattached, 2x <= 1)`. We could define in JuMP
```julia
struct DummyModel <: AbstractModel
end
function constraint_type(::DummyModel)
    # Called when the user do e.g. `@constraint(unattached, [i=1:N], x <= i)`
    error("`@constraint(JuMP.unattached, container, expr) is not supporter, use `@constraint(JuMP.unattached, expr)")
end
function add_constraint(::DummyModel, ::AbstractConstraint, ::String="") end
```
With this, no need to modify any macro code :)

mlubin:
We do have to modify the macro because the non-anonymous versions and the vector versions `@constraint(unattached, [i=1:N], ...)` won't immediately work and should give useful error messages. Given that, defining an object called `JuMP.unattached` seems a bit silly when we can detect `:unattached` in the macro anyway.

blegat:
> the vector versions `@constraint(unattached, [i=1:N], ...)` won't immediately work and should give useful error messages

This will call `constraint_type(::unattached)` so we can implement this method to give the useful error as done in https://github.com/JuliaOpt/JuMP.jl/pull/1630#issuecomment-439691643

mlubin:
@odow, how about `@build_constraint`?

odow:
works for me

PR #1657, opened 2018-12-03
[WIP] Add CustomIndexing.SparseArray

blegat:
Implement solution 3. of https://github.com/JuliaOpt/JuMP.jl/pull/1537#issuecomment-443438792 
All the sophisticated indexing of JuMP macros as well as `JuMPArray` and `JuMPDict` are not relatated to optimization.
There does not seem to be any package that we could rely on of this so we have to implement everything in-house.
It seems to me that moving most of the logic into a separate submodule that is independent of the rest of the JuMP code might help readability and improve the code mainainability.
In the long term, we might move this submodule into a separate package `CustomIndexing.jl` to make it usable outside JuMP and make contribution to it easier.

In the long term, it would be nice to have a macro
```julia
@container [i=..., j=...) f(i, j)
```
in `CustomIndexing` that creates a container and fill it with `f(i, j)`. The element type would be determined from the first entry like what is done for `map` and `broadcast` in base.
This macro would rely on a function in `CustomIndexing` similar to `JuMP.generatecontainer` what JuMP could use.
As it would not need the element type as argument, we will be able to get rid of `variable_type` and `constraint_type` and have containers with concrete element types (which is not the same at the moment with `@constraint`).

TODO:
- [ ] Add helpful error for unsupported operations
- [ ] inplement indexing with `:`
- [ ] Use SparseArray in `JuMP.generatecontainer`instead of `Dict`
- [ ] Move `JuMPArray` into `CustomIndexing.DenseArray`

What do you think ?

### DIFF
--- a/src/CustomIndexing/CustomIndexing.jl
+++ b/src/CustomIndexing/CustomIndexing.jl
@@ -1,0 +1,130 @@
+module CustomIndexing
+
+"""
+    struct SparseArray{T,N} <: AbstractArray{T,N}
+        data::Dict{Tuple,T}
+    end
+
+`N`-dimensional array with elements of type `T` where only a subset of the entries are defined.
+The entries with indices `idx = (i1, i2, ..., iN)` in `keys(data)` has value `data[idx]`.
+Note that as opposed to `SparseArrays.AbstractSparseArray`, the missing entries are not assumed
+to be `zero(T)`. The indexing structure of the array is not rectangular and the missing entries
+are simply not part of the array. This means that the result of `map(f, sa::SparseArray)` or
+`f.(sa::SparseArray)` has the same sparsity structure than `sa` even if `f(zero(T))` is not zero.
+"""
+struct SparseArray{T,N} <: AbstractArray{T,N}
+    data::Dict{Tuple,T}
+end
+
+Base.length(sa::SparseArray) = length(sa.data)
+Base.iterate(sa::SparseArray, args...) = iterate(sa, args...)
+function Base.mapreduce(f, op, sa::SparseArray)
+    mapreduce(f, op, values(sa.data))
+end
+Base.:(==)(sa1::SparseArray, sa2::SparseArray) = sa1.data == sa2.data
+
+########
+# Show #
+########
+
+# Inspired from Julia SparseArrays stdlib package
+function Base.show(io::IO, ::MIME"text/plain", sa::SparseArray)
+    num_entries = length(sa.data)
+    print(io, typeof(sa), " with ", num_entries, " stored ",
+              isone(num_entries) ? "entry" : "entries")
+    if !iszero(num_entries)
+        println(io, ":")
+        show(io, sa)
+    end
+end
+Base.show(io::IO, x::SparseArray) = show(convert(IOContext, io), x)
+function Base.show(io::IOContext, x::SparseArray)
+    # TODO: make this a one-line form
+    if isempty(x)
+        return show(io, MIME("text/plain"), x)
+    end
+    limit::Bool = get(io, :limit, false)
+    half_screen_rows = limit ? div(displaysize(io)[1] - 8, 2) : typemax(Int)
+    key_string(key::Tuple) = join(key, ", ")
+    print_entry(i) = i < half_screen_rows || i > length(x) - half_screen_rows
+    pad = maximum(Int[print_entry(i) ? length(key_string(key)) : 0 for (i, key) in enumerate(keys(x.data))])
+    if !haskey(io, :compact)
+        io = IOContext(io, :compact => true)
+    end
+    for (i, (key, value)) = enumerate(x.data)
+        if print_entry(i)
+            print(io, "  ", '[', rpad(key_string(key), pad), "]  =  ", value)
+            if i != length(x)
+                println(io)
+            end
+        elseif i == half_screen_rows
+            println(io, "   ", " "^pad, "   \u22ee")
+        end
+    end
+end
+
+############
+# Indexing #
+############
+
+Base.setindex!(d::SparseArray, value, idx) = setindex!(d.data, value, idx)
+Base.getindex(d::SparseArray, idx) = getindex(d.data, idx)
+Base.eachindex(d::SparseArray) = keys(d.data)
+
+# Need to define it as indices may be non-integers
+Base.to_index(d::SparseArray, idx) = idx
+
+# Arbitrary typed indices. Linear indexing not supported.
+struct IndexAnyCartesian <: Base.IndexStyle end
+Base.IndexStyle(::IndexAnyCartesian, ::IndexAnyCartesian) = IndexAnyCartesian()
+Base.IndexStyle(::Type{<:SparseArray}) = IndexAnyCartesian()
+# eachindex redirect to keys
+Base.keys(::IndexAnyCartesian, d::SparseArray) = keys(d)
+
+################
+# Broadcasting #
+################
+
+# Need to define it as indices may be non-integers
+Base.Broadcast.newindex(d::SparseArray, idx) = idx
+
+struct BroadcastStyle{N} <: Broadcast.BroadcastStyle end
+# TODO throw an helpful error when trying to mix SparseArray with a non-scalar
+# Scalars can be used with SparseArray in broadcast
+Base.BroadcastStyle(::BroadcastStyle{N}, ::Base.Broadcast.DefaultArrayStyle{0}) where {N} = BroadcastStyle{N}()
+Base.BroadcastStyle(::Type{<:SparseArray{T, N}}) where {T, N} = BroadcastStyle{N}()
+function Base.similar(b::Base.Broadcast.Broadcasted{BroadcastStyle{N}}, ::Type{T}) where {T, N, Ax}
+    SparseArray{T, N}(Dict{Tuple, T}())
+end
+
+# Check that all SparseArrays involved have the same indices. The other
+# arguments are scalars
+function check_same_eachindex(each_index) end
+check_same_eachindex(each_index, not_sa, args...) = check_same_eachindex(eachindex, args...)
+function check_same_eachindex(each_index, sa::SparseArray, args...)
+    if Set(each_index) != Set(eachindex(sa))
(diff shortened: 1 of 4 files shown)


joaquimg:
Having this abstracted from JuMP would be a great improvement. It always annoyed me a little that we needed to have a JuMPArray.

src/CustomIndexing/CustomIndexing.jl
```diff
@@ -0,0 +1,130 @@
+module CustomIndexing
```
mlubin: Maybe `Containers`, i.e., `JuMP.Containers`?

mlubin: Consider adding a docstring on the module describing its intent. Also missing a license header.

src/CustomIndexing/CustomIndexing.jl
```diff
@@ -0,0 +1,130 @@
...
+
+"""
+    struct SparseArray{T,N} <: AbstractArray{T,N}
+        data::Dict{Tuple,T}
```
mlubin: Tuple` could be `NTuple{N}`.

src/CustomIndexing/CustomIndexing.jl
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+`N`-dimensional array with elements of type `T` where only a subset of the entries are defined.
+The entries with indices `idx = (i1, i2, ..., iN)` in `keys(data)` has value `data[idx]`.
+Note that as opposed to `SparseArrays.AbstractSparseArray`, the missing entries are not assumed
+to be `zero(T)`. The indexing structure of the array is not rectangular and the missing entries
```
mlubin: "not rectangular" is a bit unclear. I think "The missing entries are simply not part of the array." says the same thing.

mlubin: What is a "missing entry" defined as? This implicitly assumes some axis structure, no? Is there a difference between a missing entry and an out-of-bounds entry like there are for regular sparse arrays?

src/CustomIndexing/CustomIndexing.jl
```diff
@@ -0,0 +1,130 @@
+module CustomIndexing
+
+"""
+    struct SparseArray{T,N} <: AbstractArray{T,N}
```
mlubin: Does it really make sense to call this an `AbstractArray` if `size` and `axes` aren't implemented?

src/CustomIndexing/CustomIndexing.jl
```diff
@@ -0,0 +1,130 @@
...
+    return each_index
+end
+# Need to define it as it falls back to `axes` by default
+function Base.eachindex(bc::Base.Broadcast.Broadcasted{<:BroadcastStyle})
+    return _eachindex(bc.args...)
+end
+# Need to define it as `axes` is not defined
+Base.Broadcast.instantiate(bc::Base.Broadcast.Broadcasted{<:BroadcastStyle}) = bc
+# Need to define it as indices are not integers
+function Base.copyto!(dest::SparseArray{T, N}, bc::Base.Broadcast.Broadcasted{BroadcastStyle{N}}) where {T, N}
```
mlubin: Is this part of the normal broadcasting interface? Will the code break between minor Julia releases?

mlubin:
I like where this is going, but I'm a bit worried that we have to call internal broadcast functions like `_broadcast_getindex`. If we do this, we should also commit to putting in the effort to try and make the necessary changes in Julia's broadcasting system so we don't have to permanently maintain code that is likely to break in unpredictable ways between Julia minor releases.

> In the long term, it would be nice to have a macro `@container [i=..., j=...) f(i, j)`in CustomIndexing that creates a container and fill it with f(i, j). The element type would be determined from the first entry like what is done for map and broadcast in base. This macro would rely on a function in CustomIndexing similar to JuMP.generatecontainer what JuMP could use. As it would not need the element type as argument, we will be able to get rid of variable_type and constraint_type and have containers with concrete element types (which is not the same at the moment with `@constraint`).

This is a good idea but it's mostly orthogonal to the point of the PR which is how we represent dictionaries.

src/CustomIndexing/CustomIndexing.jl
blegat: Yes, I was thinking we should already think about which name the package would have if we move it to avoid breakage but users will need to add `using PackageName` anyway so we can call it `Containers` here and rename later

src/CustomIndexing/CustomIndexing.jl
blegat: Not if indices have different types, e.g. `(1, 1.0) isa NTuple{2}` -> `false`

src/CustomIndexing/CustomIndexing.jl
blegat: No, there are no difference between missing and out-of-bound entries. In fact, we should throws `BoundsError` when attempting to access such entries

src/CustomIndexing/CustomIndexing.jl
blegat: If it is not an `AbstractArray` then `map(f, d)` will store the result in a `Vector` and it's difficult to make it behave differently except by being very hacky with `Base._..` functions

src/CustomIndexing/CustomIndexing.jl
blegat: Yes, `copyto!` and `instantiate` are the place where you are expected to add new methods for new `BroadcastStyle`, it is written in their docstrings

blegat:
> I'm a bit worried that we have to call internal broadcast functions like _broadcast_getindex. If we do this, we should also commit to putting in the effort to try and make the necessary changes in Julia's broadcasting system so we don't have to permanently maintain code that is likely to break in unpredictable ways between Julia minor releases.

I agree, once we have the `Containers` set up we should check if `Base` can be made nicer

src/Containers/Containers.jl
```diff
@@ -0,0 +1,204 @@
...
+#############################################################################
+# JuMP
+# An algebraic modeling language for Julia
+# See http://github.com/JuliaOpt/JuMP.jl
+#############################################################################
+
+"""
+    Containers
+
+Module defining containers `DenseArray` and `SparseArray` that behaves as
```
mlubin: Nit: This doesn't accurately describe the code because `DenseArray` isn't part of the module yet.

mlubin: Could we maybe call these `DenseAxisArray` and `SparseAxisArray`?

blegat:
I would require some word to make this work on Julia v0.6 and provide support for it for JuMP v0.19 but I don't think it is worth it. Any objection to drop support for Julia v0.6 ?

src/CustomIndexing/CustomIndexing.jl
blegat: In fact we can use `NTuple{N, Any}` but not `NTuple{N}`, see https://discourse.julialang.org/t/calling-a-funcion-with-tuple-types/1745/2
```julia
julia> Tuple{Int, Symbol} <: NTuple{2}
false

julia> Tuple{Int, Symbol} <: NTuple{2, Any}
true
```

mlubin:
No objections to dropping 0.6 support from JuMP. We should continue to
support 0.6 in MOI for a bit longer to avoid forcing solver wrappers to
drop 0.6.

blegat:
Ready to merge on my side, any comment ? :)

src/Containers/Containers.jl
```diff
@@ -0,0 +1,226 @@
...
+
+Base.length(sa::SparseAxisArray) = length(sa.data)
+Base.IteratorSize(::Type{<:SparseAxisArray}) = Base.HasLength()
+# By default `IteratorSize` for `Generator{<:AbstractArray{T,N}}` is
+# `HasShape{N}`
+Base.IteratorSize(::Type{Base.Generator{<:SparseAxisArray}}) = Base.HasLength()
+# Needed in `collect_to_with_first!`
+Base.eachindex(g::Base.Generator{<:SparseAxisArray}) = eachindex(g.iter)
+Base.iterate(sa::SparseAxisArray, args...) = iterate(values(sa.data), args...)
+# Used by JuMP.getloopedcode
```
mlubin: I don't think the comments on eachindex and haskey are needed. These methods will be used in multiple places, and the comment will go out of date.

src/Containers/Containers.jl
```diff
@@ -0,0 +1,226 @@
...
+end
+
+# A `length` argument can be given because `IteratorSize` is `HasLength`
+function Base.similar(sa::SparseAxisArray{S,N,K}, ::Type{T},
+                      length::Integer=0) where {S, T, N, K}
+    d = Dict{K, T}()
+    if !iszero(length)
+        sizehint!(d, length)
+    end
+    return SparseAxisArray(d)
```
mlubin: How does the `N` parameter get inferred? Through `K`?

src/Containers/Containers.jl
```diff
@@ -0,0 +1,226 @@
...
+function Base.mapreduce(f, op, sa::SparseAxisArray)
+    mapreduce(f, op, values(sa.data))
+end
+Base.:(==)(sa1::SparseAxisArray, sa2::SparseAxisArray) = sa1.data == sa2.data
+
+########
+# Show #
+########
+
+# Inspired from Julia SparseAxisArrays stdlib package
```
mlubin: SparseArrays?

docs/src/variables.md
```diff
@@ -221,7 +221,7 @@ datastructures.
```
mlubin: I thought there were more docs to update, but maybe they're in #1597. @odow

odow: Yeah, they're still in the PR.

odow: I just fixed the remaining doctests, so assuming CI passes, #1597 is also good to merge. Depending on which one goes first, the other will have to be rebased on top with the updated docs.

mlubin: #1597 has been in progress for longer so ok with me if it gets merged first (assuming CI passes).

blegat: Ok to merge #1597 merged first

src/Containers/Containers.jl
blegat: Yes. I have added many `@inferred` in the tests to make sure it works

PR #1658, opened 2018-12-03
Fix KeyIterator on 0.7+ in JuMP 0.18

coroa:
`KeyIterator`s in the 0.18 release which are created by calling `keys(b)` on a `JuMPArray` `b` are broken on Julia 0.7+, as illustrated by the tests added in the first commit.

```julia
   @testset "Key iterators of JuMPContainers" begin
       m = Model()
       @variable(m, x[2:3, 1:2])
       @variable(m, y[2:3])
       @variable(m, z[-1:1,[:red,"blue"]])
       @variable(m, w[i=1:2,j=1:2 ; (j+i) % 2 == 1])

       @test all(collect(keys(x)) .== [(2,1), (3,1), (2,2), (3,2)])
       @test all(collect(keys(y)) .== [(2,), (3,)])
       @test all(collect(keys(z)) .== [(-1,:red), (0,:red), (1,:red), (-1,"blue"), (0,"blue"), (1,"blue")])
       @test all(collect(keys(w)) .== [(1,2),(2,1)])
    end
```

The tests on variables `x` to `z` throw a `MethodError`, since the used __next function refers to the deprecated `next` function:
```
Key iterators of JuMPContainers: Error During Test at /home/coroa/.julia/dev/JuMP/test/variable.jl:229
  Test threw exception MethodError(next, ( x[2,1]  x[2,2]
 x[3,1]  x[3,2], 2), 0x0000000000006c5c)
  Expression: all(collect(keys(x)) .== [(2, 1), (3, 1), (2, 2), (3, 2)])
  MethodError: no method matching next(::Array{Variable,2}, ::Int64)
  Closest candidates are:
    next(::I, !Matched::Base.LegacyIterationCompat{I,T,S}) where {I, T, S} at essentials.jl:887
    next(!Matched::Union{Process, ProcessChain}, ::Int64) at deprecated.jl:199
    next(!Matched::AbstractString, ::Integer) at deprecated.jl:53
  Stacktrace:
   [1] __next(::JuMP.JuMPArray{Variable,2,Tuple{UnitRange{Int64},UnitRange{Int64}}}, ::Int64) at /home/coroa/.julia/dev/JuMP/src/JuMPContainer.jl:346
   [2] iterate at /home/coroa/.julia/dev/JuMP/src/JuMPContainer.jl:355 [inlined]
[...]

PR #1659, opened 2018-12-03
Remove broken KeyIterator on 0.7+ in JuMP 0.18

coroa:
As an alternative to #1658, we here remove the `KeyIterator` struct altogether and provide `keys` on a `JuMPArray` directly by `Iterators.product`.

As explained in the discusson of #1658, the only drawback is that `ProductIterator` has a shape.

For instance, `size(product(1:2, 3:5)) = (2, 3)`, s.t. `collect(product(1:2, 3:5))` becomes
```
julia> collect(Iterators.product(1:2, 3:5))
2×3 Array{Tuple{Int64,Int64},2}:
 (1, 3)  (1, 4)  (1, 5)
 (2, 3)  (2, 4)  (2, 5)
```
contrarily to what the current and previous `KeyIterator` returned
```
julia> collect(keys(JuMPArray(rand(2,3), (1:2, 3:5))))
6-element Array{Any,1}:
 (1, 3)
 (2, 3)
 (1, 4)
 (2, 4)
 (1, 5)
 (2, 5)
```

For its main purpose, i.e. the iteration over all index pairs, that would not have any impact, of course.


src/JuMPContainer.jl
```diff
@@ -220,140 +220,3 @@ else
     Base.iterate(it::ValueIterator, k) = iterate(it.x, k)
 end
 Base.length(it::ValueIterator) = length(it.x)
-
-mutable struct KeyIterator{JA<:JuMPArray}
```
mlubin: I wonder why we had such a complicated implementation before.

test/variable.jl
```diff
@@ -219,6 +219,19 @@ end
...
     end
 
+    @testset "Key iterators of JuMPContainers" begin
+        m = Model()
+        @variable(m, x[2:3, 1:2])
+        @variable(m, y[2:3])
+        @variable(m, z[-1:1,[:red,"blue"]])
+        @variable(m, w[i=1:2,j=1:2 ; (j+i) % 2 == 1])
+
+        @test all(reshape(collect(keys(x)), :) .== [(2,1), (3,1), (2,2), (3,2)])
```
mlubin: Is `all` needed, or could we just check for equality of the vectors and sets?

mlubin: Could we replace `reshape(foo, :)` with `vec(foo)`?

coroa: didn't remember about vec, sure

coroa: sure

test/variable.jl
```diff
@@ -219,6 +219,19 @@ end
...
     end
 
+    @testset "Key iterators of JuMPContainers" begin
+        m = Model()
+        @variable(m, x[2:3, 1:2])
+        @variable(m, y[2:3])
+        @variable(m, z[-1:1,[:red,"blue"]])
+        @variable(m, w[i=1:2,j=1:2 ; (j+i) % 2 == 1])
+
+        @test vec(collect(keys(x))) == [(2,1), (3,1), (2,2), (3,2)]
```
mlubin: Nit: The whitespace doesn't follow the JuMP style guide (there should be spaces after commas). However, not a big deal for the release-0.18 branch.

mlubin:
Thanks!

PR #1669, opened 2018-12-10
Move JuMPArray to Containers.DenseAxisArray

blegat:
JuMPArray was exported from JuMP. Now DenseAxisArray and SparseAxisArray are exported from Containers which means that if the user does `using JuMP.Containers` then we has access to `DenseAxisArray` directly. Otherwise he needs to do `JuMP.Containers.DenseAxisArray`.
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1664


docs/src/constraints.md
```diff
@@ -215,7 +215,7 @@ So far, we've added constraints one-by-one. However, just like
 [Variable containers](@ref), JuMP provides a mechanism for building groups of
 constraints compactly. References to these groups of constraints are returned in
 *containers*. Three types of constraint containers are supported: `Array`s,
-`JuMPArray`s, and `Dict`ionaries. We explain each of these in the following.
+`DenseAxisArray`s, and `Dict`ionaries. We explain each of these in the following.
```
mlubin: Outdated mention of `Dict`.

PR #1678, opened 2018-12-13
Add force option to JuMP.fix

### LINKED ISSUE #1634: Can't fix variable when bounds are defined.
If a group of variables has been defined within certain bounds, then it is not possible to fix the value of one of them. In many applications, it is valuable to create a group of variables over a set and then be able to fix only of them.

This code will reproduce the error on Master

```Julia
m = Model() 
@variable(m, 0.0 <= x[1:2] <= 10.0)
JuMP.fix(x[1],0.0) 


AssertionError: !(has_upper_bound(v)) && !(has_lower_bound(v))

Stacktrace:
 [1] fix(::VariableRef, ::Float64) at /Users/jdlara/.julia/packages/JuMP/aU7VH/src/variables.jl:407
 [2] top-level scope at In[50]:3
```

odow:
Closes #1634

### DIFF
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -189,7 +189,7 @@ julia> JuMP.lower_bound(x)
 2.0
 ```
 
-Finally, we can delete variable bounds using `JuMP.delete_lower_bound` and
+We can delete variable bounds using `JuMP.delete_lower_bound` and
 `JuMP.delete_upper_bound`:
 ```jldoctest; setup=:(model=Model())
 julia> @variable(model, 1 <= x <= 2)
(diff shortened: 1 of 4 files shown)


iaravena:
Do we want the bounds to be removed when fixing a variable? If I were to unfix the fixed variable afterwards, I would expect the bounds to remain in the model.

odow:
> Do we want the bounds to be removed when fixing a variable? If I were to unfix the fixed variable afterwards, I would expect the bounds to remain in the model.

See the discussion in https://github.com/JuliaOpt/JuMP.jl/issues/1634

odow:
I've updated both the error message and the docstring to clarify that the existing bounds are deleted. Because `force` is opt-in, users should only use it once they are aware of the consequences. If it becomes important, we can always cache the previous bounds in a future PR.

iaravena:
thank you @odow!

PR #1684, opened 2018-12-15
🚸 Add deprecations for JuMP v0.18 functions

### LINKED ISSUE #1095: deprecation/error message improvements for JuMP/MOI
Let's keep a list of errors and deprecations that need to be added before the JuMP 0.19 release:
- [ ] Removal of `norm` (https://github.com/JuliaOpt/JuMP.jl/pull/1093)
- [ ] Statuses (https://github.com/JuliaOpt/JuMP.jl/issues/1046)

blegat:
This should help with the transition.

Closes #1095

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -61,6 +61,9 @@ moi_set(::RotatedSecondOrderCone, dim::Int) = MOI.RotatedSecondOrderCone(dim)
 
 # Deprecation for JuMP v0.18 -> JuMP v0.19 transition
 function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
-    error("`@constraint(model, norm(x) <= t)` should now be written as " *
+    error("JuMP no longer performs automatic transformation of `norm()` ",
+          "expressions into second-order cone constraints. They should now ",
+          "be expressed using the SecondOrderCone() set. For example, ",
+          "`@constraint(model, norm(x) <= t)` should now be written as ",
           "`@constraint(model, [t; x] in SecondOrderCone())`")
 end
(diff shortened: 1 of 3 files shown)


blegat:
The deprecation for `solve` could be improved by also mentioning that the status has changed and we still need to add a deprecation for `norm` so this PR does not close #1095

blegat:
It now completely addresses #1095 
```julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, x);

julia> using LinearAlgebra

julia> norm([x])
ERROR: `@constraint(model, norm(x) <= t)` should now be written as `@constraint(model, [t; x] in SecondOrderCone())`
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] norm(::Array{VariableRef,1}) at /home/blegat/.julia/dev/JuMP/src/sets.jl:64
 [3] top-level scope at none:0

julia> @constraint(model, norm([x]) <= 1)
ERROR: `@constraint(model, norm(x) <= t)` should now be written as `@constraint(model, [t; x] in SecondOrderCone())`
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] norm(::Array{VariableRef,1}) at /home/blegat/.julia/dev/JuMP/src/sets.jl:64
 [3] top-level scope at /home/blegat/.julia/dev/JuMP/src/macros.jl:325
```

src/sets.jl
```diff
@@ -58,3 +58,9 @@ julia> @constraint(model, [t, x, x-1, x-2] in RotatedSecondOrderCone())
 """
 struct RotatedSecondOrderCone <: AbstractVectorSet end
 moi_set(::RotatedSecondOrderCone, dim::Int) = MOI.RotatedSecondOrderCone(dim)
+
+# Deprecation for JuMP v0.18 -> JuMP v0.19 transition
+function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
+    error("`@constraint(model, norm(x) <= t)` should now be written as " *
```
mlubin: Clarify that this error message is an example of how to reformulate `norm()`, not that it is a direct suggestion for the user's code. It might be confusing if they already have a variable named `x`. E.g.,
"JuMP no longer performs automatic transformation of norm() expressions into second-order cone constraints. They should now be expressed using the SecondOrderCone() set. For example, ..."

blegat:
@mlubin I think we should get this merged before releasing JuMP v0.19-β

PR #1694, opened 2018-12-20
Add variable_with_name

blegat:
Resolve the first part of https://github.com/JuliaOpt/JuMP.jl/issues/1184

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -227,9 +227,9 @@ end
     variable_by_name(model::AbstractModel,
                      name::String)::Union{AbstractVariableRef, Nothing}
 
-Returns the reference of the variable with name `name`. Throws an error if
-several variables have `name` as name. Returns `Nothing` of no variable have
-this name.
+Returns the reference of the variable with name attribute `name` or `Nothing` if
+no variable have this name attribute. Throws an error if several variables have
+`name` as name attribute.
 
 ```jldoctest objective_function; setup = :(using JuMP), filter = r"Stacktrace:.*"s
 julia> model = Model()


docs/src/variables.md
```diff
@@ -246,6 +246,21 @@ julia> JuMP.fix_value(x)
 2.0
 ```
 
+## Variable names
+
+The name of a variable can retrieved using [`JuMP.name`](@ref), set by
```
mlubin: Clarify that this refers to the name attribute.

src/variables.jl
```diff
@@ -223,6 +223,52 @@ function set_name(v::VariableRef, s::String)
     return MOI.set(owner_model(v), MOI.VariableName(), v, s)
 end
 
+"""
+    variable_with_name(model::AbstractModel, name::String)::AbstractVariableRef
```
mlubin: I'd suggest calling this `variable_by_name`.

src/variables.jl
```diff
@@ -223,6 +223,52 @@ function set_name(v::VariableRef, s::String)
...
+
+Returns the reference of the variable with name `name`. Throws an error if
+several variables have the same name (even if it is different from `name`).
```
mlubin: > Throws an error if several variables have the same name

I thought we changed this behavior. It's only an error if multiple variables have the name that you're looking for, right?

src/variables.jl
```diff
@@ -223,6 +223,52 @@ function set_name(v::VariableRef, s::String)
     return MOI.set(owner_model(v), MOI.VariableName(), v, s)
 end
 
+"""
+    variable_with_name(model::AbstractModel, name::String)::AbstractVariableRef
+
+Returns the reference of the variable with name `name`. Throws an error if
```
mlubin: What happens if the name doesn't exist? What error is thrown?

src/variables.jl
blegat: Indeed!

src/variables.jl
```diff
@@ -223,6 +223,73 @@ function set_name(v::VariableRef, s::String)
...
+ u[2]
+
+julia> JuMP.variable_by_name(model, "u[2]")
+u[2]
+```
+"""
+function variable_by_name(model::Model, name::String)
+    index = MOI.get(backend(model), MOI.VariableIndex, name)
+    if index isa Nothing
+        return nothing
```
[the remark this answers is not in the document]
blegat: This is more consistent with the behavior of search function in Base which returns `nothing` when the element is not present, e.g. `findfirst(isequal(3), [1, 2, 4])`.
The user will get an error later on if he uses the reference.

PR #1698, opened 2018-12-20
Fix bug handling Adjoints in macros

### LINKED ISSUE #1667: setobjective issue
Hi I came across this issue while migrating my code from Julia 0.6 to Julia 1.0. I don't know if this is intentional as my code runs fine in Julia 0.6. Here is some code: 
```
using LinearAlgebra;
using JuMP;
using Gurobi;

m = Model(solver=GurobiSolver())
@variable(m, x[i=1:3], lowerbound=0, upperbound=1);
Q = [1 0 0; 0 1 0; 0 0 1;];
c = [3,2,1];
@objective(m, Min, x'*(Q+Matrix(Diagonal([1,2,3])))*x + c'*x - sum(i*x[i] for i=1:3));
status = solve(m);
```
will trigger the following error (Julia 0.7):
> ERROR: LoadError: in setobjective: array of size (1,) passed as objective; only scalar objectives are allowed

However, everything works fine with the following equivalent code, which defines a quadratic expression first before putting into macro "@objective":
```
using LinearAlgebra;
using JuMP;
using Gurobi;

m = Model(solver=GurobiSolver())
@variable(m, x[i=1:3], lowerbound=0, upperbound=1);
Q = [1 0 0; 0 1 0; 0 0 1;];
c = [3,2,1];
qobj = x'*(Q+Matrix(Diagonal([1,2,3])))*x + c'*x - sum(i*x[i] for i=1:3) ;
@objective(m, Min, qobj);
status = solve(m);
```

odow:
Not sure if this is the best way to achieve this, but it seems to work.

Closes #1667.

### DIFF
--- a/src/parse_expr.jl
+++ b/src/parse_expr.jl
@@ -285,14 +285,12 @@ function destructive_add!(ex::AbstractArray{<:GenericAffExpr}, c::Number,
     return result
 end
 
-# For cases such as x' * x, broadcasting the ex .+ will cause the Adjoint to be
-# collected into an Array{T, 2}. The easiest way to work around this is to undo
-# the adjoint, perform the operation, and then re-apply the adjoint.
-function destructive_add!(ex::Number, c::Number, x::Adjoint{T, Vector{T}}) where {T}
-    return (ex .+ c * x')'
+# For some reason, the broadcast syntax `ex .+ c * x` fails if `x` is an
+# `Adjoint`. But if we explicitly call `broadcast` it seems to work.
+# See JuMP PR #1698 for more discussion.
+function destructive_add!(ex, c, x)
+    return Broadcast.materialize(Broadcast.broadcast(+, ex, c * x))
 end
-
-destructive_add!(ex, c, x) = ex .+ c * x
 
 destructive_add_with_reorder!(ex, arg) = destructive_add!(ex, 1.0, arg)
 # Special case because "Val{false}()" is used as the default empty expression.
--- a/src/parse_expr.jl
+++ b/src/parse_expr.jl
@@ -285,6 +285,12 @@ function destructive_add!(ex::AbstractArray{<:GenericAffExpr}, c::Number,
     return result
 end
 
+# For cases such as x' * x, broadcasting the ex .+ will cause the Adjoint to be
+# collected into an Array{T, 2}. The easiest way to work around this is to undo
+# the adjoint, perform the operation, and then re-apply the adjoint.
+function destructive_add!(ex::Number, c::Number, x::Adjoint{T, Vector{T}}) where {T}
+    return (ex .+ c * x')'
+end
 
 destructive_add!(ex, c, x) = ex .+ c * x
 
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -274,15 +274,14 @@ end
     @testset "Adjoints" begin
         model = Model()
         @variable(model, x[1:2])
-        Q = [1.0 0.0; 0.0 1.0]
-        obj = @objective(model, Min, x' * Q * x)
-        @test JuMP.isequal_canonical(obj, sum(x.^2))
-        cref = @constraint(model, x' * Q * x <= 1)
+        obj = @objective(model, Min, x' * ones(2, 2) * x)
+        @test JuMP.isequal_canonical(obj, x[1]^2 + 2 * x[1] * x[2] + x[2]^2)
+        cref = @constraint(model, x' * ones(2, 2) * x <= 1)
         c = JuMP.constraint_object(cref)
-        @test JuMP.isequal_canonical(c.func, sum(x.^2))
+        @test JuMP.isequal_canonical(c.func, x[1]^2 + 2 * x[1] * x[2] + x[2]^2)
         @test c.set == MOI.LessThan(1.0)
         @test JuMP.isequal_canonical(
-            JuMP.destructive_add!(0.0, x', Q), (1.0 .* x)'
+            JuMP.destructive_add!(0.0, x', ones(2, 2)), x' * ones(2, 2)
         )
     end
 end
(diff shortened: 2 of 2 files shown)


src/parse_expr.jl
```diff
@@ -285,6 +285,12 @@ function destructive_add!(ex::AbstractArray{<:GenericAffExpr}, c::Number,
     return result
 end
 
+# For cases such as x' * x, broadcasting the ex .+ will cause the Adjoint to be
+# collected into an Array{T, 2}. The easiest way to work around this is to undo
+# the adjoint, perform the operation, and then re-apply the adjoint.
+function destructive_add!(ex::Number, c::Number, x::Adjoint{T, Vector{T}}) where {T}
```
mlubin: Don't we need to do the same for lots of other cases, like where `ex` is not ` number` or when `c` is `AffExpr`?

blegat:
If you replace `0.0` here:
https://github.com/JuliaOpt/JuMP.jl/blob/b78da6727823b594ef020997e6e48c3e7d15ed90/src/parse_expr.jl#L98
By `Val(false)`, the method called is now `JuMP.destructive_add_with_reorder!(::Val{false}, ::Adjoint)` which calls
https://github.com/JuliaOpt/JuMP.jl/blob/b78da6727823b594ef020997e6e48c3e7d15ed90/src/parse_expr.jl#L295
The broadcast still collects the adjoint into a matrix but overloading this one may be easier.
By the way, it seems like a bug that broadcasting an adjoint lead to collect.

odow:
I don't really know what I'm doing because I haven't played around with the broadcasting stuff, but:
```julia
julia> f(x) = x' .+ 1

julia> g(x) = Broadcast.materialize(Broadcast.broadcast(+, adjoint(x), 1))
g (generic function with 1 method)

julia> x = [1, 2]
2-element Array{Int64,1}:
1
2

julia> f(x)
1×2 Array{Int64,2}:
2  3

julia> g(x)
1×2 LinearAlgebra.Adjoint{Int64,Array{Int64,1}}:
2  3

julia> @code_lowered f(x)
CodeInfo(
1 1 ─ %1 = Base.Broadcast.materialize                                       │
  │   %2 = Base.Broadcast.broadcasted                                       │
  │   %3 = (Base.adjoint)(x)                                                │
  │   %4 = (%2)(Main.:+, %3, 1)                                             │
  │   %5 = (%1)(%4)                                                         │
  └──      return %5                                                        │
)

julia> @code_lowered g(x)
CodeInfo(
1 1 ─ %1 = (Base.getproperty)(Main.Broadcast, :materialize)                 │
  │   %2 = (Base.getproperty)(Main.Broadcast, :broadcast)                   │
  │   %3 = (Main.adjoint)(x)                                                │
  │   %4 = (%2)(Main.:+, %3, 1)                                             │
  │   %5 = (%1)(%4)                                                         │
  └──      return %5                                                        │
)
```

Related issues: https://github.com/JuliaLang/julia/issues/25671, https://github.com/JuliaArrays/AxisArrays.jl/pull/131

src/parse_expr.jl
```diff
@@ -285,8 +285,12 @@ function destructive_add!(ex::AbstractArray{<:GenericAffExpr}, c::Number,
     return result
 end
 
-
-destructive_add!(ex, c, x) = ex .+ c * x
+# For some reason, the broadcast syntax `ex .+ c * x` fails if `x` is an
+# `Adjoint`. But if we explicitly call `broadcast` it seems to work.
+# See JuMP PR #1698 for more discussion.
+function destructive_add!(ex, c, x)
+    return Broadcast.materialize(Broadcast.broadcast(+, ex, c * x))
```
blegat: Isn't `return broadcast(+, ex, c * x)` sufficient ?

blegat:
It seems to be thank to this:
https://github.com/JuliaLang/julia/blob/163b0196b12a84143388fcc0037873ef68f0beec/stdlib/LinearAlgebra/src/adjtrans.jl#L185-L186
which was introduced in:
https://github.com/JuliaLang/julia/pull/25083
A special method is defined for `broadcast` with Adjoint vectors and scalars.
However, it seems that `.+` is lowered to `broadcasted` followed by `materialize` instead of `broadcast` so the method is not called.
I don't understand why it's not lowered to `broadcast` since it is basically equivalent:
https://github.com/JuliaLang/julia/blob/163b0196b12a84143388fcc0037873ef68f0beec/base/broadcast.jl#L707

test/macros.jl
```diff
@@ -270,6 +270,17 @@ end
             "DenseAxisArray, or SparseAxisArray.")
         @test_throws exception @variable(model, x[1:3], container=Oops)
     end
+
+    @testset "Adjoints" begin
+        model = Model()
+        @variable(model, x[1:2])
+        obj = @objective(model, Min, x' * x)
```
blegat: For the bug to be triggered, you need something at least as complex as `x' * ones(2, 2) * x`.

blegat: To confirm, on JuMP master:
```julia
julia> using JuMP; model = Model(); @variable(model, x[1:2]);

julia> @objective(model, Min, x' * x)
x[1]² + x[2]²

julia> typeof(@objective(model, Min, x' * x))
JuMP.GenericQuadExpr{Float64,VariableRef}
```

odow: I've modified the test to include a `Q` matrix and added an explicit call to `destructive_add!` in the tests.

test/macros.jl
```diff
@@ -274,12 +274,16 @@ end
     @testset "Adjoints" begin
         model = Model()
         @variable(model, x[1:2])
-        obj = @objective(model, Min, x' * x)
-        @test obj == sum(x.^2)
-        cref = @constraint(model, x' * x <= 1)
+        Q = [1.0 0.0; 0.0 1.0]
+        obj = @objective(model, Min, x' * Q * x)
```
blegat: `x' Q x` doesn't trigger the bug, we need at least `x' ones(2, 2) x`. If every factor is just a symbol it won't trigger the bug

PR #1703, opened 2018-12-21
Remove weird items

matbesancon:
Odd items in documentation files

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -125,7 +125,7 @@ specified by using the `SecondOrderCone` set.
 
 ### Gurobi
 
-￼Requires a working installation of Gurobi with an activated license (free for
+Requires a working installation of Gurobi with an activated license (free for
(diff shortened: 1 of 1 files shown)


docs/src/installation.md
```diff
@@ -125,7 +125,7 @@ specified by using the `SecondOrderCone` set.
 
 ### Gurobi
 
-￼Requires a working installation of Gurobi with an activated license (free for
+Requires a working installation of Gurobi with an activated license (free for
```
blegat: I fail to see what changed

matbesancon: Check http://www.juliaopt.org/JuMP.jl/dev/installation/#MOSEK-1
There is a weird character

blegat: I don't see it, where is it ?

mlubin: I see it in Firefox, not Chrome

PR #1704, opened 2018-12-22
Printing for the nonlinear part of the model

### LINKED ISSUE #1648: Add nonlinear part in model printing
The nonlinear model stored in `model.nlp_data` is not currently printed.

mlubin:
Closes #1648

@blegat note the commented-out test. `JuMPExtension.MyModel` wasn't actually being tested for printing. Should it be? The nonlinear printing tests will need to be separate in that case.


blegat:
Yes it should. The nonlinear part should be separate indeed.

PR #1707, opened 2018-12-26
Pass keyword argument in build_constraint

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1706

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -213,14 +213,14 @@ function _localvar(x::Expr)
 end
 
 """
-    extract_kwargs(args)
+    extract_kw_args(args)
 
 Process the arguments to a macro, separating out the keyword arguments.
 Return a tuple of (flat_arguments, keyword arguments, and requestedcontainer),
 where `requestedcontainer` is a symbol to be passed to `getloopedcode`.
 """
-function extract_kwargs(args)
-    kwargs = filter(x -> isexpr(x, :(=)) && x.args[1] != :container , collect(args))
+function extract_kw_args(args)
+    kw_args = filter(x -> isexpr(x, :(=)) && x.args[1] != :container , collect(args))
     flat_args = filter(x->!isexpr(x, :(=)), collect(args))
     requestedcontainer = :Auto
     for kw in args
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -231,7 +231,7 @@ function extract_kwargs(args)
     return flat_args, kwargs, requestedcontainer
 end
 
-function addkwargs!(call, kwargs)
+function add_keyword_arguments(call, kwargs)
```
mlubin: (optional) kw seems fine as an abbreviation of keyword

src/macros.jl
```diff
@@ -474,10 +474,12 @@ Returns the code for the macro `@constraint_like args...` of syntax
...
 where `@constraint_like` is either `@constraint` or `@SDconstraint`.
-The expression `con` is parsed by `parsefun` which returns a code that, when
-executed, returns an `AbstractConstraint`. This `AbstractConstraint` is passed
-to `add_constraint` with the macro keyword arguments (except the `container`
-keyword argument which is used to determine the container type).
+The expression `con` is parsed by `parsefun` which returns a `build_constraint`
+call code that, when executed, returns an `AbstractConstraint`. The macro
+keyword arguments (except the `container` keyword argument which is used to
+determine the container type) are added to the `build_constraint` call. The
+returned value of this call is passed to µ `add_constraint` which returns a
```
mlubin: Extra mu?

src/macros.jl
mlubin: This function could use documentation, especially since it modifies the call argument.

src/macros.jl
```diff
@@ -231,7 +231,35 @@ function extract_kwargs(args)
     return flat_args, kwargs, requestedcontainer
 end
 
-function addkwargs!(call, kwargs)
+"""
+    add_kwargs(call, kwargs)
```
mlubin: Would it be `kw_args` according to JuMP style?

src/macros.jl
```diff
@@ -231,7 +231,35 @@ function extract_kwargs(args)
...
+
+Add the keyword arguments `kwargs` to the function call expression `call`,
+escaping the expressions. The keyword arguments can be obtained with
+[`extract_kwargs`](@ref).
+
+## Examples
+
+```jldoctest; setup = :(using JuMP)
```
mlubin: Does this doctest pass? It still refers to `add_keyword_arguments`.

src/macros.jl
```diff
@@ -231,7 +231,35 @@ function extract_kwargs(args)
...
+julia> call = :(f(1, a=2))
+:(f(1, a=2))
+
+julia> using JuMP
+
+julia> macro m(args...)
+           flag_args, kwargs, requested_container
+           JuMP.extract_kwargs(args); JuMP.add_keyword_arguments(call, kwargs)
```
mlubin: It would be sufficient to provide a standalone example without macros. There's a lot of nontrivial things going on in this example (referring to an object in global scope, using `extract_kwargs`. If you provide a more complex example then it needs more explanation.

Issue #1711, opened 2018-12-29
Make JuMP tests callable from an external package


blegat:
The tests that are run for `JuMP.Model` and `JuMPExtension.MyModel` should also work for `StructJuMP.StructuredModel`. Currently, if StructJuMP wanted to run these tests, it would have to include, e.g. `constraint.jl` and then call `constraints_tests(StructuredModel, ...)` but including `constraint.jl` also run the tests for `JuMP.Model` and `JuMPExtension.MyModel`. It would be nice to define the tests and run the tests in separate files so that StructJuMP can define them without running them on `JuMP.Model` and `JuMPExtension.MyModel`.

odow:
I'm actually going to re-open this until we have something documented. Were essentially there, though.

```julia
import JuMP
include(joinpath(dirname(dirname(pathof(JuMP))), "test", "Kokako.jl"))
const MODULES_TO_TEST = Kokako.include_modules_to_test(JuMP)
Kokako.run_tests(
    MODULES_TO_TEST,
    JuMPExtension.MyModel,
    JuMPExtension.MyVariableRef;
    test_prefix = "test_extension_",
)
```

PR #1713, opened 2018-12-30
✅ Fix JuMPExtension print tests

blegat:
The only change in `JuMP.Model` printing done by this PR is that the objective function type displayed is now the JuMP type and not the MOI type but I think it is better that way.

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1710

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -71,6 +71,20 @@ mutable struct NLPData
     evaluator
 end
 
+"""
+    nlp_objective_function(model::Model)
+
+Returns the nonlinear objective function or `nothing` if no nonlinear objective
+function is set.
+"""
+function nlp_objective_function(model::Model)
+    if model.nlp_data isa Nothing
+        return nothing
+    else
+        return model.nlp_data.nlobj
+    end
+end
+
 function create_nlp_block_data(m::Model)
     @assert m.nlp_data !== nothing
     bounds = MOI.NLPBoundsPair[]
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -78,7 +78,7 @@ Returns the nonlinear objective function or `nothing` if no nonlinear objective
 function is set.
 """
 function nlp_objective_function(model::Model)
-    if model.nlp_data isa Nothing
+    if model.nlp_data === nothing
         return nothing
     else
         return model.nlp_data.nlobj
--- a/src/print.jl
+++ b/src/print.jl
@@ -163,13 +163,7 @@ function Base.show(io::IO, model::AbstractModel)
     println(io, "Variable", plural(num_variables(model)), ": ",
             num_variables(model))
     if sense != MOI.FEASIBILITY_SENSE
-        nlobj = nlp_objective_function(model)
-        if nlobj === nothing
-            println(io, "Objective function type: ",
-                    objective_function_type(model))
-        else
-            println(io, "Objective function type: Nonlinear")
-        end
+        show_objective_function_summary(io, model)
     end
     show_constraints_summary(io, model)
     show_backend_summary(io, model)
@@ -216,12 +210,7 @@ function model_string(print_mode, model::AbstractModel)
             str *= "\\quad"
         end
         str *= sep
-        nlobj = nlp_objective_function(model)
-        if nlobj === nothing
-            str *= function_string(print_mode, objective_function(model))
-        else
-            str *= nl_expr_string(model, print_mode, nlobj)
-        end
+        str *= objective_function_string(print_mode, model)
     end
     str *= eol
     str *= ijl ? "\\text{Subject to} \\quad" : "Subject to" * eol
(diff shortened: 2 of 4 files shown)


src/print.jl
```diff
@@ -370,6 +354,38 @@ end
...
+        println(io, "Nonlinear: ", num_nl_constraints(model), " constraint",
+                plural(num_nl_constraints(model)))
+    end
+end
+
+function constraints_string(print_mode, model::Model, sep, eol)
+    str = ""
+    for (F, S) in MOI.get(model, MOI.ListOfConstraints())
+        for idx in MOI.get(model, MOI.ListOfConstraintIndices{F, S}())
+            # FIXME the shape may be incorrect here
```
odow: Does an issue need to be opened for this?

src/print.jl
```diff
@@ -370,6 +354,38 @@ end
...
+            shape = S <: MOI.AbstractScalarSet ? ScalarShape() : VectorShape()
+            cref = ConstraintRef(model, idx, shape)
+            con = constraint_object(cref)
+            str *= sep * constraint_string(print_mode, con) * eol
+        end
+    end
+    if model isa Model && model.nlp_data !== nothing
```
odow: `model isa Model` is always true because of the type annotation in the function arguments?

test/print.jl
```diff
@@ -202,7 +202,8 @@ end
     end
 end
 
-function printing_test(ModelType::Type{<:JuMP.AbstractModel})
+function printing_test(ModelType::Type{<:JuMP.AbstractModel},
+                       moi_backend::Bool)
```
odow: Docstring for when `moi_backend` should be used?

src/print.jl
```diff
@@ -370,6 +354,38 @@ end
 ## Constraints
 #------------------------------------------------------------------------
 
+function show_constraints_summary(io::IO, model::Model)
```
odow: These need docstrings if external packages are expected to extend them.

src/nlp.jl
```diff
@@ -71,6 +71,20 @@ mutable struct NLPData
...
 end
 
+"""
+    nlp_objective_function(model::Model)
+
+Returns the nonlinear objective function or `nothing` if no nonlinear objective
+function is set.
+"""
+function nlp_objective_function(model::Model)
+    if model.nlp_data isa Nothing
```
odow: Not sure if we have a style point on `x isa Nothing` or `x === nothing`?

blegat: I have always wondered what was the recommended way. Is there one way that is better recognized by the compiler ?

mlubin: I've always used `x === nothing`. It seems simpler and I know that it's fast. Not sure about `x isa Nothing`.

src/print.jl
```diff
@@ -162,22 +163,25 @@ function Base.show(io::IO, model::Model)
     println(io, "Variable", plural(num_variables(model)), ": ",
             num_variables(model))
     if sense != MOI.FEASIBILITY_SENSE
-        if model.nlp_data !== nothing && model.nlp_data.nlobj !== nothing
-            println(io, "Objective function type: Nonlinear")
-        else
+        nlobj = nlp_objective_function(model)
```
mlubin: So `AbstractModel`s have to implement `nlp_objective_function`? That's weird. I'm tempted to say that we shouldn't implement `show` for an `AbstractModel`.

src/nlp.jl
odow: Julia 1.1 introduces `isnothing`: https://github.com/JuliaLang/julia/pull/29679

src/print.jl
blegat: Done: https://github.com/JuliaOpt/JuMP.jl/issues/1723

src/print.jl
blegat: I have added `show_objective_function_summary` an `AbstractModel` should implement that instead

test/print.jl
```diff
@@ -202,7 +202,17 @@ end
...
+    printing_test(ModelType::Type{<:JuMP.AbstractModel},
+                  moi_backend::Bool)
+
+Test printing of models of type `ModelType`. The expected model printing depends
+on `moi_backend` which indicates whether the model is stored in an MOI backend
+or is stored in its JuMP form (i.e. as `AbstractVariable`s,
+`AbstractConstraint`s and `AbstractJuMPScalar` for the objective function).
+"""
+function printing_test(ModelType::Type{<:JuMP.AbstractModel},
+                       moi_backend::Bool)
```
mlubin: The `moi_backend == false` case seems pretty specialized to `JuMPExtension.MyModel`. IMO it would make more sense to completely split the printing tests for `JuMP.Model` and `JuMPExtension.MyModel`.

test/print.jl
```diff
@@ -479,8 +473,93 @@ function printing_test(ModelType::Type{<:JuMP.AbstractModel})
...
+    Feasibility problem with:
+    Variable: 1
+    Constraint: 1
+    Names registered in the model: x""", repl=:show)
+    end
+end
+
 @testset "Printing for JuMP.Model" begin
     printing_test(Model)
+    model_printing_test(Model)
```
mlubin: How do we decide which tests go inside `model_printing_test` and which tests are listed like the testsets below? It seems arbitrary given that `model_printing_test` is only called with `Model`.

This structure creates fuzzy abstraction categories ("models of type `ModelType` for which the model is stored in MOI form" and "... where the model is stored in its JuMP form") that have only one concrete example (`Model` and `JuMPExtension.MyModel`, resp.). Why create the abstraction for categories that have only one example?

IMO all the tests `model_printing_test` can be moved under "Printing for JuMP.Model" and all the tests in `model_extension_printing_test` can be moved under "Printing for JuMPExtension.MyModel" unless there's some consideration that I'm missing.

blegat: See https://github.com/JuliaOpt/JuMP.jl/issues/1711

mlubin: Could you add a link to #1711 in the code to explain the design?

PR #1712, opened 2018-12-29
Automate testing of examples

odow:
This PR adds a build stage to Linux/Julia 1.0 on Travis that runs the examples in `/examples`. This requires the installation of some solvers. So far, I have added GLPK (for LP/MIP), Ipopt (for NLP), and SCS (for SDP).

List of examples and whether they have been updated:
- [x] basic
- [x] cannery
- [x] cluster
- [x] corr_sdp
- [x] diet
- [x] knapsack
- [x] maxcut_sdp
- [x] mindistortion
- [x] minellipse
- [x] mle
- [x] multi
- [x] optcontrol
- [x] prod
- [x] qcp
- [x] robust_uncertainty
- [x] rosenbrock
- [x] steelT3
- [x] sudoku
- [x] transp
- [x] urbanplan

Closes #699 
Closes #1175 

- [x] PR currently includes `Manifest.toml` because latest GLPK tag isn't in METADATA. Once it is, we can remove `Manifest.tom`.

PR #1720, opened 2019-01-01
Fix spelling of language

odow:
@IainNZ was responsible for this cracker a very long time ago: https://github.com/JuliaOpt/JuMP.jl/commit/7e49d65b4c80cdb25c0a18ed7741e9dbb4eadbfb

I kinda feel like we should leave it in perpetuity.

There are some other instances in the examples, but they are fixed in #1712.


IainNZ:
Obviously intentional, trying to create a new word

PR #1721, opened 2019-01-02
Add a style point on return statements

### LINKED ISSUE #1718: Missing style point on explicit return
Use `return` explicitly except for assignment-form one-line functions. We decided on this but looks like it never made it to the style guide.

odow:
Closes #1718

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -114,8 +114,10 @@ Bad:
 
 #### Return statements
 
-Use an explicit `return` statement to exit from a function.  If the return from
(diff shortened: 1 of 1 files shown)


docs/src/style.md
```diff
@@ -112,6 +112,37 @@ Bad:
 2(x + 1)
 ```
 
+#### Return statements
+
+Use an explicit `return` statement to exit from a function.  If the return from
```
mlubin: It helps to mention the motivation: without an explicit `return`, it could be confusing whether the author intended to return a certain value or not.

PR #1722, opened 2019-01-02
Mention removal of column generation syntax

### LINKED ISSUE #1719: Mention removal of column generation syntax in NEWS
We had keyword arguments to the `@variable` macro that no longer exist. The plan for column generation has been to call `MOI.add_variable` and then make a `MultirowChange`, but I don't think any solvers have implemented that yet.

odow:
Closes #1719

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -94,6 +94,10 @@ Breaking changes:
   for consistency with JuMP's new
   [style recommendations](http://www.juliaopt.org/JuMP.jl/dev/style/).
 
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -94,6 +94,10 @@ Breaking changes:
   for consistency with JuMP's new
   [style recommendations](http://www.juliaopt.org/JuMP.jl/dev/style/).
 
+- The column generation syntax in `@variable` has been removed (i.e., the
```
mlubin: This should go under the "known regressions" section.

PR #1728, opened 2019-01-04
✨ Add value for constraints

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1394

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -373,6 +373,28 @@ function set_coefficient(constraint::ConstraintRef{Model, MOICON{F, S}},
 end
 
 """
+    value(cref::ConstraintRef)
+
+Get the primal value of this constraint in the result returned by a solver. That
+is, if `cref` is the reference of a constraint `func`-in-`set`, it returns the
+value of `func` evaluated at the value of the variables (given by
+[`value(::VariableRef)`](@ref)).
+Use [`has_values`](@ref) to check if a result exists before asking for values.
+
+## Note
+
+For scalar contraints, the constant is moved to the `set` so it is not taken
+into account in the primal value of the constraint. For instance, the constraint
+`@constraint(model, 2x + 3y + 1 == 5)` is transformed into
+`2x + 3y`-in-`MOI.EqualTo(4)` so the value returned by this function is the
+evaluation of `2x + 3y`.
+```
+"""
+function value(cref::ConstraintRef{Model, <:MOICON})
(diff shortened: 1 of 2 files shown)


src/constraints.jl
```diff
@@ -372,6 +372,28 @@ function set_coefficient(constraint::ConstraintRef{Model, MOICON{F, S}},
...
+
+For scalar contraints, the constant is moved to the `set` so it is not taken
+into account in the primal value of the constraint. For instance, the constraint
+`@constraint(model, 2x + 3y + 1 == 5)` is transformed into
+`2x + 3y`-in-`MOI.EqualTo(4)` so the value returned by this function is the
+evaluation of `2x + 3y`.
+```
+"""
+function value(cref::ConstraintRef{Model, <:MOICON})
+    reshape(MOI.get(cref.model, MOI.ConstraintPrimal(), cref), cref.shape)
```
odow: Nit: missing `return`

blegat:
> What happens if a solver hasn't implemented `ConstraintPrimal`?

That should normally not happen since if it is not natively supported by the solver, the wrapper should use `get_fallback` but the error is
```
ModelLike of type ... does not support accessing the attribute MathOptInterface.ConstraintPrimal(1)
```

mlubin:
The MethodError principle would tell us to check for an `ArgumentError` (https://github.com/JuliaOpt/MathOptInterface.jl/blob/70cc356dc8fcd79b20250ecfed0d1576e7d5b36f/src/attributes.jl#L203) and give a more relevant error message. This could be left as a TODO.

blegat:
What would you change in the error message ? The only thing we can add is that this `ModelLike` is the optimizer used by the JuMP model but the user can probably guess that.
To be consistent, the `try`-`catch` should be added here:
https://github.com/JuliaOpt/JuMP.jl/blob/05bfd2a0c04217c6964fe925a2f86c9668771604/src/JuMP.jl#L517
but having a `try`-`catch` for every attribute access to slightly change the error message seems a bit overkill IMO.
`JuMP.objective_value`, `JuMP.dual`, `JuMP.value`, ... should be understood as thin wrapper on top of the corresponding MOI attributes so it should really be the same as doing the MOI call directly hence I don't think the MethodError applies here.

blegat:
I have moved the discussion to https://github.com/JuliaOpt/JuMP.jl/issues/1739

PR #1736, opened 2019-01-05
Fix julia 1.0 note

joaquimg:


docs/src/index.md
```diff
@@ -80,11 +80,11 @@ JuMP's features include:
         it can be embedded in commercial software that complies with the terms
         of the license.
 
-While neither Julia nor JuMP have reached version 1.0 yet, the releases are
-stable enough for everyday use and are being used in a number of research
-projects and neat applications by a growing community of users who are early
-adopters. JuMP remains under active development, and we welcome your feedback,
-suggestions, and bug reports.
+Although Julia has recenlty reached version 1.0, JuMP has not reached it
```
blegat: It should "Although JuMP is not v1.0, it is stable" but here it reads "Although Julia is v1.0, JuMP is not" and the "it is stable" is not linked by any link-word.
I would just remove "Julia is not v1.0"

Issue #1737, opened 2019-01-05
clear out test/old


mlubin:
Either revive or delete the tests there.

PR #1741, opened 2019-01-06
move over relevant old NLP tests

mlubin:
This revealed an issue. Tests are failing because I don't know how to clear the existence of an MOI objective when setting a nonlinear objective:

https://github.com/JuliaOpt/JuMP.jl/blob/3657606561b74e05e2d062e5fc8e35f33d3a16cb/src/nlp.jl#L17

Ref #1737 

Note: The failure won't appear on this PR because test/nlp_solver.jl isn't run by travis.


blegat:
Clearing the objective should be done by `MOI.set(model, MOI.ObjectiveSense(), MOI.FEASIBILITY_SENSE)` I'd say.
If you want to change it to NL and keep the objective sense, you should do `MOI.set(model, MOI.ObjectiveFunction(), nonlinear_function)`. As this is not yet the API for nonlinear function, I would say that the temporary rule for an optimizer is that if a nonlinear objective is given in the nlp data, it should ignore the MOI.ObjectiveFunction. Would that work ?

mlubin:
Good to go now.

PR #1743, opened 2019-01-06
Document build_constraint in extensions.md

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1740

### DIFF
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -34,7 +34,7 @@ The [`@constraint`](@ref) macro always calls the same three functions:
 * [`build_constraint`](@ref): given the functions and sets involved in the
   constraints, it returns a `AbstractConstraint`;
 * [`add_constraint`](@ref): given the model, the `AbstractConstraint`
-  constructedi in [`build_constraint`](@ref) and the constraint name, it stores
+  constructed in [`build_constraint`](@ref) and the constraint name, it stores
   them in the model and returns a `ConstraintRef`.
 
 Adding methods to these functions is the recommended way to extend the
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -14,7 +14,7 @@ TODO: Create new MOI function/sets, how to use it in JuMP, create new bridge
 
 ## Extending JuMP macros
 
-In order to provide the most convenient syntax for the user to create variables,
+In order to provide a convenient syntax for the user to create variables,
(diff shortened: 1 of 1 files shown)


docs/src/extensions.md
```diff
@@ -1,5 +1,93 @@
...
 TODO: How to extend JuMP: discussion on different ways to build on top of JuMP.
 How to extend JuMP's macros and how to avoid doing this.
+
+## Extending MOI
+
+TODO: Create new MOI function/sets, how to use it in JuMP, create new bridge
+
+## Extending JuMP macros
+
+In order to provide the most convenient syntax for the user to create variables,
```
mlubin: "Most convenient" implies that we think that the macros are the best possible syntax. Maybe "a convenient syntax".

docs/src/extensions.md
```diff
@@ -1,5 +1,93 @@
...
+`AbstractConstraint` with a corresponding [`add_constraint`](@ref) method (see
+[Adding `add_constraint` methods](@ref)).
+
+```@docs
+build_constraint
+```
+
+#### Shapes
+
+To convert the constraint to a `VectorConstraint`, it needs to be converted to
```
mlubin: To convert which constraint? Maybe say instead, "shapes allow vector constraints, which are represented as flat vectors in MOI, to retain a matrix shape at the JuMP level".

docs/src/expressions_draft.md
```diff
@@ -0,0 +1,12 @@
+### Defining a new MOI function
```
mlubin: Is this file referenced anywhere?

blegat: Oups, it shouldn't have been added

docs/src/extensions.md
```diff
@@ -1,5 +1,91 @@
...
+
+## Extending the `@constraint` macro
+
+The [`@constraint`](@ref) macro always calls the same three functions:
+* `parse_constraint`: is called at parsing time, it parses the constraint
+  expression and returns a [`build_constraint`](@ref) call expression;
+* [`build_constraint`](@ref): given the functions and sets involved in the
+  constraints, it returns a `AbstractConstraint`;
+* [`add_constraint`](@ref): given the model, the `AbstractConstraint`
+  constructedi in [`build_constraint`](@ref) and the constraint name, it stores
```
mlubin: typo: "constructedi"

PR #1745, opened 2019-01-07
Pukeko-ize tests

IainNZ:
Requires master of Pukeko, one commit ahead of 0.2.0, will tag if I find any more glitches.

I did find two strange things:
* expr.jl is, as written, run basically twice but it seems like most of the tests don't use the extensions? I reduced it down to running only things once if possible.
* A `(true, true)` in `test/model.jl` that is probably a mistake?

Going to do more.

Running tests in parallel with

```
#! /bin/bash

PUKEKO_TESTS=$'expr.jl\nmodel.jl\nvariable.jl\n'
echo "$PUKEKO_TESTS" | xargs -P 8 -I % julia --project=. test/%
```

### DIFF
--- a/test/expr.jl
+++ b/test/expr.jl
@@ -1,3 +1,183 @@
+#  Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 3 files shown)


mlubin:
Cool! I'll have to check this out and play around with it before approving. Did you measure the speedup in the runtime of the tests?

IainNZ:
Might still be worth to Pukeko-ize from Base.Test but @odow has started doing a good first step which is just function-izing things which is part of the speedup.

PR #1750, opened 2019-01-07
Enable test/expr for extension models (excluding copy)

### LINKED ISSUE #1747: test/expr.jl isn't really using the extension tests?
Only a small fraction seem to make use of the extension - is this intentional?

odow:
Copy test fails with extension because `copy(::GenericAffExpr{Float64, MyVariableRef}, ::MyModel)` isn't defined.

@blegat do we need to implement/test this?

Closes #1747


blegat:
Looking at the `copy` code, it should work for `AbstractModel`, we can just modify its signature

PR #1757, opened 2019-01-09
implement all_variables

### LINKED ISSUE #1612: Feature request: get all variables in the model
```julia
function all_variables(model::Model)
    return [Variable(model, idx) for idx in MOI.get(model, MOI.ListOfVariableIndices())]
end
```

mlubin:
Closes #1612

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -694,3 +694,28 @@ function add_variable(m::Model, v::ScalarVariable, name::String="")
     end
     return vref
 end
+
+"""
+    all_variables(model::Model)::Vector{VariableRef}
+
+Returns a list of all variables currently in the model. The variables are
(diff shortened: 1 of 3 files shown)


src/variables.jl
```diff
@@ -694,3 +694,28 @@ function add_variable(m::Model, v::ScalarVariable, name::String="")
...
+
+# output
+
+2-element Array{VariableRef,1}:
+ x
+ y
+```
+"""
+function all_variables(model::Model)
+    return [VariableRef(model, idx) for idx in
```
blegat: It might be more efficient to annotate the return type of the comprehension, se`
https://github.com/JuliaOpt/JuMP.jl/pull/1639
That is, doing `VariableRef[...]`

PR #1759, opened 2019-01-09
Discuss PackageName.Optimizer in NEWS

odow:


NEWS.md
```diff
@@ -25,6 +25,9 @@ Breaking changes:
   updated for MOI. See the
   [installation guide](http://www.juliaopt.org/JuMP.jl/dev/installation/)
   for a list of solvers that have and have not yet been updated.
+  
+- Most solvers have been renamed to `PackageName.Optimizer`. For example, 
+  `GurobiSolver(OutputFlag=0)` is now `Gurobi.Optimizer(OutputFlag=0)`.
```
blegat: The `Gurobi.Optimizer` isn't really the replacement for `GurobiSolve()r`. The solver is the factory so the replacement for `GurobiSolver` is `with_optimizer(Gurobi.Optimizer)`.
Maybe that is just a technical detail that shouldn't be mentioned here

blegat: In view of https://github.com/JuliaOpt/JuMP.jl/pull/1758, I think it's ok, we don't need to dig into details here

mlubin:
Needs to be updated to resolve merge conflicts.

PR #1765, opened 2019-01-10
🚸 Throw error if a variable is not owned

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1762

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -261,8 +261,15 @@ Replaces `getvalue` for most use cases.
 """
 value(a::GenericAffExpr) = value(a, value)
 
+function verify_ownership(model::AbstractModel, a::GenericAffExpr)
+    for variable in keys(a.terms)
+        verify_ownership(model, variable)
+    end
+end
+
 # Note: No validation is performed that the variables in the AffExpr belong to
-# the same model.
+# the same model. The verification is done in `verify_ownership` which should be
+# called before calling `moi_function`.
 function MOI.ScalarAffineFunction(a::AffExpr)
     assert_isfinite(a)
     terms = MOI.ScalarAffineTerm{Float64}[MOI.ScalarAffineTerm(t[1],
@@ -274,6 +281,7 @@ moi_function(a::GenericAffExpr) = MOI.ScalarAffineFunction(a)
 function moi_function_type(::Type{<:GenericAffExpr{T}}) where T
     return MOI.ScalarAffineFunction{T}
 end
+
 
 function AffExpr(m::Model, f::MOI.ScalarAffineFunction)
     aff = AffExpr()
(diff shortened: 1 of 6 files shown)


test/model.jl
```diff
@@ -42,6 +42,74 @@ end
...
+                @test_throws err @constraint(model_x, y*y + x + y <= 1)
+                @test_throws err begin
+                    @constraint(model_x, [x, y*y + x + y] in MOI.Zeros(2))
+                end
+                @test_throws err begin
+                    @constraint(model_x, [y*y + x + y, x] in MOI.Zeros(2))
+                end
+            end
+            @testset "objective" begin
+                @test_throws err @objective(model_x, Min, x*y)
```
odow: Nit: spaces around binary operators (here and above).

src/constraints.jl
```diff
@@ -106,6 +106,36 @@ struct ConstraintRef{M <: AbstractModel, C, Shape <: AbstractShape}
...
+Returns the model to which `cref` belongs.
+"""
+owner_model(cref::ConstraintRef) = cref.model
+
+"""
+    verify_ownership(model::AbstractModel, cref::ConstraintRef)
+
+Throw `ConstraintNotOwned` if `owner_model(cref)` is not `model`.
+"""
+function verify_ownership(model::AbstractModel, cref::ConstraintRef)
```
mlubin: `verify_ownership_or_throw` would make it clearer what the function does. Or maybe `check_belongs_to_model(::Ref, ::Model)`.

src/aff_expr.jl
```diff
@@ -261,8 +261,15 @@ Replaces `getvalue` for most use cases.
...
+function verify_ownership(model::AbstractModel, a::GenericAffExpr)
+    for variable in keys(a.terms)
+        verify_ownership(model, variable)
+    end
+end
+
 # Note: No validation is performed that the variables in the AffExpr belong to
-# the same model.
+# the same model. The verification is done in `verify_ownership` which should be
+# called before calling `moi_function`.
```
mlubin: The comment is confusing because this function is `MOI.ScalarAffineFunction`, not `moi_function`.

src/variables.jl
```diff
@@ -119,6 +119,37 @@ true
...
+
+The variable `variable` was used in a model different to
+`owner_model(variable)`.
+"""
+struct VariableNotOwned{V <: AbstractVariableRef} <: Exception
+    variable::V
+end
+
+"""
+    check_belongs_to_model(model::AbstractModel, func::AbstractJuMPScalar)
```
mlubin: The order of the parameters should be switched to match the order of words in the function name, i.e., check that the variable belongs to the model.

mlubin:
Please fix travis failure before merging.

PR #1771, opened 2019-01-13
edits for the quickstart section

mlubin:


docs/src/quickstart.md
```diff
@@ -49,10 +54,8 @@ end
 !!! note
     Your model doesn't have to be called `model` - it's just a name.
 
-There are a few options for defining a variable, depending on whether you want
-to have lower bounds, upper bounds, both bounds, or even no bounds. The
-following commands will create two variables, `x` and `y`, with both lower and
-upper bounds. Note the first argument is our model variable ``model``. These
+The following commands will create two variables, `x` and `y`, with both lower
+and upper bounds. Note the first argument is our model variable ``model``. These
```
blegat: Why the double backsticks for model ?

mlubin: Dunno. That was already there. Fixed.

docs/src/quickstart.md
```diff
@@ -35,9 +35,14 @@ CachingOptimizer state: NO_OPTIMIZER
...
 
+!!! note
+    The term "solver" is used as a synonym for "optimizer". The convention in
+    code, however, is to always use "optimizer", e.g., `with_optimizer` and
+    `GLPK.Optimizer`.
+
 ```@meta
 DocTestSetup = quote
-    # Using a caching optimizer removes the need to # load a solver such as GLPK
+    # Using a caching optimizer removes the need to load a solver such as GLPK
```
odow: Using a *mock* optimizer (this is probably my bad)

docs/src/quickstart.md
```diff
@@ -49,10 +54,8 @@ end
 !!! note
     Your model doesn't have to be called `model` - it's just a name.
 
-There are a few options for defining a variable, depending on whether you want
-to have lower bounds, upper bounds, both bounds, or even no bounds. The
-following commands will create two variables, `x` and `y`, with both lower and
-upper bounds. Note the first argument is our model variable ``model``. These
+The following commands will create two variables, `x` and `y`, with both lower
+and upper bounds. Note the first argument is our model variable `model`. These
```
odow: "is our model `model`". Calling the model a "variable" is confusing in the next sentence

docs/src/quickstart.md
```diff
@@ -49,10 +54,8 @@ end
...
 variables are associated with this model and cannot be used in another model.
```
odow: These variables (`x` and `y`) are associated

docs/src/quickstart.md
```diff
@@ -112,8 +119,8 @@ DocTestSetup = quote
 end
 ```
 
-After the call to `JuMP.optimize!` has finished, we need to understand why the
-optimizer stopped. This can be for a number of reasons. First, the solver might
+After the call to `JuMP.optimize!` has finished, we need to query what happened.
+The solve could terminate for a number of reasons. First, the solver might
```
odow: Does the solve terminate, or the optimizer terminate the solve? "The optimizer can terminate the solve for a number of reasons."?

mlubin: "to terminate" (intransitive) means to come to an end (https://www.merriam-webster.com/dictionary/terminate), i.e., "the solve could come to an end for a number of reasons." Seems fine to me.

PR #1772, opened 2019-01-13
🎨 quadterm -> quad_terms

blegat:
In accordance to the [JuMP style guide](http://www.juliaopt.org/JuMP.jl/dev/style/).

### DIFF
--- a/src/quad_expr.jl
+++ b/src/quad_expr.jl
@@ -92,14 +92,14 @@ linear_terms(quad::GenericQuadExpr) = LinearTermIterator(quad.aff)
 struct QuadTermIterator{GQE<:GenericQuadExpr}
     quad::GQE
 end
-# TODO: rename to quad_terms
+
 """
-    quadterms(quad::GenericQuadExpr{C, V})
+    quad_terms(quad::GenericQuadExpr{C, V})
 
 Provides an iterator over tuples `(coefficient::C, var_1::V, var_2::V)` in the
 quadratic part of the quadratic expression.
 """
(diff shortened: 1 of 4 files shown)


blegat:
`quadratic_terms` or `quad_term` ? Now is the time for bikeshedding

mlubin:
We have QuadExpr so quad_terms seems fine.

blegat:
> We have QuadExpr so quad_terms seems fine

Makes sense

src/quad_expr.jl
```diff
@@ -269,28 +269,28 @@ last term added.
...
+    return offset + length(quad_terms(aff))
 end
 
 function MOI.VectorQuadraticFunction(quads::Vector{QuadExpr})
-    num_quad_terms = sum(quad -> length(quadterms(quad)), quads)
-    quad_terms = Vector{MOI.VectorQuadraticTerm{Float64}}(undef, num_quad_terms)
-    num_aff_terms = sum(quad -> length(linear_terms(quad)), quads)
-    lin_terms = Vector{MOI.VectorAffineTerm{Float64}}(undef, num_aff_terms)
+    num_qua_terms = sum(quad -> length(quad_terms(quad)), quads)
+    qua_terms = Vector{MOI.VectorQuadraticTerm{Float64}}(undef, num_qua_terms)
```
mlubin: Not sure I'm a fan of `qua_terms`. The fact that `quad_terms` conflicts doesn't make it any better to use `qua`. What about `quadratic_terms`?

PR #1777, opened 2019-01-14
Update URL to the stable version of Julia manual

petershintech:
The URL to the stable version of Julia manual website, https://docs.julialang.org/en/stable/manual/, is not generated any more. Instead, use the new URL, https://docs.julialang.org/en/v1/manual/.

Discussed at JuliaLang/julia#30524


mlubin:
Thanks!

PR #1778, opened 2019-01-15
style point on using vs import

mlubin:


IainNZ:
This was in the Google Doc draft style guide right? Was there anything else in there worth bringing over that hasn't been?

odow:
> This was in the Google Doc draft style guide right? Was there anything else in there worth bringing over that hasn't been?

I've opened an issue with the text from the doc: https://github.com/JuliaOpt/JuMP.jl/issues/1780

blegat:
In view of https://discourse.julialang.org/t/exporting-symbols-of-imported-module/19651, we should probably make an exception for JuMP extensions to allow them to do `@reexport using JuMP`.
But this is only inside the code of the extension. Non throw-away script should do
```julia
import SumOfSquares
import JuMP
model = JuMP.Model()
JuMP.@constraint(model, poly in SumOfSquares.SOSCone())
```
and throw-away scripts can do
```julia
using SumOfSquares
model = Model()
@constraint(model, poly in SOSCone())
```

mlubin:
How does someone writing the non-throw-away script know that they should write
```julia
import SumOfSquares
import JuMP
model = JuMP.Model()
JuMP.@constraint(model, poly in SumOfSquares.SOSCone())
```
instead of
```julia
import SumOfSquares
model = SumOfSquares.Model()
SumOfSquares.@constraint(model, poly in SumOfSquares.SOSCone())
```
?

blegat:
> instead of
> ```julia
> import SumOfSquares
> model = SumOfSquares.Model()
> SumOfSquares.@constraint(model, poly in SumOfSquares.SOSCone())
> ```

This would also work as `using JuMP` is done in `SumOfSquares` (even if it is not done in the submodules).

mlubin:
> This would also work as using JuMP is done in SumOfSquares (even if it is not done in the submodules).

Yes, but I'm questioning if we want that to work. Seems to cause confusion about where the symbols are coming from. What if you use two JuMP extensions at the same time? You could have `ExtensionA.Model` and `ExtensionB.Model` both being the same as `JuMP.Model`.

blegat:
Why is `ExtensionA.Model` and `ExtensionB.Model` being the same thing an issue ? If one does `using ExtensionA, ExtensionB` then it needs to link to the same thing otherwise there is a clash.

mlubin:
> Why is ExtensionA.Model and ExtensionB.Model being the same thing an issue ?

It's a code readability issue, although not that serious.

Anyway, making an exception in the style guide for `@reexport using JuMP` feels too much like a special case. I'd like a more general answer to how to extend modules. It's maybe too early to understand all the trade-offs involved before we propagate the change in this PR first. How about an agreement that JuMP extensions are welcome to use `@reexport using JuMP` but should have a comment on that line explaining the deviation from the style guide?

blegat:
> How about an agreement that JuMP extensions are welcome to use `@reexport using JuMP` but should have a comment on that line explaining the deviation from the style guide?

That would work

mlubin:
Last call for objections.

PR #1779, opened 2019-01-15
Export all non-internal functions.

### LINKED ISSUE #1761: Decide what to export
JuMP currently exports a pretty arbitrary set of symbols:

https://github.com/JuliaOpt/JuMP.jl/blob/040841ef2ca2f0a4054bde4da4eff1be90fe0948/src/JuMP.jl#L26-L36

We should try to come up with a coherent convention before the 0.19 release. 

Options:
1. Export nothing
2. Export everything except internal functions
3. Export everything except internal functions and names that conflict with Base/stdlib
4. Export everything except internal functions and names that might conflict with Base/stdlib or other packages.

We could define internal functions as anything whose name begins with `_`. (So far we haven't consistently been following this convention.)

4 Is very vague. 3 is still vague and not future-proof because it's hard to keep track of what names exist in Base/stdlib. 2 is principled but makes it hard to import JuMP and anything else via `using`. 1 is principled but less user friendly. No clear winner.

mlubin:
Surprisingly not very painful. This feels like a step in the right direction. We'll likely be prompted to rename some things with underscores.

Note: `MOI` is exported from JuMP after this change. This needs more consideration.

Closes #1761


src/JuMP.jl
```diff
@@ -735,7 +717,26 @@ include("macros.jl")
 include("optimizer_interface.jl")
 include("nlp.jl")
 include("print.jl")
+##########################################################################
```
odow: Nit: It really bothers me that there are 6 `#`s missing from this line.

mlubin: I removed all of these.

mlubin:
After going through and updating the docs, I'm more convinced that this is a good idea. We no longer have `JuMP.` littered everywhere except some arbitrary cases. Users just need to figure out once that these symbols are exported from the module.

odow:
> I'm more convinced that this is a good idea

Seconded. I've changed my earlier mind. 

p.s. I also fixed the failing example so hopefully the tests pass this time.

Issue #1780, opened 2019-01-15
Missing style guide points


odow:
The following points were discussed at a committee meeting 1 July 2018 in Bordeaux.

### Best practice when developing
If you touch old code, fix style.

### Source code
80 character line limit. 
~~4 spaces for indentation.~~
~~No one-letter variable names. This includes (m = Model()). Exception for indices in loops (like for i in 1:N), maybe arguments to one-line anonymous functions?~~
~~Which abbrev. should we allow in variable names? (con vs. constr, aff vs. affine)~~
~~lowercase_with_underscore for variables and functions~~
~~CamelCase for types and module names~~
~~UPPERCASE for constants~~
~~Never use “0.”~~
Start internal functions with underscore, e.g., _add_constraint().
Use three backticks for name of internal function docstrings because underscores break markdown parsing. The 4-space indent doesn’t display with code highlighting in Documenter html docs. Maybe this is a documenter bug that can be fixed instead.
Variables can’t start with underscore
~~No bangs (!) in function names. Use docstrings to indicate what’s modified and variable lifetime issues.~~
Never more than one consecutive blank line inside a function
~~Always use return from a function unless it is a one-line function~~
Don’t split a one-line function over lines. Only if less than 80 chars
~~Type arguments to user-facing functions so that users don’t get internal errors when passing wrong types~~
~~Prefer “for foo in 1:n” over “for foo = 1:n”~~
(suggestion/proposal by @IainNZ, please comment/disagree): prefer “functional” style over “C” style where performance isn’t critical? E.g., use map + anonymous?
~~(suggestion/proposal by @IainNZ, please comment/disagree): Comments should end in `.`, i.e., use punctuation. (This is the most MINOR THING EVER but it really bugs me when it is missing now.)~~
Imports: use import instead of using, importing specific things from a module is fine, stdlib imports first, alphabetical order after that. 
~~Vector instantiation : T[] over Vector{T}()~~
Function appearance order in the files : ???
Prefer short functions : less than 40 lines ? 

### Inside macros
Never use 1.0 when 1 is okay
2x + 3y not x*2 + y*3
Parentheses in macros
`@variables` over lots of `@variable`

### Models
~~Don’t use m for model~~
If the model fits on one screen, short letter variables are okay. If it is decomposed over multiple functions, long variable names.
Splicing into nonlinear stuff???

### Doc strings
` ```julia` not ` ``` ` or 4 spaces indent for pretty printing
[...]

PR #1781, opened 2019-01-15
Updates to the style guide

odow:
Ref #1780.

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -144,8 +144,16 @@ function foo(x)
 end
 ```
 
+#### Line length
 
-#### Line breaks
+Following the Google style guide for [line length](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length)
+in python, keep lines under an 80 character limit.
+
+We make exceptions for
+ - URLs
+ - pathnames
+ - long string constants not containing whitespace that would be inconvenient to
+   split across multiple lines.
 
 For readability, don't split a one-line function over multiple lines.
 
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -260,6 +260,12 @@ some_local_variable = ...
 some_file.jl # Except for ModuleName.jl.
 ```
 
+Never begin a variable name with an underscore. We make an exception for
+non-exported constants.
+```julia
+const _PRIVATE_JUMP_CONSTANT = 3.14159
+```
+
 #### Use of underscores
 
 The Julia style guide recommends avoiding underscores "when readable", for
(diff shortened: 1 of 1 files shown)


docs/src/style.md
```diff
@@ -167,6 +167,76 @@ empty vector with element type `T`. Prefer `T[]` because it is more concise.
...
+```julia
+@variable(model, x >= 0)
+```
+
+Bad:
+```julia
+@variable(model, x >= 0.0)
+```
+
+Always use parentheses.
```
blegat: This is only in the context of macros right ? Otherwise it is ambiguous

docs/src/style.md
```diff
@@ -167,6 +167,76 @@ empty vector with element type `T`. Prefer `T[]` because it is more concise.
 Both `1.0` and `1.` create a `Float64` with value `1.0`. Prefer `1.0` over
 `1.` because it is more easily distinguished from the integer constant `1`.
 
+#### Comments
+
+Comments in code must be proper English sentences with appropriate punctuation.
```
mlubin: Provide a motivation, e.g., for nonnative speakers and for general clarity.

docs/src/style.md
```diff
@@ -167,6 +167,76 @@ empty vector with element type `T`. Prefer `T[]` because it is more concise.
...
+```julia
+# This is a comment demonstrating a good comment.
+```
+
+Bad:
+```julia
+# a bad comment
+```
+
+#### Macro syntax
```
mlubin: Is this supposed to apply to all macros or just JuMP's DSL?

docs/src/style.md
```diff
@@ -167,6 +167,76 @@ empty vector with element type `T`. Prefer `T[]` because it is more concise.
...
+
+Never use `1.0` when `1` is okay.
```
mlubin: Why would the use of 1.0 or 1 depend on if it's inside a macro?

mlubin: Motivation?

docs/src/style.md
```diff
@@ -179,6 +249,18 @@ some_local_variable = ...
 some_file.jl # Except for ModuleName.jl.
 ```
 
+Begin private functions and constants with an underscore.
```
mlubin: Explain that all other functions are exported (after #1779)

docs/src/style.md
```diff
@@ -167,6 +167,76 @@ empty vector with element type `T`. Prefer `T[]` because it is more concise.
...
+```julia
+@constraint(model, 3 * x <= 1)
+```
+
+Bad:
+```julia
+@constraint(model, x * 3 <= 1)
+```
+
+Prefer plural form of macros over lots of repeated calls to singular forms.
```
mlubin: Motivation? Easier to read? What about calls with lots of keyword arguments?

PR #1782, opened 2019-01-16
Add style guide section on JuMP macros

odow:
Ref issue: https://github.com/JuliaOpt/JuMP.jl/issues/1780
Ref previous PR: https://github.com/JuliaOpt/JuMP.jl/pull/1781

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -181,6 +181,18 @@ Bad:
 @variable model x >= 0
 ```
 
+Use `constant * variable` as opposed to `variable * constant`.
+
+Good:
+```julia
+@constraint(model, 3 * x <= 1)
+```
+
+Bad:
+```julia
+@constraint(model, x * 3 <= 1)
+```
+
 ### Naming
 
 ```julia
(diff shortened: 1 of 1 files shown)


blegat:
Use doctests would probably be even better to make sure that we don't break these syntaxes in the future (e.g. I didn't know the keyword argument syntax for the multiline `@variable`).

docs/src/style.md
```diff
@@ -167,6 +167,70 @@ empty vector with element type `T`. Prefer `T[]` because it is more concise.
...
+```julia
+@variable(model, x >= 0)
+```
+
+Bad:
+```julia
+@variable model x >= 0
+```
+
+Use `constant * variable` as opposed to `variable * constant`.
```
mlubin: Motivation: doing this consistently makes it easier to read models.

docs/src/style.md
```diff
@@ -167,6 +167,70 @@ empty vector with element type `T`. Prefer `T[]` because it is more concise.
 Both `1.0` and `1.` create a `Float64` with value `1.0`. Prefer `1.0` over
 `1.` because it is more easily distinguished from the integer constant `1`.
 
+#### JuMP macro syntax
+
+Always use parentheses.
```
mlubin: Motivation: consistency

PR #1785, opened 2019-01-16
[style] Maximum line length

odow:
This one is going to be contentious (looking at @chriscoey).

Ref issue: https://github.com/JuliaOpt/JuMP.jl/issues/1780
Ref previous PR: https://github.com/JuliaOpt/JuMP.jl/pull/1781

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -144,8 +144,16 @@ function foo(x)
 end
 ```
 
+#### Line length
 
-#### TODO: Line breaks
+Following the Google style guide for [line length](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length)
+in python, keep lines under an 80 character limit.
+
+We make exceptions for
+ - URLs
+ - pathnames
+ - long string constants not containing whitespace that would be inconvenient to
+   split across multiple lines.
 
 ### Syntax
 
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -156,17 +156,11 @@ Therefore, aim to keep line lengths under 80 characters by breaking lines
 for maximum readability (examples are given in the [Line breaks](@ref) section),
 but don't treat this as a hard rule.
 
-We make exceptions for
-
- - URLs
- - path names
- - long string constants not containing whitespace
- - lines for which breaking over multiple lines worsens readability
-
 #### Line breaks
 
 The "readability" of a line is subjective. In this section we give suggestions
-of good and bad style of how to break a line.
+of good and bad style of how to break a line. These suggestions are inspired by
+Google's [Python style guide](https://google.github.io/styleguide/pyguide.html).
 
 When defining functions, align arguments vertically after the opening
 parenthesis, or list all arguments on a new (indented) line.
@@ -208,24 +202,6 @@ Better:
 ```julia
 a_very_long_variable_name = a_long_variable_name_with_arguments(
     first_argument, second_argument)
-```
-
-Only use vertical alignment for function arguments. Don't use it in any other
-situation.
-
-Bad:
-```julia
-# Vertical alignment used in a non-function context. Also, indent is not a
-# multiple of 4 spaces.
-comprehension = [(first, second) for (first, second) in list_of_things
-                 if isodd(first)]
-```
-
-Better
-```julia
-comprehension = [
-    (first, second) for (first, second) in list_of_things if isodd(first)
-]
 ```
 
 Don't use vertical alignment if it would be more readable to place all arguments
@@ -270,21 +246,6 @@ f(x) = 1 + x +
 Better:
 ```julia
 f(x) = 1 + x + x^2 + x^3 + x^3
-```
-
-Drop trailing parentheses onto a new line if it improves readability.
-
-Good:
-```julia
-con_index = MOI.add_constraint(
-    backend(owner_model(variable)), MOI.SingleVariable(index(variable)), set)
-```
-
-Also good:
-```julia
-con_index = MOI.add_constraint(
-    backend(owner_model(variable)), MOI.SingleVariable(index(variable)), set
-)
 ```
 
 ### Syntax
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -162,6 +162,10 @@ The "readability" of a line is subjective. In this section we give suggestions
 of good and bad style of how to break a line. These suggestions are inspired by
 Google's [Python style guide](https://google.github.io/styleguide/pyguide.html).
 
+!!! note
+    If you're unsure about how format your code, you can experiment (in Python)
+    using [YAPF](https://yapf.now.sh/).
+
 When defining functions, align arguments vertically after the opening
 parenthesis, or list all arguments on a new (indented) line.
 
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -146,14 +146,127 @@ end
 
 #### Line length
 
-Following the Google style guide for [line length](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length)
-in python, keep lines under an 80 character limit.
+Line lengths are a contentious issue. Our foremost goal is to maximize code
+readability. Very long line lengths can be hard to easily comprehend. However,
+arbitrarily enforcing a maximum line length (like 80 characters) inevitably
+leads to cases in which slightly longer lines (e.g. 81 characters) might be more
+readable.
 
-We make exceptions for
- - URLs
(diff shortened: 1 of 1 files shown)


chriscoey:
IMO harder to read and archaic given modern code viewers and editors (wrapping), but I won't put up a fight.

docs/src/style.md
```diff
@@ -144,8 +144,16 @@ function foo(x)
 end
 ```
 
+#### Line length
 
-#### TODO: Line breaks
+Following the Google style guide for [line length](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length)
+in python, keep lines under an 80 character limit.
```
mlubin: This is a contentious point so we need to give our own motivation and not just defer to the google style guide.
My thinking is that for consistency, we need *some* guideline on line length. 80 characters is a reasonable pre-existing convention.

odow: Maybe we just make this a recommendation. What about:

Our foremost goal is to maximize code readability. Very long line lengths are hard to read since they either require horizontal scrolling, or are wrapped uncontrollably by a text editor. Thus, where possible, keep line lengths under 80 characters (an arbitrary, but pre-existing convention). We make an exception for lines in which manually breaking the line over multiple lines is worse for readability.

chriscoey: "wrapped uncontrollably by a text editor" is pretty unconvincing

chriscoey: yes, please just make it a recommendation. I don't plan to start enforcing line lengths in any packages I author. though for packages that aren't mine, I will follow the surrounding style.

joaquimg:
Maybe we should consider that we use verbose names for everything. We also have many parametric types. And 4 spaces tabs...
I don't like that limit.
I would like to vote for a larger limit at least.
Why 80? wasn't that chosen in the 1930's with no widescreens?

docs/src/style.md
chriscoey: @mlubin it is also "consistent" to not have line limits at all

chriscoey:
@joaquimg makes a great point I hadn't even thought of. 

though I prefer no line limit, I would be OK with a "target/soft maximum line length" of n characters (n = 120 maybe?). if the limit is arbitrary anyway, we may as well be a little soft about it. then n chars is a goal, and clearly a line of 1.5*n chars is too long and should in good conscience be broken up. guidelines like "verbose variable names" require judgment and subjectivity, so why not line lengths too?

the code is littered with ugliness like https://github.com/JuliaOpt/MathOptInterface.jl/blob/f4db9518c49a85b70f72043f01a2bce93d44d981/src/Test/contlinear.jl#L1482 where with a target line length of 80 chars we would be allowed to just keep it on the same line of ~85 chars.

odow:
> the code is littered with ugliness like 
```julia
    obj = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([2.0, 1.0], [x, y]),
                                   0.0)
```

True, that's not very nice. A better way (to my eyes) is:
```julia
    obj = MOI.ScalarAffineFunction(
        MOI.ScalarAffineTerm.([2.0, 1.0], [x, y]), 
        0.0)
```
Breaking things up into smaller lines makes each statement easier to read. Consider:
```julia
    obj = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([2.0, 1.0], [x, y]), 0.0)
    obj = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([2.0, 1.0], [x, y], 0.0))
```

mlubin:
@chriscoey the reference on semantic line breaks is specifically for text. What do semantic line breaks mean for code?

hochB:
Line length style could be copied from @jrevels YASGuide. https://github.com/jrevels/YASGuide
> Try to limit line length to 80 characters per line. This rule should not be enforced arbitrarily, but is rather a soft guideline.


Actually, YASGuide and JuMP's guide have lots in common.
And since JuMP's is now hammering the last nails in the coffin ( https://github.com/JuliaOpt/JuMP.jl/issues/1780 ) it might be worth unifying the guides.

chriscoey:
> > the code is littered with ugliness like
> 
> ```julia
>     obj = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([2.0, 1.0], [x, y]),
>                                    0.0)
> ```
> 
> True, that's not very nice. A better way (to my eyes) is:
> 
> ```julia
>     obj = MOI.ScalarAffineFunction(
>         MOI.ScalarAffineTerm.([2.0, 1.0], [x, y]), 
>         0.0)
> ```

agreed with @odow, that looks nicer. so why don't we recommend *how* to break up lines, so that this outcome results? in particular, 4 spaces indenting only.

joaquimg:
Recommendations on how to break lines would be great!
Softer limit in 80 might be good also, allowing to go to 100 sometimes. (Of course I’d like more columns ;D )

docs/src/style.md
```diff
@@ -144,8 +144,129 @@ function foo(x)
...
+Only use vertical alignment for function arguments. Don't use if in any other
+situation.
+
+Bad:
+```julia
+comprehension = [(first, second) for (first, second) in list_of_things
+                 if isodd(first)]
+```
+
+Beter
```
blegat: Beter -> Better

docs/src/style.md
```diff
@@ -144,8 +144,148 @@ function foo(x)
...
+
+Only use vertical alignment for function arguments. Don't use it in any other
+situation.
+
+Bad:
+```julia
+# Vertical alignment used in a non-function context. Also, indent is not a
+# multiple of 4 spaces.
+comprehension = [(first, second) for (first, second) in list_of_things
+                 if isodd(first)]
```
mlubin: Agreed, why is this bad? https://yapf.now.sh/ aligns python comprehensions vertically with the opening `[`.

docs/src/style.md
```diff
@@ -144,8 +144,148 @@ function foo(x)
...
+```julia
+con_index = MOI.add_constraint(
+    backend(owner_model(variable)), MOI.SingleVariable(index(variable)), set)
+```
+
+Also good:
+```julia
+con_index = MOI.add_constraint(
+    backend(owner_model(variable)), MOI.SingleVariable(index(variable)), set
+)
```
mlubin: https://yapf.now.sh/ never does this for python code, from what I can tell.

docs/src/style.md
```diff
@@ -144,8 +144,148 @@ function foo(x)
...
+
+ - URLs
+ - path names
+ - long string constants not containing whitespace
+ - lines for which breaking over multiple lines worsens readability
+
+#### Line breaks
+
+The "readability" of a line is subjective. In this section we give suggestions
+of good and bad style of how to break a line.
```
mlubin: Maybe cite https://yapf.now.sh/ as a place to be inspired from? Python syntax is typically not that different from Julia.

docs/src/style.md
```diff
@@ -144,8 +144,148 @@ function foo(x)
...
+Therefore, aim to keep line lengths under 80 characters by breaking lines
+for maximum readability (examples are given in the [Line breaks](@ref) section),
+but don't treat this as a hard rule.
+
+We make exceptions for
+
+ - URLs
+ - path names
+ - long string constants not containing whitespace
+ - lines for which breaking over multiple lines worsens readability
```
mlubin: The automatic formatters don't make exceptions like this. I'm ok with this for now but when we have an autoformatter this exception will probably go away.

odow:
Okay, I've removed the contentious points. I'm in favour of merging this so we can move past the discussion.

docs/src/style.md
```diff
@@ -156,17 +156,11 @@ Therefore, aim to keep line lengths under 80 characters by breaking lines
 for maximum readability (examples are given in the [Line breaks](@ref) section),
 but don't treat this as a hard rule.
 
-We make exceptions for
-
- - URLs
```
mlubin: URLs are a valid reason to not break a line. No need to remove this exception.

mlubin:
@joaquimg @blegat PTAL.

@chriscoey Your major objection of not having providing guidelines for how to break a line is now mostly addressed. https://yapf.now.sh/ seems to provide a reasonable point of reference to the extent that julia syntax overlaps with python syntax. What do you think?

joaquimg:
Looks good.

chriscoey:
looks good. 

aside: would be nice to have a Julia version of YAPF

odow:
Merged. Anyone wanting to re-ignite the debate can make a new PR.

PR #1791, opened 2019-01-17
✨ Allow constraint with constant vector

blegat:
Currently, `@constraint(model, 1 >= 0)` works but `@constraint(model, [1 1; 1 1] in PSDCone())` does not.
Constraint with constants are particularly useful in SumOfSquares as we often want to check if a constant polynomial is SOS.

### DIFF
--- a/test/constraint.jl
+++ b/test/constraint.jl
@@ -50,36 +50,57 @@ function constraints_test(ModelType::Type{<:JuMP.AbstractModel},
     end
 
     @testset "AffExpr constraints" begin
-        m = ModelType()
-        @variable(m, x)
+        model = ModelType()
+        @variable(model, x)
 
-        cref = @constraint(m, 2x <= 10)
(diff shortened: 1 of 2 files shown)


test/constraint.jl
```diff
@@ -50,36 +50,57 @@ function constraints_test(ModelType::Type{<:JuMP.AbstractModel},
...
-        m = ModelType()
-        @variable(m, x)
+        model = ModelType()
+        @variable(model, x)
 
-        cref = @constraint(m, 2x <= 10)
-        @test JuMP.name(cref) == ""
-        JuMP.set_name(cref, "c")
-        test_constraint_name(cref, "c", JuMP.AffExpr, MOI.LessThan{Float64})
+        @testset "Scalar" begin
```
mlubin: It's easier to reproduce test failures if the testsets are standalone, i.e., a new model for each testset.

PR #1795, opened 2019-01-19
Fix @SDconstraint with import

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1792

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -651,7 +651,7 @@ function parseSDconstraint(_error::Function, sense::Symbol, lhs, rhs)
         _error("Invalid sense $sense in SDP constraint")
     end
     vectorized = false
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -651,7 +651,7 @@ function parseSDconstraint(_error::Function, sense::Symbol, lhs, rhs)
         _error("Invalid sense $sense in SDP constraint")
     end
     vectorized = false
-    parsecode, buildcall = parse_one_operator_constraint(_error, false, Val(:in), aff, :(PSDCone()))
+    parsecode, buildcall = parse_one_operator_constraint(_error, false, Val(:in), aff, :(JuMP.PSDCone()))
```
mlubin: Doesn't this indicate an escaping issue?

blegat: From the viewpoint of `parse_one_perator_constraint`, the last argument is the set given by the user so it should be in the user scope, not JuMP scope.

PR #1800, opened 2019-01-21
minor edits to constraints.md

mlubin:


docs/src/constraints.md
```diff
@@ -289,11 +284,11 @@ And data, a 2×2 Array{ConstraintRef{Model,C,Shape} where Shape<:AbstractShape w
...
 
-### Dictionaries
+### SparseAxisArrays
 
-The syntax for constructing a dictionary of constraints is very similar to the
-[syntax for constructing](@ref variable_sparseaxisarrays) a dictionary of
-variables.
+The syntax for constructing a `SparseAxisArray` of constraints is very similar
+to the [syntax for constructing](@ref variable_sparseaxisarrays) a
+SparseAxisArray of variables.
```
blegat: Missing backsticks

docs/src/constraints.md
```diff
@@ -289,11 +284,11 @@ And data, a 2×2 Array{ConstraintRef{Model,C,Shape} where Shape<:AbstractShape w
  con[2,2] : 2 x <= 3.0  con[2,3] : 2 x <= 4.0
 ```
 
-### Dictionaries
+### SparseAxisArrays
 
-The syntax for constructing a dictionary of constraints is very similar to the
-[syntax for constructing](@ref variable_sparseaxisarrays) a dictionary of
-variables.
+The syntax for constructing a `SparseAxisArray` of constraints is very similar
```
blegat: Missing `@ref`

docs/src/constraints.md
```diff
@@ -594,16 +592,10 @@ julia> is_valid(model, con)
...
 
-DRAFT: Describe how constraints are represented (link to MOI docs). Constraints
-are very similar to variables in (1) how names work (2) how attributes work, and
-(3) the macro syntax for constructing them. They're a bit different because
-they're parameterized by function-set type. Describe constraints vs.
-`ConstraintRefs`. Describe `JuMP.constraint_object`. How to delete constraints.
-How to modify constraints by setting attributes and `MOI.modifyconstraint!`.
-Describe semidefinite constraints and symmetry handling. Refer to NLP docs for
-nonlinear constraints.
+TODO: Describe constraints vs. `ConstraintRefs`. Describe `JuMP.constraint_object`.
```
blegat: `ConstraintRefs` -> `ConstraintRef`s

docs/src/constraints.md
```diff
@@ -478,14 +437,51 @@ julia> @constraint(model, [t, u, x[1], x[2]] in RotatedSecondOrderCone())
...
+
+julia> @constraint(model, x in MOI.SOS1([1.0, 2.0, 3.0]))
+[x[1], x[2], x[3]] in MathOptInterface.SOS1{Float64}([1.0, 2.0, 3.0])
+```
+Note that we have to pass `MOI.SOS1` a *weight* vector. This vector implies an
+ordering on the variables. If the decision variables are related and have a
+physical ordering (e.g., they correspond to the size of a factory to be built,
+and the SOS-I constraint enforces that only one factory can be built), then the
+weight vector, although not used directly in the constraint, can help the solver
+make better decision in the solution process.
```
Peiffap:
```suggestion
make a better decision in the solution process.
```

mlubin: Fixed, good catch!

mlubin:
@blegat Is the intention that all backticked references to JuMP functions should have a `@ref`? If so that should go in the mini documentation style guide in index.md. We don't do this consistently at all.

PR #1805, opened 2019-01-23
all_constraints and list_of_constraint_types

mlubin:
IMO this is blocking for 0.19 to address gaps in the documentation on how to query what's stored in the model (referenced in constraints.md and nlp.md).

@blegat This gets tricky for vector constraints because we don't store the shape anywhere. What about storing a dictionary from constraint index to shape in Model?

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -693,12 +693,12 @@ julia> @variable(model, x >= 0, Bin);
 julia> @constraint(model, 2x <= 1);
 
 julia> list_of_constraint_types(model)
-3-element Array{Any,1}:
+3-element Array{Tuple{DataType,DataType},1}:
  (VariableRef, MathOptInterface.ZeroOne)
  (VariableRef, MathOptInterface.GreaterThan{Float64})
  (GenericAffExpr{Float64,VariableRef}, MathOptInterface.LessThan{Float64})
 """
 function list_of_constraint_types(model::Model)
     list = MOI.get(model, MOI.ListOfConstraints())
-    return Any[(jump_function_type(model, f), s) for (f,s) in list]
+    return [(jump_function_type(model, f), s) for (f,s) in list]
 end
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -269,7 +269,6 @@ end
 
 #############################################################################
 # AbstractConstraint
-#
 """
     abstract type AbstractConstraint
 
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -274,7 +274,7 @@ end
 
 An abstract base type for all constraint types. `AbstractConstraint`s store the
 function and set directly, unlike [`ConstraintRef`](@ref)s that are merely
-references to constraints stored in a model. `AbstractConstraint` do not need
+references to constraints stored in a model. `AbstractConstraint`s do not need
 to be attached to a model.
 """
 abstract type AbstractConstraint end
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -269,7 +269,15 @@ end
 
 #############################################################################
 # AbstractConstraint
-# Abstract base type for all constraint types
+#
+"""
+    abstract type AbstractConstraint
+
+An abstract base type for all constraint types. `AbstractConstraint`s store the
+function and set directly, unlike [`ConstraintRef`](@ref)s that are merely
+references to constraints stored in a model. `AbstractConstraint` do not need
+to be attached to a model.
+"""
 abstract type AbstractConstraint end
 
 """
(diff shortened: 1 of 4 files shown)


src/constraints.jl
```diff
@@ -587,3 +587,85 @@ function shadow_price(constraint::ConstraintRef{Model, MOICON{F, S}}
...
+        error("$t is not a concrete type. Did you miss a type parameter?")
+    end
+    return
+end
+
+"""
+    all_constraints(model::Model, function_type, set_type)::Vector{VariableRef}
+
+Return a list of all constraints currently in the model where the function
+has type `function_type` and the set has type `set_type`. The constrtaints are
```
odow: s/constrtaints/constraints

src/constraints.jl
```diff
@@ -587,3 +587,85 @@ function shadow_price(constraint::ConstraintRef{Model, MOICON{F, S}}
...
+
+julia> list_of_constraint_types(model)
+3-element Array{Tuple{DataType,DataType},1}:
+ (VariableRef, MathOptInterface.ZeroOne)
+ (VariableRef, MathOptInterface.GreaterThan{Float64})
+ (GenericAffExpr{Float64,VariableRef}, MathOptInterface.LessThan{Float64})
+"""
+function list_of_constraint_types(model::Model)
+    list = MOI.get(model, MOI.ListOfConstraints())
+    return [(jump_function_type(model, f), s) for (f,s) in list]
```
blegat: Annotate the element type `Tuple{DataType, DataType}[(jump....`

test/constraint.jl
```diff
@@ -377,6 +377,31 @@ end
...
+    @testset "all_constraints (scalar)" begin
+        model = Model()
+        @variable(model, x >= 0)
+        ref = all_constraints(model, VariableRef, MOI.GreaterThan{Float64})
+        @test ref == [LowerBoundRef(x)]
+        aff_constraints = all_constraints(model, AffExpr,
+                                          MOI.GreaterThan{Float64})
+        @test isempty(aff_constraints)
+        # GreaterThan isn't a concrete type.
+        @test_throws ErrorException all_constraints(model, AffExpr,
```
blegat: Since we dropped Julia v0.6, we can now do
```julia
err = ErrorException("MOI.GreaterThan is not a concrete type. Did you miss a type parameter?")
@test_throws err ...
```
To make sure that this error is thrown instead of another one

src/constraints.jl
```diff
@@ -587,3 +587,85 @@ function shadow_price(constraint::ConstraintRef{Model, MOICON{F, S}}
         return shadow_price_less_than_(dual_val, sense)
     end
 end
+
+function _error_if_not_concrete_type(t)
+    if !isconcretetype(t)
+        error("$t is not a concrete type. Did you miss a type parameter?")
```
blegat: Backsticks around `$t`

src/constraints.jl
mlubin: That type seems more confusing than helpful. Using `Any` instead.

src/constraints.jl
```diff
@@ -587,3 +587,85 @@ function shadow_price(constraint::ConstraintRef{Model, MOICON{F, S}}
...
+
+julia> list_of_constraint_types(model)
+3-element Array{Any,1}:
+ (VariableRef, MathOptInterface.ZeroOne)
+ (VariableRef, MathOptInterface.GreaterThan{Float64})
+ (GenericAffExpr{Float64,VariableRef}, MathOptInterface.LessThan{Float64})
+"""
+function list_of_constraint_types(model::Model)
+    list = MOI.get(model, MOI.ListOfConstraints())
+    return Any[(jump_function_type(model, f), s) for (f,s) in list]
```
blegat: What is the point of using `Any` ? It will be is more efficient to use `Tuple{DataType, DataType}` and IMO, it also tells the user that he gets a pair on concrete types so it is helpful.

mlubin: Ok, maybe it's not necessary to use `Any`. I don't understand the suggestion to annotate with an ugly element type though. Is there an implicit style point on annotating all comprehensions with types?

odow: I don't think it's necessary to type. The docstring should tell the user, they shouldn't have to guess from complicated type information.

> Is there an implicit style point on annotating all comprehensions with types?

Only do it if the return type is important and would otherwise be type-unstable.

blegat: It is not about style nor for user documentation, it is for type stability/performance

mlubin: @blegat Given how generic the type is, how can it specifying it have a performance impact? It's not a concrete type of small union of concrete types so it's unlikely the compiler will specialize any code for it.

blegat: Not sure but having the function inferrable certainly wouldn't hurt. If there are tests with `@inferred` that pass without the annotation, we can remove it

mlubin: > getindex on an element will be slow since it will have to find at runtime which getindex method to call

So? I'm having difficulty imagining a realistic case where that has any meaning impact on runtime. I would need to see a benchmark to be convinced otherwise. IMO it's inconsistent to test this one relatively minor method with `@inferred` when we don't test other much more important methods like `value` with `@inferred`.

mlubin:
This is ready to go on my side. The vector versions are blocked by not having shape information, which can be addressed separately.

src/constraints.jl
odow: I agree with @mlubin.

PR #1808, opened 2019-01-23
No need for localvar with Julia >= 1.0

xhub:
As noted in
https://github.com/JuliaOpt/JuMP.jl/pull/1216#issuecomment-379089680
there is no need for localvar anymore.

This change is need for the EMP.jl package, in which I have macros that are calling JuMP macros. Without this fix, the JuMP macros is trying to create a local variable like ```EMP:.##43##``` with Julia >= 0.7, and it fails. With Julia 0.6, the code works fine.

There may be another way to fix this, but removing the function call seems to be the way forward.

This pass all tests. If this change is accepted, I'll do a PR for the master branch ( where localvar is also marked as removable, see https://github.com/JuliaOpt/JuMP.jl/blob/master/src/macros.jl#L192 )

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -234,25 +234,30 @@ function getloopedcode(varname, code, condition, idxvars, idxsets, idxpairs, sym
     end
 end
 
-localvar(x::Symbol) = _localvar(x)
-localvar(x::Expr) = Expr(:block, _localvar(x)...)
-_localvar(x::Symbol) = :(local $(esc(x)))
-function _localvar(x::Expr)
-    @assert x.head in (:escape,:tuple)
-    args = Any[]
-    for t in x.args
-        if isa(t, Symbol)
-            push!(args, :(local $(esc(t))))
-        else
-            @assert isa(t, Expr)
-            if t.head == :tuple
-                append!(args, map(_localvar, t.args))
+# There is no need for localvar with >=0.7 since the loop variable are always local
+if VERSION < v"0.7"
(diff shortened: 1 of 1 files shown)


xhub:
Thanks, will do.

xhub:
Need to have a look at the failures ...

odow:
> This change is need for the EMP.jl package, in which I have macros that are calling JuMP macros.

I also had this problem with SDDP.jl! It blocked me from upgrading it to Julia 1.0. I tried a few things but got confused with all the scoping/hygiene issues so never tracked down the root cause. I'll be pleased if this works!

xhub:
Hm ... I need to better with the new Pkg, I tested the wrong version (at least on 0.7).

So the change of behaviour happened in 1.0, not 0.7, as one can see with
```julia
toto = 4
for inner toto=1:3
       println(toto)
 end
println(toto)
```

Therefore, I'm going to update the version check to < v"1.0" and call it a day, I am fine with my software not working on 0.7.

xhub:
> 
> I also had this problem with SDDP.jl! It blocked me from upgrading it to Julia 1.0. I tried a few things but got confused with all the scoping/hygiene issues so never tracked down the root cause. I'll be pleased if this works!

I made it work for my use case. I still need to polish one aspect, I'll ping you when I am done

odow:
Closing as out-dated. We won't be making new releases to the 0.18 branch.

Issue #1790, opened 2019-01-17
sanity check JuMP's exports


mlubin:
Now that we export everything without a starting underscore, we should sanity check that all exported functions are intended to be public. Removing an export is usually considered a breaking change, so this is blocking for 0.19.

Example candidates to unexport in macros.jl: `issum`, `isprod`, `error_curly`, `buildrefsets`, `getloopedcode`. (This is just from a quick look.)

Sanitized files
- [ ] JuMP.jl
- [x] aff_expr.jl https://github.com/JuliaOpt/JuMP.jl/pull/1833
- [x] constraints.jl
- [x] copy.jl https://github.com/JuliaOpt/JuMP.jl/pull/1834
- [x] macros.jl https://github.com/JuliaOpt/JuMP.jl/pull/1813
- [x] nlp.jl
- [x] objective.jl
- [ ] operators.jl https://github.com/JuliaOpt/JuMP.jl/pull/1837
- [x] optimizer_interface.jl
- [x] parse_expr.jl https://github.com/JuliaOpt/JuMP.jl/pull/1838
- [x] parse_nlp.jl
- [x] print.jl https://github.com/JuliaOpt/JuMP.jl/pull/1835
- [x] quad_expr.jl https://github.com/JuliaOpt/JuMP.jl/pull/1833
- [x] sd.jl
- [x] sets.jl
- [x] utils.jl https://github.com/JuliaOpt/JuMP.jl/pull/1814
- [x] variables.jl https://github.com/JuliaOpt/JuMP.jl/pull/1830

odow:
Question. Structs like `VariableInfoExpr` are private right? In which case, should they be renamed `_VariableInfoExpr`?
https://github.com/JuliaOpt/JuMP.jl/blob/a53cf4d9e88b682578a1139740fb1b1f0e0506f0/src/variables.jl#L9-L20

There is a loooot of renaming to do...

mlubin:
> Structs like VariableInfoExpr are private right? In which case, should they be renamed _VariableInfoExpr?

I'd say yes. This struct is used internally for macros. It might be used by JuMP extensions but I'd still say that's private.

blegat:
Yes, I'd say it's private

odow:
Previously we had things like `JuMP.delete` so we didn't have to overload `Base.delete`. But now export user-facing functions suggests that we should export `delete` which implies that we should overload `Base.delete`... Thoughts?

mlubin:
[...]

PR #1813, opened 2019-01-28
[refactor] make methods private in macros.jl

odow:
Part of #1790

What should we do with extendable methods like `_build_constraint` and `_build_variable`?

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -714,7 +714,7 @@ macro SDconstraint(args...)
 end
 
 """
-    @_build_constraint(constraint_expr)
+    @build_constraint(constraint_expr)
 
 Constructs a `ScalarConstraint` or `VectorConstraint` using the same
 machinery as [`@constraint`](@ref) but without adding the constraint to a model.
@@ -727,12 +727,12 @@ create arrays of `ScalarConstraint` or `VectorConstraint`.
 ```jldoctest; setup = :(using JuMP)
 model = Model();
 @variable(model, x);
-@_build_constraint(2x >= 1)
+@build_constraint(2x >= 1)
 
 # output
 ScalarConstraint{GenericAffExpr{Float64,VariableRef},MathOptInterface.GreaterThan{Float64}}(2 x, MathOptInterface.GreaterThan{Float64}(1.0))
 """
-macro _build_constraint(constraint_expr)
+macro build_constraint(constraint_expr)
     _error(str...) = _macro_error(:_build_constraint, args, str...)
 
     if isa(constraint_expr, Symbol)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -337,7 +337,7 @@ function _check_vectorized(sense::Symbol)
     end
 end
 
-# two-argument _build_constraint is used for one-sided constraints.
+# two-argument build_constraint is used for one-sided constraints.
 # Right-hand side is zero.
 _sense_to_set(_error::Function, ::Union{Val{:(<=)}, Val{:(≤)}}) = MOI.LessThan(0.0)
 _sense_to_set(_error::Function, ::Union{Val{:(>=)}, Val{:(≥)}}) = MOI.GreaterThan(0.0)
@@ -349,9 +349,9 @@ function _parse_one_operator_constraint(_error::Function, vectorized::Bool,
     newaff, parseaff = parseExprToplevel(aff, :q)
     parsecode = :(q = Val{false}(); $parseaff)
     if vectorized
-        buildcall = :(_build_constraint.($_error, $newaff, Ref($(esc(set)))))
+        buildcall = :(build_constraint.($_error, $newaff, Ref($(esc(set)))))
     else
-        buildcall = :(_build_constraint($_error, $newaff, $(esc(set))))
+        buildcall = :(build_constraint($_error, $newaff, $(esc(set))))
     end
     parsecode, buildcall
 end
@@ -363,7 +363,7 @@ function _parse_one_operator_constraint(_error::Function, vectorized::Bool, sens
     _parse_one_operator_constraint(_error, vectorized, Val(:in), aff, set)
 end
 
-function _parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
+function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
     (sense, vectorized) = _check_vectorized(sense)
     vectorized, _parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
 end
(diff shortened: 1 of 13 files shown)


blegat:
How does one distinguish between a method meant to be extended and an internal method ? Maybe we shouldn't underscore functions of the JuMP extension API.

blegat:
Note that it is perfectly fine to create constraint with build_constraint and add_constraint. The @constraint macro simply adds convenience. So it is not like private functions that the user shouldn't use

blegat:
> Will you commit to not break anything between minor JuMP releases? If so and if the function is documented then it can be exported.

Certainly, that would allow JuMP extension to safely use e.g. `JuMP 0.19 0.20` in their `REQUIRE` to prevent breakage.

> I suggest we keep them as internal features for now, and export them in a future release once we're convinced that the API is stable.

I am not excited to go back and forth with renaming with corresponding updates to extensions, clashes if the wrong branches are checkout... I would prefer keeping extension functions exported.

mlubin:
@blegat effectively owns this part of the API, so I'll leave the decision on underscoring to him.

odow:
@blegat, can you make a list of extension functions you consider to be part of the public API? I restored `build_constraint` and `parse_constraint`.

src/macros.jl
```diff
@@ -339,13 +339,13 @@ end
...
 # two-argument build_constraint is used for one-sided constraints.
 # Right-hand side is zero.
-sense_to_set(_error::Function, ::Union{Val{:(<=)}, Val{:(≤)}}) = MOI.LessThan(0.0)
-sense_to_set(_error::Function, ::Union{Val{:(>=)}, Val{:(≥)}}) = MOI.GreaterThan(0.0)
-sense_to_set(_error::Function, ::Val{:(==)}) = MOI.EqualTo(0.0)
-sense_to_set(_error::Function, ::Val{S}) where S = _error("Unrecognized sense $S")
+_sense_to_set(_error::Function, ::Union{Val{:(<=)}, Val{:(≤)}}) = MOI.LessThan(0.0)
+_sense_to_set(_error::Function, ::Union{Val{:(>=)}, Val{:(≥)}}) = MOI.GreaterThan(0.0)
+_sense_to_set(_error::Function, ::Val{:(==)}) = MOI.EqualTo(0.0)
+_sense_to_set(_error::Function, ::Val{S}) where S = _error("Unrecognized sense $S")
```
blegat: I would leave this one exported

src/macros.jl
```diff
@@ -339,13 +339,13 @@ end
...
 
-function parse_one_operator_constraint(_error::Function, vectorized::Bool,
-                                       ::Union{Val{:in}, Val{:∈}}, aff, set)
+function _parse_one_operator_constraint(_error::Function, vectorized::Bool,
```
blegat: This one too

src/macros.jl
```diff
@@ -356,19 +356,19 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
...
 end
 
 function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
     (sense, vectorized) = _check_vectorized(sense)
-    vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
+    vectorized, _parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
 end
 
-function parse_ternary_constraint(_error::Function, vectorized::Bool, lb, ::Union{Val{:(<=)}, Val{:(≤)}}, aff, rsign::Union{Val{:(<=)}, Val{:(≤)}}, ub)
+function _parse_ternary_constraint(_error::Function, vectorized::Bool, lb, ::Union{Val{:(<=)}, Val{:(≤)}}, aff, rsign::Union{Val{:(<=)}, Val{:(≤)}}, ub)
```
blegat: This one too

src/macros.jl
```diff
@@ -563,28 +563,28 @@ function constraint_macro(args, macro_name::Symbol, parsefun::Function)
...
+        macro_code = _macro_assign_and_return(creationcode, variable, name,
                                              model_for_registering = m)
     end
-    return assert_validmodel(m, macro_code)
+    return _assert_valid_model(m, macro_code)
 end
 
 # This function needs to be implemented by all `AbstractModel`s
-constraint_type(m::Model) = ConstraintRef{typeof(m)}
+_constraint_type(m::Model) = ConstraintRef{typeof(m)}
```
blegat: This one too

src/macros.jl
```diff
@@ -995,44 +982,42 @@ macro expression(args...)
...
+_esc_non_constant(x::Expr) = isexpr(x,:quote) ? x : esc(x)
+_esc_non_constant(x) = esc(x)
 
-# Returns the type of what `add_variable(::Model, build_variable(...))` would return where `...` represents the positional arguments.
-# Example: `@variable m [1:3] foo` will allocate an vector of element type `variable_type(m, foo)`
+# Returns the type of what `add_variable(::Model, _build_variable(...))` would return where `...` represents the positional arguments.
+# Example: `@variable m [1:3] foo` will allocate an vector of element type `_variable_type(m, foo)`
 # Note: it needs to be implemented by all `AbstractModel`s
-variable_type(m::Model) = VariableRef
+_variable_type(m::Model) = VariableRef
```
blegat: This one too

src/macros.jl
```diff
@@ -995,44 +982,42 @@ macro expression(args...)
...
 # Returns a new variable. Additional positional arguments can be used to dispatch the call to a different method.
 # The return type should only depends on the positional arguments for `variable_type` to make sense. See the @variable macro doc for more details.
-# Example: `@variable m x` foo will call `build_variable(_error, info, foo)`
-function build_variable(_error::Function, info::VariableInfo; extra_kw_args...)
+# Example: `@variable m x` foo will call `_build_variable(_error, info, foo)`
+function _build_variable(_error::Function, info::VariableInfo; extra_kw_args...)
```
blegat: This one too

src/macros.jl
```diff
@@ -1048,21 +1033,21 @@ function namecall(base_name, idxvars)
...
-reverse_sense(::Val{:(==)}) = Val(:(==))
+_reverse_sense(::Val{:<=})   = Val(:>=)
+_reverse_sense(::Val{:≤})    = Val(:≥)
+_reverse_sense(::Val{:>=})   = Val(:<=)
+_reverse_sense(::Val{:≥})    = Val(:≤)
+_reverse_sense(::Val{:(==)}) = Val(:(==))
 
 """
-    parse_one_operator_variable(_error::Function, infoexpr::VariableInfoExpr, sense::Val{S}, value) where S
+    _parse_one_operator_variable(_error::Function, infoexpr::VariableInfoExpr, sense::Val{S}, value) where S
```
blegat: All the `parse_variable` API too

odow:
Can we have a rule for which ones should and shouldn't be? A lot of these are undocumented... Maybe you want to take over this PR and export the ones you want and add docstrings to them? I'm flying blind here...

blegat:
All the parse/build/add API should the exported and also variable_type and constraint_type since it is used to extend the variable and constraint macros.
I can take over on Sundays but not before

odow:
Can we make a rule that all exported functions need to have docstrings? For example, why is `sense_to_set` exported? What is the use-case for overloading this?

blegat:
> Can we make a rule that all exported functions need to have docstrings?

I have created an issue. Edit the issue if I missed anything.

> For example, why is sense_to_set exported? What is the use-case for overloading this?

Defining the meaning of a new sense symbol.

PR #1815, opened 2019-01-28
Tweak computation of products.

### LINKED ISSUE #1811: Nonzero elements in the Hessian that should be zero
The code
```
using JuMP, MathProgBase
model = Model()
@variable(model, x[1:2])
@NLobjective(model, Min, 0.0)
@NLconstraint(model, x[1] * x[2] == 2)
d = JuMP.NLPEvaluator(model)
MathProgBase.initialize(d, [:Hess])
hx = zeros(3)
MathProgBase.eval_hesslag(d, hx, [0.659; 0.702], 1.0, [1.0])
```
corresponds to problem
```
min 0   s.t.  x_1 * x_2 == 2
```
which has Hessian of Lagrangian
```
\nabla^2 L(x,y) = [0 y_1; y_1 0]
```
i.e., only one nonzero is stored. However, a small error appears on the diagonal. Varies with the inputs.

Related to JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/13.

mlubin:
Remove a source of numerical imprecision in derivatives of products.
Closes #1811 

This was arguably not a bug because the derivatives were correct up to numerical tolerances, but I made a small tweak to remove the source of error in this case.

@abelsiqueira

### DIFF
--- a/src/Derivatives/forward.jl
+++ b/src/Derivatives/forward.jl
@@ -83,7 +83,13 @@ function forward_eval(storage::AbstractVector{T}, partials_storage::AbstractVect
                 for c_idx in children_idx
                     @inbounds tmp_prod *= storage[children_arr[c_idx]]
                 end
-                if tmp_prod == zero(T) # inefficient
+                if tmp_prod == zero(T) || n_children <= 2
+                    # This is inefficient if there are a lot of children.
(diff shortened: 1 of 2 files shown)


abelsiqueira:
Will this still be released for 0.18?

PR #1816, opened 2019-01-29
Remove dead code

odow:
I found this, which appears to be a hold-over from the early MOI days and doesn't seem used in the current code base. Is it used in extensions of something? @blegat


mlubin:
This was originally intended for representing solution vectors, e.g., you could create one of these and set values, then provide it to the model as a starting point or for checking feasibility (https://github.com/JuliaOpt/JuMP.jl/issues/693). Or you get a few of these back if the solver finds multiple solutions. Anyway this didn't go far in 0.19. It's safe to remove.

PR #1828, opened 2019-02-06
rename private functions in parse_nlp.jl and nlp.jl

mlubin:
Ref #1790


blegat:
Missing an update here:
https://github.com/JuliaOpt/JuMP.jl/blob/f46a461c126fd1a7c309fb773a00b5dc529632b9/src/optimizer_interface.jl#L105

PR #1829, opened 2019-02-06
Remove unused method check_belongs_to_model

ndinsmore:
I noticed that the vector case of check_belongs_to_model had a bug in it.   So I fixed it and added tests.

Please note that the vector method returns a Bool while the scalar method throws an exception or returns nothing.  I left that implementation detail the same.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -475,11 +475,7 @@ function Base.showerror(io::IO, ex::VariableNotOwnedError)
 end
 
 function check_belongs_to_model(vec::Vector{VariableRef}, m::Model)
-    n = length(vec)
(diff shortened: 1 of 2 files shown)


odow:
Was this specific method used anywhere? If not, it can probably be removed. 

@blegat probably knows after touching https://github.com/JuliaOpt/JuMP.jl/pull/1765

blegat:
This method should be removed, it was a function called `verify_ownership` that wasn't used but was renamed in https://github.com/JuliaOpt/JuMP.jl/pull/1765.

ndinsmore:
I have updated the pull to just remove the method instead.

PR #1835, opened 2019-02-07
Sanitize print

blegat:
See https://github.com/JuliaOpt/JuMP.jl/issues/1790

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -29,25 +29,25 @@ abstract type IJuliaMode <: PrintMode end
 
 # Whether something is zero or not for the purposes of printing it
 # oneunit is useful e.g. if coef is a Unitful quantity.
-is_zero_for_printing(coef) = abs(coef) < 1e-10 * oneunit(coef)
+_is_zero_for_printing(coef) = abs(coef) < 1e-10 * oneunit(coef)
 # Whether something is one or not for the purposes of printing it.
-is_one_for_printing(coef) = is_zero_for_printing(abs(coef) - oneunit(coef))
-sign_string(coef) = coef < zero(coef) ? " - " : " + "
+_is_one_for_printing(coef) = _is_zero_for_printing(abs(coef) - oneunit(coef))
+_sign_string(coef) = coef < zero(coef) ? " - " : " + "
 
 # Helper function that rounds carefully for the purposes of printing
 # e.g.   5.3  =>  5.3
 #        1.0  =>  1
-function string_round(f::Float64)
+function _string_round(f::Float64)
     iszero(f) && return "0" # strip sign off zero
     str = string(f)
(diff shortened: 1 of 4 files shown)


src/print.jl
```diff
@@ -140,10 +140,10 @@ function math_symbol(::Type{IJuliaMode}, name::Symbol)
     end
 end
 
-wrap_in_math_mode(str) = "\$\$ $str \$\$"
-wrap_in_inline_math_mode(str) = "\$ $str \$"
+_wrap_in_math_mode(str) = "\$\$ $str \$\$"
+_wrap_in_inline_math_mode(str) = "\$ $str \$"
 
-plural(n) = (n==1 ? "" : "s")
+plural(n) = (n == 1 ? "" : "s")
```
mlubin: Unexport this?

src/print.jl
```diff
@@ -283,39 +283,39 @@ end
 
 Base.show(io::IO, a::GenericAffExpr) = print(io, aff_string(REPLMode,a))
 function Base.show(io::IO, ::MIME"text/latex", a::GenericAffExpr)
-    print(io, wrap_in_math_mode(aff_string(IJuliaMode, a)))
+    print(io, _wrap_in_math_mode(aff_string(IJuliaMode, a)))
 end
 
 function aff_string(mode, a::GenericAffExpr, show_constant=true)
```
mlubin: I'd vote to unexport `aff_string`.

src/print.jl
```diff
@@ -326,7 +326,7 @@ end
 #------------------------------------------------------------------------
 Base.show(io::IO, q::GenericQuadExpr) = print(io, quad_string(REPLMode,q))
 function Base.show(io::IO, ::MIME"text/latex", q::GenericQuadExpr)
-    print(io, wrap_in_math_mode(quad_string(IJuliaMode, q)))
+    print(io, _wrap_in_math_mode(quad_string(IJuliaMode, q)))
 end
 
 function quad_string(mode, q::GenericQuadExpr)
```
mlubin: Same for `quad_string`.

src/print.jl
```diff
@@ -484,21 +484,21 @@ function function_string(print_mode, constraint::AbstractConstraint)
 end
 
 function in_set_string(print_mode, set::MOI.LessThan)
```
mlubin: Unexport?

src/print.jl
blegat: I was hesitant since it is used in `JuMPExtension.jl`

src/print.jl
blegat: It is part of the extension API, when an extension defines a new set, it can a a method to this function to give a nice printing

src/print.jl
blegat: Then we might as well remove it, see https://github.com/JuliaOpt/JuMP.jl/pull/1836

PR #1836, opened 2019-02-07
Remove aff_string, quad_string and var_string

blegat:


blegat:
An extension defining a new scalar function should add a method to `function_string`

mlubin:
Bigger picture we should have a way to distinguish between functions that come up in normal JuMP use and functions that are useful only for extensions. But this is ok for now.

PR #1837, opened 2019-02-07
🚸 Sanitize operators

blegat:
See https://github.com/JuliaOpt/JuMP.jl/issues/1790


mlubin:
Approval assumes failing tests will be fixed :)

PR #1840, opened 2019-02-07
avoid creating objects with duplicate names

### LINKED ISSUE #1839: variables with duplicate names are added to the model anyway
```julia

julia> m = Model();

julia> @variable(m, x)
x

julia> @variable(m, x)
ERROR: An object of name x is already attached to this model. If this is intended, consider using the anonymous construction syntax, e.g., x = @variable(model, [1:N], ...) where the name of the object does not appear inside the macro.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] register_object(::Model, ::Symbol, ::VariableRef) at /home/mlubin/.julia/dev/JuMP/src/JuMP.jl:620
 [3] top-level scope at /home/mlubin/.julia/dev/JuMP/src/macros.jl:322

julia> all_variables(m)
2-element Array{VariableRef,1}:
 x
 x
```

Reported by @mtanneau

mlubin:
Closes #1839

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -303,6 +303,21 @@ function _macro_return(code, variable)
     end
 end
 
+function _error_if_cannot_register(model::AbstractModel, name::Symbol)
+    obj_dict = object_dictionary(model)
+    if haskey(obj_dict, name)
+        error("An object of name $name is already attached to this model. " *
+              "If this is intended, consider using the anonymous construction" *
+              " syntax, e.g., x = @variable(model, [1:N], ...) where the " *
+              "name of the object does not appear inside the macro.")
+    end
+    return
+end
+
+function _error_if_cannot_register(model::AbstractModel, name)
+    error("Invalid name $name.")
+end
+
 """
     _macro_assign_and_return(code, variable, name;
                              final_variable=variable,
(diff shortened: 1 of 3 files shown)


src/macros.jl
```diff
@@ -303,25 +303,44 @@ function _macro_return(code, variable)
...
+        $(if model_for_registering !== nothing
+            :(_error_if_cannot_register($model_for_registering,
+                                        $(quot(name))))
+          end)
         $variable = $macro_code
         $(if model_for_registering !== nothing
-              :(register_object($model_for_registering, $(quot(name)),
-                $variable))
+            :(object_dictionary($model_for_registering)[$(quot(name))] =
+              $final_variable)
```
blegat: Should it be `variable` instead of `final_variable` ?

mlubin: The only case I found where `final_variable` is different from `variable` is here:
https://github.com/JuliaOpt/JuMP.jl/blob/2239b48ab4bb4b753c712a9080f6343545e29fbb/src/macros.jl#L1383

We want the lookup to return the `Symmetric` object in this case, right?

blegat: Indeed, we should have a test for that

mlubin: Opened https://github.com/JuliaOpt/JuMP.jl/issues/1842.

PR #1841, opened 2019-02-08
Sanitize JuMP.jl

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1790

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -46,7 +46,7 @@ const _MOIFIX = _MOICON{MOI.SingleVariable,MOI.EqualTo{Float64}}
 const _MOIINT = _MOICON{MOI.SingleVariable,MOI.Integer}
 const _MOIBIN = _MOICON{MOI.SingleVariable,MOI.ZeroOne}
 
-MOIU.@model(JuMPMOIModel,
+MOIU.@model(_MOIModel,
             (MOI.ZeroOne, MOI.Integer),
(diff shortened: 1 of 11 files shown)


src/JuMP.jl
```diff
@@ -38,13 +38,13 @@ Base.@deprecate(linearterms,       JuMP.linear_terms)
...
-const MOIBIN = MOICON{MOI.SingleVariable,MOI.ZeroOne}
+const _MOIVAR = MOI.VariableIndex
+const _MOICON{F,S} = MOI.ConstraintIndex{F,S}
+const _MOILB = _MOICON{MOI.SingleVariable,MOI.GreaterThan{Float64}}
+const _MOIUB = _MOICON{MOI.SingleVariable,MOI.LessThan{Float64}}
+const _MOIFIX = _MOICON{MOI.SingleVariable,MOI.EqualTo{Float64}}
+const _MOIINT = _MOICON{MOI.SingleVariable,MOI.Integer}
+const _MOIBIN = _MOICON{MOI.SingleVariable,MOI.ZeroOne}
 
 MOIU.@model(JuMPMOIModel,
```
mlubin: Unexport `JuMPMOIModel` also?

blegat: I am unsure since it might be useful if someone want to use a CachingOptimizer in direct mode. However, why would someone use direct mode if a CachingOptimizer is used anyway...
If we unexport, it should be `JuMP._MOIModel` since `JuMP._JuMPMOIModel` is redundant.

PR #1844, opened 2019-02-10
✅ Add test for variable lookup with Symmetric

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1842


test/variable.jl
```diff
@@ -462,6 +462,9 @@ function test_variable_symmetric(ModelType)
     @variable model x[1:2, 1:2] Symmetric
     @test x isa Symmetric
     @test x[1, 2] === x[2, 1]
+    @testset "Variable lookup with Symmetric" begin
```
mlubin: Why is this test in a testset while the others aren't? It should either be standalone with a testset or listed like the other tests in this function.

PR #1847, opened 2019-02-10
Fix list_of_constraint_types with vector constraints

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1845

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -295,15 +295,16 @@ function jump_function_type(::Model,
                             ::Type{MOI.ScalarAffineFunction{T}}) where T
     return GenericAffExpr{T, VariableRef}
 end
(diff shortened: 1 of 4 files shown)


src/aff_expr.jl
```diff
@@ -291,13 +291,17 @@ function AffExpr(m::Model, f::MOI.ScalarAffineFunction)
     aff.constant = f.constant
     return aff
 end
-function jump_function_type(::AbstractModel,
+function jump_function_type(::Model,
```
mlubin: Why does `jump_function` use `AbstractModel` and `jump_function_type` use `Model`? Is this explained anywhere?

blegat:
Merged as the `jump_function_type` methods were needed for https://github.com/JuliaOpt/JuMP.jl/pull/1850

PR #1848, opened 2019-02-11
Fix bug in build_constraint error

odow:


test/macros.jl
```diff
@@ -270,6 +270,12 @@ function macros_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Typ
         @test con[2].set == MOI.LessThan(2.0)
     end
 
+    @testset "@build_constraint error" begin
+        model = ModelType()
+        @variable(model, x)
+        @test_throws LoadError @macroexpand @build_constraint(2x + 1)
```
blegat: We have test_macro_throws in utilities to check for errors in macros

PR #1849, opened 2019-02-11
DNMY: add solver-independent callbacks

odow:
This PR re-introduces a limited form of solver-independent(-ish) callbacks in JuMP.

See the following PR for the API added to MOI.

MathOptInterface: https://github.com/JuliaOpt/MathOptInterface.jl/pull/670

See the following PRs for the solver-specific implementations.

GLPK: https://github.com/JuliaOpt/GLPK.jl/pull/91
Gurobi: https://github.com/JuliaOpt/Gurobi.jl/pull/194

#### What these are 

A convenient way of accessing some limited functionality of the lazy constraint and heuristic callbacks. 

These are solver-independent to the extent that

*Lazy callback*
- the lazy constraint callback *may* be called when the solver has an integer primal solution
- you can only access the primal solution through `JuMP.value`. That is, `JuMP.dual` etc will not work.
- when the solver terminates, it will provide a solution that satisfies all lazy constraints
- we make no claims about what happens to solutions during the B&B tree

*Heuristic callback*
- the heuristic callback *may* be called when the solver has a non-integer solution
- the solver may accept or reject the solution
- some solvers require a complete solution, others only partial solutions. It's up to you to provide the appropriate one. If in doubt, give a complete solution

#### What these are not

Generic solver-independent callbacks. If you want more control, use the solver-specific ones.

#### Requirements before merging

- [ ] MOI PR is merged and tagged
- [ ] GLPK PR is merged and tagged
- [ ] documentation

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1553

### DIFF
--- a/examples/callbacks_glpk.jl
+++ b/examples/callbacks_glpk.jl
@@ -28,7 +28,7 @@ model = JuMP.direct_model(GLPK.Optimizer())
 # gets called.
 cb_calls = Symbol[]
 
-function my_lazy_callback(model, cb_data)
+function my_lazy_callback(cb_data)
     push!(cb_calls, :lazy)
     # Double check for sanity's sake that we have a feasible (given the
     # current constraints) point.
@@ -44,7 +44,7 @@ function my_lazy_callback(model, cb_data)
     end
 end
 
-function my_heuristic_callback(model, cb_data)
+function my_heuristic_callback(cb_data)
     push!(cb_calls, :heuristic)
     # Double check for sanity's sake that we have a feasible (given the
     # current constraints) point.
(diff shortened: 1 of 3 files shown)


blegat:
Why do we need to call `GLPK.@load_extensions` and instead of just including `GLPKExtensions.jl` in `GLPK.jl` ? Is it because we don't want JuMP to be a GLPK dependency ?
Then why don't we define the callback API in MOI instead of JuMP ?

odow:
@blegat I have removed the fancy macros and added a callback API to MOI.

examples/callbacks_glpk.jl
```diff
@@ -0,0 +1,69 @@
...
+
+# This vector is going to cache a descriptive symbol everytime our callback gets
+# called.
+cb_calls = Symbol[]
+
+function my_lazy_callback(cb_data)
+    push!(cb_calls, :lazy)
+    # Double check for sanity's sake that we have a feasible (given the
+    # current constraints) point.
+    @assert primal_status(model) == MOI.FEASIBLE_POINT
```
odow: Q: what should `termination_status(model)` be at this point? `MOI.INTERRUPTED`?

PR #1850, opened 2019-02-11
✨ Store shape information

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1817
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1723

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -337,7 +337,7 @@ function MOI.VectorAffineFunction(affs::Vector{AffExpr})
     MOI.VectorAffineFunction(terms, constant)
 end
 moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
-function moi_function_type(::Type{Vector{Aff}}) where {T, Aff <: GenericAffExpr{T}}
+function moi_function_type(::Type{<:Vector{<:GenericAffExpr{T}}}) where {T}
     return MOI.VectorAffineFunction{T}
 end
 
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -9,95 +9,6 @@
 #############################################################################
 
 """
-    AbstractShape
-
-Abstract vectorizable shape. Given a flat vector form of an object of shape
-`shape`, the original object can be obtained by [`reshape_result`](@ref).
-"""
-abstract type AbstractShape end
-
-"""
-    dual_shape(shape::AbstractShape)::AbstractShape
-
-Returns the shape of the dual space of the space of objects of shape `shape`. By
-default, the `dual_shape` of a shape is itself. See the examples section below
-for an example for which this is not the case.
-
-## Examples
-
-Consider polynomial constraints for which the dual is moment constraints and
-moment constraints for which the dual is polynomial constraints. Shapes for
-polynomials can be defined as follows:
-```julia
-struct Polynomial
-    coefficients::Vector{Float64}
-    monomials::Vector{Monomial}
-end
-struct PolynomialShape <: AbstractShape
-    monomials::Vector{Monomial}
-end
-JuMP.reshape_result(x::Vector, shape::PolynomialShape) = Polynomial(x, shape.monomials)
-```
-and a shape for moments can be defined as follows:
-```julia
-struct Moments
-    coefficients::Vector{Float64}
-    monomials::Vector{Monomial}
-end
-struct MomentsShape <: AbstractShape
-    monomials::Vector{Monomial}
-end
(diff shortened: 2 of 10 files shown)


src/aff_expr.jl
```diff
@@ -337,7 +337,7 @@ function MOI.VectorAffineFunction(affs::Vector{AffExpr})
     MOI.VectorAffineFunction(terms, constant)
 end
 moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
-function moi_function_type(::Type{Vector{Aff}}) where {T, Aff <: GenericAffExpr{T}}
+function moi_function_type(::Type{<:Vector{<:GenericAffExpr{T}}}) where {T}
```
mlubin: Why is the `<:` needed before Vector?

src/constraints.jl
```diff
@@ -628,15 +545,66 @@ function _error_if_not_concrete_type(t)
...
+julia> @constraint(model, y in MOI.GreaterThan(1.0));
+
+julia> @constraint(model, y <= 1.0);
+
+julia> @constraint(model, 2x <= 1);
+
+julia> num_constraints(model, VariableRef, MOI.GreaterThan{Float64})
+2
+
+julia> all_constraints(model, VariableRef, MOI.ZeroOne)
```
mlubin: `num_constraints` here and below?

src/constraints.jl
```diff
@@ -628,15 +545,66 @@ function _error_if_not_concrete_type(t)
...
+    f_type = moi_function_type(function_type)
+    return MOI.get(model, MOI.NumberOfConstraints{f_type, set_type}())
+end
+
 
 """
-    all_constraints(model::Model, function_type, set_type)::Vector{VariableRef}
+    all_constraints(model::Model, function_type, set_type)::Vector{
+        ConstraintRef{Model,
+            MOI.ConstraintIndex{moi_function_type(function_type), set_type}}}
```
mlubin: It's pretty ugly to list this return type. Not sure what the right way to handle this is.

src/print.jl
```diff
@@ -386,11 +386,10 @@ end
 Write to `io` a summary of the number of constraints.
 """
 function show_constraints_summary(io::IO, model::Model)
-    for (F, S) in MOI.get(model, MOI.ListOfConstraints())
-        num_constraints = MOI.get(model, MOI.NumberOfConstraints{F, S}())
-        # TODO: Print jump_function_type(model, F) instead of F.
```
mlubin: Thanks for fixing this TODO :)

src/quad_expr.jl
```diff
@@ -295,7 +295,7 @@ function MOI.VectorQuadraticFunction(quads::Vector{QuadExpr})
     MOI.VectorQuadraticFunction(lin_terms, quadratic_terms, constants)
 end
 moi_function(a::Vector{<:GenericQuadExpr}) = MOI.VectorQuadraticFunction(a)
-function moi_function_type(::Type{Vector{Quad}}) where {T, Quad <: GenericQuadExpr{T}}
+function moi_function_type(::Type{<:Vector{<:GenericQuadExpr{T}}}) where {T}
```
mlubin: Same question about `<:` before Vector.

src/aff_expr.jl
blegat: Consider this example:
```julia
julia> f(::Type{Vector{<:Any}}) = 1

julia> f(Vector{Int})
ERROR: MethodError: no method matching f(::Type{Array{Int64,1}})
Closest candidates are:
  f(::Type{Array{T,1} where T}) at REPL[1]:1
Stacktrace:
 [1] top-level scope at none:0

julia> f(::Type{<:Vector{<:Any}}) = 1
f (generic function with 2 methods)

julia> f(Vector{Int})
1
```
The type `Vector{Int}` is the the same as the type `Vector{<:Any}` but it is a subtype of `Vector{<:Any}`.

src/constraints.jl
blegat: I wasn't sure either, @odow any idea ?

blegat: I have simplified it. Since the shape is different from vector to scalar, it is too complex to put in a signature

src/aff_expr.jl
odow: What was wrong  with
```julia
::Type{Vector{S}}) where {T, S <: GenericAffExpr{T}}
```
It is easier to understand than the two `<:`

(Although I now see this is used a lot.)

blegat: I find it confusing to create a type variable that is not used when we can avoid it. I can revert it if you don't share this feeling

odow: I'm not fussy. We probably need a style point on type signatures.

PR #1854, opened 2019-02-12
Fix corner cases for optimizer_index

blegat:
The PR also adds the `NoSolution` error which is thrown in `optimize!` if not optimizer is attached (instead of an AssertionError because the state is not `MOIU.ATTACHED_OPTIMIZER`)

### DIFF
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -125,7 +125,7 @@ function optimize!(model::Model,
         return model.optimize_hook(model)
     end
 
-    if MOIU.state(backend(model)) == MOIU.NO_OPTIMIZER
+    if model(model) != DIRECT && MOIU.state(backend(model)) == MOIU.NO_OPTIMIZER
         throw(NoOptimizer())
     end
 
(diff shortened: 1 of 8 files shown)


src/optimizer_interface.jl
```diff
@@ -67,15 +67,16 @@ function solve(::Model)
...
-
 """
     optimize!(model::Model,
               optimizer_factory::Union{Nothing, OptimizerFactory}=nothing;
               ignore_optimize_hook=(model.optimize_hook === nothing))
 
 Optimize the model. If `optimizer_factory` is not `nothing`, it first sets the
 optimizer to a new one created using the optimizer factory. The factory can be
-created using the [`with_optimizer`](@ref) function.
+created using the [`with_optimizer`](@ref) function. If `optimizer_factor` is
```
mlubin: s/optimizer_factor/optimizer_factory/

src/variables.jl
```diff
@@ -189,6 +189,11 @@ function Base.isequal(v1::VariableRef, v2::VariableRef)
     return owner_model(v1) === owner_model(v2) && v1.index == v2.index
 end
 
+"""
+    index(v::VariableRef)::MOI.VariableIndex
+
+Return the index of the variable `v` for the MOI backend.
```
mlubin: The index of the variable that corresponds to `v` in the MOI backend.

PR #1855, opened 2019-02-12
Minor documentation fixes

rtwalker:
Thanks for JuMP! 

I've been reading through the documentation over the last few weeks and collecting commits for correcting small typos, missing commas, formatting issues, etc., along the way. Now that I've made it to the end, it's time to share!  

I also noticed that there were a few inconsistencies in the examples with respect to preferring `x[j=1:2]` or  `x[j = 1:2]`. However I didn't see a note in the style guide, so I left them alone for now.

### DIFF
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -5,16 +5,22 @@ CurrentModule = JuMP
 Extending JuMP
 ==============
 
+```@meta
 TODO: How to extend JuMP: discussion on different ways to build on top of JuMP.
 How to extend JuMP's macros and how to avoid doing this.
+```
 
 ## Extending MOI
 
+```@meta
 TODO: Create new MOI function/sets, how to use it in JuMP
+```
 
 ### Adding a bridge
 
+```@meta
 TODO: create new bridge
+```
 
 See the [bridge section in the MOI manual](http://www.juliaopt.org/MathOptInterface.jl/v0.8/apimanual/#Constraint-bridges-1).
 
@@ -35,7 +41,9 @@ provide a more consistent interface to the user.
 
 ## Extending the `@variable` macro
 
+```@meta
 TODO: parse/build/add
+```
 
 ## Extending the `@constraint` macro
 
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -13,7 +13,7 @@ Extending JuMP
 ## Extending MOI
 
 ```@meta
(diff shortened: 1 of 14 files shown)


docs/src/extensions.md
```diff
@@ -5,16 +5,22 @@ CurrentModule = JuMP
 Extending JuMP
 ==============
 
+```@meta
 TODO: How to extend JuMP: discussion on different ways to build on top of JuMP.
```
mlubin: `@meta` introduces a code block, so you need a `#` to write text comments inside. This is likely causing the build failures: https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/492287976#L650.

docs/src/variables.md
```diff
@@ -42,7 +42,7 @@ This code does three things:
 2. it creates two *JuMP* variables that act as references to those optimization
    variables
 3. it binds those JuMP variables as a vector with two elements to the *Julia*
```
mlubin: The formatting of this list is inconsistent with the one below. The suggestion at https://developers.google.com/style/lists is to use complete sentences with proper capitalization for all list items. Could you make this change here also?

mlubin:
Also, good point about inconsistent use of `x[j=1:2]` or `x[j = 1:2]`. According to the [style guide](http://www.juliaopt.org/JuMP.jl/dev/style/#Whitespace-1), I would view `=` as a binary operator, so there should be spaces around it.

docs/src/extensions.md
rtwalker: Thanks for pointing that out

docs/src/variables.md
rtwalker: Done!

rtwalker:
Thanks for the feedback! I'll try to find the places where `x[j=1:2]` style indexing is used and correct those as well.

docs/src/extensions.md
```diff
@@ -5,16 +5,22 @@ CurrentModule = JuMP
...
-How to extend JuMP's macros and how to avoid doing this.
+```@meta
+# TODO: How to extend JuMP: discussion on different ways to build on top of JuMP.
+# How to extend JuMP's macros and how to avoid doing this.
+```
 
 ## Extending MOI
 
+```@meta
 TODO: Create new MOI function/sets, how to use it in JuMP
```
mlubin: Missing comment. This is causing the build to fail.

PR #1856, opened 2019-02-13
Add SeDuMi

blegat:


blegat:
Yes it's better to redirect to this one

PR #1857, opened 2019-02-13
Improve printing of SDP constraints

blegat:
Currently all SDP constraints are displayed in their vectorized form `[...] in MOI.PositiveSemidefiniteCone...` even if in the `shape` of the constraint, we have the information of whether it was constructed in this vectorized form or in matrix form.
In this PR, the function and set of a constraint are reshaped before being passed to `function_string` and `in_set_string` (hence the need to rename `reshape_result` into `reshape_vector` since the function is not a result and the need to create `reshape_set`).
For PSD constraint, in case the constraint was created in matrix form, the vector is reshaped to a matrix and the set is replaced by `PSDCone()`.
Finally, the PR implements the printing of matrices similarly to how it was done in JuMP v0.18 (except I couldn't find how it was done for the IJuliaMode). The `constraints_string` function had to be tweaked to work with multiline constraint string.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -400,7 +400,7 @@ end
 """
     constraints_string(print_mode, model::AbstractModel)::Vector{String}
 
-Return a list of `String`s describing each constraints of the model.
+Return a list of `String`s describing each constraint of the model.
 """
 function constraints_string(print_mode, model::Model)
     strings = String[]
--- a/src/print.jl
+++ b/src/print.jl
@@ -220,7 +220,17 @@ function model_string(print_mode, model::AbstractModel)
     end
     str *= eol
     str *= ijl ? "\\text{Subject to} \\quad" : "Subject to" * eol
-    str *= constraints_string(print_mode, model, sep, eol)
+    constraints = constraints_string(print_mode, model)
+    if print_mode == REPLMode
+        constraints = map(str -> replace(str, '\n' => eol * sep), constraints)
+    end
+    if !isempty(constraints)
+        str *= sep
+    end
+    str *= join(constraints, eol * sep)
+    if !isempty(constraints)
+        str *= eol
+    end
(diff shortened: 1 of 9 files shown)


src/print.jl
```diff
@@ -398,26 +398,25 @@ function show_constraints_summary(io::IO, model::Model)
 end
 
 """
-    constraints_string(print_mode, model::AbstractModel, sep, eol)::String
+    constraints_string(print_mode, model::AbstractModel)::Vector{String}
 
-Return a `String` describing the constraints of the model, each on a line
-starting with `sep` and ending with `eol` (which already contains `\n`).
+Return a list of `String`s describing each constraints of the model.
```
mlubin: Nit: each constraint

src/sd.jl
```diff
@@ -24,17 +24,38 @@ julia> a = [ x 2x
 julia> b = [1 2
             2 4];
 
-julia> @SDconstraint(model, a ⪰ b)
-[x - 1, 2 x - 2, 2 x - 2, x - 4] ∈ MathOptInterface.PositiveSemidefiniteConeSquare(2)
+julia> cref = @SDconstraint(model, a ⪰ b)
+ x - 1    2 x - 2
+ 2 x - 2  x - 4   ∈ PSDCone()
```
mlubin: I'd argue for a format like:
```
[x - 1,    2 x - 2;
 2 x - 2,  x - 4 ]  ∈ PSDCone()
```
This is more immediately understandable because it uses existing Julia syntax. I had to spend a second to parse the output format above.

blegat: We rely on Julia printing of matrices ro leverage its alignment of columns and its use of dots for too long matrices.
That makes it difficult to dectect change of columns so it's hard to add the commas

PR #1866, opened 2019-02-15
⚡️ Avoid creating temporary expression in matrix multiplications

blegat:
With the benchmark in https://github.com/JuliaOpt/JuMP.jl/issues/1403#issuecomment-463857803:
Before:
```
julia> @time x*A;
0.559247 seconds (8.21 M allocations: 668.261 MiB, 40.98% gc time)

julia> @time A*x*A;
33.784054 seconds (14.60 M allocations: 9.227 GiB, 51.83% gc time)
```
After:
```
julia> @time x*A;
0.153061 seconds (210.03 k allocations: 73.168 MiB)

julia> @time A*x*A;
16.037227 seconds (630.06 k allocations: 6.126 GiB, 14.55% gc time)
```

### DIFF
--- a/src/quad_expr.jl
+++ b/src/quad_expr.jl
@@ -126,55 +126,64 @@ function Base.eltype(qti::QuadTermIterator{GenericQuadExpr{C, V}}
     return Tuple{C, V, V}
 end
 
-# With one factor
+# With one factor.
 
 function add_to_expression!(quad::GenericQuadExpr{C}, other::C) where C
     return add_to_expression!(quad.aff, other)
 end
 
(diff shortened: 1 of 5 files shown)


src/quad_expr.jl
```diff
@@ -151,8 +143,110 @@ function add_to_expression!(q::GenericQuadExpr{T,S}, other::GenericQuadExpr{T,S}
...
+            add_to_expression!(quad.aff, c*lhscoef, lhsvar)
+        end
+    end
+
+    quad.aff.constant += lhs.constant * rhs.constant
+
+    return quad
+end
+
+# With three factor
```
mlubin: Typo: factors. Missing periods on all the comments like these.

mlubin:
Could we add this case to `test/perf`?

PR #1872, opened 2019-02-18
Added type info in NL error message

coreysharris:
Type checking for parsing nonlinear expression was missng the GenericAffExpr case.  Also added type information to the catch-all error.

### DIFF
--- a/src/parse_nlp.jl
+++ b/src/parse_nlp.jl
@@ -199,13 +199,13 @@ end
 function _parse_NL_expr_runtime(m::Model, x::GenericQuadExpr, tape, parent, values)
     error("Unexpected quadratic expression $x in nonlinear expression. " *
           "Quadratic expressions (e.g., created using @expression) and " *
-          "nonlinear expression cannot be mixed.")
+          "nonlinear expressions cannot be mixed.")
 end
 
 function _parse_NL_expr_runtime(m::Model, x::GenericAffExpr, tape, parent, values)
     error("Unexpected affine expression $x in nonlinear expression. " *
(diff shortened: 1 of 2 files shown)


src/parse_nlp.jl
```diff
@@ -202,8 +202,14 @@ function _parse_NL_expr_runtime(m::Model, x::GenericQuadExpr, tape, parent, valu
           "nonlinear expression cannot be mixed.")
 end
 
+function _parse_NL_expr_runtime(m::Model, x::GenericAffExpr, tape, parent, values)
+    error("Unexpected affine expression $x in nonlinear expression. " *
+          "Affine expressions (e.g., created using @expression) and " *
+          "nonlinear expression cannot be mixed.")
```
odow: Nit: nonlinear expression*s*

coreysharris: Agreed... fixed.

test/nlp.jl
```diff
@@ -619,4 +619,30 @@
...
+        model = Model()
+        @variable(model, x)
+        @variable(model, y)
+        A = x + y
+        expected_exception = ErrorException(
+            "Unexpected affine expression x + y in nonlinear expression. " *
+            "Affine expressions (e.g., created using @expression) and " *
+            "nonlinear expressions cannot be mixed."
+        )
+        @test_throws expected_exception @NLexpression(m, A)
```
mlubin: `m` should be `model` (also below)

coreysharris: Fixed in both places

mlubin:
Thanks!

PR #1878, opened 2019-02-25
remove localvar hack

mlubin:
The desired behavior of loop variables not leaking outside their scope is now the default in Julia.


mlubin:
Added a test. I thought there was one but maybe it got dropped in the transition to 0.19.

PR #1886, opened 2019-03-02
Add point in style guide for naming models

### LINKED ISSUE #1885: Style for naming model
We should mention in the style guide that the model should be named `model` instead of `m` as the variable name is more representative with `model`.

barpit20:
Fixes #1885

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -284,6 +284,10 @@ some_local_variable = ...
 some_file.jl # Except for ModuleName.jl.
 ```
 
+#### Don’t use m for model
+
+Models should be named `model` instead of `m` as the variable name is more representative with `model`.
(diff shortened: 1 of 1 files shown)


mlubin:
Cannot be merged as is, see https://github.com/JuliaOpt/JuMP.jl/issues/1885#issuecomment-468942470. I guess we need a style guide on how to write the style guide also. Always give a reason for a rule instead of saying "don't do X".

barpit20:
Sorry, I read the comment after sending the PR. Should I wait for it to be decided what to add with regards to meaningful names or close this PR?

mlubin:
Deciding what to say is the hard part, so we can close this PR until we have a conclusion on that.

PR #1900, opened 2019-03-13
Creation of ExprTerms alias and cleanup of interface

ndinsmore:
This is a simple change that cleans up the definition of the collection used to store expression terms

It creates a type alias `ExprTerms` that can be used where `OrderDict` was previously used.

This will enable easy changes to the storage of expression terms in the future.

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -15,45 +15,38 @@
 # Operator overloads in src/operators.jl
 #############################################################################
 
+const ExprTerms{K,V} = OrderedDict{K,V} where {K, V}
+
 # Utilities for OrderedDict
-function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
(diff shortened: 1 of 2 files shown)


mlubin:
Could you separate the improvements (e.g., saving an extra lookup) from the alias change?

I'm not convinced that we need an alias. It's not really so hard to experiment with a new type that we need to commit these changes to the code in anticipation of experimenting.

ndinsmore:
Beyond just the ease of experimentation, it adds to the clarity of the code.  

I think it also makes it clear that the terms in both the affine and quadratic versions are using the same type.
Furthermore, I don't know how JuMP extensions have done it in the past but this seems like a good practice so that have access to the same machinery.

mlubin:
Closing as the change is stale. Please discuss with us at https://gitter.im/JuliaOpt/JuMP-dev if you want to revive this change.

PR #1901, opened 2019-03-13
Improvement to _add_or_set

ndinsmore:
per @mlubin request in #1900

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -16,12 +16,13 @@
 #############################################################################
 
 # Utilities for OrderedDict
-function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
-    # TODO: This unnecessarily requires two lookups for k.
+function _add_or_set!(exprterms::OrderedDict{K,V}, k::K, v::V) where {K,V}
     # TODO: Decide if we want to drop zeros here after understanding the
     # performance implications.
-    dict[k] = get!(dict, k, zero(V)) + v
-    return dict
+    v0 = get!(exprterms, k, v)
+    # This short circuit makes sure only one hash lookup is need when inserting a value
+    v0 == v || (exprterms[k] = v0 + v)
+    return exprterms
 end
 
 function _new_ordered_dict(::Type{K}, ::Type{V}, kv::AbstractArray{<:Pair}) where {K,V}
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -16,13 +16,16 @@
 #############################################################################
 
 # Utilities for OrderedDict
-function _add_or_set!(exprterms::OrderedDict{K,V}, k::K, v::V) where {K,V}
+function _add_or_set!(expr_terms::OrderedDict{K,V}, k::K, v::V) where {K,V}
     # TODO: Decide if we want to drop zeros here after understanding the
     # performance implications.
-    v0 = get!(exprterms, k, v)
-    # This short circuit makes sure only one hash lookup is need when inserting a value
-    v0 == v || (exprterms[k] = v0 + v)
-    return exprterms
+    v0 = get!(expr_terms, k, v)
+    # This makes sure only one hash lookup is need when inserting a value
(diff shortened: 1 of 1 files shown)


src/aff_expr.jl
```diff
@@ -16,12 +16,13 @@
 #############################################################################
 
 # Utilities for OrderedDict
-function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
-    # TODO: This unnecessarily requires two lookups for k.
+function _add_or_set!(exprterms::OrderedDict{K,V}, k::K, v::V) where {K,V}
```
mlubin: Either `expr_terms` (http://www.juliaopt.org/JuMP.jl/v0.19.0/style/#Use-of-underscores-within-names-1) or leave as `dict` .

src/aff_expr.jl
```diff
@@ -16,12 +16,13 @@
...
     # TODO: Decide if we want to drop zeros here after understanding the
     # performance implications.
-    dict[k] = get!(dict, k, zero(V)) + v
-    return dict
+    v0 = get!(exprterms, k, v)
+    # This short circuit makes sure only one hash lookup is need when inserting a value
+    v0 == v || (exprterms[k] = v0 + v)
```
mlubin: IMO the following is easier to read than the short circuit:
```julia
if v0 !== v
    expr_terms[k] = v0 + v
end
```
(`===`/`!==` can be slightly faster than `==`/`!=` in this case, and is safe to use because the documentation for `get!` says that `v` itself will be returned if there's no key defined.)

src/aff_expr.jl
```diff
@@ -16,12 +16,13 @@
...
 # Utilities for OrderedDict
-function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
-    # TODO: This unnecessarily requires two lookups for k.
+function _add_or_set!(exprterms::OrderedDict{K,V}, k::K, v::V) where {K,V}
     # TODO: Decide if we want to drop zeros here after understanding the
     # performance implications.
-    dict[k] = get!(dict, k, zero(V)) + v
-    return dict
+    v0 = get!(exprterms, k, v)
+    # This short circuit makes sure only one hash lookup is need when inserting a value
```
mlubin: Nit: missing period at the end of the comment (http://www.juliaopt.org/JuMP.jl/v0.19.0/style/#Comments-1).

mlubin:
I can't reproduce the CI failure either on JuMP `master` or with this PR patched in. Not sure what the issue is.

src/aff_expr.jl
```diff
@@ -16,12 +16,16 @@
...
-function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
-    # TODO: This unnecessarily requires two lookups for k.
+function _add_or_set!(expr_terms::OrderedDict{K,V}, k::K, v::V) where {K,V}
     # TODO: Decide if we want to drop zeros here after understanding the
     # performance implications.
-    dict[k] = get!(dict, k, zero(V)) + v
-    return dict
+    v0 = get!(expr_terms, k, v)
+    # This makes sure only one hash lookup is need when inserting a value.
+    if v0 !== v
```
odow: This doesn't work when `expr_terms[k]` is equal to `v`.

mlubin: Good catch. Glad we have tests :)

odow: So it isn't an unnecessary double look-up. You need one to get the current value (if it exists), and the second to set the new value. I don't see a way to get around this.

mlubin: It's not necessary to compute the hash of `k` twice, and if `k` is already a valid key, it's not necessary to look up where its value is stored twice. Neither `Dict` nor `OrderedDict` provide an API to do this, however.

odow: I guess I should have clarified "contingent on the current API."

ndinsmore:
Per odow, currently this is incorrect

Issue #1899, opened 2019-03-13
Refactor sum


blegat:
There is currently three sum methods:
https://github.com/JuliaOpt/JuMP.jl/blob/c01802926c34aa30e5d272ac1063f1595084aa4a/src/operators.jl#L254-L269
The three methods interact the same way with the arrays, they just iterate over it, so the first method should be defined on `AbstractArray`.
Then it has the same signature than the first method so we can just keep the fastest one.
Lastly, the third method could also work for `GenericQuadExpr` so we could widen the type.

Peiffap:
Did #1902 solve this issue?

blegat:
It did !
Closed by #1902

PR #1902, opened 2019-03-13
generalized sum for GenericQuadExpr

guilhermebodin:
I'm trying to help with [#1899](https://github.com/JuliaOpt/JuMP.jl/issues/1899). One thing I didn't understand is how the first `sum` method should be defined on `AbstractVector`.

I tried `Base.sum(vars::AbstractVector{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])` but this is not the same thing as `Base.sum(vars::Array{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])`. 

```julia
julia> Array <: AbstractVector
false
```

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -251,7 +251,13 @@ _sizehint_expr!(q, n) = nothing
 #############################################################################
 
 # TODO: specialize sum for DenseAxisArray and SparseAxisArray of JuMP objects?
-Base.sum(vars::AbstractArray{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])
+function Base.sum(array::AbstractArray{<:AbstractVariableRef}) 
+    result_expression = zero(eltype(array)) 
+    for variable in array 
+        add_to_expression!(result_expression, variable) 
+    end 
+    return result_expression 
+end 
 # TODO: Specialize for iterables.
 function Base.sum(affs::AbstractArray{T}) where {T <: AbstractJuMPScalar}
     new_aff = zero(T)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -252,13 +252,6 @@ _sizehint_expr!(q, n) = nothing
 
 # TODO: specialize sum for DenseAxisArray and SparseAxisArray of JuMP objects?
 Base.sum(vars::AbstractArray{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])
-function Base.sum(array::AbstractArray{<:AbstractVariableRef})
-    result_expression = zero(eltype(array))
-    for variable in array
-        add_to_expression!(result_expression, variable)
-    end
-    return result_expression
-end
 # TODO: Specialize for iterables.
 function Base.sum(affs::AbstractArray{T}) where {T <: AbstractJuMPScalar}
     new_aff = zero(T)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -251,7 +251,7 @@ _sizehint_expr!(q, n) = nothing
 #############################################################################
 
 # TODO: specialize sum for DenseAxisArray and SparseAxisArray of JuMP objects?
-Base.sum(vars::Array{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])
+Base.sum(vars::AbstractArray{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])
 function Base.sum(array::AbstractArray{<:AbstractVariableRef})
     result_expression = zero(eltype(array))
     for variable in array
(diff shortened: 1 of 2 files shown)


blegat:
@guilhermebodin You are right, I meant `AbstractArray`, I have corrected the issue

blegat:
The other change can be done in a separate PR, this one is self-contained

src/operators.jl
```diff
@@ -251,7 +251,7 @@ _sizehint_expr!(q, n) = nothing
 #############################################################################
 
 # TODO: specialize sum for DenseAxisArray and SparseAxisArray of JuMP objects?
-Base.sum(vars::Array{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])
+Base.sum(vars::AbstractArray{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])
 function Base.sum(array::AbstractArray{<:AbstractVariableRef})
```
blegat: Now this one can be removed

blegat:
See CI failure

src/operators.jl
guilhermebodin: Removing this sum method is failing because creating this kind of array `[v => 1.0 for v in x]` 
in a `JuMPDict{VariableRef}`
```julia
m = Model()
@variable(m, x[1:3])
[v => 1.0 for v in x]
S = [1, 3]
@variable(m, y[S])
[v => 1.0 for v in y] #cannot create the array
[v => 1.0 for v in y.data] #creates what we want
```
Maybe we could leave this method and add more specialized ones, something like 
```julia
Base.sum(vars::Array{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars]) 
Base.sum(vars::Containers.DenseAxisArray{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars.data])
function Base.sum(array::AbstractArray{<:AbstractVariableRef}) 
     result_expression = zero(eltype(array)) 
     for variable in array 
         add_to_expression!(result_expression, variable) 
     end 
     return result_expression 
 end 
```

guilhermebodin: The other option would be 
```julia
function Base.sum(array::AbstractArray{<:AbstractVariableRef}) 
     result_expression = zero(eltype(array)) 
     for variable in array 
         add_to_expression!(result_expression, variable) 
     end 
     return result_expression 
 end 
```
and rely that the `add_to_expression!` function is well specialized.

guilhermebodin:
Removing this sum method is failing because creating this kind of array `[v => 1.0 for v in x]` in a `JuMPDict{VariableRef}`
```julia
m = Model()
@variable(m, x[1:3])
[v => 1.0 for v in x]
S = [1, 3]
@variable(m, y[S])
[v => 1.0 for v in y] #cannot create the array
[v => 1.0 for v in y.data] #creates what we want
```
Maybe we could leave this method and add more specialized ones, something like 
```julia
Base.sum(vars::Array{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars]) 
Base.sum(vars::Containers.DenseAxisArray{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars.data])
function Base.sum(array::AbstractArray{<:AbstractVariableRef}) 
     result_expression = zero(eltype(array)) 
     for variable in array 
         add_to_expression!(result_expression, variable) 
     end 
     return result_expression 
 end 
```

The other option would be 
```julia
function Base.sum(array::AbstractArray{<:AbstractVariableRef}) 
     result_expression = zero(eltype(array)) 
     for variable in array 
         add_to_expression!(result_expression, variable) 
     end 
     return result_expression 
 end 
```
and rely that the `add_to_expression!` function is well specialized.

blegat:
We could leave only the second one. In fact, because now the list of terms is represented with a dictionary and not an array, creating an intermediate array is not a good idea.
In fact, it turns out that the second one is more efficient even for a vector of variables:
```julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, x[1:1000000]);

julia> function a(array::AbstractArray{<:AbstractVariableRef})        
           result_expression = zero(eltype(array))        
           for variable in array        
               add_to_expression!(result_expression, variable)        
           end        
           return result_expression        
       end
a (generic function with 1 method)

julia> b(vars::Array{<:AbstractVariableRef}) = GenericAffExpr(0.0, [v => 1.0 for v in vars])
b (generic function with 1 method)

julia> @time f = a(x);
  0.575067 seconds (328.02 k allocations: 49.837 MiB)

julia> @time f = a(x);
  0.441960 seconds (64 allocations: 33.336 MiB)

julia> @time f = a(x);
  0.449861 seconds (64 allocations: 33.336 MiB)

julia> @time f = a(x);
  0.417298 seconds (64 allocations: 33.336 MiB)

julia> @time f = b(x);
  0.811685 seconds (1.14 M allocations: 86.536 MiB, 36.19% gc time)

julia> @time f = b(x);
  0.451770 seconds (1.00 M allocations: 79.483 MiB)

julia> @time f = b(x);
  0.465811 seconds (1.00 M allocations: 79.483 MiB)
```

PR #1904, opened 2019-03-14
Improve error on getting duals from variables

### LINKED ISSUE #1903: Improve error on getting duals from variables
Currently this is a no method error.
It could return a proper error saying that the user should refer to `UpperBoundRef`, `LowerBoundRef` or `FixRef` to get reduced the dual from these constraints.

Example that returns the error:
```julia
m = Model(with_optimizer(SOLVER))
@variable(m, x == 0)
optimize!(m)
dual(x)
```
The error:
```julia
ERROR: MethodError: no method matching dual(::VariableRef)
Closest candidates are:
  dual(::ConstraintRef{Model,NonlinearConstraintIndex,Shape} where Shape<:AbstractShape) at C:\Users\joaquimgarcia\.julia\packages\JuMP\jnmGG\src\nlp.jl:162
  dual(::ConstraintRef{Model,#s1044,Shape} where Shape<:AbstractShape where #s1044<:MathOptInterface.ConstraintIndex) at C:\Users\joaquimgarcia\.julia\packages\JuMP\jnmGG\src\constraints.jl:480
Stacktrace:
 [1] top-level scope at none:0
```

barpit20:
Fixes #1903

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -790,3 +790,10 @@ function all_variables(model::Model)
         model, MOI.ListOfVariableIndices())::Vector{MOI.VariableIndex}
     return VariableRef[VariableRef(model, idx) for idx in all_indices]
 end
+
+function dual(vref::VariableRef)
+    error("To query the dual variables associated with a variable bound, first " *
+          "obtain a constraint reference using one of `UpperBoundRef`, `LowerBoundRef`, " *
+          "or `FixRef`, and then call `dual` on the returned constraint reference.\nFor " *
+          "example, if `x <= 1`, instead of `dual(x)`, call `dual(UpperBoundRef(x))`.")
+end
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -191,7 +191,7 @@ julia> shadow_price(con)
 2.0
 ```
 
-To query the dual variables associated a variable bound, first obtain a
+To query the dual variables associated with a variable bound, first obtain a
 constraint reference using one of [`UpperBoundRef`](@ref),
 [`LowerBoundRef`](@ref), or [`FixRef`](@ref), and then call [`dual`](@ref) on
 the returned constraint reference. Note that in linear programming,
--- a/test/variable.jl
+++ b/test/variable.jl
@@ -599,3 +599,14 @@ end
 @testset "Variables for JuMPExtension.MyModel" begin
     variables_test(JuMPExtension.MyModel, JuMPExtension.MyVariableRef)
 end
+
+@testset "Dual from Variable" begin
+    model = Model()
+    @variable(model, x == 0)
+    exception = ErrorException(
+        "To query the dual variables associated with a variable bound, first " *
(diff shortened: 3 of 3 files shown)


src/variables.jl
```diff
@@ -790,3 +790,15 @@ function all_variables(model::Model)
         model, MOI.ListOfVariableIndices())::Vector{MOI.VariableIndex}
     return VariableRef[VariableRef(model, idx) for idx in all_indices]
 end
+
+"""
+    dual(vref::VariableRef)
+
+Returns a proper error saying that the user should refer to UpperBoundRef,
+LowerBoundRef or FixRef to get the reduced dual from a constraint.
+"""
```
odow: This function doesn't need a doctstring since it just implements a helpful error fallback.

src/variables.jl
```diff
@@ -790,3 +790,15 @@ function all_variables(model::Model)
...
+function dual(vref::VariableRef)
+    error("To query the dual variables associated with a variable bound, first " *
+    "obtain a constraint reference using one of UpperBoundRef, LowerBoundRef, " *
+    "or FixRef, and then call dual on the returned constraint reference.")
```
odow: These lines should be indented to line up with the `"` on the first line. You could also add an example:
```
For example, if `x <= 1`, instead of `dual(x)`, call `dual(UpperBoundRef(x))`.
```

odow:
You should also add a test to https://github.com/JuliaOpt/JuMP.jl/blob/master/test/variable.jl

Here is an example for how to test exceptions:
https://github.com/JuliaOpt/JuMP.jl/blob/c01802926c34aa30e5d272ac1063f1595084aa4a/test/macros.jl#L367-L370

barpit20:
Updated with the required changes. I also think that I noticed a grammatical mistake in the docs while working on this issue.
Screenshot from 2019-03-14 23-11-54

I think this line should be - To query the dual variables associated **with** a variable bound, first obtain ...
Should I add a commit to fix this as well?

blegat:
> Should I add a commit to fix this as well?

Yes, seems like a typo indeed

src/variables.jl
```diff
@@ -790,3 +790,10 @@ function all_variables(model::Model)
         model, MOI.ListOfVariableIndices())::Vector{MOI.VariableIndex}
     return VariableRef[VariableRef(model, idx) for idx in all_indices]
 end
+
+function dual(vref::VariableRef)
+    error("To query the dual variables associated with a variable bound, first " *
+          "obtain a constraint reference using one of UpperBoundRef, LowerBoundRef, " *
```
blegat: Put backsticks around `UpperBoundRef`, ...

src/variables.jl
```diff
@@ -790,3 +790,10 @@ function all_variables(model::Model)
...
+          "or FixRef, and then call dual on the returned constraint reference.\nFor " *
```
blegat: Backsticks around `dual`

barpit20:
Updated :smiley:

PR #1909, opened 2019-03-18
Fix import expression for Julia 1.2

iamed2:
Fixes https://github.com/JuliaOpt/JuMP.jl/issues/1908

This alternate import syntax appears to be valid on Julia 1.0+


iamed2:
Ah, I missed that `release-0.18` still runs on 0.6. I'll fix.

Issue #1907, opened 2019-03-18
writeLP and writeMPS lost with JuMP 0.19


dourouc05:
I'm a long-time user of the `writeLP` function to ease model debugging (especially when a problem happens deep inside a loop, where getting access to the model from the REPL is not always easy — and it is so convenient to load a LP file in CPLEX interactive or Gurobi shell!). However, when migrating to JuMP 0.19, this function (along with `writeMPS`) seems to have disappeared. I could not see any trace of it in the documentation (nor a related title), in NEWS, in MathOptInterface, in LinQuadOptInterface, so I feel a bit lost… Has this functionality been deprecated or moved to another package? I guess it should be documented (`@deprecate`).

odow:
See https://discourse.julialang.org/t/writemps/21669/4

https://github.com/odow/MathOptFormat.jl

joaquimg:
Should we move to juliOpt, tag and add to news?

odow:
Probably. It's a good candidate to join the new JuMP-dev organisation. I also want to rename it to `MathOptFormats.jl`.

dourouc05:
What about providing an error message for the old functions that point to this new package? 

```
writeLP(x…) = error("writeLP is no more provided by JuMP; instead, you can check MathOptFormat.jl")
writeMPS(x…) = error("writeMPS is no more provided by JuMP; instead, you can check MathOptFormat.jl")
```

odow:
Sure, we would accept a PR for this. If MathOptFormats.jl gets tagged, we will probably re-introduce this functionality to JuMP via a package dependency.

mlubin:
Fixed by #1910

PR #1910, opened 2019-03-18
Deprecation message for writeLP and writeMPS

dourouc05:
Following #1907


src/print.jl
```diff
@@ -638,3 +638,9 @@ end
 function Base.show(io::IO, evaluator::NLPEvaluator)
     Base.show(io, "A JuMP.NLPEvaluator")
 end
+
+#------------------------------------------------------------------------
+## Deprecated functions to export a model
+#------------------------------------------------------------------------
+writeLP(_...) = error("writeLP is no more provided by JuMP; instead, you can check https://github.com/odow/MathOptFormat.jl")
+writeMPS(_...) = error("writeMPS is no more provided by JuMP; instead, you can check https://github.com/odow/MathOptFormat.jl")
```
odow: These should probably go after these: 
https://github.com/JuliaOpt/JuMP.jl/blob/c01802926c34aa30e5d272ac1063f1595084aa4a/src/JuMP.jl#L29-L37
No need for the comments.

Also, please make the functions `writeLP(args...; kwargs...)` so that it catches any usage.

As an English note, `is no more provided` isn't great. How about something like
```julia
function writeLP(args...; kargs...)
    error("writeLP has been removed from JuMP. Use `MathOptFormat.jl` instead.")
end
```

odow:
Great! I added a new line to `print.jl`

PR #1912, opened 2019-03-19
Promote eltype of SOS sets to Float64

### LINKED ISSUE #1911: Promote eltype of MOI sets like SOS to Float64
As raised here: https://discourse.julialang.org/t/sos1-variables-are-not-supported/22012

```julia
model = Model()
@variable(model, x[1:3])
@constraint(model, x in MOI.SOS1(1:3))  # should create `SOS1{Float64}`
@constraint(model, x in MOI.SOS2([4, 5, 6])) # should create `SOS2{Float64}`
```

odow:
Closes #1911

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -519,12 +519,12 @@ function build_constraint(_error::Function, expr, lb, ub)
 end
 
 function build_constraint(
-        error_func::Function, x::Vector{<:AbstractJuMPScalar}, set::MOI.SOS1)
+        ::Function, x::Vector{<:AbstractJuMPScalar}, set::MOI.SOS1)
     return VectorConstraint(x, MOI.SOS1{Float64}(set.weights))
 end
 
 function build_constraint(
-        error_func::Function, x::Vector{<:AbstractJuMPScalar}, set::MOI.SOS2)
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -518,6 +518,16 @@ function build_constraint(_error::Function, expr, lb, ub)
     end
 end
 
+function build_constraint(
+        error_func::Function, x::Vector{<:AbstractJuMPScalar}, set::MOI.SOS1)
```
blegat: `error_func` should be `_error` for consistency.

blegat: If there is a good argument for `error_func`, we should change everything to `error_func`

odow: The argument is ["never begin the name of a local variable with an underscore"](https://www.juliaopt.org/JuMP.jl/latest/style/#Exported-and-non-exported-names-1).

blegat: Ok, then I suggest making a decision on how to name this argument and stick with it because it will affect many packages (PolyJuMP, SumOfSquares, SetProg, ...). `error_func` looks good to me.

blegat: Note that for these two methods, it's probably better not to give any name since it is not used

Issue #1892, opened 2019-03-09
variable lower/upper bound inconsistency


EdvinAblad:
I am under the impression that a single variable has at most one lower and upper bound.

Hence, the code
```julia
m = Model(with_optimizer(Clp.Optimizer))
@variable(m, x)
@constraint(m, x in MOI.LessThan(7.0))
@constraint(m, x in MOI.LessThan(5.0))
@constraint(m, x in MOI.LessThan(10.0))
@objective(m, Max, x)
optimize!(m)
```

Is a little weird, and should probably warn the user about its behaviour. 
From the results (Clp and GLPK tested), it is only the last variable constraint that is used (seems like a sane choice).

1. `has_upper_bound(x)`  returns `false`, (It only seems to concern the bound introduced in `@variable` and not abound `@constraint` bounds)
2. `all_constraints(m, VariableRef, MOI.LessThan{Float64})` returns all three constraints
3. `value(x) ==10` returns `true` 

This is related to #1586, where it seems that `@constraint(m, x <= b) ` always introduces an affine constraint rather than a single variable constraint. Maybe this is the expected behaviour? 

At least this inconsistency is clearly a bug.

mlubin:
I believe this issue will be resolved by https://github.com/JuliaOpt/MathOptInterface.jl/issues/570, which would enforce that at most one variable bound can exist at a time.

EdvinAblad:
Sorry missed that ticket. 
I believe so too.

holy-motors:
I think the issue is resolved. Let's close this.

mlubin:
@holy-motors, agreed.

```julia
julia> m = Model();

julia> @variable(m, x)
x

julia> @constraint(m, x in MOI.LessThan(7.0))
x ≤ 7.0

julia> @constraint(m, x in MOI.LessThan(5.0))
ERROR: MathOptInterface.UpperBoundAlreadySet{MathOptInterface.LessThan{Float64},MathOptInterface.LessThan{Float64}}: Cannot add `SingleVariable`-in`MathOptInterface.LessThan{Float64}` constraint for variable MathOptInterface.VariableIndex(1) as a `SingleVariable`-in`MathOptInterface.LessThan{Float64}` constraint was already set for this variable and both constraints set an upper bound.
```

Issue #1890, opened 2019-03-08
Modify RHS (and define RHS)


joaquimg:
In #1183 we decided to not implement: `ScalarConstantChange`.
Probably following: #470 
According to the manual (http://www.juliaopt.org/JuMP.jl/v0.19.0/constraints/#Modifying-a-constant-term-1):
In this constraint: `@constraint(model, 2x + 1 <= x - 3)`, why the coefficient of `x` is well defined and the constant of the function is not?

Moreover, in #1332 we are considering bounds on RHS, but we have not defined RHS.

Maybe we should reconsider?
At least bring back `add_constant` (or `add_to_rhs`)? https://github.com/JuliaOpt/JuMP.jl/pull/1351#discussion_r224945107

blegat:
It seems to me that `add_constant` is clearer. You add a constant to the the function. For vector constraints, this is what will be done. For scalar constraints, it will substract this constant from the set. From the user perspective, he needs to understand that `add_constant` is defined based on the `@constraint(model, func in set)` syntax so he needs to understand that `@constraint(model, a <= b)` is translated to `@constraint(model, a - b in MOI.LessThan(0.0))`.

odow:
I'm not a fan of `add_to_rhs` without a nice way to query the current (standard-form) rhs.

How about `set_standard_form_rhs` and `set_standard_form_coefficient`. That makes it pretty clear something weird is going on, and hopefully people go read the documentation.

joaquimg:
The functionality is essencial.
I think is reasonable to have a verbose name to highlight some possibly confusing behavior.

blegat:
> without a nice way to query the current (standard-form) rhs.

What prevents us from defining `constant(::ConstraintRef)` is interval constraints. But `rhs` wont work for interval constraint either so we could have
* a `constant` function that works with every non-interval constraint;
* a `add_constant` function that works even with interval constraints and is as described in https://github.com/JuliaOpt/JuMP.jl/issues/1890#issuecomment-470877385 and
* a `set_function_constant` that only works for non-interval constraints and sets the function constant where the set constant is zero.

`set_standard_from_rhs(cref, a)` is equivalent to `set_function_constant(cref, -a)`.

DillonJ:
For what it's worth, I support the addition of _any_ functionality that allows you to modify the constant term of a constraint that doesn't involve adding dummy variables to the problem.

joaquimg:
[...]

Issue #1332, opened 2018-06-09
Feature Request: sensitivity summary for LPs


lkapelevich:
It would be convenient to have an interface to get bounds on the RHS / objective vector where the current duals/reduced costs are valid.

mlubin:
This is a very consistent request for people using JuMP for teaching, so let's just implement it (like we did for `shadow_price`).

CC @EdvinAblad

EdvinAblad:
Agree,  this should not be too cumbersome to do.
That is if we assume that ConstraintBasisStatus is implemented by the solver. (JuliaOpt/MathOptInterface#655)

However, if ConstraintBasisStatus is unsupported/not implemented by the solver how should this be handled?
1. warn/throw error
2. implement a fallback (easy for non-degenerate solutions, brute-force(?) for degenerate ones)

I vote for option 1. it encourage the solver to implement a rather simple feature and do not require a potentially slow/unstable fallback method.

(To my knowledge the basis is needed for the sensitivity analysis of changes in rhs and cost of active constraints and basic variables, respectively. i.e., if not another LP is to be solved..., enlighten me if I'm wrong here=) )

mlubin:
> However, if ConstraintBasisStatus is unsupported/not implemented by the solver how should this be handled?

Throw an error.

mlubin:
This was implemented in https://github.com/JuliaOpt/JuMP.jl/pull/1917.

PR #1917, opened 2019-03-23
Lp sensitivity summary

EdvinAblad:
Adds two functionalities:
`perturbation_range_of_feasibility(constraint)`
`perturbation_range_of_optimality(variable)`
(Targets the issue #1332.)

The choice to present the perturbation (differences) rather than the actual ranges is mainly because the rhs is not necessarily a constant, related to #1890. (It seems sane that these two functions have consistent output).
Are the names too verbose? is e.g., `range_of_feasibility` better?

The functions can be optimized and probably simplified, currently the `Lower/UpperBoundRef` is not used due to #1892. 

The perturbation_range_of_feasibility is implemented for affine and variable constraints. However, the interval sets are intentionally not supported (the interpretation of shifting the entire interval seems far-fetched).

Maybe all sensitivity analysis (possibly also `shadow_price`) should be moved from constraints.jl, this since they are closely connected (using the same internal functions) and in particular `perturbation_range_of_optimality` feels out of place. To where?  

Tests to check these functions are also added. The Clp solver (JuliaOpt/Clp.jl#54) was used to test some larger instances (everything seems fine). 

Documentation is intentionally postponed until e.g., the desired output, names, etc. are decided. 

A last note, these functions can be heavily optimized if vectorized, i.e., if the ranges of all/many coefficients are desired. So it can be implemented if needed.


src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+            return (-Inf, Inf)
+        end
+        if (var_basis_status[j] == MOI.NONBASIC_AT_LOWER && objective_sense(model) == MOI.MIN_SENSE) ||
+            (var_basis_status[j] == MOI.NONBASIC_AT_UPPER && objective_sense(model) == MOI.MAX_SENSE)
+            return (-c_red[j], Inf)
+        end
+        return (-Inf, -c_red[j])
+    end
+
+    basic = [var_basis_status[i] == MOI.BASIC for i in 1:length(var_basis_status)] # .== throws ERROR: MethodError: no method matching length(::MathOptInterface.BasisStatusCode)
```
odow: `basic = var_basis_status .== Ref(MOI.BASIC)`

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+    perturbation_range_of_optimality(var::VariableRef)
+
+Gives the range by which the cost coefficient can change and the current lp-basis
+remains optimal, i.e., the reduced costs remain valid.
+
+## Notes
+- The range denote valid changes, Δ in [l, u], for which c[var] += Δ do not violate the current optimality conditions.
+
+"""
+function perturbation_range_of_optimality(var)
```
odow: Type the input since it's a user-facing function

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+    model = var.model
```
odow: `JuMP.owner_model(var)`

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+    # The variable is in the basis for a minimization problem with no variable upper bounds we need:
+    # c_N - N^T B^-T (c_B + Δ e_j ) >= 0
+    # c_red - Δ N^T (B^-T e_j) = c_red - Δ N^T ρ = c_red - Δ N_red >= 0
+    # c_red >= Δ N_red
+    # c_red >= Δ N_red
+    # maximum( c_red_- ./ N_red_- ) <= Δ <= minimum( c_red_+ ./ N_red_+ )
+    # If upper bounds are present (and active), those inequalities are flipped.
+    # If the problem is a maximization problem all inequalities are flipped.
+    vars = all_variables(model)
+    j = findfirst(vars .== var)
```
odow: `findfirst(isequal(var), vars)` avoids creating a new vector

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+
+    A, L, U, constraints = _std_matrix(model)
+    var_basis_status = _std_basis_status(model)
+    c_red = _std_reduced_costs(model, constraints)
+
+    if var_basis_status[j] != MOI.BASIC
+        if L[j] == U[j]
+            return (-Inf, Inf)
```
odow: Nit: no need for `(` and `)` for returns

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+        if (var_basis_status[j] == MOI.NONBASIC_AT_LOWER && objective_sense(model) == MOI.MIN_SENSE) ||
+            (var_basis_status[j] == MOI.NONBASIC_AT_UPPER && objective_sense(model) == MOI.MAX_SENSE)
+            return (-c_red[j], Inf)
+        end
+        return (-Inf, -c_red[j])
+    end
+
+    basic = [var_basis_status[i] == MOI.BASIC for i in 1:length(var_basis_status)] # .== throws ERROR: MethodError: no method matching length(::MathOptInterface.BasisStatusCode)
+    upper = [var_basis_status[.!basic][i] == MOI.NONBASIC_AT_UPPER for i in 1:length(var_basis_status[.!basic])] # .== throws ERROR: MethodError: no method matching length(::MathOptInterface.BasisStatusCode)
+    ej = zeros(size(A)[1])
```
odow: Instead of `size(A)[1]`, `size(A, 1)`

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+    ej[findfirst(vars[basic[1:length(vars)]] .== var)] = 1.0
+    N_red = A[:, .!basic]' * (A[:, basic]' \ ej)
+    c_red = c_red[.!basic]
+    pos = N_red .> 1e-7
+    neg = N_red .< -1e-7
```
odow: Is `1e-7` special? What is the implication of a value that is `9e-8`?

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+    in_lb = (neg .& .!upper) .| (pos .& upper)
+    in_ub = (pos .& .!upper) .| (neg .& upper)
+    if objective_sense(model) == MOI.MAX_SENSE
+        in_lb, in_ub = in_ub, in_lb
+    end
+    unfixed_vars = (L .< U)[.!basic]
+    in_lb .&= unfixed_vars
+    in_ub .&= unfixed_vars
+    return (maximum([-Inf; c_red[in_lb] ./ N_red[in_lb]]),
+            minimum([Inf; c_red[in_ub] ./ N_red[in_ub]]))
```
odow: Can you add some more comments through here / use more explicit variable names?

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+end
+
+"""
+Returns the optimal reduced costs for the variables of the problem in standard form, c.f. `_std_matrix(model)`
+"""
+function _std_reduced_costs(model::Model, constraints::Vector{ConstraintRef})
+    con_types = list_of_constraint_types(model)
+    vars = all_variables(model)
+    var2idx = Dict{VariableRef, Int64}()
+    for (j, var) in pairs(vars)
```
odow: `in enumerate(vars)` is more explicit. I had to look up what the `pairs` function did.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+            minimum([Inf; UmX_B[pos] ./ rho[pos]; LmX_B[neg] ./ rho[neg]]))
+end
+
+"""
+Returns the optimal reduced costs for the variables of the problem in standard form, c.f. `_std_matrix(model)`
+"""
+function _std_reduced_costs(model::Model, constraints::Vector{ConstraintRef})
+    con_types = list_of_constraint_types(model)
+    vars = all_variables(model)
+    var2idx = Dict{VariableRef, Int64}()
```
odow: The [JuMP style guide](https://www.juliaopt.org/JuMP.jl/latest/style/#Abbreviations-1) would prefer `variable_to_column`

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+function perturbation_range_of_feasibility(constraint::ConstraintRef{Model, _MOICON{F, S}}
+                      ) where {S <: Union{MOI.LessThan, MOI.GreaterThan, MOI.EqualTo}, T, F <: Union{MOI.ScalarAffineFunction{T}, MOI.SingleVariable}}
+    model = constraint.model
+    if termination_status(model) != MOI.OPTIMAL
+        error("The range of feasibility is not available because the current solution "*
+              "is not optimal.")
+    end
+
+    # This throw the error: ... do not support accessing ... ConstraintBasisStatus(), if its not implemented by the solver.
+    con_status = MOI.get(model, MOI.ConstraintBasisStatus(), constraint)
```
odow: Can we catch and throw a more informative error? This function will like be used by new users who don't know that `ConstraintBasisStatus` means.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+    var_basis_status = _std_basis_status(model)
+    x = _std_primal_solution(model, constraints)
+    # Ax = 0 => B*x_B = -N x_N =: b_N
+    # Perturb row i by Δ => ̃x_B = B^-1 (b_N + Δ e_i) = x_B + Δ ρ
+    # Perturb var bound j by Δ => ̃x_B = B^-1 (b_N - Δ N_j) = x_B + Δ ρ
+    # Find min and max values of Δ : L_B <= ̃x_B <= U_B
+    # L_B <= x_B + Δ ρ <= U_B  -> L_B - x_B <= Δ ρ <= U_B - x_B
+    # Δ <= (U_B - x_B)_+ ./ ρ_+,  (L_B - x_B)_- ./ p_-
+    # Δ >= (U_B - x_B)_- ./ ρ_-,  (L_B - x_B)_+ ./ p_+
+    basic = [var_basis_status[i] == MOI.BASIC for i in 1:length(var_basis_status)] # .== throws ERROR: MethodError: no method matching length(::MathOptInterface.BasisStatusCode)
```
odow: DItto comment about `Ref(MOI.BASIC)`

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+        ei = zeros(size(A)[1])
+        ei[i] = 1.0
+        rho = B \ ei
+    end
+    UmX_B = U[basic] - x_B
+    LmX_B = L[basic] - x_B
+    pos = rho .> 1e-7
+    neg = rho .< -1e-7
+    return (maximum([-Inf; UmX_B[neg] ./ rho[neg]; LmX_B[pos] ./ rho[pos]]),
+            minimum([Inf; UmX_B[pos] ./ rho[pos]; LmX_B[neg] ./ rho[neg]]))
```
odow: See my comments about the other function.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+end
+
+
+"""
+Returns the basis status of all variables of the problem in standard form, c.f. `_std_matrix(model)`.
+"""
+function _std_basis_status(model::Model)::Vector{MOI.BasisStatusCode}
+    con_types = list_of_constraint_types(model)
+    vars = all_variables(model)
+    var2idx = Dict{VariableRef, Int64}()
```
odow: see my comments about `_std_reduced_costs`

test/constraint.jl
```diff
@@ -561,3 +561,220 @@ end
...
+
+        @test true_range[1] ≈ range[1]
+        @test true_range[2] ≈ range[2]
+    end
+end
+
+
+@testset "range_of_feasibility" begin
+    test_perturbation_range_of_feasibility("""
+    variables: x, y, z, w
```
odow: Indentation in these functions?

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
     return Tuple{DataType, DataType}[(jump_function_type(model, f), s)
                                      for (f,s) in list]
 end
+
+
```
mlubin: I'd recommend moving this code into a new file called `lp_sensitivity.jl`. It's pretty specialized and doesn't have much to do with the rest of constraints.jl.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+"""
+    perturbation_range_of_feasibility(constraint::ConstraintRef)
+
+Gives the range by which the rhs coefficient can change and the current lp-basis
```
mlubin: What is the definition of the rhs coefficient of a constraint? (I know what it is, readers might not.)

mlubin: What is the return type of this function?

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
     return Tuple{DataType, DataType}[(jump_function_type(model, f), s)
                                      for (f,s) in list]
 end
+
+
+"""
+    perturbation_range_of_feasibility(constraint::ConstraintRef)
```
mlubin: Maybe call this `lp_rhs_perturbation_range`? "Range of feasibility" could also be interpreted as the slack in the constraint, i.e., how much you can perturb the constraint to invalidate the current solution.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+- The range denote valid changes, e.g., for a*x <= b + Δ, the lp-basis remains feasible for all Δ in [l, u].
+
+"""
+function perturbation_range_of_feasibility(constraint::ConstraintRef{Model, <:_MOICON})
+    error("The range of feasibility is not defined or not implemented for this type " *
+          "of constraint.")
+end
+
+"""
+Builds the standard form constraint matrix, varaible bounds (rhs := 0, by choice of slack bounds), and a vector of constraint reference one for each row.
```
mlubin: Please be more specific about how the slack variables are set up.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+i.e., the LP-problem feasible set on the form
+s.t.  Ax == 0
+L <= x <= U
```
mlubin: Document the four things that are returned as a tuple.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+"""
+function _std_matrix(model::Model)
+    con_types = list_of_constraint_types(model)
+    con_func_type = first.(con_types)
+    if !all(broadcast(<:, AffExpr, con_func_type) .| broadcast(<:, VariableRef, con_func_type))
+        error("Range of feasibility and optimality require all constraints functions to be affine (the problem to be linear).")
+    end
+    vars = all_variables(model)
+    var2idx = Dict{VariableRef, Int64}()
```
mlubin: Nit: var_to_idx

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+                if S <: MOI.Interval
+                    var_lower[j] = constr_obj.set.lower
+                    var_upper[j] = constr_obj.set.upper
+                end
+                if S <: MOI.LessThan
+                    var_upper[j] = constr_obj.set.upper
+                end
+                if S <: MOI.GreaterThan
+                    var_lower[j] = constr_obj.set.lower
+                end
```
mlubin: Error if the set isn't one of the 4 types checked for. This code will silently ignore integrality constraints.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+                push!(var_upper, constr_obj.set.upper)
+            end
+            if S <: MOI.LessThan
+                push!(var_lower, -Inf)
+                push!(var_upper, constr_obj.set.upper)
+            end
+            if S <: MOI.GreaterThan
+                push!(var_lower, constr_obj.set.lower)
+                push!(var_upper, Inf)
+            end
```
mlubin: Error if the set isn't one of the 4 types checked for.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+        end
+    end
+
+    A = sparse(row_i, col_j, coeffs, m, n)
+    return A, var_lower, var_upper, affine_constraints
+end
+"""
```
mlubin: Nit: Missing newline after `end`.

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+    end
+
+    if objective_sense(model) == MOI.MAX_SENSE
+        y .*= -1.0
+    end
+    return y
+end
+
+"""
+    perturbation_range_of_optimality(var::VariableRef)
```
mlubin: `lp_objective_perturbation_range`?

src/constraints.jl
EdvinAblad: 9e-8 is treated as zero. (neither negative nor positive)
Don't know if 1e-7 is the best choice. (But it usually works =) )
(This is related to the minimum ratio test in the simplex method.)
Choosing the epsilon too small makes the analysis break down, since a number that should be positive is due to numerical reasons deemed negative.  And too large epsilon is bad since elements that is clearly non-zero is regarded to be zero. 

But yes, making use of absolute tolerance assumes that the input is not strangely scaled...
Any ideas, experience, or other input are most welcomed.
(The most common argument for allowing the use of absolute tolerance within the simplex method is by first rescaling the columns and rows in A)

src/constraints.jl
```diff
@@ -727,3 +727,313 @@ function list_of_constraint_types(model::Model)
...
+          "of constraint.")
+end
+
+"""
+Builds the standard form constraint matrix, varaible bounds (rhs := 0, by choice of slack bounds), and a vector of constraint reference one for each row.
+i.e., the LP-problem feasible set on the form
+s.t.  Ax == 0
+L <= x <= U
+"""
+function _std_matrix(model::Model)
```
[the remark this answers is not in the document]
EdvinAblad: That would indeed be neat. However, I can't get it to work, I might be missing some thing (not so used with automatic bridges...)

Main issue: Every constraint should get an associated slack variable.
My idea was to copy the model to a model having only interval constraints, (Since you can create an MOI.Interval, from other the other sets.) However, I always get an unsupported constraint error. Moreover, I can't find a bridge that creates intervals (only the ones that splits intervals). 

If this was accomplished then the slack bridge would accomplish the rest (I think).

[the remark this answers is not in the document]
blegat: No there is not any bridge that transforms `GreaterThan` or `LessThan` to `Interval` yet. You can try with
```julia
MOIU.@model(MyModel,
    (),
    (MOI.Interval, MOI.GreaterThan, MOI.LessThan, MOI.EqualTo),
    (),
    (),
    (MOI.SingleVariable,),
    (),
    (MOI.Zeros,),
    (MOI.VectorAffineFunction)
)
```

EdvinAblad:
Revised according to most of your comments. 
The simplification of _std_matrix still remains (I did some minor simplifications).
Have also realized that this solution currently don't support models containing vectorized constraints, this could probably be resolved by revising _std_matrix...
Also the tolerance issue should probably be targeted.

Also, (as you noted) I tend to use abbreviations, e.g., var for variable.
Since I think it makes the code more readable, but I really don't care, if you want verbose, I revise again.

Thank you for your feedback =) It's very appreciated.

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+#############################################################################
+
+"""
+    lp_rhs_perturbation_range(constraint::ConstraintRef)::Tuple{Float64, Float64}
+
+Gives the range by which the rhs coefficient can change and the current lp-basis
+remains feasible, i.e., where the shadow prices apply.
+
+## Notes
+- The rhs coefficient is the value right of the relation, i.e., b for the constraint when on the form a*x □ b, where □ is ≤, =, or ≥.
```
odow: s/when on the/when of the

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+- The range denote valid changes, e.g., for a*x <= b + Δ, the lp-basis remains feasible for all Δ ∈ [l, u].
```
odow: s/denote/denotes

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+    return A, var_lower, var_upper, affine_constraints
+end
+
+"""
+Returns the optimal primal values for the problem in standard form, c.f. `_std_matrix(model)`.
+"""
+function _std_primal_solution(model::Model, constraints::Vector{ConstraintRef})
+    vars = all_variables(model)
+    x = [value.(vars); value.(constraints)]
+    return x
```
odow: Nit: no need to have `x`, just return the vector directly

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+                j = var_to_idx[con_obj.func]
+                basis_status[j] = MOI.get(model, MOI.ConstraintBasisStatus(), constraint)
+                if S <: MOI.LessThan && basis_status[j] == MOI.NONBASIC
+                    basis_status[j] = MOI.NONBASIC_AT_UPPER
+                elseif S <: MOI.GreaterThan && basis_status[j] == MOI.NONBASIC
+                    basis_status[j] = MOI.NONBASIC_AT_LOWER
+                end
+            end
+            #issue 1892 makes is hard to know how to handle variable constraints, atm. assume unique variable constraints.
+            continue
```
odow: Instead of this continue,
```julia
if F <: VariableRef
    # ... what you have ...
else
    for constraint in constraints
        # ...
    end
end
```

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+        if F <: VariableRef
+            for constraint in constraints
+                con_obj = constraint_object(constraint)
+                j = var_to_idx[con_obj.func]
+                basis_status[j] = MOI.get(model, MOI.ConstraintBasisStatus(), constraint)
+                if S <: MOI.LessThan && basis_status[j] == MOI.NONBASIC
+                    basis_status[j] = MOI.NONBASIC_AT_UPPER
+                elseif S <: MOI.GreaterThan && basis_status[j] == MOI.NONBASIC
+                    basis_status[j] = MOI.NONBASIC_AT_LOWER
+                end
```
odow: Should you have an `else`? What about `VariableRef`-in-`MOI.ZeroOne` constraints for example? Or will this be caught on the get for `ConstraintBasicStatus`? What is the error that will be thrown?

Edit: I see later on that you catch the error. Maybe leave a note here saying that it is already checked.

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+            #issue 1892 makes is hard to know how to handle variable constraints, atm. assume unique variable constraints.
+            continue
+        end
+        for constraint in constraints
+            con_basis_status = MOI.get(model, MOI.ConstraintBasisStatus(), constraint)
+            if S <: MOI.LessThan && con_basis_status == MOI.NONBASIC
+                con_basis_status = MOI.NONBASIC_AT_UPPER
+            elseif S <: MOI.GreaterThan && con_basis_status == MOI.NONBASIC
+                con_basis_status = MOI.NONBASIC_AT_LOWER
+            end
```
odow: Ditto on the `else`

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+                return value(constraint) - constraint_object(constraint).set.upper, Inf
+            end
+            if S <: MOI.GreaterThan
+                return -Inf, value(constraint) - constraint_object(constraint).set.lower
+            end
+            return 0.0, 0.0
+        end
+    catch e
+        error("The perturbation range of rhs is not available because the solver doesn't "*
+              "support ´ConstraintBasisStatus´.")
```
odow: You should probably just `try-catch` the `get. If an error is thrown by, for example, `value(constraint)`, this error is incorrect.

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+end
+
+"""
+    lp_objective_perturbation_range(var::VariableRef)::Tuple{Float64, Float64}
+
+Gives the range by which the cost coefficient can change and the current lp-basis
+remains optimal, i.e., the reduced costs remain valid.
+
+## Notes
+- The range denote valid changes, Δ in [l, u], for which c[var] += Δ do not violate the current optimality conditions.
```
odow: s/denote/denotes

What is `c`?

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+        constraints = all_constraints(model, F, S)
+        if F <: VariableRef
+            for constraint in constraints
+                con_obj = constraint_object(constraint)
+                j = var_to_idx[con_obj.func]
+                range = MOI.Interval(con_obj.set)
+                var_lower[j] = max(var_lower[j], range.lower)
+                var_upper[j] = min(var_upper[j], range.upper)
+            end
+            #issue 1892 makes is hard to know how to handle variable constraints, atm. assume unique variable constraints.
```
mlubin: Typo: "is" -> "it"

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+    if objective_sense(model) == MOI.FEASIBILITY_SENSE
+        error("The perturbation range of the objective is not applicable on feasibility problems.")
+    end
+    if !has_duals(model)
+        error("The perturbation range of the objective is not available because no dual result is " *
+              "available.")
+    end
+    # Optimization possible: since the has_upper_bound(var) and has_lower_bound(var)
+    # Do not capture all variable bounds (issue 1892), it's hard to directly
+    # access the reduced cost of a variable, now the complete matrix is always built.
```
mlubin: I don't understand this comment. Doesn't `_std_reduced_costs` directly access the reduced costs of the variables?

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+
+    # The variable is in the basis for a minimization problem with no variable upper bounds we need:
```
mlubin: This is a run-on sentence ("the variable is... we need").

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+    # c_N - N^T B^-T (c_B + Δ e_j ) >= 0
+    # c_red - Δ N^T (B^-T e_j) = c_red - Δ N^T ρ = c_red - Δ N_red >= 0
+    # c_red >= Δ N_red
+    # c_red >= Δ N_red
```
mlubin: Is this line meant to be repeated?

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+            end
+            return 0.0, 0.0
+        end
+    catch e
+        error("The perturbation range of rhs is not available because the solver doesn't "*
+              "support ´ConstraintBasisStatus´.")
+    end
+    # The constraint is active.
+
+    # Optimization possible, only basic columns needed
```
mlubin: Rephrase "optimization possible" as a TODO, e.g., "TODO: This could be made more efficient because actually we only need the basic columns."

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,322 @@
...
+        rho = -B \ Vector(A[:, j])
+    else
+        i = findfirst(isequal(constraint), constraints)
+        ei = zeros(size(A, 1))
+        ei[i] = 1.0
+        rho = B \ ei
+    end
+    UmX_B = U[basic] - x_B
+    LmX_B = L[basic] - x_B
+    pos = rho .> 1e-7
```
mlubin: I would argue that we should stick to the textbook version and warn if we detect possible numerical instability. Otherwise, these numerical constants need a justification.

mlubin:
> Also why do the 32bit version fail to build?

The [error](https://ci.appveyor.com/project/mlubin/jump-jl/builds/23483452/job/xjwigjyndl0imigj#L196) is `MethodError: no method matching lu!(::SparseMatrixCSC{Float64,Int64}, ::Val{true}; check=true)` and if you dig into the stack trace you see,
```julia
 [7] lp_rhs_perturbation_range(::ConstraintRef{Model,MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.EqualTo{Float64}},ScalarShape}) at C:\projects\jump-jl\src\lp_sensitivity.jl:203
```

Because Julia doesn't follow JuMP's style recommendation ["A user should see a MethodError only for methods that they called directly."](http://www.juliaopt.org/JuMP.jl/v0.19.0/style/#User-facing-MethodError-1), it's on you to infer which internal assumption of backslash is violated. I would guess that you need to use `SparseMatrixCSC{Float64,Int32}` on a 32-bit system.

src/lp_sensitivity.jl
EdvinAblad: I would claim that this 1e-7 is the textbook version =)
C.f. "Computational Techniques of the Simplex Method" by István Maros section 9.3.4.
Out of context but this regards the values for the feasibility range:
> In the latter case a typical values are 1e-8  <= e_f <= 1e-6 if
standard double precision arithmetic with 53 bit mantissa is used.

EdvinAblad:
Regarding the tolerances used:
-  These are the primal/dual feasibility tolerances. 
-  Without them the sensitivity analysis is likely to frequently fail since, e.g., solvers has a tolerance for <= constraints, allowing them to be slightly violated.   
- Now I use the smallest suggested (1e-8) this allow a slightly wider range of applicable models. But might return too narrow bounds. (C.f. "Computational Techniques of the Simplex Method" by István Maros section 9.3.4)

Some alternative remedies:
- Solvers has these as parameters, e.g., Clp has `PrimalTolerance `and `DualTolerance`, could it be possible to get these? Note that the default values varies among solvers, e.g., Clp use 1e-7 and Gurobi use 1e-6. 
- Since these are model dependent, the best option could be to expose them to the user. (As an optional argument with a sane default value).

Regarding getting the input (constraint matrix, variable bounds, constraint bounds):
-  Currently, Vectorized constraints are yet not supported in the lp_sensitivity. Moreover, should the ConstraintBasisStatus for vectorized constraint be implemented in each solver interface or could e.g., LQOI do some translation. (Are vectorized constraints necessary to support atm? e.g., `shadow_price` do not yet support them)
- I'm getting increasingly unsure of how the `copy_to` would simplify the code... If we can't transform it to only interval constraints then it feels like there will be as much code needed to loop through the new transformed problem... (Of course it could be used to generalize the procedure, if e.g., there would be a "unvectorize bridge") 

Again thanks for the feedback =)

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,340 @@
...
+        rho = B \ ei
+    end
+    UmX_B = U[basic] - x_B
+    LmX_B = L[basic] - x_B
+    pos = rho .> 0.0
+    neg = rho .< 0.0
+    # Find the first basic variable bound that is strictly violated.
+    # Use the vanilla 1e-8 tolerance (c.f. "Computational Techniques of the Simplex Method" by István Maros section 9.3.4.)
+    # to determine this strict violation.
+    tol = 1e-8
```
odow: Maybe the best way forward is to pass in `tol` as a keyword argument with `1e-8` as the default tolerance.

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,340 @@
...
+        in_lb, in_ub = in_ub, in_lb
+    end
+    # The reduced cost of variables fixed at equality do not be accounted for (they only change binding direction).
+    unfixed_vars = (var_lower .< var_upper)[.!basic]
+    in_lb .&= unfixed_vars
+    in_ub .&= unfixed_vars
+    # Find the first reduced cost that is strictly violated
+    # Use the vanilla 1e-8 tolerance (c.f. "Computational Techniques of the Simplex Method" by István Maros section 9.3.4.)
+    # to determine this strict violation.
+    tol = 1e-8
```
odow: Ditto on `tol`.

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,340 @@
...
+    if objective_sense(model) == MOI.FEASIBILITY_SENSE
+        error("The perturbation range of the objective is not applicable on feasibility problems.")
+    end
+    if !has_duals(model)
+        error("The perturbation range of the objective is not available because no dual result is " *
+              "available.")
+    end
+#    TODO: This could be made more efficient for non-basic variables by checking them
+#          before building calling ´_std_matrix´ and ´_std_reduced_costs´ and only check
+#          the reduced cost of that variable. (issue 1892 makes this somewhat non-trivial)
```
odow: Nit: indentation is out.

src/lp_sensitivity.jl
```diff
@@ -0,0 +1,340 @@
...
+        ei = zeros(size(A, 1))
+        ei[i] = 1.0
+        rho = B \ ei
+    end
+    UmX_B = U[basic] - x_B
+    LmX_B = L[basic] - x_B
+    pos = rho .> 0.0
+    neg = rho .< 0.0
+    # Find the first basic variable bound that is strictly violated.
+    # Use the vanilla 1e-8 tolerance (c.f. "Computational Techniques of the Simplex Method" by István Maros section 9.3.4.)
```
mlubin: Please move this tolerance up to where the `1e-8` appears, and mention that this tolerance should match the feasibility tolerance used by the LP solver.

test/lp_sensitivity.jl
```diff
@@ -0,0 +1,214 @@
+function test_lp_rhs_perturbation_range(model_string, primal_solution, basis_status, feasibility_ranges)
```
mlubin: Please copy the standard license header for this file.

docs/src/solutions.md
```diff
@@ -17,6 +17,7 @@ optimization step. Typical questions include:
  - Do I have a solution to my problem?
  - Is it optimal?
  - Do I have a dual solution?
+ - How sensitive is the solution to data uncertainty?
```
mlubin: I would call this "data perturbations" instead of "data uncertainty". Uncertainty is an overloaded term and makes me think of robust or stochastic programming.

docs/src/solutions.md
```diff
@@ -116,6 +117,84 @@ end
 # TODO: How to accurately measure the solve time.
 ```
 
+## Sensitivity analysis
```
mlubin: "Sensitivity analysis for LP" to make the scope clear from the title.

docs/src/solutions.md
```diff
@@ -116,6 +117,84 @@ end
...
+
+Given an LP problem and an optimal solution corresponding to a basis, we can
+question how much an objective coefficient or standard form rhs coefficient
+(c.f., [`standard_form_rhs`](@ref)) can change without violating primal or dual
+feasibility of the basic solution.
+
```
mlubin: Please add a note that this form of sensitivity analysis appears often in textbooks but has limited use in practice because optimal bases are often degenerate.

docs/src/solutions.md
```diff
@@ -116,6 +117,84 @@ end
...
+Given an LP optimal solution (and both [`has_values`](@ref) and
+[`has_duals`](@ref) returns `true`) [`lp_objective_perturbation_range`](@ref)
```
mlubin: Also mention that you must use a solver that computes and returns the optimal basis. Many of the solvers available in JuMP for solving LPs either don't compute an optimal basis at all or don't yet have this exposed.

docs/src/solutions.md
mlubin: (If you object to this characterization, I'd be interested to hear the argument.)

EdvinAblad: True, it depends quite a lot on the instance if these ranges are practically useful, I added a paragraph discussing briefly their limitations.

Issue #1920, opened 2019-03-30
Set silent


blegat:
We now have a [`Silent` MOI attribute](http://www.juliaopt.org/MathOptInterface.jl/stable/apireference/#MathOptInterface.Silent) (even if it is not yet implemented by many optimizers).
It can be set to JuMP with `MOI.set(model, MOI.Silent(), true)` and unset with `MOI.set(model, MOI.Silent(), false)`.
It would be nice to have a `set_silent` and `unset_silent` that redirect to these `MOI.set` calls.

mlubin:
Closed by #1921

PR #1921, opened 2019-04-02
MOI Silent

guilhermebodin:
This is a start to solve #1920. 

I'm not sure how exactly to test this feature as JuMP doesn't depend on solvers to test it. Maybe testing an error the same way it is tested in MOI? 
```julia
attr = MOI.Silent()
exception = ErrorException(
    "Cannot query $(attr) from caching optimizer because no optimizer" *
    " is attached.")
@test_throws exception MOI.get(model, attr)
```

Also not sure where to document this feature in the manual.

### DIFF
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -95,3 +95,28 @@ JuMP.direct_model
 ```@docs
 JuMP.backend
 ```
+
+## Solver attributes
+
+Some solvers attributes can be queried through JuMP Models. Some solver attributes can setted through JuMP Models.
+
+```@docs
+solver_name
+```
+
+```@docs
+bridge_constraints
+```
+
+```@docs
+num_variables
+```
+
+```@docs
+set_silent
+```
+
+
+```@docs
+unset_silent
+```
--- a/docs/src/solvers.md
+++ b/docs/src/solvers.md
@@ -98,25 +98,13 @@ JuMP.backend
 
 ## Solver attributes
 
-Some solvers attributes can be queried through JuMP Models. Some solver attributes can setted through JuMP Models.
+Some solver attributes can be queried and set through JuMP models.
 
 ```@docs
 solver_name
-```
 
-```@docs
 bridge_constraints
-```
 
-```@docs
-num_variables
-```
-
-```@docs
 set_silent
-```
-
-
-```@docs
 unset_silent
 ```
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -316,6 +316,30 @@ function _try_get_solver_name(model_like)
     end
 end
 
+function _try_set_silent(model_like)
+    try
+        return MOI.set(model_like, MOI.Silent(), true)
+    catch ex
+        if isa(ex, ArgumentError)
+            return "Silent() attribute not implemented by the optimizer."
+        else
+            rethrow(ex)
+        end
+    end
+end
+
+function _try_unset_silent(model_like)
+    try
+        return MOI.set(model_like, MOI.Silent(), false)
+    catch ex
+        if isa(ex, ArgumentError)
+            return "Silent() attribute not implemented by the optimizer."
+        else
+            rethrow(ex)
+        end
+    end
+end
+
 """
     solver_name(model::Model)
 
@@ -445,7 +469,12 @@ Takes precedence over any other attribute controlling verbosity
 and requires the solver to produce no output.
 """
 function set_silent(model::Model)
-    return MOI.set(model, MOI.Silent(), true)
+    if mode(model) != DIRECT &&
+        MOIU.state(backend(model)) == MOIU.NO_OPTIMIZER
+        return "No optimizer attached."
+    else
+        return _try_set_silent(backend(model))
+    end
 end
 
 """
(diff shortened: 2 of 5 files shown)


blegat:
> I'm not sure how exactly to test this feature

You can use a `MOIU.MockOptimizer(MOIU.UniversalFallback(JuMP._MOIModel{Float64}()))`, it should support it.

> not sure where to document this feature in the manual.

I would suggest adding a subsection in the solver section.

src/JuMP.jl
```diff
@@ -438,6 +462,35 @@ end
...
+"""
+    set_silent(model::Model)
+
+Takes precedence over any other attribute controlling verbosity 
+and requires the solver to produce no output.
+"""
+function set_silent(model::Model)
+    if mode(model) != DIRECT &&
+        MOIU.state(backend(model)) == MOIU.NO_OPTIMIZER
+        return "No optimizer attached."
```
blegat: If no optimizer is attached, the silent attribute will be copied to the optimizer when it is attached so I don't think we need to make an exception

src/JuMP.jl
```diff
@@ -316,6 +316,30 @@ function _try_get_solver_name(model_like)
     end
 end
 
+function _try_set_silent(model_like)
+    try
+        return MOI.set(model_like, MOI.Silent(), true)
+    catch ex
+        if isa(ex, ArgumentError)
+            return "Silent() attribute not implemented by the optimizer."
```
blegat: What is the `ArgumentError` thrown ?

src/JuMP.jl
guilhermebodin: Ok, I thought it should follow the same logic as the solver_name function, but indeed it makes sense that the silent works even if no optimizer is attached

```julia
function solver_name(model::Model)
    if mode(model) != DIRECT &&
        MOIU.state(backend(model)) == MOIU.NO_OPTIMIZER
        return "No optimizer attached."
    else
        return _try_get_solver_name(backend(model))
    end
end
```

src/JuMP.jl
guilhermebodin: Also tried following the same logic as `solver_name` in the auxiliary `_try_get_solver_name` function but this also does not make much sense, as the argument Silent() should always work independently of the optimizer

src/JuMP.jl
guilhermebodin: For what I understood now this is what we want to work
```julia
model = Model()
MOI.set(model, MOI.Silent(), true)
```
but unfortunately it gives this error
```julia
ERROR: MethodError: no method matching set(::Model, ::MathOptInterface.Silent, ::Bool)
Closest candidates are:
  set(::Model, ::MathOptInterface.AbstractModelAttribute, ::Any) at /Users/guilhermebodin/Documents/JuMP.jl/src/JuMP.jl:643
  set(::MathOptInterface.Utilities.MockOptimizer, ::Union{AbstractModelAttribute, AbstractOptimizerAttribute}, ::Any) at /Users/guilhermebodin/.julia/packages/MathOptInterface/C3lip/src/Utilities/mockoptimizer.jl:157
  set(::MathOptInterface.Utilities.CachingOptimizer, ::MathOptInterface.AbstractOptimizerAttribute, ::Any) at /Users/guilhermebodin/.julia/packages/MathOptInterface/C3lip/src/Utilities/cachingoptimizer.jl:511
```
Because `Silent` attribute is an `AbstractOptimizerAttribute`
```julia
struct Silent <: AbstractOptimizerAttribute end
```
I tested implementing the following function and now the first code works
```julia
MOI.set(m::Model, attr::MOI.AbstractOptimizerAttribute, value) = MOI.set(backend(m), attr, value)
model = Model()
MOI.set(model, MOI.Silent(), true)
```
and then I can write the `set_silent` as 

```julia
function set_silent(model::Model)
    return MOI.set(model, MOI.Silent(), true)
end
```
have I got it right?

blegat: Yes, that looks correct :)

blegat:
Looks good, just need to be added to the doc

docs/src/solvers.md
```diff
@@ -95,3 +95,28 @@ JuMP.direct_model
 ```@docs
 JuMP.backend
 ```
+
+## Solver attributes
+
+Some solvers attributes can be queried through JuMP Models. Some solver attributes can setted through JuMP Models.
```
mlubin: "Some solver attributes can be queried and set through JuMP models."

docs/src/solvers.md
```diff
@@ -95,3 +95,28 @@ JuMP.direct_model
...
+```@docs
+solver_name
+```
+
+```@docs
+bridge_constraints
+```
+
+```@docs
+num_variables
```
mlubin: Is `num_variables` really a solver attribute? You can query it when no solver is attached.

docs/src/solvers.md
```diff
@@ -95,3 +95,28 @@ JuMP.direct_model
...
+```
+
+```@docs
+set_silent
+```
+
+
+```@docs
```
mlubin: Why separate `@docs` blocks for `set_silent` and `unset_silent`?

docs/src/solvers.md
guilhermebodin: True, this could be in the variables section

docs/src/solvers.md
guilhermebodin: No special reason, I didn't realize that this
```julia
 @docs
set_silent
unset_silent
```
and this
```julia
@docs
set_silent
@docs
unset_silent
```
produce the same output

PR #1922, opened 2019-04-03
Development roadmap

mlubin:
This is a bit drafty but has been sitting on a local branch for a few weeks. I didn't copy over the MOI 1.0 points. Would we rather list them here or in MOI?

Do we want the roadmap to be in the official documentation? Another option is to have `roadmap.md`, which would make it less visible to users. Unclear if this is a good outcome.

@odow @blegat @joaquimg @juan-pablo-vielma

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -28,7 +28,8 @@ makedocs(
         "Query Solutions" => "solutions.md",
         "Nonlinear Modeling" => "nlp.md",
         "Style Guide" => "style.md",
-        "Extensions" => "extensions.md"
+        "Extensions" => "extensions.md",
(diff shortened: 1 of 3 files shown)


odow:
> Would we rather list them here or in MOI?

Here.

> Do we want the roadmap to be in the official documentation? 

Yes.

joaquimg:
Agree with @odow in both.
Also liked the idea of adding beyond JuMP 1.0

juan-pablo-vielma:
+1 on including it in the official documentation, including MOI and beyond v1.0. 
Would also be a good place to elaborate on plans for popular issues (e.g. callbacks) as they become clearer (with ad-hoc notifications of updates to the file).

PR #1929, opened 2019-04-08
Add start_value to docs

### LINKED ISSUE #1871: start_value and set_start_value are not listed in the documentation

odow:
Closes #1871.

I made one substantive change that we can debate: `set_start_value` now always returns `nothing`. Previously, it returned whatever the return of `MOI.set` was, which is solver-specific. We should standardize on return conventions.

In support of my change, `set_upper_bound`, `set_binary` et al all return `nothing`.

### DIFF
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -629,6 +629,13 @@ julia> start_value(y)
 2.0
 ```
 
+Prior to JuMP 0.19, the previous solution to a solve was automatically set as
+the new starting value. JuMP 0.19 no longer does this automatically. To
+reproduce the functionality, use:
+```julia
+set_start_value.(all_variables(model), value.(all_variables(model)))
+```
+
 ## Reference
 
 ```@docs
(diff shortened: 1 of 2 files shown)


docs/src/variables.md
```diff
@@ -600,6 +600,35 @@ in the model. This is useful for performing operations like:
...
+y
+
+julia> start_value(y)
+1.0
+
+julia> set_start_value(y, 2)
+
+julia> start_value(y)
+2.0
+```
```
mlubin: It may be worth adding a note that pre-0.19, the optimal solution from a solve would serve as the starting point for the next solve. To get the same thing now, you can write `set_start_value.(m, value.(all_variables(m)))`.

docs/src/variables.md
```diff
@@ -600,6 +600,42 @@ in the model. This is useful for performing operations like:
...
+
+Prior to JuMP 0.19, the previous solution to a solve was automatically set as
```
mlubin: Format in a `!!! note`?

Issue #1906, opened 2019-03-15
more tools for determining if expressions are zero


ExpandingMan:
It would be nice to have a function that determines if, for example, a `GenericAffExpr` is truly zero.  For example
```julia               
function expr_iszero(expr::GenericAffExpr)  
    iszero(expr.constant) || (return false)  
    for c ∈ values(expr.terms)
        iszero(c) || (return false) 
    end 
    true
end
```
The current behavior is to check whether the `terms` are empty.  Presumably this is for performance reasons.  Since the current `iszero` doesn't give the correct answer, I wonder if it makes sense to change `Base.iszero` to the above, but to use a different internal function with the current definition in cases where the efficiency is needed.  If so, at the same time we should expose functions for pruning zeros from expressions (perhaps `iszero!`).  Obviously it's still possible to get unexpected answers due to floating point error, but there's nothing we can do about that.

Thoughts?

odow:
Sure, you could implement `_is_zero(expr)` as the fast check and then overload `Base.iszero` for the comprehensive.

I think we would probably consider a PR with tests and docs. I've needed this myself at some points.

> we should expose functions for pruning zeros from expressions (perhaps iszero!)

The JuMP style guide would suggest `drop_zeros(expr)`.

mlubin:
@odow, is this resolved now?

PR #1933, opened 2019-04-08
Fix iszero for expressions

odow:
Part I of #1906 

To answer the question:
> The current behavior is to check whether the terms are empty. Presumably this is for performance reasons. 

Nope. It's just legacy: https://github.com/JuliaOpt/JuMP.jl/commit/09383f2fcacb69aaccde0b7d4c34489927489c59

It's more surprising that no one has complained about this before. As a side note, it doesn't appear to be called in JuMP.jl at all, so there aren't any performance concerns on our part. Better to be correct than fast.

### DIFF
--- a/test/expr.jl
+++ b/test/expr.jl
@@ -28,6 +28,7 @@ function expressions_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType
         @test !iszero(x + 1)
         @test !iszero(x + 0)
         @test iszero(0 * x + 0)
+        @test iszero(x - x)
     end
 
     @testset "isequal(::GenericQuadExpr)" begin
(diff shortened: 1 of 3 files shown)


test/expr.jl
```diff
@@ -22,6 +22,14 @@ function expressions_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType
         @test hash(x + 1) == hash(x + 1)
     end
 
+    @testset "iszero(::GenericAffExpr)" begin
+        m = ModelType()
+        @variable(m, x)
+        @test !iszero(x + 1)
+        @test !iszero(x + 0)
+        @test iszero(0 * x + 0)
```
blegat: I would also add `x - x`

PR #1935, opened 2019-04-08
Add standard_form modifications

### LINKED ISSUE #1890: Modify RHS (and define RHS)

odow:
Closes #1890

I have deprecated `set_coefficient` in favor of `set_standard_form_coefficient`.

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -395,20 +395,19 @@ function add_constraint(model::Model, c::AbstractConstraint, name::String="")
 end
 
 """
-    set_coefficient(constraint::ConstraintRef, variable::VariableRef, value)
+    set_standard_form_coefficient(constraint::ConstraintRef, variable::VariableRef, value)
 
 Set the coefficient of `variable` in the constraint `constraint` to `value`.
 
 Note that prior to this step, JuMP will aggregate multiple terms containing the
 same variable. For example, given a constraint `2x + 3x <= 2`,
-`set_coefficient(c, x, 4)` will create the constraint `4x <= 2`.
-
+`set_standard_form_coefficient(c, x, 4)` will create the constraint `4x <= 2`.
 
 ```jldoctest; setup = :(using JuMP), filter=r"≤|<="
 model = Model()
 @variable(model, x)
 @constraint(model, con, 2x + 3x <= 2)
-set_coefficient(con, x, 4)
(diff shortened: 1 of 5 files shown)


src/constraints.jl
```diff
@@ -395,36 +395,94 @@ function add_constraint(model::Model, c::AbstractConstraint, name::String="")
...
+    MOI.modify(backend(owner_model(constraint)), index(constraint),
+               MOI.ScalarCoefficientChange(index(variable), convert(T, value)))
+    return
+end
+@deprecate set_coefficient set_standard_form_coefficient
+
+"""
+    standard_form_coefficient(constraint::ConstraintRef, variable::VariableRef)
+
+Return the coefficient associated with `variable` in `cosntraint` after JuMP has
```
joaquimg: constraint

blegat:
I don't support renaming `set_coefficient` to set_standard_form_coefficient`.
The "standard form" is confusing because of the confusion with https://en.m.wikipedia.org/wiki/Linear_programming#Standard_form

odow:
Not sure if this counts as breaking because it adds deprecations.

blegat:
Do you think it should be included in v0.19.1 ?

odow:
I don't mind. It isn't urgent, so it can probably wait. (Selfish reason, I'll have to update SDDP.jl, and relying on a patch 0.19.1 isn't fun.)

blegat:
Even if there is warning, if someone release a package using this function and people use this package, when updating JuMP, the package will throw warning so it is kind of breaking.
I'd like to release MOI v0.9 in a few days so if we give a week for packages to update to it, we can release JuMP v0.20 in about a week.

PR #1947, opened 2019-04-18
Ignore terms with a zero coefficient when adding to an `OrderedDict`

### LINKED ISSUE #1946: Extreme performance degradation from v0.18.5 -> v0.19
Heya JuMP team!

I've currently noticed a large performance regression (which has been cited in a few cases, notably #1403 and #1905). I was planning on releasing some code for a paper, which was originally written with v0.18.5, but, with the upgrade, decided to do the necessary edits for v0.19. The code is essentially (sadly :( ) unusable in its current form, due to the speed degradation.

Two MWEs showing a large difference (I will note that I'm using two slightly different environments, with Julia 1.0 and Julia 1.1)

[Julia 1.0, JuMP v0.18.5](https://gist.github.com/angeris/8ac92dee79081c946979bcce8160f656). Time to complete: <2s. Roughly .0001s/constraint.

[Julia 1.1, JuMP v0.19](https://gist.github.com/angeris/2f3ae08642c001d8f0ccf7aad03a6a8f). ETA for completing this (as reported by `ProgressMeter`): 18 minutes. Roughly 1.3s/constraint.

The problem in question solved was much, much larger and would take several days to formulate at this pace. I'd also be happy to contribute to a fix, but I'd need to learn a bit more about profiling and performance in Julia (and definitely more about the current JuMP implementation) in order to help out!

Anyways, thank you so much for this project, it's been super useful :)

odow:
This addresses the `TODO` listed in `_add_or_set!`. 

The answer is not to call `drop_zeros`, but just to exclude terms with a zero coefficient. Thus, `x + 0y` will return `x`, but `x + y - y` will return `x + 0 y`.

Closes #1946.

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -19,10 +19,11 @@
 function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
     # Adding zero terms to this dictionary leads to unacceptable performance
     # degradations. See, e.g., https://github.com/JuliaOpt/JuMP.jl/issues/1946.
-    if !iszero(v)
-        # TODO: This unnecessarily requires two lookups for k.
-        dict[k] = get!(dict, k, zero(V)) + v
+    if iszero(v)
+        return dict  # No-op.
     end
+    # TODO: This unnecessarily requires two lookups for k.
+    dict[k] = get!(dict, k, zero(V)) + v
     return dict
 end
 
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -17,10 +17,12 @@
 
 # Utilities for OrderedDict
 function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
-    # TODO: This unnecessarily requires two lookups for k.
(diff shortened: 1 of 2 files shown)


src/aff_expr.jl
```diff
@@ -17,10 +17,12 @@
 
 # Utilities for OrderedDict
 function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
-    # TODO: This unnecessarily requires two lookups for k.
-    # TODO: Decide if we want to drop zeros here after understanding the
-    # performance implications.
-    dict[k] = get!(dict, k, zero(V)) + v
+    # Adding zero terms to this dictionary leads to unacceptable performance
+    # degradations. See, e.g., https://github.com/JuliaOpt/JuMP.jl/issues/1946.
+    if !iszero(v)
```
mlubin: This could be made more clear as:
```julia
if iszero(v)
    return dict # No-op.
end
dict[k] = get!(dict, k, zero(V)) + v
return dict
```

angeris:
I think this is generally expected behavior (vs. not dropping zeros), especially in the case of sparse matrix multiplication :)

Anyways, thanks again !

Issue #1942, opened 2019-04-17
raw_status


blegat:
We have a [`RawStatusString`](http://www.juliaopt.org/MathOptInterface.jl/stable/apireference/#MathOptInterface.RawStatusString) in MOI which can be accessed with `MOI.get(model, MOI.RawStatusString())` but we should have a `raw_status` function similar to `termination_status`/`primal_status`/`dual_status`.

rtwalker:
I'd like to try my hand at this! I'll plan to have something in the next week if it's not too urgent.

blegat:
@rtwalker Sound perfect :)

blegat:
Closed by #1949

PR #1949, opened 2019-04-26
Add raw_status

rtwalker:
My attempt at #1942. 
 
With respect to the tests, I tried to replicate what was done with `TerminationStatus`, but I'm unsure whether I did so successfully.

### DIFF
--- a/docs/src/solutions.md
+++ b/docs/src/solutions.md
@@ -52,6 +52,12 @@ This function will return a `MOI.TerminationStatusCode` `enum`.
 MOI.TerminationStatusCode
 ```
 
+Additionally, we can receive a solver specific string explaning why the
+optimization stopped with
+```julia
+raw_status(model)
+```
+
 ## Solution statuses
 
 These statuses indicate what kind of result is available to be queried
--- a/docs/src/solutions.md
+++ b/docs/src/solutions.md
@@ -53,10 +53,7 @@ MOI.TerminationStatusCode
 ```
 
 Additionally, we can receive a solver specific string explaning why the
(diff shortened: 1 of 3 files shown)


docs/src/solutions.md
```diff
@@ -52,6 +52,12 @@ This function will return a `MOI.TerminationStatusCode` `enum`.
 MOI.TerminationStatusCode
 ```
 
+Additionally, we can receive a solver specific string explaning why the
+optimization stopped with
+```julia
+raw_status(model)
+```
```
odow: It's better to go 
```
optimization stopped with [`raw_status`](@ref).
```

and then add `raw_status` to the reference list at the bottom.

blegat:
> Could you also test calling RawStatusString when it hasn't been set so that it returns nothing?

It would rather have it either return a String or throw OptimizeNotCalled. It should not return `nothing`

blegat:
The issue with the failing tests is due to https://github.com/JuliaOpt/MathOptInterface.jl/issues/707 which will be fixed in https://github.com/JuliaOpt/MathOptInterface.jl/pull/726.
We can wait for MOI v0.9 to be released.

blegat:
@rtwalker You should be able to finish it now as MOI v0.9 has been released and JuMP master has been update to MOI v0.9

rtwalker:
Hi, I just opened this to look at what needed finishing up and saw that this was merged. I guess everything was ok?

blegat:
Yes, I just relaunch CI now that MOI v0.9 was released and it worked :)

PR #1953, opened 2019-05-06
Giving guidance on naming of fields in structs

### LINKED ISSUE #1952: Style Guide makes no recommendation for fields in structs
Your Style Guide is helpful for me as a newcomer. However you could make it even more helpful:
 
``` julia
module SomeModule end
function some_function end
const SOME_CONSTANT = ...
struct SomeStruct end
@enum SomeEnum ENUM_VALUE_A ENUM_VALUE_B
some_local_variable = ...
some_file.jl # Except for ModuleName.jl.
```
does not specify what naming convention to pick for fields.

If you would change it to:
``` julia
module SomeModule end
function some_function end
const SOME_CONSTANT = ...
struct SomeStruct 
   some_field::SomeType
end
@enum SomeEnum ENUM_VALUE_A ENUM_VALUE_B
some_local_variable = ...
some_file.jl # Except for ModuleName.jl.
```

It would be even more helpful

freemin7:
Closes #1952

### DIFF
--- a/docs/src/style.md
+++ b/docs/src/style.md
@@ -387,7 +387,9 @@ end)
 module SomeModule end
 function some_function end
 const SOME_CONSTANT = ...
-struct SomeStruct end
+struct SomeStruct 
+  some_field::SomeType
+end
 @enum SomeEnum ENUM_VALUE_A ENUM_VALUE_B
 some_local_variable = ...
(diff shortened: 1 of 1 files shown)


odow:
Great. Thanks @freemin7

PR #1955, opened 2019-05-07
Fix handling of summations with an empty range

### LINKED ISSUE #1954: Problem with Empty Summations
In the following example the constraint `x == sum(1.0 for i in 1:0)` raises a loading error, `Variable MathOptInterface.VariableIndex(1) is fixed. Cannot also set lower bound.`.  An equivalent form of the constraint `x == 0.0` works fine.

```julia
using JuMP
using Ipopt

m = Model()

@variable(m, 0.0 <= x <= 1.0)

# works fine
# @constraint(m, x == 0.0)

# breaks
@constraint(m, x == sum(1.0 for i in 1:0))

println(m)

optimize!(m, with_optimizer(Ipopt.Optimizer))
```

odow:
Okay, so I thought of a few fixes for this, and I went for the easiest (if the most hacky).

This code should only ever get called for constraints with an (in)equality sense. So `a <= b`, `a >= b`, or `a == b`. (`a in b` goes through a different code path.) Given we have a comparison, then we always want to obtain a `GenericAffExpr`. The easiest way to do this is to just add a dummy `+ 0` term to the expression. 

I haven't benchmarked this, but it should be negligible.

The alternative is to only include this in the `parse_generator_sum` method, and maybe only if the generator has zero terms.

The other argument for doing what I've done here is that there might be other edge-cases (other than `sum` where it might occur. This approach is a panacea.

Closes #1954

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -393,8 +393,6 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense
     # the `lhs` is a `VariableRef` and the `rhs` is a summation with no terms.
     # Without the `+0` term, `aff` would evaluate to a `VariableRef` when we
     # really want it to be a `GenericAffExpr`.
-    #
-    # See JuMP issue #1954 for more information.
     aff = :($lhs - $rhs + 0)
     set = sense_to_set(_error, sense)
     parse_one_operator_constraint(_error, vectorized, Val(:in), aff, set)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -387,8 +387,15 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
 end
 
 function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Val, lhs, rhs)
-    # Simple comparison - move everything to the LHS
-    aff = :($lhs - $rhs)
+    # Simple comparison - move everything to the LHS.
+    #
+    # Note: We add the +0 to this term to account for the diabolical case that
+    # the `lhs` is a `VariableRef` and the `rhs` is a summation with no terms.
+    # Without the `+0` term, `aff` would evaluate to a `VariableRef` when we
+    # really want it to be a `GenericAffExpr`.
+    #
+    # See JuMP issue #1954 for more information.
+    aff = :($lhs - $rhs + 0)
     set = sense_to_set(_error, sense)
     parse_one_operator_constraint(_error, vectorized, Val(:in), aff, set)
 end
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -389,7 +389,7 @@ end
 function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Val, lhs, rhs)
     # Simple comparison - move everything to the LHS.
     #
-    # Note: We add the +0 to this term to account for the diabolical case that
+    # Note: We add the +0 to this term to account for the pathological case that
     # the `lhs` is a `VariableRef` and the `rhs` is a summation with no terms.
     # Without the `+0` term, `aff` would evaluate to a `VariableRef` when we
     # really want it to be a `GenericAffExpr`.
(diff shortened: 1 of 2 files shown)


blegat:
If we always want it to be an expression and not a VariableRef, we can do
```julia
functionize(f::SingleVariable) = convert(AffExpr, f)
functionize(f::Any) = f
...
aff = :(functionize($lhs - $rhs))
```

ccoffrin:
I would also recommend adding this test to avoid any regressions on the NL side,
```
    @testset "Empty summation in @NLconstraints" begin
        model = Model()
        @variable(model, x)
        c = @NLconstraint(model, x == sum(1.0 for i in 1:0))
    end
```

odow:
So I've added the NL test. But going down @blegat's route got complicated since it makes the constraints parse differently.

odow:
There is no change. 

Current master
```julia
Oscars-MBP:JuMP oscar$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
Oscars-MBP:JuMP oscar$ julia test/perf/speed.jl
P-Median(100 facilities, 100 customers, 5000 locations) benchmark:
BenchmarkTools.Trial: 
  memory estimate:  714.13 MiB
  allocs estimate:  12025585
  --------------
  minimum time:     1.491 s (51.07% GC)
  median time:      1.679 s (54.64% GC)
  mean time:        1.664 s (54.87% GC)
  maximum time:     1.808 s (58.46% GC)
  --------------
  samples:          4
  evals/sample:     1
Cont5(n=500) benchmark:
BenchmarkTools.Trial: 
  memory estimate:  402.78 MiB
  allocs estimate:  6785862
  --------------
  minimum time:     682.780 ms (37.28% GC)
  median time:      834.134 ms (49.49% GC)
  mean time:        797.730 ms (45.13% GC)
  maximum time:     900.495 ms (49.59% GC)
  --------------
  samples:          7
  evals/sample:     1
```

This PR
```Julia
Oscars-MBP:JuMP oscar$ git checkout od/fix_summation
Switched to branch 'od/fix_summation'
Your branch is up to date with 'origin/od/fix_summation'.
Oscars-MBP:JuMP oscar$ julia test/perf/speed.jl
P-Median(100 facilities, 100 customers, 5000 locations) benchmark:
BenchmarkTools.Trial: 
  memory estimate:  714.13 MiB
  allocs estimate:  12025585
  --------------
  minimum time:     1.484 s (51.12% GC)
  median time:      1.699 s (54.34% GC)
  mean time:        1.680 s (54.58% GC)
  maximum time:     1.838 s (57.81% GC)
  --------------
  samples:          4
  evals/sample:     1
Cont5(n=500) benchmark:
BenchmarkTools.Trial: 
  memory estimate:  402.78 MiB
  allocs estimate:  6785862
  --------------
  minimum time:     710.835 ms (37.76% GC)
  median time:      848.776 ms (49.96% GC)
  mean time:        809.624 ms (45.11% GC)
  maximum time:     894.192 ms (49.03% GC)
  --------------
  samples:          7
  evals/sample:     1
```

blegat:
> There is no change.

Good! That means it is modifying the expression in-place by adding `0.0` to the constant which has negligible effect on performance.

blegat:
Could you give the effect of this PR to the `speed.jl` benchmark ?

src/macros.jl
```diff
@@ -387,8 +387,15 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
...
-    # Simple comparison - move everything to the LHS
-    aff = :($lhs - $rhs)
+    # Simple comparison - move everything to the LHS.
+    #
+    # Note: We add the +0 to this term to account for the pathological case that
+    # the `lhs` is a `VariableRef` and the `rhs` is a summation with no terms.
+    # Without the `+0` term, `aff` would evaluate to a `VariableRef` when we
+    # really want it to be a `GenericAffExpr`.
+    #
+    # See JuMP issue #1954 for more information.
```
mlubin: There's actually not much extra information in the issue discussion that's not already covered by the comment above, so this reference could be removed.

PR #1959, opened 2019-05-12
Add Project.toml and remove REQUIRE files

blegat:
See https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/159.
Generated by https://github.com/JuliaLang/Pkg.jl/blob/master/bin/gen_project.jl

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "JuMP"
 uuid = "4076af6c-e467-56ae-b986-b466b2749572"
 repo = "https://github.com/JuliaOpt/JuMP.jl.git"
-version = "0.18.0"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -0,0 +1,33 @@
+name = "JuMP"
+uuid = "4076af6c-e467-56ae-b986-b466b2749572"
+repo = "https://github.com/JuliaOpt/JuMP.jl.git"
+version = "0.18.0"
```
frapac: I do not understand why the version is `0.18` instead of `0.19`?

blegat: Ups :-P

PR #1967, opened 2019-05-17
Add SCIP

blegat:
cc @rschwarz


rschwarz:
Cool, thanks!

We might want to remove the `LP`, as SCIP does not give proper access to dual solution values, but other than that, this is correct.

PR #1968, opened 2019-05-17
Add CDCS

blegat:
cc @giofantuzzi @zhengy09


mlubin:
CDCS.jl doesn't appear to be a registered package.

blegat:
It is now registered :)

Issue #1965, opened 2019-05-17
set_parameter


blegat:
Add a function `set_parameter` that sets the `RawParameter` attribute that will be added in MOI v0.9: https://github.com/JuliaOpt/MathOptInterface.jl/pull/733

mlubin:
Once this is in place, we should strongly consider removing `with_optimizer` (and adding any necessary helper functions to make the transition easier).

blegat:
Closed by #1974

PR #1974, opened 2019-05-23
adding set_parameter function and tests

raphaelsaavedra:
My attempt at #1965. 

@blegat is this what you wanted?

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -648,14 +648,7 @@ end
 Return the value of the attribute `attr` from the model's MOI backend.
 """
 function MOI.get(model::Model, attr::MOI.AbstractOptimizerAttribute)
-    if MOI.is_set_by_optimize(attr) &&
-       !(attr isa MOI.TerminationStatus) && # Before `optimize!` is called, the
-       !(attr isa MOI.PrimalStatus) &&      # statuses are `OPTIMIZE_NOT_CALLED`
-       !(attr isa MOI.DualStatus)           # and `NO_SOLUTION`
-        _moi_get_result(backend(model), attr)
-    else
-        MOI.get(backend(model), attr)
-    end
+    MOI.get(backend(model), attr)
 end
 function MOI.get(model::Model, attr::MOI.AbstractVariableAttribute,
                  v::VariableRef)
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -451,6 +451,15 @@ function solve_time(model::Model)
 end
 
 """
+    set_parameter(model::Model, name, value)
+
+Sets solver-specific parameter identified by `name` to `value`.
+"""
+function set_parameter(model::Model, name, value)
+    return MOI.set(model, MOI.RawParameter(name), value)
+end
+
+"""
     set_silent(model::Model)
 
 Takes precedence over any other attribute controlling verbosity 
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -642,6 +642,21 @@ function MOI.get(model::Model, attr::MOI.AbstractModelAttribute)
         MOI.get(backend(model), attr)
     end
 end
+"""
+    get(model::Model, attr::MathOptInterface.AbstractOptimizerAttribute)
+
+Return the value of the attribute `attr` from the model's MOI backend.
+"""
+function MOI.get(model::Model, attr::MOI.AbstractOptimizerAttribute)
+    if MOI.is_set_by_optimize(attr) &&
+       !(attr isa MOI.TerminationStatus) && # Before `optimize!` is called, the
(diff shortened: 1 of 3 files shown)


raphaelsaavedra:
I believe it is erroring because `RawParameter` wasn't added to MOI yet. Testing with MOI's latest versions returns no errors.

test/model.jl
```diff
@@ -367,6 +367,13 @@ function test_model()
         @test !JuMP.unset_silent(model)
         @test !MOI.get(backend(model), MOI.Silent())
     end
+
+    @testset "set_parameter" begin
+        mock = MOIU.UniversalFallback(JuMP._MOIModel{Float64}())
+        model = Model(with_optimizer(MOIU.MockOptimizer, mock))
+        @test JuMP.set_parameter(model, "aaa", "bbb") == "bbb"
+        @test MOI.get(backend(model), MOI.RawParameter("aaa")) == "bbb"
```
blegat: Can you test that `MOI.get(model, ...)` also works ?

src/JuMP.jl
```diff
@@ -450,6 +450,15 @@ function solve_time(model::Model)
     return MOI.get(model, MOI.SolveTime())
 end
 
+"""
+    set_parameter(model::Model, name, value)
+
+Sets solver-specific parameter identified by `name` to `value`.
+"""
+function set_parameter(model::Model, name, value)
```
blegat: Add it do the documentation besides `set_silent`

test/model.jl
raphaelsaavedra: It does not, I think it's because there's no `MOI.get(model::Model, attr::MOI.AbstractOptimizerAttribute)`. There's only for model, variable, and constraint.

If you want, I can add the following function:

```
"""
    get(model::Model, attr::MathOptInterface.AbstractOptimizerAttribute)

Return the value of the attribute `attr` from the model's MOI backend.
"""
function MOI.get(model::Model, attr::MOI.AbstractOptimizerAttribute)
    if MOI.is_set_by_optimize(attr) &&
       !(attr isa MOI.TerminationStatus) && # Before `optimize!` is called, the
       !(attr isa MOI.PrimalStatus) &&      # statuses are `OPTIMIZE_NOT_CALLED`
       !(attr isa MOI.DualStatus)           # and `NO_SOLUTION`
        _moi_get_result(backend(model), attr)
    else
        MOI.get(backend(model), attr)
    end
end
```

This seems to work.

src/JuMP.jl
raphaelsaavedra: Done.

test/model.jl
raphaelsaavedra: Done in f6a9164d66781d554527f6b1c97644e3323a70cd, please tell me if it's okay

raphaelsaavedra: Added 4b2f34ca123bf88357bd6a06d2c4b1e6c53d620b because if we're doing this might as well add the tests for `Silent` too

src/JuMP.jl
```diff
@@ -633,6 +642,21 @@ function MOI.get(model::Model, attr::MOI.AbstractModelAttribute)
...
     end
 end
+"""
+    get(model::Model, attr::MathOptInterface.AbstractOptimizerAttribute)
+
+Return the value of the attribute `attr` from the model's MOI backend.
+"""
+function MOI.get(model::Model, attr::MOI.AbstractOptimizerAttribute)
+    if MOI.is_set_by_optimize(attr) &&
+       !(attr isa MOI.TerminationStatus) && # Before `optimize!` is called, the
```
blegat: You can remove this because `MOI.TerminationStatus` is not an optimizer attribute

raphaelsaavedra: True. 30f020c7a71eb2357af7117fa2e46b5afff4981a should be fine now.

blegat:
@raphaelsaavedra  You should be able to finish it now as MOI v0.9 has been released and JuMP master has been update to MOI v0.9

raphaelsaavedra:
Sure @blegat, what else do you need me to do?

test/model.jl
```diff
@@ -364,8 +364,18 @@ function test_model()
...
         @test JuMP.set_silent(model)
         @test MOI.get(backend(model), MOI.Silent())
+        @test MOI.get(model, MOI.Silent())
         @test !JuMP.unset_silent(model)
         @test !MOI.get(backend(model), MOI.Silent())
+        @test !MOI.get(model, MOI.Silent())
+    end
+
+    @testset "set_parameter" begin
+        mock = MOIU.UniversalFallback(JuMP._MOIModel{Float64}())
```
blegat: `JuMP._MOIModel` should be replaced by `MOIU.Model`

raphaelsaavedra: FIxed it in 11f047af2f8422640fbae67c12bdedb21a743e48

PR #1975, opened 2019-05-24
add Project.toml so we can tag 0.18.6

mlubin:
Given that we recommend users to stick with 0.18 until callbacks are supported, we need to tag 0.18.6 to support Julia 1.2.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,7 +16,7 @@ ForwardDiff = "~0.5.0, ~0.6, ~0.7, ~0.8, ~0.9, ~0.10"
 MathProgBase = "~0.6.0, ~0.7"
 OffsetArrays = "≥ 0.2.13"
 ReverseDiffSparse = "~0.8.0"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -0,0 +1,32 @@
...
+MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
+ReverseDiffSparse = "89212889-6d3f-5f97-b412-7825138f6c9c"
+
+[compat]
+Compat = "≥ 1.0.0"
+ForwardDiff = "~0.5.0, ~0.6, ~0.7, ~0.8, ~0.9, ~0.10"
+MathProgBase = "~0.6.0, ~0.7"
+OffsetArrays = "≥ 0.2.13"
+ReverseDiffSparse = "~0.8.0"
+julia = "≥ 0.6.0"
```
blegat: I would rather do "0.6, 1` so that we do not say that we support Julia v2

PR #1977, opened 2019-05-26
Add nice error message for splatting in macros

### LINKED ISSUE #1964: Splatting does not work inside [] for variable declarations
Write
```Julia
using JuMP
using Gurobi

model = Model(with_optimizer(Gurobi.Optimizer))

c = rand(2,3)
```
now the following
```Julia
julia> @variable(model, X[axes(c)...])
ERROR: syntax: "..." expression outside call
```
does not work, while
```Julia
julia> @variable(model, X[axes(c,1), axes(c,2)])
2×3 Array{VariableRef,2}:
 X[1,1]  X[1,2]  X[1,3]
 X[2,1]  X[2,2]  X[2,3]
```
does.

odow:
Ref #1964. <s>This adds a nice error message, but doesn't enable the functionality, so I haven't gone `Closing #1964`.</s>

Closes #1964

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -83,13 +83,15 @@ Helper function for macros to construct container objects. Takes an `Expr` that 
 """
 _build_ref_sets(c) = _build_ref_sets(c, _get_name(c))
 
-function _expr_contains_splat(ex::Expr)
+function _expr_is_splat(ex::Expr)
     if ex.head == :(...)
         return true
+    elseif ex.head == :escape
+        return _expr_is_splat(ex.args[1])
     end
-    return any(_expr_contains_splat.(ex.args))
+    return false
 end
-_expr_contains_splat(::Any) = false
+_expr_is_splat(::Any) = false
 
 """
     JuMP._get_looped_code(varname, code, condition, idxvars, idxsets, sym, requestedcontainer::Symbol; lowertri=false)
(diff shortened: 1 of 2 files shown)


mlubin:
The recursive check might be too aggressive. Does it also error at `x[some_index_set_generator(args...)]`?

odow:
> Does it also error at x[some_index_set_generator(args...)]?

D'oh. Much simpler...

mlubin:
To clarify, `x[some_index_set_generator(args...)]` should not be an error. The current recursive check for ... would error in this case, right? I think the right check is if any of the index sets are exactly splatting expressions, without a recursive dive into the expression.

src/macros.jl
```diff
@@ -592,6 +602,9 @@ function _constraint_macro(args, macro_name::Symbol, parsefun::Function)
     # Strategy: build up the code for add_constraint, and if needed
     # we will wrap in loops to assign to the ConstraintRefs
     refcall, idxvars, idxsets, condition = _build_ref_sets(c, variable)
+    if any(_expr_is_splat.(idxsets))
+        _error("cannot use splatting operator `...`.")
```
mlubin: Nit: I would clarify on the error message that you can't use the splatting operator *in the definition of an index set*. There are other cases where splatting works inside a macro.

PR #1978, opened 2019-05-26
Error on complex values in NLP

### LINKED ISSUE #1973: InexactError: Float64 when problem gets big
I'm running an optimization problem on some simulated data:

```julia
using LinearAlgebra, JuMP, Ipopt

d = 10
n = 1000
A = I + 0.5*(rand(d,d)-0.5*ones(d,d))
A = Symmetric(A)
J = sqrt(A)
eps = randn(d,n)
z = J*eps

m = Model(with_optimizer(Ipopt.Optimizer))
@variable(m, theta >= 0, start = 1.0)
@variable(m, beta[1:d-1], start = 0.0)
@NLobjective(m, Min, theta/n*sum( (z[1,j] + sum(beta[i-1] * z[i,j] for i=2:d))^2 for j=1:n) - log(theta))

optimize!(m)
```

This code runs great when d=10, but change to d=100, and I get the following:

```
InexactError: Float64(0.3362800773416026 + 0.03631914074869952im)

Stacktrace:
 [1] Type at .\complex.jl:37 [inlined]
 [2] convert at .\number.jl:7 [inlined]
 [3] push!(::Array{Float64,1}, ::Complex{Float64}) at .\array.jl:853
 [4] _parse_NL_expr_runtime(::Model, ::Complex{Float64}, ::Array{JuMP._Derivatives.NodeData,1}, ::Int64, ::Array{Float64,1}) at C:\Users\<Username>\.julia\packages\JuMP\tyMag\src\parse_nlp.jl:171
 [5] top-level scope at C:\Users\<Username>\.julia\packages\JuMP\tyMag\src\parse_expr.jl:358
```
Not sure what is going on here.

odow:
Closes #1973

Related:
```Julia
model = Model()
@variable(model, x >= 1im)
# ERROR: InexactError: Float64(0 + 1im)
```
Probably in lots of other places as well. What do we want to do about this?

### DIFF
--- a/src/parse_nlp.jl
+++ b/src/parse_nlp.jl
@@ -167,11 +167,7 @@ function _parse_NL_expr(m, x, tapevar, parent, values)
 
 end
 
-function _parse_NL_expr_runtime(m::Model, x::Complex, tape, parent, values)
-    error("Complex numbers are not supported in nonlinear expressions")
-end
-
-function _parse_NL_expr_runtime(m::Model, x::Number, tape, parent, values)
+function _parse_NL_expr_runtime(m::Model, x::Real, tape, parent, values)
     push!(values, x)
     push!(tape, NodeData(VALUE, length(values), parent))
     nothing
--- a/src/parse_nlp.jl
+++ b/src/parse_nlp.jl
@@ -167,6 +167,10 @@ function _parse_NL_expr(m, x, tapevar, parent, values)
 
 end
 
+function _parse_NL_expr_runtime(m::Model, x::Complex, tape, parent, values)
+    error("Complex numbers are not supported in nonlinear expressions")
+end
+
 function _parse_NL_expr_runtime(m::Model, x::Number, tape, parent, values)
     push!(values, x)
     push!(tape, NodeData(VALUE, length(values), parent))
--- a/test/nlp.jl
+++ b/test/nlp.jl
@@ -619,7 +619,7 @@
         evaluator = JuMP.NLPEvaluator(model)
         @test !(:Hess in MOI.features_available(evaluator))
     end
-    
+
     @testset "Error on using AffExpr in NLexpression" begin
         model = Model()
         @variable(model, x)
@@ -632,7 +632,7 @@
         )
         @test_throws expected_exception @NLexpression(model, A)
     end
-    
+
     @testset "Error on using QuadExpr in NLexpression" begin
         model = Model()
         @variable(model, x)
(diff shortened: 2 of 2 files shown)


src/parse_nlp.jl
```diff
@@ -167,6 +167,10 @@ function _parse_NL_expr(m, x, tapevar, parent, values)
 
 end
 
+function _parse_NL_expr_runtime(m::Model, x::Complex, tape, parent, values)
+    error("Complex numbers are not supported in nonlinear expressions")
+end
+
 function _parse_NL_expr_runtime(m::Model, x::Number, tape, parent, values)
```
mlubin: Can we achieve the same result by changing `Number` to `Real` in this signature? There's a fallback below (`_parse_NL_expr_runtime(m::Model, x, tape, parent, values)`) that would catch complex numbers then.

mlubin:
> Probably in lots of other places as well. What do we want to do about this?

Check all values that come into a macro, and throw an error if they match a condition that we know will throw an unhelpful internal error message (e.g., not being convertible to `Float64`). However, doing this well requires a bit of a refactoring of the macros so perhaps not high priority for now.

Bondan000:
hi,
should this merge solve the problem with the complex numbers? i am using JuMP v0.20.1 and Julia v1.2.0 and still having this problem:
` julia> mdl=Model()`
`julia> @variable(mdl, x >= 1*im)`
`ERROR: InexactError: Float64(0 + 1im)`
is there any work around to use complex numbers while defining expressions?

odow:
> should this merge solve the problem with the complex numbers

No. JuMP does not support complex numbers.

> is there any work around to use complex numbers while defining expressions?

Define two variables, one for the real part and one for the imaginary part

```Julia
model = Model()
@variable(model, x_real >= 0)
@variable(model, x_imag >= 1)
```
See https://discourse.julialang.org/t/how-define-a-complex-jump-variable/14362

Bondan000:
> should this merge solve the problem with the complex numbers
> 
> No. JuMP does not support complex numbers.
> 
> is there any work around to use complex numbers while defining expressions?
> 
> Define two variables, one for the real part and one for the imaginary part
> model = Model()
> @variable(model, x_real >= 0)
> @variable(model, x_imag >= 1)
> See https://discourse.julialang.org/t/how-define-a-complex-jump-variable/14362

Thank you for the quick reply! will go through the discourse and try to solve my specific problem!

PR #1979, opened 2019-06-01
Fix a bug in subexpression handling

mlubin:
Subexpressions in some cases were not correctly linked to the constraints that they appeared in via a transitive dependency (chain of subexpressions). This could cause incorrect derivative computations for models with this structure.

Thanks to @sanderclaeys for providing the test case.

The change includes style fixes, so the diff is larger than the bug fix itself. The bug has been around since JuMP 0.12.

### DIFF
--- a/src/_Derivatives/subexpressions.jl
+++ b/src/_Derivatives/subexpressions.jl
@@ -15,56 +15,69 @@ function list_subexpressions(nd::Vector{NodeData})
     return sort(collect(indices))
 end
 
-# order the subexpressions which main_expressions depend on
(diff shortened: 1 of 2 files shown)


ccoffrin:
CC @kaarthiksundar

odow:
@ccoffrin is this potentially the cause of your "AMPLNLWriter != JuMP" bugs?

ccoffrin:
@odow maybe, worth revisiting that case.  If I recall correctly in that case JuMP == Gravity, but NL != Gravity / JuMP, which suggests there was no issue in JuMP there.

PR #1980, opened 2019-06-01
remove unused argument in _hessian_slice_inner

mlubin:


blegat:
We might as well give the `breaking` label to avoid including it in v0.19.2 even if it only changes an internal function

mlubin:
> We might as well give the breaking label to avoid including it in v0.19.2 even if it only changes an internal function

Changes/refactoring of internal functions should not be included in the definition of breaking.

PR #1993, opened 2019-06-20
Implement algebra with UniformScaling

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1989
Also fixes https://github.com/JuliaLang/julia/issues/32374 for JuMP variables

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -9,52 +9,58 @@
 #############################################################################
 
 const _JuMPTypes = Union{AbstractJuMPScalar, NonlinearExpression}
+const Constant = Union{Number, UniformScaling}
+_float(x::Number) = convert(Float64, x)
+_float(J::UniformScaling) = _float(J.λ)
 
 # Overloads
 #
 # Different objects that must all interact:
-# 1. Number
+# 1. Constant
 # 2. AbstractVariableRef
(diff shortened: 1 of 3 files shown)


src/operators.jl
```diff
@@ -576,6 +582,16 @@ function Base.:-(x::AbstractArray{T}) where {T <: _JuMPTypes}
     return ret
 end
 
+# Fix https://github.com/JuliaLang/julia/issues/32374 as done in
+# https://github.com/JuliaLang/julia/pull/32375. This hack should
+# be removed once we drop Julia v1.0.
+function Base.:-(A::Symmetric{<:JuMP.AbstractVariableRef})
```
mlubin: It would be nice to have a comment on https://github.com/JuliaLang/julia/pull/32375 that this is the right fix, or even better have the PR merged before we patch it into JuMP.

blegat: Even if it is merged, it will be long before we can assume that the fix is in people binaries. I agree that we could wait for a review from the Julia PR.

blegat: It is merged

blegat: It is merged

mlubin: LGTM. Please fix the merge conflicts.

PR #1995, opened 2019-06-20
Refactor matmul code with mul!

blegat:
The new implementation catches `mul!` calls for which the eltype of the result is a JuMP affine or quadratic expressions (resolving https://github.com/JuliaOpt/JuMP.jl/issues/1990).
This allows the generic codes using `mul!` to benefit from our specialized multiplication methods.
For SparseArrays, the situation is less ideal as some calls are redirected to `mul!` and some have custom implementation that do not work for JuMP as they start by promoting the arrays to the same eltype...
This PR redirects some of them so that all tests pass but more are needed (more tests and more redirections) but this is left for future work. I don't think there is any regression as we already had this issue: https://github.com/JuliaOpt/JuMP.jl/issues/1276

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1990
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1716

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -405,8 +405,8 @@ function _mul!(ret::AbstractVecOrMat{<:GenericAffOrQuadExpr},
     size(A, 2) == size(ret, 1) || throw(DimensionMismatch())
     size(A, 1) == size(B, 1) || throw(DimensionMismatch())
     size(B, 2) == size(ret, 2) || throw(DimensionMismatch())
-    nzv = nonzeros(A)
-    rv = rowvals(A)
+    A_nonzeros = nonzeros(A)
+    A_rowvals = rowvals(A)
     # See SparseArrays/src/linalg.jl
     if !isone(β)
         !iszero(β) ? rmul!(ret, β) : _fill_with_zeros!(ret)
@@ -416,7 +416,7 @@ function _mul!(ret::AbstractVecOrMat{<:GenericAffOrQuadExpr},
         @inbounds for col ∈ 1:A.n
             tmp = zero(eltype(ret))
             for j ∈ A.colptr[col]:(A.colptr[col + 1] - 1)
-                add_to_expression!(tmp, adjoint(nzv[j]), B[rv[j], k])
+                add_to_expression!(tmp, adjoint(A_nonzeros[j]), B[A_rowvals[j], k])
             end
             add_to_expression!(ret[col, k], α, tmp)
         end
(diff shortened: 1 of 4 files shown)


src/operators.jl
```diff
@@ -328,119 +328,44 @@ end
...
     for I in eachindex(A)
-        A[I] = zero(T)
+        A[I] = zero(eltype(A))
     end
     return A
 end
 
 ###############################################################################
-# `_A_mul_B!(ret, A, B)` adds the result of `A*B` into the buffer `ret`. We roll our own
+# `_mul!(ret, A, B)` adds the result of `A*B` into the buffer `ret`. We roll our own
```
mlubin: s/adds/sets/. The difference is `ret = A*B` vs `ret += A*B`.

src/operators.jl
```diff
@@ -467,78 +392,188 @@ end
...
+    A = parent(adjA)
+    size(A, 2) == size(ret, 1) || throw(DimensionMismatch())
+    size(A, 1) == size(B, 1) || throw(DimensionMismatch())
     size(B, 2) == size(ret, 2) || throw(DimensionMismatch())
-    nzv = A.nzval
-    rv = A.rowval
-    # `ret` is already filled with zeros by `_return_arrayt`.
-    for k = 1:size(ret, 2)
-        @inbounds for col = 1:A.n
+    nzv = nonzeros(A)
```
mlubin: These names don't follow JuMP conventions: `A_nonzeros`, `A_rowvals` etc

src/operators.jl
```diff
@@ -467,78 +392,188 @@ end
 > OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 > WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 =#
-function _At_mul_B!(ret::StridedVecOrMat{<:GenericAffOrQuadExpr}, A::SparseMatrixCSC, B::StridedVecOrMat)
-    A.n == size(ret, 1) || throw(DimensionMismatch())
-    A.m == size(B, 1) || throw(DimensionMismatch())
+
+const TrAdj{T<:Union{GenericAffOrQuadExpr, Real}, MT} = Union{Transpose{T, MT}, Adjoint{T, MT}}
```
mlubin: TrAdj is not a meaningful name (is it transposed adjoint?). What about `TransposeOrAdjoint`?

src/operators.jl
```diff
@@ -467,78 +392,188 @@ end
...
+    T = _A_mul_B_eltype(eltype(A), eltype(B))
+    ret = Array{T}(undef, _A_mul_B_ret_dims(A, B))
+    return mul!(ret, A, B)
 end
 
-function Base.:*(A::Union{Matrix, SparseMatrixCSC},
-                 B::Union{Matrix{<:_JuMPTypes}, Vector{<:_JuMPTypes}, SparseMatrixCSC{<:_JuMPTypes}})
-    return _A_mul_B(A, B)
-end
+# A few a overwritten below but many more need to be redirected to `_mul` in
```
mlubin: s/a/are/

src/operators.jl
blegat: Good point

src/operators.jl
chriscoey: The name is pretty tradj

PR #2000, opened 2019-06-26
Improve NLP documentation about ForwardDiff

odow:
Addresses this Discourse post: https://discourse.julialang.org/t/dual-forwarddiff-type-error-when-jump-tries-autodiff-user-defined-function/25639

### DIFF
--- a/docs/src/nlp.md
+++ b/docs/src/nlp.md
@@ -196,11 +196,13 @@ for a description of how to write a function suitable for automatic
 differentiation.
 
 !!! note
-    The general guideline for writing ForwardDiff-compatible code is is to write
-    code that is generic with respect to the number type; don't assume that the 
-    input to the function is `Float64`.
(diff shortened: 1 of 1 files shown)


docs/src/nlp.md
```diff
@@ -193,9 +193,19 @@ JuMP uses [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl) to
 perform automatic differentiation; see the ForwardDiff.jl
 [documentation](http://www.juliadiff.org/ForwardDiff.jl/v0.10.2/user/limitations.html)
 for a description of how to write a function suitable for automatic
-differentiation. The general guideline is to write code that is generic with
-respect to the number type; don't assume that the input to the function is
-`Float64`. To register a user-defined function with derivatives computed by
+differentiation.
+
+!!! note
```
mlubin: It might be more useful to word this as, "If you see MethodErrors with ForwardDiff.Duals, then see guidelines at X, Y, Z."

PR #2003, opened 2019-07-08
DNMY: Update to MOI v0.9

coroa:
Might be a bit premature, since not all solvers have been updated (see https://github.com/JuliaOpt/MathOptInterface.jl/issues/736), but I wanted to play with https://github.com/JuliaOpt/Gurobi.jl/pull/216 .

All tests pass. ~~, except for https://github.com/coroa/JuMP.jl/blob/c8784cf3e9fe5936e652139a5ae73caa48d6cfb4/test/print.jl#L351-L509 , where constraints appear with a different order; and I do not know, whether that is meant to stay!?~~ . 

All doc-tests pass as well.

All examples run fine.

## Next steps
- [x] Wait for MOI and solvers to tag compatible versions, then
- [x] Remove `Manifest.toml` from top-level, `docs` and `examples`
- [x] Remove MOI from `Project.toml` files
- [x] Rebased on `master`

### DIFF
--- a/examples/diet.jl
+++ b/examples/diet.jl
@@ -95,7 +95,7 @@ function example_diet(; verbose = true)
     @constraint(model, buy["milk"] + buy["ice cream"] <= 6)
     verbose && println("Solving dairy-limited problem...")
     JuMP.optimize!(model)
-    @test JuMP.termination_status(model) == MOI.INFEASIBLE_OR_UNBOUNDED
+    @test JuMP.termination_status(model) == MOI.INFEASIBLE
     @test JuMP.primal_status(model) == MOI.NO_SOLUTION
     verbose && print_solution(false, foods, buy)
 end
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -524,7 +524,7 @@ end
 function _moi_optimizer_index(model::MOI.Bridges.LazyBridgeOptimizer,
                               index::MOI.Index)
     if index isa MOI.ConstraintIndex &&
-        MOI.Bridges.is_bridged(model, typeof(index))
+        MOI.Bridges.is_bridged(model, index)
         error("There is no `optimizer_index` for $(typeof(index)) constraints",
               " because they are bridged.")
     else
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -50,21 +50,6 @@ const _MOIFIX = _MOICON{MOI.SingleVariable,MOI.EqualTo{Float64}}
 const _MOIINT = _MOICON{MOI.SingleVariable,MOI.Integer}
 const _MOIBIN = _MOICON{MOI.SingleVariable,MOI.ZeroOne}
 
-MOIU.@model(_MOIModel,
-            (MOI.ZeroOne, MOI.Integer),
-            (MOI.EqualTo, MOI.GreaterThan, MOI.LessThan, MOI.Interval),
-            (MOI.Zeros, MOI.Nonnegatives, MOI.Nonpositives, MOI.SecondOrderCone,
-             MOI.RotatedSecondOrderCone, MOI.GeometricMeanCone,
-             MOI.PositiveSemidefiniteConeTriangle,
-             MOI.PositiveSemidefiniteConeSquare,
-             MOI.RootDetConeTriangle, MOI.RootDetConeSquare,
-             MOI.LogDetConeTriangle, MOI.LogDetConeSquare),
-            (),
-            (),
-            (MOI.ScalarAffineFunction, MOI.ScalarQuadraticFunction),
-            (MOI.VectorOfVariables,),
-            (MOI.VectorAffineFunction, MOI.VectorQuadraticFunction))
-
 """
     OptimizerFactory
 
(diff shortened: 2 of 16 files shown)


odow:
As long as they are all there, I don't think the order matters. I would vote to update. It is likely because of the re-jig of the `MOI.@model` backend.

src/JuMP.jl
```diff
@@ -59,7 +60,7 @@ MOIU.@model(_MOIModel,
              MOI.RootDetConeTriangle, MOI.RootDetConeSquare,
              MOI.LogDetConeTriangle, MOI.LogDetConeSquare),
             (),
-            (MOI.SingleVariable,),
+            (),
```
blegat: You can get rid of this and use `MOIU.Model` instead of `_MOIModel`

blegat:
`SingleVariable` constraints are now treated differently because their index should be equal to the variable index so the order has changed for these constraints.

src/JuMP.jl
coroa: Great, that works.

mlubin:
This is great but we can't merge it until there's an MOI release and necessary solvers tagged so that Travis passes.

examples/diet.jl
```diff
@@ -95,7 +95,7 @@ function example_diet(; verbose = true)
     @constraint(model, buy["milk"] + buy["ice cream"] <= 6)
     verbose && println("Solving dairy-limited problem...")
     JuMP.optimize!(model)
-    @test JuMP.termination_status(model) == MOI.INFEASIBLE_OR_UNBOUNDED
+    @test JuMP.termination_status(model) == MOI.INFEASIBLE
     @test JuMP.primal_status(model) == MOI.NO_SOLUTION
```
coroa: @odow It's correct that GLPK now returns a more specialized status code, isn't it?

odow: Yes. I clarified a few of the statuses.

blegat:
MOI v0.9 is released and solvers used for the examples have been updated and released so you should be able to finalize the PR :)

test/nonnegative_bridge.jl
```diff
@@ -44,7 +45,7 @@ function MOI.supports_constraint(::Type{NonnegativeBridge{T}},
                                  ::Type{Nonnegative}) where T
     return true
 end
-function MOIBC.added_constraint_types(::Type{NonnegativeBridge{T, F}}) where {T, F}
+function MOIB.added_constraint_types(::Type{NonnegativeBridge{T, F}}) where {T, F}
```
coroa: Seems to work, but I didn't check, whether that's correct.

coroa:
I am not familiar enough with bridges to fix the new errors in the bridging test. Any hints?

blegat:
`typeof(index)` should be replaced by `index` here:
https://github.com/JuliaOpt/JuMP.jl/blob/0565f26eb5aca10c92516143f8a40280257b49b9/src/JuMP.jl#L541
You also need to add
```julia
function MOIB.added_constraint_types(::Type{NonnegativeBridge{T, F}}) where {T, F}
    return [(F, MOI.GreaterThan{T})]
end
```
just before this line:
https://github.com/JuliaOpt/JuMP.jl/blob/0565f26eb5aca10c92516143f8a40280257b49b9/test/nonnegative_bridge.jl#L43
You also need to replace `add_scalar_constraint` here:
https://github.com/JuliaOpt/JuMP.jl/blob/0565f26eb5aca10c92516143f8a40280257b49b9/test/nonnegative_bridge.jl#L34
and here:
https://github.com/JuliaOpt/JuMP.jl/blob/0565f26eb5aca10c92516143f8a40280257b49b9/test/nonnegative_bridge.jl#L50
with `normalize_and_add_constraint`.

test/nonnegative_bridge.jl
blegat: It's correct

coroa:
> You also need to add
> [...]

You meant:
```julia
function MOIB.added_constrained_variable_types(::Type{NonnegativeBridge{T, F}}) where {T, F}
    return [(F, MOI.GreaterThan{T})]
end
```
didn't you?

blegat:
Yes, I meant:
```julia
function MOIB.added_constrained_variable_types(::Type{<:NonnegativeBridge})
    return Tuple{DataType}[]
end
```

test/nonnegative_bridge.jl
```diff
@@ -45,14 +45,18 @@ function MOI.supports_constraint(::Type{NonnegativeBridge{T}},
                                  ::Type{Nonnegative}) where T
     return true
 end
+
+function MOIB.added_constrained_variable_types(::Type{NonnegativeBridge{T, F}}) where {T, F}
+    return []
```
blegat: Use `Tuple{DataType}[]`

blegat:
You should add the following methods in JuMP: https://github.com/JuliaOpt/MathOptInterface.jl/issues/844
This is type piracy but it will be removed once we release MOI v0.9.1

test/nonnegative_bridge.jl
```diff
@@ -40,14 +45,18 @@ function MOI.supports_constraint(::Type{NonnegativeBridge{T}},
                                  ::Type{Nonnegative}) where T
     return true
 end
+
+function MOIB.added_constrained_variable_types(::Type{NonnegativeBridge{T, F}}) where {T, F}
```
blegat: You can just do `Type{<:NonnegativeBridge}` as you don't use `T` and `F` inside the function

coroa:
@blegat Thx for pulling it over the finish line.

PR #2008, opened 2019-07-11
Implement set_objective_coefficient

dourouc05:
Currently, only works for affine objectives: I could not find the relevant functions to call for quadratic objectives, and looking at https://github.com/JuliaOpt/CPLEX.jl/blob/d837d8fc090f223f7467e65fd175a14958e41b9e/src/MOI_wrapper.jl or https://github.com/JuliaOpt/LinQuadOptInterface.jl/blob/a81b31235bac72e180ca7586746b2e5b5e910bee/src/objective.jl doesn't show like anything like this is implemented…

### DIFF
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -184,7 +184,7 @@ function set_objective_coefficient(model::Model, variable::VariableRef, coeff::R
         end
     elseif obj_fct_type == AffExpr || obj_fct_type == QuadExpr
         MOI.modify(backend(model),
-            MOI.ObjectiveFunction{moi_function_type(objective_function_type(model))}(),
+            MOI.ObjectiveFunction{moi_function_type(obj_fct_type)}(),
             MOI.ScalarCoefficientChange(index(variable), coeff)
         )
     else
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -172,14 +172,13 @@ coefficient of a quadratic term where there is only a linear objective.
 function set_objective_coefficient end
 
 function set_objective_coefficient(model::Model, variable::VariableRef, coeff)
+    if model.nlp_data !== nothing && model.nlp_data.nlobj !== nothing
+        error("A nonlinear objective is already set in the model")
+    end
+
     MOI.modify(backend(model),
             MOI.ObjectiveFunction{moi_function_type(objective_function_type(model))}(),
             MOI.ScalarCoefficientChange(index(variable), coeff)
     )
-    # Nonlinear objectives override regular objectives, so if there was a
-    # nonlinear objective set, we must clear it.
-    if model.nlp_data !== nothing
-        model.nlp_data.nlobj = nothing
-    end
     return
 end
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -169,7 +169,7 @@ Set the linear objective coefficient associated with `Variable` to `coefficient`
 Note: this function will throw an error if a nonlinear objective is set.
 """
 function set_objective_coefficient(model::Model, variable::VariableRef, coeff::Real)
-    if model.nlp_data !== nothing && model.nlp_data.nlobj !== nothing
+    if model.nlp_data !== nothing && _nlp_objective_function(model) !== nothing
         error("A nonlinear objective is already set in the model")
     end
 
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -180,13 +180,12 @@ function set_objective_coefficient(model::Model, variable::VariableRef, coeff::R
         if index(current_obj) == index(variable)
             set_objective_function(model, coeff * variable)
         else
-            obj_expr = objective_function(model)
-            add_to_expression!(coeff * variable, obj_expr)
+            set_objective_function(model, add_to_expression!(coeff * variable, current_obj))
         end
     elseif obj_fct_type == MOI.ScalarAffineFunction{Float64} || obj_fct_type == MOI.ScalarQuadraticFunction{Float64}
         MOI.modify(backend(model),
-                MOI.ObjectiveFunction{moi_function_type(objective_function_type(model))}(),
-                MOI.ScalarCoefficientChange(index(variable), coeff)
+            MOI.ObjectiveFunction{moi_function_type(objective_function_type(model))}(),
+            MOI.ScalarCoefficientChange(index(variable), coeff)
         )
     else
         error("Objective function type not supported: $(obj_fct_type)")
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -176,7 +176,12 @@ function set_objective_coefficient(model::Model, variable::VariableRef, coeff::R
     obj_fct_type = moi_function_type(objective_function_type(model))
     if obj_fct_type == MOI.SingleVariable
         # Promote the objective function to be an affine expression.
-        set_objective_function(model, coeff * variable)
+        current_obj = objective_function(model)
+        if index(current_obj) == index(variable)
+            set_objective_function(model, coeff * variable)
+        else
+            set_objective_function(model, current_obj + coeff * variable)
+        end
     elseif obj_fct_type == MOI.ScalarAffineFunction{Float64} || obj_fct_type == MOI.ScalarQuadraticFunction{Float64}
         MOI.modify(backend(model),
                 MOI.ObjectiveFunction{moi_function_type(objective_function_type(model))}(),
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -162,16 +162,13 @@ function objective_function(model::Model,
 end
 
 """
-    set_objective_coefficient(model::Model, term, coeff)
+    set_objective_coefficient(model::Model, variable::VariableRef, coefficient::Real)
 
-Updates the objective function to change the coefficient of the given term.
-If the objective does not have the right type, an error will be thrown:
-changing a variable coefficient for a constant objective, changing the
-coefficient of a quadratic term where there is only a linear objective.
+Set the linear objective coefficient associated with `Variable` to `coefficient`.
+
+Note: this function will throw an error if a nonlinear objective is set.
 """
-function set_objective_coefficient end
-
-function set_objective_coefficient(model::Model, variable::VariableRef, coeff)
+function set_objective_coefficient(model::Model, variable::VariableRef, coeff::Real)
     if model.nlp_data !== nothing && model.nlp_data.nlobj !== nothing
         error("A nonlinear objective is already set in the model")
     end
--- a/src/objective.jl
+++ b/src/objective.jl
@@ -173,8 +173,8 @@ function set_objective_coefficient(model::Model, variable::VariableRef, coeff::R
         error("A nonlinear objective is already set in the model")
     end
 
-    obj_fct_type = moi_function_type(objective_function_type(model))
-    if obj_fct_type == MOI.SingleVariable
+    obj_fct_type = objective_function_type(model)
+    if obj_fct_type == VariableRef
         # Promote the objective function to be an affine expression.
         current_obj = objective_function(model)
         if index(current_obj) == index(variable)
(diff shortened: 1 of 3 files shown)


src/objective.jl
```diff
@@ -160,3 +160,26 @@ function objective_function(model::Model,
...
+function set_objective_coefficient end
+
+function set_objective_coefficient(model::Model, variable::VariableRef, coeff)
+    MOI.modify(backend(model),
+            MOI.ObjectiveFunction{moi_function_type(objective_function_type(model))}(),
+            MOI.ScalarCoefficientChange(index(variable), coeff)
+    )
+    # Nonlinear objectives override regular objectives, so if there was a
+    # nonlinear objective set, we must clear it.
+    if model.nlp_data !== nothing
```
blegat: Maybe throw an error ?

src/objective.jl
```diff
@@ -160,3 +160,26 @@ function objective_function(model::Model,
...
+    set_objective_coefficient(model::Model, term, coeff)
+
+Updates the objective function to change the coefficient of the given term.
+If the objective does not have the right type, an error will be thrown:
+changing a variable coefficient for a constant objective, changing the
+coefficient of a quadratic term where there is only a linear objective.
+"""
+function set_objective_coefficient end
+
+function set_objective_coefficient(model::Model, variable::VariableRef, coeff)
```
odow: There should be a specific docstring here instead of the generic one above:

```Julia
"""
    set_objective_coefficient(model::Model, variable::VariableRef, coefficient::Real)

Set the linear objective coefficient associated with `Variable` to `coefficient`.

Note: this function will throw an error if a nonlinear objective is set.
"""

File: src/objective.jl [hunk 3]
```diff
@@ -160,3 +160,26 @@ function objective_function(model::Model,
...
+
+"""
+    set_objective_coefficient(model::Model, term, coeff)
+
+Updates the objective function to change the coefficient of the given term.
+If the objective does not have the right type, an error will be thrown:
+changing a variable coefficient for a constant objective, changing the
+coefficient of a quadratic term where there is only a linear objective.
+"""
+function set_objective_coefficient end
```
odow: You don't need this stub method and docstring.

src/objective.jl
odow: Shouldn't this just check if the objective is set? You could have a nonlinear constraint and a linear objective, for example.

test/objective.jl
```diff
@@ -64,6 +64,20 @@ function objectives_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType:
             4x + 1, @inferred JuMP.objective_function(m, AffExprType))
     end
 
+    @testset "Linear objective changes" begin
+        m = ModelType()
+        @variable(m, x)
+
+        # No change possible for a single variable.
+        @objective(m, Max, x)
+        @test_throws MethodError set_objective_coefficient(m, x, 4.0)
```
odow: Hmm. Users could reasonably expect that this would work. I would suggest that the solvers promote the objective to `ScalarAffine`.

dourouc05: I did that at the JuMP level -- unless there is a reason to let the solvers handle that case?

src/objective.jl
dourouc05: I just implemented this idea: if there is a nonlinear objective, then throw an error; otherwise, don't care.

src/objective.jl
dourouc05: I did that because I wanted this function to also deal with the coefficients of the quadratic terms in the objective. I suspect this will require another method, but I have no idea about how to implement it…

src/objective.jl
```diff
@@ -160,3 +160,34 @@ function objective_function(model::Model,
...
+
+function set_objective_coefficient(model::Model, variable::VariableRef, coeff)
+    if model.nlp_data !== nothing && model.nlp_data.nlobj !== nothing
+        error("A nonlinear objective is already set in the model")
+    end
+
+    obj_fct_type = moi_function_type(objective_function_type(model))
+    if obj_fct_type == MOI.SingleVariable
+        # Promote the objective function to be an affine expression.
+        set_objective_function(model, coeff * variable)
```
odow: What if the single variable is not `variable`?
```Julia
model = Model()
@variable(model, x)
@variable(model, y)
@objective(model, Min, x)
set_objective_coefficient(model, y, 2.0)
@test JuMP.isequal_canonical(JuMP.objective_function(m), x + 2.0 * y)
```

src/objective.jl
odow: We can add the extra method and docstring in the future. For now, just document the method that is implemented.

src/objective.jl
dourouc05: I didn't think of that case, thanks!

dourouc05:
To make the tests pass, I'm not sure what is missing from JuMPExtension.MyModel…

src/objective.jl
```diff
@@ -160,3 +160,36 @@ function objective_function(model::Model,
...
+
+"""
+    set_objective_coefficient(model::Model, variable::VariableRef, coefficient::Real)
+
+Set the linear objective coefficient associated with `Variable` to `coefficient`.
+
+Note: this function will throw an error if a nonlinear objective is set.
+"""
+function set_objective_coefficient(model::Model, variable::VariableRef, coeff::Real)
+    if model.nlp_data !== nothing && model.nlp_data.nlobj !== nothing
```
mlubin: It's slightly cleaner to check `_nlp_objective_function(model) != nothing` instead of accessing the fields of `model`

mlubin:
> To make the tests pass, I'm not sure what is missing from JuMPExtension.MyModel…

Unless @blegat objects, you can have these tests run only for `Model` and not for `JuMPExtension.MyModel`.

dourouc05:
I just did that (moving the relevant tests to a new function, I thought it was the cleanest way of only testing these functions for the "base" JuMP model).

dourouc05:
Now that the tests pass, is this ready for merge?

src/objective.jl
```diff
@@ -160,3 +160,36 @@ function objective_function(model::Model,
...
+Set the linear objective coefficient associated with `Variable` to `coefficient`.
+
+Note: this function will throw an error if a nonlinear objective is set.
+"""
+function set_objective_coefficient(model::Model, variable::VariableRef, coeff::Real)
+    if model.nlp_data !== nothing && _nlp_objective_function(model) !== nothing
+        error("A nonlinear objective is already set in the model")
+    end
+
+    obj_fct_type = moi_function_type(objective_function_type(model))
```
blegat: You could just work with `objective_function_type` and do `obj_fuct_type == VariableRef`, ...

src/objective.jl
```diff
@@ -160,3 +160,36 @@ function objective_function(model::Model,
...
+    if obj_fct_type == MOI.SingleVariable
+        # Promote the objective function to be an affine expression.
+        current_obj = objective_function(model)
+        if index(current_obj) == index(variable)
+            set_objective_function(model, coeff * variable)
+        else
+            set_objective_function(model, current_obj + coeff * variable)
```
blegat: You will allocate two AffExpr here. Do `add_to_expression!(coeff * variable, current_obj)` to allocate only one

src/objective.jl
```diff
@@ -160,3 +160,37 @@ function objective_function(model::Model,
...
+    end
+
+    obj_fct_type = moi_function_type(objective_function_type(model))
+    if obj_fct_type == MOI.SingleVariable
+        # Promote the objective function to be an affine expression.
+        current_obj = objective_function(model)
+        if index(current_obj) == index(variable)
+            set_objective_function(model, coeff * variable)
+        else
+            obj_expr = objective_function(model)
```
odow: You don't need to query this twice? Just use `current_obj`

src/objective.jl
```diff
@@ -160,3 +160,37 @@ function objective_function(model::Model,
...
+            add_to_expression!(coeff * variable, obj_expr)
+        end
+    elseif obj_fct_type == MOI.ScalarAffineFunction{Float64} || obj_fct_type == MOI.ScalarQuadraticFunction{Float64}
+        MOI.modify(backend(model),
+                MOI.ObjectiveFunction{moi_function_type(objective_function_type(model))}(),
```
odow: Indentation is off.

test/objective.jl
```diff
@@ -64,6 +64,26 @@ function objectives_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType:
             4x + 1, @inferred JuMP.objective_function(m, AffExprType))
     end
 
+    @testset "Linear objective changes" begin
+        m = ModelType()
```
odow: Just use `m = Model()` instead.  Then you don't need the `== VariableRef` below.

mlubin:
LGTM. Please add the new function to the documentation, i.e. [here](http://www.juliaopt.org/JuMP.jl/v0.19.2/objective/).

dourouc05:
Is this enough for the documentation? 

I partly reverted a previous commit when updating the objective expression in the single variable case: applying @odow's comment made the tests fail (I don't understand why they passed before), so I use the return value of `add_to_expression!` to reset the objective to a true expression (and no more just a variable reference).

src/objective.jl
```diff
@@ -160,3 +160,36 @@ function objective_function(model::Model,
...
+    end
+
+    obj_fct_type = moi_function_type(objective_function_type(model))
+    if obj_fct_type == MOI.SingleVariable
+        # Promote the objective function to be an affine expression.
+        current_obj = objective_function(model)
+        if index(current_obj) == index(variable)
+            set_objective_function(model, coeff * variable)
+        else
+            set_objective_function(model, add_to_expression!(coeff * variable, current_obj))
```
mlubin: This should be `add_to_expression!(current_obj, coef, variable)`. The expression being added to is the first argument.

dourouc05: If the `else` reduces to `add_to_expression!(current_obj, coeff, variable)`, then I get this "nice" method error: 

```
  MethodError: no method matching add_to_expression!(::VariableRef, ::Float64, ::VariableRef)
  Closest candidates are:
    add_to_expression!(!Matched::GenericAffExpr{C,V}, ::Real, ::V) where {C, V} at C:\Users\…\.julia\dev\JuMP\src\aff_expr.jl:236
    add_to_expression!(!Matched::GenericQuadExpr{C,V}, ::Real, ::V) where {C, V} at C:\Users\…\.julia\dev\JuMP\src\quad_expr.jl:174
    add_to_expression!(!Matched::GenericQuadExpr{C,V}, ::C, ::V, !Matched::V) where {C, V} at C:\Users\…\.julia\dev\JuMP\src\quad_expr.jl:278
    ...
  Stacktrace:
   [1] set_objective_coefficient(::Model, ::VariableRef, ::Float64) at C:\Users\…\.julia\dev\JuMP\src\objective.jl:183
```

test/objective.jl
```diff
@@ -65,23 +65,22 @@ function objectives_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType:
     end
 
     @testset "Linear objective changes" begin
-        m = ModelType()
+        # Only works with JuMP model, not extensions.
```
mlubin: These testsets should be outside `objectives_test` and inside the `"Objectives for JuMP.Model"` testset. Currently the tests are run twice for no reason.

dourouc05: Did you prefer this version? https://github.com/JuliaOpt/JuMP.jl/pull/2008/commits/a3694f93e1ad3138226366d3c66ca2197dbaa24a

src/objective.jl
mlubin: Oh, good point. LGTM as is.

test/objective.jl
mlubin: Yes

dourouc05: Done in the last commit!

mlubin:
I'll merge by the end of the day unless there are additional comments.

dourouc05:
Great, thanks :)!

mlubin:
> Great, thanks :)!

Thanks for the contribution and good job addressing our comments!

src/objective.jl
blegat: Why use the MOI function type and not the JuMP function type ?

src/objective.jl
```diff
@@ -160,3 +160,36 @@ function objective_function(model::Model,
...
+        end
+    elseif obj_fct_type == AffExpr || obj_fct_type == QuadExpr
+        MOI.modify(backend(model),
+            MOI.ObjectiveFunction{moi_function_type(objective_function_type(model))}(),
```
blegat: Replace `objective_function_type(model)` by `obj_fct_type`

PR #2015, opened 2019-07-29
Constraints: use `cref` for constraint references.

dourouc05:
Used to be a mix of cr, ref, cref, constraint_ref, and v. This looked especially strange in the documentation. `cref` seems to be the most commonly used name in the code.

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -347,7 +347,7 @@ reshape_set(set::MOI.AbstractVectorSet, ::VectorShape) = set
 shape(c::VectorConstraint) = c.shape
 function constraint_object(cref::ConstraintRef{Model, _MOICON{FuncType, SetType}}) where
         {FuncType <: MOI.AbstractVectorFunction, SetType <: MOI.AbstractVectorSet}
-    model = ref.model
+    model = cref.model
     f = MOI.get(model, MOI.ConstraintFunction(), cref)::FuncType
     s = MOI.get(model, MOI.ConstraintSet(), cref)::SetType
     return VectorConstraint(jump_function(model, f), s, ref.shape)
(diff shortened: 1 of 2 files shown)


mlubin:
Thanks for the tidying :).
How do we abbreviate "constraint" throughout the code? Since the style guide says to use an underscore to separate words in names, the consistent naming would be `[cons/constr]_ref`.

dourouc05:
I saw a lot of `con`, too. Numerically: 28 `c` (18 in the code), 20 `con` (4 in the code), 0 `cons`, 0 `constr`, 82 `constraint` (12 in the code).

mlubin:
If we want to declare `con` as the standard abbreviation for `constraint` (e.g., adding it [here](http://www.juliaopt.org/JuMP.jl/dev/style/#Abbreviations-1)), then `con_ref` would be the standard abbreviation. That sounds fine with me. 
@blegat @odow

odow:
I agree with @mlubin

blegat:
`con` and `con_ref` sounds good to me

dourouc05:
Here it is, then, with `con_ref` and a note in the docs! Tests do pass locally.

src/constraints.jl
```diff
@@ -459,35 +459,35 @@ con : 2 x <= 4.0
...
 end
 
 """
-    standard_form_rhs(constraint::ConstraintRef)
+    standard_form_rhs(con_ref::ConstraintRef)
 
-Return the right-hand side term of `constraint` after JuMP has converted the
-constraint into its standard form. See also [`set_standard_form_rhs`](@ref).
+Return the right-hand side term of the constraint after JuMP has converted it
+its standard form. See also [`set_standard_form_rhs`](@ref).
```
mlubin: "converted it *into* its standard form"

mlubin:
Please fix the test failures, e.g., https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/565181397#L424

PR #2018, opened 2019-08-06
Clarified modifying variable coefficient example

a-torgovitsky:
The example now shows that this can be done to each component of an array of variables.
https://discourse.julialang.org/t/workaround-for-modifying-coefficients-in-constraints-in-jump/27123

### DIFF
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -608,17 +608,17 @@ The constraint `con` is now equivalent to `2x <= 2`.
 To modify the scalar coefficients of a cosntraint (but notably *not yet* the
 quadratic coefficients), use [`set_standard_form_coefficient`](@ref). To query
 the current coefficient, use [`standard_form_coefficient`](@ref).
-```jldoctest; setup = :(model = Model(); @variable(model, x))
-julia> @constraint(model, con, 2x <= 1)
-con : 2 x <= 1.0
+```jldoctest; setup = :(model = Model(); @variable(model, x[1:2]))
+julia> @constraint(model, con, 2x[1] + x[2] <= 1)
+con : 2 x[1] + x[2] ≤ 1.0
 
-julia> set_standard_form_coefficient(con, x, 3)
+julia> set_standard_form_coefficient(con, x[2], 0)
 
 julia> con
-con : 3 x <= 1.0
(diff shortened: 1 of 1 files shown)


docs/src/constraints.md
```diff
@@ -608,17 +608,17 @@ The constraint `con` is now equivalent to `2x <= 2`.
 To modify the scalar coefficients of a cosntraint (but notably *not yet* the
```
mlubin: This wording about scalar coefficients of a constraint doesn't really make sense to me. I see why you were confused. What about:

To modify the linear coefficients of a variable in a constraint (but notably *not yet* the quadratic coefficients) ...

a-torgovitsky: That would make more sense to me.
I usually interpret scalar as in opposite to vector (or array).

Since it's the constraint not the coefficient that is linear, the following might be better still:

> To modify the coefficients on a variable in a linear constraint (but notably not yet the coefficients in a quadratic constraint) ...

mlubin: This method works for both linear and quadratic constraints. It modifies the coefficients on the variable in the linear part of the constraint:
```julia
model = Model()
@variable(model, x)
@constraint(model, con, 2x + x^2 <= 10)
set_standard_form_coefficient(con, x, 3.0)
julia> con
con : x² + 3 x ≤ 10.0
```

Open to suggestions if you have better wording.

a-torgovitsky: Good point. How about this:

> To modify the coefficients for a linear term in a constraint (but notably not yet the coefficients on a quadratic term) ...

mlubin: Sounds good

odow:
Thanks for working on this @a-torgovitsky!

a-torgovitsky:
Glad I could help in a small way. I'll keep an eye out for other ways to improve the documentation.

docs/src/constraints.md
mlubin: @a-torgovitsky, do you mind modifying your pull request with the new wording? This way you'll receive the credit in the commit log for your suggestion.

a-torgovitsky: Ok, credit is good I suppose ;)

PR #2028, opened 2019-08-15
added documentation for categories

matbesancon:
JuMP `@variable` can be odd for newcomers because there seems to be bindings that don't exist:

```julia
@variable(m, x[1:10], Bin)

JuMP.Bin # not found
?JuMP.Bin # not found either
```

This allows having a short docstring for these categories

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -67,3 +67,18 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
           "`@constraint(model, norm(x) <= t)` should now be written as ",
           "`@constraint(model, [t; x] in SecondOrderCone())`")
 end
+
+"""
+    Bin
+
+Binary variable category, used in the `@variable` macro.
+"""
+Bin
+
+"""
+    PSD
+
+Postitive semi-definite variable category, used in the `@variable` macro,
+only valid for square matrix variables.
+"""
+PSD


matbesancon:
These haven't been exported, so users still need the fully qualified `JuMP.PSD` to find

src/sets.jl
```diff
@@ -67,3 +67,18 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
           "`@constraint(model, norm(x) <= t)` should now be written as ",
           "`@constraint(model, [t; x] in SecondOrderCone())`")
 end
+
+"""
+    Bin
+
+Binary variable category, used in the `@variable` macro.
+"""
+Bin
```
odow: I didn't know you could declare variables like this! 

But, maybe we should make it a type? Because it still has this:
```Julia
julia> JuMP.Bin
ERROR: UndefVarError: Bin not defined
```

We also don't really want to declare an `Int` JuMP-local variable. It would get exported and might cause all sorts of issues.

matbesancon: It's a bizarre trickery, it's documented, but not an identifier. You get it in REPL auto-completion, but raises UndefVar

matbesancon: We can make them types indeed

mlubin:
I agree that the use of `Bin` and `Int` can be a bit confusing in the macros, but we cannot export `Int` in any case because of the conflict with the name in `Base`.

> These haven't been exported, so users still need the fully qualified JuMP.PSD to find them

All symbols without preceding underscores are exported from JuMP:
https://github.com/JuliaOpt/JuMP.jl/blob/0565f26eb5aca10c92516143f8a40280257b49b9/src/JuMP.jl#L757-L758

I've thought about `@variable(model, x, :Bin)` and `@variable(model, x, :Int)`. Which is more confusing for new users, Julia symbol syntax with a colon or seemingly referring to undefined bindings like `Bin`?

matbesancon:
> Which is more confusing for new users, Julia symbol syntax with a colon or seemingly referring to undefined bindings like Bin?

I would say symbols are less confusing, they can experiment with them in the REPL, get `typeof(:Int)` and then check the documentation for Symbol.

The fact that Int is among the categories is even more confusing in the current state, because users might think the last argument passed is a type

blegat:
`PSD` should be replaced by `PSDCone()`. In fact, we should allow any `MOI.AbstractVectorSet` and then reshape just like with `@constraint`. We could do the same for `Int` and `Bin`. We expect ` MOI.ZeroOne()` and `JuMP.Bin()` just redirects to it

matbesancon:
So what should do here?

matbesancon:
This may be personal but I find it less of a problem than `Bin` because you know there is "something"  (an identifier) which is `Int`, while you need to understand how macro operate to get that `Bin` doesn't exist

matbesancon:
bump here?

odow:
I think we're going to punt on this until we come up with a better solution. Exporting a symbol for `Bin` is weird because we don't for `Int`. And we won't be changing the syntax before JuMP 1.0. So maybe just close?

odow:
Closing until we come up with a better solution in the future.

PR #2032, opened 2019-08-19
Remove dictionaries tracking SingleVariable constraints

mlubin:
This follows from https://github.com/JuliaOpt/MathOptInterface.jl/issues/570.

Unfortunately the performance issue in https://github.com/JuliaOpt/JuMP.jl/issues/1607 is not resolved:
```julia
using JuMP
using BenchmarkTools

function plain(N)
    model = Model()
    x = @variable(model, [1:N])
    return x
end
@btime plain(1_000_000)

function with_int_and_bounds(N)
    model = Model()
    x = @variable(model, [1:N], Int, lower_bound=-1, upper_bound=1)
    return x
end
@btime with_int_and_bounds(1_000_000)
```
yields
```
  67.775 ms (2000220 allocations: 73.43 MiB)
  674.142 ms (18000220 allocations: 305.57 MiB)
```
So it's still about 10x slower to create variables with bounds than without them. I haven't started looking into the reasons for this.

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -328,18 +328,20 @@ Return `true` if `v` has a lower bound. If `true`, the lower bound can be
 queried with [`lower_bound`](@ref). See also [`LowerBoundRef`](@ref).
 """
 function has_lower_bound(v::VariableRef)
-    return haskey(owner_model(v).variable_to_lower_bound, index(v))
+    return _moi_has_lower_bound(backend(owner_model(v)), v)
 end
 
-function lower_bound_index(v::VariableRef)
-    if !has_lower_bound(v)
-        error("Variable $(v) does not have a lower bound.")
-    end
-    return owner_model(v).variable_to_lower_bound[index(v)]
+# _moi_* methods allow us to work around the type instability of the backend of
+# a model.
+function _moi_has_lower_bound(backend, v::VariableRef)
+    return MOI.is_valid(backend, _lower_bound_index(v))
 end
-function set_lower_bound_index(v::VariableRef, cindex::_MOILB)
-    owner_model(v).variable_to_lower_bound[index(v)] = cindex
+
+function _lower_bound_index(v::VariableRef)
+    return _MOICON{MOI.SingleVariable, MOI.GreaterThan{Float64}}(index(v).value)
 end
+
+
 
 """
     set_lower_bound(v::VariableRef, lower::Number)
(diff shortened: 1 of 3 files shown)


src/variables.jl
```diff
@@ -600,10 +581,8 @@ function set_integer(variable_ref::VariableRef)
         error("Cannot set the variable_ref $(variable_ref) to integer as it " *
               "is already binary.")
     end
-    constraint_ref = MOI.add_constraint(backend(owner_model(variable_ref)),
-                                        MOI.SingleVariable(index(variable_ref)),
-                                        MOI.Integer())
-    set_integer_index(variable_ref, constraint_ref)
+    MOI.add_constraint(backend(owner_model(variable_ref)),
```
blegat: The `is_integer` and `is_binary` above might be the performance bottleneck. It should be replaced by errors at the MOI level like:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/3f4857816281a79744af0170391c435e318123f1/src/Utilities/model.jl#L431-L445

src/variables.jl
```diff
@@ -351,13 +345,12 @@ function set_lower_bound(v::VariableRef, lower::Number)
     newset = MOI.GreaterThan(convert(Float64, lower))
     # do we have a lower bound already?
     if has_lower_bound(v)
-        cindex = lower_bound_index(v)
+        cindex = _lower_bound_index(v)
         MOI.set(backend(owner_model(v)), MOI.ConstraintSet(), cindex, newset)
     else
         @assert !is_fixed(v)
-        cindex = MOI.add_constraint(
+        MOI.add_constraint(
```
blegat: We should have a shortcut in `JuMP.add_variable` to skip the branching on `has_lower_bound` and directly go in the `else`, e.g., by creating an auxiliary function for this `else`

src/variables.jl
mlubin: Why is `is_valid` expensive? I thought it would just look up an element in an array.

blegat: `is_integer` called `backend` which is type unstable. Maybe we should have `moi_set_integer` etc... to make sure we call `backend` only once in `JuMP.add_variable`

mlubin:
I did a big refactoring to avoid type instabilities due to accessing the backend. It made a big difference:
```julia
  100.329 ms (2000220 allocations: 126.46 MiB)
  291.198 ms (4000220 allocations: 154.98 MiB)
```
There are still two additional allocations per variable that are unaccounted for.

PR #2041, opened 2019-08-23
Add dual_objective_function and update doc links to MOI

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/2030

### DIFF
--- a/docs/src/solutions.md
+++ b/docs/src/solutions.md
@@ -31,7 +31,7 @@ see the [Termination statuses](@ref) section below.
 
 Second, we can query the [`primal_status`](@ref) and the [`dual_status`](@ref),
 which will tell us what kind of results we have for our primal and dual
-solutions. This might be an optimal primal-dual pair, a primal solution without 
+solutions. This might be an optimal primal-dual pair, a primal solution without
 a corresponding dual solution, or a certificate of primal or dual infeasibility.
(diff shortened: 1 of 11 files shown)


docs/src/solutions.md
```diff
@@ -95,10 +95,12 @@ The container type (e.g., scalar, vector, or matrix) of the returned solution
 
 The objective value of a solved problem can be obtained via
 [`objective_value`](@ref). The best known bound on the optimal objective
-value can be obtained via [`objective_bound`](@ref).
+value can be obtained via [`objective_bound`](@ref). If dual values are
+available, the value of the dual objective can be obtained via
```
mlubin: Solvers could provide duals without providing a dual objective (e.g., OSQP), so this isn't precise.

src/objective.jl
```diff
@@ -24,6 +24,14 @@ Return the objective value after a call to `optimize!(model)`.
 """
 objective_value(model::Model)::Float64 = MOI.get(model, MOI.ObjectiveValue())
 
+"""
+    dual_objective_value(model::Model)
+
+Return the value of the objective of the dual problem after a call to
+`optimize!(model)`.
```
mlubin: Maybe a TODO, but the docstrings for `objective_bound` and `dual_objective_value` should both mention the error cases, i.e., what happens if the attribute isn't available.

Issue #2042, opened 2019-08-23
Style: interpolated variables in strings need parentheses


odow:
I propose that all interpolated variables in strings need parentheses.

> Good:
> ```Julia
> x = 1
> "Interpolated x=$(x) with parentheses.
> ```
>
> Bad:
> ```Julia
> x = 1
> "Interpolated x=$x without parentheses.
> ```

This stems from a potential breaking change between Julia 1.2 and Julia 1.3. See https://github.com/JuliaLang/julia/issues/33044.

PR #2043, opened 2019-08-23
Correcting string interpolations failing on julia 1.3

matbesancon:


matbesancon:
Closes #2042

blegat:
Can you enable `1.3` on Travis and AppVeyor ?

matbesancon:
To go a bit deeper on the original bug, it's not all identifiers, it's just var being broken:
```julia
julia> println("$var.")
ERROR: syntax: unexpected ")"
Stacktrace:
 [1] top-level scope at REPL[3]:0

julia> println("$val.")
2.
```

matbesancon:
Given the answer on https://github.com/JuliaLang/julia/issues/33044, do we still want this?
cc @odow

src/parse_nlp.jl
```diff
@@ -22,7 +22,7 @@ function _parse_NL_expr(m, x, tapevar, parent, values)
         elseif _is_prod(header)
             operatorid = operator_to_id[:*]
         else
-            error("Unrecognized expression $header(...)")
+            error("Unrecognized expression $(header...)")
```
odow: Is it meant to be `$(header)(...)`?

src/parse_nlp.jl
```diff
@@ -48,7 +48,7 @@ function _parse_NL_expr(m, x, tapevar, parent, values)
             opname = x.args[1]
             errorstring = "$opname() can appear in nonlinear expressions " *
             " only if the argument is a generator statement, for example, " *
-            "$opname(x[i] for i in 1:N)."
+            "$(opname(x[i] for i in 1:N))."
```
odow: I think this is meant to be `$(opname)(x[i] for i in 1:N)`

src/parse_nlp.jl
matbesancon: fixed thanks

c42f:
Thanks for reporting the `var` bug upstream.

The rule for where string interpolations stop is meant to be super simple: they consume exactly one identifier or parenthesized expression.

matbesancon:
the other errors on 1.3 are about a `_mul!` undefined

Issue #2014, opened 2019-07-29
Allow `@constraint` without relational operator?


dourouc05:
Thinking longer term: why is `@constraint` limited to using relational operators (<=, >=, and ==)? Getting rid of this limitation would be very useful if adding constraint programming into JuMP (I have plans for that, but not quite soon): you assert that something is true (like `alldifferent`). 

Closer to present, I'm working on https://github.com/dourouc05/MOModelingExtensions.jl, to allow beginners to write some expressions more easily (with brute-force formulations that work everywhere, even though they could be simplified in most cases — but I'm not analysing the formulation as a whole to improve it). It would be nice to write something like `@constraint(m, any([x, y, z]))` instead of `@constraint(m, any([x, y, z]) == true)`.

So, the functionality as I imagine it: allow JuMP to parse things like `@constraint(m, any([x, y, z]))`, and then let extensions do the rest of the work.

(I prefer to bring this issue sooner than later, in case this would entail design decisions that would require breaking changes and thus would have to wait for JuMP 2.0…)

mlubin:
The `@constraint` macro is due for a rewrite, although it's currently not on the [1.0 roadmap](http://www.juliaopt.org/JuMP.jl/dev/roadmap/). That doesn't mean it won't happen before 1.0, but it means that none of the core developers have committed to do it.

My first question for this is, why are macros needed? The two reasons that we have macros are (1) for performance, because naive summation of affine expressions using operator overloading is O(n^2), while we can rewrite sums to be O(n) using macros, and (2) a macro gives us a syntactic context in which we can change the definition of the relational operators. In regular Julia code, the operators return bools, while inside the macros they're reinterpreted as mathematical constraints. It's not clear that either of these two motivations still holds for the examples you mentioned.

 Could you do `add_constraint(m, any([x, y, z]))` or `add_alldifferent(m, [x, y, z])`? What kind of object does `any([x, y, z])` return?

dourouc05:
Indeed, for this kind of constraint, the usual rationale behind macros is not so relevant: usually, just "raw" variables are passed as arguments, not really expressions (for a rare need, I suppose forcing users to use `@expression` is not such a big deal). Also, there is no need for macros for these expressions to be considered different than the usual boolean operators: in my implementation, even outside macros, `any([x, y, z])` directly returns a new variable, not a boolean. 

[...]

PR #2051, opened 2019-08-29
Add extension points for more generic `@constraint` syntax

dourouc05:
Prototype for #2014. Comments are welcome! This PR is mostly to enable more things from extensions.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -192,12 +192,6 @@ function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
     vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
 end
 
-function parse_constraint(_error::Function, sense::Val, F...)
-    sense_symbol = typeof(sense).parameters[1]
-    (sense, vectorized) = _check_vectorized(sense_symbol)
-    vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), F...)...
-end
-
 function parse_ternary_constraint(_error::Function, vectorized::Bool, lb, ::Union{Val{:(<=)}, Val{:(≤)}}, aff, rsign::Union{Val{:(<=)}, Val{:(≤)}}, ub)
     newaff, parseaff = _MA.rewrite(aff)
     newlb, parselb = _MA.rewrite(lb)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -246,7 +246,16 @@ function parse_constraint_head(_error::Function, ::Val, args...)
     _unknown_constraint_expr(_error)
 end
 function parse_constraint(_error::Function, args...)
-    _unknown_constraint_expr(_error)
+    # Define this as the last fallback: either this is a function call that may
+    # be overridden by extensions, or a syntax that is not recognised.
+    # Multiple dispatch does not work here, due to ambiguity with:
+    #     parse_constraint(_error::Function, lb, lsign::Symbol, aff, rsign::Symbol, ub)
+    if args[1] isa Symbol
+        (sense, vectorized) = _check_vectorized(args[1])
+        vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), args[2:end]...)...
+    else
+        _unknown_constraint_expr(_error)
+    end
 end
 
 # Generic fallback.
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -181,13 +181,18 @@ function parse_constraint_expr(_error::Function, expr::Expr)
     return parse_constraint_head(_error, Val(expr.head), expr.args...)
 end
 function parse_constraint_head(_error::Function, ::Val{:call}, args...)
-    return parse_constraint_call(_error, Val(args[1]), args[2:end]...)
-end
-function parse_constraint(_error::Function, args...)
-    return parse_constraint_call(_error, Val(args[1]), args[2:end]...)
+    return parse_constraint(_error, args...)
 end
 
-function parse_constraint_call(_error::Function, sense::Val, F...)
+function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
+    (sense, vectorized) = _check_vectorized(sense)
+    vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
+end
+function parse_constraint(_error::Function, ::Val{:call}, args...)
+    return parse_constraint(_error, Val(args[1]), args[2:end]...)
+end
+
+function parse_constraint(_error::Function, sense::Val, F...)
     sense_symbol = typeof(sense).parameters[1]
     (sense, vectorized) = _check_vectorized(sense_symbol)
     vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), F...)...
@@ -240,7 +245,7 @@ end
 function parse_constraint_head(_error::Function, ::Val, args...)
     _unknown_constraint_expr(_error)
 end
-function parse_constraint_call(_error::Function, args...)
+function parse_constraint(_error::Function, args...)
     _unknown_constraint_expr(_error)
 end
 
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -191,9 +191,6 @@ function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
     (sense, vectorized) = _check_vectorized(sense)
     vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
 end
-function parse_constraint(_error::Function, ::Val{:call}, args...)
-    return parse_constraint(_error, Val(args[1]), args[2:end]...)
-end
 
 function parse_constraint(_error::Function, sense::Val, F...)
     sense_symbol = typeof(sense).parameters[1]
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -181,18 +181,13 @@ function parse_constraint_expr(_error::Function, expr::Expr)
     return parse_constraint_head(_error, Val(expr.head), expr.args...)
 end
 function parse_constraint_head(_error::Function, ::Val{:call}, args...)
-    return parse_constraint(_error, args...)
+    return parse_constraint_call(_error, Val(args[1]), args[2:end]...)
+end
+function parse_constraint(_error::Function, args...)
+    return parse_constraint_call(_error, Val(args[1]), args[2:end]...)
 end
 
-function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
-    (sense, vectorized) = _check_vectorized(sense)
-    vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
-end
-function parse_constraint(_error::Function, ::Val{:call}, args...)
-    return parse_constraint(_error, Val(args[1]), args[2:end]...)
-end
-
-function parse_constraint(_error::Function, sense::Val, F...)
+function parse_constraint_call(_error::Function, sense::Val, F...)
     sense_symbol = typeof(sense).parameters[1]
     (sense, vectorized) = _check_vectorized(sense_symbol)
     vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), F...)...
@@ -245,7 +240,7 @@ end
 function parse_constraint_head(_error::Function, ::Val, args...)
     _unknown_constraint_expr(_error)
 end
-function parse_constraint(_error::Function, args...)
+function parse_constraint_call(_error::Function, args...)
     _unknown_constraint_expr(_error)
 end
 
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -155,6 +155,9 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
     variable, parse_code = _MA.rewrite(func)
     return parse_code, _build_call(_error, vectorized, variable, set)
 end
+function parse_one_operator_constraint(_error::Function, args...)
+    _unknown_constraint_expr(_error)
+end
 
 _functionize(v::VariableRef) = convert(AffExpr, v)
 _functionize(v::AbstractArray{VariableRef}) = _functionize.(v)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -188,6 +188,15 @@ function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
     (sense, vectorized) = _check_vectorized(sense)
     vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
 end
+function parse_constraint(_error::Function, ::Val{:call}, args...)
+    return parse_constraint(_error, Val(args[1]), args[2:end]...)
+end
+
+function parse_constraint(_error::Function, sense::Val, F...)
+    sense_symbol = typeof(sense).parameters[1]
+    (sense, vectorized) = _check_vectorized(sense_symbol)
+    vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), F...)...
+end
 
 function parse_ternary_constraint(_error::Function, vectorized::Bool, lb, ::Union{Val{:(<=)}, Val{:(≤)}}, aff, rsign::Union{Val{:(<=)}, Val{:(≤)}}, ub)
     newaff, parseaff = _MA.rewrite(aff)
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -176,6 +176,20 @@ function custom_expression_test(ModelType::Type{<:JuMP.AbstractModel})
     end
 end
 
+function JuMP.parse_constraint_call(_error::Function, ::Val{:f}, x)
+    return false, :(), :(build_constraint($_error, $(esc(x)), $(esc(CustomType()))))
+end
+function custom_function_test(ModelType::Type{<:JuMP.AbstractModel})
+    @testset "Custom function" begin
+        model = ModelType()
+        @variable(model, x)
+        @constraint(model, con_ref, f(x))
+        con = JuMP.constraint_object(con_ref)
+        @test jump_function(con) == x
+        @test moi_set(con) isa CustomSet
+    end
+end
+
 function macros_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Type{<:JuMP.AbstractVariableRef})
     @testset "build_constraint on variable" begin
         m = ModelType()
@@ -357,8 +371,8 @@ function macros_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Typ
     end
 
     build_constraint_keyword_test(ModelType)
-
     custom_expression_test(ModelType)
+    custom_function_test(ModelType)
 end
 
 @testset "Macros for JuMP.Model" begin
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -188,7 +188,7 @@ function custom_function_test(ModelType::Type{<:JuMP.AbstractModel})
         @test jump_function(con) == x
         @test moi_set(con) isa CustomSet
 
-        @test_macro_throws LoadError @constraint(model, g(x))
+        @test_macro_throws ErrorException @constraint(model, g(x))
     end
 end
 
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -176,8 +176,8 @@ function custom_expression_test(ModelType::Type{<:JuMP.AbstractModel})
     end
 end
 
-function JuMP.parse_constraint_head(_error::Function, ::Val{:call}, ::Val{:f}, x)
-    return false, :(), :(build_constraint($_error, $(esc(x)), $(esc(CustomType()))))
+function JuMP.parse_one_operator_constraint(_error::Function, ::Bool, ::Val{:f}, x)
+    return :(), :(build_constraint($_error, $(esc(x)), $(esc(CustomType()))))
 end
 function custom_function_test(ModelType::Type{<:JuMP.AbstractModel})
     @testset "Custom function" begin
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -188,7 +188,7 @@ function custom_function_test(ModelType::Type{<:JuMP.AbstractModel})
         @test jump_function(con) == x
         @test moi_set(con) isa CustomSet
 
-        @test_throws LoadError eval(:(@constraint(model, g(x))))
+        @test_macro_throws LoadError @constraint(model, g(x))
     end
 end
 
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -176,7 +176,7 @@ function custom_expression_test(ModelType::Type{<:JuMP.AbstractModel})
     end
 end
 
-function JuMP.parse_constraint_call(_error::Function, ::Val{:f}, x)
+function JuMP.parse_constraint_head(_error::Function, ::Val{:call}, ::Val{:f}, x)
     return false, :(), :(build_constraint($_error, $(esc(x)), $(esc(CustomType()))))
 end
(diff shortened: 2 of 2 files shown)


src/macros.jl
```diff
@@ -543,6 +556,18 @@ function build_constraint(
     return VectorConstraint(x, MOI.SOS2{Float64}(set.weights))
 end
 
+function build_non_relational_constraint(_error::Function, head, args...) # This would remain in JuMP.
+    _error("Constraint $(head) is not recognised.")
+end
+
+function build_non_relational_constraint(_error::Function, head::Val{:alldifferent}, args::VariableRef...) # This would be an extension to JuMP.
```
blegat: We could redirect `isequal` and `isless` to the build call with `EqualTo`/`LessThan`/`GreaterThan`

src/macros.jl
```diff
@@ -543,6 +556,18 @@ function build_constraint(
     return VectorConstraint(x, MOI.SOS2{Float64}(set.weights))
 end
 
+function build_non_relational_constraint(_error::Function, head, args...) # This would remain in JuMP.
```
blegat: Why not dispatch on the `head` in `parse_...` as done in `parse_one_operator_constraint` ?
`build_constraint` should have 3 arguments: `_error`, `func`, `set` that builds the constraint `func`-in-`set`.
`parse_..._constraint` is where the function and sets are defined.

dourouc05:
@blegat I did as you suggested (at least, I hope so…), which yielded a simpler implementation. Not all tests still pass, but quite a few of them do (issues with bridging).

dourouc05:
Most of the tests pass now, except one for the `@constraint(model, [x == 1-x, 2x == 3])` syntax, but I've been unable to understand why it worked previously… 

So, for now, this PR allows to write JuMP extensions to allow global constraints like `alldifferent` and reification (binary variable indicating if a condition is met). I think the implementation is rather clean in its principles, only added overloads to `parse_constraint` (and changing its interface to allow much finer dispatch). What's missing right now for full constraint-programming modelling: logical operator parsing (including implications), allowing operators almost everywhere (with constraints like `x >= min{y, z*z}` or having `IloSizeOf` with CPLEX), maybe allow other types of variables (again, with CPLEX, you can have intervals as variables).

src/macros.jl
```diff
@@ -456,6 +471,9 @@ function parse_constraint(_error::Function, args...)
           "       expr1 == expr2\n" * "       lb <= expr <= ub")
 end
 
+function_to_set(_error::Function, ::Val{:alldifferent}) = nothing # TODO: Needs a MOI extension (new set alldifferent).
```
blegat: Why is this method needed ?

src/macros.jl
```diff
@@ -327,12 +327,14 @@ function _macro_assign_and_return(code, variable, name;
     end
 end
 
-function _check_vectorized(sense::Symbol)
+function _check_vectorized(sense::Val)
```
blegat: Do `Val{S}) where S` to get the symbol `S` instead of doing `string(sense)` and then needed to do weird string processing

src/macros.jl
```diff
@@ -408,11 +410,24 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense
...
+function parse_constraint(_error::Function, ::Val{:call}, fct::Union{Val{:in}, Val{:∈}}, args...)
+    false, parse_one_operator_constraint(_error, false, Val(:in), args...)...
+end
+
+function parse_constraint(_error::Function, ::Val{:call}, fct::Val, args...)
+    set = function_to_set(_error, fct)
+    false, parse_one_operator_constraint(_error, false, Val(:in), args, set)...
+end
+
+function parse_constraint(_error::Function, ::Val{:(:=)}, var, args...)
```
blegat: Which macro calls this method ?

src/macros.jl
```diff
@@ -408,11 +410,24 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense
...
     (sense, vectorized) = _check_vectorized(sense)
     vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
 end
 
+function parse_constraint(_error::Function, ::Val{:call}, fct::Union{Val{:in}, Val{:∈}}, args...)
+    false, parse_one_operator_constraint(_error, false, Val(:in), args...)...
+end
+
+function parse_constraint(_error::Function, ::Val{:call}, fct::Val, args...)
+    set = function_to_set(_error, fct)
```
blegat: Move the `function_to_set` function right before this function and give an example in comments

src/macros.jl
```diff
@@ -408,11 +410,24 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense
     parse_one_operator_constraint(_error, vectorized, Val(:in), aff, set)
 end
 
-function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
+function parse_constraint(_error::Function, ::Val{:call}, sense::Union{Val{:(<=)}, Val{:(≤)}, Val{:(==)}, Val{:(>=)}, Val{:(≥)}, Val{:(.<=)}, Val{:(.≤)}, Val{:(.==)}, Val{:(.>=)}, Val{:(.≥)}}, lhs, rhs)
```
blegat: Why specifying this union here ? If an extension adds a new `parse_one_operator_constraint`, it needs this method to also work with symbols out of this union.

dourouc05: Actually, this union is required to ensure that this function is not called for cases it should not (for instance, for `alldifferent` with two arguments).

src/macros.jl
dourouc05: It is there to make the translation between the syntax offered by JuMP and the sets that will be defined on MOI's side (or rather an extension, I suppose), in the same spirit as `sense_to_set`.

src/macros.jl
blegat: It seems we should merge the three methods if we replace `lhs, rhs` by `args...`.
Then, instead of creating a custom `parse_constraint` method for `Val{:alldifferent}`, you can create a custom `parse_one_operator_constraint`.

src/macros.jl
```diff
@@ -327,12 +327,12 @@ function _macro_assign_and_return(code, variable, name;
...
 
-function _check_vectorized(sense::Symbol)
-    sense_str = string(sense)
+function _check_vectorized(sense::Val{S}) where S
+    sense_str = string(S)
     if sense_str[1] == '.'
         Symbol(sense_str[2:end]), true
     else
-        sense, false
+        Symbol(sense_str), false
```
blegat: What is the difference between `Symbol(sense_str)` and `S` ?

src/macros.jl
```diff
@@ -327,12 +327,12 @@ function _macro_assign_and_return(code, variable, name;
     end
 end
 
-function _check_vectorized(sense::Symbol)
-    sense_str = string(sense)
+function _check_vectorized(sense::Val{S}) where S
```
blegat: Do you have an example for which the heuristic "vectorized iff `S` starts by a `.`" does not work ?

dourouc05: I don't, for now, but you may have some with more developments of constraint programming, like `alldifferent.(vectorA, vectorB)`. I don't think we need to care that much about this case for now.

blegat: Then why not keep `Symbol` ? Using `Val` means that you want to allow it to be extended.

dourouc05: I had problems at some point and had to use Val (or do tricks I didn't know at the time). I've reverted to symbols.

src/macros.jl
```diff
@@ -396,7 +405,7 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
     parsecode, buildcall
 end
 
-function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Val, lhs, rhs)
+function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Sense, lhs, rhs)
```
blegat: The [SetProg](https://github.com/blegat/SetProg.jl) extension uses sense which is `Val{:⊆}` so I would prefer to keep `Val` here. The method is more specific than the other one anyway since it has `lhs, rhs` instead of `args...`.

Wikunia:
As I'll not be able to attend the developers call and currently can't check this as I'm on my phone: is it already possible to write a not equal constraint with this/JuMP?

blegat:
@dourouc05 In https://github.com/JuliaOpt/JuMP.jl/pull/2132, I was able to make `@constraint(model, complements(..., ...))` work without needing to modify the `parse_constraint` API, couldn't this also work for `alldifferent` ?

dourouc05:
@blegat thanks for your ping. However, this case is slightly different: for many constraints, you have notions similar to a left-hand and a right-hand sides. 

* for complementarity constraints, F(x) ⟂ x
* for indicator constraints, condition ⇒ { constraints }

Here, there is just one set of expressions, not two. 

An easy modification would be to force the user to just have `alldifferent([x, y])`, with a vector of expressions as an argument. This way, I think that it would be sufficient to have a definition like this to extend the `parse_constraint` API: 

```julia
function parse_constraint(_error::Function, sense::Symbol, F)
    (sense, vectorized) = _check_vectorized(sense)
    vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), F)...
end
```

I made a new patch to still allow syntax like `alldifferent(x, y)`. I did not see any other way than to slightly modify `_constraint_macro` to allow passing all the arguments as one array. This is much cleaner than previously, at least!

src/macros.jl
```diff
@@ -368,7 +376,13 @@ function _constraint_macro(args, macro_name::Symbol, parsefun::Function)
     # in a function returning `ConstraintRef`s and give it to `Containers.container`.
     idxvars, indices = Containers._build_ref_sets(_error, c)
 
-    vectorized, parsecode, buildcall = parsefun(_error, x.args...)
+    if x.head == :(:=)
+        vectorized, parsecode, buildcall = parsefun(_error, x.head, x.args...)
+    elseif x.args[1] != :call  || !is_one_argument_constraint(Val(x.args[1]))
+        vectorized, parsecode, buildcall = parsefun(_error, x.args...)
```
blegat: I tend to prefer having simpler macro code and move the complexity to method dispatch.
Here, you could change
https://github.com/JuliaOpt/JuMP.jl/blob/ed22a474d0285cf3993ecb75e57a70408c5c8316/src/macros.jl#L179-L182
into
```julia
function parse_constraint(_error::Function, sense::Symbol, args...)
    (sense, vectorized) = _check_vectorized(sense)
    vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), args...)...
end
```

dourouc05: This allows to merge two definitions of `parse_constraint`. However, I don't think it's enough: for `:=`, the dispatch must be done on the expression's head.

blegat: Could you give a macro call that dispatches to the first one with `:=`

dourouc05: Sorry, I missed your comment. Here are a few examples: 

```julia
@constraint(m, x := { y <= 4 })
@constraint(m, x := { alldifferent(y, z) })
```

blegat: Then a simple and easy to extend solution would be to call
```julia
parsefun(_error, Val(x.head), x.args...)
```
and then have a
```julia
parse_constraint(_error::Function, ::Val{:call}, args...) = parse_constraint(_error, args...)
```
Maybe we could have a different name for the `parse_constraint` that takes the `Val` if that simplifies dispatch

dourouc05:
@blegat I made your solution work, it is in the current version of this PR. I'm not sure that `parse_call_constraint` is strictly necessary, it would be enough to have a `::Val{:call}` on all these methods, but I find this solution cleaner. 

With this, I have the majority of the extension points I need, with very little code in JuMP. It would be nice to allow a syntax like `@constraint(m, x == count(y, 1))` or `@constraint(m, x == count(y .== 1))`, a much more natural syntax, but it would require deeper changes. I'm having a look at it.

src/macros.jl
```diff
@@ -156,29 +156,47 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
...
-    set = sense_to_set(_error, sense)
-    variable, parse_code = _MA.rewrite(func)
-    return parse_code, _build_call(_error, vectorized, :(_functionize($variable)), set)
+
+    build_code = _build_call(_error, vectorized, :(_functionize($variable)), set)
+    return :($parse_code_rhs; $parse_code), :($build_code_rhs; $build_code)
+end
+
+function parse_constraint(_error::Function, ::Val{:call}, args...)
+    return parse_call_constraint(_error, Val(args[1]), args[2:end]...)
```
blegat: I would rather have this call `parse_constraint(_error, args...)` to avoid breaking existing code

dourouc05: How can you then dispatch on the function being called (i.e. `args[1]`)?

blegat: You can just keep the existing `parse_constraint(_error::Function, sense::Symbol, lhs, rhs)`, the call will be dispatched to it, isn't it ?

dourouc05: Sorry, I skipped your comment… I don't see how Julia could dispatch on a symbol without Val. Or would you rather have this function called `parse_constraint`?

blegat: It is dispatched in `parse_one_operator_constraint` as it is currently.

dourouc05: Is the new version what you meant ?

odow:
Can this be closed? Experimentation for constraint programming should take place in https://github.com/dourouc05/ConstraintProgrammingExtensions.jl first.

odow:
What I mean is, we should have MOI sets, along with solvers that support them first. And once that is done, we can have another think about how best to support them via JuMP.

dourouc05:
There are already sets, the first solver binding is being written, so it's just a matter of time, and it shouldn't be long. Nevertheless, this PR has nothing related to constraint programming (apart from its title); the new functionalities could be used by totally unrelated projects.

src/macros.jl
```diff
@@ -156,29 +156,47 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
     return parse_code, _build_call(_error, vectorized, variable, set)
 end
 
+function rewrite_call_expression(_error::Function, head::Val{F}, args...) where F
```
blegat, line 159: Could this part be moved to a separate PR ?

blegat:
IIUC, you want to allow 3 new syntaxes in `@constraint`:
1) Allow non-`:call` expressions, could you provide an example use case for this one ?
2) Allow `:call` expressions with new functions, e.g. `@constraint(model, alldifferent(x, y, z))`.
3) Allow comparison with new expressions, e.g. `@constraint(model, x >= min{y, z})`.

It would be easier to work on this with 3 separate PR's.

src/macros.jl
dourouc05: Here it is: https://github.com/JuliaOpt/JuMP.jl/pull/2229

dourouc05:
Actually, I'm starting to get lost in the implications of various parts of this PR… For now, I think that #2228 and #2229 would cover all this original PR was about.

The next step would be to allow to rewrite more expressions, to allow things like `@constraint(m, x == y[z])` where at least one of `y` and `z` is a JuMP variable (either an integer index or a vector). However, to do this, you need information which does not seem to be available to a macro… (Maybe with generated functions? They work like macros, but have dispatch on the type of the arguments.)

dourouc05:
I've just rewritten the PR to bring back the functionality of having constraints with just a function (i.e. point 2 for @blegat's message), now with a test case.

dourouc05:
All PRs for new syntax (this one and #2229) have been merged in: https://github.com/dourouc05/JuMP.jl/tree/constraints

test/macros.jl
```diff
@@ -176,6 +176,20 @@ function custom_expression_test(ModelType::Type{<:JuMP.AbstractModel})
     end
 end
 
+function JuMP.parse_constraint_call(_error::Function, ::Val{:f}, x)
```
blegat, line 179: Why not `JuMP.parse_constraint_head(_error::Function, ::Val{:call}, ::Val{:f}, x)` ?

dourouc05: If I remove my changes in `src/macros.jl` and adapt the test as you suggest, this is what I get: 

```
ERROR: LoadError: LoadError: LoadError: In `@constraint(model, con_ref, f(x))`: Constraints must be in one of the following forms:
       expr1 <= expr2
       expr1 >= expr2
       expr1 == expr2
       lb <= expr <= ub
Stacktrace:
 [1] error(::String, ::String) at .\error.jl:42
 [2] _macro_error(::Symbol, ::Array{Any,1}, ::String) at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:891
 [3] (::JuMP.var"#_error#68"{Symbol})(::String) at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:363
 [4] _unknown_constraint_expr(::JuMP.var"#_error#68"{Symbol}) at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:240
 [5] parse_constraint(::Function, ::Symbol, ::Symbol) at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:249
 [6] parse_constraint_head(::Function, ::Val{:call}, ::Symbol, ::Symbol) at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:184
 [7] parse_constraint_expr at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:181 [inlined]
 [8] _constraint_macro(::Tuple{Symbol,Symbol,Expr}, ::Symbol, ::typeof(parse_constraint_expr)) at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:404
 [9] @constraint(::LineNumberNode, ::Module, ::Vararg{Any,N} where N) at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:487
 [10] include(::String) at .\client.jl:439
 [11] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\test\runtests.jl:23
 [12] include(::String) at .\client.jl:439
 [13] top-level scope at none:6
in expression starting at C:\Users\Thibaut\.julia\dev\JuMP\test\macros.jl:186
in expression starting at C:\Users\Thibaut\.julia\dev\JuMP\test\macros.jl:182
in expression starting at C:\Users\Thibaut\.julia\dev\JuMP\test\runtests.jl:23
```

src/macros.jl
```diff
@@ -188,6 +188,15 @@ function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
     (sense, vectorized) = _check_vectorized(sense)
     vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), lhs, rhs)...
 end
+function parse_constraint(_error::Function, ::Val{:call}, args...)
```
blegat, line 191: Sorry, my recommendation doesn't work. Why not replace
```julia
function parse_constraint(_error::Function, args...)
    _unknown_constraint_expr(_error)
end
```
by
```julia
function parse_constraint(_error::Function, args...)
    if args[1] isa Symbol
        (sense, vectorized) = _check_vectorized(args[1])
        vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), args[2:end]...)...
    else
        _unknown_constraint_expr(_error)
    end
end
```
and then in the tests, do
```julia
function JuMP.parse_one_operator_constraint(_error::Function, ::Val{:f}, x)
    return :(), :(build_constraint($_error, $(esc(x)), $(esc(CustomType()))))
end
```

dourouc05: At least, this solution allows the test to pass, so I guess this is enough, thanks for your continued support :)!

src/macros.jl
```diff
@@ -188,6 +188,15 @@ function parse_constraint(_error::Function, sense::Symbol, lhs, rhs)
...
+    return parse_constraint(_error, Val(args[1]), args[2:end]...)
+end
+
+function parse_constraint(_error::Function, sense::Val, F...)
```
blegat, line 195: Why are these two methods needed ?

src/macros.jl
```diff
@@ -237,7 +246,16 @@ function parse_constraint_head(_error::Function, ::Val, args...)
...
 end
 function parse_constraint(_error::Function, args...)
-    _unknown_constraint_expr(_error)
+    # Define this as the last fallback: either this is a function call that may
+    # be overridden by extensions, or a syntax that is not recognised.
+    # Multiple dispatch does not work here, due to ambiguity with:
+    #     parse_constraint(_error::Function, lb, lsign::Symbol, aff, rsign::Symbol, ub)
+    if args[1] isa Symbol
+        (sense, vectorized) = _check_vectorized(args[1])
+        vectorized, parse_one_operator_constraint(_error, vectorized, Val(sense), args[2:end]...)...
```
blegat, line 255: I guess that
```julia
julia> @constraint(model, g(1))
ERROR: LoadError: In `@constraint(model, g(1))`: Constraints must be in one of the following forms:
       expr1 <= expr2
       expr1 >= expr2
       expr1 == expr2
       lb <= expr <= ub
Stacktrace:
 [1] error(::String, ::String) at ./error.jl:42
 [2] _macro_error(::Symbol, ::Array{Any,1}, ::String) at /home/blegat/.julia/packages/JuMP/MnJQc/src/macros.jl:856
 [3] (::JuMP.var"#_error#68"{Symbol})(::String) at /home/blegat/.julia/packages/JuMP/MnJQc/src/macros.jl:331
 [4] parse_constraint(::JuMP.var"#_error#68"{Symbol}, ::Symbol, ::Int64) at /home/blegat/.julia/packages/JuMP/MnJQc/src/macros.jl:220
 [5] _constraint_macro(::Tuple{Symbol,Expr}, ::Symbol, ::Function) at /home/blegat/.julia/packages/JuMP/MnJQc/src/macros.jl:372
 [6] @constraint(::LineNumberNode, ::Module, ::Vararg{Any,N} where N) at /home/blegat/.julia/packages/JuMP/MnJQc/src/macros.jl:455
in expression starting at REPL[4]:1
```
now throws a MethodError.
Can you as it as a test, and add a fallback method for `parse_one_operator_constraint` that calls `_unknown_constraint_expr` ?

dourouc05: Is the new commit good enough?

src/macros.jl
dourouc05: Indeed, they are not needed, I've just removed them.

test/macros.jl
```diff
@@ -176,6 +176,22 @@ function custom_expression_test(ModelType::Type{<:JuMP.AbstractModel})
...
+function custom_function_test(ModelType::Type{<:JuMP.AbstractModel})
+    @testset "Custom function" begin
+        model = ModelType()
+        @variable(model, x)
+        @constraint(model, con_ref, f(x))
+        con = JuMP.constraint_object(con_ref)
+        @test jump_function(con) == x
+        @test moi_set(con) isa CustomSet
+
+        @test_throws LoadError eval(:(@constraint(model, g(x))))
```
blegat, line 191: Use `@test_macro_throws` defined in `test/utilities.jl`.

test/macros.jl
```diff
@@ -176,6 +176,22 @@ function custom_expression_test(ModelType::Type{<:JuMP.AbstractModel})
...
+function custom_function_test(ModelType::Type{<:JuMP.AbstractModel})
+    @testset "Custom function" begin
+        model = ModelType()
+        @variable(model, x)
+        @constraint(model, con_ref, f(x))
+        con = JuMP.constraint_object(con_ref)
+        @test jump_function(con) == x
+        @test moi_set(con) isa CustomSet
+
+        @test_macro_throws LoadError @constraint(model, g(x))
```
blegat, line 191: Replace `LoadError` by `ErrorException`. This is why travis is failing.

dourouc05: I didn't expect the macro to be so smart! Done in the next commit.

Issue #2031, opened 2019-08-18
TimeLimitSec


blegat:
Now that a [TimeLimitSec](https://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#MathOptInterface.TimeLimitSec) attribute in MOI, we could have JuMP function with a syntax more user friendly than `MOI.get(model, MOI.TimeLimitSec())` and `MOI.set(model, MOI.TimeLimitSec(), value)`

Wikunia:
`JuMP.get_time_limit()` and `JuMP.set_time_limit(limit)` ?

blegat:
`time_limit` instead `get_time_limit` would be more consistent with other JuMP functions, e.g. we have `value` instead of `get_value`.

Wikunia:
This can be closed, right?

blegat:
Indeed, thanks!

PR #2053, opened 2019-09-03
[[un]set_]time_limit_sec

Wikunia:
Using `set_time_limit` to set the time limit of the solver
and `time_limit` to retrieve it.
Implements #2031

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -460,7 +460,8 @@ end
 """
     set_time_limit_sec(model::Model, limit)
 
-Sets the time limit (in seconds) of the solver. Can be unset using `unset_time_limit_sec`.
+Sets the time limit (in seconds) of the solver. 
+Can be unset using `unset_time_limit_sec` or with `limit` set to `nothing`.
 """
 function set_time_limit_sec(model::Model, limit)
     return MOI.set(model, MOI.TimeLimitSec(), limit)
@@ -478,7 +479,7 @@ end
 """
     time_limit_sec(model::Model)
 
-Gets the time limit (in seconds) of the model (`nothing` if undefined). Can be set using `set_time_limit_sec`.
+Gets the time limit (in seconds) of the model (`nothing` if unset). Can be set using `set_time_limit_sec`.
 """
 function time_limit_sec(model::Model)
     return MOI.get(model, MOI.TimeLimitSec())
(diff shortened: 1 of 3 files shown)


src/JuMP.jl
```diff
@@ -457,6 +457,24 @@ function unset_silent(model::Model)
     return MOI.set(model, MOI.Silent(), false)
 end
 
+"""
+    set_time_limit(model::Model, limit)
+
+Sets the time limit (in seconds) of the solver.
```
blegat: Explain also how to unset the the time limit. Maybe with a `unset_time_limit`.

test/model.jl
```diff
@@ -382,6 +382,11 @@ function test_model()
         model = Model(with_optimizer(MOIU.MockOptimizer, mock))
         JuMP.set_time_limit(model, 12.0)
         @test JuMP.time_limit(model) == 12.0
+        JuMP.set_time_limit(model, nothing)
+        @test JuMP.time_limit(model) === nothing
+        JuMP.set_time_limit(model, 12.0)
```
blegat: Test that getting time limit here give 12

Wikunia: Isn't that done by line 384? Or shell I do it after setting the time limit to nothing as well? I though `nothing` is the standard so this wouldn't test anything new or am I wrong?

mlubin:
Please use `time_limit_sec` and `set_time_limit_sec`.

test/model.jl
blegat: Yes, indeed, I was just reviewing the diff and didn't see the above lines

src/JuMP.jl
```diff
@@ -457,6 +457,33 @@ function unset_silent(model::Model)
...
+Unsets the time limit of the solver. Can be set using `set_time_limit_sec`.
+"""
+function unset_time_limit_sec(model::Model)
+    return MOI.set(model, MOI.TimeLimitSec(), nothing)
+end
+
+"""
+    time_limit_sec(model::Model)
+
+Gets the time limit (in seconds) of the model (`nothing` if undefined). Can be set using `set_time_limit_sec`.
```
mlubin: s/undefined/unset/

src/JuMP.jl
```diff
@@ -457,6 +457,33 @@ function unset_silent(model::Model)
     return MOI.set(model, MOI.Silent(), false)
 end
 
+"""
+    set_time_limit_sec(model::Model, limit)
+
+Sets the time limit (in seconds) of the solver. Can be unset using `unset_time_limit_sec`.
```
mlubin: Also document that `nothing` unsets the time limit (or disallow this behavior).

PR #2063, opened 2019-09-17
Fix product of sparse matrices for Julia v1.3

### LINKED ISSUE #2042: Style: interpolated variables in strings need parentheses

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1276
Closes https://github.com/JuliaOpt/JuMP.jl/issues/2054
Closes #2042 
Closes #2043

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -395,6 +395,8 @@ end
 =#
 
 const TransposeOrAdjoint{T<:Union{GenericAffOrQuadExpr, Real}, MT} = Union{Transpose{T, MT}, Adjoint{T, MT}}
+_mirror_transpose_or_adjoint(x, ::Transpose) = transpose(x)
+_mirror_transpose_or_adjoint(x, ::Adjoint) = adjoint(x)
 
 function _mul!(ret::AbstractVecOrMat{<:GenericAffOrQuadExpr},
                adjA::TransposeOrAdjoint{<:Any, <:SparseMatrixCSC},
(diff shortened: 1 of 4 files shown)


src/operators.jl
```diff
@@ -449,23 +452,54 @@ function _mul!(ret::AbstractVecOrMat{<:GenericAffOrQuadExpr},
 end
 
 function _mul!(ret::AbstractMatrix{<:GenericAffOrQuadExpr},
-               A::AbstractMatrix, B::SparseMatrixCSC)
-    _fill_with_zeros!(ret)
+               A::AbstractMatrix, B::SparseMatrixCSC,
+               α_expr=one(eltype(ret)),
+               β=zero(eltype(ret)))
```
odow: When are these called with anything other than the defaults?

odow:
You can probably add
```
Closes #2042 
Closes #2043
```
It doesn't look like we'll move forward, given the resolution up-stream. It's something `JuliaFormatter` can handle as well.

src/operators.jl
blegat: I can't find anywhere it's called with values different from alpha=one and beta=zero in `SparseArrays/src/linalg.jl`. I don't know where it's documented either

PR #2069, opened 2019-09-22
Remove SCIP from list of outdated solvers

blegat:
cc @rschwarz


rschwarz:
Thanks. Outdated here means not supporting JuMP v0.19, right?

blegat:
Yes, it means not supporting MOI. I forgot to remove it in https://github.com/JuliaOpt/JuMP.jl/pull/1967

PR #2070, opened 2019-09-23
Create containers with map instead of for loops

blegat:
Currently, when creating a container, JuMP macros first allocate the container and then fill it with for loops.
The disadvantage of this approach is that we need to determine the type beforehand so the `eltype` is for instance not concrete for `@constraint` and `@expression` only works for `AffExpr` (see https://github.com/JuliaOpt/JuMP.jl/issues/525).
With this PR, we rely on `map` to create the array holding the values of the container hence we don't need to determine the `eltype` beforehand hence also removing the need for `variable_type` and `constraint_type` (but we should keep them to avoid making a breaking change).

The PR also lays out foundations for creating constrained variables as it was needed to update the creation of SDP and symmetric matrix of variables with the new design.

As a proof of concept that the container logic is now completely decoupled from JuMP, the `Containers` submodule has now a `@container`.

- [x] Fix tests.
- [x] Update doc.

Closes https://github.com/JuliaOpt/JuMP.jl/issues/525

## Benchmarks

For the `test/perf/speed.jl` benchmark, before the PR:
```julia
julia> include("speed.jl")
P-Median(100 facilities, 100 customers, 5000 locations) benchmark:
BenchmarkTools.Trial: 
  memory estimate:  703.37 MiB
  allocs estimate:  11015593
  --------------
  minimum time:     1.624 s (51.79% GC)
  median time:      2.267 s (57.83% GC)
  mean time:        2.137 s (58.00% GC)
  maximum time:     2.519 s (62.15% GC)
  --------------
  samples:          3
  evals/sample:     1

Cont5(n=500) benchmark:
BenchmarkTools.Trial: 
  memory estimate:  365.94 MiB
  allocs estimate:  5779814
  --------------
  minimum time:     1.061 s (50.03% GC)
  median time:      1.134 s (46.80% GC)
  mean time:        1.142 s (47.05% GC)
  maximum time:     1.206 s (46.29% GC)
  --------------
  samples:          5
  evals/sample:     1
```
After the PR:
```julia
julia> include("speed.jl")
P-Median(100 facilities, 100 customers, 5000 locations) benchmark:
BenchmarkTools.Trial: 
  memory estimate:  703.39 MiB
  allocs estimate:  11016099
  --------------
  minimum time:     1.533 s (57.35% GC)
  median time:      1.706 s (61.08% GC)
  mean time:        1.693 s (60.98% GC)
  maximum time:     1.826 s (62.87% GC)
  --------------
  samples:          4
  evals/sample:     1

Cont5(n=500) benchmark:
BenchmarkTools.Trial: 
  memory estimate:  365.94 MiB
  allocs estimate:  5779820
  --------------
  minimum time:     761.274 ms (40.83% GC)
  median time:      808.268 ms (41.89% GC)
  mean time:        821.060 ms (44.11% GC)
  maximum time:     883.443 ms (48.30% GC)
  --------------
  samples:          7
  evals/sample:     1
```
The speedup is between 25% and 50% (i.e. 2x) which is nice to see!

|          | Median | Mean |
|----------|--------|------|
| P-Median | 25 %   | 47 % |
| Cont5    | 29 %   | 28 % |

I suspect the speedup to come from the fact that the containers of constraint now have concrete element type. This was the case in JuMP v0.18 IIRC but not anymore for JuMP v0.19 so this PR may help with https://github.com/JuliaOpt/JuMP.jl/issues/1403

For the `test/perf/axis_constraints.jl` benchmark, before this PR:
```julia
julia> dense_axis_constraints(1000)
  0.001575 seconds (8.01 k allocations: 272.016 KiB)
  708.939 μs (13986 allocations: 234.14 KiB)
  700.468 μs (13986 allocations: 234.14 KiB)

julia> dense_axis_constraints(1000)
  0.001671 seconds (8.01 k allocations: 272.016 KiB)
  707.259 μs (13986 allocations: 234.14 KiB)
  711.204 μs (13986 allocations: 234.14 KiB)

julia> sparse_axis_constraints(1000)
  0.001006 seconds (5.83 k allocations: 138.234 KiB)
  361.992 μs (7000 allocations: 117.19 KiB)
  389.133 μs (7193 allocations: 120.20 KiB)

julia> sparse_axis_constraints(1000)
  0.001039 seconds (5.83 k allocations: 138.234 KiB)
  361.963 μs (7000 allocations: 117.19 KiB)
  390.414 μs (7193 allocations: 120.20 KiB)
```
After this PR:
```julia
julia> dense_axis_constraints(1000)
  0.000709 seconds (8.02 k allocations: 272.281 KiB)
  669.011 μs (12987 allocations: 218.53 KiB)
  668.274 μs (12987 allocations: 218.53 KiB)

julia> dense_axis_constraints(1000)
  0.001555 seconds (8.02 k allocations: 272.281 KiB)
  666.685 μs (12987 allocations: 218.53 KiB)
  670.274 μs (12987 allocations: 218.53 KiB)

julia> dense_axis_constraints(1000)
  0.001748 seconds (8.02 k allocations: 272.281 KiB)
  681.249 μs (12987 allocations: 218.53 KiB)
  674.972 μs (12987 allocations: 218.53 KiB)

julia> sparse_axis_constraints(1000)
  0.001240 seconds (5.78 k allocations: 145.906 KiB)
  352.693 μs (6500 allocations: 109.38 KiB)
  351.343 μs (6500 allocations: 109.38 KiB)

julia> sparse_axis_constraints(1000)
  0.001090 seconds (5.78 k allocations: 145.906 KiB)
  347.866 μs (6500 allocations: 109.38 KiB)
  355.565 μs (6500 allocations: 109.38 KiB)

julia> sparse_axis_constraints(1000)
  0.001250 seconds (5.78 k allocations: 145.906 KiB)
  348.161 μs (6500 allocations: 109.38 KiB)
  359.230 μs (6500 allocations: 109.38 KiB)
```
We see a speedup on computing the sum of the duals and less allocation because the `eltype` of the containers are concrete.


src/Containers/macro.jl
```diff
@@ -0,0 +1,187 @@
...
+    else
+        return :(Containers.container($func, $indices, $requested_container))
+    end
+end
+function parse_container(_error, var, value, requested_container)
+    idxvars, indices = _build_ref_sets(_error, var)
+    return container_code(idxvars, indices, value, requested_container)
+end
+
+macro container(args...)
```
mlubin: Missing docs for this macro.

src/Containers/container.jl
```diff
@@ -0,0 +1,41 @@
...
+function container(f::Function, indices::ArrayIndices, ::Type{Array})
+    return map(I -> f(I...), indices)
+end
+function _oneto(indices)
+    if indices isa UnitRange{Int} && indices == 1:length(indices)
+        return Base.OneTo(length(indices))
+    end
+    error("Index set for array is not one-based interval.")
+end
+function container(f::Function, indices::VectorizedProductIterator,
```
mlubin: `container` looks important enough to have docs

src/Containers/vectorized_product_iterator.jl
```diff
@@ -0,0 +1,32 @@
+"""
+    struct VectorizedProductIterator{T}
+        prod::Iterators.ProductIterator{T}
+    end
+
+Same as `Base.Iterators.ProductIterator` except that it is independent
+on the `IteratorSize` of the elements of `prod.iterators`.
```
mlubin: "dependent on" or "independent of"?

mlubin: Could this description be made more self contained? Currently it refers to an undocumented internal field of `ProductIterator`.

src/Containers/nested_iterator.jl
```diff
@@ -0,0 +1,67 @@
...
+        for i3 in iterator[3](i1, i2)
+            if condition(i1, i2, i3)
+                # produces (i1, i2, i3)
+            end
+        end
+    end
+end
+```
+"""
+struct NestedIterator{T}
```
mlubin: This iterator could use unit tests.

src/Containers/vectorized_product_iterator.jl
```diff
@@ -0,0 +1,32 @@
...
+For instance:
+* `size(Iterators.product(1, 2))` is `tuple()` while
+  `size(VectorizedProductIterator(1, 2))` is `(1, 1)`.
+* `size(Iterators.product(ones(2, 3)))` is `(2, 3)` while
+  `size(VectorizedProductIterator(ones(2, 3)))` is `(1, 1)`.
+"""
+struct VectorizedProductIterator{T}
```
mlubin: This iterator could use unit tests.

src/Containers/vectorized_product_iterator.jl
```diff
@@ -1,10 +1,16 @@
...
 """
     struct VectorizedProductIterator{T}
         prod::Iterators.ProductIterator{T}
     end
 
-Same as `Base.Iterators.ProductIterator` except that it is independent
-on the `IteratorSize` of the elements of `prod.iterators`.
+Cartesian product of the iterators `prod.iterators`. It is the same iterator as
+`Base.Iterators.ProductIterator` except that it is independent of the
+`IteratorSize` of the elements of `prod.iterators`.
```
mlubin: I think this could benefit from a longer comment about why the behavior of `Base.Iterators.ProductIterator` wasn't sufficient and required a wrapper. (Maybe outside the docstring.) What breaks if we use `Base.Iterators.ProductIterator`?

blegat: In fact, every test passed with `Iterators.ProductIterator` but
https://github.com/JuliaOpt/JuMP.jl/blob/1a4f663504606464292caf2db2d60acba6991af3/examples/cutting_stock_column_generation.jl#L149
fails.
Now I have added tests in `test/Containers/` covering this feature.

blegat:
I added benchmarks in the initial PR comment.

mlubin:
The numbers from `test/perf/speed.jl` look great. However, they don't appear to cover `SparseAxisArray` or `DenseAxisArray`. Could you also test those?

blegat:
There is a small performance hit for creating `SparseAxisArray`. The problem is
```julia
function container(f::Function, indices,
                   ::Type{SparseAxisArray})
    mappings = map(I -> I => f(I...), indices)
    data = Dict(mappings)
    if length(mappings) != length(data)
        error("Repeated index ...")
    end
    return SparseAxisArray(Dict(data))
end
```
because we allocate `mappings`. Instead, we could do
```julia
function container(f::Function, indices,
                   ::Type{SparseAxisArray})
    mappings = Base.Generator(I -> I => f(I...), indices)
    data = Dict(mappings)
    if length(mappings) != length(data)
        error("Repeated index ...")
    end
    return SparseAxisArray(Dict(data))
end
```
but then `length(mappings)` is not O(1) anymore.

src/Containers/no_duplicate_dict.jl
```diff
@@ -0,0 +1,40 @@
...
+
+"""
+    struct NoDuplicateDict{K, V} <: AbstractDict{K, V}
+        dict::Dict{K, V}
+    end
+
+Same as `Dict{K, V}` but errors if constructed from an iterator with duplicate
+keys.
+"""
+struct NoDuplicateDict{K, V} <: AbstractDict{K, V}
```
mlubin: Do we need to define a new type that's only used in one function? What about a function that takes a generator, constructs a `Dict` with errors on duplicates, and returns `Dict`?

blegat: > constructs a `Dict` with errors on duplicates

How do you do that ? That's exactly `NoDuplicateDict` is trying to do

blegat:
There seems to be something fishy with `NestedIterator`:
```julia
julia> f() = (it = Containers.nested(() -> 1:1000, condition = iseven); @time map(i -> 1, it); return)
f (generic function with 2 methods)

julia> f()
  0.020841 seconds (79.05 k allocations: 3.991 MiB)

julia> f()
  0.000392 seconds (3.50 k allocations: 125.422 KiB)

julia> f() = (it = Iterators.filter(iseven, 1:1000); @time map(i -> 1, it); return)
f (generic function with 2 methods)

julia> f()
  0.000013 seconds (10 allocations: 8.406 KiB)

julia> f()
  0.000023 seconds (10 allocations: 8.406 KiB)
```

blegat:
The performance issue with `SparseAxisArray` is now resolved, I have updated the benchmarks.

src/Containers/no_duplicate_dict.jl
mlubin: I didn't dig too deeply, but my question was why you can't define a function like:
```julia
function dict_no_duplicates{K, V}(it) where {K, V}
    dict = Dict{K, V}()
    for (k, v) in it
        if haskey(dict, k)
            error(...)
        else
            dict[k] = v
        end
    end
    return dict
end
```

This doesn't require defining a new type. Maybe I'm missing something related to `dict_with_eltype`.

blegat: The issue is that we don't know the type of the keys and values. The challenge is that Julia wants to infer it in a way that's friendly with Julia compiler and inference system.
So it starts iterating, assumes the type is always the same so it build the container with the type of the first element and then if a new element does not fit, it creates a new container with a wider type, copy and continue.
This is what is done in `map` (that DenseAxisArray rely on) and `dict_with_eltype` (that SparseAxisArray rely on).

PR #2076, opened 2019-10-08
Querying after modifying is undefined behaviour

henriquebecker91:
I do believe the change is self-explanatory. The only issue may be that it is too much text for a smal detail. However, the user needs the documentation to know what is undefined behavior, as if they test it programmatically they can be led to believe it works some way and not
another.

### DIFF
--- a/docs/src/solutions.md
+++ b/docs/src/solutions.md
@@ -121,7 +121,7 @@ end
 
 !!! warning "Querying after modification"
     If a solved model is modified, then querying the solution is undefined
-    behaviour. Adding, deleting, or modifying a constraint (or variable)
+    behavior. Adding, deleting, or modifying a constraint (or variable)
     may invalidate any part of the solution.
 
 ```@meta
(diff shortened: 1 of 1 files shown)


docs/src/solutions.md
```diff
@@ -119,6 +119,12 @@ else
 end
 ```
 
+!!! warning "Querying after modification"
+    If a solved model is modified, then querying the solution is undefined
+    behaviour. Adding, deleting, or modifying a constraint (or variable)
+    may invalidate the any part of the solution (e.g., the values of
```
odow: "invalidate the any" should be replaced by "invalidate any". You can probably also remove the "(e.g., ...)". "Any part" is descriptive enough.

blegat:
@odow Should the optimizer just clear it's result when something is deleted so that the error is "optimize not called" or "result out of bound" instead of getting an arbitrary value ?

odow:
> Should the optimizer just clear it's result when something is deleted so that the error is "optimize not called" or "result out of bound" instead of getting an arbitrary value ?

I've resisted this because it's going to require a lot of solvers to set a flag and check/set that on every call. But it might be better than undefined behavior.

blegat:
Still not convinced that it's not better to just add MOI tests, there are not so many solvers supporting modification. On the short term, adding this warning is ok

henriquebecker91:
Surely it would make the debug easier for people that fall for this pitfall. It really would add a lot of code as odow mentioned. I have no time to contribute to it for now, but I may try to help after 30 november.

PR #2079, opened 2019-10-11
Now deepcopy(::AbstractModel) calls error

henriquebecker91:
There is no plan to support copying the underlying solver and, if this is not done, then the object returned by deepcopy does not behave as would be expected (both copy and original edit the same underlying model).
This has been a source of frustration before:
https://discourse.julialang.org/t/jump-gurobi-bug-value-of-variable-in-solved-model-changes-after-deleting-another-variable/29619/7
https://discourse.julialang.org/t/strange-behavior-with-deepcopy/23342/3
https://discourse.julialang.org/t/workaround-for-modifying-coefficients-in-constraints-in-jump/27123/9

### DIFF
--- a/src/copy.jl
+++ b/src/copy.jl
@@ -138,14 +138,10 @@ function Base.copy(model::AbstractModel)
     return new_model
 end
 
-"""
-    deepcopy(_::AbstractModel)
-
-Calls error. Calling `deepcopy` over a JuMP model is not supported, nor
-planned to be supported, because it would involve making a deep copy of
-the underlying solver (behind a C pointer). Check [`copy`](@ref) instead.
-"""
-function Base.deepcopy(_::AbstractModel)
-    error("JuMP.AbstractModel does not support deepcopy, see documentation")
+# Calling `deepcopy` over a JuMP model is not supported, nor planned to be
+# supported, because it would involve making a deep copy of the underlying
+# solver (behind a C pointer).
+function Base.deepcopy(_::Model)
+    error("JuMP.AbstractModel does not support deepcopy as the reference to the underlying solver cannot be deep copied, use copy instead.")
 end
 
--- a/src/copy.jl
+++ b/src/copy.jl
@@ -141,7 +141,6 @@ end
 # Calling `deepcopy` over a JuMP model is not supported, nor planned to be
 # supported, because it would involve making a deep copy of the underlying
 # solver (behind a C pointer).
-function Base.deepcopy(_::Model)
+function Base.deepcopy(::Model)
     error("JuMP.AbstractModel does not support deepcopy as the reference to the underlying solver cannot be deep copied, use copy instead.")
 end
-
--- a/src/copy.jl
+++ b/src/copy.jl
@@ -142,5 +142,5 @@ end
 # supported, because it would involve making a deep copy of the underlying
 # solver (behind a C pointer).
 function Base.deepcopy(::Model)
-    error("`JuMP.AbstractModel` does not support `deepcopy` as the reference to the underlying solver cannot be deep copied, use `copy` instead.")
+    error("`JuMP.Model` does not support `deepcopy` as the reference to the underlying solver cannot be deep copied, use `copy` instead.")
 end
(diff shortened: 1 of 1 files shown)


src/copy.jl
```diff
@@ -137,3 +137,15 @@ function Base.copy(model::AbstractModel)
     new_model, _ = copy_model(model)
     return new_model
 end
+
+"""
+    deepcopy(_::AbstractModel)
+
+Calls error. Calling `deepcopy` over a JuMP model is not supported, nor
```
mlubin: I don't see a need to have a docstring. The error message should be reasonably descriptive, and if necessary we can have a comment above the error explaining the technical issue with copying solvers that we have C pointers to.

src/copy.jl
```diff
@@ -137,3 +137,15 @@ function Base.copy(model::AbstractModel)
...
+planned to be supported, because it would involve making a deep copy of
+the underlying solver (behind a C pointer). Check [`copy`](@ref) instead.
+"""
+function Base.deepcopy(_::AbstractModel)
```
mlubin: The issue only applies for `Model`, not necessarily `AbstractModel`.

mlubin: `Base.deepcopy(::AbstractModel)` instead of `Base.deepcopy(_::AbstractModel)`

odow: This approach forces people sub-typing `AbstractModel` to explicitly opt-in to `deepcopy` which I think is a good thing.

mlubin: Yes, but `deepcopy` is opt-out in Julia, so not sure if we should override that just for `JuMP.AbstractModel`.

mlubin: Unless we want to create a style guideline about `deepcopy`.

src/copy.jl
```diff
@@ -137,3 +137,11 @@ function Base.copy(model::AbstractModel)
...
     return new_model
 end
+
+# Calling `deepcopy` over a JuMP model is not supported, nor planned to be
+# supported, because it would involve making a deep copy of the underlying
+# solver (behind a C pointer).
+function Base.deepcopy(_::Model)
+    error("JuMP.AbstractModel does not support deepcopy as the reference to the underlying solver cannot be deep copied, use copy instead.")
+end
+
```
blegat: Remove trailing empty line

src/copy.jl
henriquebecker91: So, what will be? I have changed to Model, but I may change back to AbstractModel. I had made the change before seeing @odow's comment. I think  both sides have some points, and no decision will be final anyway, as new subclasses of AbstractModel may define specializations of `deepcopy`. This is basically my first contribution, so I think the decision should not be mine.

odow: Leave it as `Model`. You contribution is appreciated! Sorry for the back-and-forward.

henriquebecker91: I think it can be merged then. I have left it as Model, and changed the parameter style and removed the trailing line, as asked by others. I did not intend to sound rude, XD, it is just I was worried the merge was waiting because you people were expecting my input.

henriquebecker91: Done.

src/copy.jl
henriquebecker91: Done.

src/copy.jl
```diff
@@ -137,3 +137,10 @@ function Base.copy(model::AbstractModel)
     new_model, _ = copy_model(model)
     return new_model
 end
+
+# Calling `deepcopy` over a JuMP model is not supported, nor planned to be
+# supported, because it would involve making a deep copy of the underlying
+# solver (behind a C pointer).
+function Base.deepcopy(::Model)
+    error("JuMP.AbstractModel does not support deepcopy as the reference to the underlying solver cannot be deep copied, use copy instead.")
```
blegat: Use backsticks around code, i.e. `JuMP.AbstractModel`, `deepcopy`, `copy`

henriquebecker91: In the error message string you say? Or in the comment? Or both?
I did not know the error message made use of this formatting.

blegat: > I did not know the error message made use of this formatting.

It does not but we expect the developer reading the error message to understand the meaning. That is consistent with what we do in other places in JuMP and MOI.

PR #2080, opened 2019-10-11
Add Tulip.jl to list of supported solvers

mtanneau:


blegat:
Add it in the README too

PR #2082, opened 2019-10-18
Fill NEWS entry for v0.20.1

blegat:


NEWS.md
```diff
@@ -1,6 +1,17 @@
 JuMP release notes
 ==================
 
+Version 0.20.1 (Oct 18, 2019)
+-----------------------------
+
+- Documentation updates.
+- Fix product of sparse matrices for Julia v1.3 (#2063).
+- Always use concrete element type for containers created by macros (#2070).
```
mlubin: This could be considered a breaking change. User code that depends on the types of objects returned by the macros could break.

blegat: Let's backport the other PR's then

PR #2084, opened 2019-10-19
Document set_optimizer and remove optimizer argument from optimize!

### LINKED ISSUE #2073: set_optimizer is undocumented
http://www.juliaopt.org/JuMP.jl/v0.20.0/search/?q=set_optimizer yields no results.

### LINKED ISSUE #2074: Switching solver malfunctioning
Previous to v.0.19 to switch from a solver to another, or modify the solver parameter, all we have to do is: (the following is an example)
```julia
mp = Model(solver = GurobiSolver(OutputFlag = 0));
mp.solver = IpoptSolver(print_level = 0);
```
Once JuMP was upgraded, MOI employs the solver in this way:
```julia
mp = Model();
optimize!(with_optimizer(Gurobi.Optimizer, OutputFlag = 0));
```

Now if I change the solver by running optimize! again, it throws me an error. Here is the code I run to generate this error:
```julia
using JuMP, Ipopt, Gurobi;
mp = Model();
@variable(mp, x1 >= 0);
@variable(mp, x2 >= 0);
@constraint(mp, x1 + 2x2 <= 1);
@constraint(mp, x2 + 2x1 <= 1);
@objective(mp, Max, x1 + x2);
optimize!(mp, with_optimizer(Gurobi.Optimizer));
optimize!(mp, with_optimizer(Ipopt.Optimizer));
```
Here is the printout of the error. The first solve is successful and then the second fails. If I want to solve the model again, I have to use Gurobi.Optimizer without changing any parameters/to another solver. It appears the same if I set up an optimizer when setting up the JuMP model instead of leaving it blank initially.

> ERROR: An optimizer factory cannot both be provided in the `Model` constructor and at the `optimize` call.
> Stacktrace:
>  [1] error(::String) at ./error.jl:33
>  [2] #optimize!#78(::Bool, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(optimize!), ::Model, ::OptimizerFactory) at /home/haoxiang/.julia/packages/JuMP/iGamg/src/optimizer_interface.jl:121
>  [3] optimize!(::Model, ::OptimizerFactory) at /home/haoxiang/.julia/packages/JuMP/iGamg/src/optimizer_interface.jl:111
>  [4] top-level scope at REPL[162]:1

mlubin:
Closes #2073
Closes #2074

### DIFF
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -32,6 +32,24 @@ function MOIU.attach_optimizer(model::Model)
     MOIU.attach_optimizer(backend(model))
 end
 
+"""
+    set_optimizer(model::Model, optimizer_factory::OptimizerFactory;
+                  bridge_constraints::Bool=true)
+
+Creates a new `MathOptInterface.AbstractOptimizer` instance using the optimizer
+factory and sets it as the optimizer of `model`.
+
+If `bridge_constraints` is true, constraints that are not supported by the
+optimizer are automatically bridged to equivalent supported constraints when
+an appropriate transformation is defined in the `MathOptInterface.Bridges`
+module or is defined in another module and is explicitly added.
+
+## Examples
+```julia
+model = Model()
+set_optimizer(model, with_optimizer(GLPK.Optimizer))
+```
+"""
 function set_optimizer(model::Model, optimizer_factory::OptimizerFactory;
                        bridge_constraints::Bool=true)
     error_if_direct_mode(model, :set_optimizer)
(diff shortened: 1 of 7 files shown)


src/optimizer_interface.jl
```diff
@@ -114,15 +112,9 @@ function optimize!(model::Model,
     end
 
     if optimizer_factory !== nothing
-        if mode(model) == DIRECT
-            error("An optimizer factory cannot be provided at the `optimize` call in DIRECT mode.")
-        end
-        if MOIU.state(backend(model)) != MOIU.NO_OPTIMIZER
-            error("An optimizer factory cannot both be provided in the `Model` constructor and at the `optimize` call.")
```
blegat: We could also just remove this condition and `optimize!(model, with_optimizer(...))` would be a shortcut for `set_optimizer(with_optimizer(...)); optimizer!(model)`. What would be wrong with that ?

mlubin: This won't work when we remove `with_optimizer`, which I would like to do after this PR. I can toss this PR and go straight to removing `with_optimizer` if that makes more sense.

blegat: > This won't work when we remove `with_optimizer`

You mean when we replace `Model(with_optimizer(GLPK.Optimizer))` by `Model(GLPK.Optimizer)` ? Why wouldn't `optimize!(model, GLPK.Optimizer)` work ?

mlubin: The non-obvious part is how to deal with parameters. We could take a dictionary and automate the call to `set_parameter` for each key/value in the dictionary. 

However, I think there's a bigger issue this API. Keeping the discussion to `with_optimizer`, consider:
```julia
optimize!(model, with_optimizer(GLPK.Optimizer, some_parameter=some_value))
...
optimize!(model)
```
It's not really clear that the solver or parameter choice in the first call to `optimize!` would apply to the second call. Indeed, this is inconsistent with cvxpy and Convex.jl where I believe that the second call would fall back to a default solver with default parameters instead of the last solver used. Hiding the fact that the JuMP model is stateful with respect to the optimizer will cause more confusion like https://github.com/JuliaOpt/JuMP.jl/issues/2074.

blegat: The only way to make it non-confusing is to keep a flag in the JuMP Model that the optimizer should be discarded before the next `optimize!` call (we need to keep it until then in case the user query results). I agree that we don't really want to go in that direction.

PR #2090, opened 2019-10-29
Deprecate with_optimizer

mlubin:
`with_optimizer(Ipopt.Optimizer)` becomes simply `Ipopt.Optimizer`

`set_optimizer(model, with_optimizer(Ipopt.Optimizer, print_level=0))` becomes `set_optimizer(model, Ipopt.Optimizer); set_parameter(model, "print_level", 0)`

`set_optimizer(model, with_optimizer(Gurobi.Optimizer, gurobi_env))` becomes `set_optimizer(model, () -> Gurobi.Optimizer(gurobi_env))`. (This should be a relatively rare case.)

Motivation:
The ability to pass solver parameters to solver constructors is now redundant with `set_parameter`. Passing these through `with_optimizer` is cute syntax but IMO worse than `set_parameter` because 1) the syntax is a bit surprising (it's the only instance of a [fluent](https://en.wikipedia.org/wiki/Fluent_interface)-like API in JuMP), and 2) the parameters are harder to access and manipulate inside of the `OptimizerFactory`. Solvers should stop accepting parameters in their constructors once `with_optimizer` is fully removed (in JuMP 0.22?).

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -423,13 +423,16 @@ end
     set_parameters(model::Model, pairs::Pair...)
 
 Given a list of `parameter_name => value` pairs, calls
-`set_parameters(model, parameter_name, value)` for each pair. This is a
-convenience function only. See [`set_parameter`](@ref).
+`set_parameter(model, parameter_name, value)` for each pair. See
+[`set_parameter`](@ref).
 
 ## Example
 ```julia
 model = Model(Ipopt.Optimizer)
 set_parameters(model, "tol" => 1e-4, "max_iter" => 100)
+# The above call is equivalent to:
+set_parameter(model, "tol", 1e-4)
+set_parameter(model, "max_iter", 100)
 ```
 """
 function set_parameters(model::Model, pairs::Pair...)
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -46,23 +46,8 @@ include("utils.jl")
 const _MOIVAR = MOI.VariableIndex
 const _MOICON{F,S} = MOI.ConstraintIndex{F,S}
 
-"""
-    OptimizerFactory
-
-User-friendly closure that creates new MOI models. New `OptimizerFactory`s are
-created with [`with_optimizer`](@ref) and new models are created from the
-optimizer factory `optimizer_factory` with `optimizer_factory()`.
-
-## Examples
-
-The following construct an optimizer factory and then use it to create two
-independent `Ipopt.Optimizer`s:
-```julia
-optimizer_factory = with_optimizer(Ipopt.Optimizer, print_level=0)
-optimizer1 = optimizer_factory()
-optimizer2 = optimizer_factory()
-```
-"""
+# OptimizerFactory was deprecated in JuMP 0.21 and should be removed when
+# with_optimizer is removed.
 struct OptimizerFactory
     # The constructor can be
     # * `Function`: a function, or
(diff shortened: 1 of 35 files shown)


rschwarz:
Should the current `Optimizer` constructors also be deprecated in sync? And the new constructors would no longer support keyword arguments?

mlubin:
> Should the current Optimizer constructors also be deprecated in sync?

They shouldn't be deprecated until at least MOI 0.10. Assuming both JuMP 0.20 and 0.21 are based on MOI 0.9, we don't want to break code that's passing parameters through constructors in JuMP 0.20.

> And the new constructors would no longer support keyword arguments?

More precisely, they wouldn't support keyword arguments that can also be set through `JuMP.set_parameter`. They could have keyword arguments for other reasons.

ericphanson:
What about adding a convenience function

```julia
set_parameters(model, first_param = 0, second_param = 2)
```

so that the new API isn't more clunky than the old in the case of many parameters?

mlubin:
@ericphanson Good suggestion. Two comments: 

- Parameters are now identified by [strings or enums](http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#MathOptInterface.RawParameter), not by symbols. Confusing a symbol with a string *usually* works, but there are some solvers whose parameter names don't convert to symbols, e.g., Ipopt's "resto.max_iter". We just removed hacks to around around this so I'd rather not re-introduce the friction between strings and symbols. I don't see how we could convert a symbol to an enum, although I don't think any solvers use enums yet.

- For meta-solvers that take sub-solvers as parameters, we'll want to have a standard way to pass parameters to the sub-solvers. An obvious way to do this is a list of pairs, e.g., `["first_param" => 1, "second_param" => 2]`. If we decide to make this the canonical way to group parameters, this would suggest:
```julia
set_parameters(model, ["first_param" => 1, "second_param" => 2])
```
  This version has one more symbolic layer of complexity (`[]`) than your proposal but is a lot less magical.

rschwarz:
> * Confusing a symbol with a string _usually_ works, but there are some solvers whose parameter names don't convert to symbols, e.g., Ipopt's "resto.max_iter".

Yes, SCIP has that issue too, where parameters have a path-like name using `/` as separators.

ericphanson:
That makes sense. While I feel that for general Julia code, symbols / NamedTuples / keyword arguments all come together nicely as a way to choose parameters, I can see that in this case it's better to allow the same kinds of parameters as the solvers use.

To me, it seems convenient to also accept a varargs form (i.e. `set_parameters(model, "first.param" => 1, "second.param" => 2)`, without the `[]`), with the idea that one could think of there being an implicit `[]` which needs to be made explicit in the case of passing parameters sub-solvers. But maybe you'd prefer things to be explicit.

mlubin:
The varargs form of `set_parameters` makes sense to me. I'll add that.

src/JuMP.jl
```diff
@@ -436,6 +419,25 @@ function set_parameter(model::Model, name, value)
     return MOI.set(model, MOI.RawParameter(name), value)
 end
 
+"""
+    set_parameters(model::Model, pairs::Pair...)
+
+Given a list of `parameter_name => value` pairs, calls
+`set_parameters(model, parameter_name, value)` for each pair. This is a
+convenience function only. See [`set_parameter`](@ref).
```
blegat: What does "convenience function only" means. It seems  to say there is a restriction to using it while I guess you mean to say it's "just I convenience function" (maybe I'm biased by French here)

mlubin: I reworded this, should be clearer now.

PR #2092, opened 2019-11-01
Add indicator syntax to @constraint

### LINKED ISSUE #2049: Add an example of using indicator constraints
Now that this is supported in MOI, we should have an example of how to do it in JuMP. Perhaps also special syntax.

blegat:
Add the syntax
```julia
julia> @constraint(model, a => x + y <= 1)
[a, x + y] ∈ MathOptInterface.IndicatorSet{MathOptInterface.ACTIVATE_ON_ONE,MathOptInterface.LessThan{Float64}}(MathOptInterface.LessThan{Float64}(1.0))

julia> @constraint(model, !a => x + y <= 1)
[a, x + y] ∈ MathOptInterface.IndicatorSet{MathOptInterface.ACTIVATE_ON_ZERO,MathOptInterface.LessThan{Float64}}(MathOptInterface.LessThan{Float64}(1.0))
```
to create indicator constraints.

Closes #2049

### DIFF
--- a/src/indicator.jl
+++ b/src/indicator.jl
@@ -1,3 +1,11 @@
+#  Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file extends JuMP to indicator constraints. It is a good example of how
+# JuMP can be extended.
+
 function build_indicator_constraint(
     _error::Function, variable::JuMP.AbstractVariableRef, constraint::JuMP.ScalarConstraint, ::Type{MOI.IndicatorSet{A}}) where A
 
@@ -34,4 +42,15 @@ function parse_one_operator_constraint(
         buildcall = :(build_indicator_constraint($_error, $(esc(variable)), $rhs_buildcall, $S))
     end
     return rhs_parsecode, buildcall
+end
+
+function constraint_string(
+    print_mode, constraint::VectorConstraint{F, <:MOI.IndicatorSet{A}}) where {F, A}
+    var_str = function_string(print_mode, constraint.func[1])
+    if A == MOI.ACTIVATE_ON_ZERO
+        var_str = "!" * var_str
+    end
+    con = ScalarConstraint(constraint.func[2], constraint.set.set)
+    con_str = constraint_string(print_mode, con)
+    return var_str * " => " * con_str
 end
--- a/src/indicator.jl
+++ b/src/indicator.jl
@@ -1,0 +1,37 @@
+function build_indicator_constraint(
+    _error::Function, variable::JuMP.AbstractVariableRef, constraint::JuMP.ScalarConstraint, ::Type{MOI.IndicatorSet{A}}) where A
+
+    set = MOI.IndicatorSet{A}(moi_set(constraint))
+    return VectorConstraint([variable, jump_function(constraint)], set)
+end
+function _indicator_variable_set(::Function, variable::Symbol)
+    return variable, MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}
+end
+function _indicator_variable_set(_error::Function, expr::Expr)
+    if expr.args[1] == :¬ || expr.args[1] == :!
+        if length(expr.args) != 2
+            _error("Invalid binary variable expression `$(expr)` for indicator constraint.")
+        end
+        return expr.args[2], MOI.IndicatorSet{MOI.ACTIVATE_ON_ZERO}
+    else
+        return expr, MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}
+    end
+end
+function parse_one_operator_constraint(
+    _error::Function, vectorized::Bool, ::Union{Val{:(=>)}, Val{:⇒}}, lhs, rhs)
+
+    variable, S = _indicator_variable_set(_error, lhs)
+    if !(rhs isa Expr)
+        _error("Invalid right-hand side `$(rhs)`.")
+    end
+    rhs_vectorized, rhs_parsecode, rhs_buildcall = parse_constraint(_error, rhs.args...)
+    if vectorized != rhs_vectorized
+        _error("Inconsistent use of `.` in symbols to indicate vectorization.")
+    end
+    if vectorized
+        buildcall = :(build_indicator_constraint.($_error, $(esc(variable)), $rhs_buildcall, $S))
+    else
+        buildcall = :(build_indicator_constraint($_error, $(esc(variable)), $rhs_buildcall, $S))
+    end
+    return rhs_parsecode, buildcall
+end
--- a/src/indicator.jl
+++ b/src/indicator.jl
@@ -7,7 +7,8 @@
 # JuMP can be extended.
 
 function build_indicator_constraint(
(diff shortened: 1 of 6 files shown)


mlubin:
I find the symmetry of `a => x + y <= 1` a bit confusing to read. What about ` a --> x + y <= 1`? This appears to parse correctly.

blegat:
We could add it too and use it for printing. Do you also want to remove support for `=>` ? People might expect it to work

mlubin:
> Do you also want to remove support for => ?

We should only have one operator for indicator constraints, unless there's a very good reason to have two.

> People might expect it to work

For indicator constraints, I find it hard to imagine someone guessing the syntax a-priori. I'd expect them to see an example and copy it. From this perspective, any operator we choose that looks like an implication should be fine.

Maybe we need more opinions here on the use of `>=`, but I'm really finding `X => Y <= Z` confusing to read as "X implies Y is less than Z". It looks more like "both X and Z imply Y".

blegat:
> We should only have one operator for indicator constraints, unless there's a very good reason to have two.

We have several operators for `<=` as we also have the unicode `≤` so I thought we could at least also have: `⇒`.

> `X => Y <= Z` looks more like "both X and Z imply Y".

Indeed, we don't recognize `<=` as the rhs implies the lhs but rather that the the `lhs` is less than the `rhs`. In view of this it would be inconsistent to recognize `=>` as an implication.

mlubin:
I meant that we should only have one ascii and one unicode operator for the
same operation :). ⇒ is a reasonable unicode operator for implies, although
it's a bit small relative to the font size.

matbesancon:
The pair `→ ` with `->` or `-->`would be more consistent IMO. Agree that the ASCII double arrow can be confusing

juan-pablo-vielma:
+1 on having ⇒ (or similar Unicode?) and an ascii version. I am not sure about what the ascii version should be. I don’t like => because = and > already have a meaning and the convention of treating concatenation as “or” is common (i.e. <= ). However, it could be the first thing people may try. Are there any example notations from other languages?

ccoffrin:
I would like there to be an ASCII version.  I also find overloading `=>` confusing in the JuMP DSL.  I do like `->`, I think it's acceptable to overload the standard julia meaning in this context.

Another point to consider, require explicit parentheses (or some other brackets) for the RHS.  e.g.,
```
@constraint(model, a => (x + y <= 1))
@constraint(model, a -> (x + y <= 1))
```

joaquimg:
I was about to write the same as @ccoffrin . I really like the parentheses (or some other brackets).
```julia
@constraint(model, a -> {x + y <= 1})
@constraint(model, a -> (x + y <= 1))
```
I like the `->` !

mlubin:
Parenthesis won't work because they don't change the expression generated by the parser. I don't have a strong opinion on whether we should use braces.

odow:
We discussed this offline, and settled on:
```Julia
@constraint(model, a => {x + y <= 1})
```

The main reason is so that broadcasting still works:
```Julia
@constraint(model, a .=> {A * x .<= b})
```

Writing the indicator constraints without braces is an error:
```Julia
@constraint(model, a .=> A * x .<= b)
```

src/indicator.jl
```diff
@@ -0,0 +1,57 @@
...
+    _error::Function, variable::JuMP.AbstractVariableRef, constraint::JuMP.ScalarConstraint, ::Type{MOI.IndicatorSet{A}}) where A
+
+    set = MOI.IndicatorSet{A}(moi_set(constraint))
+    return VectorConstraint([variable, jump_function(constraint)], set)
+end
+function _indicator_variable_set(::Function, variable::Symbol)
+    return variable, MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}
+end
+function _indicator_variable_set(_error::Function, expr::Expr)
+    if expr.args[1] == :¬ || expr.args[1] == :!
```
mlubin: The ¬ version appears to be undocumented.

src/indicator.jl
```diff
@@ -0,0 +1,57 @@
...
+        return expr.args[2], MOI.IndicatorSet{MOI.ACTIVATE_ON_ZERO}
+    else
+        return expr, MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}
+    end
+end
+function parse_one_operator_constraint(
+    _error::Function, vectorized::Bool, ::Union{Val{:(=>)}, Val{:⇒}}, lhs, rhs)
+
+    variable, S = _indicator_variable_set(_error, lhs)
+    if !(rhs isa Expr) || rhs.head != :braces || length(rhs.args) != 1
```
mlubin: `!(rhs isa Expr) || rhs.head != :braces` can be written as `!isexpr(rhs, :braces)`. (`Meta.isexpr`)

src/indicator.jl
```diff
@@ -0,0 +1,57 @@
...
+    if vectorized
+        buildcall = :(build_indicator_constraint.($_error, $(esc(variable)), $rhs_buildcall, $S))
+    else
+        buildcall = :(build_indicator_constraint($_error, $(esc(variable)), $rhs_buildcall, $S))
+    end
+    return rhs_parsecode, buildcall
+end
+
+function constraint_string(
+    print_mode, constraint::VectorConstraint{F, <:MOI.IndicatorSet{A}}) where {F, A}
```
mlubin: Leave a TODO for pretty ijulia printing?

test/constraint.jl
```diff
@@ -239,6 +239,41 @@ function constraints_test(ModelType::Type{<:JuMP.AbstractModel},
...
+            @constraint(model, a ⇒  {x + 2y ≤ 1})
+        ]
+            c = JuMP.constraint_object(cref)
+            @test c.func == [a, x + 2y]
+            @test c.set == MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}(MOI.LessThan(1.0))
+        end
+        for cref in [
+            @constraint(model, !b => {2x + y <= 1});
+            @constraint(model, ¬b ⇒  {2x + y ≤ 1});
+            @constraint(model, ![b, b] .=> {[2x + y, 2x + y] .≤ 1});
```
mlubin: Why doesn't this `@constraint` return a vector of `ConstraintRef`s? I'm confused how it's passing tests.

test/print.jl
```diff
@@ -344,6 +344,17 @@ function printing_test(ModelType::Type{<:JuMP.AbstractModel})
         io_test(REPLMode, quad_constr, "2 x$sq $le 1.0")
         # TODO: Test in IJulia mode.
     end
+    @testset "Scalar QuadExpr constraints" begin
```
mlubin: s/QuadExpr/indicator

test/constraint.jl
blegat: I am using `;` so that it concatenates

src/indicator.jl
```diff
@@ -0,0 +1,57 @@
+#  Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file extends JuMP to indicator constraints. It is a good example of how
+# JuMP can be extended.
+
+function build_indicator_constraint(
```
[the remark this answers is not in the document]
blegat: The `_error` function is usually named with an underscore throught the JuMP code.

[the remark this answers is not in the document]
mlubin: I was suggesting `_build_indicator_constraint`.

PR #2100, opened 2019-11-12
Add support for multiple solutions

### LINKED ISSUE #2099: Multiobjective support in JuMP
This issue is for a discussion on introducing multiobjective support to JuMP (and
MOI).

## Passing multiple objectives to the solver

There are two possibilities:

1. Extending `@objective` to support vector-valued functions, so the user would
   write:
   ```julia
   @objective(model, Min, C * x)
   ```
2. Adding an `index` keyword, so the user would write:
   ```julia
   @objective(model, Min, C[1, :] * x, index = 1)
   @objective(model, Min, C[2, :] * x, index = 2)
   ```

Option (1) is probably the easiest, since it fits quite well into the MOI 
framework, and JuMP already has support for parsing vector-valued functions.

M.O. solvers would just declare:
```julia
function MOI.supports(
    ::Optimizer, ::MOI.ObjectiveFunction{MOI.VectorAffineFunction{Float64}}
)
    return true
end
```

The biggest downside with (1) is that it would only support linear and quadratic
objectives. We wouldn't be able to directly support nonlinear objectives. (You
could, of course, introduce a dummy variable with a nonlinear equality 
constraint.)

## Querying results from the solver

I think the easiest way for M.O. solvers to return the Pareto frontier is for 
them to sligthly abuse the notion of `ResultCount`.

We should implement 
```julia
function result_count(model::Model)
    return MOI.get(model, MOI.ResultCount())
end
```
And then add keyword arguments to `value`, `objective_value`, so we have:
```julia
function value(x::VariableRef; result=1)
    return MOI.get(owner_model(x), MOI.VariablePrimal(result), index(x))
end
```

The only issue with `objective_value` (and `objective_bound` and 
`dual_objective_bound`) is that they expect a `Float64` return type. This would
be relaxed to depend on the type of the objective function.

## Example

If `MultiJuMP` implemented a `MOI.Optimizer` solver, we could write:
```Julia
using JuMP
using MultiJuMP

model = Model(() -> MultiJuMP.Optimizer(GLPK.Optimizer))
@variable(model, x[1:2] >= 0)
@objective(model, Min, [2x[1] + x[2]; x[1] + 2x[2]])
@constraint(model, sum(x) >= 1)
optimize!(model)

pareto_frontier = []
for i = 1:result_count(model)
    @assert primal_status(model; result = i) == MOI.OPTIMAL
    push!(pareto_frontier, (
        x_val = value.(x; result = i),
        obj = objective_value(model; result = i)
    ))
end
```

cc @anriseth, @matbesancon

odow:
<s>This is a WIP PR to add multiple objective support to JuMP.

Currently blocked by changed needed in MOI.</s>

This PR adds support for querying multiple results form solvers via `result_index`.


src/constraints.jl
```diff
@@ -589,26 +589,35 @@ See also [`dual`](@ref) and [`shadow_price`](@ref).
...
-    dual(con_ref::ConstraintRef)
+    dual(con_ref::ConstraintRef; result::Int = 1)
 
 Get the dual value of this constraint in the result returned by a solver.
 Use `has_dual` to check if a result exists before asking for values.
 See also [`shadow_price`](@ref).
 """
-function dual(con_ref::ConstraintRef{Model, <:_MOICON})
-    return reshape_vector(_constraint_dual(con_ref), dual_shape(con_ref.shape))
+function dual(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)
```
blegat: I noticed performance issue with keyword argument in https://github.com/JuliaOpt/MathOptInterface.jl/pull/587
We need to check that this is not affecting performance here

odow: The kwarg hit on dual, if there is one, is tiny compared to everything else. This isn't going to be a bottleneck in anyone's code.

src/constraints.jl
```diff
@@ -545,14 +545,19 @@ function add_to_function_constant(constraint::ConstraintRef{Model}, value)
...
 
 """
-    value(con_ref::ConstraintRef)
+    value(con_ref::ConstraintRef; result::Int = 1)
 
-Get the primal value of this constraint in the result returned by a solver. That
-is, if `con_ref` is the reference of a constraint `func`-in-`set`, it returns the
-value of `func` evaluated at the value of the variables (given by
+Return the primal value of constraint `con_ref` associated with result index
+`result` of the most-recent returned by the solver.
```
blegat: most-recent what?

src/constraints.jl
```diff
@@ -562,60 +567,82 @@ into account in the primal value of the constraint. For instance, the constraint
...
-has_duals(model::Model) = dual_status(model) != MOI.NO_SOLUTION
+function has_duals(model::Model; result::Int = 1)
+    return dual_status(model; result = result) != MOI.NO_SOLUTION
+end
 
 """
-    dual(con_ref::ConstraintRef)
+    dual(con_ref::ConstraintRef; result::Int = 1)
+
+Return the dual value of constriant `con_ref` associated with result index
```
blegat: constriant -> constraint

src/constraints.jl
```diff
@@ -562,60 +567,82 @@ into account in the primal value of the constraint. For instance, the constraint
...
+    return MOI.get(con_ref.model, MOI.ConstraintDual(result), con_ref)
 end
 
 
 """
-    shadow_price(con_ref::ConstraintRef)
+    shadow_price(con_ref::ConstraintRef; result::Int = 1)
+
+Return the change in the objective from an infinitesimal relaxation of the
+constraint.
```
blegat: `result` is not used in the docstring

docs/src/solutions.md
```diff
@@ -219,6 +219,32 @@ lp_objective_perturbation_range
 lp_rhs_perturbation_range
 ```
 
+## Multiple solutions
+
+Some solvers support returning multiple solutions. You can check how many
+solutions are available to query using [`result_count`](@ref).
+
+Then, functions for querying the solutions, e.g., [`primal_status`](@ref) and
```
mlubin: Nit: "Then" isn't needed. Just: "Functions for querying solutions ..."

src/constraints.jl
mlubin: It's also not clear if we need to expose `result` from `shadow_price`. I would be surprised if anyone ever asks for it. It may be better to keep this interface as simple as possible because it's for teaching use only.

src/constraints.jl
```diff
@@ -667,37 +694,42 @@ function shadow_price_greater_than_(dual_value, sense::MOI.OptimizationSense)
...
 end
 
-function shadow_price(con_ref::ConstraintRef{Model, _MOICON{F, S}}
-                      ) where {S <: MOI.LessThan, F}
+function shadow_price(
+    con_ref::ConstraintRef{Model, _MOICON{F, S}}
+) where {S <: MOI.LessThan, F}
     model = con_ref.model
-    if !has_duals(model)
+    if !has_duals(model; result = 1)
```
mlubin: Not necessary to pass the default argument.

PR #2101, opened 2019-11-12
Callbacks Episode II: revenge of the solver-independent implementation

### LINKED ISSUE #1553: [docs] Callbacks
I am getting often questions about why generic callbacks were dropped from JuMP. I think we need to have in the doc an example that shows the solver dependent aspect of callbacks and the inconsistency issues it can raise.

Also I wonder if we can restrict the scope of that dependancy. 

For example instead of the old:

```julia
function user_routine(cb::MathProgBase.MathProgCallbackData)
    x_val = getvalue(x)
    y_val = getvalue(y)
    ...
    @lazyconstraint(cb, y - x <= 1)
    ...
end

# Tell JuMP/Cplex to use our callback function
addlazycallback(m, user_routing)
```

we can imagine something like:

```julia
struct Cplex.LazyConstraintCallback <: MOI.ConstraintGenerationCallback
...
end

struct Cplex.LazyConstraintCallbackData <: MOI.ConstraintGenerationCallbackData 
...
end

function user_routine(cb)
    x_val = getvalue(x)
    y_val = getvalue(y)
    ...
    @generated_constraint(cb, y - x <= 1)
    ...
end

# Tell JuMP/Cplex to use our callback function
add_constraint_generation_callback(m, CPLEX.LazyConstraintCallback, user_routine)
```

While `add_constraint_generation_callback` will work for any constraint generation callback ( lazy constraint, a user cut, ... )

Even if we don't develop it right now, pointing on how the new callbacks might look like in a year or so, may encourage people who are still hesitant to switch to JuMP.

### LINKED ISSUE #1880: @lazyconstraint not defined in Julia 1.1
Hello

I try to use @lazyconstraint to version Julia 1.1 but obtain this error:
LoadError: UndefVarError: @lazyconstraint not defined

odow:
It's _finally_ time to re-introduce callbacks into JuMP.

I've tried to stay as light as possible. I think it's reasonable to leave a fairly heavy MOI influence in the syntax since this makes it easier to transition to solver-dependent callbacks. It's kind of nice that the JuMP side requires all of four lines of code.

The biggest question is how to test this. I"m open to ideas. For now, I've just added an example. The issue is that JuMP tests don't have a solver, and the MOI.Utilities mock solver doesn't understand callbacks.

TODOs
 - [x] how to test?
 - [x] documentation

Closes #1553
Closes #1849 
Closes #1880

### DIFF
--- a/docs/src/callbacks.md
+++ b/docs/src/callbacks.md
@@ -23,6 +23,10 @@ callbacks:
  2. user-cuts
  3. heuristic solutions
 
+!!! warning
+    Using callbacks requires a solver in [Direct mode](@ref). A direct-mode
+    model is created using [`JuMP.direct_model`](@ref).
+
 ## Available solvers
 
 Callback support is limited to a few solvers. This includes
@@ -31,10 +35,12 @@ Callback support is limited to a few solvers. This includes
 [Gurobi](https://github.com/JuliaOpt/Gurobi.jl).
 
 !!! warning
-    While JuMP provides a solver-independent way of accessing callbacks, the
-    particular implementations are dependent on the solver. You should not use
-    the callbacks without understanding how your particular solver handles
-    each callback.
+    While JuMP provides a solver-independent way of accessing callbacks, you
+    should not assume that you will see identical behavior when running the same
+    code on different solvers. For example, some solvers may ignore user-cuts
+    for various reasons, while other solvers may add every user-cut. Read the
+    underlying solver's callback documentation to understand details specific to
+    each solver.
 
 ## Information that can be queried during callbacks
 
@@ -42,6 +48,11 @@ In a callback, the only thing you may query is the primal value of the
 variables using [`callback_value`](@ref).
 
 If you need any other information, use a solver-dependent callback instead.
+
+!!! info
+    Solver-dependent callbacks are mostly un-documented. Using them will require
+    you to read and understand the source-code of solver's Julia wrapper (i.e.,
+    the `Solver.jl` package).
 
 ## Lazy constraints
 
--- a/docs/src/callbacks.md
+++ b/docs/src/callbacks.md
@@ -1,0 +1,162 @@
+```@meta
+CurrentModule = JuMP
+DocTestSetup = quote
+    using JuMP
+end
+DocTestFilters = [r"≤|<=", r"≥|>=", r" == | = ", r" ∈ | in ", r"MathOptInterface|MOI"]
+```
+
+# Callbacks
+
+Many mixed-integer (linear, conic, and nonlinear) programming solvers offer
+the ability to modify the solve process. Examples include changing branching
+decisions in branch-and-bound, adding custom cutting planes, providing custom
+heuristics to find feasible solutions, or implementing on-demand separators to
+add new constraints only when they are violated by the current solution (also
+known as lazy constraints).
+
+While historically this functionality has been limited to solver-specific
+interfaces, JuMP provides solver-independent support for three types of
+callbacks:
+
+ 1. lazy constraints
+ 2. user-cuts
+ 3. heuristic solutions
+
+## Available solvers
+
+Callback support is limited to a few solvers. This includes
+[CPLEX](https://github.com/JuliaOpt/CPLEX.jl),
+[GLPK](https://github.com/JuliaOpt/GLPK.jl), and
+[Gurobi](https://github.com/JuliaOpt/Gurobi.jl).
+
+!!! warning
+    While JuMP provides a solver-independent way of accessing callbacks, the
+    particular implementations are dependent on the solver. You should not use
+    the callbacks without understanding how your particular solver handles
+    each callback.
+
+## Information that can be queried during callbacks
+
+In a callback, the only thing you may query is the primal value of the
+variables using [`callback_value`](@ref).
+
+If you need any other information, use a solver-dependent callback instead.
+
+## Lazy constraints
+
+Lazy constraints are useful when the full set of constraints is too large to
+explicitly include in the initial formulation. When a MIP solver reaches a new
+solution, for example with a heuristic or by solving a problem at a node in
+the branch-and-bound tree, it will give the user the chance to provide
+constraint(s) that would make the current solution infeasible. For some more
+information about lazy constraints, see this [blog post by Paul Rubin](http://orinanobworld.blogspot.com/2012/08/user-cuts-versus-lazy-constraints.html).
+
+A lazy constraint callback can be set using the following syntax:
+
+```julia
+model = direct_model(GLPK.Optimizer)
+@variable(model, x <= 10, Int)
+@objective(model, Max, x)
+function my_callback_function(cb_data)
+    x_val = callback_value(cb_data, x)
+    if x_val > 2 + 1e-6
+        con = @build_constraint(x <= 2)
+        MOI.submit(model, MOI.LazyConstraint(cb_data), con)
+    end
+end
+MOI.set(model, MOI.LazyConstraintCallback(), my_callback_function)
+```
+
+!!! info
+    The lazy constraint callback _may_ be called at fractional or integer
+    nodes in the branch-and-bound tree. There is no guarantee that the
+    callback is called at _every_ feasible primal solution.
+
+## User cuts
+
+User cuts, or simply cuts, provide a way for the user to tighten the LP
+relaxation using problem-specific knowledge that the solver cannot or is
+unable to infer from the model. Just like with lazy constraints, when a MIP
+solver reaches a new node in the branch-and-bound tree, it will give the user
+the chance to provide cuts to make the current relaxed (fractional) solution
+infeasible in the hopes of obtaining an integer solution. For more details
+about the difference between user cuts and lazy constraints see the
+aforementioned [blog post](http://orinanobworld.blogspot.com/2012/08/user-cuts-versus-lazy-constraints.html).
+
+A user-cut callback can be set using the following syntax:
+
+```julia
+model = direct_model(GLPK.Optimizer)
+@variable(model, x <= 10.5, Int)
+@objective(model, Max, x)
+function my_callback_function(cb_data)
+    x_val = callback_value(cb_data, x)
+    con = @build_constraint(x <= floor(x_val))
+    MOI.submit(model, MOI.UserCut(cb_data), con)
+end
+MOI.set(model, MOI.UserCutCallback(), my_callback_function)
+```
+
+!!! warning
+    Your user cuts should not change the set of integer feasible solutions.
+    Equivalently, your cuts can only remove fractional solutions. If you add a
+    cut that removes an integer solution, the solver may return an incorrect
+    solution.
+
+!!! info
+    The user-cut callback _may_ be called at fractional nodes in the
+    branch-and-bound tree. There is no guarantee that the callback is called
+    at _every_ fractional primal solution.
+
+## Heuristic solutions
+
+Integer programming solvers frequently include heuristics that run at the
+nodes of the branch-and-bound tree. They aim to find integer solutions quicker
+than plain branch-and-bound would to tighten the bound, allowing us to fathom
+nodes quicker and to tighten the integrality gap.
+
+Some heuristics take integer solutions and explore their "local neighborhood"
+(e.g., flipping binary variables, fix some variables and solve a smaller MILP)
+and others take fractional solutions and attempt to round them in an
+intelligent way.
+
+You may want to add a heuristic of your own if you have some special insight
+into the problem structure that the solver is not aware of, e.g. you can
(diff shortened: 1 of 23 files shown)


mtanneau:
> The biggest question is how to test this.

I'm not too familiar with the JuMP side of things. Exactly what needs to be tested here?
That a callback defined in JuMP is correctly passed to the MOI backend? That the callback performs as expected during an `optimize!` call? Something else?

For instance, verifying that user cuts and heuristics perform doesn't seem trivial to me, since they only affect performance, not the result (assuming for simplicity there is a unique optimal solution).

matbesancon:
Seeing the set syntax:
```julia
     lazy_called = false
    MOI.set(model, MOI.LazyConstraintCallback(), (cb_data) -> begin
        con = @build_constraint(y - x <= 1)
        MOI.submit(model, MOI.LazyConstraint(cb_data), con)
    end) 
```

This would lend itself to having the callback as first argument to leverage 
```julia
set_callback(model) do cb_data
    # ...
end
```

odow:
> This would lend itself to having the callback as first argument to leverage

I thought of this idea:
```Julia
lazy_constraint_callback(model) do cb_data
  ...
end

heuristic_solution_callback(model) do cb_data
  ...
end

user_cut_callback(model) do cb_data
  ...
end
```

The only real argument against this is that it differs from the rest of the JuMP code-base, and it departs from solver-dependent callbacks.

For example, the solver-dependent GLPK callback is:
```Julia
function example_lazy_constraint()
    model = direct_model(GLPK.Optimizer())
    @variable(model, 0 <= x <= 2.5, Int)
    @variable(model, 0 <= y <= 2.5, Int)
    @objective(model, Max, y)
    lazy_called = false
    MOI.set(model, GLPK.CallbackFunction(), (cb_data) -> begin
        lazy_called = true
        reason = GLPK.ios_reason(cb_data.tree)
        if reason != GLPK.IROWGEN
            return
        end
        x_val = callback_value(cb_data, x)
        y_val = callback_value(cb_data, y)
        if y_val - x_val > 1 + 1e-6
            con = @build_constraint(y - x <= 1)
            MOI.submit(model, MOI.LazyConstraint(cb_data), con)
        elseif y_val + x_val > 3 + 1e-6
            con = @build_constraint(y - x <= 1)
            MOI.submit(model, MOI.LazyConstraint(cb_data), con)
        end
    end)
    optimize!(model)
    @test termination_status(model) == MOI.OPTIMAL
    @test primal_status(model) == MOI.FEASIBLE_POINT
    @test lazy_called
    @test value(x) == 1
    @test value(y) == 2
end
```
So it's pretty crazy how little changes.

matbesancon:
I see the point yes, I don't have a too strong opinion on the syntax but `v-> begin # .. end` is a bit heavy

odow:
I can change the examples to define actual functions. That might be clearer.

src/callbacks.jl
```diff
@@ -0,0 +1,42 @@
...
+    callback_value(cb_data, x::VariableRef)
+
+Return the primal solution of a variable inside a callback.
+
+`cb_data` is the argument to the callback function, and the type is dependent on
+the solver.
+"""
+function callback_value(cb_data, x::VariableRef)
+    return MOI.get(
+        backend(owner_model(x)),
```
blegat: Why not `MOI.get(owner_model(x), MOI.CallbackVariablePrimal(cb_data), x)` ?
Accessing `backend` is type unstable so it's better to handle it carefully in the generic `MOI.get` and use it in a minimal number of other places

blegat:
> The biggest question is how to test this.

We could do something like
```julia
struct DummyCallbackData end
mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
MOIU.set_mock_optimize!(mock, mock -> begin
    MOI.get(mock, MOI.LazyConstraintCallback())(DummyCallbackData())
end)
model = direct_model(mock)
...
@test mock.submitted[MOI.LazyConstraint()] == ...
```

src/callbacks.jl
```diff
@@ -17,7 +17,18 @@ Return the primal solution of a variable inside a callback.
...
-    return MOI.get(owner_model(x), MOI.CallbackVariablePrimal(cb_data), x)
+    # TODO(odow):
+    # MOI defines `is_set_by_optimize(::CallbackVariablePrimal) = true`.
+    # This causes problems for JuMP because it checks the termination_status to
+    # see if optimize! has been called. Solutions are:
+    # 1) defining is_set_by_optimize = false
+    # 2) adding a flag to JuMP to store whether it is in a callback
+    # 3) adding IN_OPTIMIZE to termination_status for callbacks
+    # Once this is resolved, we can replace the current function with:
+    #     MOI.get(owner_model(x), MOI.CallbackVariablePrimal(cb_data), x)
```
odow: @blegat: thoughts?

test/callbacks.jl
```diff
@@ -0,0 +1,63 @@
...
+struct DummyCallbackData end
+
+@testset "LazyConstraint" begin
+    mock = MOI.Utilities.MockOptimizer(
+        MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}())
+    )
+    model = direct_model(mock)
+    @variable(model, 0 <= x <= 2.5, Int)
+    con = @build_constraint(x <= 2)
+    c = MOI.submit(model, MOI.LazyConstraint(DummyCallbackData()), con)
```
blegat: Isn't submit not supposed to return anything ? If we make it return nothing in MOI then it might break JuMP tests

rschwarz:
> For instance, verifying that user cuts and heuristics perform doesn't seem trivial to me, since they only affect performance, not the result (assuming for simplicity there is a unique optimal solution).

With cuts I agree, but heuristics can be tested (sort of) by submitting a sub-optimal solution, while also limiting the solver to stop after the first solution was found. In addition, it might be necessary to disable all of the solver's other heuristics, so that the heuristic callback is really the first in line.

odow:
@rschwarz: I was referring to testing the JuMP functions. It's up to the solvers to test that they implement callbacks correctly. I've now added tests to `JuMP/test/callbacks.jl` to check we're mapping the `submittable`s correctly.

src/callbacks.jl
blegat: Good point,
1) would not work as otherwise the attribute will be queried to the cache instead of the optimizer in `CachingOptimizer`.
2) I would prefer a more general solution so that other attributes can be defined outside of MOI.
3) I like his idea. It could also be called `OPTIMIZE_IN_PROGRESS` to be consistent with the `OptimizeInProgress` error.

docs/src/callbacks.md
```diff
@@ -0,0 +1,168 @@
...
+ 1. lazy constraints
+ 2. user-cuts
+ 3. heuristic solutions
+
+## Available solvers
+
+Callback support is limited to a few solvers. This includes
+[CPLEX](https://github.com/JuliaOpt/CPLEX.jl),
+[GLPK](https://github.com/JuliaOpt/GLPK.jl), and
+[Gurobi](https://github.com/JuliaOpt/Gurobi.jl).
```
blegat: As with the solver table, this kind of information can become quickly outdated and would rather belong to the website IMO. At the time the user reads it, more solvers could implement the callbacks.

src/callbacks.jl
odow: I also vote 3).

docs/src/callbacks.md
```diff
@@ -0,0 +1,168 @@
...
+    particular implementations are dependent on the solver. You should not use
+    the callbacks without understanding how your particular solver handles
+    each callback.
+
+## Information that can be queried during callbacks
+
+In a callback, the only thing you may query is the primal value of the
+variables using [`callback_value`](@ref).
+
+If you need any other information, use a solver-dependent callback instead.
```
mlubin: What about some hints on how to use solver-dependent callbacks? (Or if they're not ready/documented, say so.)

docs/src/callbacks.md
```diff
@@ -0,0 +1,168 @@
...
+add new constraints only when they are violated by the current solution (also
+known as lazy constraints).
+
+While historically this functionality has been limited to solver-specific
+interfaces, JuMP provides solver-independent support for three types of
+callbacks:
+
+ 1. lazy constraints
+ 2. user-cuts
+ 3. heuristic solutions
```
mlubin: Do callbacks require using `direct_mode`?

blegat: It should work with cache and bridges too

mlubin: This is untested, at least.

blegat: There are unit tests in MOI but indeed no integration test at the JuMP level.

odow: Hmm. I can't get this to work. I think `OptimizerAttributes` aren't being copied? https://github.com/JuliaOpt/MathOptInterface.jl/blob/1755bf88b05499aced42b4ca1c15267045b038ed/src/Utilities/copy.jl#L331-L340

blegat: Good point, that will be an issue if the CachingOptimizer state is `EMPTY_OPTIMIZER`, we should fix this in MOI v0.9.8. However, if it is `ATTACHED_OPTIMIZER` then the attributes will be passed to the optimizer during `MOI.set`.

odow: Question: why don't we call `attach_optimizer` on the creation of `Model`? Because if we try to add a variable and it isn't supported, it will detach and we're back to the current approach. Whereas the current approach causes a big `copy_to` operation on the first `optimize!`, even if incremental building is supported.

mlubin: The big `copy_to` has the potential to be faster than incremental building because you can pass the problem all at once to efficient methods like `GRBloadmodel`.

blegat: Yes, the reasoning is that the incremental model building can only be slower than copying it all at once. There is a slight inconsistency with
https://github.com/JuliaOpt/MathOptInterface.jl/blob/1755bf88b05499aced42b4ca1c15267045b038ed/src/Utilities/cachingoptimizer.jl#L173-L175
where we attach the solver when we empty the caching optimizer.

ferrolho:
I have been searching through the docs and issues about generic callbacks. I don't want to create a new issue for this, and this seems to be the latest and most relevant existing issue to ask the question at, so here it goes.

How can users define a custom function to be called at every new iterate? For example, how can I have a callback that pushes to an array the value of a certain decision variable for every new solver iteration?

As far as I know, there is no way to do this in a generic way for any solver supported. But how can I do this for a specific solver then? Thank you in advance!

mtanneau:
At the lowest level, callbacks are functions that are called by a solver at certain points.
Thus, the first question you need to ask is: does solver `XXX` support callback functionality `YYY`? If the answer is no, you're done (or you need to implement it yourself)

If the functionality `YYY` is supported by solver `XXX`, and you only want to use it with solver `XXX`, then you probably don't need a solver-agnostic callback for doing so.
If you really want to, then docs for [`MOI.AbstractCallback`](https://jump.dev/MathOptInterface.jl/stable/apireference/#MathOptInterface.AbstractCallback) is where you'll want to start. Just define your own callback, and update your solver's MOI wrapper accordingly.

ferrolho:
Thank you for your reply, @mtanneau. I guess my question is related with the latter case, i.e., where solver `XXX` supports functionality `YYY`. How can I add that solver-specific-callback to my existing and generic JuMP model?

mtanneau:
> is the answer to go with MOI.AbstractCallback?

Yes, if you want a nice-looking callback that you can add to your solver-agnostic JuMP model, then you need to go through an MOI callback, and then port that MOI callback to a JuMP callback. There's a bit of interfacing to do here.

Another way (but more risky), is to use JuMP to model the problem (preferably in direct mode), then access the low-level solver object, and attach a callback to it directly. You don't need an MOI callback to do this, because you're interfacing with the solver, but... now you can't use any of the nice JuMP objects that you have. For instance, you can't query the value of a JuMP variable `x` (I think doing this would throw an `OptimizeInProgress` error), you'd have to know the index of that variable in the solver object.
There's less interfacing but more risk for fatal bugs I would say. Not recommended unless you really know what you are doing.

odow:
> Just define your own callback, and update your solver's MOI wrapper accordingly.

Please don't do this. Use the existing solver-dependent callbacks instead.

Here is an example for GLPK:
https://github.com/jump-dev/JuMP.jl/blob/b34e040ac2220a5a4e6e39ebb1464c054f34596c/examples/callbacks.jl#L121-L155

Note that only GLPK, Gurobi, and CPLEX have solver-dependent callbacks at present.

Please post questions like this on the forum https://discourse.julialang.org/c/domain/opt/13 to avoid cluttering these old pull requests.

ferrolho:
@odow, thank you for including the link to `example_solver_dependent_callback()`. And I will use the discourse for these sort of questions instead next time. :+1:

PR #2103, opened 2019-11-13
Update FUNDING.yml

Dr-G:
add NumFOCUS github sponsors button (recurring donations only)

This feature is launching today at GitHub Universe!

cc @mlubin


odow:
Thanks @Dr-G

PR #2108, opened 2019-11-23
Print constraint name in model string

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1767

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -402,8 +402,7 @@ function constraints_string(print_mode, model::Model)
     strings = String[]
     for (F, S) in list_of_constraint_types(model)
         for cref in all_constraints(model, F, S)
-            con = constraint_object(cref)
-            push!(strings, constraint_string(print_mode, con))
+            push!(strings, constraint_string(print_mode, cref, in_math_mode = true))
         end
     end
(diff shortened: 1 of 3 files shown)


src/print.jl
```diff
@@ -557,19 +556,21 @@ function constraint_string(print_mode, constraint_object::AbstractConstraint)
...
-                           constraint_object::AbstractConstraint)
+                           constraint_object::AbstractConstraint;
+                           in_math_mode = false)
     constraint_without_name = constraint_string(print_mode, constraint_object)
-    if print_mode == IJuliaMode
+    if print_mode == IJuliaMode && !in_math_mode
         constraint_without_name = _wrap_in_inline_math_mode(constraint_without_name)
     end
-    if isempty(constraint_name)
+    # Names don't print well in LaTeX math mode
```
mlubin: It looks like this changes existing printing behavior. Why didn't any tests break?

blegat: When printing constraint alone, the constraint name is not in math mode so we add the name (no change here)
When printing the full model in math mode, the whole model is in math mode so we don't print the name (no change here)
When printing the full model in REPL mode, the name has been added and the tests updated

PR #2111, opened 2019-12-01
[RFC] Interpret NaN as no bound

blegat:
What do you think of the following change of behavior ?

Before:
```julia
julia> @variable(model, x >= 1, upper_bound = NaN)
x

julia> print(model)
Feasibility
Subject to
 x ≥ 1.0
 x ≤ NaN
```
After:
```julia
julia> @variable(model, x >= 1, upper_bound = NaN)
x

julia> print(model)
Feasibility
Subject to
 x ≥ 1.0
```
This fixes the use cases exposed by @ccoffrin in https://github.com/lanl-ansi/PowerModels.jl/issues/630

Have a bound of value `NaN` will have inconsistent behaviors across solvers so it makes sense to interpret it as no bounds but the issue is that this does not work:
```julia
julia> a = NaN
NaN

julia> @variable(model, x >= 1, upper_bound = a)
x

julia> print(model)
Feasibility
Subject to
 x ≥ 1.0
 x ≤ NaN
```

Once we have a consensus, I will do:

- [ ] Add tests
- [ ] Add doc

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -51,9 +51,11 @@ end
 function _keywordify(kw::Expr)
     (kw.args[1], _esc_non_constant(kw.args[2]))
 end
+_is_nan_expr(expr::Number) = isnan(expr)
+_is_nan_expr(expr::Symbol) = expr == :NaN
+_is_nan_expr(expr) = isexpr(expr, :escape) && length(expr.args) == 1 && _is_nan_expr(expr.args[1])
 function _VariableInfoExpr(; lower_bound=NaN, upper_bound=NaN, start=NaN, binary=false, integer=false)
-    # isnan(::Expr) is not defined so we need to do !== NaN
-    _VariableInfoExpr(lower_bound !== NaN, lower_bound, upper_bound !== NaN, upper_bound, false, NaN, start !== NaN, start, binary, integer)
+    _VariableInfoExpr(!_is_nan_expr(lower_bound), lower_bound, !_is_nan_expr(upper_bound), upper_bound, false, NaN, !_is_nan_expr(start), start, binary, integer)
 end
 
 struct VariableInfo{S, T, U, V}


mlubin:
Following the discussion in https://github.com/lanl-ansi/PowerModels.jl/issues/630, it looks like this is a workaround for solvers not supporting intervals like `(-inf, x]` or `[x, inf)`. I think they should. This is clearly well defined and useful in many contexts.

blegat:
Having `SplitIntervalBridge` ignore the bounds if it is infinite would help for solvers relying on that bridge.

mlubin:
Regarding this PR, if we want a special value at the JuMP level for "no bound", it should be `nothing`, not `NaN`. `nothing` is consistent with how we specify missing values in MOI.

ccoffrin:
A note on Miles' remark, the most common use case in PowerModels is `(-inf, inf)` which I have been using as a proxy for explicitly unbounded, which I presumed was indistinguishable from omitting the `upper_bound` or `lower_bound` keyword args.  No issue with this approach was observed until lanl-ansi/PowerModels.jl#630.

Functionally what I need is a value that is indistinguishable from omitting the `upper_bound` or `lower_bound` keyword args.  `nothing` is fine by me.

mlubin:
@ccoffrin, instead of a value indistinguishable from omitting the `lower_bound` argument to `@variable`, what about always omitting it and conditionally calling `set_lower_bound`? The code complexity should be about the same, if you also have to conditionally replace `-Inf` with `nothing`.

ccoffrin:
@mlubin, that is also a viable approach and I probably use that in the near term.  I expect that there are some cases when the `nothing`-like solution would lead to cleaner looking comprehensions but this case is not one of them.

odow:
I think the consensus is that we won't be implementing this change. Use `set_lower_bound` instead.

PR #1982, opened 2019-06-07
[WIP] Implementation of write_to_file that use MathOptFormat.

dourouc05:
Follows https://github.com/odow/MathOptFormat.jl/issues/70. Not mergeable right now due to MOF not being registered (but it may happen soon, as I understood).

PR #2114, opened 2019-12-03
Add write_to_file and read_from_file

### LINKED ISSUE #1075: Problem writers
For a different project I needed light-weight LP and MPS file writers. These have inputs similar to the MPB linear quadratic interface.

https://github.com/odow/LPWriter.jl
https://github.com/odow/MPSWriter.jl

At some point these should be used by JuMP, or put into some sort of "MathOptFileFormats" package that allows reading and writing of common MathOpt file formats like LP, MPS, NL, and OSIL.

This is probably actionable after the re-write, so leaving this as a reminder.

I did pick up a few issues with the current implementations:
- LP writer doesn't support binary variables
- LP writer doesn't check for invalid names properly
- MPS format doesn't support user specified names

odow:
Closes #1075
Closes #1982

cc @dourouc05


mlubin:
My only concern is that we're not following the standard Julia API by allowing reading and writing from an IO object. It's inevitable that we'll get requests for this. Does MOF support writing to streams?

dourouc05:
Looks good to me (even though the failures on Windows are quite strange). Maybe refactor a bit the test suite to avoid having three times the exact same model? (Sorry for the delay, I still have the tab for my PR opened, but had too little time to work on it, unfortunately :(.)

I think this API is already quite good as it is. For streams, I suspect `write` and `read` overloads should be provided, but these cannot support the convenient inference from file name. This calls for a slightly different interface, like `read/write(io::IO, m::Model, compression, format). 

Another though: wouldn't the order format-compression for the arguments be easier to grasp? The file names almost invariably have first the file format, then the compression format (like .lp.gz).

odow:
> Does MOF support writing to streams?

I've added the read/write to streams.

>  isn't there any other solution, like including parts of MOF docs from JuMP?

Longer term, we should probably pull MOF into MOI. Our longer term goal is to unify the JuMP/MOI documentation in a nice way.

dourouc05:
> Longer term, we should probably pull MOF into MOI. Our longer term goal is to unify the JuMP/MOI documentation in a nice way.

Nice to hear! 

> I've added the read/write to streams.

Why aren't you using the "standard" way of defining IO operations on streams (with read and write methods)? This would also imply to have the IO stream as a first argument, but would be more convenient for people picking up JuMP after Julia.

odow:
> Why aren't you using the "standard" way of defining IO operations on streams (with read and write methods)?

Mainly to keep consistency with `write_to_file` and `read_from_file`. It's a small change to swap `filename` to `io`.  I'll defer to @mlubin if he prefers `read`/`write` though.

src/file_formats.jl
```diff
@@ -0,0 +1,134 @@
+#  Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+const MOF = MathOptFormat
```
mlubin: I'm still uncomfortable with exporting other modules (https://github.com/JuliaOpt/JuMP.jl/issues/1789). Do we need to export MOF or can we just re-export the two enums used here?

odow:
I'm making changes to MathOptFormat.jl which should simplify JuMP's implementation. See: https://github.com/odow/MathOptFormat.jl/pull/100.

src/file_formats.jl
```diff
@@ -0,0 +1,134 @@
...
+
+"""
+    write_to_file(
+        model::Model,
```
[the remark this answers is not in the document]
odow: `foo`. We're not going to do modify user input.

odow:
I've merged and released the changes to MathOptFormat.jl, so now things are now considerably simpler on the JuMP side.

src/copy.jl
```diff
@@ -144,3 +144,12 @@ end
 function Base.deepcopy(::Model)
     error("`JuMP.Model` does not support `deepcopy` as the reference to the underlying solver cannot be deep copied, use `copy` instead.")
 end
+
+function MOI.copy_to(dest::MOI.ModelLike, src::Model)
+    bridged_dest = MOI.Bridges.full_bridge_optimizer(dest, Float64)
```
blegat: What if `dest` is already bridged ? There cannot be multiple layers of bridges because variables bridges assume that the  underlying model does not return variable with negative indices.

docs/src/solvers.md
```diff
@@ -112,3 +112,15 @@ set_time_limit_sec
 unset_time_limit_sec
 time_limit_sec
 ```
+
+## File formats
+
+JuMP can write models to a variety of file-formats using [`write_to_file`](@ref).
+```@docs
+write_to_file
```
mlubin: Also include the methods for `write` and `read!`.

src/file_formats.jl
```diff
@@ -0,0 +1,98 @@
...
+    MOI.copy_to(model, src)
+    return model
+end
+
+"""
+    Base.read(io::IO, ::Type{Model}; format::MathOptFormat.FileFormat)
+
+Return a JuMP model read from `io` in the format `format`.
+"""
+function Base.read(io::IO, ::Type{Model}; format::MathOptFormat.FileFormat)
```
mlubin: What does it mean to have a keyword argument without a default value? I haven't seen this before.

odow: It has been valid base Julia syntax for a while. It throws a nice error if you don't provide a value.

PR #2116, opened 2019-12-14
fix knapsack example (with_optimizer)

igormcoelho:
First time experimenting Julia and JuMP... looks like knapsack example is broken. According to QuickStart, we need "with_optimizer": http://www.juliaopt.org/JuMP.jl/v0.20.0/quickstart/

This explains an ugly error I was getting:
```
julia> model = Model(GLPK.Optimizer)
ERROR: MethodError: no method matching Model(::Type{GLPK.Optimizer})
Closest candidates are:
  Model(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any) at /home/imcoelho/.julia/packages/JuMP/MsUSY/src/JuMP.jl:126
  Model(; caching_mode, solver) at /home/imcoelho/.julia/packages/JuMP/MsUSY/src/JuMP.jl:163
  Model(::MathOptInterface.AbstractOptimizer, ::Dict{MathOptInterface.ConstraintIndex,AbstractShape}, ::Set{Any}, ::Any, ::Any, ::Dict{Symbol,Any}, ::Int32, ::Dict{Symbol,Any}) at /home/imcoelho/.julia/packages/JuMP/MsUSY/src/JuMP.jl:126
  ...
Stacktrace:
 [1] top-level scope at none:0
```

### DIFF
--- a/examples/knapsack.jl
+++ b/examples/knapsack.jl
@@ -22,7 +22,7 @@ function example_knapsack(; verbose = true)
     profit = [5, 3, 2, 7, 4]
     weight = [2, 8, 4, 2, 5]
     capacity = 10
-    model = Model(GLPK.Optimizer)
+    model = Model(with_optimizer(GLPK.Optimizer))
     @variable(model, x[1:5], Bin)
     # Objective: maximize profit
     @objective(model, Max, profit' * x)


mlubin:
Thanks for making this PR. The example on the `master` branch is compatible with the rest of the code on the `master` branch, where we've deprecated `with_optimizer`. See https://github.com/JuliaOpt/JuMP.jl/tree/release-0.20/examples for examples compatible with the 0.20 release.

igormcoelho:
I suspected of something like that @mlubin ... can you explain me which is current valid version number on master?

mlubin:
The code on the master branch hasn't been released yet, so it doesn't correspond with any existing version number. It will become 0.21 when it's released.

igormcoelho:
So, I would suggest to update project README with this information, like working on `0.21-dev`, and perhaps add some links for past branches, for example this `release-0.20` because I couldn't easily find them (and I tried a little bit when I suspected it could be the issue...)
Anyway, congratulations for the amazing project

mlubin:
We have links to the examples and documentation for 0.20: 
image

Do you mean we should point the `release-0.20` text to https://github.com/JuliaOpt/JuMP.jl/tree/release-0.20? That would be an easy change.

igormcoelho:
ohh!! I see... I'm very used to Github, yet I got very confused at first... some things are worth improving I guess :)

In this case, my suggestions are:

> Do you mean we should point the release-0.20 text to https://github.com/JuliaOpt/JuMP.jltree/release-0.20? That would be an easy change.

On "release-2.0 branch", point to that branch as a link.

On "Documentation" for release-2.0, write something like "Documentation (0.20)".
On "Examples" for release-2.0, write something like "Examples (0.20)"

It would certainly help a lot... being more verbose, but for newcomers (like me) it's worth a lot.
Now I understood the project... but perhaps these can help others in the future ;)

PR #2119, opened 2019-12-16
Clearer warning about examples version mismatch

mlubin:
See discussion at PR #2116.

@igormcoelho, does this address your concern?

> On "Documentation" for release-2.0, write something like "Documentation (0.20)".
> On "Examples" for release-2.0, write something like "Examples (0.20)"

I'd prefer not to add more hard-coded version numbers to the README, it makes the release process more complicated and has the potential to accidentally fall out of sync.


igormcoelho:
Very good idea! Looks nice, thanks!

PR #2126, opened 2019-12-28
Add documentation on performance and work-around for sparse variable creation

hellemo:
Add note on possible performance implications and suggested work-around following the discussion here:
https://discourse.julialang.org/t/working-efficiently-with-sparse-variables-in-jump/32240

### DIFF
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -387,35 +387,33 @@ JuMP.Containers.SparseAxisArray{VariableRef,1,Tuple{Int64}} with 2 entries:
   [2]  =  x[2]
 ```
 
-Note that with many indices and large amount of sparsity, variable construction
-may take more time than expected, as JuMP iterates over all indices and 
-evaluates the conditional for each combination. The recommended work-around
-in such cases is to work directly with a list of tuples or create a 
-dictionary. Consider the following examples:
+Note that with many index dimensions and a large amount of sparsity,
+variable construction may be unnecessarily slow if the semi-colon syntax is
+naively applied. When using the semi-colon as a filter, JuMP iterates over
+*all* indices and evaluates the conditional for each combination. When this
+is undesired, the recommended work-around is to work directly with a list
+of tuples or create a dictionary. Consider the following examples:
+
+```@meta
+# TODO: Reformat the code below as a proper doctest.
+```
 
 ```jldoctest; setup=:(model=Model())
 N = 10
-S = [(1,1,1),(N,N,N)]
-# slow as it evaluates conditional N^3 times:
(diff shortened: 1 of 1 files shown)


docs/src/variables.md
```diff
@@ -387,6 +387,12 @@ JuMP.Containers.SparseAxisArray{VariableRef,1,Tuple{Int64}} with 2 entries:
   [2]  =  x[2]
 ```
 
+Note that with many indices and large amount of sparcity, variable construction may take more time than expected, as JuMP is effectively iterating over all indices and evaluating the conditional for each combination. The recommended work-around in such cases is to work directly with a list of tuples. For example:
```
blegat: sparcity -> sparsity

docs/src/variables.md
```diff
@@ -387,6 +387,12 @@ JuMP.Containers.SparseAxisArray{VariableRef,1,Tuple{Int64}} with 2 entries:
...
+
+```jldoctest; setup=:(model=Model())
+S = [(1,1,1),(10,10,10)]
+@variable(model, x[S])
```
blegat: Could you make an example with both approach and then show that e.g. one has quadratic complexity and the other one has linear complexity? For instance with a condition i == j and n elements for i and j, it will be quadratic complexity vs linear complexity

docs/src/variables.md
mlubin: Please follow the 80-character limit used in the rest of the documentation.

mlubin: "is effectively iterating" is unnecessarily imprecise, just "iterates"

docs/src/variables.md
```diff
@@ -387,6 +387,12 @@ JuMP.Containers.SparseAxisArray{VariableRef,1,Tuple{Int64}} with 2 entries:
   [2]  =  x[2]
 ```
 
+Note that with many indices and large amount of sparcity, variable construction may take more time than expected, as JuMP is effectively iterating over all indices and evaluating the conditional for each combination. The recommended work-around in such cases is to work directly with a list of tuples. For example:
+
+```jldoctest; setup=:(model=Model())
```
mlubin: Another recommended workaround is to create your own dictionary however you like:
```julia
x = Dict()
for it in complex_list()
   if complex_condition(it)
     i, j, v = complex_indices(it)
     x[i, j, v] = @variable(model)
     # Optional: if you care about pretty printing.
     set_name(x[i, j, v], "x[$i, $j, $v]")
   end
end
```

mlubin:
I don't want to merge this until we have a green light on the documentation build.

blegat:
@hellemo can you rebase against master ?

Issue #2123, opened 2019-12-26
Add support for CallbackNodeStatus


harshangrjn:
On JuMP v0.20.1: Attaching a small lazy callback implementation example below (modified based on https://github.com/JuliaOpt/JuMP.jl/blob/master/examples/callbacks.jl) and the output of CPLEX. `x_val` values are taking fractional values inside the lazy callback while this must be strictly integral as this callback must be invoked only at incumbents. Typically user callback has to do this.  

Posting here since I am using solver independent callback. This issue appears in both CPLEX and Gurobi. Looks like a bug in JuMP but let me know if I am missing anything.  
```
using JuMP, CPLEX, Gurobi

function example_lazy_constraint_mod()
    n=4
    A = [0.0 0.83513 0.564369 0.432627; 0.83513 0.0 1.93858 1.77313; 0.564369 1.93858 0.0 0.718298; 0.432627 1.77313 0.718298 0.0]
    
    model = direct_model(CPLEX.Optimizer())
    # model = direct_model(Gurobi.Optimizer())
    @variable(model, x[1:n, 1:n], Bin)
    @variable(model, y >= 0)
    @variable(model, W[1:n, 1:n])

    @objective(model, Max, y)

    @constraint(model, [i=1:n], x[i,i] == 0)
    @constraint(model, [i=1:(n-1), j=(i+1):n], x[i,j] == x[j,i])
    @constraint(model, sum(x) == 2*(n-1))   
    @constraint(model, [i=1:n], W[i,i] >= 0)
    @constraint(model, [i=1:(n-1), j=(i+1):n], W[i,j] == W[j,i])
    @constraint(model, [i=1:(n-1), j=(i+1):n], W[i,j] == -A[i,j]*x[i,j] + y)
    @constraint(model, [i=1:n], W[i,i] == sum(A[i,:] .* x[i,:]) - y)

    function my_callback_function(cb)     
        x_val = callback_value.(Ref(cb), x)      
        @show x_val
    end
    MOI.set(model, MOI.LazyConstraintCallback(), my_callback_function)
    optimize!(model)

end
example_lazy_constraint_mod()
```
CPLEX output
```
        Nodes                                         Cuts/
   Node  Left     Objective  IInf  Best Integer    Best Bound    ItCnt     Gap

*     0+    0                            0.8351        1.2827            53.59%
      0     0        1.2827     3        0.8351        1.2827        3   53.59%
[...]

PR #2127, opened 2019-12-28
Update callbacks.jl

harshangrjn:
Related to #2123 :  A small update on example_lazy_constraint() function to make sure the cuts are added only at integral solutions (see https://www.juliaopt.org/JuMP.jl/dev/callbacks/#Lazy-constraints-1)

### DIFF
--- a/examples/callbacks.jl
+++ b/examples/callbacks.jl
@@ -28,12 +28,14 @@ function example_lazy_constraint()
         lazy_called = true
         x_val = callback_value(cb_data, x)
         y_val = callback_value(cb_data, y)
-        if y_val - x_val > 1 + 1e-6
-            con = @build_constraint(y - x <= 1)
-            MOI.submit(model, MOI.LazyConstraint(cb_data), con)
-        elseif y_val + x_val > 3 + 1e-6
-            con = @build_constraint(y - x <= 1)
-            MOI.submit(model, MOI.LazyConstraint(cb_data), con)
+        if (isinteger.(x_val) == true) && (isinteger.(y_val) == true) # Add the cut only on integral solutions
+            if y_val - x_val > 1 + 1e-6
+                con = @build_constraint(y - x <= 1)
+                MOI.submit(model, MOI.LazyConstraint(cb_data), con)
+            elseif y_val + x_val > 3 + 1e-6
+                con = @build_constraint(y - x <= 1)
+                MOI.submit(model, MOI.LazyConstraint(cb_data), con)
+            end
         end
     end
     MOI.set(model, MOI.LazyConstraintCallback(), my_callback_function)


examples/callbacks.jl
```diff
@@ -28,12 +28,14 @@ function example_lazy_constraint()
         lazy_called = true
         x_val = callback_value(cb_data, x)
         y_val = callback_value(cb_data, y)
-        if y_val - x_val > 1 + 1e-6
-            con = @build_constraint(y - x <= 1)
-            MOI.submit(model, MOI.LazyConstraint(cb_data), con)
-        elseif y_val + x_val > 3 + 1e-6
-            con = @build_constraint(y - x <= 1)
-            MOI.submit(model, MOI.LazyConstraint(cb_data), con)
+        if (isinteger.(x_val) == true) && (isinteger.(y_val) == true) # Add the cut only on integral solutions
```
odow:
```Julia
if isinteger(x_val) && isinteger(y_val)
```

mlubin: Using `isinteger` isn't correct. If you don't use the same integrality tolerance as the solver, then you might fail to add cuts on a solution that the solver thinks is integral. Maybe we need to provide better support for adding a lazy constraint callback that is called only at integer solutions. We previously had the `fractional` keyword argument: 
https://github.com/JuliaOpt/JuMP.jl/blob/a9de4d0de1b34b85d35c3624bea236afec7e5bec/src/callbacks.jl#L32

blegat:
The failure on Julia v1.1 seems to be https://github.com/JuliaOpt/MathOptInterface.jl/issues/980

examples/callbacks.jl
odow: How about a `callback_status(cb_data)` which returns an enum `CBNODE_FRACTIONAL`, `CBNODE_INTEGER`, or `CBNODE_UNKNOWN`.

mlubin: I haven't thought through all the implications, but that could work also.

blegat: +1 for something like `MOI.CallbackNodeStatus`

harshangrjn: I agree that `isinteger` was just a hack for that simple example which was there. I actually wrote my own function to handle the integrality tolerance for my problem. It would be ideal to invoke lazy callback at only integral solutions, which was also my original concern.

YP-Ye: I think callback function form `function callback_function(cb_data, cb_where)` can be a proper choice, `cb_where` can be used to control where we can add constraints.

odow: @YP-Ye, currently `cb_where` is for solver-dependent callbacks. We're looking at introducing a way for solver-independent callbacks (which only take `(cb_data)`) to identify where they are being called from.

odow:
Closing as out-dated. The fix for https://github.com/jump-dev/JuMP.jl/issues/2123 is to support https://github.com/jump-dev/MathOptInterface.jl/pull/1199.

PR #2128, opened 2019-12-29
Constrained variables with @variable

blegat:
Showcase:
```julia
julia> @variable(model, [1:2] in SecondOrderCone())
2-element Array{VariableRef,1}:
 noname
 noname

julia> @variable(model, [1:2] in MOI.SecondOrderCone(2))
2-element Array{VariableRef,1}:
 noname
 noname

julia> @variable(model, p in MOI.Semiinteger(1.0, 2.0))
p

julia> model
A JuMP Model
Feasibility problem with:
Variables: 5
`Array{VariableRef,1}`-in-`MathOptInterface.SecondOrderCone`: 2 constraints
`VariableRef`-in-`MathOptInterface.Semiinteger{Float64}`: 1 constraints
Model mode: AUTOMATIC
Names registered in the model: p
```

- [x] Add tests
- [x] Add doc

Closes https://github.com/JuliaOpt/JuMP.jl/issues/2105

### DIFF
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -607,7 +607,10 @@ julia> @variable(model, y[1:3])
 julia> @constraint(model, y in SecondOrderCone())
 [y[1], y[2], y[3]] ∈ MathOptInterface.SecondOrderCone(3)
 ```
-The variables `x` are called *constrained variables*.
+The variables `x` in the example above are called *constrained variables* in
+contrast which the variables `y` which are called *free variables* because
+the constraint added on the variables is added after their creation as free
+variables.
 
 !!! warn
     When using JuMP in [Direct mode](@ref), it may be required to create
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -402,7 +402,7 @@ of tuples or create a dictionary. Consider the following examples:
 N = 10
 S = [(1, 1, 1),(N, N, N)]
 # Slow. It evaluates conditional N^3 times.
-@variable(model, x1[i=1:N, j=1:N, k=1:N; (i, j, k) in S]) 
+@variable(model, x1[i=1:N, j=1:N, k=1:N; (i, j, k) in S])
 # Fast.
 @variable(model, x2[S])
 # Fast. Manually constructs a dictionary and fills it.
@@ -611,6 +611,16 @@ The variables `x` in the example above are called *constrained variables* in
 contrast which the variables `y` which are called *free variables* because
 the constraint added on the variables is added after their creation as free
 variables.
+
+!!! note
+    Variable bounds and integrality constraints are specified when variables
+    are created, but they don't involve constrained variables, unless you
+    explicitly use the `in` syntax or the `set` keyword. For instance,
+    `@variable(model, z <= 1, Int)` and `@variable(model, upper_bound = 1,
+    integer = true)` are not constrained variables but `@variable(model,
+    z in MOI.LessThan(1.0), Int)` is a constrained variable in
+    `MOI.LessThan(1.0)` and `@variable(model, lower_bound = 1.0,
+    set = MOI.Integer())` is a constrained variable in `MOI.Integer()`.
 
 !!! warn
     When using JuMP in [Direct mode](@ref), it may be required to create
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -521,6 +521,13 @@ julia> @variable(model, x[1:2, 1:2], PSD)
  x[1,1]  x[1,2]
  x[1,2]  x[2,2]
 ```
+or using the [Constrained variables](@ref) syntax:
+```jldoctest; setup=:(model=Model())
+julia> @variable(model, x[1:2, 1:2] in PSDCone())
+2×2 LinearAlgebra.Symmetric{VariableRef,Array{VariableRef,2}}:
+ x[1,1]  x[1,2]
+ x[1,2]  x[2,2]
+```
 
 Note that `x` must be a square 2-dimensional `Array` of JuMP variables; it
 cannot be a DenseAxisArray or a SparseAxisArray.
@@ -579,6 +586,44 @@ julia> x = @variable(model, [i=1:2], base_name="x", lower_bound=i, integer=true)
 !!! warn
     Creating two named JuMP variables with the same name results in an error at
     runtime. Use anonymous variables as an alternative.
+
+## Constrained variables
+
+To create variables constrained to belong to the [`SecondOrderCone`](@ref) either
+of the following two methods can be used:
+```jldoctest constrained_variables; setup=:(model=Model())
+julia> @variable(model, x[1:3] in SecondOrderCone())
+3-element Array{VariableRef,1}:
+ x[1]
+ x[2]
+ x[3]
+
+julia> @variable(model, y[1:3])
+3-element Array{VariableRef,1}:
+ y[1]
+ y[2]
+ y[3]
+
+julia> @constraint(model, y in SecondOrderCone())
+[y[1], y[2], y[3]] ∈ MathOptInterface.SecondOrderCone(3)
+```
+The variables `x` are called *constrained variables*.
+
+!!! warn
+    When using JuMP in [Direct mode](@ref), it may be required to create
+    constrained variables instead of constraining free variables as the solver
+    may only support constrained variables. In [Automatic and Manual modes](@ref),
+    both ways of adding constraints on variables are equivalent. Indeed, during
+    the copy of the cache to the optimizer, the choice of the constraints on
+    variables that are copied as constrained variables does not depend on
+    how it was added to the cache.
+
+Additionally to the `in` syntax, the set can be specified with the `set` keyword
+argument. This is useful when creating anonymous scalar variables:
+```jldoctest constrained_variables
+julia> z = @variable(model, set = MOI.Semiinteger(1.0, 2.0))
+noname
+```
 
 ## User-defined containers
 
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -589,6 +589,15 @@ julia> x = @variable(model, [i=1:2], base_name="x", lower_bound=i, integer=true)
 
 ## Variables constrained on creation
 
+!!! info
+    When using JuMP in [Direct mode](@ref), it may be required to constrain
+    variables on creation instead of constraining free variables as the solver
+    may only support variables constrained on creation. In [Automatic and Manual
+    modes](@ref), both ways of adding constraints on variables are equivalent.
+    Indeed, during the copy of the cache to the optimizer, the choice of the
+    constraints on variables that are copied as variables constrained on creation
+    does not depend on how it was added to the cache.
+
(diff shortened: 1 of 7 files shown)


src/variables.jl
```diff
@@ -807,6 +807,46 @@ function _moi_constrain_variable(backend::MOI.ModelLike, index, info)
     end
 end
 
+"""
+    ConstrainedVariables <: AbstractVariable
```
mlubin: `AbstractVariable` is missing a docstring. Readers might get confused about how this relates to a `AbstractVariableRef`.

odow:
Does the following work (or can it be made to work)?
```julia
@variable(model, x[1:2] in SecondOrderCone())
```

blegat:
@odow Yes, I added it as a test

odow:
Ah. I missed that bit. Cool cool

docs/src/variables.md
```diff
@@ -580,6 +587,44 @@ julia> x = @variable(model, [i=1:2], base_name="x", lower_bound=i, integer=true)
...
+julia> @variable(model, y[1:3])
+3-element Array{VariableRef,1}:
+ y[1]
+ y[2]
+ y[3]
+
+julia> @constraint(model, y in SecondOrderCone())
+[y[1], y[2], y[3]] ∈ MathOptInterface.SecondOrderCone(3)
+```
+The variables `x` are called *constrained variables*.
```
mlubin: This is not consistent with the definition in MOI, "constrained variables are the variables created with add_constrained_variable or add_constrained_variables" (http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#Variables-1). I'd like to have a clear definition. Is it any variable that belongs to *at least* one `SingleVariable` or `VectorOfVariables` constraint?

src/variables.jl
```diff
@@ -3,6 +3,13 @@
 #  License, v. 2.0. If a copy of the MPL was not distributed with this
 #  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+"""
+    AbstractVariable
+
+Variable returned by [`build_variable`](@ref). It represent a variable that has
```
mlubin: s/represent/represents/

src/variables.jl
```diff
@@ -3,6 +3,13 @@
...
+not been added yet to any model, it can be added to a given `model` with
```
mlubin: run-on sentence: "not been added yet to any model. It can ..."

src/variables.jl
```diff
@@ -807,6 +814,46 @@ function _moi_constrain_variable(backend::MOI.ModelLike, index, info)
     end
 end
 
+"""
+    ConstrainedVariables <: AbstractVariable
+
+Variable `scalar_variables` constrained to belong to `set`.
+Adding this variable can be thought as doing:
```
mlubin: Nit: s/thought as/understood as/

docs/src/variables.md
blegat: Maybe the wording isn't clear but what I mean here is that `x` is called *constrained variables* as opposed to `y` (which is created free and then constrained) which isn't called like this.
Was that the confusion ?

mlubin: Ah, I misread this as saying that `y` as also a constrained variable. Perhaps clarify that `x` is a constrained variable while `y` is not.

docs/src/variables.md
```diff
@@ -580,6 +587,47 @@ julia> x = @variable(model, [i=1:2], base_name="x", lower_bound=i, integer=true)
...
+ y[1]
+ y[2]
+ y[3]
+
+julia> @constraint(model, y in SecondOrderCone())
+[y[1], y[2], y[3]] ∈ MathOptInterface.SecondOrderCone(3)
+```
+The variables `x` in the example above are called *constrained variables* in
+contrast which the variables `y` which are called *free variables* because
+the constraint added on the variables is added after their creation as free
```
mlubin: This is still not quite clear. Variable bounds and integrality constraints are specified when variables are created, but they don't involve constrained variables, unless you explicitly use `in` or `set=`.

src/macros.jl
```diff
@@ -1132,9 +1156,10 @@ macro variable(args...)
     end
 
     info_kw_args = filter(_is_info_keyword, kw_args)
-    extra_kw_args = filter(kw -> kw.args[1] != :base_name && kw.args[1] != :variable_type && !_is_info_keyword(kw), kw_args)
+    extra_kw_args = filter(kw -> kw.args[1] != :base_name && kw.args[1] != :variable_type && kw.args[1] != :set && !_is_info_keyword(kw), kw_args)
```
mlubin: Nit: Character limit

src/macros.jl
```diff
@@ -1143,17 +1168,19 @@ macro variable(args...)
...
-        var = parse_variable(_error, infoexpr, x.args...)
+        var, set = parse_variable(_error, infoexpr, x.args...)
     else
         var = x
+        set = nothing
     end
 
     anonvar = isexpr(var, :vect) || isexpr(var, :vcat) || anon_singleton
-    anonvar && explicit_comparison && _error("Cannot use explicit bounds via >=, <= with an anonymous variable")
+    anonvar && explicit_comparison && set === nothing && _error("Cannot use explicit bounds via >=, <= with an anonymous variable")
```
mlubin: Nit: character limit. I think it would be more clear to structure this as:
```julia
if ...
  error(...)
end
```

docs/src/variables.md
```diff
@@ -580,6 +587,57 @@ julia> x = @variable(model, [i=1:2], base_name="x", lower_bound=i, integer=true)
...
+
+!!! note
+    Variable bounds and integrality constraints are specified when variables
+    are created, but they don't involve constrained variables, unless you
+    explicitly use `in` or the `set` keyword. Therefore `@variable(model, z <= 1,
+    Int)` and `@variable(model, upper_bound = 1, integer = true)` are not
+    constrained variables but `@variable(model, z in MOI.LessThan(1.0), Int)` is
+    a constrained variable in `MOI.LessThan(1.0)` and `@variable(model,
+    lower_bound = 1.0, set = MOI.Integer())` is a constrained variable in
+    `MOI.Integer()`.
```
mlubin: I think the name "constrained variables" is deceivingly simple for the concept it represents. Could we open this for bikeshedding? What about "variables constrained on creation"?

blegat: Yes, let's discuss the name. "variable added constrained" is another possibility

odow: How about

---

## Restricted variables

By default, `@variable(model, x)` creates a _free_ variable that belongs to the 
set of real numbers.

If you add constraints, such as `@variable(model, x >= 0, Int)`, then the 
`@variable` macro is equivalent to:
```julia
@variable(model, x)
@constraint(model, x in MOI.GreaterThan(0.0))
@constraint(model, x in MOI.Integer())
```
Importantly, the bound and integrality constraints are added _after_ the 
variable has been created.

However, in some cases, it is necessary to supply a constraining set _at 
creation time_. We call these variables _restricted_ variables to differentiate 
them from _free_ variables.

For example, one way to create a vector of variables constrained to the second 
order cone is as follows:
```jldoctest constrained_variables; setup=:(model=Model())
julia> @variable(model, x[1:3])
3-element Array{VariableRef,1}:
x[1]
x[2]
x[3]

julia> @constraint(model, x in SecondOrderCone())
[x[1], x[2], x[3]] ∈ MathOptInterface.SecondOrderCone(3)
```

An alternative approach, creating a vector of variables _restricted_ to the 
[`SecondOrderCone`](@ref) is:
```jldoctest constrained_variables
julia> @variable(model, y[1:3] in SecondOrderCone())
3-element Array{VariableRef,1}:
y[1]
y[2]
y[3]
```
Importantly, _the second-order cone restriction on `y` cannot be deleted.

!!! info
    Restricted variables are most helpful for creating PSD matrices, where 
    creating a matrix of variables _restricted_ to the PSD cone can result in a
    smaller problem than _constraining_ the matrix to the PSD cone after 
    variable creation. Therefore, prefer
    ```julia
    @variable(model, x[1:2, 1:2] in PSDCone())
    ```
    over
    ```julia
    @variable(model, x[1:2, 1:2], PSD)
    ```

### The `set` keyword

An alternate syntax to `x in Set` is to use the `set` keyword of 
[`@variable`](@ref). This is most useful when creating anonymous variables:
```julia
x = @variable(model, [1:2, 1:2], set = PSDCone())
```

mlubin: Changing "constrained" to "restricted" doesn't really resolve my concern. Mathematically, variables can be restricted by adding constraints after they're created. Also, if you ask anyone doing linear programming if `@variable(model, x >= 0)` creates a free variable, they'll guess no. 

What we're *actually* trying to distinguish is constraints specified at creation time, so why not put the creation-time concept in the name? Deceptively short names create confusion.

We can have "variables constrained on creation" and "variables free on creation". "variable added constrained" doesn't quite parse as well for me.

odow: > Changing "constrained" to "restricted" doesn't really resolve my concern. 

I guess I was aiming at "constrained" being an overloaded term, and "restricted" has a similar meaning but we don't use it presently.

> We can have "variables constrained on creation" and "variables free on creation". 

I'm good with this.

blegat: I have updated the PR with the new name and improved the doc with @odow's suggestions

docs/src/variables.md
```diff
@@ -580,6 +587,67 @@ julia> x = @variable(model, [i=1:2], base_name="x", lower_bound=i, integer=true)
...
+An alternative approach, creating a vector of variables constrained on creation
+to belong to the [`SecondOrderCone`](@ref) is:
+```jldoctest constrained_variables
+julia> @variable(model, y[1:3] in SecondOrderCone())
+3-element Array{VariableRef,1}:
+ y[1]
+ y[2]
+ y[3]
+```
+Importantly, in [Direct mode](@ref) and for some solver, the second-order
```
mlubin: s/solver/solvers/

src/variables.jl
```diff
@@ -807,6 +814,46 @@ function _moi_constrain_variable(backend::MOI.ModelLike, index, info)
     end
 end
 
+"""
+    ConstrainedVariables <: AbstractVariable
```
mlubin: We should also propagate the new naming here, e.g., `VariableConstrainedOnCreation`.

docs/src/variables.md
```diff
@@ -580,6 +587,67 @@ julia> x = @variable(model, [i=1:2], base_name="x", lower_bound=i, integer=true)
...
+### The `set` keyword
+
+An alternate syntax to `x in Set` is to use the `set` keyword of
+[`@variable`](@ref). This is most useful when creating anonymous variables:
+```julia
+x = @variable(model, [1:2, 1:2], set = PSDCone())
+```
+
+!!! info
+    When using JuMP in [Direct mode](@ref), it may be required to constrain
```
mlubin: Perhaps note at the beginning of this section that the whole discussion is relevant only for direct mode, so that readers can skip it if they don't care about direct mode.

PR #2132, opened 2020-01-01
[RFC] Add complementarity constraints

blegat:
```julia
julia> using JuMP; model = Model(); @variable(model, x);

julia> @constraint(model, complements(x, x >= 1))
[x, x - 1] ∈ MathOptInterface.Complements(1)

julia> @constraint(model, complements(x, x <= 1))
[x, -x + 1] ∈ MathOptInterface.Complements(1)
```
What about `⟂` (that can be entered with `\perp<TAB>`) ?
Do you prefer:
```julia
@constraint(model, x ⟂ x <= 1)
```
or, similarly to the indicator constraint `@constraint(x => {x <= 1})`,
```julia
@constraint(model, x ⟂ {x <= 1})
```
The argument for the indicator constraint that the two signs were two symmetric may not hold here, see the discussion in https://github.com/JuliaOpt/JuMP.jl/pull/2092.

### DIFF
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -819,7 +819,7 @@ MathOptInterface.LessThan{Float64}(1.0)
 
 ## Complementarity constraints
 
-A mixed complementarity problem `F(x) ⟂ x` consists of finding `x` in the
+A mixed complementarity constraint `F(x) ⟂ x` consists of finding `x` in the
 interval `[lb, ub]`, such that the following holds:
 
 - `F(x) == 0` if `lb < x < ub`
@@ -829,19 +829,31 @@ interval `[lb, ub]`, such that the following holds:
 For more information, see the [`MOI.Complements` documentation](https://www.juliaopt.org/MathOptInterface.jl/v0.9/apireference/#MathOptInterface.Complements).
 
 JuMP supports mixed complementarity constraints via `complements(F(x), x)` or
-`F(x) ⟂ x` in the [`@constraint`](@ref) macro. For example:
+`F(x) ⟂ x` in the [`@constraint`](@ref) macro. The interval set `[lb, ub]` is
+obtained from the variable bounds on `x`.
 
+For example, to define the problem `2x - 1 ⟂ x` with `x ∈ [0, ∞)`, do:
 ```jldoctest complementarity; setup=:(model=Model())
 julia> @variable(model, x >= 0)
 x
-
-julia> @constraint(model, complements(2x - 1, x))
-[2 x - 1, x] ∈ MathOptInterface.Complements(1)
 
 julia> @constraint(model, 2x - 1 ⟂ x)
 [2 x - 1, x] ∈ MathOptInterface.Complements(1)
 ```
 This problem has one solution at `x = 0.5`.
+
+The perp operators `⟂` can be entered in most editors (and the Julia REPL) by
+typing `\perp<tab>`.
+
+An alternative approach that does not require the `⟂` symbol uses the
+`complements` function as follows:
+```jldoctest complementarity
+julia> @constraint(model, complements(2x - 1, x))
+[2 x - 1, x] ∈ MathOptInterface.Complements(1)
+```
+
+In both cases, the mapping `F(x)` is supplied as the first argument, and the
+matching variable `x` is supplied as the second.
 
 Vector-valued complementarity constraints are also supported:
 ```jldoctest complementarity
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -817,6 +817,52 @@ julia> con.set
 MathOptInterface.LessThan{Float64}(1.0)
 ```
 
+## Complementarity constraints
+
+A mixed complementarity problem `F(x) ⟂ x` consists of finding `x` in the
+interval `[lb, ub]`, such that the following holds:
+
+- `F(x) == 0` if `lb < x < ub`
+- `F(x) >= 0` if `lb == x`
+- `F(x) <= 0` if `x == ub`
+
+For more information, see the [`MOI.Complements` documentation](https://www.juliaopt.org/MathOptInterface.jl/v0.9/apireference/#MathOptInterface.Complements).
+
+JuMP supports mixed complementarity constraints via `complements(F(x), x)` or
+`F(x) ⟂ x` in the [`@constraint`](@ref) macro. For example:
+
+```jldoctest complementarity; setup=:(model=Model())
+julia> @variable(model, x >= 0)
+x
+
+julia> @constraint(model, complements(2x - 1, x))
+[2 x - 1, x] ∈ MathOptInterface.Complements(1)
+
+julia> @constraint(model, 2x - 1 ⟂ x)
+[2 x - 1, x] ∈ MathOptInterface.Complements(1)
+```
+This problem has one solution at `x = 0.5`.
+
+Vector-valued complementarity constraints are also supported:
+```jldoctest complementarity
+julia> @variable(model, -2 <= y[1:2] <= 2)
+2-element Array{VariableRef,1}:
+ y[1]
+ y[2]
+
+julia> M = [1 2; 3 4]
+2×2 Array{Int64,2}:
+ 1  2
+ 3  4
+
+julia> q = [5, 6]
+2-element Array{Int64,1}:
+ 5
+ 6
+
+julia> @constraint(model, M * y + q ⟂ y)
+[y[1] + 2 y[2] + 5, 3 y[1] + 4 y[2] + 6, y[1], y[2]] ∈ MathOptInterface.Complements(2)
+```
 
 ## Reference
 
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -840,9 +840,9 @@ x
 julia> @constraint(model, 2x - 1 ⟂ x)
 [2 x - 1, x] ∈ MathOptInterface.Complements(1)
 ```
-This problem has one solution at `x = 0.5`.
+This problem has a unique solution at `x = 0.5`.
 
-The perp operators `⟂` can be entered in most editors (and the Julia REPL) by
+The perp operator `⟂` can be entered in most editors (and the Julia REPL) by
 typing `\perp<tab>`.
 
 An alternative approach that does not require the `⟂` symbol uses the
--- a/src/complement.jl
+++ b/src/complement.jl
@@ -3,27 +3,54 @@
 #  License, v. 2.0. If a copy of the MPL was not distributed with this
 #  file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# This file extends JuMP to indicator constraints. It is a good example of how
-# JuMP can be extended.
+# This file extends JuMP to complementarity constraints. It is a good example of
+# how JuMP can be extended. In particular, it allows constraints of the form:
+#
+# @constraint(model, complements(2x - 1, x))
+# @constraint(model, 2x - 1 ⟂ x)
 
 function _build_complements_constraint(
-    _error::Function, funcs::Vector{<:AbstractJuMPScalar},
-    variables::Vector{<:AbstractVariableRef})
-    n = length(variables)
-    if length(funcs) != n
-        _error("Number of variables ($n) does not match number of constraints ($(length(funcs))).")
+    _error::Function,
+    F::Vector{<:AbstractJuMPScalar},
+    x::Vector{<:AbstractVariableRef}
+)
+    n = length(x)
+    if length(F) != n
+        _error(
+            "Length of mapping ($(length(F))) is not equal to number of " *
+            "matching variables ($(n))."
+        )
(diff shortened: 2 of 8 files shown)


joaquimg:
In MOI the complements and variable bounds are defined separately.

In JuMP they will be defined simultaneously?

joaquimg:
I like the perp syntax, I think it should be a possibility.
I agree that the curly braces are not as important here.

blegat:
> In JuMP they will be defined simultaneously?

No, the example is rather
```julia
model = Model()
@variable(model, x >= 0)
@variable(model, y)
@variable(model, z)
@constraint(model, complements(x, y + z <= 1))
```
meaning that `x * (1 - y - z) = 0`.

joehuchette:
What is the working definition for "complementarity constraints" used here? FWIW I found it confusing that not all the arguments to ``complements`` needed to be constraints.

+1 to the \perp syntax.

ccoffrin:
Being old fashioned, I am hesitant to use special characters like `⟂`.  As long as there is a ASCII alternative (i.e. `complements()`) that is fine by me.

Also will `complements()` support more than two arguments?

odow:
> Also will complements() support more than two arguments?

@ccoffrin do you have a definition of what that would mean and what the argument types would be?

ccoffrin:
@odow `complements(a, b, c, ... , z) -> a*b*c*...*z = 0`

odow:
> @odow `complements(a, b, c, ... , z) -> a*b*c*...*z = 0`

So to re-iterate, our definition of complements is not `complements(x, y) => x * y = 0`. 

In particular, we care _alot_ about the bound sense of `y` (see the MOI documentation for the definition):
`complements(x, y) => x * y == 0` iff `y >= 0`.

To model `x[1] * x[2] * x[3] = 0`, you can write:
```julia
model = Model()
@variable(model, x[1:3] >= 0)
@constraint(
    model, 
    complements([x[2] * x[3], x[1] * x[3], x[1] * x[2]], x)
)
```
But notice how it needs bounds on the variables.

The signature of `complements` should be the equivalent of
```Julia
function complements(::AbstractScalarFunction, ::VariableRef)
end

function complements(::AbstractVectorFunction, ::Vector{VariableRef})
end
```
where the first argument is the mapping, and the second argument is the associated variables.

ccoffrin:
@odow, ok so then it this definition of complementarity is new to me.  Is there a seminal reference you can point me to?  Is this semantics inspired by a specific solver?

odow:
> Is there a seminal reference you can point me to?

Not sure about seminal, but see, e.g., Chapter 1 of http://pages.cs.wisc.edu/~ferris/theses/ToddMunson.pdf.

> this definition of complementarity is new to me

The "classical" complementarity constraint:
```Julia
0 <= F(x) ⟂ x >= 0
```
Is an MCP with the matching (F, x) and bounds on x of `[0, inf)`, i.e.,
```Julia
@variable(model, x >= 0)
@constraint(model, F(x) ⟂ x)
```

> Is this semantics inspired by a specific solver?

I spent a lot of time talking to Michael Ferris and @xhub about how we should represent complementarity. In particular, we discussed how [AMPL](https://ampl.com/BOOK/CHAPTERS/22-complement.pdf) and [GAMS](https://www.gams.com/latest/docs/S_PATH.html) represent complementarity, and why our approach is the way to go. 

GAMS needs a matching between a constraint and a variable, but the constraint sense doesn't matter because they use `F(x) = lhs - rhs`, so GAMS suggest you use `=N=` as the constraint sense. Except that the constraint sense does matter because if you pair a `=L=` constraint with a lower bounded error GAMS will throw an error (because `x>=l` can only result in `F(x) = 0` or `F(x) >= 0`). This causes confusion for users.

The AMPL approach asks for bounds with the complementarity constraint (actually, there are multiple ways of specifying the constraint). This also leads to a multitude of modeling errors, e.g., how should we interpret:
```Julia
@variable(model, x >= 1)
@constraint(model, complements(2x - 8, x >= 2))
```
There are also a lot of different interpretations depending on the types of the LHS and RHS of the complementarity constraints. (See [Section 19.2 of the AMPL book](https://ampl.com/BOOK/CHAPTERS/22-complement.pdf).)

So, given the definition of a complementarity constraint that solvers actually implement, the simplest way for users to supply is just by matching `F` and `x`, and supplying `l` and `u` from the variable bounds on `x`.

This has numerous benefits:
 - it directly matches the form required by solvers
 - there is only one way to supply the constraint, leading to less potential for modeling errors
 - it doesn't require additional machinery in JuMP

The biggest downside for users coming to JuMP from GAMS/AMPL is that the syntax is different.

blegat:
PR updated. The now syntax is:
```julia
model = Model()
@variable(model, x >= 0)
@variable(model, y)
@variable(model, z)
@constraint(model, complements(1 - y - z, x))
# or
@constraint(model, 1 - y - z ⟂ x)
```

src/complement.jl
```diff
@@ -0,0 +1,29 @@
...
+# JuMP can be extended.
+
+function _build_complements_constraint(
+    _error::Function, funcs::Vector{<:AbstractJuMPScalar},
+    variables::Vector{<:AbstractVariableRef})
+    n = length(variables)
+    if length(funcs) != n
+        _error("Number of variables ($n) does not match number of constraints ($(length(funcs))).")
+    end
+    return VectorConstraint([variables; funcs], MOI.Complements(n))
```
odow: These are the wrong way round: `[funcs; variables]`.

src/complement.jl
```diff
@@ -0,0 +1,29 @@
...
+end
+function _build_complements_constraint(
+    _error::Function, func::AbstractJuMPScalar,
+    variable::AbstractVariableRef)
+    return VectorConstraint([variable, func], MOI.Complements(1))
```
odow: `[func, variable]`

blegat:
> Do you want me to finish this PR? I have time

The floor is yours :)

ccoffrin:
@odow, thanks for the detailed explanation.  I had a quick discussion with @hhijazi.  We were both under the impression that the most general form of complementary is `f(x)*g(x) = 0` which is logically equivalent to the disjunction `f(x) = 0 or g(x) = 0`; but I for one and not an expert in this topic.  

I had a quick skim of the AMPL and GAMS docs and it does seem that the term `complements` there is consistent with the semantics you define here.  So in that sense, this proposal is a nice way of replicating that functionality in JuMP, I agree with your arguments on how to specify the bounds.  I am wondering if `linear_complements` would be a more precise name for this feature, that seems to be a common name for the "classical" complementarity constraint from the thesis you reference?

My question is then, down the road how might JuMP reconcile the semantics difference for complements of `f(x), x` and `f(x), g(x)`?  One option might be to say that linear complements is the only set that solvers have special support for and hence all more general complements should be written in NLP form.

odow:
> My question is then, down the road how might JuMP reconcile the semantics difference for complements of f(x), x and f(x), g(x)? 

If you want to write `0 <= f(x) ⟂ g(x) >= 0`, you can formulate:
```Julia
@variable(model, x)
@variable(model, y >= 0)
@constraint(model, g(x) == y)
@constraint(model, f(x) ⟂ y)
```
The alternative is to go the AMPL route, which essentially does this reformulation anyway, but leads to confusion for users for the reasons discussed above.

> One option might be to say that linear complements is the only set that solvers have special support for and hence all more general complements should be written in NLP form.

It seems you really want to define a `ProductZero` set. That seems more explicit compared to complements.
```Julia
"""
    ProductZero(dimension::Int)

The set `{x ∈ R^{n}: x_1 * x_2 * ... * x_n == 0}`.
"""
struct ProductZero <: AbstractVectorSet
    dimension::Int
end

model = Model()
@variable(model, x[1:2])
@constraint(model, x in ProductZero(2))
@constraint(model, [f(x), x] in ProductZero(2))
@constraint(model, [f(x), g(x)] in ProductZero(2))
```

ccoffrin:
> It seems you really want to define a ProductZero set. That seems more explicit compared to complements.

So it sounds like `@constraint(model, [f(x), x] in ProductZero(2))` is just a more general form of `@constraint(model, complements(f(x), x))`, so why should we add `complements`?

odow:
> So it sounds like `@constraint(model, [f(x), x] in ProductZero(2))` is just a more general form of `@constraint(model, complements(f(x), x))` so why should we add complements?

They're different concepts with a slight overlap.

`f(x) * x == 0` is equivalent to
```Julia
@variable(model, x)
@constraint(model, f(x) * x ⟂ x)
```
_iff_ `x` is free.

In contrast, the following MCP
```Julia
@variable(model, x >= 1)
@constraint(model, 2x + 1 ⟂ x)
```
has one solution, `x = 1`. Therefore `F(x) = 3` so `F(x) * x = 3`.

Are there any solvers with special support (not reformulations) for `f(x) * g(x) == 0` that would benefit from knowing `f(x)` and `g(x)` separately and support bounds on `x`?

blegat:
Approved ! (I can't do an approval review to my own PR so I just write it)

xhub:
First thanks guys for pushing this forward! General +1 for the explanations from @odow 
 
> Are there any solvers with special support (not reformulations) for `f(x) * g(x) == 0` that would benefit from knowing `f(x)` and `g(x)` separately and support bounds on `x`?

Inconsistent terminology is a pain point of the complementarity world.
If we follow the following review http://pages.cs.wisc.edu/~ferris/papers/sirev-comp-app.pdf  in section 2.3m the problem `f(x) >=0, g(x) >= 0, f(x) * g(x) == 0` would be known as a _vertical complementarity problem_. 

As in the aforementioned review, it can be extended to `n` terms by using the `min` formulation: `min(f,g,h,...) = 0`. I don't know of any solver exploiting this structure. The review paper is more than 20 years old and hinges on the fact that there may be specialized algorithms for this problem class. I am not aware of any such development either on the theoretical side or in the implementation.

Note that this is more specific than the "disjunction" `f*g*h*... = 0`, since the latter doesn't have a restriction on the signs of each functional `f, g, h, ...`

IIRC, once the NLP codepath in MOI https://github.com/JuliaOpt/MathOptInterface.jl/issues/846 has been reworked, one can use the same syntax to define nonlinear complementarity problems.

docs/src/constraints.md
```diff
@@ -817,6 +817,64 @@ julia> con.set
...
+For example, to define the problem `2x - 1 ⟂ x` with `x ∈ [0, ∞)`, do:
+```jldoctest complementarity; setup=:(model=Model())
+julia> @variable(model, x >= 0)
+x
+
+julia> @constraint(model, 2x - 1 ⟂ x)
+[2 x - 1, x] ∈ MathOptInterface.Complements(1)
+```
+This problem has one solution at `x = 0.5`.
+
```
xhub: [pedantic] This problem has a unique solution at (?)

ccoffrin:
> Are there any solvers with special support (not reformulations) for f(x) * g(x) == 0 that would benefit from knowing f(x) and g(x) separately and support bounds on x?

I am not aware of any such solvers.  @frapac do you have any insights?  I once heard that KNITRO had some special support for complementary constraints in NLPs.

My only question at this point is if `complements` is the best name for this feature, which seems have a fairly specific semantics for a concept with a variety of interpretations.

odow:
KNITRO supports complementarity constraints between two variables iff they both have lower bounds of `0.0`:
https://www.artelys.com/docs/knitro/2_userGuide/complementarity.html

So you can _only_ do this:
```Julia
@variable(model, x >= 0)
@variable(model, y >= 0)
@constraint(model, x ⟂ y)
```

The KNITRO C function for adding complementarity constraints follows our approach: you just give the two variables, and the bounds are added/checked separately.

> My only question at this point is if complements is the best name for this feature, which seems have a fairly specific semantics for a concept with a variety of interpretations.

"Complements" seems like the widely accepted term amongst people solving complementarity problems, and I don't think we should move away for it. In addition, it's what all solvers who support this feature call it. 

The confusion seems to be from the NLP modeling side that `f(x) * g(x) = 0` is a complementarity constraint, when really this is only true in a few special cases.

ccoffrin:
> "Complements" seems like the widely accepted term amongst people solving complementarity problems, and I don't think we should move away for it. In addition, it's what all solvers who support this feature call it.

Ok, sounds good to me.

> The confusion seems to be from the NLP modeling side that f(x) * g(x) = 0 is a complementarity constraint, when really this is only true in a few special cases.

Indeed, I will keep exploring this to understand better.  It sounds like the term "complementarity constraint" is being used in some cases where "disjunctive constraint" would be more appropriate.

odow:
@blegat, this seems like some MutableArithmetics issue with Julia 1.3?

odow:
@blegat how does broadcasting fit into this?

e.g, how can we catch and handle:
```Julia
@constraint(model, complements.(2x + 1, x))
```

blegat:
The broadcasting on the complement operator is ignored as we are just concatenating the lhs and rhs. So you can use it or not, both should work

PR #2134, opened 2020-01-07
Update documentation and examples for callbacks in automatic mode

### LINKED ISSUE #2124: Callbacks in automatic and manual mode
It should work now with MOI v0.9.8, we should

- [ ] Update MOI bound to v0.9.8
- [ ] Add test in automatic mode
- [ ] Add example in autimatic mode
- [ ] Remove warning in doc

odow:
<s>Example is going to fail, because GLPK is missing some supports: https://github.com/JuliaOpt/GLPK.jl/pull/125

I've verified that things work with that PR merged, so we just need to wait for a new release and tag of GLPK.</s>

Closes #2124

### DIFF
--- a/docs/src/callbacks.md
+++ b/docs/src/callbacks.md
@@ -23,10 +23,6 @@ callbacks:
  2. user-cuts
  3. heuristic solutions
 
-!!! warning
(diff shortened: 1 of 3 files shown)


odow:
Woo! Callbacks in automatic mode 🚀

PR #2135, opened 2020-01-08
Create a version of method delete that deletes a vector of variables.

henriquebecker91:
This calls the version of MOI.delete that also works over a vector of variables, and is specialized by solvers for greater performance than repeatedly calling the single variable version.

The code was suggested by @odow in a discussion with him and @mlubin.

This is related to:
https://github.com/JuliaOpt/MathOptInterface.jl/pull/989
https://github.com/JuliaOpt/Gurobi.jl/pull/282
https://github.com/JuliaOpt/CPLEX.jl/pull/273

The CPLEX and Gurobi PR's are already closed, the MathOptInterface is just waiting more reviewers (I think). Anyway, MathOptInterface already has a "delete a vector of variables" fallback method  (https://github.com/JuliaOpt/MathOptInterface.jl/blob/c4beceb4e9444c752f7e655f7a50a38a70181b06/src/indextypes.jl#L124-L128). So this addition to JuMP would already be working if it was approved immediately. I suppose tests may be necessary, but it is MOI that does the heavy lifting and this method is basically the same as the single variable version except by the type of the parameters and the version of the MOI method it calls.

I imported the JuMP package changed by my branch in the REPL and called the method to test against typos.

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -173,6 +173,22 @@ function delete(model::Model, con_ref::ConstraintRef{Model})
 end
 
 """
+    delete(model::Model, con_refs::Vector{ConstraintRef})
+
+Delete the constraints associated with `con_refs` from the model `model`.
+Solvers may implement methods for deleting multiple constraints that are
+more efficient than repeatedly calling the single constraint delete method.
+"""
+function delete(model::Model, con_refs::Vector{<:ConstraintRef{Model}})
+    # This is just a fallback of the fallback. To be really useful,
+    # instead of calling JuMP.delete repeatedly it needs to call a
+    # (currently unimplemented) fallback in MOI like the one for
+    # batch variable deletion: https://github.com/JuliaOpt/MathOptInterface.jl/blob/78c15baee/src/indextypes.jl#L119
+    # Then the solvers may reimplement the fallback in MOI level.
+    delete.(model, con_refs)
+end
+
+"""
     is_valid(model::Model, con_ref::ConstraintRef{Model})
 
 Return `true` if `constraint_ref` refers to a valid constraint in `model`.
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -180,12 +180,12 @@ Solvers may implement methods for deleting multiple constraints that are
 more efficient than repeatedly calling the single constraint delete method.
 """
 function delete(model::Model, con_refs::Vector{<:ConstraintRef{Model}})
-    # This is just a fallback of the fallback. To be really useful,
-    # instead of calling JuMP.delete repeatedly it needs to call a
-    # (currently unimplemented) fallback in MOI like the one for
-    # batch variable deletion: https://github.com/JuliaOpt/MathOptInterface.jl/blob/78c15baee/src/indextypes.jl#L119
-    # Then the solvers may reimplement the fallback in MOI level.
-    delete.(model, con_refs)
+    if any(c -> model !== c.model, con_refs)
+        error("A constraint reference you are trying to delete does not" * "
+            belong to the model.")
+    end
+    MOI.delete(backend(model), index.(con_refs))
+    return
 end
 
 """
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -177,8 +177,9 @@ end
 
 Delete the constraints associated with `con_refs` from the model `model`.
 Solvers may implement specialized methods for deleting multiple constraints of
-the same concrete type. These may be more efficient than repeatedly calling the
-single constraint delete method.
+the same concrete type, i.e., when `isconcretetype(eltype(con_refs))`. These
+may be more efficient than repeatedly calling the single constraint delete
+method.
 """
 function delete(model::Model, con_refs::Vector{<:ConstraintRef{Model}})
     if any(c -> model !== c.model, con_refs)
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -164,7 +164,7 @@ end
 
 Delete the constraint associated with `constraint_ref` from the model `model`.
 """
-function delete(model::Model, con_ref::ConstraintRef{Model})
+function delete(model::Model, con_ref::ConstraintRef)
     if model !== con_ref.model
         error("The constraint reference you are trying to delete does not " *
               "belong to the model.")
@@ -173,11 +173,12 @@ function delete(model::Model, con_ref::ConstraintRef{Model})
 end
 
 """
-    delete(model::Model, con_refs::Vector{ConstraintRef})
+    delete(model::Model, con_refs::Vector{<:ConstraintRef})
 
 Delete the constraints associated with `con_refs` from the model `model`.
-Solvers may implement methods for deleting multiple constraints that are
-more efficient than repeatedly calling the single constraint delete method.
+Solvers may implement specialized methods for deleting multiple constraints of
+the same concrete type. These may be more efficient than repeatedly calling the
+single constraint delete method.
 """
 function delete(model::Model, con_refs::Vector{<:ConstraintRef{Model}})
     if any(c -> model !== c.model, con_refs)
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -180,8 +180,8 @@ end
     delete(model::Model, variable_refs::Vector{VariableRef})
 
 Delete the variables associated with `variable_refs` from the model `model`.
-The Solvers may implement methods for deleting multiple variables that are
-more efficient than repeatedely calling the single variable delete method.
+Solvers may implement methods for deleting multiple variables that are
+more efficient than repeatedly calling the single variable delete method.
 """
 function delete(model::Model, variable_refs::Vector{VariableRef})
     if any(model !== owner_model(v) for v in variable_refs)
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -177,6 +177,22 @@ function delete(model::Model, variable_ref::VariableRef)
 end
 
 """
+    delete(model::Model, variable_refs::VariableRef)
+
+Delete the variables associated with `variable_refs` from the model `model`.
+The solvers may implement methods for deleting multiple variables that are
+more efficient than repeatedely calling the single variable delete method.
+"""
+function delete(model::Model, variable_refs::Vector{VariableRef})
+    if any(model !== owner_model(v) for v in variable_refs)
+        error("A variable reference you are trying to delete does not " *
+              "belong to the model.")
+    end
+    MOI.delete(backend(model), index.(variable_refs))
+    return
+end
+
+"""
     is_valid(model::Model, variable_ref::VariableRef)
 
 Return `true` if `variable` refers to a valid variable in `model`.
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -177,10 +177,10 @@ function delete(model::Model, variable_ref::VariableRef)
 end
 
 """
-    delete(model::Model, variable_refs::VariableRef)
+    delete(model::Model, variable_refs::Vector{VariableRef})
 
 Delete the variables associated with `variable_refs` from the model `model`.
-The solvers may implement methods for deleting multiple variables that are
+The Solvers may implement methods for deleting multiple variables that are
 more efficient than repeatedely calling the single variable delete method.
 """
 function delete(model::Model, variable_refs::Vector{VariableRef})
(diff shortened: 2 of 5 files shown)


odow:
Can you add a test to test/variables.jl please? Can just be the same one as the one you added to MOI (but with JuMP syntax).

henriquebecker91:
Will do it.

henriquebecker91:
Added the tests and changed JuMPExtensions to implement the batch delete and pass the tests too. The tests are similar to the [ones for a single variable in JuMP](https://github.com/JuliaOpt/JuMP.jl/blob/7a656f69cebaa718edba061523281007304fd0cc/test/variable.jl#L182) and to the [ones over multiple variables in MOI](https://github.com/henriquebecker91/MathOptInterface.jl/blob/hb/batch_delete_test/src/Test/UnitTests/modifications.jl#L357).

I think it is now ready for review.

src/variables.jl
```diff
@@ -176,6 +176,22 @@ function delete(model::Model, variable_ref::VariableRef)
     MOI.delete(backend(model), variable_ref.index)
 end
 
+"""
+    delete(model::Model, variable_refs::VariableRef)
```
mlubin: The signature doesn't match. Should be `Vector{VariableRef}`.

src/variables.jl
```diff
@@ -176,6 +176,22 @@ function delete(model::Model, variable_ref::VariableRef)
...
+
+Delete the variables associated with `variable_refs` from the model `model`.
+The solvers may implement methods for deleting multiple variables that are
```
mlubin: "The solvers" -> "Solvers"

test/JuMPExtension.jl
```diff
@@ -75,6 +75,13 @@ function JuMP.delete(model::MyModel, vref::MyVariableRef)
     delete!(model.variables, vref.idx)
     delete!(model.var_to_name, vref.idx)
 end
+function JuMP.delete(model::MyModel, vrefs::Vector{MyVariableRef})
+    @assert all(JuMP.is_valid.(model, vrefs))
```
mlubin: I don't see a need for a specialized implementation in this case. The implementation could just be `JuMP.delete.(model, vrefs)`. This avoids the need to duplicate the deletion logic.

test/variable.jl
```diff
@@ -500,6 +500,19 @@ function test_variables_constrained_on_creation(ModelType)
...
+function test_batch_delete_variables(ModelType)
+    model = ModelType()
+    @variable(model, x[1:3] >= 1)
+    @objective(model, Min, sum([1, 2, 3] .* x))
+    @test all(is_valid.(model, x))
+    delete(model, x[[1, 3]])
+    @test all((!is_valid).(model, x[[1, 3]]))
+    @test is_valid(model, x[2])
+    second_model = ModelType()
+    @test_throws Exception JuMP.delete(second_model, x[2])
```
mlubin: We usually have more specific tests that include the error messages. See examples above.

mlubin: I see that `test_variable_is_valid_delete` doesn't do any better, so feel free to ignore this advice.

henriquebecker91: The problem with this approach is that the current JuMPExtension does not throw the same error message, nor the same specific type of exception. For the single variable deletion (that was already there and I did not touch), [it is an error in the JuMP](https://github.com/JuliaOpt/JuMP.jl/blob/master/src/variables.jl#L173) and [an assertion in JuMPExtension](https://github.com/JuliaOpt/JuMP.jl/blob/master/test/JuMPExtension.jl#L74). I think its is not the place of this PR to fix this.

test/JuMPExtension.jl
henriquebecker91: Fixed.

src/variables.jl
henriquebecker91: Fixed.

src/variables.jl
henriquebecker91: Fixed.

src/variables.jl
```diff
@@ -176,6 +176,22 @@ function delete(model::Model, variable_ref::VariableRef)
     MOI.delete(backend(model), variable_ref.index)
 end
 
+"""
+    delete(model::Model, variable_refs::Vector{VariableRef})
+
+Delete the variables associated with `variable_refs` from the model `model`.
+The Solvers may implement methods for deleting multiple variables that are
```
mlubin: Remove "The", just "Solvers"

src/variables.jl
```diff
@@ -176,6 +176,22 @@ function delete(model::Model, variable_ref::VariableRef)
...
+more efficient than repeatedely calling the single variable delete method.
```
mlubin: Typo: repeatedly

src/variables.jl
henriquebecker91: Fixed.

src/variables.jl
henriquebecker91: Fixed.

henriquebecker91:
I will implement the batch constraint deletion in JuMP too. However, I think it would be good to keep in mind that:
1) I do not know if there are advantages for batch constraint deletion. Probably there are, I just really never checked.
2) I already wondered why there was not a batch deletion method for variables without existing one for constraints or something similar, XD.
3) I am a little worried that having a batch deletion of constraints will lure the users to think this has, or will have in the close future, a performance boost, when we do not have immediate plans for that nor are sure there will be such gain. On the plus side, if there is a gain and it is implemented in the future, the sooner we have such method, the more users that will use it without thinking and will be positively affected when it is specialized.
4) I will:
4.1) Implement a trivial fallback delete. Make the same comment about performance I did in the batch variable delete version. 
4.2) Do the same for JuMPExtensions.
4.3) Add tests for this method to be executed over both a JuMP model and a JuMPExtensions model.

src/constraints.jl
```diff
@@ -172,6 +172,22 @@ function delete(model::Model, con_ref::ConstraintRef{Model})
...
+Solvers may implement methods for deleting multiple constraints that are
+more efficient than repeatedly calling the single constraint delete method.
+"""
+function delete(model::Model, con_refs::Vector{<:ConstraintRef{Model}})
+    # This is just a fallback of the fallback. To be really useful,
+    # instead of calling JuMP.delete repeatedly it needs to call a
+    # (currently unimplemented) fallback in MOI like the one for
+    # batch variable deletion: https://github.com/JuliaOpt/MathOptInterface.jl/blob/78c15baee/src/indextypes.jl#L119
+    # Then the solvers may reimplement the fallback in MOI level.
+    delete.(model, con_refs)
```
odow: https://github.com/JuliaOpt/MathOptInterface.jl/blob/2a8603407fdabce4f322cd90fe646f3a0efdc6a6/src/indextypes.jl#L126-L130
The MOI fallback actually covers constraint indices.

You you need something like:
```Julia
function delete(model::Model, con_refs::Vector{<:ConstraintRef})
    if any(c -> model !== c.model, con_refs)
        error("A constraint reference you are trying to delete does not belong to the model.")
    end
    MOI.delete(backend(model), index.(con_refs))
    return
end
```

henriquebecker91:
I implemented the items 4.1, 4.2, and 4.3 from my last comment.

To make it work, I defined the method signature as:
`function delete(model::Model, con_refs::Vector{<:ConstraintRef{Model}})`
instead of
`function delete(model::Model, con_refs::Vector{ConstraintRef{Model}})`
that was my first try. I am not sure if what I did is correct (I have not a good mental model for how these types should interact in JuMP design). Should I change the variable deletion methods to work in subtypes too? Or should I make the code work without accepting constraint reference subtypes in the constraint deletion methods ? (The tests fail with the second signature, not sure why.) Or only the constraint deletion methods need to work on subtypes (like it is now) and variable deletion should also stay the way it is now?

src/constraints.jl
henriquebecker91: Will do. In the meanwhile, can you check the doubts from my last comment?

henriquebecker91: Done.

odow:
`ConstraintRef` has multiple type parameters:
https://github.com/JuliaOpt/JuMP.jl/blob/7a656f69cebaa718edba061523281007304fd0cc/src/constraints.jl#L16-L20
whereas `VariableRef` doesn't have any
https://github.com/JuliaOpt/JuMP.jl/blob/7a656f69cebaa718edba061523281007304fd0cc/src/variables.jl#L105-L108

An easier way to reason about this is that `[1, 2] isa Vector{Int} == true`, `[1, 2] isa Vector{Integer} == false`, and `[1, 2] isa Vector{<:Integer} == true`. The second case is saying "is this a vector with eltype exactly `Integer`?" To which the answer is no, because `Int != Integer`. However, the third case is saying "is this a vector with eltype that is a subtype of `Integer`?" To which the answer is true because `Int <: Integer`.

So `Vector{VariableRef}` is okay, but `Vector{ConstraintRef}` isn't, because it's really `Vector{ConstraintRef{M, C, Shape} where {M, C, Shape}}`. You need `Vector{<:ConstraintRef}`.

test/variable.jl
henriquebecker91: For now, I will leave this out of this PR.

henriquebecker91:
Thanks for the explanation, @odow. Taking this into account, then there is no change to be done in the signature of the methods.

src/constraints.jl
```diff
@@ -172,6 +172,22 @@ function delete(model::Model, con_ref::ConstraintRef{Model})
     MOI.delete(backend(model), index(con_ref))
 end
 
+"""
+    delete(model::Model, con_refs::Vector{ConstraintRef})
```
mlubin: The signature in the documentation should match the function signature (`Vector{<:ConstraintRef{Model}}` vs `Vector{ConstraintRef}`)

src/constraints.jl
```diff
@@ -172,6 +172,22 @@ function delete(model::Model, con_ref::ConstraintRef{Model})
...
+
+Delete the constraints associated with `con_refs` from the model `model`.
+Solvers may implement methods for deleting multiple constraints that are
+more efficient than repeatedly calling the single constraint delete method.
```
mlubin: It's important to note that solvers are only expected to implement more efficient fallbacks when the element type of the index vector is concrete ([docs](http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#MathOptInterface.delete-Tuple{MathOptInterface.ModelLike,Array{#s1,1}%20where%20#s1%3C:Union{VariableIndex,%20ConstraintIndex}})). I.e., all constraints referenced are of the same function-in-set type.

src/constraints.jl
henriquebecker91: I can do it. However, I think it is worth pointing that the single variable delete has signature:
`delete(model::Model, con_ref::ConstraintRef{Model})`
but the documentation comment is:
`delete(model::Model, con_ref::ConstraintRef)`
So I thought the preferred style was to drop the type parameters if they are little relevant.

Should I fix the documentation of the single variable method too?

src/constraints.jl
henriquebecker91: Is this a good wording?

> Solvers may implement specialized methods for deleting multiple constraints of
> the same concrete type. These may be more efficient than repeatedly calling the
> single constraint delete method.

src/constraints.jl
mlubin: These are two different cases. No instance of `Vector{<:ConstraintRef{Model}}` will match to  `Vector{ConstraintRef}`, so this is plain incorrect. See the discussion [here](https://docs.julialang.org/en/v1/manual/types/index.html#Parametric-Composite-Types-1).

For the single-variable delete, any `ConstraintRef{Model}` is also a `ConstraintRef`, so this is broadening the scope of the method. I'm not sure how I feel about the documentation not matching the method signature, but this case is much less misleading.

src/constraints.jl
mlubin: Yes, sounds good.

src/constraints.jl
henriquebecker91: You are correct about the differences between the cases, I had not considered it that way. I will drop the the type parameters to keep conformance with how it is already done and because I have no opinion on this (nor I think this PR should bother about this). If you think this is not good you can change it at any moment.

src/constraints.jl
henriquebecker91: Will be done then.

henriquebecker91:
Thanks for the support and agility @mlubin, unfortunately, I will probably not contribute much for some  time but I hope in a more distant future I can help to enhance some other areas (like the lazy update layer of Gurobi.jl or something about the solver specific callbacks).

PR #2141, opened 2020-01-22
Use UninstantiatedOptimizer of MOI

blegat:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/2138#issuecomment-575680880
See https://github.com/JuliaOpt/MathOptInterface.jl/pull/1008


src/JuMP.jl
```diff
@@ -53,38 +53,34 @@ include("utils.jl")
...
+    if isempty(kwargs)
+        deprecation_message = """
+with_optimizer is deprecated, replace `with_optimizer(Ipopt.Optimizer)` by `Ipopt.Optimizer`.
+"""
+        Base.depwarn(deprecation_message, :with_optimizer)
+        return constructor
+    else
+        deprecation_message = """
+`with_optimizer` is deprecated, replace `with_optimizer(Ipopt.Optimizer, max_cpu_time=60.0)`
+by `MOI.OptimizerWithAttributes(Ipopt.Optimizer, "max_cpu_time" => 60.0)`.
```
mlubin: Let's re-export OptimizerWithAttributes. This is essential and shouldn't be hidden in a submodule.

src/JuMP.jl
```diff
@@ -53,38 +53,34 @@ include("utils.jl")
...
+end
+function with_optimizer(constructor, args...; kwargs...)
     deprecation_message = """
-with_optimizer is deprecated. The examples below demonstrate how to update to the new syntax:
-- 'with_optimizer(Ipopt.Optimizer)' becomes 'Ipopt.Optimizer'.
-- 'set_optimizer(model, with_optimizer(Ipopt.Optimizer, print_level=1, tol=1e-5))' becomes 'set_optimizer(model, Ipopt.Optimizer); set_parameters(model, \"print_level\" => 1, \"tol\" => 1e-5)'.
-- In rare cases where an argument must be passed to the constructor, use an anonymous function. For example, 'env = Gurobi.Env(); set_optimizer(model, with_optimizer(Gurobi.Optimizer, env))' becomes 'env = Gurobi.Env(); set_optimizer(model, () -> Gurobi.Optimizer(env))'.
-    """
-    Base.depwarn(deprecation_message, :with_optimizer)
+`with_optimizer` is deprecated, replace `with_optimizer(Ipopt.Optimizer, args...)` by `() -> Ipopt.Optimizer(args...)`.
```
mlubin: This message implies that keyword arguments should be passed to the constructor, but we're trying to deprecate that as well.

src/optimizer_interface.jl
```diff
@@ -63,38 +56,19 @@ model = Model()
...
-        error("The provided optimizer_factory returned an object of type " *
-              "$(typeof(optimizer)). Expected a " *
-              "MathOptInterface.AbstractOptimizer.")
-    end
-    if !MOI.is_empty(optimizer)
-        error("The provided optimizer_factory returned a non-empty optimizer.")
-    end
     if bridge_constraints
         # The names are handled by the first caching optimizer.
         # If default_copy_to without names is supported, no need for a second
```
mlubin: Is this comment still relevant? The corresponding `if` statement was removed.

src/JuMP.jl
blegat: This is positional for arguments

src/optimizer_interface.jl
blegat: This is to explain why we have `with_names = false`

src/JuMP.jl
blegat: From the JuMP-dev called, it seems it should be `JuMP.optimizer_with_attributes`

mlubin: That works for me also.

src/JuMP.jl
mlubin: If you call `with_optimizer(Ipopt.Optimizer, arg1, kwarg1 = arg2)` then you'll see this message that could easily be understood as implying you should write `() -> Ipopt.Optimizer(arg1, kwarg1 = arg2)`.

src/optimizer_interface.jl
mlubin: It doesn't explicitly mention `with_names`. What about formulating it as "We set with_names=false because ..."

src/JuMP.jl
blegat: You'll see the other warning too because we redirect to with_optimizer

mlubin: Seeing two deprecation warnings, neither of which provides clear guidance for your situation, isn't a good user experience.

src/JuMP.jl
```diff
@@ -53,38 +53,59 @@ include("utils.jl")
...
-    if !applicable(constructor, args...)
-        error("$constructor does not have any method with arguments $args.",
-              "The first argument of `with_optimizer` should be callable with",
-              " the other argument of `with_optimizer`.")
-    end
-    return OptimizerFactory(constructor, args, kwargs)
+"""
+    optimizer_with_attributes(optimizer_constructor, params::Pair...)
+
+Create an `MOI.OptimizerWithAttributes` grouping an optimizer constructor with
```
mlubin: Example(s) needed. Also, this is basic JuMP functionality; we should try to first describe it in a way that doesn't require someone to understand MOI. We can say that it's equivalent to `MOI.OptimizerWithAttributes`, but that shouldn't be the only definition.

src/JuMP.jl
```diff
@@ -56,8 +56,15 @@ const _MOICON{F,S} = MOI.ConstraintIndex{F,S}
...
-the list of parameters `params`.
+Groups an optimizer constructor with the list of parameters `params`. Note that
+it is equivalent to `MOI.OptimizerWithAttributes`.
+
+## Examples
+
+The call `optimizer_with_attributes(Gurobi.Optimizer, "Presolve" => 0)` groups
+a Gurobi optimizer with the parameter `Presolve` with value `0`. When
+instantiated, it creates a Gurobi optimizer and then set the `Presolve`
+parameter to value `0`.
```
mlubin: It would also be helpful to mention/show that you can pass this object to `set_optimzier` or the model constructor.

src/JuMP.jl
```diff
@@ -53,38 +53,92 @@ include("utils.jl")
...
+    optimizer_with_attributes(optimizer_constructor, params::Pair...)
+
+Groups an optimizer constructor with the list of parameters `params`. Note that
+it is equivalent to `MOI.OptimizerWithAttributes`.
+
+## Examples
+
+The call `optimizer_with_attributes(Gurobi.Optimizer, "Presolve" => 0)` groups
+a Gurobi optimizer with the parameter `Presolve` with value `0`. When
+provided to to `Model` constructor, it creates a Gurobi optimizer and then set
```
mlubin: or `set_optimizer`

src/JuMP.jl
```diff
@@ -53,38 +53,92 @@ include("utils.jl")
...
+```
+Alternatively, the parameter can be provided separately with either
+[`set_parameter`](@ref):
+```julia
+julia> model = Model(Gurobi.Optimizer);
+
+julia> set_parameter(model, "Presolve", 0)
+0
+
+julia> MOI.get(model, MOI.RawParameter("Presolve"))
```
odow: I guess we should have `get_parameter`?

Also, did we make a decision about deprecating `set_parameter` in favor of `set_attribute`?

src/JuMP.jl
```diff
@@ -53,38 +53,92 @@ include("utils.jl")
...
+0
+```
+or [`set_parameters`](@ref):
+```julia
+julia> model = Model(Gurobi.Optimizer);
+
+julia> set_parameters(model, "Presolve" => 0)
```
odow: This example should really show that multiple parameters can be set.

src/JuMP.jl
blegat: Not sure, we should open an issue to discuss this

blegat: See https://github.com/JuliaOpt/JuMP.jl/issues/2149

blegat:
Merging to avoid conflict with https://github.com/JuliaOpt/JuMP.jl/pull/2150

PR #2147, opened 2020-02-01
Support for skew symmetric variables

### LINKED ISSUE #2146: Support for skew-symmetric matrix variables
I contribute to [PowerModelsDistribution](https://github.com/lanl-ansi/PowerModelsDistribution.jl). Ideally, we would like to use JuMP's support for matrix variables with symmetry more broadly. 

As we start from a complex-value problem with Hermitian matrix variables, we would love to see support for both symmetric (real part) and skew-symmetric (imaginary part) matrix variables. We have a work-around, but the printing of the variables looks different from other variables.

We can do something like the following to make sets of full and symmetric matrix variables:
```
Dict(i => JuMP.@variable(pm.model,
            [c in 1:N, d in 1:N], base_name="P_$(i)",
        ) for i in ids  )
Dict(i => JuMP.@variable(pm.model,
            [c in 1:N, d in 1:N], base_name="Wr_$(i)", Symmetric
        ) for i in ids  )
```
but the following is not supported:
```
Dict(i => JuMP.@variable(pm.model,
            [c in 1:N, d in 1:N], base_name="Wi_$(i)", SkewSymmetric
        ) for i in ids  )
```
Ideally, this would have (support for) zeros on the diagonal as well.

After some preliminary digging, it seems Julia does not have skew symmetric matrix variable primitives (https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#Special-matrices-1, somewhat related discussion in https://github.com/JuliaLang/julia/issues/21224)

frederikgeth:
This is an attempt to support skew symmetric variables (https://github.com/JuliaOpt/JuMP.jl/issues/2146).

Note that the return type (affine expression) is different from the input type (variable ref) [here]( https://github.com/frederikgeth/JuMP.jl/blob/22fc8d29e549d7fee9666fc4d925b8eb03be785c/src/sd.jl#L113).

Keen to hear your feedback!

Closes #2146

### DIFF
--- a/docs/src/reference/solutions.md
+++ b/docs/src/reference/solutions.md
@@ -47,4 +47,6 @@ MOI.ConflictParticipationStatusCode
 
 lp_objective_perturbation_range
 lp_rhs_perturbation_range
+lp_sensitivity_report
+SensitivityReport
 ```
--- a/docs/src/solutions.md
+++ b/docs/src/solutions.md
@@ -173,88 +173,107 @@ optimize!(model)
 ## Sensitivity analysis for LP
 
 Given an LP problem and an optimal solution corresponding to a basis, we can
-question how much an objective coefficient or standard form rhs coefficient
-(c.f., [`normalized_rhs`](@ref)) can change without violating primal or dual
-feasibility of the basic solution. Note that not all solvers compute the basis,
-and for sensitivity analysis, the solver interface must implement
-`MOI.ConstraintBasisStatus`.
+question how much an objective coefficient or standard form right-hand side
+coefficient (c.f., [`normalized_rhs`](@ref)) can change without violating primal
+or dual feasibility of the basic solution.
 
-Given an LP optimal solution (and both [`has_values`](@ref) and
-[`has_duals`](@ref) returns `true`) [`lp_objective_perturbation_range`](@ref)
-returns a range of the allowed perturbation of the cost coefficient
-corresponding to the input variable. Note that the current primal solution
-remains optimal within this range, however the corresponding dual solution might
-change since a cost coefficient is perturbed. Similarly,
-[`lp_rhs_perturbation_range`](@ref) returns a range of the allowed perturbation
-of the rhs coefficient corresponding to the input constraint. And in this range
-the current dual solution remains optimal but the primal solution might change
-since a rhs coefficient is perturbed.
-
-However, if the problem is degenerate, there are multiple optimal bases and
-hence these ranges might not be as intuitive and seem too narrow. E.g., a larger
-cost coefficient perturbation might not invalidate the optimality of the current
-primal solution. Moreover, if a problem is degenerate, due to finite precision,
-it can happen that, e.g., a perturbation seems to invalidate a basis even though
-it doesn't (again providing too narrow ranges). To prevent this
-`feasibility_tolerance` and `optimality_tolerance` is introduced, which in turn,
-might make the ranges too wide for numerically challenging instances. Thus do not
-blindly trust these ranges, especially not for highly degenerate or numerically
-unstable instances.
+Note that not all solvers compute the basis, and for sensitivity analysis, the
+solver interface must implement `MOI.ConstraintBasisStatus`.
 
 To give a simple example, we could analyze the sensitivity of the optimal
 solution to the following (non-degenerate) LP problem:
 
 ```julia
-julia> model = Model();
-julia> @variable(model, x[1:2]);
-julia> @constraint(model, c1, x[1] + x[2] <= 1);
-julia> @constraint(model, c2, x[1] - x[2] <= 1);
-julia> @constraint(model, c3, -0.5 <= x[2] <= 0.5);
-julia> @objective(model, Max, x[1]);
+model = Model();
+@variable(model, x[1:2])
+set_lower_bound(x[2], -0.5)
+set_upper_bound(x[2], 0.5)
+@constraint(model, c1, x[1] + x[2] <= 1);
+@constraint(model, c2, x[1] - x[2] <= 1);
+@objective(model, Max, x[1]);
 ```
 
 ```@meta
 DocTestSetup = quote
     using JuMP
-    model = Model(() -> MOIU.MockOptimizer(MOIU.Model{Float64}(),
-                          eval_variable_constraint_dual=true));
+    mock = MOIU.MockOptimizer(
+        MOIU.Model{Float64}(), eval_variable_constraint_dual=false
+    )
+    model = direct_model(mock)
     @variable(model, x[1:2]);
+    set_lower_bound(x[2], -0.5)
+    set_upper_bound(x[2], 0.5)
     @constraint(model, c1, x[1] + x[2] <= 1);
     @constraint(model, c2, x[1] - x[2] <= 1);
-    @constraint(model, c3, -0.5 <= x[2] <= 0.5);
     @objective(model, Max, x[1]);
     optimize!(model);
-    mock = backend(model).optimizer.model;
-    MOI.set(mock, MOI.TerminationStatus(), MOI.OPTIMAL);
-    MOI.set(mock, MOI.DualStatus(), MOI.FEASIBLE_POINT);
-    MOI.set(mock, MOI.VariablePrimal(), JuMP.optimizer_index(x[1]), 1.0);
-    MOI.set(mock, MOI.VariablePrimal(), JuMP.optimizer_index(x[2]), 0.0);
-    MOI.set(mock, MOI.ConstraintBasisStatus(), JuMP.optimizer_index(c1), MOI.NONBASIC);
-    MOI.set(mock, MOI.ConstraintBasisStatus(), JuMP.optimizer_index(c2), MOI.NONBASIC);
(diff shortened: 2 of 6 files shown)


src/sd.jl
```diff
@@ -167,6 +202,25 @@ function build_variable(_error::Function, variables::Matrix{<:ScalarVariable}, :
...
+constrained after their creation.
+
+This function is used by the [`@variable`](@ref) macro as follows:
+```julia
+@variable(model, Q[1:2, 1:2] in SkewSymMatrixSpace())
+```
+"""
+function build_variable(_error::Function, variables::Matrix{<:ScalarVariable}, ::SkewSymMatrixSpace)
+    n = _square_side(_error, variables)
+    set = MOI.Reals((n^2-n)/2)
```
blegat: It should be `div(n^2 - n, 2)`, `n^2 - n` is even for any `n`.

src/sd.jl
```diff
@@ -96,6 +97,40 @@ function vectorize(matrix::Matrix, ::SymmetricMatrixShape)
...
+Shape object for a skew symmetric square matrix of `side_dimension` rows and columns.
+The vectorized form contains the entries of the upper-right triangular part of
+the matrix (without the diagonal) given column by column (or equivalently, the
+entries of the lower-left triangular part given row by row). The diagonal is zero.
+"""
+struct SkewSymmetricMatrixShape <: AbstractShape
+    side_dimension::Int
+end
+function reshape_vector(vectorized_form::Vector{T}, shape::SkewSymmetricMatrixShape) where T
+    matrix = Matrix{JuMP.GenericAffExpr{Float64,JuMP.VariableRef}}(undef, shape.side_dimension, shape.side_dimension)
```
blegat: It should be `Matrix{Base.promote_type(T, MA.promote_operation(-, T), MA.promote_operation(zero, T))}` so that it works for any `T`

src/sd.jl
```diff
@@ -96,6 +97,40 @@ function vectorize(matrix::Matrix, ::SymmetricMatrixShape)
...
+    k = 0
+    for j in 1:shape.side_dimension
+        for i in 1:j-1
+            k += 1
+            matrix[i, j] =  vectorized_form[k]
+            matrix[j, i] = -vectorized_form[k]
+        end
+    end
+    for j in 1:shape.side_dimension
+        matrix[j, j] =  0
```
blegat: Replace `0` by `zero(T)`

src/sd.jl
frederikgeth: ok!

src/sd.jl
frederikgeth: ok

src/sd.jl
frederikgeth: thanks!

src/sd.jl
```diff
@@ -96,6 +97,40 @@ function vectorize(matrix::Matrix, ::SymmetricMatrixShape)
...
+Shape object for a skew symmetric square matrix of `side_dimension` rows and columns.
+The vectorized form contains the entries of the upper-right triangular part of
+the matrix (without the diagonal) given column by column (or equivalently, the
+entries of the lower-left triangular part given row by row). The diagonal is zero.
+"""
+struct SkewSymmetricMatrixShape <: AbstractShape
+    side_dimension::Int
+end
+function reshape_vector(vectorized_form::Vector{T}, shape::SkewSymmetricMatrixShape) where T
+    matrix = Matrix{Base.promote_type(T, _MA.promote_operation(-, T), _MA.promote_operation(zero, T))}(undef, shape.side_dimension, shape.side_dimension)
```
blegat: This breaks 80-char limit

src/sd.jl
```diff
@@ -96,6 +97,40 @@ function vectorize(matrix::Matrix, ::SymmetricMatrixShape)
...
+    k = 0
+    for j in 1:shape.side_dimension
+        for i in 1:j-1
+            k += 1
+            matrix[i, j] =  vectorized_form[k]
+            matrix[j, i] = -vectorized_form[k]
+        end
+    end
+    for j in 1:shape.side_dimension
+        matrix[j, j] =  zero(T)
```
blegat: Remove space before `zero`

src/sd.jl
```diff
@@ -167,6 +202,25 @@ function build_variable(_error::Function, variables::Matrix{<:ScalarVariable}, :
     return VariablesConstrainedOnCreation(_vectorize_variables(_error, variables), set, shape)
 end
 
+"""
+    build_constraint(_error::Function, variables, ::SkewSymMatrixSpace)
+
+Return a `VariablesConstrainedOnCreation` of shape [`SkewSymMatrixSpace`](@ref)
```
blegat: `Space` ->`Shape`

test/variable.jl
```diff
@@ -467,6 +467,19 @@ function test_variable_symmetric(ModelType)
...
+    model = ModelType()
+
+    @variable model x[1:2, 1:2] in SkewSymMatrixSpace()
+    @test x[1, 2] == -x[2, 1]
+    @test x[1, 1] == 0
+    @test x[2, 2] == 0
+    @test model[:x] === x
+
+    y = @variable model [1:2, 1:2] in SkewSymMatrixSpace()
+    @test y[1, 2] == -y[2, 1]
```
blegat: I would add a 3 x 3 example as the 2 x 2 examples only has 1 non-diag entry so it may not cover all corner cases

frederikgeth:
Thanks, @blegat , I will update the pr and add the docs tomorrow

frederikgeth:
So tests fail for `test_variable_skewsymmetric({JuMPExtension.MyModel})` for 3x3 matrices. This is the error:
```
  DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 6 and 3")
  Stacktrace:
   [1] add_variable(::Main.JuMPExtension.MyModel, ::VariablesConstrainedOnCreation{MathOptInterface.Reals,SkewSymmetricMatrixShape,ScalarVariable{Float64,Float64,Float64,Float64}}, ::Array{String,2}) at ./broadcast.jl:490
```

Not sure where to go with this?

test/variable.jl
```diff
@@ -467,6 +467,30 @@ function test_variable_symmetric(ModelType)
...
 end
 
+function test_variable_skewsymmetric(ModelType)
+    model = ModelType()
+
+    @variable model x[1:2, 1:2] in SkewSymMatrixSpace()
+    @test x[1, 2] == -x[2, 1]
+    @test iszero(x[1, 1])
+    @test iszero(x[2, 2])
+    @test model[:x] === x
```
blegat: Add test for the number of variables actually added to the inner model

src/sd.jl
```diff
@@ -167,6 +203,25 @@ function build_variable(_error::Function, variables::Matrix{<:ScalarVariable}, :
...
+This function is used by the [`@variable`](@ref) macro as follows:
+```julia
+@variable(model, Q[1:2, 1:2] in SkewSymMatrixSpace())
+```
+"""
+function build_variable(_error::Function, variables::Matrix{<:ScalarVariable}, ::SkewSymMatrixSpace)
+    n = _square_side(_error, variables)
+    set = MOI.Reals(div(n^2 - n, 2))
+    shape = SkewSymmetricMatrixShape(n)
+    return VariablesConstrainedOnCreation(_vectorize_variables(_error, variables), set, shape)
```
blegat: `vectorize_variables` also include the diagonal, this why the test are failing. I would guess because of this, you add to many variables so test the number of variables added

test/variable.jl
frederikgeth: I'm not sure how to access those?

src/sd.jl
frederikgeth: ok

src/sd.jl
frederikgeth: ok

src/sd.jl
```diff
@@ -167,6 +205,25 @@ function build_variable(_error::Function, variables::Matrix{<:ScalarVariable}, :
...
+This function is used by the [`@variable`](@ref) macro as follows:
+```julia
+@variable(model, Q[1:2, 1:2] in SkewSymMatrixSpace())
+```
+"""
+function build_variable(_error::Function, variables::Matrix{<:ScalarVariable}, ::SkewSymMatrixSpace)
+    n = _square_side(_error, variables)
+    set = MOI.Reals(div(n^2 - n, 2))
+    shape = SkewSymmetricMatrixShape(n)
+    return VariablesConstrainedOnCreation(vectorize(variables, SkewSymmetricMatrixShape(n)), set, shape)
```
blegat: Here, you will ignore `variables` outside the upper triangle, add a TODO for throwing error in case the user set something that does not suit the skew symmetry like we do in `_vectorize_variables`

test/variable.jl
blegat: https://www.juliaopt.org/JuMP.jl/dev/variables/#JuMP.num_variables

martincornejo:
Any updates on this PR? This feature could be convenient for me

odow:
Moved to https://github.com/jump-dev/JuMP.jl/pull/2416

PR #2150, opened 2020-02-03
set_parameter -> set_optimizer_attribute

### LINKED ISSUE #2149: Rename `set_parameter` to `set_attributes`
https://github.com/JuliaOpt/JuMP.jl/pull/2141 adds `optimizer_with_attributes`, so one can go:
```Julia
model = Model(optimizer_with_attributes(Gurobi.Optimizer, "OutputFlag" => 0))
```
However, we still have `set_parameter`, which means
```Julia
set_parameter(model, "OutputFlag", 0)
```

It would be more consistent if we renamed `set_parameter` to `set_attribute`, with the implication that passing a string was sugar for passing `MOI.RawParameter`.

In addition, we should add `get_attribute`.

odow:
Closes #2149


src/JuMP.jl
```diff
@@ -424,36 +424,79 @@ function solve_time(model::Model)
...
+function set_optimizer_attribute(model::Model, name::String, value)
+    return set_optimizer_attribute(model, MOI.RawParameter(name), value)
+end
+
+"""
+    set_optimizer_attribute(
+        model::Model, attr::MOI.AbstractOptimizerAttribute, value
+    )
+
+Set the solver-specific attribute `attr` in `model` to `value`.
```
mlubin: We could give an example that `set_silent` is equivalent to `set_optimizer_attribute` with `MOI.Silent()`.

blegat:
Looking forward, we would have `set_optimizer_attribute`, `set_model_attribute`, `set_variable_attribute` and `set_constraint_attribute`.
I may be redundant that we specify both in the function name and in the signature the different of attribute type.
Why not have a single `set_attribute` function with 4 different methods ?

odow:
> Why not have a single set_attribute function with 4 different methods ?

People can always just do `MOI.set(model, attr, index, value)`, which is actually less characters. 

The reason for making `optimizer_attribute` special is so that it lines up with `optimizer_with_attributes` and because passing a string is fundamentally different.

For non-optimizer attributes, we either have things like `set_start_value`, or, for the more obscure ones, people should use `MOI.set` and know what they are doing.

odow:
Not sure what went wrong with coverage, because all of the lines I added got hit.

src/JuMP.jl
```diff
@@ -54,54 +54,54 @@ const _MOIVAR = MOI.VariableIndex
...
 
 ## Examples
 
 The call `optimizer_with_attributes(Gurobi.Optimizer, "Presolve" => 0, "OutputFlag" => 0)`
-groups a Gurobi optimizer with the parameters `Presolve` with value `0` and
+groups a Gurobi optimizer with the attributes `Presolve` with value `0` and
 `OutputFlag` with value `0`. When provided to the `Model` constructor or to
 [`set_optimizer`](@ref), it creates a Gurobi optimizer and then set the
-`Presolve` parameter to value `0` and the `OutputFlag` parameter to value `0`:
+`Presolve` attribute to value `0` and the `OutputFlag` attribute to value `0`:
```
mlubin: Note that any string attributes used here are specific to the solver. One should consult the solver's documentation to find the parameters of interest.

PR #2175, opened 2020-02-22
Remove a comment that is no more relevant.

dourouc05:


src/macros.jl
```diff
@@ -161,11 +161,6 @@ _functionize(v::AbstractArray{VariableRef}) = _functionize.(v)
 _functionize(x) = x
 function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Val, lhs, rhs)
     # Simple comparison - move everything to the LHS.
-    #
-    # Note: We add the +0 to this term to account for the pathological case that
```
blegat: It is still relevant but it is `_functionize`, not `+0` anymore

dourouc05: Is the new formulation correct?

PR #2177, opened 2020-02-24
Add @variable to (model, x) in anon var doc

mlhetland:
This expression seems to be missing the macro?


mlubin:
@mlhetland Thanks!
@blegat Did a change in MutableArithmetics break JuMP's tests?

odow:
I kicked off Travis again and it passed. Couldn't figure out how to restart AppVeyor though? Anyway, merged since this is just documentation.

PR #2196, opened 2020-03-07
Implementation of Jac-vec products

amontoison:
+ Non optimized Jac-vec products.

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -618,37 +618,35 @@ function MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
 end
 
 function MOI.eval_constraint_jacobian_product(d::NLPEvaluator,
-                                              Jv::AbstractVector{Float64},
+                                              y::AbstractVector{Float64},
                                               x::AbstractVector{Float64},
-                                              v::AbstractVector{Float64})
-    fill!(Jv, 0.0)
+                                              w::AbstractVector{Float64})
+    fill!(y, 0.0)
     jac_struct = MOI.jacobian_structure(d)
-    nnzj = length(jac_struct)
-    J = zeros(Float64, nnzj)
-    MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
-    for k = 1 : nnzj
-        i = jac_struct[k][1]
-        j = jac_struct[k][2]
-        Jv[i] += J[k] * v[j]
+    nnz_jac = length(jac_struct)
+    J = zeros(Float64, nnz_jac)
+    MOI.eval_constraint_jacobian(d, J, x)
+    for k in 1:nnz_jac
+        i, j = jac_struct[k]
+        y[i] += J[k] * w[j]
     end
-    return Jv
+    return y
 end
 
 function MOI.eval_constraint_jacobian_transpose_product(d::NLPEvaluator,
-                                                        Jtv::AbstractVector{Float64},
+                                                        y::AbstractVector{Float64},
                                                         x::AbstractVector{Float64},
-                                                        v::AbstractVector{Float64})
-    fill!(Jtv, 0.0)
+                                                        w::AbstractVector{Float64})
+    fill!(y, 0.0)
     jac_struct = MOI.jacobian_structure(d)
-    nnzj = length(jac_struct)
-    J = zeros(Float64, nnzj)
+    nnz_jac = length(jac_struct)
+    J = zeros(Float64, nnz_jac)
     MOI.eval_constraint_jacobian(d, J, x)
-    for k = 1 : nnzj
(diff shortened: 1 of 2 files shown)


src/nlp.jl
```diff
@@ -619,7 +617,39 @@ function MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
     return
 end
 
+function MOI.eval_constraint_jacobian_product(d::NLPEvaluator,
+                                              Jv::AbstractVector{Float64},
```
mlubin, line 621: If you're going to use one-letter variable names, then it's better to use the ones that are consistent with the MOI documentation: http://www.juliaopt.org/MathOptInterface.jl/stable/apireference/#MathOptInterface.eval_constraint_jacobian_product
i.e., `eval_constraint_jacobian_product(d, y, x, w)`. Same comment for below.

src/nlp.jl
```diff
@@ -619,7 +617,39 @@ function MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
...
+                                              x::AbstractVector{Float64},
+                                              v::AbstractVector{Float64})
+    fill!(Jv, 0.0)
+    jac_struct = MOI.jacobian_structure(d)
+    nnzj = length(jac_struct)
```
mlubin, line 626: Use underscores to separate words in names: http://www.juliaopt.org/JuMP.jl/dev/style/#Use-of-underscores-within-names-1. I'd suggest naming this `nnz_jac`.

src/nlp.jl
```diff
@@ -619,7 +617,39 @@ function MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
...
+    J = zeros(Float64, nnzj)
+    MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
```
mlubin, line 628: The `::NLPEvaluator` has no effect and should be removed.

test/nlp.jl
```diff
@@ -328,6 +337,67 @@
...
+        @NLconstraint(m, a^2/2 <= 1)
+        d = JuMP.NLPEvaluator(m)
+        MOI.initialize(d, [:JacVec, :Jac])
+
+        values = [1.0, 2.0, 3.0] # For a, b, c.
+        jacobian_sparsity = MOI.jacobian_structure(d)
+        V = zeros(length(jacobian_sparsity))
+        MOI.eval_constraint_jacobian(d, V, values)
+        correct_jacobian = [0.0 3.0 2.0; 1.0 0.0 0.0]
+        @test isapprox(dense_jacobian(jacobian_sparsity, V, 2, 3), correct_jacobian)
```
mlubin, line 357: Please use ≈ instead of isapprox (see the TODO at the top of the file).

test/nlp.jl
```diff
@@ -328,6 +337,67 @@
...
+
+        v = [2.4, 3.5, 1.2]
+        Jv = zeros(2)
```
mlubin, line 360: Instead of `Jv`, what about a more readable name like `product_storage`?

test/nlp.jl
```diff
@@ -328,6 +337,67 @@
...
+        MOI.eval_constraint_jacobian_product(d, Jv, values, v)
+        @test isapprox(Jv, correct_jacobian * v)
+
+        w = [0.6, 4.3]
+        Jᵀw = zeros(3)
```
mlubin, line 365: The ᵀ symbol could potentially confuse a reader. It's not obvious that it's part of the name and not an arithmetic operator. I would just re-use the name `product_storage`.

amontoison:
Thanks for the comments Miles !
I updated the pull request.

Do I need to dig a lot in the code to understand how to use forward / reverse pass for Jv and J'v products? Optimized JacVec products will be relevant for [NLPModelsJuMP](https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl). We recently updated it for MathOptInterface with Dominique and Abel.

mlubin:
> Do I need to dig a lot in the code to understand how to use forward / reverse pass for Jv and J'v products?

In short, yes, you'll have to dig a lot because the code is undocumented and was written in 2016 before JuMP had a style guide. But it's feasible to do. I would start with Jv. Make sure you understand the math first (i.e., using forward-mode AD to compute a directional derivative of the constraints -- there are a number of blog posts on this). Then in `nlp.jl` you basically need to call `_forward_eval_all` and then `forward_eval_ϵ` where the `input_ϵ` array is filled in with the vector you're taking a product with. No reverse pass is needed.

amontoison:
Following your recommendations, I implemented the optimized version for Jv products in #2239. From what I understand, It's only the reverse pass that we require for J'v products.

PR #2198, opened 2020-03-10
Added Base.empty! method for JuMP.Model.

henriquebecker91:
Not only the method but also its documentation and tests. The documentation is generating and tests passing.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -396,6 +396,30 @@ function add_bridge(model::Model,
 end
 
 """
+    empty!(model::Model) -> model
+
+Empty the model, that is, remove all variables, constraints and model
+attributes but not optimizer attributes. Always return the argument.
+
+Note: removes extensions data.
+"""
+function Base.empty!(model::Model)::Model
+    # The method changes the Model object to, basically, the state it was when
+    # created (if the optimizer was already pre-configured). The two exceptions
+    # are: optimize_hook and operator_counter. One is left alone because it is
+    # related to optimizer attributes and the other is just a counter for a
+    # single-time warning message (so keeping it helps to discover
+    # inneficiencies).
+    MOI.empty!(model.moi_backend)
+    empty!(model.shapes)
+    empty!(model.bridge_types)
+    model.nlp_data = nothing
+    empty!(model.obj_dict)
+    empty!(model.ext)
+    return model
+end
+
+"""
     num_variables(model::Model)::Int64
 
 Returns number of variables in `model`.
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -405,14 +405,16 @@ Note: removes extensions data.
 """
 function Base.empty!(model::Model)::Model
     # The method changes the Model object to, basically, the state it was when
-    # created (if the optimizer was already pre-configured). The two exceptions
-    # are: optimize_hook and operator_counter. One is left alone because it is
-    # related to optimizer attributes and the other is just a counter for a
-    # single-time warning message (so keeping it helps to discover
-    # inneficiencies).
+    # created (if the optimizer was already pre-configured). The exceptions
+    # are:
+    # * optimize_hook: it is basically an optimizer attribute and we promise
+    #   to leave them alone (as do MOI.empty!).
+    # * bridge_types: for consistency with MOI.empty! for
+    #   MOI.Bridges.LazyBridgeOptimizer.
+    # * operator_counter: it is just a counter for a single-time warning
+    #   message (so keeping it helps to discover inneficiencies).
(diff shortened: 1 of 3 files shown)


henriquebecker91:
This was discussed with @odow over Discourse (https://discourse.julialang.org/t/empty-for-jump-model-or-jump-direct-model/35673).

src/JuMP.jl
```diff
@@ -395,6 +395,30 @@ function add_bridge(model::Model,
...
+function Base.empty!(model::Model)::Model
+    # The method changes the Model object to, basically, the state it was when
+    # created (if the optimizer was already pre-configured). The two exceptions
+    # are: optimize_hook and operator_counter. One is left alone because it is
+    # related to optimizer attributes and the other is just a counter for a
+    # single-time warning message (so keeping it helps to discover
+    # inneficiencies).
+    MOI.empty!(model.moi_backend)
+    empty!(model.shapes)
+    empty!(model.bridge_types)
```
blegat: I would keep `brodge_types` unaltered to be consistent witn `MOI.empty!` for `MOI.Bridges.LazyBridgeOptimizer`

henriquebecker91: Done. Commit message is missing a `model.bridge_types` after the comma because bash interpreted it as a subshell invocation (sorry) and I did not perceive until after push.

henriquebecker91: I tested with just the tests in `model.jl`, those passed.

mlubin:
For consistency with MOI, I don't think we should extend `Base.empty!`. `Base.empty!` is defined as:
"Remove all elements from a collection.". A JuMP model is not a collection (it doesn't implement any sort of collection-like interface), so I don't think this applies. We could define `JuMP.empty!` instead.

henriquebecker91:
I will change this (I do not know if today, because I will be traveling
without internet connection most of the day), but note that doing this has
some consequences: all occurrences of `empty!` inside JuMP must be changed
to `Base.empty!` (the same applies to future occurrences referring to
`Base.empty!`); if `JuMP.empty!` is exported (what makes sense), it will
make all code with `using JuMP` and `empty!` will error about lack of
qualification. MOI is often not expected to be imported with "using" and/or
directly by the user (most users will use the `const MOI = ...` trick to
just copy and paste code freely from other places), but JuMP is often
imported with "using" and is probably used in many codes that make use of
`empty!`.

Em qui., 12 de mar. de 2020 às 09:58, Miles Lubin <notifications@github.com>
escreveu:

odow:
I agree with @henriquebecker91. Exporting an identical symbol to base Julia is very bad.

We either overload `Base.empty!`, or we provide `JuMP.empty_model!`. I vote for `Base.empty!`, even though it isn't exactly the same semantics.

mlubin:
Ah right, I wasn't thinking about our auto-export. Not much we can do about Julia's weird global namespace. We can go with `Base.empty!` but I'm not too satisfied with it.

henriquebecker91:
I understand your annoyance with this choice. I would like to point that:

* There is a `Base.isempty` and a `Base.length` for String but not a
`Base.empty!` (as it is immutable). Even if it is just three methods to
implement a [`General Collection Interface`](
https://docs.julialang.org/en/v1/base/collections/#General-Collections-1),
some Types of Base do not implement it fully (but use the same qualified
name), the "Fully implemented by:" section after the documentation of these
three functions implicitly acknowledge this.
* The JuMP.Model type is not from Base, so it is not type piracy at least.
* Another name may be used, like `reset_model!`, `reset!`, `empty_model!`,
but the average user of JuMP will probably guess `empty!`, not qualify it
and be right about their guess, rarely being negatively affected by this
choice in any fashion. The intuitiveness of the name is attested by the
fact that MOI use the same name (just does not have the same export policy).
* As you perceived, a middle ground is not possible with this name, as
either: we export `JuMP.empty!` and break a considerable amount of code in
a way that will be hard for non-CS users to exactly understand what is
happening; or we add a `JuMP.empty!` but not export it, but for so we need
to break the "export all names non-prefixed-with-underscore" rule.

Em qui., 12 de mar. de 2020 às 13:32, Miles Lubin <notifications@github.com>
escreveu:

henriquebecker91:
I will leave to you two to decide if the name will be changed and, if this is the case, to which name. The current code seems to only need changes if the function name is changed.

mlubin:
The consensus is to not change the name.

odow:
Thanks for your help @henriquebecker91. I appreciate it.

henriquebecker91:
Thanks for being open to my rants, @odow.

Em sex., 13 de mar. de 2020 às 14:21, Oscar Dowson <notifications@github.com>
escreveu:

Issue #2158, opened 2020-02-09
Add a nice way to access MOI attributes


odow:
Some MOI attributes have nice accessors, e.g.:
https://github.com/JuliaOpt/JuMP.jl/blob/c7b5d930037db4f304cf061fddc233d8974dcdd8/src/objective.jl#L18-L20
However, others such as `RelativeGap` are missing.

TODO: Add accessors for missing attributes.
- [x] RelativeGap https://github.com/JuliaOpt/JuMP.jl/pull/2199
- [ ] SimplexIterations
- [ ] BarrierIterations
- [ ] NodeCount

x-ref: https://github.com/JuliaOpt/JuMP.jl/pull/2156#discussion_r376813037

PR #2199, opened 2020-03-11
#2158 Added relative_gap() to access MOI.RelativeGap() attribute

vrngupta53:
Added getter function for MOI.RelativeGap() attribute with documentation

### DIFF
--- a/docs/src/solutions.md
+++ b/docs/src/solutions.md
@@ -141,8 +141,8 @@ using JuMP
 model = Model()
 # ...
 optimize!(model)
-JuMP.relative_gap(model)
-MOI.get(model, MOI.SimplexIterations())
+@show relative_gap(model)
+@show MOI.get(model, MOI.SimplexIterations())
 ```
 
 ## Sensitivity analysis for LP
--- a/docs/src/solutions.md
+++ b/docs/src/solutions.md
@@ -133,7 +133,7 @@ end
 [MathOptInterface](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/) defines a large
 number of model attributes that can be queried. Examples include
 [`MOI.RelativeGap`](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/apireference/#MathOptInterface.RelativeGap) and
-[`MOI.SimplexIterations`](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/apireference/#MathOptInterface.SimplexIterations).
+[`MOI.SimplexIterations`](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/apireference/#MathOptInterface.SimplexIterations). Some attributes can be directly accessed by getter functions like `JuMP.objective_bound()` and `JuMP.relative_gap()`.
 
 To query these attributes, use:
 ```julia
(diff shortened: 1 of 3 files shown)


odow:
Great! You have a couple of other things to do.

1. We use automated tests to reduce the chance of bugs in the code.

Take a look at the file https://github.com/JuliaOpt/JuMP.jl/blob/master/test/generate_and_solve.jl.

https://github.com/JuliaOpt/JuMP.jl/blob/6f9192a07cffde8fb071855d540341a49ca91ecd/test/generate_and_solve.jl#L50-L54
i.e., you want something like
```Julia
MOI.set(mockoptimizer, MOI.RelativeGap(), 0.123)
```

Then, you can add a test that you can query the attribute using your function. Take a look at:
https://github.com/JuliaOpt/JuMP.jl/blob/6f9192a07cffde8fb071855d540341a49ca91ecd/test/generate_and_solve.jl#L74
i.e., you want something like
```Julia
@test  0.123 == @inferred JuMP.relative_gap(m)
```

2. Documentation

Edit this section of `solutions.md` to explain the function you added:
https://github.com/JuliaOpt/JuMP.jl/blob/master/docs/src/solutions.md#accessing-mathoptinterface-attributes

vrngupta53:
I have written the testcase only for the Integer Program case because relative_gap() only applies for Mixed Integer Programs

docs/src/solutions.md
```diff
@@ -133,15 +133,16 @@ end
 [MathOptInterface](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/) defines a large
 number of model attributes that can be queried. Examples include
 [`MOI.RelativeGap`](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/apireference/#MathOptInterface.RelativeGap) and
-[`MOI.SimplexIterations`](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/apireference/#MathOptInterface.SimplexIterations).
+[`MOI.SimplexIterations`](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/apireference/#MathOptInterface.SimplexIterations). Some attributes can be directly accessed by getter functions like `JuMP.objective_bound()` and `JuMP.relative_gap()`.
```
odow: Instead of "`JuMP.realative_gap()`", use "[`relative_gap`](@ref)". This will link to the documentation. Same for `objective_bound`.

vrngupta53: done

docs/src/solutions.md
```diff
@@ -133,15 +133,16 @@ end
...
+[`MOI.SimplexIterations`](https://www.juliaopt.org/MathOptInterface.jl/v0.9.10/apireference/#MathOptInterface.SimplexIterations). Some attributes can be directly accessed by getter functions like [`objective_bound`](@ref) and [`relative_gap`](@ref).
 
 To query these attributes, use:
 ```julia
 using JuMP
 model = Model()
 # ...
 optimize!(model)
-MOI.get(model, MOI.RelativeGap())
+JuMP.relative_gap(model)
```
blegat: Why the `JuMP.` prefix?

PR #2201, opened 2020-03-12
#2158 Added functions simplex_iterations(), barrier_iterations() and node_count()

### LINKED ISSUE #2158: Add a nice way to access MOI attributes

vrngupta53:
Test cases added as well.

Closes #2158

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -635,6 +635,37 @@ function time_limit_sec(model::Model)
     return MOI.get(model, MOI.TimeLimitSec())
 end
 
+"""
+    simplex_iterations(model::Model)
+
+Gets the cumulative number of simplex iterations during the optimization process after a call to `optimize!(model)`.
+Returns the total simplex iterations for all nodes in case of a mixed-integer program (MIP). Solver must implement
+MathOptInterface.SimplexIterations() to use this function.
+"""
+function simplex_iterations(model::Model)::Int64
+    return MOI.get(model, MOI.SimplexIterations())
+end
+
+"""
+    barrier_iterations(model::Model)
+
+Gets the cumulative number of barrier iterations while solving a problem after a call to `optimize!(model)`.Solver must
+implement MathOptInterface.BarrierIterations() to use this function.
+"""
+function barrier_iterations(model::Model)::Int64
+    return MOI.get(model, MOI.BarrierIterations())
+end
+
+"""
+    node_count(model::Model)
+
+Gets the total number of branch-and-bound nodes explored while solving a mixed-integer program (MIP) after a call
+to `optimize!(model)`.Solver must implement MathOptInterface.NodeCount() to use this function.
+"""
+function node_count(model::Model)::Int64
+    return MOI.get(model, MOI.NodeCount())
+end
+
 # Abstract base type for all scalar types
 # The subtyping of `AbstractMutable` will allow calls of some `Base` functions
 # to be redirected to a method in MA that handles type promotion more carefuly
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -638,31 +638,31 @@ end
 """
     simplex_iterations(model::Model)
 
-Gets the cumulative number of simplex iterations during the optimization process after a call to `optimize!(model)`.
-Returns the total simplex iterations for all nodes in case of a mixed-integer program (MIP). Solver must implement
-MathOptInterface.SimplexIterations() to use this function.
+Gets the cumulative number of simplex iterations during the most-recent optimization.
+Solver must implement `MOI.SimplexIterations()` to use this function.
 """
-function simplex_iterations(model::Model)::Int64
+function simplex_iterations(model::Model)
     return MOI.get(model, MOI.SimplexIterations())
 end
 
 """
     barrier_iterations(model::Model)
 
-Gets the cumulative number of barrier iterations while solving a problem after a call to `optimize!(model)`.Solver must
-implement MathOptInterface.BarrierIterations() to use this function.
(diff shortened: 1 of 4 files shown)


vrngupta53:
I will add the documentation as well if these changes are fine.

src/JuMP.jl
```diff
@@ -635,6 +635,37 @@ function time_limit_sec(model::Model)
...
 end
 
+"""
+    simplex_iterations(model::Model)
+
+Gets the cumulative number of simplex iterations during the optimization process after a call to `optimize!(model)`.
+Returns the total simplex iterations for all nodes in case of a mixed-integer program (MIP). Solver must implement
+MathOptInterface.SimplexIterations() to use this function.
+"""
+function simplex_iterations(model::Model)::Int64
```
odow: Two things:
- Instead of using `Int64`, you should generally use `Int`.
- Gurobi actually returns a `Float64` here to avoid overflow:
https://github.com/JuliaOpt/Gurobi.jl/blob/5822a697c02884f3310a0fc3f6a0e6d4a5942df1/src/MOI_wrapper.jl#L2089
 so it's not clear we want to have a return-type annotation.

I would remove.

src/JuMP.jl
```diff
@@ -635,6 +635,37 @@ function time_limit_sec(model::Model)
     return MOI.get(model, MOI.TimeLimitSec())
 end
 
+"""
+    simplex_iterations(model::Model)
+
+Gets the cumulative number of simplex iterations during the optimization process after a call to `optimize!(model)`.
+Returns the total simplex iterations for all nodes in case of a mixed-integer program (MIP). Solver must implement
+MathOptInterface.SimplexIterations() to use this function.
```
odow: Maybe just:
```
Gets the cumulative number of simplex iterations during the most-recent 
optimization.

Solvers must implement `MOI.SimplexIterations()` to use this function.
```
Use back-ticks around all code, e.g., ```` `MathOptInterface.SimplexIterations()` ````.

src/JuMP.jl
```diff
@@ -635,6 +635,37 @@ function time_limit_sec(model::Model)
...
+"""
+function simplex_iterations(model::Model)::Int64
+    return MOI.get(model, MOI.SimplexIterations())
+end
+
+"""
+    barrier_iterations(model::Model)
+
+Gets the cumulative number of barrier iterations while solving a problem after a call to `optimize!(model)`.Solver must
+implement MathOptInterface.BarrierIterations() to use this function.
```
odow: See my suggested change for `SimplexIterations`.

src/JuMP.jl
```diff
@@ -635,6 +635,37 @@ function time_limit_sec(model::Model)
...
+"""
+function barrier_iterations(model::Model)::Int64
```
odow: Similarly, remove `::Int64`.

src/JuMP.jl
```diff
@@ -635,6 +635,37 @@ function time_limit_sec(model::Model)
...
+    return MOI.get(model, MOI.BarrierIterations())
+end
+
+"""
+    node_count(model::Model)
+
+Gets the total number of branch-and-bound nodes explored while solving a mixed-integer program (MIP) after a call
+to `optimize!(model)`.Solver must implement MathOptInterface.NodeCount() to use this function.
+"""
+function node_count(model::Model)::Int64
```
odow: Ditto doc-string and return type.

test/generate_and_solve.jl
```diff
@@ -78,6 +82,9 @@ using JuMP
         @test -1.0 == @inferred JuMP.dual(c)
         @test  0.0 == @inferred JuMP.dual(JuMP.UpperBoundRef(x))
         @test  1.0 == @inferred JuMP.dual(JuMP.LowerBoundRef(y))
+        @test  Int64(1) == @inferred JuMP.simplex_iterations(m)
+        @test  Int64(1) == @inferred JuMP.barrier_iterations(m)
+        @test  Int64(1) == @inferred JuMP.node_count(m)
```
odow: Others may chime in, but not sure if we need to `@inferred` these.

blegat:
Should we use the `num_` prefix for consistency with `num_variables`, ...?

src/JuMP.jl
vrngupta53: I agree with no return type. I actually had not specified a return type but then did so later, assuming it as good practice.

odow:
Actually. Can you add some documentation to `/docs` like you did for relative gap?

vrngupta53:
Sure, done

vrngupta53:
Unable to understand why the buid failed.

odow:
I meant that we want to add `simplex_iterations`, `barrier_iterations`, and `node_count` to the documentation, rather than `relative_gap`.

The error is because the other PR isn't merged yet. Once the CI finishes, I will merge: https://github.com/JuliaOpt/JuMP.jl/pull/2199.

Then, you can rebase your PR (let me know if you need help with that).

odow:
Thanks @vrngupta53

Issue #2142, opened 2020-01-30
Print bridge graph


blegat:
Add a `print_bridge_graph` functions that either
* throws an error in case there is no bridge layer used
* calls `MOI.Bridges.print_graph` on the bridge layer otherwise.

To implement this, inspiration can be taken from
https://github.com/JuliaOpt/JuMP.jl/blob/dd68f8107339b15b41be147d6454202530606a3f/src/JuMP.jl#L309-L345

This is a good first step for the "Visualization of bridge graph" GSoC 2020 project.

See https://github.com/JuliaOpt/MathOptInterface.jl/issues/680

PR #2202, opened 2020-03-13
Adding print_bridge_graph

yashcodes:
Closes [#2142](https://github.com/jump-dev/JuMP.jl/issues/2142)

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -398,10 +398,27 @@ function add_bridge(model::Model,
 end
 
 """
-     print_bridge_graph(model::Model)
+     print_bridge_graph([io::IO,] model::Model)
 
-Print out the bridge graph explaining which bridges are used (and in what order
-thay are used) to bridge relevant constraints.
+Print the hyper-graph containing all variable, constraint, and objective types 
+that could be obtained by bridging the variables, constraints, and objectives 
+that are present in the model.
+
+Each node in the hyper-graph corresponds to a variable, constraint, or objective 
+type.
+  * Variable nodes are indicated by `[ ]`
+  * Constraint nodes are indicated by `( )`
+  * Objective nodes are indicated by `| |`
+The number inside each pair of brackets is an index of the node in the 
+hyper-graph.
+
+Note that this hyper-graph is the full list of possible transformations. When 
+the bridged model is created, we select the shortest hyper-path(s) from this 
+graph, so many nodes may be un-used.
+
+For more information, see Legat, B., Dowson, O., Garcia, J., and Lubin, M. 
+(2020).  "MathOptInterface: a data structure for mathematical optimization 
+problems." URL: https://arxiv.org/abs/2002.03447
 """
 print_bridge_graph(model::Model) = print_bridge_graph(Base.stdout, model)
 
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -379,6 +379,7 @@ function _moi_add_bridge(caching_opt::MOIU.CachingOptimizer,
     return
 end
 
+
 """
      add_bridge(model::Model,
                 BridgeType::Type{<:MOI.Bridges.AbstractBridge})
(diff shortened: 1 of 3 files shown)


yashcodes:
@blegat Please check, does it look ok?

yashcodes:
@odow I think it is good to go forward now.

src/JuMP.jl
```diff
@@ -395,6 +396,36 @@ function add_bridge(model::Model,
     return
 end
 
+function _moi_add_bridge(model::Nothing,
+                        BridgeType::Type{<:MOI.Bridges.AbstractBridge})
+    # No optimizer is attached, the bridge graph will be printed if one is attached
+    return
+end
```
odow: What is this function for?

src/JuMP.jl
```diff
@@ -378,6 +378,7 @@ function _moi_add_bridge(caching_opt::MOIU.CachingOptimizer,
     return
 end
 
+
```
odow: Remove new line.

src/JuMP.jl
```diff
@@ -395,6 +396,36 @@ function add_bridge(model::Model,
...
+end
+function _moi_print_bridge_graph(io::IO, caching_opt::MOIU.CachingOptimizer)
+    _moi_print_bridge_graph(io, caching_opt.optimizer)
+    return
+end
+
+"""
+     print_bridge_graph(model::Model)
+
+Print out the bridge graph explaining which bridges and in what order they are used to bridge constrained function or variable
```
odow: Wrap to 80 characters and end with a full stop.

Maybe just:
```
Print a summary of the bridges added to the model, explaining which constraints 
they bridge and in what order.
```

src/JuMP.jl
yashcodes: Honestly speaking I am also not very sure abt that, I added it because the same kind of the function is also there for adding bridges (maybe it is there to not get an error if we try to add bridge in empty Model)  
`function _moi_add_bridge(model::Nothing,
                        BridgeType::Type{<:MOI.Bridges.AbstractBridge})
    # No optimizer is attached, the bridge will be added when one is attached
    return
end`

src/JuMP.jl
yashcodes: I think I made a mistake the correct statement should be
`Print graph created by added bridges in `model::Model.``

yashcodes: I think the correct statement should be
`Print a summary of the bridges added to the model, explaining which constraint they bridge.`

src/JuMP.jl
odow: Please remove.

src/JuMP.jl
odow: It should be the plural `constraints`.

src/JuMP.jl
yashcodes: cool.

yashcodes:
@odow Hope its good to go now.

src/JuMP.jl
```diff
@@ -396,6 +397,37 @@ function add_bridge(model::Model,
     return
 end
 
+"""
+     print_bridge_graph(model::Model)
+
+Print out the bridge graph explaining which bridges are used (and in what order
+thay are used) to bridge relevant constraints.
```
blegat, line 404: thay -> they

blegat: Maybe add a note that this is the full bridge graph containing all the constraint, variable and objective types that could be obtained by bridging the variable, constraint and objective that are present in the model but this does not mean that all these bridges  and nodes are used. This is all the nodes of the full graph but then we select the shortest path and we only use this one so many nodes may be unused.

odow: Is there a way to print the optimal hyper-paths? That's perhaps more useful than the complete graph.

blegat: Not yet, That would be a nice addition indeed. Other useful printing would be one showing alternative paths with the same or lower cost than the one chosen. That's the second item of https://github.com/jump-dev/MathOptInterface.jl/issues/680

blegat (approved):

src/JuMP.jl
yashcodes: Please checkout the PR [jump-dev/MathOptInterface.jl#1055](https://github.com/jump-dev/MathOptInterface.jl/pull/1055), I think it is  solving the second item of [jump-dev/MathOptInterface.jl#680](https://github.com/jump-dev/MathOptInterface.jl/issues/680).

Issue #2191, opened 2020-03-05
Feature Request: Report table for query solutions


yordiak:
The last few weeks I have to switched to JuMP in order to perform my (optimization) analysis. Having a look at the documentation related to "Query Solutions" I found that you can extract very useful information regarding the optimizations results.

Thus, I was thinking that it would be a nice all of this information to be presented in a table. I understand that not everything is useful for everyone, and querying solutions might be context-specific. However, I believe that there are parts of a solution that are important to (almost) every one. These parts should be included in the report table.

I believe that a `report!()` command can generate such table. Here, arguments will denote the information someone wants to include in the table. For instance,

`report!()` = universal results, like `objective_value` and `value.(x)`
`report!(primal, dual)` = `objective_value` and `value.(x)` and also, `primal_status` and `dual_status'

odow:
This issue is still up for grabs. I think we want a function
```Julia
function solution_summary(model::Model; verbose::Bool = false)
    # ...
end
```
that produces a report like the following:
```
solution_summary(model)

--------------------------------------------------------------------------------
* Solver : MOI.get(model, MOI.Name())

* Status

  termination_status : MOI.OPTIMAL
  primal_status      : MOI.FEASIBLE_POINT
  dual_status        : MOI.NO_SOLUTION
  Message from the solver:
    "... the output of MOI.get(model, MOI.RawTerminationStatus()) ..."

* Candidate solution

  Objective value : xxx
  Objective bound : xxx

* Work counters

  Solve time (sec)   : xxx
  Simplex iterations : xxx
  Barrier iterations : xxx
  Node count         : xxx
--------------------------------------------------------------------------------

solution_summary(model; verbose = true)

--------------------------------------------------------------------------------
* Solver : MOI.get(model, MOI.Name())

* Status

  termination_status : MOI.OPTIMAL
  primal_status      : MOI.FEASIBLE_POINT
  dual_status        : MOI.NO_SOLUTION
  Message from the solver:
    "... the output of MOI.get(model, MOI.RawTerminationStatus()) ..."

* Candidate solution

  Objective value : xxx
  Objective bound : xxx
  Primal solution : 
     x[1] : xxx
     x[2] : xxx
     ...
  Dual solution : 
     c[1] : xxx
     c[2] : xxx
     ...

* Work counters

  Solve time (sec)   : xxx
  Simplex iterations : xxx
  Barrier iterations : xxx
  Node count         : xxx
--------------------------------------------------------------------------------
```

PR #2204, opened 2020-03-14
Report table to query solution

yashcodes:
Solves [#2191](https://github.com/JuliaOpt/JuMP.jl/issues/2191)
This PR adds a function `result!(model)` whose purpose is to structure the result obtained from the solver in a proper Tabular form, this function still have lots of possibilities to extend that's why we can call it as a basic version of `result!` like the accounting of multiple numbers of results is still pending, accessing MOI attribute likes RelativeGap(), SimplexIterations() etc.
From the current version of `result!` we can access dual solutions by setting keyword argument `dual_query` equals to `true` (`result!(model, dual_query = true)` )


yashcodes:
@odow @mtanneau @blegat 
Can anyone review this PR?

mtanneau:
I'm not sure I understand the motivation here.

In what context would this be used? And to which extent are existing functions not sufficient for doing so?

Project.toml
```diff
@@ -11,6 +11,7 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 MutableArithmetics = "d8a4904e-b15c-11e9-3269-09a3773c0cb0"
 NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
+PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
```
odow: This also needs an entry in the `[compat]` section.

src/JuMP.jl
```diff
@@ -474,6 +475,38 @@ function solve_time(model::Model)
     return MOI.get(model, MOI.SolveTime())
 end
 
+
+function result!(model::Model; dual_query::Bool = false)
```
odow: What about: `print_solution_summary(model::Model)`?

src/JuMP.jl
```diff
@@ -474,6 +475,38 @@ function solve_time(model::Model)
...
+    data = ["termination status" termination_status(model)]
+
+    if data[2] != MOI.OPTIMIZE_NOT_CALLED
```
odow: Instead of long `if` blocks, prefer
```Julia
if data[2] == MOI.OPTIMIZE_NOT_CALLED
    return pretty_table(data)
end

data = ...
```

yashcodes:
`result!` function basically show the complete result status in a clean tabular form, which also give the option to the user of which data you want to see or not (like the result of dual of the problem)(suggested by [#2191](https://github.com/JuliaOpt/JuMP.jl/issues/2191)).
I agree that the function can be improved, I just wanted reviews on the idea of this kind of function.

yashcodes:
Ok @odow that was a mistake to ping people for reviews, I'll take care of that in future.

Project.toml
yashcodes: Yes right!!

odow:
So the action item for this is to find what other modeling languages do. e.g., AMPL, GAMS, Excel, Pyomo, CVXPY... They may have good ideas about what information to present.

yashcodes:
The idea behind this PR was simply just to present whatever results a solver can give us by using just one function `JuMP.result(m::Model)`

yashcodes:
I added a few tests, hope it will make clear the aim of the PR.

odow:
Closing because this is not what were are after. We can revisit in a new PR after some discussion in the issue.

PR #2205, opened 2020-03-17
Function `reduced_cost` (code, documentation, and tests)

henriquebecker91:
Incentivized by @odow, also discussed with @mlubin, both in the [old MathOptOInterface issue #776](https://github.com/JuliaOpt/MathOptInterface.jl/issues/776).

The code was suggested by me, corrected by @odow (two times), and in this last commit I just added an error message for the case in which the user changed the sense to feasibility sense after solving with min or max sense. I followed the definition given by @odow in the documentation (basically used the the one suggested by him, and added reference to the new function in related places, like `shadow_price`). I had difficulties with testing the method without using a solver underneath and created my own mock-up (I tried to use the MockOptimizer but soon I got frustrated). Maybe this was wildly wrong, and the JuMP developers will prefer to remove these tests, however, I really cannot see myself remaking all the tests (I need to work on my PhD now, and I will be able to make a significant changes/rewrites to this PR only by the end of may, probably). Just tell me what you people think needs to be changed that I will do what I think I can do and leave the rest for the maintainers (I allowed edits to this branch/PR from the JuMP maintainers).

All tests are passing (new and old), and from what I have seen the documentation is being correctly generated.

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -926,7 +926,7 @@ function _moi_add_constrained_variables(
 end
 
 """
-    reduced_cost(x::AbstractVariableRef)::Float64
+    reduced_cost(x::VariableRef)::Float64
 
 Return the reduced cost associated with variable `x`.
 
@@ -935,19 +935,11 @@ Equivalent to querying the shadow price of the active variable bound
 
 See also: [`shadow_price`](@ref).
 """
-function reduced_cost(x::AbstractVariableRef)::Float64
+function reduced_cost(x::VariableRef)::Float64
     model = owner_model(x)
     if !has_duals(model)
         error("Unable to query reduced cost of variable because model does" *
               " not have duals available.")
-    end
-    # I am not sure if a feasibility sense model may have duals, but the
-    # sense may be changed after solving and having an error message help
-    # to discover this mistake.
-    obj_sense = objective_sense(model)
-    if obj_sense != MOI.MIN_SENSE && obj_sense != MOI.MAX_SENSE
-        error("The reduced cost is not available because the objective" *
-              " sense $obj_sense is not minimization or maximization.")
     end
     sign = objective_sense(model) == MOI.MIN_SENSE ? 1.0 : -1.0
     if is_fixed(x)
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -928,37 +928,36 @@ end
 """
     reduced_cost(x::VariableRef)::Float64
 
-Concerning an optimally solved linear model, the reduced cost of a variable
-is how much the objective function coefficient of that variable has to change
-for the variable appear in an optimal solution (i.e., have a positive value,
-become a basic variable).
+Return the reduced cost associated with variable `x`.
 
-Querying the reduced cost of a variable that is neither fixed nor has upper
-or lower bounds will always return zero.
+Equivalent to querying the shadow price of the active variable bound
+(if one exists and is active).
 
-The pitfalls and restrictions to the usage of
-[`shadow_price`](@ref JuMP.shadow_price) are also applicable to this method
-(which uses it internally).
-
-# Notes
-
-Follows the definition given above that:
-
-* If a variable has value zero and a reduced cost that is also zero, then
-  there is another optimal solution in which the variable has a non-zero value.
-* If a variable has a positive value, its reduced cost is always zero.
-
+See also: [`shadow_price`](@ref).
 """
 function reduced_cost(x::VariableRef)::Float64
-    !has_duals(owner_model(x)) && error(
-        "The reduced cost is not available because no dual result is" *
-        " available."
-		)
-    is_fixed(x) && return shadow_price(FixRef(x))
+    model = owner_model(x)
+    if !has_duals(model)
+        error("Unable to query reduced cost of $(x) because duals are not" *
+              " available.")
+    end
+		obj_sense = objective_sense(model)
+		if obj_sense != MOI.MIN_SENSE && obj_sense != MOI.MAX_SENSE
+        error("The reduced cost is not available because the objective" *
+              " sense $sense is not minimization or maximization.")
+		end
+    sign = objective_sense(model) == MOI.MIN_SENSE ? 1.0 : -1.0
+    if is_fixed(x)
+        return sign * dual(FixRef(x))
+    end
     rc = 0.0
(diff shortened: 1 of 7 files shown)


src/variables.jl
```diff
@@ -925,6 +925,44 @@ function _moi_add_constrained_variables(
...
+    end
+    # I am not sure if a feasibility sense model may have duals, but the
+    # sense may be changed after solving and having an error message help
+    # to discover this mistake.
+    obj_sense = objective_sense(model)
+    if obj_sense != MOI.MIN_SENSE && obj_sense != MOI.MAX_SENSE
+        error("The reduced cost is not available because the objective" *
+              " sense $obj_sense is not minimization or maximization.")
+    end
+    sign = objective_sense(model) == MOI.MIN_SENSE ? 1.0 : -1.0
```
odow: If, by "FEASIBILITY", we mean `min 0' x`, then sure, feasibility problems can have duals. I would just say that 
`sign = objective_sense(model) == MOI.MAX_SENSE ? -1.0 : 1.0` to cover both min and feasibility.

test/variable.jl
```diff
@@ -673,3 +674,97 @@ end
         "Julia's broadcast syntax instead: `JuMP.value.(x)`.")
     @test_throws exception JuMP.value(x)
 end
+
+struct MockModel <: JuMP.AbstractModel
+    has_duals::Bool
+    obj_sense::MathOptInterface.OptimizationSense
+end
```
odow: Take a look at mock optimizer in https://github.com/JuliaOpt/JuMP.jl/blob/master/test/generate_and_solve.jl

src/variables.jl
henriquebecker91: I was not sure this is the meaning of FEASIBILITY_SENSE and for this reason I included the error. I removed it in the last commit, as it is not of my interest I will not dispute this.

henriquebecker91:
I changed the tests code to use the `MockOptimizer`, however, I left the dual bonded variable tests commented because the `MockOptimizer` fails to query the bound duals (this happens not at the mock model building, but when the mock is passed to `reduced_cost`). [This is the error `reduced_cost` raises.](https://github.com/JuliaOpt/MathOptInterface.jl/blob/80fc0eaa40b8fb6924be41d94c5cdd606d244998/src/Utilities/results.jl#L331) I am not sure if I did something wrong, or just `MockOptimizer` has no support for this kind of testing.

test/variable.jl
```diff
@@ -673,3 +673,121 @@ end
...
+		m = JuMP.direct_model(mockoptimizer)
+		if var_bound_type === :lower
+				@variable(m, x >= 0)
+				if has_duals
+						@assert isa(var_bounds_dual, Float64)
+						has_duals && MOI.set(
+								mockoptimizer, MOI.ConstraintDual(),
+								JuMP.optimizer_index(JuMP.LowerBoundRef(x)), var_bounds_dual
+						)
+				end
```
odow: Indentation thought this function seems off?

henriquebecker91:
Sorry, I use tabs by default and change them to spaces before sending the
PR. Today I am a little tired and forgot to fix it before sending. I will
fix it asap.

Em ter., 17 de mar. de 2020 às 15:47, Oscar Dowson <notifications@github.com>
escreveu:

henriquebecker91:
I talked to my co-advisor briefly today, and he found this definition of reduced cost to be very strange. For him, you need to look to all constraints in which the variable appears to be able to deduce a reduced cost (not only the variable bounds). He said he will study this question a little and give me a return.

odow:
Most solvers agree with our definition.

For example, CPLEX says ["The values returned by this method are defined to be the dual multipliers for bound constraints on the specified variables."](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.10.0/ilog.odms.cplex.help/refcallablelibrary/cpxapi/getdj.html).

A variable can be at a bound with a reduced cost of 0 if the dual is degenerate (i.e., there are multiple optimal primal solutions).

test/variable.jl
```diff
@@ -673,3 +673,121 @@ end
...
+            @assert eltype(var_bounds_dual) == Float64
+            lb_dual, ub_dual = var_bounds_dual
+            MOI.set(
+                mockoptimizer, MOI.ConstraintDual(),
+                JuMP.optimizer_index(JuMP.LowerBoundRef(x)), lb_dual
+            )
+            MOI.set(
+                mockoptimizer, MOI.ConstraintDual(),
+                JuMP.optimizer_index(JuMP.UpperBoundRef(x)), ub_dual
+            )
```
odow: So what error does this throw?

src/variables.jl
blegat: I agree with odow, using `-1`on `MAX_SENSE` only seems more sensible

test/variable.jl
```diff
@@ -673,3 +673,121 @@ end
...
+)
+    mockoptimizer = MOIU.MockOptimizer(
+        MOIU.Model{Float64}(), eval_objective_value = false
+    )
+    m = JuMP.direct_model(mockoptimizer)
+    if var_bound_type === :lower
+        @variable(m, x >= 0)
+        if has_duals
+            @assert isa(var_bounds_dual, Float64)
+            has_duals && MOI.set(
```
blegat, line 694: `has_dual &&` here is useless as we are inside `if has_duals`

test/variable.jl
blegat: Which error ?

odow:
Hey @henriquebecker91, I went to fix up a couple of things to get this merged. Where did the repo go? :(

Otherwise, @blegat I vote we merge this, and I'll make another PR cleaning up some of the tests. The content in `src/` is good.

henriquebecker91:
My bad @odow, I forgot about this PR and when I finished a previous PR I deleted my fork of JuMP because I thought everything was merged. Can I do something to help you? I see the changes were not lost, do you want me to create a new fork (and a new PR)?

I did not go back to this because it is overkill in my case (I know the maximization sense and I am sure no variable is from other model), because my co-advisor has not yet returned to me about his impressions of this reduced cost definition, and because I am focusing in my PhD now, but if you need a little of my work to finish this you can just ask.

odow:
Will follow up with another PR fixing @blegat's comments.

PR #2211, opened 2020-03-27
Intial work for batch versions of fix, unfix, delete_lower/upper_bound and similar

henriquebecker91:
Based on a gitter discussion. Work in progress.

Adds a new (untested) internal fix method for a list of variables and values (to be yet exposed). Adds list/batch variants for unfix, delete_lower_bound, and delete_upper_bound.

Seems like MOI needs no changes, it already predict batch versions. The only things that need to be changed are JuMP (to give the end user access) and the solver package (to provide an efficient batch variant, avoiding the MOI fallback).

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -157,6 +157,18 @@ function check_belongs_to_model(v::AbstractVariableRef, model::AbstractModel)
     end
 end
 
+# Used in the batch methods.
+function _check_share_model(variables::Vector{VariableRef}, method_name)
+    common_owner = owner_model(first(variables))
+    if any(var -> owner_model(var) != common_owner, variables)
+        error(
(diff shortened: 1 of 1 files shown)


henriquebecker91:
@odow What do you think about this one? It was created with [Gurobi.jl#300](https://github.com/JuliaOpt/Gurobi.jl/pull/300) in mind, but what is done is mainly exposing the (maybe specialized) batch/vector methods for `fix`, `unfix` and `delete_{lower,upper}_bound` for which MOI already provides a fallback. Do you think it is worth some extra work (like adding tests), or should be abandoned?

odow:
I don't see a strong reason to implement this. The gurobi performance issue is now fixed. Would prefer to close and use the broadcasting fallback.

henriquebecker91:
Abandoning this then.

PR #2212, opened 2020-03-27
Added SOS1 and SOS2 struct

tomasfmg:


src/sets.jl
```diff
@@ -73,3 +73,64 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
...
+vector `x` to a set where at most 1 variable can take a non-zero value, all
+others being at 0. 
+
+The `weights`, when specified, induce an ordering of the variables; as such, 
+they should be unique values. The *k*th element in the set corresponds to the
+*k*th weight in `weights`. See [here](http://lpsolve.sourceforge.net/5.5/SOS.htm)
+for a description of SOS constraints and their potential uses.
+
+This is a shortcut for the `MathOptInterface.SOS1`
+
```
odow: Remove blank line.

src/sets.jl
```diff
@@ -73,3 +73,64 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
...
+SOS1 (Special Ordered Sets type 2) object than can be used to constrain a 
+vector `x` to a set where at most 2 variables can take a non-zero value, all
+others being at 0. In addition, if two are non-zero these must be consecutive
+in their ordering. 
+
+The `weights` induce an ordering of the variables; as such, they should be unique
+values. The *k*th element in the set corresponds to the *k*th weight in `weights`.
+See [here](http://lpsolve.sourceforge.net/5.5/SOS.htm) for a description of SOS
+constraints and their potential uses. This is a shortcut for the `MathOptInterface.SOS2`
+
```
odow: Remove blank line.

src/sets.jl
```diff
@@ -73,3 +73,64 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
...
+SOS1 (Special Ordered Sets type 1) object than can be used to constrain a 
+vector `x` to a set where at most 1 variable can take a non-zero value, all
+others being at 0. 
+
+The `weights`, when specified, induce an ordering of the variables; as such, 
+they should be unique values. The *k*th element in the set corresponds to the
+*k*th weight in `weights`. See [here](http://lpsolve.sourceforge.net/5.5/SOS.htm)
+for a description of SOS constraints and their potential uses.
+
+This is a shortcut for the `MathOptInterface.SOS1`
```
odow: "for the `MathOptInterface.SOS1` set."

All sentences should be full sentences that end in a full stop.

src/sets.jl
```diff
@@ -73,3 +73,64 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
...
+
+SOS1 (Special Ordered Sets type 2) object than can be used to constrain a 
+vector `x` to a set where at most 2 variables can take a non-zero value, all
+others being at 0. In addition, if two are non-zero these must be consecutive
+in their ordering. 
+
+The `weights` induce an ordering of the variables; as such, they should be unique
+values. The *k*th element in the set corresponds to the *k*th weight in `weights`.
+See [here](http://lpsolve.sourceforge.net/5.5/SOS.htm) for a description of SOS
+constraints and their potential uses. This is a shortcut for the `MathOptInterface.SOS2`
```
odow: Ditto on the final sentence.

test/macros.jl
```diff
@@ -265,6 +265,24 @@ function macros_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Typ
         @test con.set == MOI.SecondOrderCone(2)
     end
 
+    @testset "@build_constraint (SOS1)" begin 
+        model = ModelType() 
+        @variable(model, x[1:3]) 
+        con = @build_constraint(x in JuMP.SOS1()) 
+        @test con isa JuMP.VectorConstraint 
+        @test con.func == x 
+        @test con.set == MOI.SOS1([1.0, 2.0, 3.0])
```
odow: You need a test for the error condition.

Something like (I haven't tested):
```Julia
@test_throws(
    ErrorException("Weight vector in SOS1 is not of length 3."),
    @build_constraint(x in JuMP.SOS1([1.0])
)
```

test/macros.jl
```diff
@@ -265,6 +265,24 @@ function macros_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Typ
...
+        @test con.set == MOI.SOS1([1.0, 2.0, 3.0]) 
+    end 
+
+    @testset "@build_constraint (SOS2)" begin 
+        model = ModelType() 
+        @variable(model, x[1:3]) 
+        con = @build_constraint(x in JuMP.SOS2()) 
+        @test con isa JuMP.VectorConstraint 
+        @test con.func == x 
+        @test con.set == MOI.SOS2([1.0, 2.0, 3.0])
```
odow: Ditto error condition.

src/sets.jl
```diff
@@ -73,3 +73,64 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
...
+        new(convert(Vector{Float64}, weights))
+    end
+end
+function moi_set(set::SOS1, dim::Int)
+    if length(set.weights) == 0
+        return MOI.SOS1(collect(1:1.0:dim))
+    elseif length(weights) == dim
+        return MOI.SOS1(weights)
+    else
+        error("Weight vector in SOS1 is not of length($dim)")
```
odow: not of length $(dim)."

src/sets.jl
```diff
@@ -73,3 +73,64 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
...
+        new(convert(Vector{Float64}, weights))
+    end
+end
+function moi_set(set::SOS2, dim::Int)
+    if length(set.weights) == 0
+        return MOI.SOS2(collect(1:1.0:dim))
+    elseif length(weights) == dim
+        return MOI.SOS2(weights)
+    else
+        error("Weight vector in SOS2 is not of length($dim)")
```
odow: not of length $(dim)."

src/JuMP.jl
```diff
@@ -9,7 +9,7 @@
 #############################################################################
 
 module JuMP
-
+@show("oi")
```
odow: Please remove. Otherwise, looks good.

src/sets.jl
```diff
@@ -73,3 +73,60 @@ function LinearAlgebra.norm(::AbstractVector{<:AbstractJuMPScalar})
...
+others being at 0. 
+The `weights`, when specified, induce an ordering of the variables; as such, 
+they should be unique values. The *k*th element in the set corresponds to the
+*k*th weight in `weights`. See [here](http://lpsolve.sourceforge.net/5.5/SOS.htm)
+for a description of SOS constraints and their potential uses.
+This is a shortcut for the `MathOptInterface.SOS1` set.
+"""
+struct SOS1 <: AbstractVectorSet 
+    weights::Vector{Float64}
+    function SOS1(weights::AbstractVector = Float64[])
```
blegat: Why providing a default constructor ? It will be weird for the user to get an error "Weight vector in SOS1 is not of length ..." when doing
```
@constraint(model, x in SOS1())
```
since they do not provide any length vector

odow: In `moi_set`, if the weight vector is empty, a default one will be created.

So you can go
```Julia
model = Model()
@variable(model, x[1:3])
@constraint(model, x in SOS1())  # Good
@constraint(model, x in SOS1([1, 2, 3]))  # Good
@constraint(model, x in SOS1([1, 2, 3, 4]))  # Errors!
```

test/macros.jl
```diff
@@ -265,6 +265,32 @@ function macros_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Typ
         @test con.set == MOI.SecondOrderCone(2)
     end
 
+    @testset "@build_constraint (SOS1)" begin 
+        model = ModelType() 
+        @variable(model, x[1:3]) 
+        con = @build_constraint(x in JuMP.SOS1()) 
+        @test con isa JuMP.VectorConstraint 
+        @test con.func == x 
+        @test con.set == MOI.SOS1([1.0, 2.0, 3.0])
```
odow: Looks like you also need to test:
```Julia
con2 = @build_constraint(x in JuMP.SOS1([1, 3, 2])
@test con2.func == x
@test con2.set == MOI.SOS1([1.0, 3.0, 2.0])
```
And ditto for SOS2.

You can see which lines aren't covered by looking at the coverage report:
https://codecov.io/gh/JuliaOpt/JuMP.jl/compare/2532e46ad79bebc75407a82a69e9ee11bec5568b...96bfa1610683131c71933d8c4bbdbdd32713a03c/diff

tomasfmg: Okay! Thank you, just added it

test/macros.jl
```diff
@@ -265,6 +265,40 @@ function macros_test(ModelType::Type{<:JuMP.AbstractModel}, VariableRefType::Typ
         @test con.set == MOI.SecondOrderCone(2)
     end
 
+    @testset "@build_constraint (SOS1)" begin 
+        model = ModelType() 
+        @variable(model, x[1:3]) 
+        con = @build_constraint(x in JuMP.SOS1())
+        con2 = @build_constraint(x in JuMP.SOS1([1,2,3]))
```
odow: So one rule of testing. You want to make it so that the cases are unambiguous. In this case (I know the code is correct) if you passed a vector `[1, 2, 3]` it might still go down the `collect(1:1.0:dim)` path, and your test won't catch it. But if you pass the vector `[4.0, 6.0, 5.0]`, you're testing i) that the numbers don't have to be 1, 2, 3; and ii) that they don't have to be sorted in order.

tomasfmg: Got it, changing it right now!

odow:
Looks good. Just the doc section and we're done!

docs/src/constraints.md
```diff
@@ -876,6 +876,60 @@ julia> @constraint(model, M * y + q ⟂ y)
 [y[1] + 2 y[2] + 5, 3 y[1] + 4 y[2] + 6, y[1], y[2]] ∈ MathOptInterface.Complements(2)
 ```
 
+## Special Ordered Sets (SOS1 and SOS2)
+
+A Special Ordered Set (SOS) is an ordered set of variables with particular characteristics.
```
odow, line 881: s/with particular/with the following

docs/src/constraints.md
```diff
@@ -876,6 +876,60 @@ julia> @constraint(model, M * y + q ⟂ y)
...
+
+If a vector `x` is in a Special Ordered Set type 1 (SOS1), at most one element (or variable) of `x`
+can take a non-zero value, all others being zero, in a feasible solution. To induce ordering,
+`weights` can be provided; as such, they should be unique values. The `k`th element in the set
+corresponds to the `k`th weight in `weights`.
```
odow, line 886: Maybe: 

If a vector of variables `x` is in a Special Ordered Set of Type I (SOS1), then at most one element of `x` can take a non-zero value, and all other elements must be zero. 

Although not required for feasibility, solvers can benefit from an ordering of the variables (e.g., the variables represent different factories to build, at most one factory can be built, and the factories can be ordered according to cost). To induce an ordering, `weights` can be provided; as such, they should be unique values. The `k`th element in the ordered set corresponds to the `k`th weight in `weights` when the weights are sorted.

docs/src/constraints.md
```diff
@@ -876,6 +876,60 @@ julia> @constraint(model, M * y + q ⟂ y)
...
+A Special Ordered Set (SOS) is an ordered set of variables with particular characteristics.
+
+If a vector `x` is in a Special Ordered Set type 1 (SOS1), at most one element (or variable) of `x`
+can take a non-zero value, all others being zero, in a feasible solution. To induce ordering,
+`weights` can be provided; as such, they should be unique values. The `k`th element in the set
+corresponds to the `k`th weight in `weights`. 
+
+A SOS1 constraint is equivalent to:
+
+- `x[(i)] >= 0` for some `i`, where `(i)` is the induced ordering *i*th element
```
odow, line 890: No need for `(i)`, just `x[i]`

docs/src/constraints.md
```diff
@@ -876,6 +876,60 @@ julia> @constraint(model, M * y + q ⟂ y)
...
+- `x[(j)] == 0` for all `j != i`
+
+If a vector is in a Special Ordered Set type 2 (SOS2), no more than two adjacent elements may
+be non-zero in a feasibles solution. To induce ordering,
+`weights` can be provided; as such, they should be unique values. The `k`th element in the set
+corresponds to the `k`th weight in `weights`.
```
odow, line 896: If a vector of variables `x` is in a Special Ordered Set of Type II (SOS2), then at most two elements can be non-zero, and if two elements are non-zero, they must be adjacent.

Because of the adjacency requirement, you should supply a weight vector (with unique elements) to induce an ordering of the variables. The `k`th element in the ordered set corresponds to the `k`th weight in `weights` when the weights are sorted.

docs/src/constraints.md
```diff
@@ -876,6 +876,60 @@ julia> @constraint(model, M * y + q ⟂ y)
...
+
+If a vector is in a Special Ordered Set type 2 (SOS2), no more than two adjacent elements may
+be non-zero in a feasibles solution. To induce ordering,
+`weights` can be provided; as such, they should be unique values. The `k`th element in the set
+corresponds to the `k`th weight in `weights`. 
+
+A SOS2 constraint is equivalent to:
+
+- `x[(i)] >= 0`, `x[(i+1)] >= 0`  for some `i`, where `(i)` is the induced ordering *i*th element
+- `x[(j)] == 0` for all `j != i`, `j != i+1`
```
odow, line 901: Ditto on `(i)`

docs/src/constraints.md
```diff
@@ -876,6 +876,60 @@ julia> @constraint(model, M * y + q ⟂ y)
...
+
+To create a SOS constraint, one can do the following:
```
odow, line 903: Create an SOS constraint as follows:

docs/src/constraints.md
```diff
@@ -876,6 +876,60 @@ julia> @constraint(model, M * y + q ⟂ y)
...
+ x[1]
+ x[2]
+ x[3]
+
+julia> @constraint(model x in SOS2([3,5,2]))
+[x[1], x[2], x[3]] ∈ MathOptInterface.SOS2{Float64}([3.0, 5.0, 2.0])
+```
+
+In the case above, `x[3]` is the first variable and `x[2]` the last variable under the
+induced ordering.
```
odow, line 931: Maybe switch these examples around. Give the weight vector first, and then the `SOS2()` example.

docs/src/constraints.md
tomasfmg: the "s" refers to what?

odow: Ah. I forgot this can be non-obvious: https://unix.stackexchange.com/questions/159367/using-sed-to-find-and-replace

odow:
Why'd you close?

docs/src/constraints.md
```diff
@@ -876,6 +876,63 @@ julia> @constraint(model, M * y + q ⟂ y)
...
+induced ordering. When no ordering vector is provided, JuMP induces an ordering from `1:length(x)`.
+
+```jldoctest SOS
+julia> @variable(model, x[1:3])
+3-element Array{VariableRef,1}:
+ x[1]
+ x[2]
+ x[3]
+
+julia> @constraint(model x in SOS2())
```
odow, line 932: s/model x/model, x

docs/src/constraints.md
```diff
@@ -876,6 +876,63 @@ julia> @constraint(model, M * y + q ⟂ y)
...
+Create an SOS constraint as follows:
+
+```jldoctest SOS
+julia> @variable(model, x[1:3])
+3-element Array{VariableRef,1}:
+ x[1]
+ x[2]
+ x[3]
+
+julia> @constraint(model x in SOS2([3,5,2]))
```
odow, line 918: s/model x/model, x

PR #2214, opened 2020-03-30
closes #2094

### LINKED ISSUE #2094: Expose and document dual warm starts
Looks like we're missing a JuMP function analogous to `set_start_value` to expose `MOI.ConstraintDualStart`. It should also be documented here: http://www.juliaopt.org/JuMP.jl/dev/constraints/.

AKS1996:
Added two analogous method definition with a simple test

### DIFF
--- a/docs/src/constraints.md
+++ b/docs/src/constraints.md
@@ -235,6 +235,29 @@ The constraint can also be retrieved from its name using
 constraint_by_name
 ```
 
+## Start Values
+
+There is a way to assign a starting value to some constraint's dual (for warmstarting the optimizer) for each constraint: using [`set_dual_start_value`](@ref).
+
+The start value of a constraint's dual can be queried using [`dual_start_value`](@ref). If
+no start value has been set, [`dual_start_value`](@ref) will return `Nothing`.
+
+```jldoctest constraint_dual_start; setup=:(model=Model())
+julia> @variable(model, x)
+x
+
+julia> @constraint(model, con, x >= 10)
+con : x ≥ 10.0
+
+julia> dual_start_value(con)
+
+julia> set_dual_start_value(con, 2)
+2.0
+
+julia> dual_start_value(con)
+2.0
+```
+
 ## Constraint containers
 
 So far, we've added constraints one-by-one. However, just like
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -52,6 +52,28 @@ end
 Base.broadcastable(con_ref::ConstraintRef) = Ref(con_ref)
 
 """
+    dual_start_value(con_ref::ConstraintRef)
+
+Return the dual start value (MOI attribute `ConstraintDualStart`) of the constraint `con_ref`. See also [`set_dual_start_value`](@ref).
+
+Note: If no dual start value has been set, `dual_start_value` will return `Nothing`
+"""
+function dual_start_value(con_ref::ConstraintRef)::Union{Nothing, Float64}
+    return MOI.get(owner_model(con_ref), MOI.ConstraintDualStart(), con_ref)
+end
+
+"""
+    set_dual_start_value(con_ref::ConstraintRef, value::Number)
+
+Set the dual start value (MOI attribute `ConstraintDualStart`) of the constraint `con_ref` to `value`. See also [`dual_start_value`](@ref).
+"""
+function set_dual_start_value(con_ref::ConstraintRef, value::Number)
+    MOI.set(owner_model(con_ref), MOI.ConstraintDualStart(), con_ref,
+            Float64(value))
+    return
+end
+
+"""
     name(con_ref::ConstraintRef)
 
 Get a constraint's name attribute.
--- a/test/constraint.jl
+++ b/test/constraint.jl
@@ -119,6 +119,15 @@ function constraints_test(ModelType::Type{<:JuMP.AbstractModel},
         @test_throws Exception JuMP.delete(second_model, constraint_ref)
     end
 
+    @testset "get and set dual start" begin
+        model = ModelType()
+        @variable(model, x)
+        constraint_ref = @constraint(model, 2x <= 1)
+	con_val = 2.0
+	JuMP.set_dual_start_value(constraint_ref, con_val)
+        @test JuMP.dual_start_value(model, constraint_ref) == con_val
+    end
+
     @testset "batch delete / is_valid constraints" begin
         model = ModelType()
         @variable(model, x[1:9])


AKS1996:
Hi.
I've added two methods for exposing dual start values. But they're not accessible in tests file. Any suggestions how to solve this?

AKS1996:
I'm not sure why the function is not accessible :sweat_smile: . The test implementation is almost same as https://github.com/JuliaOpt/JuMP.jl/blob/master/test/variable.jl#L221

I guess I need to define `set_dual_start_value` in `JuMPExtension.jl` as this one: https://github.com/JuliaOpt/JuMP.jl/blob/master/test/JuMPExtension.jl#L163

test/JuMPExtension.jl
```diff
@@ -225,6 +225,14 @@ function JuMP.add_constraint(model::MyModel, c::JuMP.AbstractConstraint,
     JuMP.set_name(cref, name)
     return cref
 end
+
+function JuMP.dual_start_value(constraint_ref::MyConstraintRef)::Union{Nothing, Float64}
+    return dual_start_value(constraint_ref)
+end
+function JuMP.set_dual_start_value(constraint_ref::MyConstraintRef, value::Float64)
+    set_dual_start_value(constraint_ref, value)
```
AKS1996: I'm not sure how to edit this method. Should be similar to https://github.com/JuliaOpt/JuMP.jl/blob/master/test/JuMPExtension.jl#L163 right?

odow: Easiest thing is to probably not implement these methods, move the `@testset` to inside this testset 
https://github.com/JuliaOpt/JuMP.jl/blob/4ffcd1ff2423f69718d1ab18e3af882594bc75e7/test/constraint.jl#L484, and use `model = Model()` instead of `model = ModelType()`.

AKS1996: Thanks. works now :v:

blegat:
I am wonder whether `_value` is superfluous here. For instance, we have `dual`, not `dual_value` for accessing the value of the dual.
However, we have `start_value` and `set_start_value` so for consistency, we should probably use the `_value` suffix here too.

test/constraint.jl
```diff
@@ -563,6 +563,13 @@ end
             (Vector{AffExpr}, MOI.PositiveSemidefiniteConeSquare),
             (Vector{QuadExpr}, MOI.RotatedSecondOrderCone)])
     end
+    @testset "get and set dual start" begin
+	model = Model()
+	@variable(model, x)
+        con = @constraint(model, 2x <= 1)
```
blegat, line 569: The indenting of this line is inconsistent

test/constraint.jl
```diff
@@ -563,6 +563,13 @@ end
...
+	set_dual_start_value(con, 2)
+	@test dual_start_value(con) == 2.0
```
blegat, line 571: Does `@test 2.0 == @inferred dual_start_value(con)` pass ?

AKS1996: no the test is failing.
Should it return 0 instead of `Nothing` when no value is set. At least will be type stable then

src/constraints.jl
```diff
@@ -51,6 +51,32 @@ end
...
+"""
+    dual_start_value(con_ref::ConstraintRef)
+
+Return the dual start value (MOI attribute `ConstraintDualStart`) of the constraint `con_ref`. 
+
+Note: If no dual start value has been set, `dual_start_value` will return `Nothing`
+
+See also [`set_dual_start_value`](@ref).
+"""
+function dual_start_value(con_ref::ConstraintRef)::Union{Nothing, Float64}
```
joaquimg, line 63: What about vector constraints?
The current implementations is only for ScalarShape?

docs/src/constraints.md
```diff
@@ -235,6 +235,28 @@ The constraint can also be retrieved from its name using
 constraint_by_name
 ```
 
+## Start Values
+
+There is a way to assign a starting value to some constraint's dual (for warmstarting the optimizer) for each constraint: using [`set_dual_start_value`](@ref).
```
odow, line 240:
```
Provide a starting value (also called warmstart) for a constraint's dual using [`set_dual_start_value`](@ref).
```

test/constraint.jl
```diff
@@ -614,6 +614,25 @@ function test_Model_list_of_constraint_types(::Any, ::Any)
...
+    model = Model()
+    @variable(model, x)
+    con = @constraint(model, 2x <= 1)
+    # scalar
+    @test dual_start_value(con) === nothing
+    set_dual_start_value(con, 2)
+    @test dual_start_value(con) == 2.0
+    set_dual_start_value(con, nothing)
+    @test dual_start_value(con) === nothing
+    # vector
```
odow, line 627: Prefer breaking these into separate functions. The smaller the test function, the better.

src/constraints.jl
```diff
@@ -51,6 +51,66 @@ end
...
+end
+function set_dual_start_value(con_ref::ConstraintRef{
+    Model, <:_MOICON{<:MOI.AbstractVectorFunction, <:MOI.AbstractVectorSet}},
+    ::Nothing)
+    MOI.set(owner_model(con_ref), MOI.ConstraintDualStart(), con_ref, nothing)
+    return
+end
+function set_dual_start_value(con_ref::ConstraintRef{
+    Model, <:_MOICON{<:MOI.AbstractScalarFunction, <:MOI.AbstractScalarSet}},
+    value)
```
odow, line 109: For consistency, the `)` should be dropped onto a new line.

src/constraints.jl
```diff
@@ -51,6 +51,66 @@ end
...
 Base.broadcastable(con_ref::ConstraintRef) = Ref(con_ref)
 
+"""
+    dual_start_value(con_ref::ConstraintRef)
+
+Return the dual start value (MOI attribute `ConstraintDualStart`) of the
+constraint `con_ref`. 
+
+Note: If no dual start value has been set, `dual_start_value` will return
+`nothing`
```
odow, line 61: Missing `.`

docs/src/constraints.md
```diff
@@ -235,6 +235,28 @@ The constraint can also be retrieved from its name using
...
+julia> @constraint(model, con, x >= 10)
+con : x ≥ 10.0
+
+julia> dual_start_value(con)
+
+julia> set_dual_start_value(con, 2)
+
+julia> dual_start_value(con)
+2.0
+```
```
odow, line 258: Maybe an example setting a vector start?

docs/src/constraints.md
```diff
@@ -235,6 +235,28 @@ The constraint can also be retrieved from its name using
...
+
```
mlubin, line 259: What about explaining how to transfer an optimal dual solution to a dual warm start? This is not as easy to do as with variables.

joaquimg: Shouldn't it be trivial since we are using shapes?
`sol = dual(con)`
`sol` would be in the appropriate shape (scalar, vector, matrix, etc), which depends on the constraint type.
Then
`set_dual_start_value(con, sol)` should just work since a value in the proper shape is expected.

joaquimg: Now think you meant this due `list_of_constraint_types`

mlubin: Yeah, I meant to mention the equivalent of:
```
set_start_value.(all_variables(model), value.(all_variables(model)))
```
We discuss this at https://jump.dev/JuMP.jl/v0.21.5/variables/#Start-values-1, so we should do the same for constraints.

docs/src/constraints.md
```diff
@@ -257,6 +258,35 @@ julia> dual_start_value(con)
...
+julia> set_dual_start_value(con, [1.0, 3.0])
+
+julia> dual_start_value(con)
+2-element Array{Float64,1}:
+ 1.0
+ 3.0
+```
+
+To set the previous solution to a new solve, use:
```
mlubin, line 280: "To take the dual solution from the last solve and use it as the starting point for a new solve, use:"

docs/src/constraints.md
```diff
@@ -235,6 +235,59 @@ The constraint can also be retrieved from its name using
...
+```
+
+A vector constraint will require a vector warmstart:
+
+```jldoctest constraint_dual_start_vector; setup=:(model=Model())
+julia> @variable(model, x)
+x
+
+julia> @constraint(model, con, [x, x] in SecondOrderCone())
+con : [x, x] in MathOptInterface.SecondOrderCone(2)
```
odow, line 268: How about `@variable(model, x[1:3])`. `@constraint(model, con, x in SecondOrderCone())`. This is a weird constraint to add.

docs/src/constraints.md
```diff
@@ -235,6 +235,59 @@ The constraint can also be retrieved from its name using
...
+To take the dual solution from the last solve and use it as the starting point
+for a new solve, use:
+
+```julia
+for (F, S) in list_of_constraint_types(model)
+    for con in all_constraints(model, F, S)
+        set_dual_start_value(con, dual(con))
+    end
+end
+```
```
odow, line 289: Note that this will error if there are constraints for which the dual is not defined. You may want to choose only a subset of `F`-`S` pairs to set a dual start for.

joaquimg: This might be a little complicated...
You want to add filters in the example or just a note?

odow: I was proposing adding the sentence as a note.

joaquimg: done! let me know if its ok

src/constraints.jl
```diff
@@ -51,6 +51,69 @@ end
...
+
+Set the dual start value (MOI attribute `ConstraintDualStart`) of the constraint
+`con_ref` to `value`. To remove a dual start value set it to `nothing`.
+
+See also [`dual_start_value`](@ref).
+"""
+function set_dual_start_value(con_ref::ConstraintRef{
+    Model, <:_MOICON{<:MOI.AbstractVectorFunction, <:MOI.AbstractVectorSet}},
+    value
+    )
```
blegat, line 96: Should this closing parenthesis be unindended?

blegat (approved):

mlubin (approved):

PR #2219, opened 2020-04-03
Support MA.Zero in objective and constraints

odow:
Closes https://github.com/JuliaOpt/JuMP.jl/issues/2087


odow:
@blegat did that latest tag of MA break these tests?

odow:
Failing tests will be fixed with https://github.com/JuliaOpt/MutableArithmetics.jl/pull/46.

Issue #2035, opened 2019-08-21
Old error message on ambiguous variable name. Keyword "category" no longer exists


bhalonen:
This little bit of code 
```
using JuMP
model = Model()
indicator = @variable(model, Bin)
```
throws the error:
```
ERROR: LoadError: In `@variable(model, Bin)`: Ambiguous variable name Bin detected. Use the "category" keyword argument to specify a category for an anonymous variable.
```
Searching the word 'category' on the documentation yields no useful hits. Something should probably be added to the anonymous variable section.

mlubin:
Thanks for reporting. The `category` keyword no longer exists, so this error message needs to be updated. I believe the new recommendation for an anonymous binary variable would be:
```julia
using JuMP
model = Model()
indicator = @variable(model)
set_binary(indicator)
```

bhalonen:
ah. I actually gave up trying to track it down.
```
indicator = @variable(model, binary=true)
```
works as well.

PR #2222, opened 2020-04-07
Updating error message on ambiguous variable

yashcodes:
As keyword argument "category" no longer exist. Error message on ambiguous variable needs to be updated

Closes [#2035](https://github.com/jump-dev/JuMP.jl/issues/2035)

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1123,7 +1123,7 @@ macro variable(args...)
     else
         x = popfirst!(extra)
         if x in [:Int,:Bin,:PSD]
-            _error("Ambiguous variable name $x detected. Use the \"category\" keyword argument to specify a category for an anonymous variable.")
+            _error("Ambiguous variable name $x detected. To specify category of an anonymous variable use proper keyword argument or use appropiate function to add the required constraint on the variable.")
         end
         anon_singleton = false
     end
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -1123,7 +1123,7 @@ macro variable(args...)
     else
         x = popfirst!(extra)
         if x in [:Int,:Bin,:PSD]
-            _error("Ambiguous variable name $x detected. Use the \"category\" keyword argument to specify a category for an anonymous variable.")
+            _error("Ambiguous variable name $x detected. To specify category of an anonymous variable use proper keyword argument or use appropiate function to add the required constraint on the variable.")
```
odow, line 1126: We should instead tell people to use `is_integer = true`, `is_binary = true` etc.

```Julia
if x == :Int
    _error(
        "Ambiguous variable name $x detected. To specify an anonymous integer " *
        "variable, use `@variable(model, is_integer = true)` instead."
    )
elseif == :Bin
    ...
```

odow:
Tests? Coverage is a little confused.

yashcodes:
If you want I can add tests but I don't think tests are important for these error message.

yashcodes:
If you say I can add tests but I think tests are not important for these error messages.

yashcodes:
Yeah cool I add tests.

yashcodes:
Fine, I used `@test_macro_throws` :)

PR #2226, opened 2020-04-18
Defined Base.keys for SparseAxisArrays

### LINKED ISSUE #2117: SparseAxisArray does not implement size
As an `AbstractArray`, size should be defined for SparseAxisArray and return an `N`-sized tuple

pulsipher:
As per https://discourse.julialang.org/t/jump-container-functionality-feature-requests/37771, `Base.keys` is defined for `SparseAxisArray`s by returning the keys of the underlying dictionary. This allows for one method to provide the indices to iterate over `Array`s, `DenseAxisArray`s, and `SparseAxisArray`s.

Closes #2117

### DIFF
--- a/src/Containers/SparseAxisArray.jl
+++ b/src/Containers/SparseAxisArray.jl
@@ -114,6 +114,9 @@ Base.IndexStyle(::Type{<:SparseAxisArray}) = IndexAnyCartesian()
 # eachindex redirect to keys
 Base.keys(::IndexAnyCartesian, d::SparseAxisArray) = keys(d)
 
+# Redirect Base.keys to the underlying dictionary
+Base.keys(d::SparseAxisArray) = keys(d.data)
+
 ################
 # Broadcasting #
 ################
--- a/test/Containers/SparseAxisArray.jl
+++ b/test/Containers/SparseAxisArray.jl
@@ -75,6 +75,9 @@ SparseAxisArray{$Int,1,Tuple{Symbol}} with 2 entries:
             @test 3 * (d2 / 2) == d3
             @test (d2 / 2) * 3 == d3
         end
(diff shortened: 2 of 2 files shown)


src/Containers/SparseAxisArray.jl
```diff
@@ -114,6 +114,9 @@ Base.IndexStyle(::Type{<:SparseAxisArray}) = IndexAnyCartesian()
 # eachindex redirect to keys
 Base.keys(::IndexAnyCartesian, d::SparseAxisArray) = keys(d)
```
blegat, line 115: What was the previous behavior of `keys` ? This line redirect to `keys` so if `keys` does not work then according to the comment above, `eachindex` does not work either ?

odow:
@pulsipher why was this closed? We're thinking of merging it.

pulsipher:
I closed this based on blegat's comments on discourse, which seemed to indicate a more thorough solution was wanted and that the suggestion was just to define a '_keys' method for my needs. That being said I still support this being merged.

odow:
Having forgotten about this PR, we were about to implement this exact fix. So I guess it is reasonable. Currently, things are broken. We have a different fix incoming for keys and `DenseAxisArray`s.

odow:
@blegat happy?

blegat:
As mentioned on Discourse, this is not consistent with `keys` for `Array`. As we might may it consistent later, defining `keys` here and change it later would introduce a breaking change which is not desirable.
It's best to do `keys(d.data)` for now to cover the use case targeted by this PR.

odow:
Closing for now until we come up with a better long-term solution.

The recommendation is to do `keys(d.data)`, or to define a `keys` function in your own code that does this.

PR #2229, opened 2020-04-22
Add `rewrite_call_expression`.

dourouc05:
Excerpt from #2051.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -156,6 +156,7 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
     return parse_code, _build_call(_error, vectorized, variable, set)
 end
 
+expression_to_rewrite(head::Val{F}, args...) where F = false
 function rewrite_call_expression(_error::Function, head::Val{F}, args...) where F
     _error("function $F not implemented.")
 end
@@ -167,10 +168,10 @@ _functionize(::MutableArithmetics.Zero) = 0.0
 function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Val, lhs, rhs)
     parse_code_rhs, build_code_rhs, new_rhs = :(), :(), rhs
     parse_code_lhs, build_code_lhs, new_lhs = :(), :(), lhs
-    if isexpr(rhs, :call) && applicable(rewrite_call_expression, Val(rhs.args[1]), rhs.args[2:end]...)
+    if isexpr(rhs, :call) && expression_to_rewrite(Val(rhs.args[1]), rhs.args[2:end]...)
         parse_code_rhs, build_code_rhs, new_rhs = rewrite_call_expression(_error, Val(rhs.args[1]), rhs.args[2:end]...)
     end
-    if isexpr(lhs, :call) && applicable(rewrite_call_expression, Val(lhs.args[1]), lhs.args[2:end]...)
+    if isexpr(lhs, :call) && expression_to_rewrite(Val(lhs.args[1]), lhs.args[2:end]...)
         parse_code_lhs, build_code_lhs, new_lhs = rewrite_call_expression(_error, Val(lhs.args[1]), lhs.args[2:end]...)
     end
 
@@ -194,7 +195,7 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense
         parse_code = :($parse_code; $variable = _MA.mutable_operate!(-, convert(AffExpr, $variable), $new_rhs))
     elseif lhs != new_lhs && rhs == new_rhs
         variable, parse_code = _MA.rewrite(rhs)
-        parse_code = :($parse_code; $variable = _MA.mutable_operate!(-, $new_lhs, convert(AffExpr, $variable)))
+        parse_code = :($parse_code; $variable = _MA.mutable_operate!(-, convert(AffExpr, $new_lhs), $variable))
     else
         @assert rhs != new_rhs
         @assert lhs != new_lhs
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -156,25 +156,43 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
     return parse_code, _build_call(_error, vectorized, variable, set)
 end
 
+function rewrite_call_expression(_error::Function, head::Val{F}, args...) where F
+    _error("function $F not implemented.")
+end
+
 _functionize(v::VariableRef) = convert(AffExpr, v)
 _functionize(v::AbstractArray{VariableRef}) = _functionize.(v)
 _functionize(x) = x
 _functionize(::MutableArithmetics.Zero) = 0.0
 function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Val, lhs, rhs)
+    parse_code_rhs, build_code_rhs, new_rhs = :(), :(), rhs
+    if isexpr(rhs, :call) && applicable(rewrite_call_expression, Val(rhs.args[1]), rhs.args[2:end]...)
+        parse_code_rhs, build_code_rhs, new_rhs = rewrite_call_expression(_error, Val(rhs.args[1]), rhs.args[2:end]...)
+    end
+
     # Simple comparison - move everything to the LHS.
     #
     # `_functionize` deals with the pathological case where the `lhs` is a `VariableRef`
     # and the `rhs` is a summation with no terms. `_build_call` should be passed a
     # `GenericAffExpr` or a `GenericQuadExpr`, and not a `VariableRef` as would be the case
     # without `_functionize`.
-    if vectorized
-        func = :($lhs .- $rhs)
+    # TODO: bug in MutableArithmetics? The returned code does not find $new_rhs (ERROR: UndefVarError: #642###976 not defined).
+
+    if rhs == new_rhs
+        if vectorized
+            func = :($lhs .- $rhs)
+        else
+            func = :($lhs - $rhs)
+        end
+        variable, parse_code = _MA.rewrite(func)
     else
-        func = :($lhs - $rhs)
+        variable, parse_code = _MA.rewrite(lhs)
+        build_code = _build_call(_error, vectorized, :(_functionize($variable)), set)
+        parse_code = :($parse_code; $variable = _MA.mutable_operate!(-, convert(AffExpr, $variable), $new_rhs))
     end
+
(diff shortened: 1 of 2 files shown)


dourouc05:
I just rebased this on top of #2228, as it got merged.

src/macros.jl
```diff
@@ -156,25 +156,65 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
...
 _functionize(::MutableArithmetics.Zero) = 0.0
 function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Val, lhs, rhs)
+    parse_code_rhs, build_code_rhs, new_rhs = :(), :(), rhs
+    parse_code_lhs, build_code_lhs, new_lhs = :(), :(), lhs
+    if isexpr(rhs, :call) && expression_to_rewrite(Val(rhs.args[1]), rhs.args[2:end]...)
+        parse_code_rhs, build_code_rhs, new_rhs = rewrite_call_expression(_error, Val(rhs.args[1]), rhs.args[2:end]...)
+    end
+    if isexpr(lhs, :call) && expression_to_rewrite(Val(lhs.args[1]), lhs.args[2:end]...)
+        parse_code_lhs, build_code_lhs, new_lhs = rewrite_call_expression(_error, Val(lhs.args[1]), lhs.args[2:end]...)
+    end
```
blegat, line 184: Maybe do
```julia
parse_code_hhs, build_code_rhs, new_rhs = parse_expression(rhs)
```
to factor out the copy-past with lhs in `parse_expression`.

src/macros.jl
```diff
@@ -156,25 +156,65 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool,
...
+# To be defined if such a function must be rewritten; otherwise,
+# rewrite_call_expression will never be called.
+expression_to_rewrite(head::Val, args...) = false
+
+# When `expression_to_rewrite` called with the same arguments save the first one,
+# returns three things:
+# - code for parsing (which must be called first), if needed; otherwise, :().
+# - code for building the required constraints, if needed; otherwise, :().
+# - the symbol of the variable that replaces the expression (<: VariableRef).
+function rewrite_call_expression(_error::Function, head::Val{F}, args...) where F
```
blegat, line 168: If we remove `expression_to_rewrite`, we can give a default for this function of `:(), :(), rhs` instead of an error and everything would work with only one function to implement instead of two. Is that correct or am I missing something ?

dourouc05: If you write a constraint like `@constraint(m, sum(1 for _ in 1:10) == sum(1 for _ in 1:10))`, then you have a `:call` expression on both sides, but it should not be rewritten (or any function that should not be caught by JuMP).

blegat: Yes but if you replace the current fallback throwing an error by `return :(), :(), Expr(...)`, it will work for these cases too

dourouc05: If I rebuild the expression like this: 

```julia
function rewrite_call_expression(_error::Function, head::Val{F}, args...) where F
    return :(), :(), Expr(:call, [F, args...])
end
```

then it's MutableArithmetics that complains about the construct: 

```
MutableArithmetics.Zero (Issue #2187): Error During Test at C:\Users\Thibaut\.julia\dev\JuMP\test\macros.jl:50
  MethodError: objects of type Array{Any,1} are not callable
  Use square brackets [] for indexing an Array.
  Stacktrace:
   [1] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:218
   [2] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\test\macros.jl:52
   [3] top-level scope at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Test\src\Test.jl:1113
   [4] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\test\macros.jl:51
   [5] include(::String) at .\client.jl:439
   [6] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\test\runtests.jl:23
   [7] include(::String) at .\client.jl:439
   [8] top-level scope at none:6
   [9] eval(::Module, ::Any) at .\boot.jl:331
   [10] exec_options(::Base.JLOptions) at .\client.jl:264
   [11] _start() at .\client.jl:484
```

dourouc05: By the way, what would be your preferred way of implementing rewriting operations like needed in https://github.com/dourouc05/JuMP.jl/commit/e36c19682cf7920ee37805373a700673bfba4c8f#diff-6f46b954dc01974c3fb3f315681c3059R217-R247? Right now, the first error happens in MutableArithmetics, but I guess this package has nothing to do with this kind of rewritings, so maybe having a _rewrite_expr() function in JuMP would make sense?

dourouc05: The branch https://github.com/dourouc05/JuMP.jl/tree/dourouc05-rewrite-complex implements the idea of a rewriting function. It probably has some cost at runtime when building the model (each expression is fully traversed by this step, and it's mostly unavoidable to keep the same level of functionality), but it allows a lot of flexibility. In particular, this would allow parsing things like `a && b` without trouble. It's still orthogonal to #2228, which allows supporting new operators (like `:=`): here, it's really for the usual <=, >=, == constraints with funny things in either side. 

Also, it's easy to disable: the main calls to `_rewrite_expr` could be disabled if the user requires it, for instance. 

Shall I update this PR to include the new commits? 

However, there is still one thing to discuss: how to remove a constraint like `1 + f(x) <= y` from JuMP in case `f` adds a new variable and one constraint to do its job (like `z == x^2`, say)? When dealing with `f(x)`, `build_constraint` will have to build `z == x^2`; when dealing with the main constraint, it will only look like `1 + z <= y`, and `build_constraint` will have no idea that a new constraint was added. 

My idea would be to return a list of constraint objects along with `parse_code` and `build_code`, which would be passed on to the final `build_constraint`. ScalarConstraint and VectorConstraint would then also have a vector of MOI constraints that must be deleted at the same time.

blegat: About the constraint to remove, it seems to me that you need to create a custom set and a bridge for this set. In the bridge you can create new variables and constraints. When the constraint is deleted, the bridge is deleted too and in the deletion of the bridge, you can delete the added variables and constraints.

blegat: Traversing the expressions has no runtime cost, the macros are evaluated during compilation.

dourouc05: What bridge would be used for `1 + f(x) <= y`? What I'm pointing is that this code is used *before* bridges: a bridge can be used to map a constraint `z == x^2` to something the solver understands, but not for a constraint `1 + f(x) <= y` that it does not understand (there is no way of having a scalar like `1 + f(x)` right now in MOI). This code allows to add new constraints to get from a nice user-facing syntax to MOI-level constraints, which can then be bridged depending on the solver abilities. 

Take `1 + count(x .== 4) <= y`, for instance. This code allows to create two MOI constraints: `z = count(x .== 4)` and `1 + z <= y`, i.e. `MOI.add_constraint(model, [z, x], CP.Count(4))` and `MOI.add_constraint(model, z - y, MOI.LessThan(1.0))`. The first constraint can be bridged to provide a MILP formulation, for instance. However, MOI does not accept something like `MOI.add_constraint(model, 1 + count(x .== 4), MOI.LessThan(4.0))`. As I understand it, rewriting/lowering the constraint to MOI level (a "thin" abstraction on top of solvers) is JuMP's responsibility, as what's done for quadratic constraints (rewritten as SOCP when possible: https://www.juliaopt.org/JuMP.jl/stable/constraints/#Quadratic-constraints-1). 

It looks like compile-time performance is still a goal, as per the benchmarks of https://mlubin.github.io/pdf/jump-sirev.pdf. Or things did evolve since then? But still, I don't think performance issues should be discussed now, at least before any serious benchmarks occur and shows this part to be a significant bottleneck.

dourouc05:
I cannot yet remove the strange code for MutableArithmetics (https://github.com/JuliaOpt/JuMP.jl/pull/2229/files#diff-0033f9de61e25a98b730785191eb2c68R190-R208) due to this error: 

```
MutableArithmetics.Zero (Issue #2187): Error During Test at C:\Users\Thibaut\.julia\dev\JuMP\test\macros.jl:50
  ArgumentError: reducing over an empty collection is not allowed
  Stacktrace:
   [1] _empty_reduce_error() at .\reduce.jl:295
   [2] mapreduce_empty(::Function, ::Base.BottomRF{typeof(Base.add_sum)}, ::Type{T} where T) at .\reduce.jl:334
   [3] reduce_empty(::Base.MappingRF{var"#3#5",Base.BottomRF{typeof(Base.add_sum)}}, ::Type{T} where T) at .\reduce.jl:321
   [4] reduce_empty_iter at .\reduce.jl:347 [inlined]
   [5] reduce_empty_iter at .\reduce.jl:346 [inlined]
   [6] foldl_impl at .\reduce.jl:46 [inlined]
   [7] mapfoldl_impl at .\reduce.jl:41 [inlined]
   [8] #mapfoldl#189 at .\reduce.jl:157 [inlined]
   [9] mapfoldl at .\reduce.jl:157 [inlined]
   [10] #mapreduce#193 at .\reduce.jl:283 [inlined]
   [11] mapreduce at .\reduce.jl:283 [inlined]
   [12] sum at .\reduce.jl:486 [inlined]
   [13] sum(::Base.Generator{UnitRange{Int64},var"#3#5"}) at .\reduce.jl:503
   [14] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\src\macros.jl:192
   [15] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\test\macros.jl:52
   [16] top-level scope at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Test\src\Test.jl:1113
   [17] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\test\macros.jl:51
   [18] include(::String) at .\client.jl:439
   [19] top-level scope at C:\Users\Thibaut\.julia\dev\JuMP\test\runtests.jl:23
   [20] include(::String) at .\client.jl:439
   [21] top-level scope at none:6
   [22] eval(::Module, ::Any) at .\boot.jl:331
   [23] exec_options(::Base.JLOptions) at .\client.jl:264
   [24] _start() at .\client.jl:484
```

However, I don't have this one with https://github.com/JuliaOpt/JuMP.jl/compare/master...dourouc05:dourouc05-rewrite-complex…

PR #2231, opened 2020-04-25
callback value for expr

matbesancon:
Given that the necessary information is available, it would be make sense to access the value of an expression, through the evaluation of the composing variables, in a callback

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -32,28 +32,15 @@ function callback_value(cb_data, x::VariableRef)
 end
 
 """
-    callback_value(cb_data, expr::GenericAffExpr)
+    callback_value(cb_data, expr::Union{GenericAffExpr, GenericQuadExpr})
 
-Return the primal solution of an affine expression inside a callback by getting
+Return the primal solution of an affine or quadratic expression inside a callback by getting
 the value for each variable appearing in the expression.
 
 `cb_data` is the argument to the callback function, and the type is dependent on
 the solver.
 """
-function callback_value(cb_data, expr::GenericAffExpr)
-    return value(expr, v -> callback_value(cb_data, v))
-end
-
-"""
-    callback_value(cb_data, expr::GenericQuadExpr)
-
-Return the primal solution of a quadratic expression inside a callback by getting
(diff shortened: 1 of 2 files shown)


src/callbacks.jl
```diff
@@ -31,6 +31,32 @@ function callback_value(cb_data, x::VariableRef)
...
+    callback_value(cb_data, expr::GenericAffExpr)
+
+Return the primal solution of an affine expression inside a callback by getting
+the value for each variable appearing in the expression.
+
+`cb_data` is the argument to the callback function, and the type is dependent on
+the solver.
+"""
+function callback_value(cb_data, expr::GenericAffExpr)
+    return expr.constant + sum(callback_value(cb_data, var) * coeff for (var, coeff) in expr.terms)
```
blegat, line 44: You can use `value(expr, v - > callbak_value(cb_data, v))`

matbesancon: done thanks. For a future version of JuMP, value should take the function as first argument, it's more idiomatic and allows for the do block syntax

matbesancon:
one Format test is failing locally for me, no idea how it is linked to the changed code base

matbesancon:
ok apparently not failing on travis :man_shrugging:

src/callbacks.jl
```diff
@@ -31,6 +31,32 @@ function callback_value(cb_data, x::VariableRef)
...
+"""
+    callback_value(cb_data, expr::GenericAffExpr)
+
+Return the primal solution of an affine expression inside a callback by getting
+the value for each variable appearing in the expression.
+
+`cb_data` is the argument to the callback function, and the type is dependent on
+the solver.
+"""
+function callback_value(cb_data, expr::GenericAffExpr)
```
blegat, line 43: We can merge the two methods since they are identical

matbesancon: yup done

src/callbacks.jl
blegat: Good point, we can open an issue with the breaking tag to remind us of this.

blegat:
It might be because you need to update MutableArithmetics locally ?

PR #2232, opened 2020-04-29
Remove symbol in object dictionary with deletion of object

blegat:
~Currently failing because `hash` does not work on `DenseAxisArray`.~

Closes https://github.com/JuliaOpt/JuMP.jl/issues/1956

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -177,8 +177,10 @@ mutable struct Model <: AbstractModel
     nlp_data
     # Dictionary from variable and constraint names to objects.
     obj_dict::Dict{Symbol, Any}
-    # Dictionary from object to the `Symbol` at which they are registered in `obj_dict`.
-    registered_symbol::Dict{Any, Symbol}
+    # Dictionary from object id to the `Symbol` at which they are registered in `obj_dict`.
+    # They key is the object id (obtained with `objectid`) as hashing large containers of
+    # variables can be computationally expensive.
+    registered_symbol::Dict{UInt, Symbol}
     # Number of times we add large expressions. Incremented and checked by
     # the `operator_warn` method.
     operator_counter::Int
@@ -441,8 +443,30 @@ function num_nl_constraints(model::Model)
     return model.nlp_data !== nothing ? length(model.nlp_data.nlconstr) : 0
 end
 
-# TODO(IainNZ): Document these too.
+"""
+    object_dictionary(model::Model)
+
+Returns a dictionary mapping symbols to objects. This dictionary is used
+to retrieve the object corresponding to the symbol `symbol::Symbol` in
+`model[symbol]`. Objects are registered to a specific symbol in the macros.
+For instance, `@variable(model, x[1:2, 1:2])` registeres the array of variables
+`x` to the symbol `:x`.
+"""
 object_dictionary(model::Model) = model.obj_dict
+
+"""
+    registered_symbol(model::Model)
+
+Returns a dictionary mapping object ids (obtained with `objectid`) to
+symbols. This is the reverse mapping of the dictionary obtained
+with [`object_dictionary`](@ref). It is used to delete the `symbol => object`
+pair from the object dictionary when `object` is deleted with
+`delete(model, object)`.
+"""
+registered_symbol(model::Model) = model.registered_symbol
+# TODO To avoid breaking `AbstractModel`'s we have this fallback,
+#      that should be removed in JuMP v0.22
+registered_symbol(model::AbstractModel) = nothing
 
 """
     termination_status(model::Model)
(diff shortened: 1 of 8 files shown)


src/JuMP.jl
```diff
@@ -176,6 +176,10 @@ mutable struct Model <: AbstractModel
     nlp_data
     # Dictionary from variable and constraint names to objects.
     obj_dict::Dict{Symbol, Any}
+    # Dictionary from object id to the `Symbol` at which they are registered in `obj_dict`.
+    # They key is the object id (obtained with `objectid`) as hashing large containers of
+    # variables can be computationally expensive.
+    registered_symbol::Dict{UInt, Symbol}
```
mlubin, line 182: Base.IdDict appears to be specialized for this purpose

src/JuMP.jl
```diff
@@ -942,15 +969,39 @@ function Base.getindex(m::JuMP.AbstractModel, name::Symbol)
...
-    Base.setindex!(m::JuMP.Model, value, name::Symbol)
+    Base.setindex!(model::JuMP.Model, value, name::Symbol)
 
-stores the object `value` in the model `m` using so that it can be accessed via `getindex`.  Can be called with `[]` syntax.
+stores the object `value` in the model `model` using so that it can be accessed
+via `getindex`.  Can be called with `[]` syntax.
 """
-function Base.setindex!(m::JuMP.Model, value, name::Symbol)
+function Base.setindex!(model::AbstractModel, value, name::Symbol)
     # if haskey(m.obj_dict, name)
```
mlubin, line 978: Remove dead commented code

blegat:
> How does this work for collections of variables, especially if you delete them one-by-one?

If you multiply each entry of a container separately then you can do `unregister(model, con)` to remove the entry associated to the container in the `object_dictionary`.

mlubin:
> If you multiply each entry of a container separately then you can do unregister(model, con) to remove the entry associated to the container in the object_dictionary.

What's the point of having an automatic unregister that only works in the corner case of scalar variables? What about making it always manual?

blegat:
> What's the point of having an automatic unregister that only works in the corner case of scalar variables? What about making it always manual?

The user could also delete the container as a whole instead of deleting each separate constraint.
Given the code
```julia
@constraint(model, con[...], ...)
for ...
    delete(model, con[...])
end
```
We can say "use `delete(model, con)`" instead of saying "use `unregister(model, con)`.

mlubin:
Do we currently have specialized delete method for every type of container? So `delete(model, con)` would work but `delete.(model, con)` wouldn't work? I'm not in favor of implementing magic if it's flaky and only works if you use it in a non-obvious way.

`unregister(model, :x)` is super obvious that it deletes the symbol from the model. Do we need to make it more complicated?

odow:
:+1: in favor of `unregister(model, :x)`. Or even `delete(model, :x)` which, in addition to unregistering, will delete any elements of the container.

blegat:
We could have both and `delete(model, name::Symbol) = unregister(model, name); delete(model, model[name])`.

odow:
Why have to ways to do the same thing? 

We don't want people to go
```Julia
model = Model()
@variable(model, x)
unregister(model, :x)
@variable(model, x)
```
Now how do I access the first `x`?

Better to have
```Julia
model = Model()
@variable(model, x)
delete!(model, :x)
@variable(model, x)
```

blegat:
Does `delete!(model, :x)` also delete the corresponding object? Then if it is a container, it amounts to defining `delete!` on all containers and it's not so different with what's implemented in this PR

odow:
Maybe I do mean just `unregister` then.

mlubin:
> it amounts to defining delete! on all containers and it's not so different with what's implemented in this PR

Sort of, but `delete!(model, :x)` doesn't have the same subtle issue as `delete!(model, x)` and `delete!.(model, x)`.

blegat:
@odow @mlubin The new version should address your comments :)

test/constraint.jl
```diff
@@ -122,6 +122,29 @@ function test_delete_constraints(ModelType, ::Any)
...
+    @test JuMP.is_valid(model, con_ref)
+    @test model[:con_ref] === con_ref
+    JuMP.delete(model, :con_ref)
+    @test !JuMP.is_valid(model, con_ref)
+    @test_throws KeyError(:con_ref) model[:con_ref]
+    @constraint(model, con_ref[i=2:3], 3x >= i)
+    for i in 2:3
+        @test JuMP.is_valid(model, con_ref[i])
+        con = constraint_object(con_ref[i])
+        @test isequal_canonical(jump_function(con), 3x)
```
mlubin, line 138: Why is this loop needed? Why are we worried about `constraint_object` returning the wrong value?

src/JuMP.jl
```diff
@@ -943,15 +953,63 @@ function Base.getindex(m::JuMP.AbstractModel, name::Symbol)
...
+
+stores the object `value` in the model `model` using so that it can be accessed
+via `getindex`.  Can be called with `[]` syntax.
+"""
+function Base.setindex!(model::AbstractModel, value, name::Symbol)
+    object_dictionary(model)[name] = value
+end
+
+"""
+    delete(model::JuMP.AbstractModel, name::Symbol)
```
mlubin, line 966: I'm a bit worried about confusion from the overloaded meaning of `delete`. There's a big difference between `delete(m, x)` and `delete(m, :x)` that's understated by the 1-character difference. I'd tend to prefer a different name like `unregister` for simply deleting the key from the object dictionary (but could be convinced otherwise).

blegat: To be sure we talk about the same thing, you mean `unregister` so that the user needs to do:
```julia
delete(m, x)
unregister(m, :x)
```

test/constraint.jl
blegat: It's more for `is_valid`

src/JuMP.jl
mlubin: Yes, I'm proposing `delete` and then an explicit `unregister`.

blegat: @odow has a counter argument against `unregister` in https://github.com/jump-dev/JuMP.jl/pull/2232#issuecomment-622011209

PR #2239, opened 2020-05-08
Optimized eval_constraint_jacobian_product

amontoison:
@mlubin

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -409,7 +409,7 @@ function MOI.initialize(d::NLPEvaluator, requested_features::Vector{Symbol})
 
     max_chunk = min(max_chunk, 10) # 10 is hardcoded upper bound to avoid excess memory allocation
 
-    if d.want_hess || want_hess_storage # storage for Hess or HessVec
+    if (:JacVec in requested_features) || d.want_hess || want_hess_storage # storage for Hess or HessVec or JacVec
         d.input_ϵ = Array{Float64}(undef,max_chunk*num_variables_)
         d.output_ϵ = Array{Float64}(undef,max_chunk*num_variables_)
         d.forward_storage_ϵ = Array{Float64}(undef,max_chunk*max_expr_length)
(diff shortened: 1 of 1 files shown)


mlubin:
That was fast! Do you have any benchmarks comparing the performance of the two implementations?

amontoison:
Yes, I tested `eval_constraint_jacobian_product` on all problems of `OptimizationProblems.jl` that have nonlinear constraints : https://gist.github.com/amontoison/08ed12cf2d21ed39855e0521ae36614a

We use less memory now but I dont understand why it's slower :thinking:

mlubin:
Are you sure they're not flipped? I'd expect this version to use more memory (see all the memory allocated in the "`# storage for Hess or HessVec`" block) and run faster.

amontoison:
The memory allocated in the `# storage for Hess or HessVec` block only concerns the `MOI.initialize` function and not `MOI.eval_constraint_jacobian_product`.

I ran again the benchmarks, separately this time and I have the same results.
JacVec branch -> https://gist.github.com/amontoison/f0a7c4560cc809e3520cf5041282df7c
master branch  -> https://gist.github.com/amontoison/0f6b1421aef3da7b65a817f7f122e640

amontoison:
Miles do you have an idea why the "optimized" eval_constraint_jacobian_product is slower?

mlubin:
Sorry, no, I took a brief look and didn't see anything obviously wrong. We'd need to do careful profiling to see what's taking up the time in the new code.

Is the constraint jacobian product a real bottleneck in your work?

amontoison:
It's not really a bottleneck for the moment, but I can take advantage of optimized jacobian-vector products in my work. Instead of using direct solvers for solving linear systems inside optimization methods I use Krylov solvers which only required operator-vector products.
For instance, I already use the hessian-vector product instead of forming explicitly the hessian. For large-scale optimization, the computation and the storage for hessian / jacobian could be problematic.

odow:
@amontoison what is the status of this? Should we close this PR and open an issue instead? Or you could add a TODO to the code as a comment with a description of what could be done to speed it up.

amontoison:
@odow 
I think it's still relevant to have a  `JacVec` operator. It requires less memory because we don't need to store the coefficients of the Jacobian. I can update the PR next week and try to understand why it's so slow.

amontoison:
Sorry @odow, I completely forgot this PR. I can add it in the new Nonlinear submodule if you want and close the PR.

odow:
Let's leave this open here for now. I'll attack it once we have MOI.Nonlinear ready.

odow:
Closing in favor of https://github.com/jump-dev/MathOptInterface.jl/issues/1903

PR #2241, opened 2020-05-12
Add `rewrite_call_expression` through `_rewrite_expr`.

dourouc05:
Excerpt from #2051. New take on #2229, much more general (and not much longer — unless you take https://github.com/JuliaOpt/JuMP.jl/pull/2229/commits/fdef385b6b5de779208124f991c9a221f74ae925, which unfortunately does not work…). 

I'd prefer to see this PR merged instead of #2229, due to the functionalities that are possible with this, but maybe this is not the best way to implement it… 

The general goal is to allow rewriting expressions at the JuMP level, only based on syntactic information, to provide more convenience for users. This only provides more points for extensions to plug into JuMP. 

More in details (based on the current developments of the CP sets): the modelling layer can decide to rewrite an expression like `y <= count(x .== 5)` as two constraints, a CP one (say `z == count(x .== 5))`, and a purely linear one (say `y <= z`), all of this at JuMP layer. These new constraints can then be bridged by MOI if needed. The constraints to model in the first hand (like `y <= count(x .== 5))`) should not have MOI equivalents (because that would create a very very large number of sets to support: instead of just Count, you would need LessThanCount, maybe AffineLessThanCount later on, and so on). 

This thus raises the issue of deleting the new constraints: as a user, it only makes sense to have a reference to the constraint `y <= count(x .== 5)`; to delete it, two constraints and a variable must be deleted. The existing system of bridges does not work, as it takes as input a MOI constraint (which, by design, the user-specified constraint have a low likelihood to be). Nothing is implemented regarding this. My idea is to have `_rewrite_expr` to return the new things that the function call created (variables, constraints), so that JuMP can get rid of all of it when deleting the constraint.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -169,8 +169,12 @@ _functionize(x) = x
 _functionize(::MutableArithmetics.Zero) = 0.0
 function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense::Val, lhs, rhs)
     parse_code_rhs, build_code_rhs, new_rhs = :(), :(), rhs
+    parse_code_lhs, build_code_lhs, new_lhs = :(), :(), lhs
     if isexpr(rhs, :call) && applicable(rewrite_call_expression, Val(rhs.args[1]), rhs.args[2:end]...)
         parse_code_rhs, build_code_rhs, new_rhs = rewrite_call_expression(_error, Val(rhs.args[1]), rhs.args[2:end]...)
+    end
+    if isexpr(lhs, :call) && applicable(rewrite_call_expression, Val(lhs.args[1]), lhs.args[2:end]...)
+        parse_code_lhs, build_code_lhs, new_lhs = rewrite_call_expression(_error, Val(lhs.args[1]), lhs.args[2:end]...)
     end
 
     # Simple comparison - move everything to the LHS.
@@ -180,22 +184,29 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense
     # `GenericAffExpr` or a `GenericQuadExpr`, and not a `VariableRef` as would be the case
     # without `_functionize`.
 
-    if rhs == new_rhs
     # TODO: bug in MutableArithmetics? The returned code does not find $new_rhs (ERROR: UndefVarError: #642###976 not defined) when only using the first code path.
+    if rhs == new_rhs && lhs == new_lhs
         if vectorized
             func = :($lhs .- $rhs)
         else
             func = :($lhs - $rhs)
         end
         variable, parse_code = _MA.rewrite(func)
-    else
+    elseif rhs != new_rhs && lhs == new_lhs
         variable, parse_code = _MA.rewrite(lhs)
         parse_code = :($parse_code; $variable = _MA.mutable_operate!(-, convert(AffExpr, $variable), $new_rhs))
+    elseif lhs != new_lhs && rhs == new_rhs
+        variable, parse_code = _MA.rewrite(rhs)
+        parse_code = :($parse_code; $variable = _MA.mutable_operate!(-, $new_lhs, convert(AffExpr, $variable)))
+    else
+        @assert rhs != new_rhs
+        @assert lhs != new_lhs
+        error("Not yet implemented: expression rewriting in both members")
     end
 
     set = sense_to_set(_error, sense)
-    return :($parse_code_rhs; $parse_code), :($build_code_rhs; $build_code)
     build_code = _build_call(_error, vectorized, :(_functionize($variable)), set)
+    return :($parse_code_lhs; $parse_code_rhs; $parse_code), :($build_code_lhs; $build_code_rhs; $build_code)
 end
 
 function parse_constraint_expr(_error::Function, expr::Expr)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -159,7 +159,15 @@ function parse_one_operator_constraint(_error::Function, args...)
     _unknown_constraint_expr(_error)
 end
 
+# To be defined if such a function must be rewritten; otherwise,
+# rewrite_call_expression will never be called.
 expression_to_rewrite(head::Val{F}, args...) where F = false
+
+# When `expression_to_rewrite` called with the same arguments save the first one,
+# returns three things:
+# - code for parsing (which must be called first), if needed; otherwise, :().
+# - code for building the required constraints, if needed; otherwise, :().
+# - the symbol of the variable that replaces the expression (<: VariableRef).
 function rewrite_call_expression(_error::Function, head::Val{F}, args...) where F
     _error("function $F not implemented.")
 end
@@ -186,6 +194,7 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense
     # without `_functionize`.
 
     # TODO: bug in MutableArithmetics? The returned code does not find $new_rhs (ERROR: UndefVarError: #642###976 not defined) when only using the first code path.
+    parse_code, variable = nothing, nothing
     if rhs == new_rhs && lhs == new_lhs
         if vectorized
             func = :($lhs .- $rhs)
@@ -195,14 +204,15 @@ function parse_one_operator_constraint(_error::Function, vectorized::Bool, sense
         variable, parse_code = _MA.rewrite(func)
     elseif rhs != new_rhs && lhs == new_lhs
         variable, parse_code = _MA.rewrite(lhs)
-        parse_code = :($parse_code; $variable = _MA.mutable_operate!(-, convert(AffExpr, $variable), $new_rhs))
+        parse_code = :($parse_code; $variable = _MA.operate!(-, $variable, $new_rhs))
     elseif lhs != new_lhs && rhs == new_rhs
         variable, parse_code = _MA.rewrite(rhs)
-        parse_code = :($parse_code; $variable = _MA.mutable_operate!(-, convert(AffExpr, $new_lhs), $variable))
+        parse_code = :($parse_code; $variable = _MA.operate!(-, $variable, $new_lhs))
     else
         @assert rhs != new_rhs
         @assert lhs != new_lhs
-        error("Not yet implemented: expression rewriting in both members")
+        variable = gensym()
+        parse_code = :($parse_code; $variable = _MA.operate!(-, $new_lhs, $new_rhs))
     end
 
     set = sense_to_set(_error, sense)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -159,25 +159,43 @@ function parse_one_operator_constraint(_error::Function, args...)
     _unknown_constraint_expr(_error)
 end
 
+function rewrite_call_expression(_error::Function, head::Val{F}, args...) where F
+    _error("function $F not implemented.")
+end
+
 _functionize(v::VariableRef) = convert(AffExpr, v)
 _functionize(v::AbstractArray{VariableRef}) = _functionize.(v)
 _functionize(x) = x
 _functionize(::MutableArithmetics.Zero) = 0.0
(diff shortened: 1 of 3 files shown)


dourouc05:
@blegat @odow @mlubin Is this approach viable? 

(Thinking about it, I suppose `_rewrite_expr` should lose its starting underscore, as it would be a part of the public interface.)

dourouc05:
I just added a few tests for binary operators.

src/macros.jl
```diff
@@ -159,25 +159,56 @@ function parse_one_operator_constraint(_error::Function, args...)
...
+_rewrite_expr(_error::Function, ::Val{:call}, ::Val{OP}, args...) where OP = :(), :(), Expr(:call, OP, args...)
+_rewrite_expr(_error::Function, ::Val{HEAD}, args...) where HEAD = :(), :(), Expr(HEAD, args...)
+_rewrite_expr(_error::Function, e::Symbol) = :(), :(), e
+_rewrite_expr(_error::Function, e::Number) = :(), :(), e
+
+_rewrite_expr(_error::Function, e::Expr) = _rewrite_expr(_error, Val(e.head), e.args...)
+_rewrite_expr(_error::Function, head::Val{:call}, args...) =
+    _rewrite_expr(_error, Val(:call), Val(args[1]), args[2:end]...)
+
+function _rewrite_expr(_error::Function, ::Val{:call}, op::Union{Val{:+}, Val{:-}, Val{:*}, Val{:/}}, args...)
```
blegat, line 175: Why have a method with `OP` above which does not rewrite recursively and this one which rewrite recursively ?

dourouc05: The previous one is a fallback for unimplemented operators. I don't know if it's the most sensible default: I try to enable this code as rarely as possible, to avoid breaking things.

dourouc05:
@odow @blegat Does this approach make sense? In other words, might the ideas behind this PR get merged? What do you think of the concept of one JuMP-level constraint mapped to several MOI constraints (say, `CompoundConstraint`)?

dourouc05:
I did a little benchmark on this (https://github.com/dourouc05/JuMP.jl/blob/dourouc05-rewrite-complex/benchmark/parse_constraints.jl). Indeed, without writing complex things, this code may make parsing the constraints twice as slow. 

Before: 

```
julia> include("C:\\Users\\Thibaut\\.julia\\dev\\JuMP\\benchmark\\parse_constraints.jl")
(1/2) benchmarking "model0"...
done (took 5.4042206 seconds)
(2/2) benchmarking "model1"...
done (took 5.480212001 seconds)
2-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "model0" => Trial(28.958 ms)
  "model1" => Trial(27.174 ms)
```

With this PR:

```
julia> include("C:\\Users\\Thibaut\\.julia\\dev\\JuMP\\benchmark\\parse_constraints.jl")
(1/2) benchmarking "model0"...
done (took 5.499017699 seconds)
(2/2) benchmarking "model1"...
done (took 5.5020427 seconds)
2-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "model0" => Trial(29.167 ms)
  "model1" => Trial(55.927 ms)
```

While you don't look opposed to the idea, I guess the implementation is not performant enough. 

I'm not sure I get the details of the code you point at. You are generating "fast" code to parse the head of an expression, right? Plus using `Object` so that Julia doesn't generate too much code for different types?

blegat:
> What do you think of the concept of one JuMP-level constraint mapped to several MOI constraints (say, `CompoundConstraint`)?

You could generate a single constraint that is bridged into several constraints. I'd prefer not having constraint returning several MOI constraints

dourouc05:
> You could generate a single constraint that is bridged into several constraints. I'd prefer not having constraint returning several MOI constraints

Which "single" constraint? Does MOI need a way to represent constraints like `count(x .== 0) <= count(y .<= z) + minimum(abs.(x))`? That would require adding `count` as a new type of function, along with `minimum` and anything that might be used in an expression, but also a way to combine all these (at least in an affine way, but probably also nonlinear). It would make MOI's CP extension much much more complex. The solver interface would have more information about the constraints and could potentially use it (some CP solvers have a way to represent `x <= count(sth)` directly, but not all of them… and none can really gulp down more complex things), but implementing such an interface would be much more cumbersome (all solver interfaces would need a way to parse these expressions). 

That's why I think it would make sense, at the modelling layer, to decompose complex constraints into bite-sized MOI constraints, at the lowest level of abstraction that makes sense based on the solver APIs (that would mean adding things like `x <= count(sth)` in the MOI extensions, but nothing more complicated).

odow:
> could be a solution to implement this pattern matching, with high performance, but no extension possibilities (which is the point, here).

Yes. This is the point I settled on. Create a giant `if` statement with the common ones first. Make it easy to extend if we add stuff to JuMP, but don't make it externally extensible.

dourouc05:
> Yes. This is the point I settled on. Create a giant `if` statement with the common ones first. Make it easy to extend if we add stuff to JuMP, but don't make it externally extensible.

That would defeat the whole point of this, which is to allow extensions of JuMP syntax. Or maybe there is a way to keep a fast path for the common case, and keep the flexibility for extensions (with a performance penalty)?

blegat:
It seems your use case is similar to disciplined convex reformulation. In JuMP, you need to write your convex model as constraints that are either `VectorOfVariables`-in-`Cone` or `VectorAffineFunction`-in-`Cone`.
However, if a nonlinear expression is disciplined convex (which is stronger than convex), there is an automatic way to rewrite it as constraints of the form just mentioned.
Convex.jl does this reformulation for instance.
The plan to bring this kind of reformulation in JuMP is to wait for https://github.com/jump-dev/MathOptInterface.jl/issues/846 and then to add bridges for nonlinear expression into convex constraints.
This is a bit different from other bridges as you cannot determine just from the type of the constraints whether you will be able to transform it into convex constraints and which constraints you are going to produce so we might need to adapt `LazyBridgeOptimizer` or apply this bridge with a `SingleBridgeOptimizer`.
You could have the same approach and bridge nonlinear expression into constraint programming sets in bridges instead of doing it during the parsing of the macros.

dourouc05:
> It seems your use case is similar to disciplined convex reformulation. 

Indeed, it's the same issue, as DCP may decompose a single expression in many constraints. For now, I can live without the possibility to remove CP constraints :)! 

(By the way, I didn't see any documents giving ideas on how to implement bridges on top of these nonlinear expressions, that's why I didn't think this issue was relevant.)

dourouc05:
@odow I've been benchmarking this code a tad more, and it looks like my previous figures where wrong (I was using Revise to test the latest version of the code). When restarting Julia between runs (i.e. with and without this PR), I cannot find a real performance difference. 

Before: 

```
julia> include("C:\\Users\\Thibaut\\.julia\\dev\\JuMP\\benchmark\\parse_constraints.jl")
(1/2) benchmarking "model0"...
done (took 5.4042206 seconds)
(2/2) benchmarking "model1"...
done (took 5.480212001 seconds)
2-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "model0" => Trial(29.564 ms)
  "model1" => Trial(53.224 ms)
```

After: 

```
julia> include("C:\\Users\\Thibaut\\.julia\\dev\\JuMP\\benchmark\\parse_constraints.jl")
(1/2) benchmarking "model0"...
done (took 5.4042206 seconds)
(2/2) benchmarking "model1"...
done (took 5.480212001 seconds)
2-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "model0" => Trial(28.770 ms)
  "model1" => Trial(53.410 ms)
```

I've run both quite a few times, and the figures never wander too far from these (give or take one or two milliseconds, for both). I've also tested to replace all these functions by a single one and a large `if`, but it gave no real difference.

blegat:
I think the difference will be mainly on the compilation since it will need to compile many more methods due to `Val`

dourouc05:
By the way, is performance really important? Most use cases I can think of should rather be implemented in nonlinear bridges rather than this way (everything, except overloading operators like `&&` between binary variables: maybe that should go into MutableArithmetics?). As I understand the new nonlinear implementation is still far away, it might be good to include this in the meantime.

dourouc05:
@blegat @odow @mlubin As this code is intended to provide extension points for only a limited amount of time (until MOI fully supports nonlinearity), maybe it's in a good-enough shape to be merged? (At least as a 'private' API, i.e. not exported and potentially undocumented either.)

blegat:
I'm a bit hesitant because it significantly increase compile time

dourouc05:
I made a few benchmarks on my own. 

JuMP 0.21.5: 

* compilation from scratch: 16.980 s
* precompiled (second `using`): 5.799 s
* .ji file: 1,518,406 bytes

```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.0 (2020-08-01)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @time using JuMP
 16.980351 seconds (10.54 M allocations: 620.875 MiB, 1.16% gc time)
```

```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.0 (2020-08-01)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @time using JuMP
  5.799019 seconds (9.72 M allocations: 579.062 MiB, 2.99% gc time)
```

This PR (not rebased on current JuMP):

* compilation from scratch: 17.092 s
* precompiled (second `using`): 5.337 s
* .ji file: 1,522,398 bytes

```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.0 (2020-08-01)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @time using JuMP
 17.092835 seconds (10.54 M allocations: 620.675 MiB, 1.11% gc time)
```

```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.0 (2020-08-01)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @time using JuMP
  5.337707 seconds (9.72 M allocations: 578.967 MiB, 3.79% gc time)
```

Details of my version of Julia: 

```
julia> versioninfo()
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```

The difference in timing is not that large (although it's far from a scientific comparison).

blegat:
We should also compare the compilation time of `@constraint` calls.

dourouc05:
I already did some: https://github.com/jump-dev/JuMP.jl/pull/2241#issuecomment-661345044

PR #2245, opened 2020-06-01
Add model_from

blegat:
One target use case would be to allow to easily create JuMP from matrix input with https://github.com/JuliaOpt/MatrixOptInterface.jl/tree/master/src, e.g. to replace MathProgBase's `linprog`.
For instance:
```julia
using JuMP, MatrixOptInterface
model = model_from(LPStandardForm(MOI.MAX_SENSE, c, A, b))
set_optimizer(model, GLPK.Optimizer)
optimize!(model)
sol = value.(all_variables(model))
```

### DIFF
--- a/src/file_formats.jl
+++ b/src/file_formats.jl
@@ -60,6 +60,26 @@ function Base.write(
 end
 
 """
+    model_from(moi_model::MOI.ModelLike; kws...)
+
+Create a new JuMP model `model = Model(; kws...)` and copies `moi_model` to
+`model`. The function return `model`.
+
+## Examples
+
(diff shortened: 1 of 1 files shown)


blegat:
I thought it would be more complicated when I started this. It might obscure things more than help. Still I thought it might be worth discussing whether we want something like this as it might be a quite common operation.

odow:
This is either a documentation issue if `MOI.copy_to`, or a possible need for an exported `JuMP.copy_to`.

odow:
Can this be closed? I don't see the value in adding an extra function to save two lines.

PR #2249, opened 2020-06-08
update links following move to jump.dev

mlubin:


mlubin:
Not sure why travis stalled, but https://github.com/jump-dev/JuMP.jl/pull/2252 passed. Let's be on the lookout for CI issues.

PR #2251, opened 2020-06-08
Sensitivity part II

### LINKED ISSUE #2244: Improve performance of LP sensitivity summary
As raised on Discourse: https://discourse.julialang.org/t/performance-of-lp-rhs-perturbation-range/40551

The function for querying LP sensitivity are pretty slow, mainly because they compute the constraint matrix every call:
https://github.com/JuliaOpt/JuMP.jl/blob/7df1796d305aed88ec4f66ea52b8437fa4babe58/src/lp_sensitivity.jl#L190-L192

There is probably plenty of other low-hanging fruit we could improve.

Edit: in fact, it seems the best thing we could do is compute and return a dictionary with the perturbation range for every constraint.

odow:
Closes #2244

This PR is a rewrite of the sensitivity summary to avoid a lot of redundant computations and allocations.

I've tried to comment my understanding of the code, and compared against the solutions reported by Gurobi and Excel.

Points of contention are:
- What is the sensitivity of a RHS of an `Interval` constraint? I have left unimplemented.
- What is the objective sensitivity of a nonbasic variable with fixed bounds? Gurobi seems to report `+/-0.0`, where as Excel reports `+/-Inf`. The current behavior is `+/-Inf`.

I'm not sure I fully understand all of the edge cases, so it will be interesting to see what the coverage is.

cc @geoffleyland: Want to stress-test this?

### DIFF
--- a/src/lp_sensitivity2.jl
+++ b/src/lp_sensitivity2.jl
@@ -1,0 +1,446 @@
+#  Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import SparseArrays
+import LinearAlgebra
+
+"""
+    SensitivityReport
+
+See [`lp_sensitivity_report`](@ref).
+"""
+struct SensitivityReport
+    rhs::Dict{ConstraintRef,Tuple{Float64,Float64}}
+    objective::Dict{VariableRef,Tuple{Float64,Float64}}
+end
+
+Base.getindex(s::SensitivityReport, c::ConstraintRef) = s.rhs[c]
+Base.getindex(s::SensitivityReport, x::VariableRef) = s.objective[x]
+
+"""
+    lp_sensitivity_report(model::Model; atol::Float64 = 1e-8)::SensitivityReport
+
+Given a linear program `model` with a current optimal basis, return a
+[`SensitivityReport`](@ref) object, which maps:
+
+ - Every variable reference to a tuple `(d_lo, d_hi)::Tuple{Float64,Float64}`,
+   explaining how much the objective coefficient of the corresponding variable
+   can change by, such that the original basis remains optimal.
+ - Every constraint reference to a tuple `(d_lo, d_hi)::Tuple{Float64,Float64}`,
+   explaining how much the right-hand side of the corresponding constraint can
+   change by, such that the basis remains optimal.
+
+Both tuples are relative, rather than absolute. So given a objective coefficient
+of `1.0` and a tuple `(-0.5, 0.5)`, the objective coefficient can range
+between `1.0 - 0.5` an `1.0 + 0.5`.
+
+`atol` is the primal/dual optimality tolerance, and should match the tolerance
+of the solver used to compute the basis.
+
+Note: interval constraints are NOT supported.
+
+# Example
+
+    model = Model(GLPK.Optimizer)
+    @variable(model, -1 <= x <= 2)
+    @objective(model, Min, x)
+    optimize!(model)
+    report = lp_sensitivity_report(model; atol = 1e-7)
+    dx_lo, dx_hi = report[x]
+    println(
+        "The objective coefficient of `x` can decrease by \$dx_lo or " *
+        "increase by \$dx_hi."
+    )
+    c = LowerBoundRef(x)
+    dRHS_lo, dRHS_hi = report[c]
+    println(
+        "The lower bound of `x` can decrease by \$dRHS_lo or increase " *
+        "by \$dRHS_hi."
+    )
+"""
+function lp_sensitivity_report(model::Model; atol::Float64 = 1e-8)
+    if !_is_lp(model)
+        error(
+            "Unable to compute LP sensitivity because model is not a linear " *
+            "program (or it contains interval constraints).",
+        )
+    elseif !has_values(model)
+        error("Unable to compute LP sensitivity: no primal solution available.")
+    elseif !has_duals(model)
+        error("Unable to compute LP sensitivity: no dual solution available.")
+    end
+
+    std_form = _standard_form_matrix(model)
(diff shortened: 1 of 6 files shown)


joaquimg:
matrix building is really similar to what is done in:
https://github.com/JuliaOpt/MatrixOptInterface.jl

Matrix building (from MOI) will also be used in the differentiation GSOC.
More general version of that is replicated in solvers like SCS and ProxSDP.

I think the matrix builder should live in MOIU. Starting with LP, for this use, then do QP (for GSOC) and Conic (also fo GSOC, but potentially uses in solvers).

This could simplify some first version solver wrappers.
And could be used by students writing their first solvers.

odow:
:+1: to adding matrix-type support to MOI as a submodule.

src/lp_sensitivity2.jl
```diff
@@ -181,15 +173,14 @@ function lp_sensitivity(model::Model; atol::Float64 = 1e-6)
...
-                # - and nonbasic at the upper bound (switch if lower)
-                # If an even number of these switches have occured, then the
-                # ratio forms an upper bound for δ. Otherwise, it forms a lower
-                # bound.
+                # - and nonbasic at the lower bound (switch if upper)
+                # If an odd number of these switches is true, then the ratio
+                # forms an upper bound for δ. Otherwise, it forms a lower bound.
                 status = j <= n ? variable_status[j] : affine_status[j - n]
-                if iseven(
+                if isodd(
```
blegat, line 197: Looks like `xor(is_min, d[j][e_i] > atol, status == MOI.NONBASIC_AT_LOWER)`

odow: I always forget there is a n-ary version.

joaquimg:
I have seen that Xpress allows you to query rhs and obj sensitivities.
Also, Gurobi and Xpress allow you to pass a vector and they will multiply it by the inverse basis matrix.

odow:
Yeah some solvers do allow you to query the sensitivity. But I'm not sure how that plays with bridges, etc.

joaquimg:
Indeed, that is a bit of a pain

odow:
Anyone interested in reviewing this? If not, I'll merge in a few days. It's a significant improvement over what we have currently, which is unusable for anything but toy models.

src/lp_sensitivity2.jl
```diff
@@ -0,0 +1,428 @@
...
+            # (δ⁻, δ⁺) = (-πᵢ, ∞) because increasing the objective coefficient
+            # will only keep it at the bound.
+            report.objective[var] = (-π[i], Inf)
+        elseif basis.variables[i] == MOI.NONBASIC_AT_UPPER && !is_min
+            @assert π[i] > -atol
+            # We are maximizing and variable `i` is nonbasic at upper bound.
+            # (δ⁻, δ⁺) = (-πᵢ, ∞) because increasing the objective coefficient
+            # will only keep it at the bound.
+            report.objective[var] = (-π[i], Inf)
+        elseif (
```
geoffleyland, line 146: I don't think you need the parentheses around this condition?

src/lp_sensitivity2.jl
```diff
@@ -0,0 +1,428 @@
...
+bounds on `t` such that `x_new` maintains bound feasibility. That is, compute
+bounds on t such that:
+
+    l_B[j] <= x_B[j] + t * d_B[j] <= u_B[j].
+"""
+function _compute_rhs_range(d_B, x_B, l_B, u_B, atol)
+    t_lo, t_hi = -Inf, Inf
+    for j = 1:length(l_B)
+        if d_B[j] > atol
+            t_hi = min(t_hi, (u_B[j] - x_B[j]) / d_B[j])
```
geoffleyland, line 255: In other places in this function, you have `t_lo, t_hi`, but here you do `t_hi` first.  Also, the expressions in both branches are the same, so could compute them beforehand?

src/lp_sensitivity2.jl
odow: I ran the formatter instead.

src/lp_sensitivity2.jl
odow: I re-ordered the lines. But I didn't find computing the `t` any more readable. It just added another layer of in-direction.

src/lp_sensitivity2.jl
```diff
@@ -0,0 +1,429 @@
...
+"""
+    lp_sensitivity(model::Model; atol::Float64 = 1e-8)::SensitivityReport
+
+Given a linear program `model` containing a current optimal basis, return a
+[`SensitivityReport`](@ref) object, which maps:
+
+ - every constraint reference to a range over which the right-hand side of the
+    corresponding constraint can vary such that the basis remains optimal.
+    ```julia
+    model = lp_sensitivity(model)
```
joehuchette, line 31: Overwriting ``model`` seems incorrect.

src/lp_sensitivity2.jl
```diff
@@ -0,0 +1,429 @@
...
+    dRHS_lo, dRHS_hi = model[c]
```
joehuchette, line 32: What is ``c`` here? A ConstraintRef? (Apologies if I'm missing something in how the docs will be rendered)

src/lp_sensitivity2.jl
```diff
@@ -0,0 +1,429 @@
...
+            "program (or it contains interval constraints).",
+        )
+    elseif !has_values(model)
+        error("Unable to compute LP sensitivity: no primal solution available.")
+    elseif !has_duals(model)
+        error("Unable to compute LP sensitivity: no dual solution available.")
+    end
+
+    std_form = _standard_form_matrix(model)
+    basis = _standard_form_basis(model, std_form)
```
joehuchette, line 59: Where will this function crash in the case that a basis is not available?

src/lp_sensitivity2.jl
odow: Ooo. I don't know how I missed this one.

src/lp_sensitivity2.jl
odow: It throws an error when querying `ConstraintBasisStatus`.

odow:
Bump. Anyone still want to review this/objections to merging?

joaquimg (approved):

PR #2253, opened 2020-06-11
Explicit link to the documentation of MOI.FileFormats

dourouc05:


docs/src/solvers.md
```diff
@@ -125,6 +125,9 @@ time_limit_sec
 
 JuMP can write models to a variety of file-formats using [`write_to_file`](@ref)
 and [`Base.write`](@ref).
+
+The file formats are defined [within MathOptInterface in the FileFormats enumeration](https://jump.dev/MathOptInterface.jl/latest/apireference/index.html#MathOptInterface.FileFormats.FileFormat).
```
odow, line 129: Actually, we should probably link to a fixed version like https://jump.dev/MathOptInterface.jl/v0.9/apireference/#MathOptInterface.FileFormats.FileFormat

dourouc05: I decided against a version in the URL to avoid updating it at each version of JuMP. The ideal would be to generate the URL based on Project.toml, but I don't know how this is possible…

mlubin: It's better to have a link to old, relevant docs than a broken link.

dourouc05: Makes sense. I've modified this PR accordingly.

PR #2255, opened 2020-06-14
added note of GAMS solvers to installation doc

renkekuhlmann:
I updated the installation doc section about AMPL to also refer to the option of using GAMS. Because the AMPL and GAMS options are very similar I tried to cover this in the same paragraph. If you like me to write about GAMS in a separate paragraph, let me know. I haven't updated any solvers that are supported by GAMS and are not supported by JuMP (without GAMS.jl) in the README.md, like Antigone.

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -66,11 +66,13 @@ Where:
 -   SDP = Semidefinite programming
 -   MISDP = Mixed-integer semidefinite programming
 
-You may also use [AmplNLWriter](https://github.com/JuliaOpt/AmplNLWriter.jl) to
-access solvers that support the [nl format](https://en.wikipedia.org/wiki/Nl_(format)).
-Such solvers include [Bonmin](https://github.com/coin-or/Bonmin) and
-[Couenne](https://github.com/coin-or/Couenne). See a more complete list
-[here](https://ampl.com/products/solvers/all-solvers-for-ampl/).
+You may also use [AmplNLWriter](https://github.com/JuliaOpt/AmplNLWriter.jl) or
+[GAMS](https://github.com/GAMS-dev/gams.jl) to access additional solvers.
+Such solvers include for example [Antigone](https://www.gams.com/latest/docs/S_ANTIGONE.html),
+[Bonmin](https://github.com/coin-or/Bonmin), [CONOPT](http://www.conopt.com/) and
+[Couenne](https://github.com/coin-or/Couenne). See a more complete list of AMPL
+solvers [here](https://ampl.com/products/solvers/all-solvers-for-ampl/) and of
+GAMS solvers [here](https://www.gams.com/latest/docs/S_MAIN.html).
 
 To install Gurobi, for example, and use it with a JuMP model `model`, run:
 
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -66,13 +66,24 @@ Where:
 -   SDP = Semidefinite programming
 -   MISDP = Mixed-integer semidefinite programming
 
-You may also use [AmplNLWriter](https://github.com/JuliaOpt/AmplNLWriter.jl) or
-[GAMS](https://github.com/GAMS-dev/gams.jl) to access additional solvers.
-Such solvers include for example [Antigone](https://www.gams.com/latest/docs/S_ANTIGONE.html),
-[Bonmin](https://github.com/coin-or/Bonmin), [CONOPT](http://www.conopt.com/) and
-[Couenne](https://github.com/coin-or/Couenne). See a more complete list of AMPL
-solvers [here](https://ampl.com/products/solvers/all-solvers-for-ampl/) and of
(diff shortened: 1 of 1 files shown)


mlubin:
I think it makes sense to cover AMPL and GAMS in separate paragraphs, each with their own list of popular solvers (no need for you to add anything to the AMPL list). This will be helpful for people who want to use "solver X" in JuMP and don't necessarily know if "solver X" is accessible through GAMS or AMPL.

renkekuhlmann:
Here is a new suggestion. I find it a bit difficult to choose a list of "popular" GAMS solvers. Here is a suggestion of 10 solvers (mainly NLP, because there are already native Julia JuMP links for most LP/MIP solvers). I hope 10 is not too much. Alternatively, we could also use the list from [GAMS.jl](https://github.com/GAMS-dev/gams.jl), which should be complete (apart from GAMS solver tools or MPEC solvers that are currently not supported for Julia).

docs/src/installation.md
```diff
@@ -72,6 +72,19 @@ Such solvers include [Bonmin](https://github.com/coin-or/Bonmin) and
...
+[AlphaECP](https://www.gams.com/latest/docs/S_ALPHAECP.html), 
+[Antigone](https://www.gams.com/latest/docs/S_ANTIGONE.html), 
+[BARON](https://www.gams.com/latest/docs/S_BARON.html), 
+[CONOPT](https://www.gams.com/latest/docs/S_CONOPT.html), 
+[Couenne](https://www.gams.com/latest/docs/S_COUENNE.html), 
+[LocalSolver](https://www.gams.com/latest/docs/S_LOCALSOLVER.html), 
+[PATHNLP](https://www.gams.com/latest/docs/S_PATHNLP.html),
+[SHOT](https://www.gams.com/latest/docs/S_SHOT.html), 
+[SNOPT](https://www.gams.com/latest/docs/S_SNOPT.html), 
+[SoPlex](https://www.gams.com/latest/docs/S_SOPLEX.html).
```
dourouc05, line 85: Isn't SoPlex available through SCIP (https://github.com/SCIP-Interfaces/SCIP.jl)?

docs/src/installation.md
```diff
@@ -72,6 +72,19 @@ Such solvers include [Bonmin](https://github.com/coin-or/Bonmin) and
 [Couenne](https://github.com/coin-or/Couenne). See a more complete list
 [here](https://ampl.com/products/solvers/all-solvers-for-ampl/).
 
+Additionally, you may use [GAMS](https://github.com/GAMS-dev/gams.jl) to access additional solvers. Among them are 
+[AlphaECP](https://www.gams.com/latest/docs/S_ALPHAECP.html), 
+[Antigone](https://www.gams.com/latest/docs/S_ANTIGONE.html), 
+[BARON](https://www.gams.com/latest/docs/S_BARON.html), 
+[CONOPT](https://www.gams.com/latest/docs/S_CONOPT.html), 
+[Couenne](https://www.gams.com/latest/docs/S_COUENNE.html),
```
dourouc05, line 80: I suppose Couenne is better used through AmplNLWriter (https://github.com/jump-dev/AmplNLWriter.jl)?

odow: We should definitely recommend it through GAMS as well. The more the merrier.

mlubin: I don't see an issue with listing Couenne here, especially since the GAMS interface has commercial support whereas the AmplNLWriter interface does not.

docs/src/installation.md
```diff
@@ -72,6 +72,19 @@ Such solvers include [Bonmin](https://github.com/coin-or/Bonmin) and
...
+[Antigone](https://www.gams.com/latest/docs/S_ANTIGONE.html), 
+[BARON](https://www.gams.com/latest/docs/S_BARON.html), 
+[CONOPT](https://www.gams.com/latest/docs/S_CONOPT.html), 
+[Couenne](https://www.gams.com/latest/docs/S_COUENNE.html), 
+[LocalSolver](https://www.gams.com/latest/docs/S_LOCALSOLVER.html), 
+[PATHNLP](https://www.gams.com/latest/docs/S_PATHNLP.html),
+[SHOT](https://www.gams.com/latest/docs/S_SHOT.html), 
+[SNOPT](https://www.gams.com/latest/docs/S_SNOPT.html), 
+[SoPlex](https://www.gams.com/latest/docs/S_SOPLEX.html). 
+See a complete list [here](https://www.gams.com/latest/docs/S_MAIN.html).
```
mlubin, line 86: Could you add a note that GAMS is commercial software that requires a license? We make this distinction in the solver table as well.

docs/src/installation.md
renkekuhlmann: Probably, but I haven't checked https://github.com/SCIP-Interfaces/SCIP.jl. If you want me to remove SoPlex from the list, let me know.

odow: It's fine mentioning SoPlex here.

PR #2269, opened 2020-07-10
Reduce CI load by skipping Julia 1.1, 1.2, and 1.3

odow:
I think it is fine for us to only test the LTS release and the most current minor release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,10 @@ julia:
 # matrix:
 #     allow_failures:
 #         - julia: 0.7
+branches:
+  only:
+    - master
+    - /^release-.*$/
 notifications:
     email: false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,9 +7,9 @@ os:
           # with osx and linux should be similar.
 julia:
     - 1.0
-    - 1.1
-    - 1.2
-    - 1.3
+    # - 1.1 # Disable to speed up CI
```
blegat, line 10: Missing a `d` at the end of `Disable`

odow:
It also seems like travis builds both the PR and the push.

Should we add:
```
branches:
  only:
    - master
    - /^release-.*$/
```

PR #2275, opened 2020-07-11
Utility for relaxing integrality

### LINKED ISSUE #1611: Feature request: solve relaxation
Like 0.18's `relaxation=True` flag to `solve()`.

mlubin:
Closes #1611
CC @mtanneau

### DIFF
--- a/docs/src/variables.md
+++ b/docs/src/variables.md
@@ -448,13 +448,13 @@ julia> @variable(model, y[A], container=Array)
 JuMP now creates a vector of JuMP variables instead of a DenseAxisArray. Note
 that choosing an invalid container type will throw an error.
 
-## Integrality shortcuts
+## Integrality utilities
 
 Adding integrality constraints to a model such as `@constraint(model, x in MOI.ZeroOne())`
 and `@constraint(model, x in MOI.Integer())` is a common operation. Therefore,
 JuMP supports two shortcuts for adding such constraints.
 
-#### Binary (ZeroOne) constraints
+### Binary (ZeroOne) constraints
 
 Binary optimization variables are constrained to the set ``x \in \{0, 1\}``. (The
 `MOI.ZeroOne` set in MathOptInterface.) Binary optimization variables can be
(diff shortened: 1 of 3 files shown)


src/variables.jl
```diff
@@ -995,3 +995,104 @@ end
...
+        if info.integer
+            unset_integer(v)
+        elseif info.binary
+            unset_binary(v)
+            set_lower_bound(v, max(0.0, info.lower_bound))
+            set_upper_bound(v, min(1.0, info.upper_bound))
+        end
+    end
+    function unrelax()
+        for (v, info) in info_pre_relaxation
```
mtanneau, line 1078: Is there a risk that `info_pre_relaxation` gets garbage collected before `unrelax` is called?

src/variables.jl
```diff
@@ -995,3 +995,104 @@ end
...
+"""
+    relax_integrality(model::Model)
+
+Modifies `model` to "relax" all binary and integrality constraints on
+variables. Specifically,
+
+- Binary constraints are deleted, and variable bounds are tightened if
+  necessary to ensure the variable is constrained to the interval ``[0, 1]``.
+- Integrality constraints are deleted without modifying variable bounds.
+- Semi-continuous and semi-integer constraints are ignored (left in place).
```
mtanneau, line 1025: No error for semi-continuous, semi-integer, SOS constraints then?

docs/src/variables.md
```diff
@@ -510,6 +510,12 @@ julia> is_integer(x)
 false
 ```
 
+### Relaxing integrality
+
+The [`relax_integrality`](@ref) function relaxes all integrality constraints in
+the model, returning an function that can be called to undo the operation later
```
joehuchette, line 516: "a function"

src/variables.jl
mlubin: No, closures keep a reference to the variables they depend on: https://docs.julialang.org/en/v1/devdocs/functions/#Closures-1

docs/src/variables.md
mlubin: Fixed

src/variables.jl
mlubin: I added an error for semi-continuous and semi-integer because we might want to add support for those in the future (and removing an error isn't a breaking change). I think SOS, indicator and other constraints are clearly out of scope (the name of the function is `relax_integrality`) and should be ignored. We can't possibly enumerate all types of constraints that might cause the problem to have discrete aspects.

PR #2277, opened 2020-07-12
WIP: make tests run faster

odow:
This PR is an attempt to make the JuMP tests run faster.

The first commit is a general tidying commit:
- Removes unneeded imports
- Remove `DualNumbers` in favor of `ForwardDiff`
- Adds imports each test file so that it can be run independently of `runtests.jl`, or in an arbitrary order

The next commits refactor the worst-performing files to speed them up.

In every case, the overwhelming majority of the time is spent compiling methods instead of doing work.

For example, running `test/operator.jl` (slightly modified to call `TestOperator.runtests()` twice) on my machine yields:
```Julia
oscar@Oscars-MBP JuMP % ~/julia1.3 --project=. --color=yes test/operator.jl 
 31.475031 seconds (89.82 M allocations: 4.505 GiB, 5.54% gc time)
  0.025482 seconds (64.32 k allocations: 5.131 MiB)
```
Yes, that is a 1200 fold difference in run time the second time round 😢.
Running it with `-O0` drops the time further.
```julia
oscar@Oscars-MBP JuMP % ~/julia1.3 --project=. --color=yes -O0 test/operator.jl
 23.202877 seconds (89.82 M allocations: 4.505 GiB, 7.17% gc time)
  0.023707 seconds (64.97 k allocations: 5.151 MiB)
```

Current top-5 run times in seconds
| commit | ma | operator | file_formats | constraint | variable | TOTAL(5) |
| --: | --: | --: | --: | --: | --: | --: |
| start (s) | 200 |  76 | 64 | 59 | 33 | 432 |
| latest (s) | 101 | 40 | 35 | 51 |  34 | 261 |
|change | -50% | -47% | -45% | -13% |  +0% | -40% |

Latest `master` build of Travis:
image

Latest Travis build for this branch:
image

TODO before merging:
- [ ] Consider whether to cherry pick commits 2 to N into separate PRs
- [ ] Can we run travis with `-O0`?

PR #2279, opened 2020-07-13
Refactor files in test/ so they can be run independently

odow:
- Make it so individual test files can be run
- Time which tests take the most time
  - Currently, MutableArithmetics is the worst offender
- Remove unneeded test entries in Project.toml

Refactored out of #2277

### DIFF
--- a/test/constraint.jl
+++ b/test/constraint.jl
@@ -1,3 +1,13 @@
+using JuMP
+using LinearAlgebra
+using Test
+
+include(joinpath(@__DIR__, "utilities.jl"))
+
+@static if !(:JuMPExtension in names(Main))
+    include(joinpath(@__DIR__, "JuMPExtension.jl"))
+end
+
 function test_constraint_name(constraint, name, F::Type, S::Type)
     @test name == @inferred JuMP.name(constraint)
     model = constraint.model
--- a/test/derivatives.jl
+++ b/test/derivatives.jl
@@ -1,8 +1,11 @@
+using JuMP
 using JuMP._Derivatives
+using LinearAlgebra
 using Test
-using MathOptInterface
 
-struct ΦEvaluator <: MathOptInterface.AbstractNLPEvaluator
+const ForwardDiff = JuMP.ForwardDiff
+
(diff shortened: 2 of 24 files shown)


test/derivatives.jl
```diff
@@ -345,31 +343,34 @@ function dualforward(ex, x; ignore_nan=false)
...
 
     # compare with running dual numbers
-    forward_dual_storage = zeros(DualNumbers.Dual{Float64},length(nd))
-    partials_dual_storage = zeros(DualNumbers.Dual{Float64},length(nd))
-    output_dual_storage = zeros(DualNumbers.Dual{Float64},length(x))
-    reverse_dual_storage = zeros(DualNumbers.Dual{Float64},length(nd))
-    x_dual = [DualNumbers.Dual(x[i],1.0) for i in 1:length(x)]
+    _epsilon(x::ForwardDiff.Dual{Nothing, Float64, 1}) = x.partials[1]
+
+    forward_dual_storage = zeros(ForwardDiff.Dual{Nothing, Float64, 1},length(nd))
```
blegat, line 348: Great, it was always complaining that I didn't have `DualNumbers` because I was not running the tests with `] test`.

odow:
Objections to merging? Once this is in I can start teeing up the rest of the Test PRs.

test/constraint.jl
```diff
@@ -1,3 +1,13 @@
+using JuMP
+using LinearAlgebra
+using Test
+
+include(joinpath(@__DIR__, "utilities.jl"))
+
+@static if !(:JuMPExtension in names(Main))
```
mlubin, line 7: Why do we do this unusual check for JuMPExtension.jl but not utilities.jl?

odow: Utilities just loads some functions, whereas JuMPExtension is a module so we get annoying `WARNING: module JuMPExtension has been redefined` warnings.

The other PRs fix this by making each file a module. The module approach also prevents state leaking between files, and allows us to programmatically determine the tests to run within a particular file.

PR #2283, opened 2020-07-15
Fix failing tests on master

odow:
Previous two commits were on separate branches and I merged the
second without rebasing or re-runing the tests.


odow:
A good example of the need to isolate state between the test files.

PR #2284, opened 2020-07-15
Refactor test/mutable_arithmetics.jl

odow:
In addition, disable JuMPExtension tests. They double our testing time
without adding much in the way of extra coverage.

Extracted from #2277

### DIFF
--- a/test/mutable_arithmetics.jl
+++ b/test/mutable_arithmetics.jl
@@ -1,12 +1,13 @@
+module TestMutableArithmetics
+
 using LinearAlgebra
 using JuMP
 using Test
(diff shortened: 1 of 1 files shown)


odow:
@blegat are you okay dropping the JuMPExtension tests? Saves us nearly 2 minutes of test time.

blegat:
It's important to test that we use `AbstractVariableRef` here: https://github.com/jump-dev/JuMP.jl/blob/master/src/mutable_arithmetics.jl#L21
for instance. Maybe we could run them for `promote_operation` tests only.

Issue #1983, opened 2019-06-08
Print subexpressions in nonlinear models


mlubin:
Currently we don't expand subexpressions in the printout of nonlinear models, which makes it hard to debug them: https://github.com/JuliaOpt/JuMP.jl/issues/1981#issuecomment-499686720.

PR #2292, opened 2020-07-25
Apply JuliaFormatter to nlp.jl

mlubin:
I started to work on https://github.com/jump-dev/JuMP.jl/issues/1983 but the crazy formatting in this file was too distracting. This is a good trial run for applying JuliaFormatter to all of JuMP (I tried this and hit https://github.com/domluna/JuliaFormatter.jl/issues/264). Please check if anything looks off or inconsistent with the JuMP style guide.

I used this configuration:
```
format_file("nlp.jl", margin=80, always_for_in=true)
```

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -8,8 +8,11 @@ mutable struct _NonlinearExprData
     const_values::Vector{Float64}
 end
 
-function set_objective(m::Model, sense::MOI.OptimizationSense,
-                       ex::_NonlinearExprData)
+function set_objective(
+    m::Model,
+    sense::MOI.OptimizationSense,
+    ex::_NonlinearExprData,
+)
     _init_NLP(m)
     set_objective_sense(m, sense)
     m.nlp_data.nlobj = ex
@@ -44,7 +47,7 @@ function _sense(c::_NonlinearConstraint)
                 return :range
             end
         else
-                return :(>=)
+            return :(>=)
         end
     else #if c.lb == -Inf
         c.ub == Inf && error("'Free' constraint sense not supported")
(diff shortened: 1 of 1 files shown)


src/nlp.jl
```diff
@@ -321,10 +394,19 @@ function MOI.initialize(d::NLPEvaluator, requested_features::Vector{Symbol})
 
     num_variables_ = num_variables(d.m)
 
-    moi_index_to_consecutive_index = Dict(moi_index => consecutive_index for (consecutive_index, moi_index) in enumerate(MOI.get(d.m, MOI.ListOfVariableIndices())))
+    moi_index_to_consecutive_index = Dict(
+        moi_index => consecutive_index
+        for
+        (consecutive_index, moi_index) in
+        enumerate(MOI.get(d.m, MOI.ListOfVariableIndices()))
```
odow, line 401: This one is a bit weird. Not sure what is better though

src/nlp.jl
```diff
@@ -730,41 +1017,83 @@ function MOI.eval_hessian_lagrangian(
...
+                    _hessian_slice(
+                        d,
+                        ex,
+                        x,
+                        H,
+                        obj_factor,
+                        nzcount,
+                        recovery_tmp_storage,
+                        Val{1},
+                    )::Int
```
odow, line 1053: These ones are annoying. I have been formatting as
```Julia
nzthis = _hessian_slice(
    d,
    ex,
    x,
    H,
    obj_factor,
    nzcount,
    recovery_tmp_storage,
    Val{1},
)::Int
```

src/nlp.jl
```diff
@@ -1157,66 +1729,126 @@ function value(ex::NonlinearExpression; result::Int = 1)
...
     else
-        autodiff == false || Base.warn_once("autodiff=true ignored since gradient is already provided.")
-        m.nlp_data.largest_user_input_dimension = max(m.nlp_data.largest_user_input_dimension,dimension)
-        d = _UserFunctionEvaluator(x -> f(x...), (g,x)->∇f(g,x...), dimension)
-        _Derivatives.register_multivariate_operator!(m.nlp_data.user_operators, s, d)
+        autodiff == false ||
+            Base.warn_once("autodiff=true ignored since gradient is already provided.")
+        m.nlp_data.largest_user_input_dimension =
+            max(m.nlp_data.largest_user_input_dimension, dimension)
+        d = _UserFunctionEvaluator(
```
odow, line 1819: Here's a case where it didn't drop down. What gives?

src/nlp.jl
odow: Issue: https://github.com/domluna/JuliaFormatter.jl/issues/314

I'm not opposed to merging as-is and having this fixed in a future update of the formatter.

PR #2294, opened 2020-07-25
Print nonlinear subexpressions

### LINKED ISSUE #1983: Print subexpressions in nonlinear models

mlubin:
Closes #1983
CC @ccoffrin

### DIFF
--- a/test/print.jl
+++ b/test/print.jl
@@ -615,7 +615,8 @@ end
         model = Model()
         @variable(model, x)
         @NLobjective(model, Max, sin(x))
-        @NLconstraint(model, cos(x) == 0)
+        c = @NLexpression(model, cos(x))
+        @NLconstraint(model, c == 0)
 
         io_test(REPLMode, model, """
     A JuMP Model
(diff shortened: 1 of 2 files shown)


test/print.jl
```diff
@@ -631,12 +632,15 @@ end
         io_test(REPLMode, model, """
     Max sin(x)
     Subject to
-     cos(x) - 0.0 $eq 0
+     subexpression[1] - 0.0 $eq 0
+    With NL expressions
+     subexpression[1]: cos(x)
```
blegat, line 637: Why `:` and not ` =` ?

mlubin: I considered that but I thought it could be confused with equality constraints, especially on a large model where "With NL expressions" may not be visible on the screen.

blegat: That makes sense

Issue #1035, opened 2017-05-30
Ease the computation of IIS? 


dourouc05:
Many solvers present an option to compute an IIS in case of infeasibility (Gurobi, CPLEX, BARON at least, probably others that I don't know of). However, accessing this functionality from JuMP is a bit of a mess (like the example [`iis.jl`](https://github.com/JuliaOpt/JuMP.jl/blob/master/examples/iis.jl)). 

I think such functionality should be directly available from JuMP (when the underlying solver supports it), with a function like `computeIIS`, but I have seen no previous work in this direction. It could have two kinds of outputs: 

- either directly the IIS in the console (as in the example)
- or a new JuMP model with only the constraints and variables of the IIS (so that it could be exported as an LP file), without console output (other than what the solver does when computing the said IIS)

Which solution would be preferable? I think that the second is more general (you could get the first one by just printing the resulting model), although I am not sure having the whole model is really useful for the user (except probably that it would be consistent). 

Of course, I propose myself as a potential implementer of the needed changes (in JuMP, MathProgBase, and a few solvers).

mlubin:
For the record this is a duplicate of https://github.com/JuliaOpt/JuMP.jl/issues/235.

dourouc05:
In what form would you return this mapping? I can think of several ways: 

- a ´inIIS´ function that might be called on each and every variable and constraint (or array) — for a variable, it could also return, as second and third arguments, the upper and lower bounds —, after something like `computeIIS` has been called on the model. This would be very close to how Gurobi works. 
- `computeIIS` calls the solver and returns two vectors, one with the variables (tuple: variable, lower bound, upper bound), one with the constraints. This would be very close to how CPLEX works. 

What elements should be considered when making a choice here? Does one affect more performance than the other (is this kind of question relevant for IISes, which are not really used that often)? 

A bit of loud thinking about the implementation. 
- A removed bound would be represented as $\pm\infty$. 
- The data returned when the solver has found the IIS would be stored in new fields of Model (linconstrIis, quadconstrIis, sosconstrIis, socconstrIis, sdpconstrIis), arrays of Booleans. 
- Nonlinear models would not be supported at first. 

For MathProgBase, I would need define a few functions: one to start the computation of the IIS, one to get the UP/LB/IISness of a variable, IISness of a constraint (following the conventions of MathProgBase, which seem to be having functions that work on one item at a time for `add…` and `get…`). Would it be enough to add them like this?
[...]

PR #2300, opened 2020-07-30
Allow querying conflicts from within JuMP

dourouc05:
This is just to ease access to the conflict machinery in MOI, including documentation. Follows #1035.

By the way, in `docs/src/solutions.md`, on line 291 (313 after this PR), the docs for `MOI.optimize!` are called: shouldn't that be `JuMP.optimize!`?


docs/src/solutions.md
```diff
@@ -289,6 +311,7 @@ JuMP.dual
 JuMP.solve_time
 OptimizeNotCalled
 MOI.optimize!
```
blegat, line 313: Yes, this should be `JuMP.optimize!`

docs/src/solutions.md
```diff
@@ -249,6 +249,28 @@ lp_objective_perturbation_range
 lp_rhs_perturbation_range
 ```
 
+## Conflicts
+
+Sometimes, the model you input is infeasible, and some solvers can help you
+find the cause of this infeasibility by offering a conflict, i.e. a subset of
+the constraints that create this infeasibility.
+
+The function [`compute_conflict!`](@ref) is used to start the computation of
```
blegat, line 258: CI seems to be failing, maybe use `JuMP.compute_conflict!`

dourouc05: I just tried that, but it does not solve the problem… Also, it seems that the docs for `MOI.ConflictParticipationStatusCode` and `MOI.ConflictStatusCode` are not properly generated, unlike `MOI.TerminationStatusCode`. Is there any configuration missing?

dourouc05:
I tried many things to make Documenter work with the new function, I can't understand what causes the problem(s) and Documenter's output is very far from useful. I'm giving up.

odow:
I'll take a look

dourouc05:
Thanks! If you can find the root cause, I'm interested!

dourouc05:
@odow: did you have the opportunity to have a look at this? (Supposing you're not on holidays :)!)

odow:
Yes, you just need to merge this: https://github.com/dourouc05/JuMP.jl/pull/1

dourouc05:
Sorry, I didn't see that.

docs/src/solutions.md
```diff
@@ -249,6 +249,44 @@ lp_objective_perturbation_range
...
+
+For instance, this is how you can use this functionality: 
+
+```julia
+using JuMP
+model = Model()
+@variable(model, x >= 0)
+@constraint(model, c1, x >= 2)
+@constraint(model, c2, x <= 1)
+optimize!(model)
```
mlubin, line 275: Someone will inevitably copy-paste this code and open an issue or discourse post because `optimize!()` throws ``ERROR: NoOptimizer()``. Could you add a comment saying to set a solver that supports computing conflicts (e.g., Gurobi, ...)?

docs/src/solutions.md
```diff
@@ -249,6 +249,44 @@ lp_objective_perturbation_range
 lp_rhs_perturbation_range
 ```
 
+## Conflicts
+
+When the model you input is infeasible, some solvers can help you find the 
+cause of this infeasibility by offering a conflict, i.e., a subset of the 
+constraints that create this infeasibility.
```
mlubin, line 256: The standard name for this is irreducible inconsistent subsystem (IIS). Those key words should be used here in case someone searches the documentation for it.

docs/src/solutions.md
```diff
@@ -268,7 +269,7 @@ For instance, this is how you can use this functionality:
 
 ```julia
 using JuMP
-model = Model()
+model = Model() # With a solver that supports conflicts, like CPLEX or Gurobi
```
mlubin, line 272: I'd be more explicit, like "You must use a solver that supports conflicts like CPLEX or Gurobi."

dourouc05: Better this way? I also added the IIS keyword there.

PR #2302, opened 2020-08-02
[testing] Pukeko-ing, try2

IainNZ:
Since @odow functionized test/file_formats, decided to check if Pukeko just works, and it does.

I also functionized test/model (man those test sets went deep!) but it doesn't really seem to speed it up anyway:

```
# Test Summary: | Pass  Total
# Model         |  111    111
# Test Summary: | Pass  Total
# Model copy    |   59     59
# real    0m42.862s
# user    0m42.880s
# sys     0m0.460s
```
vs
```
# 32 test function(s) ran successfully in module TestModels (29.15 seconds)

#     real    0m40.666s
#     user    0m41.054s
#     sys     0m0.612s
```

I'm loathe to actually push for Pukeko.jl over Base.Tests without committing to actually supporting it, so feel free just to steal the functionized model tests for now.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,6 +26,7 @@ julia = "1"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Pukeko = "8cf91db1-0219-5b69-b551-5a0ff3c32cce"
 
 [targets]
-test = ["Test"]
+test = ["Test", "Pukeko"]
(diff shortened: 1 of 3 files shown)


odow:
Yeah, our issue is all compilation overhead, not the nested testsets per-se. Testing each file in parallel would be useful though.

IainNZ:
Should I pull just the function-ized model.jl out anyway?

odow:
> Should I pull just the function-ized model.jl out anyway?

Yes, please.

odow:
Closed by #2303

The current function/module approach to testing means it should be easy to change to Pukeko.jl if/when we want to switch to parallel testing.

PR #2315, opened 2020-08-20
Bump DataStructures version

jd-lara:


dourouc05:
As nothing changed in the code, it would be best to allow both 0.17 and 0.18, I suppose. (As long as the tests pass with both versions.)

jd-lara:
I am not sure what the rules are for the JuMP community but there are function name upgrades between versions that aren't compatible. For example, the function `top` changed to overload `Base.first`. 

This in my understanding is non backwards compatible between versions.

dourouc05:
As long as JuMP doesn't use these names (otherwise, your PR would be much longer), then this change is not relevant: this update might break compatibility in some cases, but not the interesting ones. As long as JuMP doesn't require any change, I don't think it's good behaviour to force users of JuMP to update: in some cases, they might even be unable to do so due to other dependencies… I suppose this is the understanding behind the large compatibility range for ForwardDiff.jl (JuMP announces compatibility for versions 0.5 to 0.10, i.e. more than three years of development with compatibility breakages).

jd-lara:
ok, I have added "0.17" to the compat.

dourouc05 (approved):

mlubin:
We do have to be a bit careful because CI won't tell us if the code stops
working with the older version of DataStructures. In this case it's
probably ok to use a wide range because the part of the code that uses
DataStructures doesn't change frequently.

andreasnoack:
Since DataStructures is so wide used, it would be helpful if you could make a patch release that allows for version 0.18.

Issue #693, opened 2016-03-07
An efficient way to verify the feasibility of provided solution


chenzjames:
Hi, 
I am looking for an automatic way to to verify whether a solution is feasible or not to a built up model, especially for model with a large number of constraints. 

I am thinking of defining a function whose input is a possible solution of decision variables, then copying all constraints defined in the built up model and associating each of them with a flag or boolean. The final output will be the union of all flags, which indicates the feasibility of the input solution.

Is there any other quick way? Or is it already provided by JuMP but I could not find it.

Thanks, all.

mlubin:
We don't currently have a tidy way to do this. Maybe most general current approach would be to use the MathProgBase interface to write a "solver" which collects all the problem data in matrix form (for LP/QP/conic models) and checks feasibility that way.

chenzjames:
Oh, I see. I will check with MathProgBase. Thank you.

gdowdy3:
@chenzjames, did you figure out how to implement an efficient check of feasibility?  If so, would you mind sharing some of the details?

chenzjames:
Sorry, not yet.

gdowdy3:
Alright, I've got a solution.  It may not be the most efficient computationally, but it is relatively easy to implement.

Suppose your model "m" has a vector of decision variables **x** defined by

    @variables(m, x[1:n])

and a bunch of constraints defined by lines of the form

    @constraint(m, ...)

Suppose you've got another vector **z** which you want to test for feasibility.  You can do this by:

(1) adding constraints fixing the values of **x**, like so

    for i = 1:n
        @constraint(m, x[i] == z[i])
    end

(2) setting the objective function to zero, like so
    
    @objective(m, Min, 0)   # it doesn't matter whether you use Min or Max

(3) and calling the solver, like so

    status = @solve(m)

If the status is :Optimal, then **z** is a feasible solution to your problem.  On the other hand, if the status is :Infeasible, then **z** is _not_ a feasible solution.

chenzjames:
@gdowdy3 I see. This should work well! Thank you very much!!! @gdowdy3

adowling2:
I put together a simple Julia package to help with this type of model diagnostics. Suggestions, comments and bug reports are welcome.
https://github.com/adowling2/DegeneracyHunter.jl

joaquimg:
@blegat maybe you can highlight here the internal MOI functions that could be used for computing value of constraints.
[...]

PR #2317, opened 2020-08-20
Add `value` for constraints with a custom `var_value` function

dourouc05:
This proposes a link between constraints and the `value` method that takes a `var_value` argument (https://github.com/jump-dev/JuMP.jl/blob/e2fd3ebd9e3f5e6913ba461c0bc54e5cc92f061e/src/aff_expr.jl#L149-L162). 

I was surprised that this was not supported; maybe there was a reason why? (It's the reason why I did not bother adding documentation or tests for now.)

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -161,6 +161,10 @@ function value(ex::GenericAffExpr{T, V}, var_value::Function) where {T, V}
     ret
 end
 
+function value(ex::Vector{GenericAffExpr{T, V}}, var_value::Function) where {T, V}
+    return value.(ex, var_value)
+end
+
 """
     constant(aff::GenericAffExpr{C, V})::C
 
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -161,10 +161,6 @@ function value(ex::GenericAffExpr{T, V}, var_value::Function) where {T, V}
     ret
 end
 
-function value(ex::Vector{GenericAffExpr{T, V}}, var_value::Function) where {T, V}
-    return value.(ex, var_value)
-end
-
 """
     constant(aff::GenericAffExpr{C, V})::C
 
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -588,6 +588,12 @@ function value(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)
     return reshape_vector(_constraint_primal(con_ref, result), con_ref.shape)
 end
 
+"""
+    value(con_ref::ConstraintRef, var_value::Function)
+
+Evaluate the primal value of the constraint `con_ref` using `var_value(v)` 
+as the value for each variable `v`.
+"""
 function value(con_ref::ConstraintRef{Model, <:_MOICON}, var_value::Function)
     f = jump_function(constraint_object(con_ref))
     return reshape_vector(value.(f, Ref(var_value)), con_ref.shape)
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -589,7 +589,8 @@ function value(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)
 end
 
 function value(con_ref::ConstraintRef{Model, <:_MOICON}, var_value::Function)
-    return value(jump_function(constraint_object(con_ref)), var_value)
+    f = jump_function(constraint_object(con_ref))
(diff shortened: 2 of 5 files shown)


odow:
I think the reason being it isn't that hard to do
```Julia
obj = constraint_object(c)
value(obj.f, var_value)
```

What were you trying to do that led to this?

dourouc05:
But why is it so complicated, while you can do `value(con_ref)` for the current solution? It's just inconsistent. (The solution you're showing is only easy when you are familiar with JuMP's internals, i.e. not for a typical user.)

I'm in the process of rebuilding a part of the constraint matrix for tight constraints. I think there are several blocks of that that might go into JuMP (`is_tight` and `coefficients`, for instance, possibly with other names). Would you welcome PRs for these?

odow:
>  It's just inconsistent

Fair point. 

Do you have a concrete example of what this would be used for?

dourouc05:
I have just added a few tests for this functionality, including support for vectors of expressions. Is the current implementation OK? I'm wondering about `value(ex::Vector{GenericAffExpr{T, V}}, var_value::Function)`, because I'm just broadcasting over the argument: may this have strange consequences?

dourouc05:
I also have created the package for the functions you would rather see them ripen before inclusion: https://github.com/dourouc05/JuMPFeasibility.jl (tests pass locally, I don't have the bandwidth right now to configure anything online…).

src/aff_expr.jl
```diff
@@ -161,6 +161,10 @@ function value(ex::GenericAffExpr{T, V}, var_value::Function) where {T, V}
     ret
 end
 
+function value(ex::Vector{GenericAffExpr{T, V}}, var_value::Function) where {T, V}
+    return value.(ex, var_value)
+end
```
odow, line 166: We shouldn't be defining functions like this.

src/constraints.jl
```diff
@@ -588,6 +588,10 @@ function value(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)
     return reshape_vector(_constraint_primal(con_ref, result), con_ref.shape)
 end
 
+function value(con_ref::ConstraintRef{Model, <:_MOICON}, var_value::Function)
+    return value(jump_function(constraint_object(con_ref)), var_value)
```
odow, line 592: Should this be something like
```Julia
f = jump_function(constraint_object(con_ref))
return reshape_vector(value.(f, Ref(var_value)), con_ref.shape)
```

odow:
> I think it would a rather simple extension to check for feasibility of a solution, which could solve #693, just by looping over the constraints

This is the idea of MathOptSetDistances, which generalizes your `is_satisfied` functions.

dourouc05:
I've updated the code in this PR, thanks for your comments!

src/constraints.jl
```diff
@@ -588,6 +588,11 @@ function value(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)
     return reshape_vector(_constraint_primal(con_ref, result), con_ref.shape)
 end
 
+function value(con_ref::ConstraintRef{Model, <:_MOICON}, var_value::Function)
```
odow, line 591: Docstring?

odow (commented):
Looks good, just needs a docstring.

dourouc05:
Better this way? I made a mix between `value(ex::GenericAffExpr{T, V}, var_value::Function)` and `value(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)`.

odow (approved):

src/constraints.jl
```diff
@@ -588,6 +588,17 @@ function value(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)
...
 
+"""
+    value(con_ref::ConstraintRef, var_value::Function)
+
+Evaluate the primal value of the constraint `con_ref` using `var_value(v)` 
+as the value for each variable `v`.
+"""
+function value(con_ref::ConstraintRef{Model, <:_MOICON}, var_value::Function)
+    f = jump_function(constraint_object(con_ref))
+    return reshape_vector(value.(f, Ref(var_value)), con_ref.shape)
```
blegat, line 599: I didn't know we needed a `Ref` with a function

blegat (approved):

src/constraints.jl
dourouc05: You're right, it's not required (i.e. tests still pass without it).

PR #2321, opened 2020-09-02
Check dimensions of functions and sets match

lkapelevich:
fixes https://github.com/jump-dev/MathOptInterface.jl/issues/1150

### DIFF
--- a/test/constraint.jl
+++ b/test/constraint.jl
@@ -547,10 +547,12 @@ function test_Model_all_constraints_vector(::Any, ::Any)
     csdp = @constraint(model, x in PSDCone())
     csoc = @constraint(model, [x[1], 1] in SecondOrderCone())
     csos = @constraint(model, [x[2]^2, 1] in MOI.SOS1([1.0, 2.0]))
-    @test_throws DimensionMismatch @constraint(model, [x[2]^2, 1] in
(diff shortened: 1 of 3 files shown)


test/constraint.jl
```diff
@@ -544,6 +547,10 @@ function test_Model_all_constraints_vector(::Any, ::Any)
     csdp = @constraint(model, x in PSDCone())
     csoc = @constraint(model, [x[1], 1] in SecondOrderCone())
     csos = @constraint(model, [x[2]^2, 1] in MOI.SOS1([1.0, 2.0]))
+    @test_throws DimensionMismatch @constraint(model, [x[2]^2, 1] in
+        MOI.SOS1([1.0, 2.0, 3.0]))
```
odow, line 551: Small thing, but:
```julia
@test_throws(
    DimensionMismatch,
    @constraint(model, [x[2]^2, 1] in MOI.SOS1([1.0, 2.0, 3.0]))
)
```
It's a bit weird breaking the line after `in`.

odow (approved):

odow:
I guess this is technically breaking, but it should only break incorrect code.

odow:
@blegat should we include this in 0.21.5?

blegat (approved):

PR #2324, opened 2020-09-03
Updates for MOI 0.9.15

odow:
The printing change is due to the way canonicalization happens. Terms are now ordered by variable creation time. To support older versions, I rearranged the constraint.

### DIFF
--- a/test/model.jl
+++ b/test/model.jl
@@ -107,7 +107,7 @@ function test_hygiene_variable()
     model_y = Model()
     @variable(model_y, y)
     err = JuMP.VariableNotOwned(y)
-    
+
     @test_throws err @constraint(model_x, y in MOI.EqualTo(1.0))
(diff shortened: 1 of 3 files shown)


test/model.jl
```diff
@@ -277,24 +277,21 @@ function test_bridges_add_before_con_set_optimizer()
...
-    @test_throws err @constraint(model, x in Nonnegative())
+    flag = true
+    try
+        @constraint(model, x in Nonnegative())
+        flag = false
+    catch err
+        @test err isa ErrorException
+        # Rather than test a particular bridging error, just check that the
+        # bridge explanation has been called. The sequence of errors could vary
+        # between MOI versions.
```
blegat, line 290: Completely agree, I was also waiting for the MOI error message to stabilize before interfacing it in JuMP which is silly. Your solution is definitely better

blegat (approved):
Well done for the reordering that supports both the old and the new version of MOI

Issue #2326, opened 2020-09-11
Make a new release?


DilumAluthge:
The latest registered release of JuMP is 0.21.3. JuMP 0.21.3 is not compatible with DataStructures 0.18.

However, the `master` branch of JuMP is compatible with DataStructures 0.18.

Therefore, would it be possible to make a new release of JuMP from the master branch?

Thanks so much!

Issue #320, opened 2014-11-24
Provide diagnostics mode to pinpoint source of NaNs


davidanthoff:
This continues a discussion started in #318.

[This](https://gist.github.com/davidanthoff/2e7a3ed62e20a6506dbb) model now solves, but we now know that there are iterations where it returns NaNs. I would like to be able to pinpoint the exact expression where the NaN is introduced, so that I can double check whether this is due to an error in my model, or whether I'm ok with this.

This is obviously not trivial at all, because the NaN could also be generated in a derivative, so in those cases ideally the diagnostic would say something like "the NaN was introduced in part x of the gradient of expression y" or something like that.

mlubin:
This discussion is very old and I don't expect any progress on providing NaN diagnostics unless someone is willing to do the work. I'm tempted to close.

PR #2327, opened 2020-09-11
Release v0.21.4

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/2326

We should get better at having more frequent patch releases.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,29 @@
 JuMP release notes
 ==================
 
+Version 0.21.4 (September 10, 2020)
+-----------------------------------
+
+- New features:
+  * Add debug info when adding unsupported constraints (#2203)
+  * Add `relax_integrality` for solving continuous relaxation (#2275)
+  * Allow querying constraint conflicts (#2300)
+- Bug fixes:
+  * Dispatch on `Real` for `MOI.submit` (#2319)
+  * Implement `copy` for `CustomSet` in tests (#2323)
+  * Don't export private macros (#2270)
+  * Fix invalid assertion in nonlinear (#2271)
+- Error if constraint has `NaN` right-hand side (#2272)
+- Improve speed of tests
+  * Lots of work modularizing files in `/test` (#2269, #2279, #2283, #2284, #2285, #2286, #2303, #2306, #2307)
+- Improve line numbers in macro error messages (#2276)
+- Print nonlinear subexpressions (#2294)
+- Various documentation updates (#2311, #2313, #2316)
+- Dependency updates:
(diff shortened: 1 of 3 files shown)


mlubin:
There's lots of good stuff in this release!
The PR numbers without links don't do much to help casual readers understand the changes. As a user, how does the text ``"#2300"`` help me understand how to use the new feature to query constraint conflicts?

odow:
I was just following the layout of previous NEWS entries.

Given that Tagbot now populates a news for each release (https://github.com/jump-dev/JuMP.jl/releases/tag/v0.21.3), I'm not sure these are required anymore. It would certainly allow us to set a faster release pace.

mlubin:
Maybe worth discussing during the next developer call. As is, I think a link to https://github.com/jump-dev/JuMP.jl/releases/tag/v0.21.4 would be more helpful than the current description with PR numbers listed as text.
Is updating NEWS really slowing down the release pace? A release like this deserves a discourse post (if not a blog post on the new jump blog), so even if we skipped NEWS we wouldn't be freed from the obligation of explaining what's in the release.

odow:
> Is updating NEWS really slowing down the release pace?

It depends if we want to move to a rolling release, where we tag a release every PR, or say once a week if there have been new PRs. At present it's a bit ad-hoc.

blegat:
I agree that the current NEWS is redundant with what TagBot does.

NEWS.md
```diff
@@ -1,6 +1,33 @@
...
+  * Don't export private macros
+  * Fix invalid assertion in nonlinear
+- Error if constraint has `NaN` right-hand side
+- Improve speed of tests
+  * Lots of work modularizing files in `/test`
+- Improve line numbers in macro error messages
+- Print nonlinear subexpressions
+- Various documentation updates
+- Dependency updates:
+  * Datastructures 0.18(#2315)
```
mlubin, line 27: Straggling PR number

mlubin (approved):

dourouc05:
The tagbot-generated content is not really redundant with `NEWS`: the issue and PR names are not always easy to understand for a newcomer. The list is also sometimes misleading: the issue #320 is just closed, it's far from solved, even though the generated list seems to indicate that the issue is no more present. 

Maybe it would help to have time-based releases, for instance every month? Maybe the person that merges code into JuMP would be responsible to update `NEWS` (if time allows) or to create an issue so that someone else knows that an update is required?

odow:
Agree on the time-based releases. We plan to discuss it on the next monthly call.

PR #2329, opened 2020-09-15
Fix DataStructues deprecation warnings

mlubin:
Should we be running CI with ``--depwarn=error``? That would have caught this.

We need a quick tag after this. NLP models are likely printing deprecation warnings on 0.21.4.


mlubin:
Opened https://github.com/jump-dev/JuMP.jl/issues/2331 to address the AppVeyor failure.

PR #2330, opened 2020-09-15
Update .travis.yml

odow:
Removes an unused key from Travis, with the side-effect of seeing if the documentation will build. 

x-ref https://github.com/jump-dev/jump-dev.github.io/pull/30#issuecomment-692346731

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ addons:
 # Override the default test script by making deprecation warnings an error.
 script:
   - julia --project -e 'import Pkg; Pkg.build();'
-  - julia --project --check-bounds=yes --depwarn=error -e 'Pkg.test(; coverage=true)'
+  - julia --project --check-bounds=yes --depwarn=error -e 'import Pkg; Pkg.test(; coverage=true)'
 jobs:
   include:
     - stage: "Documentation"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ os:
           # with osx and linux should be similar.
 julia:
     - 1.0
-    - 1.4
+    - 1  # The latest 1.x release
 # This is left as an example for the next big version switch.
 # matrix:
 #     allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ os:
 #   - osx # Disable to speed up CI. JuMP has no binary dependency so the result
           # with osx and linux should be similar.
 julia:
-    - 1.0
-    - 1  # The latest 1.x release
+    - 1.0  # The latest LTS 
+    - 1    # The latest minor release
 # This is left as an example for the next big version switch.
 # matrix:
 #     allow_failures:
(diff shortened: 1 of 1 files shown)


mlubin:
Here's how our config changed since the last release: https://github.com/jump-dev/JuMP.jl/commit/29f64d81071ad2169502c704e53ae1cfaaec0f3f. Maybe the "branches" section had some effect?

odow:
@mlubin I added depwarn. Let's see if it works.

mlubin:
Well, ``--depwarn=error` is going to fail before https://github.com/jump-dev/JuMP.jl/pull/2329 is merged.

odow:
> Well, `--depwarn=error` is going to fail before #2329 is merged.

Yes. If we added it to your PR, it would pass and we wouldn't know if it was the fix or whether the argument was ignored.

odow:
On a different note, I seem to have lost access to the JuMP-dev org on travis. I can't restart builds. On `travis-ci.org`, it says I don't belong to organizations, but on `travis-ci.com` it says I belong to JuMP-dev. But JuMP.jl is on `travis-ci.org`. Did you have any issues?

mlubin:
Oh, we also moved orgs since the last release. Maybe that caused the failure to push new docs.

No, I haven't had permissions issues with travis-ci.org. You are an owner of the JuMP-dev org so not sure what else to do.

odow:
Ah, I didn't read the error message closely enough. Although `MbedTLS` usually fails due to network issues, this time it is because of a deprecation warning:
https://travis-ci.org/github/jump-dev/JuMP.jl/jobs/727217390#L449
I guess `depwarn=error` isn't a good idea then.

mlubin:
`@fredrikekre` said on slack:
> You are not building on tags
> This filters them out: https://github.com/jump-dev/JuMP.jl/blob/a4592816023982d9979ebcd16bfc9e3bbac1f38e/.travis.yml#L18-L21
> That branch filter was not there on 0.21.3
> You need to add something like https://github.com/fredrikekre/Literate.jl/blob/5db8c22fcfe74470d1ab836d566afbc8a9e86d9f/.travis.yml#L13 to also run travis on the tags

I think we can manually push the docs for 0.21.4 if we want. But we could also fix this and then tag 0.21.5 to get https://github.com/jump-dev/JuMP.jl/pull/2329.

odow:
Agree on a `0.21.5`. We also need a `0.9.16` for MOI.

.travis.yml
```diff
@@ -19,14 +16,19 @@ branches:
   only:
     - master
     - /^release-.*$/
+    # Regex to allow building Documentation
```
mlubin, line 19: I'd word this as: "Matches tags (needed to push documentation)"

odow:
TODO: when merging, remove `[ci skip]` from commit message to ensure `master` build is triggered.

.travis.yml
```diff
@@ -19,14 +16,19 @@ branches:
   only:
     - master
     - /^release-.*$/
+    # Matches tags (needed to build documentation)
+    - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 notifications:
     email: false
-sudo: false
```
mlubin, line 24: I don't know what the effect of removing this is.

mlubin (approved):
LGTM assuming travis passes.

.travis.yml
odow: Travis told me it was okay
image

blegat: sudo: true used to be the default but that changed a few years ago so it's safe to remove

blegat (approved):

PR #2339, opened 2020-09-21
Allow copying models with a subset of constraints

dourouc05:
This brings https://github.com/jump-dev/MathOptInterface.jl/pull/1135 into JuMP. 

Next step for #1035.


dourouc05:
I plan to submit https://github.com/dourouc05/JuMP.jl/tree/iis-easy once this is merged, but I'm wondering how to test the new function…

dourouc05:
@mlubin @odow Is this good enough to be merged?

src/copy.jl
```diff
@@ -93,7 +99,10 @@ function copy_model(model::Model)
     reference_map = ReferenceMap(new_model, index_map)
 
     for (name, value) in object_dictionary(model)
-        new_model.obj_dict[name] = getindex.(reference_map, value)
+        if filter_constraints === nothing || 
+                (value isa ConstraintRef && filter_constraints(value))
```
mlubin, line 103: This doesn't catch the case where `value` is a container of ConstraintRefs (please add a test).

dourouc05: I don't get how you can generate such a case (even if it's mentioned in the docs: https://github.com/jump-dev/JuMP.jl/blob/1d36b2c3c1bb132dc2e97af3f2b17ef544eecbca/src/JuMP.jl#L932) from user code. I don't find any occurrence of this in the test suite either.

mlubin: `@constraint(model, constr[i = 1:2], x[i] <= y[i])` for example. See https://jump.dev/JuMP.jl/v0.21.5/constraints/#Constraint-containers-1.

dourouc05: Thanks for your pointer! I'm working on a patch for this.

mlubin:
Please fix the test failure: https://travis-ci.com/github/jump-dev/JuMP.jl/jobs/428464198#L501

```
test_copy_filter: Error During Test at /home/travis/build/jump-dev/JuMP.jl/test/model.jl:520
  MethodError: no method matching getindex(::ReferenceMap, ::Array{VariableRef,1})
  Closest candidates are:
    getindex(::ReferenceMap, !Matched::VariableRef) at /home/travis/build/jump-dev/JuMP.jl/src/copy.jl:27
    getindex(::ReferenceMap, !Matched::ConstraintRef) at /home/travis/build/jump-dev/JuMP.jl/src/copy.jl:31
  Stacktrace:
   [1] copy_model_style_mode(::Bool, ::MathOptInterface.Utilities.CachingOptimizerMode, ::Bool) at /home/travis/build/jump-dev/JuMP.jl/test/model.jl:453
   [2] test_copy_filter() at /home/travis/build/jump-dev/JuMP.jl/test/model.jl:512
```

docs/src/installation.md
```diff
@@ -40,6 +40,7 @@ is often more complex. We list below the currently available solvers.
 | [COSMO](https://github.com/oxfordcontrol/COSMO.jl)                             | [COSMO.jl](https://github.com/oxfordcontrol/COSMO.jl)                            | Apache   | LP, QP, SOCP, SDP                  |
 | [CPLEX](https://www.ibm.com/analytics/cplex-optimizer/)                        | [CPLEX.jl](https://github.com/jump-dev/CPLEX.jl)                                 | Comm.    | LP, MILP, SOCP, MISOCP             |
 | [CSDP](https://github.com/coin-or/Csdp)                                        | [CSDP.jl](https://github.com/jump-dev/CSDP.jl)                                   | EPL      | LP, SDP                            |
+| [EAGO](https://github.com/psorlab/EAGO.jl)                                     | [EAGO.jl](https://github.com/psorlab/EAGO.jl)                                    | CC BY-NC-SA | NLP                              |
```
mlubin, line 43: How did this diff get picked up?

src/copy.jl
```diff
@@ -36,14 +36,18 @@ Base.broadcastable(reference_map::ReferenceMap) = Ref(reference_map)
...
-    copy_model(model::Model)
+    copy_model(model::Model; filter_constraints::Union{Nothing, Function}=nothing)
 
 Return a copy of the model `model` and a [`ReferenceMap`](@ref) that can be used
 to obtain the variable and constraint reference of the new model corresponding
 to a given `model`'s reference. A [`Base.copy(::AbstractModel)`](@ref) method
 has also been implemented, it is similar to `copy_model` but does not return
 the reference map.
 
+If the `filter_constraints` arguments is given, only the constraints for which
```
mlubin, line 47: Typo: argument

test/model.jl
```diff
@@ -417,18 +417,24 @@ end
...
 
-function copy_model_style_mode(use_copy_model, caching_mode)
+function copy_model_style_mode(use_copy_model, caching_mode, filter_mode)
     model = Model(caching_mode = caching_mode)
     model.optimize_hook = dummy_optimizer_hook
     data = DummyExtensionData(model)
     model.ext[:dummy] = data
-    @variable(model, x ≥ 0, Bin)
-    @variable(model, y ≤ 1, Int)
+    @variable(model, x[i = 1:2] ≥ 0, Bin)
```
mlubin, line 425: Could you explicitly test all three types of possible containers? They're Arrays, DenseAxisArrays, and SparseAxisArrays: https://jump.dev/JuMP.jl/v0.21.5/constraints/#Constraint-containers-1.

dourouc05:
Normally, failures are now solved (I was testing the wrong version locally…). The other three comments are still outstanding.

test/model.jl
dourouc05: I just did that in the latest commit. I was expecting that I could just use an argument to `copy_model_style_mode`, but the `container` argument in both `@variable` and `@constraint` is not evaluated (i.e. the variable name is directly used as a container type, and not the contents of the variable).

src/copy.jl
```diff
@@ -138,6 +151,47 @@ function Base.copy(model::AbstractModel)
...
+two constraints `cref` and `cref2`. This model has no solution, as the two 
+constraints are mutually exclusive. The solver is asked to compute a conflict
+with [`compute_conflict!`](@ref). The parts of `model` participating in the 
+conflict are then copied into a model `new_model`.
+```julia
+model = Model()
+@variable(model, x)
+@constraint(model, cref, x >= 2)
+@constraint(model, cref2, x <= 1)
+
```
mlubin, line 184: Explain more explicitly what the blank line represents. I'm guessing it's for solving the model.

test/model.jl
```diff
@@ -492,6 +515,85 @@ function test_copy_direct_mode()
...
+    @test new_model.ext[:dummy].model === new_model
+    x1_new = reference_map[x[1]]
+    @test JuMP.owner_model(x1_new) === new_model
+    @test "x[1]" == @inferred JuMP.name(x1_new)
+    @test reference_map[JuMP.LowerBoundRef(x[1])] == @inferred JuMP.LowerBoundRef(x1_new)
+    
+    cref_1_new = reference_map[cref[1]]
+    @test cref_1_new.model === new_model
+    @test "cref[1]" == @inferred JuMP.name(cref_1_new)
+    cref_2_new = reference_map[cref[2]]
```
mlubin, line 542: Wasn't `cref[2]` filtered from the model? Why is it in the reference_map? Same for the two tests below.

test/model.jl
```diff
@@ -492,6 +515,85 @@ function test_copy_direct_mode()
     @test_throws ErrorException JuMP.copy(model)
 end
 
+function test_copy_filter()
+    copy_model_style_mode(true, MOIU.AUTOMATIC, true)
+end
+    
+function test_copy_filter_array()
+    model = Model()
+    model.optimize_hook = dummy_optimizer_hook
```
mlubin, line 524: The optimize hook and extension data aren't really relevant for these tests. The same functionality is already tested in copy_model_style_mode. Try to keep the tests focused on one thing, if possible (not all JuMP tests do this).

test/model.jl
```diff
@@ -492,6 +515,85 @@ function test_copy_direct_mode()
...
+    data = DummyExtensionData(model)
+    model.ext[:dummy] = data
+    @variable(model, x[i=1:2] ≥ 0, container=Array)
+    @constraint(model, cref[i=1:2], x[i] == 1, container=Array)
+    filter_constraints = (cr) -> cr != cref[1]
+    new_model, reference_map = JuMP.copy_model(model, filter_constraints=filter_constraints)
```
mlubin, line 530: Test that the number of AffineExpression-in-EqualTo constraints in the model is one. I think this test will fail because of my comment on `copy_model`.

src/copy.jl
```diff
@@ -80,7 +85,8 @@ function copy_model(model::Model)
     # changed, the `index_map` gives the map between the indices of
     # `backend(model` and the indices of `backend(new_model)`.
     index_map = MOI.copy_to(backend(new_model), backend(model),
-                            copy_names = true)
+                            copy_names = true, 
+                            filter_constraints = filter_constraints)
```
mlubin, line 89: There's a problem here. `MOI.copy_to` expects the filter to be on `MOI.ConstraintIndex` objects, while a JuMP user is likely to write a filter based on `JuMP.ConstraintIndex` (e.g., see the tests written). Some translation needs to be done.

src/copy.jl
dourouc05: It's not necessary to have something there (unless a specific solver requires solving before looking for conflicts?). 

However, I added the same comment around `Model()` to indicate that not all solvers are suitable. Should this example check that the conflict computation could be done properly (i.e. checking `MOI.get(model, MOI.ConflictStatus())`)?

dourouc05:
@mlubin Thanks for having a look! The new commits should fix the issues you raised.

test/model.jl
```diff
@@ -492,6 +515,73 @@ function test_copy_direct_mode()
...
+    
+    cref_2_new = reference_map[cref[2]]
+    @test cref_2_new.model === new_model
+    @test "cref[2]" == @inferred JuMP.name(cref_2_new)
+end
+
+function test_copy_filter_denseaxisarray()
+    model = Model()
+    model.optimize_hook = dummy_optimizer_hook
+    data = DummyExtensionData(model)
```
mlubin, line 544: Please remove `optimize_hook` and `DummyExtensionData` from these tests also.

mlubin (commented):
Looks like we're close to merging. Please resolve the merge conflicts so CI can run.

dourouc05:
The decrease in coverage is mostly due to `copy_conflict` not being called in the test suite. I also added a link to the new functionality from `solutions.md` in the docs.

mlubin (approved):
@odow or @blegat, could you take a quick look?

src/copy.jl
```diff
@@ -76,11 +81,26 @@ function copy_model(model::Model)
...
 
+    # At JuMP's level, filter_constraints should work with JuMP.ConstraintRef, 
+    # whereas MOI.copy_to's filter_constraints works with MOI.ConstraintIndex. 
+    if filter_constraints !== nothing
+        filter_constraints_user = filter_constraints
+        filter_constraints = (cref) -> begin
+            if cref isa MOI.ConstraintIndex
+                jump_cref = constraint_ref_with_index(model, cref)
+                return filter_constraints_user(jump_cref)
+            else
```
blegat, line 92: Could you add a comment explaining why this `else` clause is needed ? How could `cref` be something else than a `ConstraintIndex` ?

src/copy.jl
```diff
@@ -36,14 +36,18 @@ Base.broadcastable(reference_map::ReferenceMap) = Ref(reference_map)
...
 
 Return a copy of the model `model` and a [`ReferenceMap`](@ref) that can be used
 to obtain the variable and constraint reference of the new model corresponding
 to a given `model`'s reference. A [`Base.copy(::AbstractModel)`](@ref) method
 has also been implemented, it is similar to `copy_model` but does not return
 the reference map.
 
+If the `filter_constraints` argument is given, only the constraints for which
+this function returns `true` will be copied. This function is given a 
+constraint index as argument.
```
blegat, line 49: index -> reference

src/copy.jl
blegat: I get it, it's because you use it below, maybe do `moi_filter_constraints = (con_idx) -> ...` that only works with index, pass that to `copy_to` and for the object dictionary, use `filter_constraints` ?

dourouc05:
@blegat I have modified the doc. 

@mlubin too: I have added a new test based on MockOptimizer for `copy_conflict`, that should improve both coverage and confidence in the implementation. It needs a few changes in MOI, though: https://github.com/jump-dev/MathOptInterface.jl/pull/1202

src/copy.jl
odow:
```Julia
    function moi_filter_constraints(cref::MOI.ConstraintIndex)
        jump_cref = constraint_ref_with_index(model, cref)
        return filter_constraints(jump_cref)
    end
    filter = filter_constraints !== nothing ? moi_filter_constraints : nothing
    
    index_map = MOI.copy_to(
        backend(new_model),
        backend(model),
        copy_names = true,
        filter_constraints = filter
    )
```

src/copy.jl
```diff
@@ -93,7 +113,14 @@ function copy_model(model::Model)
...
     for (name, value) in object_dictionary(model)
-        new_model.obj_dict[name] = getindex.(reference_map, value)
+        if filter_constraints === nothing || 
+                (value isa ConstraintRef && filter_constraints(value)) || 
+                value isa VariableRef || 
+                eltype(value) == VariableRef
+            new_model.obj_dict[name] = getindex.(reference_map, value)
+        elseif eltype(value) == ConstraintRef && any(filter_constraints.(value))
+            new_model.obj_dict[name] = getindex.(reference_map, filter(filter_constraints, value))
+        end
```
odow, line 123: We generally try not to use `isa` like this. How about something like
```Julia
_copy_object(filter, value) = true
_copy_object(filter::Function, value::ConstraintRef) = filter(value)
function _copy_object(filter::Function, value::AbstractArray{<:ConstraintRef})
    return all(filter.(value))
end

for (name, value) in object_dictionary(model)
    if _copy_object(filter_constraints, value)
        new_model.obj_dict[name] = getindex.(reference_map, value)
    end
end
```

What does the `elseif` clause do when we only want some of the constraints? That seems like an issue. We probably want to just skip copying the name into `obj_dict` if only some of the container is copied.

dourouc05: That reasoning would lead to problems when copying just a conflict: maybe only a few constraints of the container are in the conflict, and only those should be copied. I'll propose a new iteration on the code soon.

odow: To clarify: we should copy on the constraints in the conflict, but maybe not set the name in `.obj_dict`.

src/copy.jl
dourouc05: @odow, I used your suggestion in 7601072.

src/copy.jl
dourouc05: I think I implemented your suggestion: (1) use functions instead of `isa` and (2) only copy into the dictionary the objects that are completely copied.

odow (commented):
Looking good. Can you add `copy_conflict` to https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/reference/solutions.md?

It also deserves a sentence in `docs/src/solutions.md` like: 
```
To create a new model containing only the constraints that participate in the conflict, use [`copy_conflict`](@ref). It may be helpful to write this model to file for easier debugging using [`write_to_file`](@ref).
```

dourouc05:
I did that in the last commit. (`copy_model` was already used in the sample code of `docs/src/solutions.md`.)

dourouc05:
The current test failure is likely only due to the fact that the latest tagged MOI doesn't have https://github.com/jump-dev/MathOptInterface.jl/pull/1202, on which this PR relies for the tests. (Aka: it works on my side®.)

odow (commented):
Looks good. There are some style things, I'm slowly rolling out JuliaFormatter, so I'm not worried about them now.

When a new MOI gets tagged, I can sort the rest of this out.

dourouc05:
Great, thanks! 

By the way, I was thinking of implementing CPLEX' and Gurobi's feature to easily write the IIS (i.e. a different extension when writing the files). Shall I do it in this PR or in another one? The problem is that this will depend a lot on `copy_model`.

odow:
You mean as a generic feature? Or specifically for Gurobi/CPLEX?

If we have `copy_conflict` and `write_to_file`, I don't see the benefit of a specialized `write_conflict_to_file`. You can just go:
```Julia
new_model, _ = copy_conflict(model)
write_to_file(new_model, "conflict.lp")
```

dourouc05:
These solvers' API allow you to just write the same as a one-liner (which is nicer to the user): 

```
write_to_file(model, "conflict.clp") # CPLEX extension
write_to_file(model, "conflict.ilp") # Gurobi extension
```

These could also be optimised by avoiding a copy of the model (I'll admit that I didn't do benchmarks to see if it's really worthwhile).

odow:
For the solver-specific API's people can just go
```Julia
# stuff
compute_conflict!(model)
GRBwrite(backend(model), "conflict.ilp")
```
In general, if you're debugging an infeasible model, you probably don't care about the extra overhead of the `copy_model`.

dourouc05:
My goal is not to provide access to more of the solvers' API, but only to replicate these features at JuMP's level.

dourouc05:
Too bad it didn't make it to the new version :(.

odow:
Once we make a new MOI release, we can make a new JuMP release. I'm aiming to have more smaller, more frequent releases than the ones we have been having.

odow:
@dourouc05 MOI 0.9.20 has been released, so you can rebase this on the latest master, and bump the minimum version of MOI in `Project.toml`!

odow:
There are still a few conflicts with the documentation

dourouc05:
Here it is, fully rebased :)! I squashed all commits into one to ease rebasing (many conflicts were created by changes from this PR that I ended up reverting in this PR).

odow:
Great! I'll merge this when CI passes. Thanks for sticking with this. It ticks off a big item from our TODO list!

dourouc05:
CI passes, with the exception of docs, but it really looks unrelated (doctest for `modify`).

odow:
That's my fault, sorry. Could you rebase onto latest master?

odow:
Don't worry. I realized I can do it.

odow:
Thanks!

dourouc05:
Thanks for merging this! It's the end of a journey that started nearly four years ago :D!

odow:
And all fixing https://github.com/jump-dev/JuMP.jl/issues/1035#issue-232333727 required was a complete re-write of JuMP, <s>MathProgBase</s>MOI, and CPLEX.jl!

PR #2342, opened 2020-09-21
Update aff_expr.jl to remove outdated TODO.

vtjeng:
Remove TODO about deprecations.

I can also add the deprecations for the functions based on the signatures [here](https://github.com/jump-dev/JuMP.jl/commit/72cf8b886c336d50e94b3cbdc935ddc640037180#diff-3d02cb6166feb9e45e071e40c7528596L72), if that's preferred.

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -222,8 +222,6 @@ product of two variables.
 """
 function add_to_expression! end
 
-# TODO: add deprecations for Base.push! and Base.append!
-
 # With one factor.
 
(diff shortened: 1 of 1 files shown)


mlubin:
Thanks, I think we missed our chance to add in the deprecations. The functions were deleted two years ago.

mlubin (approved):

vtjeng:
FWIW, it was still helpful because I could search for the function names `push!` and `append!` and infer useful information about how to update my code :)

blegat (approved):

PR #2345, opened 2020-09-29
Relaxation and fixed binary

### LINKED ISSUE #2344: Relaxation of fixed binary variable raises AssertionError
```Julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, x, Bin); @variable(model, y, Int);

julia> @objective(model, Max, y);

julia> @constraint(model, x + y <= 10);

julia> undo_relax = relax_integrality(model);

julia> undo_relax()

julia> fix(x, 1);

julia> println(model)
Max y
Subject to
 x + y ≤ 10.0
 x = 1.0
 y integer
 x binary


julia> undo_relax = relax_integrality(model);
ERROR: AssertionError: !(_moi_is_fixed(backend, v))
Stacktrace:
 [1] _moi_set_lower_bound(::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}, ::VariableRef, ::Float64) at /home/remi/.julia/packages/JuMP/qhoVb/src/variables.jl:385
 [2] set_lower_bound(::VariableRef, ::Float64) at /home/remi/.julia/packages/JuMP/qhoVb/src/variables.jl:376
 [3] relax_integrality(::Model) at /home/remi/.julia/packages/JuMP/qhoVb/src/variables.jl:1097
 [4] top-level scope at none:0

julia> println(model)
Max y
Subject to
 x + y ≤ 10.0
 x = 1.0
 y integer


julia> undo_relax = relax_integrality(model);

julia> println(model)
Max y
Subject to
 x + y ≤ 10.0
 x = 1.0
```

I would add to the end of 
https://github.com/jump-dev/JuMP.jl/blob/6c1a3bbe57a17b8a2b60fea6268d656b3b7bc815/test/variable.jl#L727-L777
```Julia
fix(x, 1)
unrelax = relax_integrality(model)
@test !is_binary(x)
@test is_fixed(x)
unrelax()
@test is_binary(x)
@test is_fixed(x)

fix(x, 0)
unrelax = relax_integrality(model)
@test !is_binary(x)
@test is_fixed(x)
unrelax()
@test is_binary(x)
@test is_fixed(x)

fix(a, 1)
unrelax = relax_integrality(model)
@test !is_integer(a)
@test is_fixed(a)
unrelax()
@test is_integer(a)
@test is_fixed(a)
```

remi-garcia:
Fix #2344 

Is that ok if I commit to this PR in order to change most `http://` to `https://` when websites handle it? Or should this be done in a separate PR?

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1097,6 +1097,10 @@ function relax_integrality(model::Model)
             if !info.has_fix
                 set_lower_bound(v, max(0.0, info.lower_bound))
                 set_upper_bound(v, min(1.0, info.upper_bound))
+            else
+                unfix(v)
+                set_lower_bound(v, max(0.0, info.fixed_value))
+                set_upper_bound(v, min(1.0, info.fixed_value))
             end
         end
     end
@@ -1117,6 +1121,10 @@ function relax_integrality(model::Model)
                     else
                         delete_upper_bound(v)
                     end
+                else
+                    delete_lower_bound(v)
+                    delete_upper_bound(v)
+                    fix(v, info.fixed_value)
                 end
             end
         end
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1094,8 +1094,10 @@ function relax_integrality(model::Model)
             unset_integer(v)
         elseif info.binary
             unset_binary(v)
-            set_lower_bound(v, max(0.0, info.lower_bound))
-            set_upper_bound(v, min(1.0, info.upper_bound))
+            if !info.has_fix
+                set_lower_bound(v, max(0.0, info.lower_bound))
+                set_upper_bound(v, min(1.0, info.upper_bound))
+            end
         end
     end
     function unrelax()
@@ -1104,15 +1106,17 @@ function relax_integrality(model::Model)
                 set_integer(v)
             elseif info.binary
                 set_binary(v)
-                if info.has_lb
-                    set_lower_bound(v, info.lower_bound)
-                else
-                    delete_lower_bound(v)
-                end
-                if info.has_ub
-                    set_upper_bound(v, info.upper_bound)
-                else
-                    delete_upper_bound(v)
+                if !info.has_fix
+                    if info.has_lb
+                        set_lower_bound(v, info.lower_bound)
+                    else
+                        delete_lower_bound(v)
+                    end
+                    if info.has_ub
+                        set_upper_bound(v, info.upper_bound)
+                    else
+                        delete_upper_bound(v)
+                    end
                 end
             end
         end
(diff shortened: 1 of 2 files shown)


blegat:
This is better done in a separate one

remi-garcia:
> This is better done in a separate one

Ok, it's done

src/variables.jl
```diff
@@ -1104,15 +1106,17 @@ function relax_integrality(model::Model)
...
-                if info.has_lb
-                    set_lower_bound(v, info.lower_bound)
-                else
-                    delete_lower_bound(v)
-                end
-                if info.has_ub
-                    set_upper_bound(v, info.upper_bound)
-                else
-                    delete_upper_bound(v)
+                if !info.has_fix
```
mlubin, line 1109: What happens if a binary variable is fixed to a value outside of [0, 1]? This should trigger an error, IMO.

remi-garcia: Using `relax_integrality` should trigger an error if a binary variable as been wrongly fixed? Shouldn't that be checked by `fix`?

mlubin: > Using relax_integrality should trigger an error if a binary variable as been wrongly fixed?

Using `relax_integrality` should trigger an error if the relaxation of a variable's domain is an empty set.

> Shouldn't that be checked by fix?

Not really, JuMP allows you to write models that are inconsistent (e.g., infeasible). This isn't checked when the constraints are added but later on.

remi-garcia: To be honest, I am not sure that throwing an error in `relax_integrality` is something that the user will expect. I tried something different to deal with fixed binary variables. It's obviously not perfect, yet it will behave as I would expect: the relaxation of this kind of infeasible model remains infeasible.

If you still think an error is the way to go, then I'll add that but I might need a little help to get a comprehensible sentence.

remi-garcia:
Last commit fails because I would expect that `is_fixed(v)` remains true after calling `relax_integrality`. Maybe I could `unfix(v)` if and only if `v` is a binary variable fixed out of bounds. That doesn't seem great but it is the best that I can come up with right now 😕

mlubin:
> Maybe I could unfix(v) if and only if v is a binary variable fixed out of bounds. That doesn't seem great 

Agreed, that's not great. I reiterate my suggestion to throw an error if no valid relaxation exists. I can help tweak the wording.

remi-garcia:
IMHO what would have been perfect needs the possibility to fix and to have bounds for a same variable. I'm trying to understand why this cannot (or shouldn't) be done but I applied your suggestion. Tests pass locally.

mlubin:
> I'm trying to understand why this cannot (or shouldn't) be done

Having a variable both fixed and with separate lower and upper bounds doesn't really make sense from any solver's perspective (no solver accepts this format). Fixing is a syntactic helper for setting the lower bound equal to the upper bound.

remi-garcia:
Ok, that makes sense. I get why an error is the way to go. Thanks!

PR #2352, opened 2020-10-15
Fix naming convention of internal functions

mlubin:


blegat (approved):
Ok to merge this assuming it we don't consider it breaking and it still allows us to continue with the v0.21.x series

mlubin:
The functions were already explicitly marked as internal, so I don't consider this breaking.

PR #2359, opened 2020-10-25
Add alternative syntax for "=" in macros

### LINKED ISSUE #1896: \in not supported in constraint macro
This code:
```julia
m = Model()
@variables(m, x)
@constraint(m, ctr[i ∈ 1:2], x >= i)
```
Results in error due to `∈`

If we use `∈` inside `sum` it works fine.

### LINKED ISSUE #2104: Style: in vs = for index variables in macros
https://discourse.julialang.org/t/style-guide-in-or/31155

CC @remi-garcia

joaquimg:
close #1896

### DIFF
--- a/src/Containers/macro.jl
+++ b/src/Containers/macro.jl
@@ -41,7 +41,7 @@ function _try_parse_idx_set(arg::Expr)
     if arg.head === :kw || arg.head === :(=)
         @assert length(arg.args) == 2
         return true, arg.args[1], arg.args[2]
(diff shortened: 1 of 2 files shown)


mlubin:
#2104 is also about the style guide recommendation, so I wouldn't consider it closed unless we decide which version to recommend in the style guide.

joaquimg:
My bad. Fixed that. Let's do one thing at a time.

mlubin (approved):

PR #1948, opened 2019-04-26
add additional example - factory schedule MILP

Crghilardi:
After getting help on [this discourse thread,](https://discourse.julialang.org/t/practice-jump-problem-scheduling-lp-questions/23390/9) I wanted to submit the finished problem as a 0.19 ready example. 

This pull request adds an additional complete MILP example problem.

Notes:
1. I originally wrote this using `Cbc`. I did not see `Cbc` listed in the `Project.TOML` and was not sure what the standards were on solver dependencies in testing, so I changed it to `GLPK` as other examples

2. This adds 2 data CSV files to the `/data` folder. Again, didn't want to mess with adding anything so I switched to using `DelimitedFiles` for reading in data rather than `CSV`

3. For file naming I just used `problemtype_name.jl`, not sure if there is a standard. While clicking through the example folder while working through the problem, I thought more explicit names would be helpful FWIW

This has been tested locally with `--project=examples run_examples.jl` and I get all tests passed on my machine (72/72)

PR #2366, opened 2020-10-30
add factory MIP example - take 2

Crghilardi:
this is #1948 cleaned up and submitted as a branch rather than on master.

Tested locally and passes 93/93 with `julia --project=examples examples/run_examples.jl`


odow:
Looks like there are some issues with printing constraints in Latex form with Literate/Documenter. I've reverted to a function for now.

blegat:
> Looks like there are some issues with printing constraints in Latex form with Literate/Documenter.

This was fixed in https://github.com/JuliaDocs/Documenter.jl/pull/1426 which will be included in Documenter v0.26, for SumOfSquares, I am checking out this branch of Documenter in the docs build to have the fix.

PR #2379, opened 2020-11-24
Add NLP tricks example

### LINKED ISSUE #2377: Add example on memoizing NLP functions
This is a useful example: https://discourse.julialang.org/t/how-to-set-up-nlconstraints-from-multi-output-user-defined-function/42309/5?u=odow
It's come up a few times on discourse and recently on SO (https://stackoverflow.com/questions/64947073/julia-jump-array-variable-constraint).

Wait for #2375 to be merged first.

odow:
A start at collating some of the more advanced usages of JuMP as examples in the docs.

Closes #2377

### DIFF
--- a/docs/src/examples/nlp_tricks.jl
+++ b/docs/src/examples/nlp_tricks.jl
@@ -61,29 +61,30 @@ println("Naive approach: function calls = $(function_calls)")
 # follows:
 
 """
-    memoize(foo::Function, n_outputs::Int; max_cache_size::Int = 50)
+    memoize(foo::Function, n_outputs::Int)
 
 Take a function `foo` and return a vector of length `n_outputs`, where each
 element is a function that returns the `i`'th output of `foo`.
 
-To avoid duplication of work, cache the evaluations of `foo`. `max_cache_size`
-controls the maximum number of solutions in the cache. A larger value may result
-in less function evaluations, at the cost of higher memory usage.
-
-Because `foo_i` is auto-differentiated with ForwardDiff, it needs to work when
-`x` is a `Float64` and a `ForwardDiff.Dual`.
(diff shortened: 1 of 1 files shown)


docs/src/examples/nlp_tricks.jl
```diff
@@ -0,0 +1,124 @@
...
+
+"""
+    memoize(foo::Function, n_outputs::Int)
+
+Take a function `foo` and return a vector of length `n_outputs`, where each
+element is a function that returns the `i`'th output of `foo`. To avoid
+duplication of work, cache the evaluations of `foo`.
+"""
+function memoize(foo::Function, n_outputs::Int)
+    cache = Dict{UInt, Any}()
```
mlubin, line 71: I think it's cleaner and a bit more memory efficient to store only the last evaluation point `last_x` and the value `foo(last_x)`. Then, check if x is equal to `last_x` and use `foo(last_x)` if it is.

odow: The difficulty is Ipopt doesn't call `foo_1` and `foo_2` sequentially, so checking `last_x` doesn't work. (You really need to check the last ~6, and that adds an equal amount of complexity). I've added a `max_cache_size` to deal with the memory issue.

mlubin: Why do you need to check the last 6?

odow: Hmm. When I set the `max_cache_size` to less than that it was using more than 20 evaluations. But you're right, it only needs to cache the last `Float64` evaluation and the last `ForwardDiff.Dual` evaluation.

odow:
Changed so we now cache the last `Float64` and `ForwardDiff.Dual` evaluations only.

mlubin (approved):

Issue #1062, opened 2017-06-15
Suggestions for Documentation Improvements


ccoffrin:
This is a list of points I ran into while developing the docstring-based documentation in #1058.

**Structure**
- [x] Incorporating docstrings highlights the mix of manual and library docs, consider breaking these into two sections at the top level (like the Julia manual) (#2380)
- [ ] Explore if there is a clean way to incorporate the epic notes in JuMP's code (For example, "Let S₊ be the cone of symmetric semidefinite matrices in" note in JuMP.jl) into docstrings that get into the developer docs.
- [x] The current API docs bring in methods on a case-by-case basis.  Consider using `@autodocs` to bringing in all methods (#2393)

**Maintenance**
- [x] Go through all functions without doc strings and figure which are missing documentation (New issue: #1825)
- [x] Go through all functions with empty doc strings and add documentation (e.g. `getvalue` and `getdual`) (Part of #1825)
- [x] Do the type signatures in macros make sense?  Is this the best way to document macro APIs?
- [x] Status documentation should probably link and/or pull in MPB docs
- [x] The list `const var_cats` should probably have documentation (No longer relevant)

mlubin:
Is there anything actionable to do here for 0.19?

mlubin:
Following the discussion in the gitter channel, we'll remove this from the 0.19 milestone. The main suggestion here that we should consider implementing is splitting docs into a manual and API reference like we do for MOI.

odow:
Closing since these points are now largely achieved, and other issues track more specific suggestions, e.g., https://github.com/jump-dev/JuMP.jl/issues/1825, https://github.com/jump-dev/JuMP.jl/issues/2485

PR #2380, opened 2020-11-25
Refactor documentation into separate API reference

odow:
Part of #1062 

The current docs are a confusing mix of narrative and docstrings. This PR creates a clear separation between the manual and the API reference.

My plan over the next few months is to ensure that every exported function has a decent docstring with examples, and to go through the manual cleaning it up and adding missing sections.


docs/src/reference/callbacks.md
```diff
@@ -0,0 +1,6 @@
+# Callback API
+
```
blegat, line 2: Maybe have a link to the manual section from each reference section in case someone arrives at this page from Google or whatever.

Issue #2354, opened 2020-10-16
Clarify what can be done in callbacks


odow:
In particular, clarify that `@variable` and `@constraint` can not be called, and doing so is undefined behavior.

Another option is to force solver wrappers to check if they are in a callback and throw a nice error message, but this seems like a high overhead that is liable to miss some functions. Callbacks are for advanced users, and so better documentation is all that is needed.

x-ref: https://github.com/jump-dev/CPLEX.jl/issues/329

PR #2381, opened 2020-11-26
[docs] clarify callback behavior

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/2354

### DIFF
--- a/docs/src/callbacks.md
+++ b/docs/src/callbacks.md
@@ -98,11 +98,18 @@ MOI.set(model, MOI.LazyConstraintCallback(), my_callback_function)
     model = Model(GLPK.Optimizer)
     @variable(model, x <= 10, Int)
     @objective(model, Max, x)
-    function my_callback_function(cb_data)
+    function bad_callback_function(cb_data)
+        # Don't do this!
         con = @build_constraint(x <= 2)
         MOI.submit(model, MOI.LazyConstraint(cb_data), con)
     end
-    MOI.set(model, MOI.LazyConstraintCallback(), my_callback_function)
+    function good_callback_function(cb_data)
+        if callback_value(x) > 2
+            con = @build_constraint(x <= 2)
+            MOI.submit(model, MOI.LazyConstraint(cb_data), con)
+        end
+    end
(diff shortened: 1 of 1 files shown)


docs/src/callbacks.md
```diff
@@ -38,17 +38,26 @@ Callback support is limited to a few solvers. This includes
...
-variables using [`callback_value`](@ref).
+There is a very limited range of things you can do during a callback. Only use 
+the functions and macros explicitly stated in this page of the documentation, or
+in the [Callbacks example](/examples/callbacks).
 
-If you need any other information, use a solver-dependent callback instead.
+Using any other part of the JuMP API (e.g., adding a constraint with [`@constraint`](@ref)
+or modifying a variable bound with [`set_lower_bound`](@ref)) is undefined
+behavior, and your solver may throw an error, return an incorrect solution, or 
+result in a segfault that aborts Julia.
```
blegat, line 50: There shouldn't be any segfault if the user does not use the C API. You don't punish users from not following the rules with segfault, only C/C++ does that and that's why people move to Julia.
Same think for incorrect solution, we should have errors from the wrappers.

mlubin: We can add error messages in some key places, but I don't think that providing this level of guarantees is really feasible. We would have to go through 100% of the JuMP API and add logic to check whether we're currently in a callback and what the current state of the callback is.

blegat: It should be done in the solvers wrappers, not JuMP. We can warn the user that these are still untested path but we should aim at having unit test for this. All MOI methods already start by checking whether the index is valid and the result index is valid, this would just be another test

odow: Wrappers can help by adding extra checks, but I don't think they should be unit tests. Ideally, the underlying solver would object to adding a constraint or variable inside a callback. 

Either way, having this documentation at the JuMP level is a win in the near-term.

odow:
Bump

mlubin (approved):
A discussion about how callbacks *should* work shouldn't block documentation on how they work now.

PR #2384, opened 2020-11-27
Format the Containers submodule

odow:
x-ref: #2292, #2383

### DIFF
--- a/src/Containers/Containers.jl
+++ b/src/Containers/Containers.jl
@@ -26,13 +26,16 @@ export DenseAxisArray, SparseAxisArray
 include("DenseAxisArray.jl")
 include("SparseAxisArray.jl")
 
-Base.eachindex(g::Base.Generator{<:Union{DenseAxisArray, SparseAxisArray}}) = eachindex(g.iter)
+Base.eachindex(g::Base.Generator{<:Union{DenseAxisArray,SparseAxisArray}}) =
+    eachindex(g.iter)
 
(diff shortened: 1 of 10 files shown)


mlubin:
If we doing this systematically, the formatting script should be checked in also.

odow:
Added a TOML configuration file, so now you can just go `using JuliaFormatter; format(dir)` and it will pick up the correct settings.

mlubin (approved):

src/Containers/Containers.jl
```diff
@@ -26,13 +26,16 @@ export DenseAxisArray, SparseAxisArray
 include("DenseAxisArray.jl")
 include("SparseAxisArray.jl")
 
-Base.eachindex(g::Base.Generator{<:Union{DenseAxisArray, SparseAxisArray}}) = eachindex(g.iter)
+Base.eachindex(g::Base.Generator{<:Union{DenseAxisArray,SparseAxisArray}}) =
+    eachindex(g.iter)
```
blegat, line 30: This is not allowed by our style guide, we should use `function` if ot is multiline

odow: The formatter only provides an option to re-write every single-line function, not an option to rewrite only those that are longer than the line limit.

Merging for now, and we can fix any offending examples by hand.

odow: I was wrong: https://github.com/jump-dev/JuMP.jl/pull/2385

PR #2388, opened 2020-11-28
Add vOptGeneric

freemin7:
It's a multi objective LP solver that calls other LP solvers

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -60,6 +60,7 @@ is often more complex. We list below the currently available solvers.
 | [SDPT3](https://blog.nus.edu.sg/mattohkc/softwares/sdpt3/)                     | [SDPT3.jl](https://github.com/jump-dev/SDPT3.jl)                                 | GPL      | LP, SOCP, SDP                      |
 | [SeDuMi](http://sedumi.ie.lehigh.edu/)                                         | [SeDuMi.jl](https://github.com/jump-dev/SeDuMi.jl)                               | GPL      | LP, SOCP, SDP                      |
 | [Tulip](https://github.com/ds4dm/Tulip.jl)                                     | [Tulip.jl](https://github.com/ds4dm/Tulip.jl)                                    | MPL-2    | LP                                 |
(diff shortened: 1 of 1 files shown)


mlubin:
I wouldn't classify vOptGeneric as a solver but as a JuMP extension, because it extends JuMP's syntax with `vModel`, `@addobjective`, `vSolve` etc. We should find or create an appropriate place to list JuMP extensions.

freemin7:
Fair.

blegat:
It seems they also provide a solver for multi-objective optimization. Once MOI allows multi-objective optimization, e.g. by allowing vector objective as suggested by @odow, this would allow it to be an MOI solver. But that's not the case at the moment so it shouldn't yet be in the table.

freemin7:
I recently had a conversation with people who research multiobjective optimization group at University of Jyvaskyla.
One reason for using multi objective optimization is that the proper weighing of the objectives depends on querying an decision maker. There a 4 ways how that might be done:

- The a proper weighing is know ahead of time and the problem can be solved as a single objective optimization 
- A surrogate model of the pareto front is required to that a decision maker can judge/explore it computationally cheaply
- After some sampling the decision maker get queried to might judge which objectives/parts of the parameter space are interesting to be explored more in case the space of options is really huge
- A sampling of the pareto front is required such that a decision maker can choose between those samples, possibly with some software that helps him explore that space.

I am not sure with all those variants can be accomplished with what JuMP has in mind for multi objective optimization but i found it wise to mention them, if someone else has ideas for that.

The problem is that the notion of "solving" a multiobjective optimization is a notion that has multiple approaches that results in a different types of return values.

odow:
Here's a plan for MO support in JuMP: https://github.com/jump-dev/JuMP.jl/issues/2099

odow:
I started a place to list JuMP extensions: https://github.com/jump-dev/jump-dev.github.io/pull/36

odow:
Closing in favor of https://github.com/jump-dev/jump-dev.github.io/pull/36

Issue #2025, opened 2019-08-14
Generic numeric type in JuMP


matbesancon:
MOI parameterizes function and set types with the number type used to store the data, example:

```julia
julia> using JuMP
julia> MOI.LessThan(big"3.5")
MathOptInterface.LessThan{BigFloat}(3.5)
```

For the moment, JuMP uses and assumes `Float64`:
```julia
julia> @variable(m, x[1:3])
julia> @constraint(m, sum(x) <= 1//7)
x[1] + x[2] + x[3] ≤ 0.14285714285714285
``` 

Julia-native solvers should support more numeric types (see https://github.com/JuliaOpt/Convex.jl/issues/262#issuecomment-515522868), using them through JuMP would in the current state block this genericity.

From the user perspective, generic number types could allow chosen precision, Unitful.jl and others

blegat:
One way to go for this is to parametrize the JuMP model with the coefficient type. `Model()` would give a `JuMP.Model{Float64}` and `Model{T}()` would give a `JuMP.Model{T}()`. `JuMP.VariableRef` will also be parametrized by `T` as one of its field is the `JuMP.Model`.
This means that in `operators`, we know which type to use for, e.g. `+(::VariableRef, ::VariableRef)`.

matbesancon:
I think this will be fairly easier to implement and test once we have solvers working on multiple number types, from the conversations of last week Hypatia (Float64 / Complex) and SCIP (Float64, Rational)

blegat:
`CDDLib.Optimizer{Rational{BigInt}}` and `Polyhedra.VRepOptimizer{Rational{BigInt}}` should work with `Rational{BigInt}`.

ericphanson:
@jiazhengzhu and I are working on wrapping SDPA-GMP for high precision SDP solving, which would benefit from being able to use `BigFloat`s. We’re currently using an branch of Convex with MOI support hacked on and parametrizing Convex’s `Problem` type with the numeric type, as was suggested above to do for JuMP. It would be great to be able to use JuMP to formulate problems as well.

chriscoey:
Hypatia and Tulip allow generic reals.
I just tried to use the `JuMP.Model{T}()` syntax expecting it to work but I forgot that it doesn't. My vote is for that syntax as it is quite natural.

rschwarz:
[...]

PR #2391, opened 2020-12-02
Preliminary support for complex numbers

blegat:
This PR is a proof of concept for adding complex numbers support for JuMP. We need a few changes to account for the fact that the expressions are now either `GenericAffExpr{Float64}` or `GenericAffExpr{Complex{Float64}}`.
As discussed in the README of https://github.com/jump-dev/ComplexOptInterface.jl, it is important that the JuMP variables remain real numbers, only the coefficients are complex.
Note that the printing is not working yet, I just did a few hacks in `src/print.Jl` so that it does not error but the printing is incorrect.
I started working on a nice printing for complex expressions in https://github.com/jump-dev/MathOptInterface.jl/pull/1180, it might be easier to first merge this one and use it from JuMP to avoid duplicate work.

Requires:

- [x] https://github.com/jump-dev/MathOptInterface.jl/issues/1693
- [x] https://github.com/jump-dev/JuMP.jl/issues/2814


odow:
We should rebase and take a look at this for JuMP 1.0

odow:
Here's the preview: https://jump.dev/JuMP.jl/previews/PR2391/tutorials/complex/complex/

odow:
@blegat what's the motivation for adding this to the JuMP documentation? I'd be more supportive of adding some tests instead. ComplexOptInterface should document itself, not as part of the JuMP documentation.

blegat:
The motivation was both to illustrate how to use an MOI extension from JuMP and to make it easier for users to see how to use it

odow:
I'd prefer that this didn't go into JuMP 1.0. It feels a little half-baked, and the inclusion of a v0.0.x package doesn't seem like a good idea in the JuMP 1.0 documentation. 

None of the changes in this PR are breaking, so they can be included in a 1.x release, at the same time we address https://github.com/jump-dev/JuMP.jl/issues/2025

blegat:
Yes, I'd also prefer JuMP v1.0 not depend on ComplexOptInterface. This will be integrated once ComplexOptInterface is in MOI.
Maybe we can just in JuMP what's needed (so the fixes included in this PR) but without having the docs require ComplexOptInterface.

odow:
> Maybe we can just in JuMP what's needed (so the fixes included in this PR) but without having the docs require ComplexOptInterface.

This sounds good.

blegat:
I can't approve my own PR but it looks good to merge

odow:
The failure in that last commit demonstrates that we need to be a bit more careful about this before making it official.

PR #2393, opened 2020-12-03
[docs] Add missing docstrings and fix doctests that arose as a result

odow:
Part of https://github.com/jump-dev/JuMP.jl/issues/1062

There are a couple of things that I will comment on in-line.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -6,7 +6,7 @@ using Test
 # Pass --fast as an argument to skip rebuilding the examples and running
 # doctests. Only use this argument to rapidly test small changes to the
 # Markdown. _Never_ set it in production.
-const _FAST = get(ARGS, 1, "") == "--fast"
+const _FAST = findfirst(isequal("--fast"), ARGS) !== nothing
 
 const _EXAMPLE_DIR = joinpath(@__DIR__, "src", "examples")
 
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -3,6 +3,11 @@ using Literate
 using JuMP
 using Test
 
+# Pass --fast as an argument to skip rebuilding the examples and running
+# doctests. Only use this argument to rapidly test small changes to the
+# Markdown. _Never_ set it in production.
+const _FAST = get(ARGS, 1, "") == "--fast"
+
 const _EXAMPLE_DIR = joinpath(@__DIR__, "src", "examples")
 
 function link_example(content)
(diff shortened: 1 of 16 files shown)


docs/make.jl
```diff
@@ -14,32 +19,59 @@ function link_example(content)
...
+    # don't necessarily want to document every MOI method.
+    #
+    # This is should be fine for now, because MOI doesn't export anything.
+    # However, also problematically, some doctests in MOI are not checked and
+    # are failing. Until they are fixed, we can't enable these options.
+    #
+    # TODO(odow): uncomment when possible.
+    # modules = [JuMP, MOI],
+    # checkdocs = :exports,
+    # ==========================================================================
```
odow, line 72: So we really want to enable these two options. But there are some failures in MOI doctests that I need to fix first.

docs/src/reference/variables.md
```diff
@@ -49,4 +49,15 @@ num_variables
 VariableRef
 index(::VariableRef)
 optimizer_index(::VariableRef)
+
+add_variable
+check_belongs_to_model
+build_variable
+parse_one_operator_variable
```
odow, line 56: This function is currently exported. Should it be?

docs/make.jl
```diff
@@ -3,6 +3,11 @@ using Literate
 using JuMP
 using Test
 
+# Pass --fast as an argument to skip rebuilding the examples and running
+# doctests. Only use this argument to rapidly test small changes to the
+# Markdown. _Never_ set it in production.
+const _FAST = get(ARGS, 1, "") == "--fast"
```
odow, line 9: I can remove or split this into a separate PR if needed, but it was useful when compiling the list of missing docstrings, not to have to repeatedly generate the literate files.

blegat: Why not `findfirst(isequal("--fast"), ARGS) !== nothing` ?

docs/src/reference/variables.md
blegat: Yes: https://github.com/blegat/SetProg.jl/blob/64c056f8c22b3b79fdc0862c8938a0f23bd03796/src/macros.jl

docs/make.jl
```diff
@@ -74,4 +106,5 @@ makedocs(
 
 deploydocs(
     repo   = "github.com/jump-dev/JuMP.jl.git",
+    push_preview = true,
```
blegat, line 109: I see that the doc of this PR is now available at https://jump.dev/JuMP.jl/previews/PR2393/, that's useful

mlubin (approved):

PR #2390, opened 2020-11-30
Bound StaticArrays to fix Julia 1.0 compatibility

odow:
CI is broken on Julia 1.0. We were passing 5 hours ago here: https://github.com/jump-dev/JuMP.jl/commit/77a0b82bf47f35913cc8cff0c389425948cb4fea
But failing 2 hours ago here: https://github.com/jump-dev/JuMP.jl/commit/eefafb163755b37faba5e0c0d648d77ef0a6796d

This release of ForwardDiff is the only thing I can find that changed within that time-frame: https://github.com/JuliaRegistries/General/pull/25512
https://github.com/JuliaDiff/ForwardDiff.jl/compare/v0.10.12...v0.10.13

PR #2396, opened 2020-12-06
Revert #2390

odow:
StaticArrays 1.0.1 has been tagged with a fix for the large memory usage


odow:
Seems like we're back up and running again.

blegat (approved):

PR #2411, opened 2020-12-14
Remove incorrect and un-used copy functions

### LINKED ISSUE #2407: Copying expressions between models
While working on #2405, I found theses two functions:

https://github.com/jump-dev/JuMP.jl/blob/90e63c47372e02b7093a9bf2fb4153dfa548c0a6/src/quad_expr.jl#L434-L439

https://github.com/jump-dev/JuMP.jl/blob/90e63c47372e02b7093a9bf2fb4153dfa548c0a6/src/aff_expr.jl#L425-L434

The first is invalid (the `.qvars1` fields have been replaced by an `OrderedDict`), so it clearly isn't tested or used by anyone in the last ~2 years. The second is tested here:
https://github.com/jump-dev/JuMP.jl/blob/90e63c47372e02b7093a9bf2fb4153dfa548c0a6/test/expr.jl#L157-L164

But it has an implicit assumption that the variables in the new model have the same `.value` field as the original model:
https://github.com/jump-dev/JuMP.jl/blob/90e63c47372e02b7093a9bf2fb4153dfa548c0a6/src/JuMP.jl#L796-L801

Now that we've moved to `MOI.copy_to`, can these be removed? Is anyone using them?

odow:
Closes #2407

Technically, this is a breaking change, but the definitions are old and incorrect, so no one should be relying on them.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -400,24 +400,24 @@ end
 """
      print_bridge_graph([io::IO,] model::Model)
 
-Print the hyper-graph containing all variable, constraint, and objective types 
-that could be obtained by bridging the variables, constraints, and objectives 
+Print the hyper-graph containing all variable, constraint, and objective types
+that could be obtained by bridging the variables, constraints, and objectives
 that are present in the model.
 
-Each node in the hyper-graph corresponds to a variable, constraint, or objective 
+Each node in the hyper-graph corresponds to a variable, constraint, or objective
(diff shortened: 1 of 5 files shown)


blegat (approved):

odow:
Going to wait for @mlubin to see this and decide if we want to remove these functions.

mlubin (approved):

PR #2416, opened 2020-12-15
Add support for skew symmetric variables

### LINKED ISSUE #2146: Support for skew-symmetric matrix variables

odow:
Rebased from https://github.com/jump-dev/JuMP.jl/pull/2147

Closes #2146 

cc @frederikgeth

### DIFF
--- a/test/variable.jl
+++ b/test/variable.jl
@@ -484,12 +484,13 @@ end
 
 function test_variable_skewsymmetric(ModelType, ::Any)
     model = ModelType()
-    @variable model x[1:2, 1:2] in SkewSymMatrixSpace()
+    @variable(model, x[1:2, 1:2] in SkewSymMatrixSpace())
     @test x[1, 2] == -x[2, 1]
     @test iszero(x[1, 1])
     @test iszero(x[2, 2])
     @test model[:x] === x
-    @variable model z[1:3, 1:3] in SkewSymMatrixSpace()
+    @test num_variables(model) == 1
+    @variable(model, z[1:3, 1:3] in SkewSymMatrixSpace())
     @test z[1, 2] == -z[2, 1]
     @test z[1, 3] == -z[3, 1]
     @test z[2, 3] == -z[3, 2]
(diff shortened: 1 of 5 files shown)


test/variable.jl
```diff
@@ -482,6 +482,30 @@ function test_variable_symmetric(ModelType, ::Any)
     @test y[1, 2] === y[2, 1]
 end
 
+function test_variable_skewsymmetric(ModelType, ::Any)
+    model = ModelType()
+    @variable(model, x[1:2, 1:2] in SkewSymMatrixSpace())
```
odow, line 487: So I've been playing with this, and the syntax is pretty restrictive. (This also applied to symmetric matrices.) 

For example, we don't support 
```julia
@variable(model, x[2:3, 2:3] in SkewSymMatrixSpace())
```
because it isn't apparent you can have a symmetric `DenseAxisArray`. 

But if users are forced to write 
```Julia
@variable(model, x[1:3, 1:3] in SkewSymMatrixSpace())
```
why not just 
```julia
@variable(model, X in SkewSymMatrixSpace(3))
```
For anonymous variables, we could support 
```julia
X = @variable(model, _ in SkewSymMatrixSpace(3))
```

blegat (approved):

odow:
We can keep discussing different syntax in the next monthly developer call.

PR #2417, opened 2020-12-16
Add an unregister function

### LINKED ISSUE #1956: Deleting constraint in order to modify it not possible due to name registered in model
I run an MILP several times with different parameter values. Instead of recreating the model every time, I would like to be able to just change the relevant constraints. From the documentation, I figured that I'd be able to do this by deleting the constraint and then reformulating it.

Here's a toy example:
```julia
using JuMP

model = Model(with_optimizer(Gurobi.Optimizer))
@variable(model, x[i=1:2,j=1:2])
@constraint(model, con[i=1:2,j=1:2], x[i,j] >= 1)
for i=1:2
    for j=1:2
        delete(model, con[i,j])
    end
end
@constraint(model, con[i=1:2,j=1:2], x[i,j] >= 2)
```

This produces the error:
```
ERROR: LoadError: An object of name con is already attached to this model. If this is intended, consider using the anonymous construction syntax, e.g., x = @variable(model, [1:N], ...) where the name of the object does not appear inside the macro.
```

One workaround I can see is to delete the name `con` in model, but I don't know how / cannot do this. Ideally I would like a simpler way of modifying constraints, for example as:

```julia
model = Model(with_optimizer(Gurobi.Optimizer))
@variable(model, x[i=1:2,j=1:2])
@constraint(model, con[i=1:2,j=1:2], x[i,j] >= 1)
@constraint(model, con[i=1:2,j=1:2], x[i,j] >= 2)
```

However it would appear that this is quite difficult judging from this issue: https://github.com/JuliaOpt/JuMP.jl/issues/1241

odow:
Alternative to #2232 

Closes #2232
Closes #1956

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -487,7 +487,6 @@ function num_nl_constraints(model::Model)
     return model.nlp_data !== nothing ? length(model.nlp_data.nlconstr) : 0
 end
 
-
 """
     object_dictionary(model::Model)
 
@@ -508,7 +507,7 @@ object_dictionary(model::Model) = model.obj_dict
 Unregister the name `key` from `model` so that a new variable, constraint, or
 expression can be created with the same key.
 
-Note that this will not delete the object `model[key]`, it will just remove the
+Note that this will not delete the object `model[key]`; it will just remove the
 reference at `model[key]`. To delete the object, use
 ```julia
 delete(model, model[key])
@@ -528,7 +527,7 @@ e.g., `x = @variable(model, [1:N], ...)` where the name of the object
 does not appear inside the macro.
 
 Alternatively, use `unregister(model, :x)` to first unregister the
-existing name from the model. Note that this will not delete the object,
+existing name from the model. Note that this will not delete the object;
 it will just remove the reference at `model[:x]`.
 [...]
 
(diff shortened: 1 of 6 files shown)


src/JuMP.jl
```diff
@@ -487,9 +487,74 @@ function num_nl_constraints(model::Model)
...
+2
+"""
+function unregister(model::AbstractModel, key::Symbol)
+    delete!(object_dictionary(model), key)
+    return
+end
+
+function unregister(::AbstractModel, key)
+    error(
+        "Invalid name $(key). The second argument to `unregister` must be a " *
```
mlubin, line 553: This method doesn't seem to do much. Isn't MethodError the right error when a user is calling a function with incorrect types?

src/macros.jl
```diff
@@ -61,10 +61,15 @@ end
...
-              " syntax, e.g., x = @variable(model, [1:N], ...) where the " *
-              "name of the object does not appear inside the macro.")
+        error("""An object of name $name is already attached to this model. If this
+            is intended, consider using the anonymous construction syntax, e.g.,
+            `x = @variable(model, [1:N], ...)` where the name of the object does
+            not appear inside the macro.
+
+            Alternatively, use `unregister(model, :$(name))` to first unregister
+            the existing name from the model. Note that this will not delete the
+            object, it will just remove the reference at `model[:$(name)]`.
```
mlubin, line 71: Nit: semicolon instead of comma

mlubin (approved):
LGTM but should also be reviewed by blegat@

blegat (approved):

PR #2418, opened 2020-12-16
Use setindex! internally

blegat:
These were part of https://github.com/jump-dev/JuMP.jl/pull/2232/


odow:
Looks like you need to define `setindex!` for `JuMPExtension`.

Issue #2333, opened 2020-09-16
[Feature Request] Error Function Keyword Argument for Macros


pulsipher:
**Is your feature request related to a problem? Please describe.**
Currently, extensions that employ their own macros that in turn wrap JuMP macros get error messages that are not meaningful as exemplified below:
```julia
using JuMP

macro my_new_variable(model, args...)
   # Define macro error function that is relevant to this macro
   _my_error_func(str...) = JuMP._macro_error(:my_new_variable, (model, args...), __source__, str...)

   # Do desired extension manipulations before calling JuMP.@variable (calling _my_error_func where needed)
   new_args = args # do nothing in this example for simplicity

   # Prepare and return the code wrapping JuMP.@variable 
   code = :( JuMP.@variable($model, $(new_args...)) )
   return esc(code)
end 

# Make a call that invokes an error 
model = Model()
@my_new_variable(model, [1:2] >= 0) 
```
```julia
ERROR: LoadError: At REPL[9]:9: `@variable(model, [1:2] >= 0)`: Cannot use explicit bounds via >=, <= with an anonymous variable
Stacktrace:
 [1] error(::String, ::String) at .\error.jl:42
 [2] _macro_error(::Symbol, ::Tuple{Symbol,Expr}, ::LineNumberNode, ::String) at my_pathr\.julia\packages\JuMP\e0Uc2\src\macros.jl:924
 [3] (::getfield(JuMP, Symbol("#_error#88")){LineNumberNode,Tuple{Symbol,Expr}})(::String) at my_path\.julia\packages\JuMP\e0Uc2\src\macros.jl:1198
 [4] @variable(::LineNumberNode, ::Module, ::Vararg{Any,N} where N) at my_path\.julia\packages\JuMP\e0Uc2\src\macros.jl:1253
in expression starting at REPL[9]:9
```
Notice that the the given error message doesn't refer to the macro that the user actually called and the sourcing info is not 
where the user called it (it refers to where the extension macro calls `@variable`). It would be nice to pass on `_my_error_func` to @variable such that it uses that instead of the `_error` defined within the `@variable` macro. Use cases come up for instance in InfiniteOpt.jl. 

**Describe the solution you'd like**
It would very helpful to add a keyword argument in JuMP macros with which a user-defined error function can be provided. By default if this keyword argument is not provided then just use the current implementation, otherwise if a user-defined error function is given it should be used in place of the `_error` defined inside the JuMP macro. 

[...]

PR #2421, opened 2021-01-04
Split variable macro in macro + function

joaquimg:
This way people can more easily extend the `@variable` macro.
Ideally, `_variable_macro` would not have the `_` so people extending jump could use it safely.

Extensions of this macro were seen in some extensions and can be very useful.
Could be a partial solution for #2333 
Could be used for `@state` macro in `SDDP`
Could be used for a `@parameter` macro in the future of ParametricOptInterface/ParameterJuMP

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1293,7 +1293,13 @@ end
 ```
 """
 macro variable(args...)
-    _error(str...) = _macro_error(:variable, args, __source__, str...)
+    _variable_macro(args, :variable, __source__)
+end
+
+function _variable_macro(
+    args, macro_name::Symbol, source::LineNumberNode
+)
+    _error(str...) = _macro_error(macro_name, args, source, str...)
 
     model = esc(args[1])
 
@@ -1445,7 +1451,7 @@ macro variable(args...)
         macro_code = _macro_assign_and_return(creation_code, variable, name,
                                               model_for_registering = model)
     end
-    return _finalize_macro(model, macro_code, __source__)
+    return _finalize_macro(model, macro_code, source)
 end
 
 """
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1293,10 +1293,10 @@ end
 ```
 """
 macro variable(args...)
-    _variable_macro(args, :variable, __source__)
+    variable_macro(args, :variable, __source__)
 end
 
-function _variable_macro(
+function variable_macro(
     args, macro_name::Symbol, source::LineNumberNode
 )
     _error(str...) = _macro_error(macro_name, args, source, str...)
@@ -1313,14 +1313,14 @@ function _variable_macro(
     else
         x = popfirst!(extra)
         if x == :Int
-            _error("Ambiguous variable name $x detected. To specify an anonymous integer " *
-                "variable, use `@variable(model, integer = true)` instead.")
+            _error("Ambiguous $(macro_name) name $x detected. To specify an anonymous integer " *
+                "$(macro_name), use `@$(macro_name)(model, integer = true)` instead.")
         elseif x == :Bin
-            _error("Ambiguous variable name $x detected. To specify an anonymous binary " *
-                "variable, use `@variable(model, binary = true)` instead.")
+            _error("Ambiguous $(macro_name) name $x detected. To specify an anonymous binary " *
+                "$(macro_name), use `@$(macro_name)(model, binary = true)` instead.")
         elseif x == :PSD
-            _error("Size of anonymous square matrix of positive semidefinite anonymous variables is not specified. To specify size of square matrix " *
-                "use `@variable(model, [1:n, 1:n], PSD)` instead.")
+            _error("Size of anonymous square matrix of positive semidefinite anonymous $(macro_name)s is not specified. To specify size of square matrix " *
+                "use `@$(macro_name)(model, [1:n, 1:n], PSD)` instead.")
         end
         anon_singleton = false
     end
@@ -1354,7 +1354,7 @@ function _variable_macro(
 
     anonvar = isexpr(var, :vect) || isexpr(var, :vcat) || anon_singleton
     if anonvar && explicit_comparison && set === nothing
-        _error("Cannot use explicit bounds via >=, <= with an anonymous variable")
+        _error("Cannot use explicit bounds via >=, <= with an anonymous $(macro_name)")
     end
     variable = gensym()
     # TODO: Should we generate non-empty default names for variables?
@@ -1366,7 +1366,7 @@ function _variable_macro(
     end
 
     if !isa(name, Symbol) && !anonvar
-        Base.error("Expression $name should not be used as a variable name. Use the \"anonymous\" syntax $name = @variable(model, ...) instead.")
+        Base.error("Expression $name should not be used as a $(macro_name) name. Use the \"anonymous\" syntax $name = @$(macro_name)(model, ...) instead.")
     end
 
     if !isempty(set_kw_args)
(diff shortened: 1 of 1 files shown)


odow (commented):
How does this make it easier to extend the `@variable` macro?

joaquimg:
It is easy to rename it and one could simply append something in args.
Otherwise, copying a lot of code that relies on private functions would be needed.

odow:
I still don't see how it helps. You need to pass a carefully crafted `args` to the function, and this isn't extensible. We're better off making it easier to write new macros that work with containers.

Do you have an example for how ParameterJuMP would use this?

joaquimg:
My first thought was just append something like: "SDDP.State". For instance ParameterJuMP.Param
The the macro:
`@parameter` would just append `ParameterJuMP.Param` in args and call `variable_macro`.

odow:
But what's to stop someone going
```Julia
@parameter(model, 0 <= [i = 1:2, 1:2] <= i, PSD)
```
You don't want to inject a `ParameterJuMP.Parameter()` symbol into the arguments without validating the syntax.

The macros do need cleaning up, but I don't think this is the way to do it.

joaquimg:
That would be an error because no `build_variable` would dispatch with both `PSD` and `Param`.

This was a quick fix that did not seem harmful. Probably there are better ways of doing that.

Do you have something in mind?

odow:
> That would be an error because no build_variable would dispatch with both PSD and Param

But it would be a weird error telling them about this `Param` struct which they didn't write. If we add this function as public, it will get exported and we're locked-in to supporting it going forward. That limits our ability to refactor the macros.

For now, I suggest 
```Julia
@variable(model, p == 1, ParameterJuMP.Parameter)
```

joaquimg:
I understand your point. I agree that the error with a  `Param` that the user never wrote is a bit weird.
I really don't like this syntax though: `@variable(model, p == 1, ParameterJuMP.Parameter)`.

I read more carefully the `@variable` macro, I have two comments:

1) `build_variable` is called "twice":
https://github.com/jump-dev/JuMP.jl/blob/d2398ad1a5e90a3fda51274ccc3b48b88e1abbb9/src/macros.jl#L1418
https://github.com/jump-dev/JuMP.jl/blob/d2398ad1a5e90a3fda51274ccc3b48b88e1abbb9/src/macros.jl#L1426

These two, however, have different meanings, one possible change to facilitate extension is to rename the second as:
`build_variable_in_set`, for instance.
This way these are clearly two functions that the user has to extend.

The first (`build_variable`) is extended with the `extra` positional arguments.
While the second (that am naming `build_variable_in_set`) is extended by either a new `set` or a new `scalar_variables` (or both).

2) In my view, the biggest issue stopping easy macro renaming is checking the `extra` positional arguments, since the remainder of the macro is protected already and keyword args can be checked in the `build_variable...` functions.

a) One way out here would be to pass a function as an argument to `variable_macro` that checks just the `extra` positional arguments, the default could be an empty check.

The extension would pass this function like:

```julia
function check_positional(extra)
...
end
macro parameter(args...)
    variable_macro(args, :parameter, __source__, check_positional)
end
```

b) Another solution would be to completely disallow `extra` positional arguments, and allow the user to pass a type to `variable_macro` which would be used in the first `build_variable`:
```julia
macro parameter(args...)
    variable_macro(args, :parameter, __source__, my_type)
end
```

c) As an alternative to disallowing the `extra` positional arguments, it could be allowed but it would be checked against `my_type` forcing the only element in `extra` to be `my_type`.
An argument for this would be there is not much reason to have two of these... (also, two can be messy because the order would matter on dispatch).

odow:
Re 1) Yes. Needs refactoring.

Re 2) Agree on c). I just think the macros need a thorough refactor.

Another syntax option without needing a new macro is:
```Julia
@variable(model, p in Parameter(default))
```

pulsipher:
For what it is worth, a change along the lines of what joaquimg is suggesting would be very useful for InfiniteOpt.jl as well.

odow:
I'll add this as an agenda item for the next monthly developer call.

odow:
Developer call says yes, we should refactor the macros.  

One point of the discussion was removing the ability to pass multiple positional arguments, but I've just checked, and it actually calls something reasonable:
```Julia
julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> struct Foo end

julia> struct Bar end

julia> @variable(model, x, Foo, Bar)
ERROR: MethodError: no method matching build_variable(::JuMP.var"#_error#86"{LineNumberNode,NTuple{4,Symbol}}, ::VariableInfo{Float64,Float64,Float64,Float64}, ::Type{Foo}, ::Type{Bar})
Closest candidates are:
  build_variable(::Function, ::VariableInfo; extra_kw_args...) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/macros.jl:906
  build_variable(::Function, ::Array{var"#s104",1} where var"#s104"<:ScalarVariable, ::AbstractVectorSet) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/sets.jl:4
  build_variable(::Function, ::Array{var"#s104",2} where var"#s104"<:ScalarVariable, ::SymMatrixSpace) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/sd.jl:163
  ...
Stacktrace:
 [1] top-level scope at /Users/oscar/.julia/packages/JuMP/qhoVb/src/macros.jl:91
 [2] top-level scope at REPL[32]:1

julia> @variable(model, x, Foo, Bar, Int)
ERROR: MethodError: no method matching build_variable(::JuMP.var"#_error#86"{LineNumberNode,NTuple{5,Symbol}}, ::VariableInfo{Float64,Float64,Float64,Float64}, ::Type{Foo}, ::Type{Bar})
Closest candidates are:
  build_variable(::Function, ::VariableInfo; extra_kw_args...) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/macros.jl:906
  build_variable(::Function, ::Array{var"#s104",1} where var"#s104"<:ScalarVariable, ::AbstractVectorSet) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/sets.jl:4
  build_variable(::Function, ::Array{var"#s104",2} where var"#s104"<:ScalarVariable, ::SymMatrixSpace) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/sd.jl:163
  ...
Stacktrace:
 [1] top-level scope at /Users/oscar/.julia/packages/JuMP/qhoVb/src/macros.jl:91
 [2] top-level scope at REPL[33]:1
```

Another option is to add a syntax validator, so that macros taking an expression, modifying it, and passing it to a JuMP macro can validate whether their revised expression would throw an inscrutable error for the user.

Perhaps we should close this PR in favor of an issue that summarizes the refactoring changes we want to make to the macros?

odow:
Following up, there is now a nicer fallback error for multiple positional arguments: https://github.com/jump-dev/JuMP.jl/pull/2451

joaquimg:
Yes, I agree with replacing this with an issue.

Syntax validator looks nice.

The call to multiple positional arguments is reasonable. 
However,
1 - the positional arguments are positional sometimes:
```julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, x[1:2,1:2], Int, PSD)
2×2 LinearAlgebra.Symmetric{VariableRef,Array{VariableRef,2}}:
 x[1,1]  x[1,2]
 x[1,2]  x[2,2]

julia> @variable(model, y[1:2,1:2], PSD, Int)
2×2 LinearAlgebra.Symmetric{VariableRef,Array{VariableRef,2}}:
 y[1,1]  y[1,2]
 y[1,2]  y[2,2]
```

2 - Creating an overloaded version that takes care of the multiple orders `Foo` and `Bar` can appear is error prone.

If the "extender" wants two new things they could go and define `FooBar`

Hence the argument for only allowing 1 extra positional argument, which could be in any order just like the current ones.

Keyword arguments can still be arbitrarily many.

odow:
> the positional arguments are positional sometimes

`Int`, `Bin`, and `PSD` are special and handled separately, so that don't appear as "positional" to `build_variable`.

joaquimg:
I understand, but from the user perspective is weird to be able to put these in any order, but having a special order for Foo and Bar.

odow:
Closing in favor of #2583

PR #2436, opened 2021-01-26
Generalize ConstraintRef Model type

### LINKED ISSUE #2256: Generalize modification functions
The majority of functions in https://github.com/jump-dev/JuMP.jl/blob/master/src/constraints.jl are specialized to `Model`s.

It would be very helpful for extensions (e.g., https://github.com/pulsipher/InfiniteOpt.jl/issues/55) if these were generalized to `AbstractModel`.

Examples:
https://github.com/jump-dev/JuMP.jl/blob/ecf12ab5868ba5af45e38ee721f1f5ea0c43a1b7/src/constraints.jl#L436-L437
https://github.com/jump-dev/JuMP.jl/blob/ecf12ab5868ba5af45e38ee721f1f5ea0c43a1b7/src/constraints.jl#L452-L453
https://github.com/jump-dev/JuMP.jl/blob/ecf12ab5868ba5af45e38ee721f1f5ea0c43a1b7/src/constraints.jl#L479-L480
https://github.com/jump-dev/JuMP.jl/blob/ecf12ab5868ba5af45e38ee721f1f5ea0c43a1b7/src/constraints.jl#L495
https://github.com/jump-dev/JuMP.jl/blob/ecf12ab5868ba5af45e38ee721f1f5ea0c43a1b7/src/constraints.jl#L557

cc @pulsipher

guilhermebodin:
Implements the suggestions on #2256 

Closes #2256


guilhermebodin:
Doctests fail because of the abstract type in cases like
```
Array{ConstraintRef{#s118,MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.LessThan{Float64}},ScalarShape} where #s118<:AbstractModel,1}:
```
instead of 
```
Array{ConstraintRef{Model,MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.LessThan{Float64}},ScalarShape},1}:
```

src/constraints.jl
```diff
@@ -934,10 +934,10 @@ function all_constraints(
     # TODO: Support JuMP's set helpers like SecondOrderCone().
     f_type = moi_function_type(function_type)
     if set_type <: MOI.AbstractScalarSet
-        constraint_ref_type = ConstraintRef{Model, _MOICON{f_type, set_type},
+        constraint_ref_type = ConstraintRef{<:AbstractModel, _MOICON{f_type, set_type},
```
odow, line 937: This needs to match the type of `model`. Either change it back to `Model`, or add a type parameter for `model`.

src/constraints.jl
```diff
@@ -934,10 +934,10 @@ function all_constraints(
...
                                             ScalarShape}
     else
-        constraint_ref_type = ConstraintRef{Model, _MOICON{f_type, set_type}}
+        constraint_ref_type = ConstraintRef{<:AbstractModel, _MOICON{f_type, set_type}}
```
odow, line 940: Ditto here.

odow (changes requested):
This is why the printing is failing.

src/constraints.jl
```diff
@@ -912,15 +912,15 @@ julia> @variable(model, x >= 0, Bin);
...
  x ≥ 0.0
 
 julia> all_constraints(model, VariableRef, MOI.ZeroOne)
-1-element Array{ConstraintRef{Model,MathOptInterface.ConstraintIndex{MathOptInterface.SingleVariable,MathOptInterface.ZeroOne},ScalarShape},1}:
+1-element Array{ConstraintRef{<:AbstractModel,MathOptInterface.ConstraintIndex{MathOptInterface.SingleVariable,MathOptInterface.ZeroOne},ScalarShape},1}:
  x binary
 
 julia> all_constraints(model, AffExpr, MOI.LessThan{Float64})
-1-element Array{ConstraintRef{Model,MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.LessThan{Float64}},ScalarShape},1}:
+1-element Array{ConstraintRef{<:AbstractModel,MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.LessThan{Float64}},ScalarShape},1}:
```
blegat, line 923: I don't understand why this is printed like that, it should be `Model`, not `<:AbstractModel`

src/constraints.jl
```diff
@@ -655,7 +655,7 @@ into account in the primal value of the constraint. For instance, the constraint
 evaluation of `2x + 3y`.
 ```
 """
-function value(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)
+function value(con_ref::ConstraintRef{<:AbstractModel, <:_MOICON}; result::Int = 1)
```
odow, line 658: It doesn't really make sense to make this an `AbstractModel`, unless we also change the signature on `_constraint_primal`.

src/constraints.jl
```diff
@@ -710,7 +710,7 @@ Use `has_dual` to check if a result exists before asking for values.
 
 See also: [`result_count`](@ref), [`shadow_price`](@ref).
 """
-function dual(con_ref::ConstraintRef{Model, <:_MOICON}; result::Int = 1)
+function dual(con_ref::ConstraintRef{<:AbstractModel, <:_MOICON}; result::Int = 1)
```
odow, line 713: Ditto here on `_constraint_dual`.

odow (commented):
It really only makes sense to relax type signatures if the internal function calls are also relaxed.

odow (approved):
Did CI run? This seems okay for now at least. @pulsipher, thoughts?

pulsipher:
This seems reasonable to me. It would seem that an extension implementing an `AbstractModel` would need to extend the underlying MOI methods then (e.g., `MOI.set`, `MOI.get`, etc.) right? It may be worth testing some of this explicitly via the CI tests that use `JuMPExtension.jl`, updating it to use `ConstraintRef`s with `MyModel` instead of `MyConstraintRef`.

odow:
They can also implement methods like `JuMP.dual` directly with a more-specific type. If they implement `backend`, things should work okay. 

I think this PR goes someway to reducing the number of methods they have to implement.

Feel free to suggest other changes that could make InfiniteOpt easier. But I don't think we should go through guessing which changes would make an extension easier if there is no downstream package using the changes.

blegat (approved):

PR #2441, opened 2021-01-29
[ci] Fix release branch wildcard for GH actions

odow:
I pushed 0.21.6 to the `release-0.21` branch, but CI didn't run. It appears GH actions doesn't use regex: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - master
-      - /^release-.*$/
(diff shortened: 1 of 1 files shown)


blegat (approved):

odow:
That seems to have worked: https://github.com/jump-dev/JuMP.jl/actions?query=workflow%3ACI+branch%3Arelease-0.21

PR #2442, opened 2021-01-30
Implemented coefficient for expressions

### LINKED ISSUE #2371: Feature request: implement getindex for expressions
I would like to access the objective coefficients yet it just possible to access the function as an expression.
My solution would be:
```julia
"""
    objective_coefficient(model::Model, variable::VariableRef)

Return the coefficient associated with `variable` in the objective function. See
also [`set_objective_coefficient`](@ref).

Note: this function will throw an error if a nonlinear objective is set.
"""
function objective_coefficient(model::Model, variable::VariableRef)
    return get(objective_function(model).terms, variable, 0.0)
end
```

Yet, I have no idea how quadratic objective functions should be handled and what to do if someone tries the following
```julia
model_1 = Model()
@variable(model_1, x >= 0)
@objective(model_1, Min, x)
model_2 = Model()
@variable(model_2, y >= 0)
@objective(model_2, Min, y)

objective_coefficient(model_1, y)
```

junglegobs:
Currently there is some discussion over whether to overload `getindex` or instead create a new function called `coefficient` whose use would be more transparent. Personally I would prefer the former - it makes sense to me, it's pretty syntactically and I can't see it being confused with something else you may try to do.

Edit(odow): Closes #2371

### DIFF
--- a/docs/src/expressions.md
+++ b/docs/src/expressions.md
@@ -267,7 +267,7 @@ julia> @variable(model, y)
 y
 
 julia> @expression(model, ex, 2*x*y + 3*x)
-2 x + 1
+2 x*y + 3 x
 
 julia> coefficient(ex, x, y)
 2.0
--- a/docs/src/expressions.md
+++ b/docs/src/expressions.md
@@ -132,6 +132,30 @@ julia> ex
 1
 ```
 
+### Coefficients
+
+Use [`coefficient`](@ref) to return the coefficient associated with a variable
+in an affine expression.
+
+```jldoctest
+julia> model = Model();
+
+julia> @variable(model, x)
+x
+
+julia> @variable(model, y)
+y
+
+julia> @expression(model, ex, 2x + 1)
+2 x + 1
+
+julia> coefficient(ex, x)
+2.0
+
+julia> coefficient(ex, y)
+0.0
+```
+
 ## Quadratic expressions
 
 Like affine expressions, there are four ways of constructing a quadratic
@@ -226,6 +250,33 @@ julia> drop_zeros!(ex)
 
 julia> ex
 x + 1
+```
+
+### Coefficients
+
+Use [`coefficient`](@ref) to return the coefficient associated with a pair of variables
+in a quadratic expression.
+
+```jldoctest
+julia> model = Model();
+
+julia> @variable(model, x)
+x
+
+julia> @variable(model, y)
+y
+
+julia> @expression(model, ex, 2*x*y + 3*x)
+2 x + 1
+
+julia> coefficient(ex, x, y)
+2.0
+
+julia> coefficient(ex, x, x)
+0.0
+
+julia> coefficient(ex, x)
+3.0
 ```
 
 ## Nonlinear expressions
--- a/docs/src/expressions.md
+++ b/docs/src/expressions.md
@@ -275,6 +275,9 @@ julia> coefficient(ex, x, y)
 julia> coefficient(ex, x, x)
 0.0
 
+julia> coefficient(ex, y, x)
+2.0
+
 julia> coefficient(ex, x)
 3.0
 ```
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -150,9 +150,11 @@ end
 Base.zero(::Type{GenericAffExpr{C,V}}) where {C,V} = GenericAffExpr{C,V}(zero(C), OrderedDict{V,C}())
 Base.one(::Type{GenericAffExpr{C,V}}) where {C,V}  = GenericAffExpr{C,V}(one(C), OrderedDict{V,C}())
 Base.zero(a::GenericAffExpr) = zero(typeof(a))
-Base.one( a::GenericAffExpr) =  one(typeof(a))
+Base.one(a::GenericAffExpr) = one(typeof(a))
 Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.constant), copy(a.terms))
 Base.broadcastable(a::GenericAffExpr) = Ref(a)
+Base.getindex(a::GenericAffExpr{C,V}, v::V) where {C,V} = get(a.terms, v, zero(V))
+Base.getindex(a::GenericAffExpr{C,V}, v1::V, v2::V) where {C,V} = zero(V)
 
 """
     drop_zeros!(expr::GenericAffExpr)
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -153,8 +153,8 @@ Base.zero(a::GenericAffExpr) = zero(typeof(a))
 Base.one(a::GenericAffExpr) = one(typeof(a))
 Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.constant), copy(a.terms))
 Base.broadcastable(a::GenericAffExpr) = Ref(a)
-Base.getindex(a::GenericAffExpr{C,V}, v::V) where {C,V} = get(a.terms, v, zero(V))
-Base.getindex(a::GenericAffExpr{C,V}, v1::V, v2::V) where {C,V} = zero(V)
+coefficient(a::GenericAffExpr{C,V}, v::V) where {C,V} = get(a.terms, v, zero(C))
+coefficient(a::GenericAffExpr{C,V}, v1::V, v2::V) where {C,V} = zero(C)
 
 """
     drop_zeros!(expr::GenericAffExpr)
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -153,6 +153,12 @@ Base.zero(a::GenericAffExpr) = zero(typeof(a))
 Base.one(a::GenericAffExpr) = one(typeof(a))
 Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.constant), copy(a.terms))
 Base.broadcastable(a::GenericAffExpr) = Ref(a)
+
+"""
+    coefficient(a::GenericAffExpr{C,V}, v::V) where {C,V}
+
+Return the coefficient associated with variable `v` in the affine expression `a`.
+"""
 coefficient(a::GenericAffExpr{C,V}, v::V) where {C,V} = get(a.terms, v, zero(C))
 coefficient(a::GenericAffExpr{C,V}, v1::V, v2::V) where {C,V} = zero(C)
 
--- a/src/quad_expr.jl
+++ b/src/quad_expr.jl
@@ -102,10 +102,24 @@ Base.zero(q::GenericQuadExpr) = zero(typeof(q))
 Base.one(q::GenericQuadExpr)  = one(typeof(q))
 Base.copy(q::GenericQuadExpr) = GenericQuadExpr(copy(q.aff), copy(q.terms))
 Base.broadcastable(q::GenericQuadExpr) = Ref(q)
-coefficient(q::GenericQuadExpr{C,V}, v::V) where {C,V} = coefficient(q.aff, v)
+
+"""
+    coefficient(a::GenericAffExpr{C,V}, v1::V, v2::V) where {C,V}
+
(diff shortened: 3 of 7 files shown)


src/aff_expr.jl
```diff
@@ -150,9 +150,11 @@ end
 Base.zero(::Type{GenericAffExpr{C,V}}) where {C,V} = GenericAffExpr{C,V}(zero(C), OrderedDict{V,C}())
 Base.one(::Type{GenericAffExpr{C,V}}) where {C,V}  = GenericAffExpr{C,V}(one(C), OrderedDict{V,C}())
 Base.zero(a::GenericAffExpr) = zero(typeof(a))
-Base.one( a::GenericAffExpr) =  one(typeof(a))
+Base.one(a::GenericAffExpr) = one(typeof(a))
 Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.constant), copy(a.terms))
 Base.broadcastable(a::GenericAffExpr) = Ref(a)
+Base.getindex(a::GenericAffExpr{C,V}, v::V) where {C,V} = get(a.terms, v, zero(V))
```
odow, line 156: `V` is the variable type and `C` is the coefficient type, so if our `get` fails, the default should be `zero(C)`, not `zero(V)`: `get(a.terms, v, zero(C))`.

src/aff_expr.jl
```diff
@@ -150,9 +150,11 @@ end
...
+Base.getindex(a::GenericAffExpr{C,V}, v1::V, v2::V) where {C,V} = zero(V)
```
odow, line 157: Ditto on the return value.

src/quad_expr.jl
```diff
@@ -102,6 +102,10 @@ Base.zero(q::GenericQuadExpr) = zero(typeof(q))
 Base.one(q::GenericQuadExpr)  = one(typeof(q))
 Base.copy(q::GenericQuadExpr) = GenericQuadExpr(copy(q.aff), copy(q.terms))
 Base.broadcastable(q::GenericQuadExpr) = Ref(q)
+Base.getindex(q::GenericQuadExpr{C,V}, v::V) where {C,V} = q.aff[v]
+function Base.getindex(q::GenericQuadExpr{C,V}, v1::V, v2::V) where {C,V}
+    return get(q.terms, UnorderedPair(v1,v2), zero(V))
```
odow, line 107: Ditto on the return type.

src/variables.jl
```diff
@@ -161,6 +161,8 @@ end
 Base.iszero(::VariableRef) = false
 Base.copy(v::VariableRef) = VariableRef(v.model, v.index)
 Base.broadcastable(v::VariableRef) = Ref(v)
+Base.getindex(v1::VariableRef, v2::VariableRef) = (v1 == v2 ? one(v1) : zero(v1))
+Base.getindex(v1::VariableRef, v2::VariableRef, v3::VariableRef) = zero(v1)
```
odow, line 165: These should be `v1 == v2 ? 1.0 : 0.0` and `= 0.0`. (We default to `Float64` here, because it isn't obvious what else to do. We can fix this in future if it becomes a problem.

test/expr.jl
```diff
@@ -154,6 +154,13 @@ function expressions_test(
         @test k == 0
     end
 
+    @testset "getindex(aff::AffExpr, v::VariableRef)" begin
+        m = ModelType()
+        x = @variable(m, x)
+        aff = @expression(m, 1.0 * x)
+        @test aff[x] == 1.0
+    end
```
odow, line 162: Some other tests we need 
- Affine
  - Checking that a missing term returns `0.0`
  - Checking that the quadratic term returns `0.0`
- Quadratic
  - `quad[x, y]` and `quad[y, x]` return the same
  - `quad[x]` returns the affine component
  - missing values return 0
- Variable
  - `x[x]` returns 1
  - `x[y]` returns 0
  - `x[x, y]` returns 0

odow (changes requested):
I think the `get` overload makes sense for affine and quadratic expressions. What doesn't really make sense is overloading it on `VariableRef`. 

`coefficient(expr, x)` is more typing that `expr[x]`, but it is more readable for new users. We can wait to see what other people think.

In the mean-time, we can fix the return values, and add some more tests.

blegat:
`coefficient` would be less confusing. It is also consistent with [MultivariatePolynomials](https://juliaalgebra.github.io/MultivariatePolynomials.jl/stable/types/#MultivariatePolynomials.coefficient)

junglegobs:
Thinking about this again, I also think `coefficient` would be less confusing, or at least the overload of `getindex` leaves more room for interpretation. You could always define `Base.getindex(a::GenericAffineExpr, args...) = coefficient(a,args...)` after if you really want a more compact syntax.

junglegobs:
I've been sitting on this, but if no objections I will implement @odow  s changes and change this to `coefficient(x,y,...)`.

test/expr.jl
junglegobs: Adding to this - if a variable does not exist in a model but you check an expression for its coefficient, should this throw an error?

joaquimg:
I agree than coefficient is much better than getindex

test/expr.jl
odow: You could use https://github.com/jump-dev/JuMP.jl/blob/ba17068a2f689e582604b8f030238fd19f5d8971/src/variables.jl#L155
but it's probably not important. It will return `0` anyway.

odow:
Yes please, implement `coefficient`.

junglegobs:
I think I should have pushed my changes the other day - is this ok now?

src/aff_expr.jl
```diff
@@ -150,9 +150,11 @@ end
 Base.zero(::Type{GenericAffExpr{C,V}}) where {C,V} = GenericAffExpr{C,V}(zero(C), OrderedDict{V,C}())
 Base.one(::Type{GenericAffExpr{C,V}}) where {C,V}  = GenericAffExpr{C,V}(one(C), OrderedDict{V,C}())
 Base.zero(a::GenericAffExpr) = zero(typeof(a))
-Base.one( a::GenericAffExpr) =  one(typeof(a))
+Base.one(a::GenericAffExpr) = one(typeof(a))
 Base.copy(a::GenericAffExpr) = GenericAffExpr(copy(a.constant), copy(a.terms))
 Base.broadcastable(a::GenericAffExpr) = Ref(a)
+coefficient(a::GenericAffExpr{C,V}, v::V) where {C,V} = get(a.terms, v, zero(C))
```
odow, line 156:
```Julia
"""
    coefficient(a::GenericAffExpr{C,V}, v::V) where {C,V}

Return the coefficient associated with variable `v` in the affine expression `a`.
"""
coefficient(a::GenericAffExpr{C,V}, v::V) where {C,V} = get(a.terms, v, zero(C))
```

src/aff_expr.jl
```diff
@@ -150,9 +150,11 @@ end
...
+coefficient(a::GenericAffExpr{C,V}, v1::V, v2::V) where {C,V} = zero(C)
```
odow, line 157: This one probably doesn't need a docstring, because it's really only there as a fallback.

src/quad_expr.jl
```diff
@@ -102,6 +102,10 @@ Base.zero(q::GenericQuadExpr) = zero(typeof(q))
 Base.one(q::GenericQuadExpr)  = one(typeof(q))
 Base.copy(q::GenericQuadExpr) = GenericQuadExpr(copy(q.aff), copy(q.terms))
 Base.broadcastable(q::GenericQuadExpr) = Ref(q)
+coefficient(q::GenericQuadExpr{C,V}, v::V) where {C,V} = coefficient(q.aff, v)
```
odow, line 105:
```Julia
"""
    coefficient(a::GenericQuadExpr{C,V}, v::V) where {C,V}

Return the coefficient associated with variable `v` in the affine component of `a`.
"""
```

src/quad_expr.jl
```diff
@@ -102,6 +102,10 @@ Base.zero(q::GenericQuadExpr) = zero(typeof(q))
...
+function coefficient(q::GenericQuadExpr{C,V}, v1::V, v2::V) where {C,V}
```
odow, line 106:
```Julia
"""
    coefficient(a::GenericAffExpr{C,V}, v1::V, v2::V) where {C,V}

Return the coefficient associated with the term `v1 * v2` in the quadratic expression `a`.

Note that `coefficient(a, v1, v2)` is the same as `coefficient(a, v2, v1)`.
"""
```

src/variables.jl
```diff
@@ -161,6 +161,8 @@ end
 Base.iszero(::VariableRef) = false
 Base.copy(v::VariableRef) = VariableRef(v.model, v.index)
 Base.broadcastable(v::VariableRef) = Ref(v)
+coefficient(v1::VariableRef, v2::VariableRef) = (v1 == v2 ? 1.0 : 0.0)
```
odow, line 164:
```Julia
"""
    coefficient(v1::VariableRef, v2::VariableRef)

Return `1.0` if `v1 == v2`, and `0.0` otherwise.

This is a fallback for other [`coefficient`](@ref) methods to simplify code in 
which the expression may be a single variable.
"""
```

odow (commented):
Looks good. We just need to add this function to the documentation.

* Add some docstrings like the ones I suggested. 
* Add `coefficient` to https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/reference/expressions.md (probably just after `constant`?)
* Add a section on `coefficient` to the manual: https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/expressions.md. Maybe something like this:
```
### Coefficients

Use [`coefficient`](@ref) to return the coefficient associated with a variable
in an affine expression.

```jldoctest
julia> model = Model();

julia> @variable(model, x)
x

julia> @variable(model, y)
y

julia> @expression(model, ex, 2x + 1)
2 x + 1

julia> coefficient(ex, x)
2.0

julia> coefficient(ex, y)
0.0
```
```

**Comment by junglegobs** [comment]
I implemented the docs pretty much exactly as you suggested. Thanks for all the help @odow!

When testing though I noticed this: 
```julia
Test Summary:          | Pass  Broken  Total
mutable_arithmetics.jl | 1776       6   1782
```
Which I don't remember before - but maybe I just wasn't paying attention?

docs/src/expressions.md
```diff
@@ -228,6 +252,33 @@ julia> ex
...
+julia> model = Model();
+
+julia> @variable(model, x)
+x
+
+julia> @variable(model, y)
+y
+
+julia> @expression(model, ex, 2*x*y + 3*x)
+2 x + 1
```
odow, line 270: This needs to be updated

odow (commented):
Tests are failing because of the doctest.

docs/src/expressions.md
junglegobs: You're right, but the mutable arithmetics tests were apparently failing before as well (I got confused with the two issues).

odow: Oh. I forgot to say, you can ignore the "Broken" tests. They are known to be broken. We only care about "Failed" tests, of which you now have none!

odow (approved):
I added one more test for `coefficient(aff, x, x)`. Assuming I didn't make a mistake and the tests pass, this is good to merge.

Thanks for your help!

junglegobs:
@odow thank you for yours!

docs/src/expressions.md
```diff
@@ -228,6 +252,33 @@ julia> ex
...
+
+julia> @variable(model, y)
+y
+
+julia> @expression(model, ex, 2*x*y + 3*x)
+2 x*y + 3 x
+
+julia> coefficient(ex, x, y)
+2.0
+
```
blegat, line 274: Maybe add `coefficient(y, x)` here to show that it also gives `2.0`

blegat (approved):

odow:
Thanks!

PR #2443, opened 2021-01-31
Refactor the examples by categorizing them into broader themes.

odow:
Doc preview: https://jump.dev/JuMP.jl/previews/PR2443/examples/mixed-integer-linear-programs/

Our current examples are just a large list. This makes it hard to figure out which ones are MILP/SDP/nonlinear etc. 

image

This PR copy-pastes the existing examples into broader themes (e.g., mixed-integer linear programs, nonlinear programs, and semi-definite programs) to make it easier to find ones of interest.

image


blegat:
Why do they all have to be in the same files ? Can't we do subdirs as in https://github.com/jump-dev/JuMP.jl/pull/2444 ?

odow:
I did this first, then found the subdir trick worked quite well! Here's the alternative: https://github.com/jump-dev/JuMP.jl/pull/2445

odow:
Closing in favor of #2445

PR #2444, opened 2021-02-01
Migrate non-modeling tutorials from JuMPTutorials.jl

odow:
I've just done the easy ones for now. 

I've copied them across with some minimal edits, mainly just to get them to compile and to remove obvious typos. 

Once we have the tutorials added, we can do an edit of the documentation to simplify things. (There is a fair bit of overlap between the Manual and the tutorials.)

Preview link: https://jump.dev/JuMP.jl/previews/PR2444/tutorials/Getting%20started/an_introduction_to_julia/


docs/src/tutorials/Getting started/an_introduction_to_julia.jl
```diff
@@ -0,0 +1,400 @@
...
+#-
+
+A * x == b
+
+# Note that when multiplying vectors and matrices, dimensions matter. For
+# example, you can't multiply a vector by a vector:
+
+try  #hide
+b * b
+catch err; showerror(stderr, err); end  #hide
```
odow, line 114: I had to use a hack to capture errors in the output. Here's the upstream https://github.com/JuliaDocs/Documenter.jl/issues/1420#issuecomment-770539595

docs/src/tutorials/Getting started/variables_constraints_objective.jl
```diff
@@ -0,0 +1,252 @@
...
+
+c = [1; 3; 5; 2]
+
+@variable(vector_model, x[1:4] >= 0)
+@constraint(vector_model, A * x .== b)
+@objective(vector_model, Min, c' * x)
+
+optimize!(vector_model)
+
+@show objective_value(vector_model);
```
blegat, line 252: We probably don't need the `@show` with Literate

mlubin:
The files seem to be missing license headers.

odow:
> The files seem to be missing license headers.

I've added a MIT license to the `/docs/src/tutorials` directory:
https://github.com/jump-dev/JuMP.jl/pull/2444/files#diff-13cf917ed803de6205839a3c013e751ae12de2afb78564cd279a6295d06fb3ce

I think these ones so far are okay, because they were part of JuMPTutorials. We might have to track down the original author/license of the tutorials originally contributed to `juliaopt-notebooks`.

Alternatively, we could get the authors of the tutorials I've copied (who are added as co-authors in the commits) to approve relicensing them under JuMP's MPL 2 license.

Thoughts?

mlubin:
Each file needs to have a copyright statement and license header (not necessarily a full copy of the MIT license).

I think an MIT license makes more sense for tutorial code than MPL.

odow:
I was under the impression that the MIT only required having `LICENSE.md` in the root of the directory licensed under MIT.

odow:
I'll leave this here for now. Once merged, I can go back through each tutorial with some improvements.

mlubin:
It's good practice to have the headers in each file, especially for files like these where it's very likely someone will be viewing the source and copy-pasting the code.

odow:
Ah, true. I've added the headers.

odow:
There are a few print failures:

image

blegat (approved):

odow:
Objections to merging? The printing will be fixed in a future PR, and merging will allow me to go through and tidy up individual tutorials in smaller diffs.

joaquimg:
+1 for merging and improve in smaller diffs

Issue #2446, opened 2021-02-02
Fix TODOs in print.jl


odow:
There are a number of TODO items in `print.jl`

- [ ] https://github.com/jump-dev/JuMP.jl/blob/2892b68cdd15b1632f9f9f6d5ab396484754e40b/src/print.jl#L165
- [ ] https://github.com/jump-dev/JuMP.jl/blob/2892b68cdd15b1632f9f9f6d5ab396484754e40b/src/print.jl#L573
- [ ] https://github.com/jump-dev/JuMP.jl/blob/2892b68cdd15b1632f9f9f6d5ab396484754e40b/src/print.jl#L574-L575
- [ ] https://github.com/jump-dev/JuMP.jl/blob/2892b68cdd15b1632f9f9f6d5ab396484754e40b/src/print.jl#L694
- [ ] https://github.com/jump-dev/JuMP.jl/blob/2892b68cdd15b1632f9f9f6d5ab396484754e40b/src/print.jl#L703

These cause printing issues, especially in `IJuliaMode` for the documentation: https://github.com/jump-dev/JuMP.jl/pull/2444#issuecomment-771318608

PR #2447, opened 2021-02-02
Fix printing in IJuliaMode

odow:
A few printing changes that were needed to fix the printing issues in the tutorials: https://github.com/jump-dev/JuMP.jl/pull/2444#issuecomment-771318608

Fixes some of the points in #2446 

* IJulia strings now created as a single line without `\n`. Makes no difference to latex, but Documenter struggled to print these.
* `\\begin{alignat*}{1}` becomes `\\begin{aligned}`
* `x binary` becomes `x ∈ {0, 1}`
* `x integer` becomes `x ∈ Z`
* Sets without a nice fallback are printed in `\text{}` instead of math. 
* Constraints, by themselves, without names, are printed as display equations instead of in-line.

I'll post before/after screenshots from IJulia in a bit.

## Before

image

## After 

image

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -591,6 +591,12 @@ function in_set_string(print_mode, set::Union{PSDCone, MOI.AbstractSet})
     end
 end
 
+"""
+    in_set_string(print_mode::Type{<:PrintMode}, constraint::AbstractConstraint)
+
+Return a `String` representing the membership to the set of the constraint
+`constraint` using print mode `print_mode`.
+"""
 function in_set_string(print_mode, constraint::AbstractConstraint)
     set = reshape_set(moi_set(constraint), shape(constraint))
     return in_set_string(print_mode, set)
--- a/src/print.jl
+++ b/src/print.jl
@@ -201,7 +201,7 @@ end
 function model_string(print_mode, model::AbstractModel)
     ijl = print_mode == IJuliaMode
     sep = ijl ? " & " : " "
-    eol = ijl ? "\\\\\n" : "\n"
+    eol = ijl ? "\\\\" : "\n"
     sense = objective_sense(model)
     str = ""
     if sense == MOI.MAX_SENSE
@@ -241,7 +241,7 @@ function model_string(print_mode, model::AbstractModel)
         str *= eol
     end
     if ijl
-        str = "\\begin{alignat*}{1}" * str * "\\end{alignat*}\n"
+        str = "\\begin{aligned}" * str * "\\end{aligned}"
     end
     return str
 end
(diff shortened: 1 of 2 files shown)


src/print.jl
```diff
@@ -201,7 +201,7 @@ end
 function model_string(print_mode, model::AbstractModel)
     ijl = print_mode == IJuliaMode
     sep = ijl ? " & " : " "
-    eol = ijl ? "\\\\\n" : "\n"
+    eol = ijl ? "\\\\" : "\n"
```
blegat, line 204: Is this due to a bug in Documenter that is reported somewhere or a known limitation ?

src/print.jl
```diff
@@ -570,27 +578,19 @@ end
...
+        set_str = replace(replace(string(set), "{" => "\\{"), "}" => "\\}")
+        return "\\in \\text{$(set_str)}"
+    end
 end
 
-"""
-    in_set_string(print_mode::Type{<:JuMP.PrintMode},
-                  constraint::JuMP.AbstractConstraint)
-
-Return a `String` representing the membership to the set of the constraint
```
blegat, line 591: Why removing this docstring ?

src/print.jl
```diff
@@ -549,6 +549,14 @@ function function_string(print_mode, constraint::AbstractConstraint)
     return function_string(print_mode, f)
 end
 
+"""
+    in_set_string(print_mode::Type{<:PrintMode}, set)
+
+Return a `String` representing the membership to the set `set` using print mode
+`print_mode`.
+"""
+function in_set_string end
```
odow, line 558: @blegat, there is not a generic docstring.

src/print.jl
odow: It was the same error as this: https://github.com/JuliaDocs/Documenter.jl/issues/1387.

Looking through your PR, I guess the regex needs a suffix `"s` so it can match multiline strings:
https://github.com/JuliaDocs/Documenter.jl/pull/1426/files#diff-c34d581afcb6837fd39a6b0b0428090c049d7dad13de417d96dc08a0ed01cf70R1751-R1752

odow: https://github.com/JuliaDocs/Documenter.jl/pull/1518

src/print.jl
blegat: Yes but the other case is not covered since the second argument is a constraint, not a set.

src/print.jl
odow: So the newline issue can be fixed upstream, but there is still the issue that we are using `alignat` instead of `aligned`. With this PR, the tutorials print nicely.

blegat (approved):

PR #2449, opened 2021-02-04
Refactor printing to reduce content printed in IJulia mode

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/957

Previous behavior was inconsistent between the REPL and IJulia/documentation.

This PR standardizes on the following
 * `Julia> model` or ending a cell with `model` prints the summary
 * `print(io, model)` prints the plain-text or latex version as a string to io
 * `print(model)` prints the latex version in IJulia, and the plain-text elsewhere.
 * `latex_formulation(model)` returns a type that will `show` the latex formulation in Documenter and IJulia, and the string of it in the REPL.

The user-visible change is that ending a cell with `model` prints the summary prints the plain-text summary instead of rendering the latex.

<s>We should probably merge #2447 first, and then rebase this PR on top.</s>

## IJulia

image

## REPL

```Julia
julia> model
A JuMP Model
Minimization problem with:
Variable: 1
Objective function type: GenericAffExpr{Float64,VariableRef}
`GenericAffExpr{Float64,VariableRef}`-in-`MathOptInterface.GreaterThan{Float64}`: 1 constraint
`VariableRef`-in-`MathOptInterface.GreaterThan{Float64}`: 1 constraint
Model mode: AUTOMATIC
Names registered in the model: x

julia> function foo(model)
           print(model)
           return
       end
foo (generic function with 1 method)

julia> foo(model)
Min x - 1
Subject to
 2 x ≥ 1.0
 x ≥ 0.0

julia> latex_formulation(model)
$$ \begin{aligned}\min\quad & x - 1\\\text{Subject to} \quad & 2 x \geq 1.0\\ & x \geq 0.0\\\end{aligned} $$
```

### DIFF
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -92,7 +92,8 @@ Solver name: No optimizer attached.
 Names registered in the model: x
 ```
 
-Use [`print(::AbstractModel)`](@ref) to print the formulation of the model.
+Use `print` to print the formulation of the model (in IJulia, this will render
+as LaTeX.
 ```jldoctest model_print
 julia> print(model)
 Max x
@@ -104,15 +105,20 @@ Subject to
     [`write_to_file`](@ref) instead.
 
 
-Pass `latex = true` to print the model in LaTeX form.
-```julia
-julia> print(model; latex = true)
+Use [`latex_formulation`](@ref) to display the model in LaTeX form.
+```jldoctest model_print
+julia> latex_formulation(model)
 $$ \begin{aligned}\max\quad & x\\\text{Subject to} \quad & x \geq 0.0\\\end{aligned} $$
 ```
 
-In IJulia (and Documenter), this will render as LaTeX!
-```@example getting_started_with_JuMP
-print(model; latex = true)
+In IJulia (and Documenter), ending a cell in with [`latex_formulation`](@ref)
+will render the model in LaTeX!
+```@example
+using JuMP                # hide
+model = Model()           # hide
+@variable(model, x >= 0)  # hide
+@objective(model, Max, x) # hide
+latex_formulation(model)
 ```
 
 ## Turn off output
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -105,7 +105,7 @@ Subject to
 
 
 Pass `latex = true` to print the model in LaTeX form.
-```jldoctest model_print; filter=["$$ \begin{aligned}\max\quad & x\\\text{Subject to} \quad & x \geq 0.0\\\end{aligned} $$"]
+```jldoctest model_print; filter=[raw"$$ \begin{aligned}\max\quad & x\\\text{Subject to} \quad & x \geq 0.0\\\end{aligned} $$"]
 julia> print(model; latex = true)
 $$ \begin{aligned}\max\quad & x\\\text{Subject to} \quad & x \geq 0.0\\\end{aligned} $$
 ```
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -105,7 +105,7 @@ Subject to
 
 
 Pass `latex = true` to print the model in LaTeX form.
-```jldoctesst model_print
+```jldoctest model_print
 julia> print(model; latex = true)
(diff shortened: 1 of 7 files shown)


blegat:
Printing the LaTeX by default in notebook might ibe inappropriate indeed as it might force the user to interrupt if the model is too large.

odow:
> Printing the LaTeX by default in notebook might ibe inappropriate indeed as it might force the user to interrupt if the model is too large.

This is exactly the complaint in #957. 

Now `show` prints the summary, and they can go `print(model; latex = false)` instead.

odow:
@mlubin any objections to the design?

src/print.jl
```diff
@@ -27,6 +27,62 @@ abstract type PrintMode end
...
+    print(io, model; latex = true)
+    print(model; latex = true)
+"""
+function Base.print(io::IO, model::AbstractModel; latex::Bool = false)
+    return latex ? _print_latex(io, model) : _print_model(io, model)
+end
+
+function Base.print(model::AbstractModel; latex::Union{Nothing,Bool} = nothing)
+    if latex === nothing
+        return _ModelPrinter(model)
```
blegat, line 80: If `print(model)` is called from inside a function (so the output is not returned to the REPL or Jupyter), then this won't print anything right ? So shouldn't this be `show(_ModelPrinter(model))` ?

odow: Ah. Maybe `Base.print` is the wrong name for this. Calling `show` explicitly won't work. We need an object with `show` overloaded so that IJulia can pick the `text/latex` MIME type. If we call `show` directly, it will just print the plain-text.

blegat: Maybe make it more explicit, e.g. the user should call `ModelPrinter(model)`.or `pretty_printing(model)`. `pretty_print` is missleading for the same reason `print` is.

odow: How's this now? I added `LaTeXify`.

src/print.jl
```diff
@@ -27,6 +27,46 @@ abstract type PrintMode end
...
+
+Print the plain-text formulation of `model` to `io` as a string (or to `stdout`
+if `io` is not given).
+
+See also: [LaTeXify](@ref).
+"""
+Base.print(io::IO, model::AbstractModel) = _print_model(io, model)
+
+"""
+    LaTeXify(model)
```
odow, line 43: Overload `display` instead.

odow:
We should look to `DataFrames` as an example.

odow:
Doc section: https://docs.julialang.org/en/v1/manual/types/#man-custom-pretty-printing

There are `IOContext` attributes like `:compact`:
```Julia
function Base.show(io::IO, z::Polar)
    if get(io, :compact, false)
        print(io, z.r, "ℯ", z.Θ, "im")
    else
        print(io, z.r, " * exp(", z.Θ, "im)")
    end
end
```
DataFrames uses them to decide how much to print https://github.com/JuliaData/DataFrames.jl/blob/6d1347983b67e93d50df847d2d8546a4c9a8cfd4/src/abstractdataframe/io.jl#L263-L264

but this doesn't help the IJulia thing.

Motivation for current code is "if you're printing and passing `io`, you want the string. If you're just calling `print`, do the best thing possible."

docs/src/manual/models.md
```diff
@@ -74,6 +74,44 @@ CachingOptimizer state: EMPTY_OPTIMIZER
...
+Subject to
+ x ≥ 0.0
+```
+!!! warning
+    This format is specific to JuMP. To write the model to a file, use
+    [`write_to_file`](@ref) instead.
+
+
+Pass `latex = true` to print the model in LaTeX form.
+```jldoctesst model_print
```
blegat, line 108: tesst -> test

odow: Fixed. You would think this should throw an error...

odow:
Okay. I've revised how we do this yet again. The hard part is that IJulia and Documenter don't work the same way. IJulia adds a new `AbstractDisplay`, where as Documenter just captures the first `TextDisplay`. 

`print` now prints the Latex automatically on IJulia (but the plain text in Documenter), but you can use `latex_formulation` to get the latex in IJulia and documenter (and the string of it at the REPL).

blegat (approved):

Issue #1884, opened 2019-03-01
MethodError: Unexpected keyword argument to at-variable


joehuchette:
Likely a macro hygiene issue:
```jl
using JuMP
m = Model()
v = @variable(m, variable_type=:foo_bar)
```
gives
```
ERROR: MethodError: no method matching build_variable(::getfield(JuMP, Symbol("#_error#66")){Tuple{Symbol,Expr}}, ::VariableInfo{Float64,Float64,Float64,Float64}, ::Symbol)
Closest candidates are:
  build_variable(::Function, ::VariableInfo; extra_kw_args...) at /Users/huchette/.julia/packages/JuMP/jnmGG/src/macros.jl:1079
Stacktrace:
 [1] top-level scope at /Users/huchette/.julia/packages/JuMP/jnmGG/src/macros.jl:300
```
This is on julia v1.1 and JuMP v0.19.

odow:
A related issue:
```Julia
julia> @variable(model; variable_type = :foo_bar)
ERROR: MethodError: no method matching _valid_model(::Expr; variable_type=:foo_bar)
Closest candidates are:
  _valid_model(::Any, ::Any) at /Users/oscar/.julia/dev/JuMP/src/macros.jl:39 got unsupported keyword argument "variable_type"
  _valid_model(::AbstractModel, ::Any) at /Users/oscar/.julia/dev/JuMP/src/macros.jl:38 got unsupported keyword argument "variable_type"
Stacktrace:
 [1] top-level scope at REPL[10]:1
```

PR #2451, opened 2021-02-05
Add nice fallback for build_variable

odow:
Closes Part I of #1884 (The related issue is still unresolved.)

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -998,14 +998,35 @@ _esc_non_constant(x) = esc(x)
 # Example: `@variable m [1:3] foo` will allocate an vector of element type `variable_type(m, foo)`
 # Note: it needs to be implemented by all `AbstractModel`s
 variable_type(m::Model) = VariableRef
-# Returns a new variable. Additional positional arguments can be used to dispatch the call to a different method.
-# The return type should only depends on the positional arguments for `variable_type` to make sense. See the @variable macro doc for more details.
-# Example: `@variable m x` foo will call `build_variable(_error, info, foo)`
+
+"""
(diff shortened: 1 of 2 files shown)


joaquimg:
Just commented here https://github.com/jump-dev/JuMP.jl/pull/2421#issuecomment-773802642 why I think we should not have multiple positional args.

src/macros.jl
```diff
@@ -998,16 +998,37 @@ _esc_non_constant(x) = esc(x)
...
+function build_variable(_error::Function, ::VariableInfo, args...; kwargs...)
+    _error(
+        "Unrecognized arguments: $(join(args, ", ")). (You may have passed " *
+        "these as positional arguments, or as a keyword value to " *
+        "`variable_type`.)\n\nIf you're trying to create a JuMP extension, " *
+        "you need to implement `build_variable`."
+    )
+end
+
 function build_variable(_error::Function, variable::ScalarVariable,
```
joaquimg, line 1032: Note that this `build_variable` is not documented, and as in the comment in https://github.com/jump-dev/JuMP.jl/pull/2421#issuecomment-754406314

It could be `build_variabe_in_set`. It would be probably breaking though.

It might be even good not to document to keep it less breaking.

joaquimg: it seems you hit a similar problem in: https://github.com/jump-dev/JuMP.jl/pull/2450

odow: Yeah, it's staying for now. It should be obvious that it should change if we refactor the macros.

blegat (approved):

PR #2453, opened 2021-02-06
[docs] Update done items on roadmap

odow:
Implemented in Gurobi: https://github.com/jump-dev/Gurobi.jl/pull/392
Documentation at the JuMP level already exists: https://jump.dev/JuMP.jl/stable/solutions/#Multiple-solutions


joaquimg (approved):
🎉

mlubin:
I'm a bit confused about the documentation. Are all solutions returned by a solver guaranteed to be optimal? I know that some solvers can return suboptimal solutions found during the solve.

odow:
Hmm. Yeah I don't think the documentation should say `an_optimal_solution`. The solution in `result = 1` is probably the optimal solution, and the others are sub-optimal, feasible solutions. You could figure that out by looking at the objective value?

odow:
So does this count as done now?

mlubin (approved):

PR #2454, opened 2021-02-07
[docs] Tweak wording of multiple results documentation

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/2453#issuecomment-774561885

### DIFF
--- a/docs/src/solutions.md
+++ b/docs/src/solutions.md
@@ -323,6 +323,10 @@ Functions for querying the solutions, e.g., [`primal_status`](@ref) and
 [`value`](@ref), all take an additional keyword argument `result` which can be
 used to specify which result to return.
 
+!!! warning
+    Some of the returned solutions may be suboptimal! You should check 
+    [`objective_value`](@ref) to assess the quality of each returned solution.
+
 ```julia
 using JuMP
 model = Model()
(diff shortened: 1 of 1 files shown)


docs/src/solutions.md
```diff
@@ -323,6 +323,10 @@ Functions for querying the solutions, e.g., [`primal_status`](@ref) and
 [`value`](@ref), all take an additional keyword argument `result` which can be
 used to specify which result to return.
 
+!!! warning
+    Some of the returned solutions may be suboptimal! You should check
```
mlubin, line 327: We implicitly make a guarantee that the first result is optimal (if any one is), because that's what users get when they call `value` without a `result` keyword. Something would be very wrong if `value(x)` corresponded to a suboptimal solution but `value(x, result = 2)` was the optimal solution. Should we make this more explicit? I don't really see it discussed in MOI either, but maybe I missed it.

odow: Gurobi have this guarantee: https://www.gurobi.com/documentation/9.1/refman/xn.html#attr:Xn

> When SolutionNumber is equal to its default value of 0, querying attribute Xn is equivalent to querying attribute X.

I don't think we discussed it in MOI either. But we should make this more explicit here, and in the docstrings of MOI.

mlubin (approved):

PR #2466, opened 2021-02-15
First pass at a feasibility checker.

### LINKED ISSUE #693: An efficient way to verify the feasibility of provided solution

odow:
This is a significant item for JuMP 1.0, but there is debate on how
to define and implement distances for different sets. 

See https://github.com/jump-dev/MathOptInterface.jl/pull/1023 for the 
last time this was attempted.

See https://github.com/matbesancon/MathOptSetDistances.jl for a WIP
package exploring the options.

This PR also only addresses primal feasibility. See
https://github.com/joaquimg/FeasibilityOptInterface.jl/pull/1
for a WIP attempt at dual feasibility as well.

In future, the _distance_to_set definitions will be merged into MOI.

This PR is more about defining the JuMP-level interface so we can
change internal details without breaking compatibility of JuMP 1.0.

Edit: for example, if MOI introduces a `AbstractDistance` argument, 
we can add a `; distance = DefaultDistance()` kwarg to `primal_feasibility_report`
without violating compatibility.

For now I've implemented sets for MILP, since this should cover 90% of 
users wishing to implement this. Perhaps for the trickier sets, we can tell
users to load MathOptSetDistances, and that package can contain heavier
dependencies for computing the minimum eigen-value, etc.

Closes #693 

Doc preview: https://jump.dev/JuMP.jl/previews/PR2466/manual/solutions/#Checking-feasibility-of-solutions

### DIFF
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -493,21 +493,31 @@ end
 ## Checking feasibility of solutions
 
 To check the feasibility of a primal solution, use
-[`primal_feasibility_report`](@ref). The returned `report` is `nothing` if the
-point is feasible. Otherwise, it is a dictionary mapping the infeasible
-constraint references to the distance between the point and the nearest point in
-the set.
+[`primal_feasibility_report`](@ref), which takes a `model`, a dictionary mapping
+each variable to a primal solution value, and a tolerance `atol`. If a variable
+is not given in dictionary, the value is assumed to be `0.0`.
 
-```@example
+The function returns a dictionary which maps the infeasible constraint
+references to the distance between the point and the nearest point in the
+corresponding set. A point is classed as infeasible if the distance is greater
+than a supplied tolerance `atol`, and feasible otherwise.
+```@example feasibility
 using JuMP
 model = Model()
 @variable(model, x >= 1, Int)
-@constraint(model, c1, x <= 1.95)
+@variable(model, y)
+@constraint(model, c1, x + y <= 1.95)
 point = Dict(x => 2.5)
-report = primal_feasibility_report(model, point)
+report = primal_feasibility_report(model, point; atol = 1e-6)
 ```
 
-To use the point from a previous solve, use:
+If the point is feasible, this function returns `nothing`.
+```@example feasibility
+point = Dict(x => 1.0)
+report = primal_feasibility_report(model, point; atol = 1e-6)
+```
+
+To obtain the primal point from a previous solve, use:
 ```julia
 point = Dict(v => value(v) for v in all_variables(model))
 ```
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -501,9 +501,10 @@ The function returns a dictionary which maps the infeasible constraint
 references to the distance between the point and the nearest point in the
 corresponding set. A point is classed as infeasible if the distance is greater
 than a supplied tolerance `atol`, and feasible otherwise.
+
 ```@example feasibility
-using JuMP
-model = Model()
+using JuMP, GLPK
+model = Model(GLPK.Optimizer)
 @variable(model, x >= 1, Int)
 @variable(model, y)
 @constraint(model, c1, x + y <= 1.95)
@@ -511,14 +512,14 @@ point = Dict(x => 2.5)
 report = primal_feasibility_report(model, point; atol = 1e-6)
 ```
 
-If the point is feasible, this function returns `nothing`.
+If the point is feasible, this function returns an empty dictionary.
 ```@example feasibility
 point = Dict(x => 1.0)
 report = primal_feasibility_report(model, point; atol = 1e-6)
-report === nothing
 ```
 
-To obtain the primal point from a previous solve, use:
-```julia
-point = Dict(v => value(v) for v in all_variables(model))
+To use the primal solution from a solve, omit the `point` argument:
+```@example feasibility
+optimize!(model)
+report = primal_feasibility_report(model)
 ```
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -502,7 +502,7 @@ references to the distance between the primal value of the constraint and the
 nearest point in the corresponding set. A point is classed as infeasible if the
 distance is greater than the supplied tolerance `atol`.
 
-```jldoctest feasibility
+```jldoctest feasibility; filter=["x integer         => 0.1", "c1 : x + y ≤ 1.95 => 0.01"]
 julia> model = Model(GLPK.Optimizer);
 
 julia> @variable(model, x >= 1, Int);
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -127,7 +127,7 @@ FEASIBLE_POINT::ResultStatusCode = 1
 ```
 Other common returns are `MOI.NO_SOLUTION`, and `MOI.INFEASIBILITY_CERTIFICATE`.
 The first means that the solver doesn't have a solution to return, and the
-second means that the primal solution is a certificate of dual infeasbility (a 
+second means that the primal solution is a certificate of dual infeasbility (a
 primal unbounded ray).
 
 You can also use [`has_values`](@ref), which returns `true` if there is a
@@ -200,7 +200,7 @@ FEASIBLE_POINT::ResultStatusCode = 1
 ```
 Other common returns are `MOI.NO_SOLUTION`, and `MOI.INFEASIBILITY_CERTIFICATE`.
 The first means that the solver doesn't have a solution to return, and the
-second means that the dual solution is a certificate of primal infeasbility (a 
+second means that the dual solution is a certificate of primal infeasbility (a
 dual unbounded ray).
 
 You can also use [`has_duals`](@ref), which returns `true` if there is a
@@ -258,7 +258,7 @@ And data, a 2-element Array{Float64,1}:
 
 ## Recommended workflow
 
-The recommended workflow for solving a model and querying the solution is 
+The recommended workflow for solving a model and querying the solution is
 something like the following:
 ```jldoctest solutions
 if termination_status(model) == MOI.OPTIMAL
@@ -427,7 +427,7 @@ For instance, this is how you can use this functionality:
 
 ```julia
 using JuMP
-model = Model() # You must use a solver that supports conflict refining/IIS 
+model = Model() # You must use a solver that supports conflict refining/IIS
 # computation, like CPLEX or Gurobi
 @variable(model, x >= 0)
 @constraint(model, c1, x >= 2)
@@ -488,4 +488,26 @@ for i in 2:result_count(model)
         print("Solution $(i) is also optimal!")
     end
 end
+```
+
+## Checking feasibility of solutions
+
+To check the feasibility of a primal solution, use
+[`primal_feasibility_report`](@ref). The returned `report` is `nothing` if the
+point is feasible. Otherwise, it is a dictionary mapping the infeasible
+constraint references to the distance between the point and the nearest point in
+the set.
+
+```@example
+using JuMP
+model = Model()
+@variable(model, x >= 1, Int)
+@constraint(model, c1, x <= 1.95)
+point = Dict(x => 2.5)
+report = primal_feasibility_report(model, point)
+```
+
+To use the point from a previous solve, use:
+```julia
+point = Dict(v => value(v) for v in all_variables(model))
 ```
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -516,8 +516,8 @@ julia> point = Dict(x => 1.9, y => 0.06);
 
 julia> primal_feasibility_report(model, point)
 Dict{Any,Float64} with 2 entries:
-  x integer         => 0.1
   c1 : x + y ≤ 1.95 => 0.01
+  x integer         => 0.1
 
 julia> primal_feasibility_report(model, point; atol = 0.02)
 Dict{Any,Float64} with 1 entry:
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -502,7 +502,11 @@ references to the distance between the primal value of the constraint and the
 nearest point in the corresponding set. A point is classed as infeasible if the
 distance is greater than the supplied tolerance `atol`.
 
-```jldoctest feasibility; filter=["x integer         => 0.1", "c1 : x + y ≤ 1.95 => 0.01"]
+```@meta
+# Add a filter here because the output of the dictionary is not ordered, and
+# changes in printing order will cause the doctest to fail.
+```
+```jldoctest feasibility; filter=[r"x.+?\=\> 0.1", r"c1.+? \=\> 0.01"]
 julia> model = Model(GLPK.Optimizer);
 
 julia> @variable(model, x >= 1, Int);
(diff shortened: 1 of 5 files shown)


src/feasibility_checker.jl
```diff
@@ -0,0 +1,141 @@
...
+constraint to the distance it is from being feasible.
+
+To obtain `point` from a solution of a solved model, use:
+```julia
+point = Dict(v => value(v) for v in all_variables(model))
+```
+"""
+function primal_feasibility_report(
+    model::Model,
+    point::Dict{VariableRef,Float64};
```
joaquimg, line 30: This could be an AbstractDict.

src/feasibility_checker.jl
```diff
@@ -0,0 +1,141 @@
...
+)
+    point_f = x -> get(point, x, 0.0)
+    violated_constraints = Dict{Any,Float64}()
+    for (F, S) in list_of_constraint_types(model)
+        # This loop is not type-stable because `F` and `S` change, but it should
+        # be fine because no one should be using this in performance-critical
+        # code.
+        for con in all_constraints(model, F, S)
+            obj = constraint_object(con)
+            d = _distance_to_set(value.(obj.func, point_f), obj.set)
```
joaquimg, line 41: I'd merge these two into a function:
`constraint_feasibility(con, point_f; atol)`

src/feasibility_checker.jl
```diff
@@ -0,0 +1,141 @@
...
+    set::MOI.Nonpositives,
+) where {T<:Real}
+    _check_dimension(x, set)
+    return LinearAlgebra.norm(max(xi, zero(T)) for xi in x)
+end
+
+function _distance_to_set(
+    x::Vector{T},
+    set::MOI.Zeros
+) where {T<:Real}
```
joaquimg, line 128: this can be complex...

src/feasibility_checker.jl
```diff
@@ -0,0 +1,141 @@
...
+
+function _distance_to_set(x::T, set::MOI.LessThan{T}) where {T<:Real}
+    return max(x - set.upper, zero(T))
+end
+
+function _distance_to_set(x::T, set::MOI.GreaterThan{T}) where {T<:Real}
+    return max(set.lower - x, zero(T))
+end
+
+function _distance_to_set(x::T, set::MOI.EqualTo{T}) where {T<:Real}
```
joaquimg, line 69: This can be complex

joaquimg (commented):
Great work!

Does it work for non-linear?

+1 to move `_distance_to_set` to MOI. SOC is very simple. SDP can be done with julia base, even full eig decomposition is usually faster than any solver (some even do full eig decompositions in each iteration).

src/feasibility_checker.jl
odow: At one point I had exactly `_distance_to_set(::ConstraintRef, point_f::Function)`, but I went with this for simplicity. We can refactor it out once we agree on the high-level API.

odow:
No, it doesn't work for nonlinear yet. But I could add if there is interest. For the other cones, I vote that we leave them for now. They can be added to MOI later. This is more about getting the API for JuMP sorted. Adding extra `distance_to_set` definitions is non-breaking.

docs/src/solutions.md
```diff
@@ -353,3 +353,25 @@ for i in 2:result_count(model)
     end
 end
 ```
+
+## Checking feasibility of solutions
+
+To check the feasibility of a primal solution, use
+[`primal_feasibility_report`](@ref). The returned `report` is `nothing` if the
+point is feasible. Otherwise, it is a dictionary mapping the infeasible
```
mlubin, line 361: What do "feasible" and "infeasible" mean? I don't think you can ignore discussion of tolerances here.

src/feasibility_checker.jl
```diff
@@ -0,0 +1,141 @@
...
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+using LinearAlgebra
+
+"""
+    primal_feasibility_report(
+        model::Model,
+        [point::AbstractDict{VariableRef,Float64}];
+        atol::Float64 = 1e-8,
```
mlubin, line 12: atol is not documented

src/feasibility_checker.jl
```diff
@@ -0,0 +1,141 @@
...
+
+To obtain `point` from a solution of a solved model, use:
+```julia
+point = Dict(v => value(v) for v in all_variables(model))
+```
+"""
+function primal_feasibility_report(
+    model::Model,
+    point::AbstractDict{VariableRef,Float64};
+    atol::Float64 = 1e-8,
```
mlubin, line 31: Does it make sense to have a default tolerance here? Any value we put here means that in some cases, a solver will return a solution that it thinks is feasible, and `primal_feasibility_report` will declare it infeasible. I would either force the user to provide a tolerance, or just return all infeasibilities greater than zero and let the caller decide how to filter them.

docs/src/solutions.md
odow: I had a go at revising.

src/feasibility_checker.jl
```diff
@@ -0,0 +1,161 @@
...
+function primal_feasibility_report(
+    model::Model,
+    point::AbstractDict{VariableRef,Float64};
+    atol::Float64,
+)
+    point_f = x -> get(point, x, 0.0)
+    violated_constraints = Dict{Any,Float64}()
+    for (F, S) in list_of_constraint_types(model)
+        # This loop is not type-stable because `F` and `S` change, but it should
+        # be fine because no one should be using this in performance-critical
```
blegat, line 40: Let's use a function barrier anyway ? What's the advantage of keeping this type unstable ?

blegat: The user could for instance call this in a loop with many points and the performance of this function would be important then.

src/feasibility_checker.jl
```diff
@@ -0,0 +1,161 @@
...
+    end
+    return
+end
+
+function _distance_to_set(
+    x::Vector{T},
+    set::MOI.Nonnegatives,
+) where {T<:Real}
+    _check_dimension(x, set)
+    return LinearAlgebra.norm(max(-xi, zero(T)) for xi in x)
```
blegat, line 134: If we go with an non-conic feasibility checker, it would make work sense to use infinity norm here to be consistent with the scalarized constraints. Otherwise, you would get different results whether you use functionize/scalarize bridges or not.

src/feasibility_checker.jl
```diff
@@ -0,0 +1,161 @@
...
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+using LinearAlgebra
+
+"""
+    primal_feasibility_report(
+        model::Model,
+        point::AbstractDict{VariableRef,Float64};
```
blegat, line 11: I would have a default value for `point` equal to the primal value. This is want people will need 99% of the time and it will be annoying to construct this dict (people could even construct it incorrectly and get errors).

src/feasibility_checker.jl
```diff
@@ -0,0 +1,161 @@
...
+    )::Union{Nothing,Dict{Any,Float64}}
+
+Check the primal feasibility of `model` at the point given by the dictionary
+`point` mapping variables to primal values. If a variable is not given in
+`point`, the value is assumed to be `0.0`.
+
+A point is classed as feasible if the distance between the point and the set is
+less than or equal to `atol`.
+
+If the point is feasible, this function returns `nothing`. If infeasible, this
```
joehuchette, line 22: Why not just return an empty dictionary if the point is feasible?

src/feasibility_checker.jl
odow: > you would get different results whether you use functionize/scalarize bridges or not.

Yes. But there are going to be weird artifacts due to the tolerance no matter how we pick the distance. I'd rather have a single definition "The Euclidean distance between the point and the set," rather than defining different distances for different sets.

src/feasibility_checker.jl
```diff
@@ -0,0 +1,208 @@
...
+
+"""
+    primal_feasibility_report(
+        model::Model,
+        [point::AbstractDict{VariableRef,Float64}];
+        atol::Float64 = 0.0,
+    )::Dict{Any,Float64}
+
+Given a dictionary `point`, which maps variables to primal values, return a
+dictionary mapping the constraint reference of each constraint in `model` to the
```
joehuchette, line 16: Think you want to mention that only _infeasible_ constraints will be added to the dictionary.

test/feasibility_checker.jl
```diff
@@ -0,0 +1,234 @@
...
+
+function runtests()
+    for name in names(@__MODULE__; all = true)
+        if !startswith("$(name)", "test_")
+            continue
+        end
+        @testset "$(name)" begin
+            getfield(@__MODULE__, name)()
+        end
+    end
```
matbesancon, line 229:
```suggestion
        if startswith("$(name)", "test_")
            @testset "$(name)" begin
                getfield(@__MODULE__, name)()
            end
        end
    end
```
the other block is not used

matbesancon:
> This is more about getting the API for JuMP sorted

The only risk I see is getting the API sorted for linear and then struggling to fit non-linear in there.

Also as said in the call this shouldn't be a blocker IMO, anything can be added in 1.+ while things cannot be changed in a breaking manner, which means we would be pushing features then blocked before 2.x. If it's fully there great, but I wouldn't push an interface just for having an interface defined.

I think this could be implemented (partially) at the MOI level?

odow:
I meant that we should lock in
```julia
primal_feasibility_report(
     model::Model,
     [point::AbstractDict{VariableRef,Float64}];
     atol::Float64 = 0.0,
)::Dict{Any,Float64}
```
but not necessarily define a distance for every single set.

test/feasibility_checker.jl
odow: In these situations I prefer early returns: https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

matbesancon: I tend to do too but in that case there is just no "rest body" to early terminate, it's just one statement

src/feasibility_checker.jl
blegat: It could also be the infinite norm for any set, no?

odow: Yes. But it could also be the L2 norm. As a standard definition, L2 makes the most sense for most sets, except where it is difficult to compute (e.g., PSDCone). The distance depends on the formulation, but that shouldn't be surprising. The formulations are different. The tolerances are also absolute, so they're not scale-invariant.

joaquimg: L2 is not that difficult for PSD. There just isn't a closed formula for that. Moreover, L2 is probably the easiest for PSD.

On the main discussion... I think people will end up asking for alternate distances, but that can be done in the sequence.

src/feasibility_checker.jl
```diff
@@ -0,0 +1,205 @@
...
+distance between the point and the nearest feasible point, if the distance is
+greater than `atol`.
+
+## Notes
+
+ * If a variable is not given in `point`, the value is assumed to be `0.0`.
```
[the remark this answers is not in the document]
odow: I guess we could have
```Julia
primal_feasibility_report(model, point; skip_missing::Bool = false)
```
where constraints are skipped if `skip_missing=true`, otherwise the missing values default to `0`?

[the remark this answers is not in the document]
oxinabox: Why is 0 a good default?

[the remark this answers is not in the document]
odow: `0` is a fairly conventional default. Especially for LP solvers that find basic solutions.

How about an argument: `default::Union{Float64,Missing} = 0`?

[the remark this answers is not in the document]
odow: Take a look at the latest commit. Sometimes I'm amazed at how Julia JustWorksᵀᴹ.

src/feasibility_checker.jl
```diff
@@ -0,0 +1,214 @@
...
+        _add_infeasible_constraints(
+            model, F, S, violated_constraints, point_f, atol
+        )
+    end
+    if num_nl_constraints(model) > 0
+        if ismissing(default)
+            error(
+                "`default` cannot be `missing` when nonlinear constraints " *
+                "are present.",
+            )
```
oxinabox, line 59: why not?

odow: There are checks like this: https://github.com/jump-dev/JuMP.jl/blob/2792fab24726eca47660ceb332d9d27c89ebf80e/src/nlp.jl#L779
```Julia
julia> x = [missing]
1-element Array{Missing,1}:
 missing

julia> x == x
missing
```

src/feasibility_checker.jl
```diff
@@ -0,0 +1,214 @@
...
+#  file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+using LinearAlgebra
+
+"""
+    primal_feasibility_report(
+        model::Model,
+        [point::AbstractDict{VariableRef,Float64}];
+        atol::Float64 = 0.0,
+        default::Union{Float64,Missing} = 0.0,
```
odow, line 13: Change to `skip_missing::Bool`

src/feasibility_checker.jl
```diff
@@ -0,0 +1,221 @@
...
+    )::Dict{Any,Float64}
+
+Given a dictionary `point`, which maps variables to primal values, return a
+dictionary mapping the constraint reference of each constraint in `model` to the
+distance between the point and the nearest feasible point, if the distance is
+greater than `atol`.
+
+## Notes
+
+ * If `skip_missing = true`, constraints containing variables that are not in
```
blegat, line 33: This docstring does not say what happens of `skip_missing` is `false` and the value of some variables is not given

odow: Added

odow:
Any other reviews?

docs/src/manual/solutions.md
```diff
@@ -489,3 +489,58 @@ for i in 2:result_count(model)
...
 ```
+
+## Checking feasibility of solutions
+
+To check the feasibility of a primal solution, use
+[`primal_feasibility_report`](@ref), which takes a `model`, a dictionary mapping
+each variable to a primal solution value (defaults to the last solved solution),
+and a tolerance `atol` (defaults to `0.0`).
+
+The function returns a dictionary which maps the infeasible constraint
```
mlubin, line 500: .. returns a dictionary whose keys are the constraints with an infeasibility greater than the supplied tolerance `atol`. The value corresponding to each key is the respective infeasibilility. Infeasibility is defined as the distance between the primal value of the constraint (see `MOI.ConstraintPrimal`) and the nearest point by Euclidean distance in the corresponding set.

mlubin (approved):

Issue #1825, opened 2019-02-04
Exported functions with no doc


blegat:
Edit(odow): updated 31-03-2021
 - [ ] `MOI`
 - [ ] `MOIB`
 - [ ] `MOIU`
 - [ ] `VariableNotOwnedError`
 - [ ] `moi_add_constraint`
 - [ ] `moi_add_to_function_constant`
 - [ ] `parse_SD_constraint`
 - [ ] `parse_SD_constraint_expr`
 - [ ] `parse_constraint`
 - [ ] `parse_constraint_expr`
 - [ ] `parse_constraint_head`
 - [ ] `parse_one_operator_constraint`
 - [ ] `parse_ternary_constraint`
 - [ ] `parse_ternary_variable`
 - [ ] `parse_variable`
 - [ ] `variable_type`
 - [ ] `vectorize`
  
Script
```julia
deprecated = [
    :getdual,
    :getobjectivebound,
    :getobjectivevalue,
    :getvalue,
    :linearterms,
    :numnlconstr,
    :numvar,
    :set_coefficient,
    :set_parameter,
    :set_parameters,
    :setlowerbound,
    :setupperbound,
    :setvalue,
    :solve,
    :with_optimizer,
    :writeLP,
    :writeMPS,
]
for name in filter(n -> !(n in deprecated), names(JuMP))
    f = getfield(JuMP, name)
    s = Base.Docs.doc(f)
    if s.content[1].content[1] == "No documentation found."
        println(" - [ ] `", name, "`")
    end
end
```

blegat:
Isn't the 1.0 milestone more appropriate ? This is not a regression from v0.18

mlubin:
Of these, I'd say that only `index` and `optimizer_index` need to be documented for 0.19. The rest are not that essential for typical usage.

odow:
Here's an updated list:

 -[ ] `@NLconstraints`
 -[ ] `@NLexpressions`
 -[ ] `@SDconstraints`
 -[ ] `@expressions`
 -[ ] `AUTOMATIC`
 -[ ] `AbstractJuMPScalar`
 -[ ] `AbstractModel`
 -[ ] `AbstractVectorSet`
 -[ ] `AffExpr`
 -[ ] `DIRECT`
 -[ ] `GenericAffExpr`
 -[ ] `GenericQuadExpr`
 -[ ] `IJuliaMode`
 -[ ] `JuMP`
 -[ ] `LinearTermIterator`
[...]

PR #2468, opened 2021-02-15
Add docstrings to functions in nlp.jl

odow:
Part of #1825

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -1786,6 +1786,9 @@ end
 Register the user-defined function `f` that takes `dimension` arguments in
 `model` as the symbol `s`.
 
+The function `f` must support all subtypes of `Real` as arguments. Do not assume
+that the inputs are `Float64`.
+
 ## Notes
 
  * For this method, you must explicitly set `autodiff = true`, because no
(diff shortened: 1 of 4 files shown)


src/nlp.jl
```diff
@@ -1790,7 +1844,59 @@ function register(
...
+        dimension::Integer,
+        f::Function,
+        ∇f::Function;
+        autodiff:Bool = false,
+    )
+
+Register the user-defined function `f` that takes `dimension` arguments in
+`model` as the symbol `s`. In addition, provide a gradient function `∇f`.
+
+If the function `f` is uni-variate (i.e., `dimension == 1`), `∇f` must return
```
blegat, line 1860: univariate without `-` ?

src/nlp.jl
```diff
@@ -1756,7 +1774,43 @@ function _UserFunctionEvaluator(
...
+        f::Function;
+        autodiff:Bool = false,
+    )
+
+Register the user-defined function `f` that takes `dimension` arguments in
+`model` as the symbol `s`.
+
+## Notes
+
+ * For this method, you must explicitly set `autodiff = true`.
```
blegat, line 1791: Specify that this is because `∇f` is not given. Can't this work without `autodiff = true` if the solver is zero order ?

mlubin: I could in theory, but it doesn't in practice. We also don't have any zero-order solvers connected. There's no real reason to use JuMP with a zero-order solver.

PR #2472, opened 2021-02-16
Add missing docstrings in print.jl

odow:
Part of https://github.com/jump-dev/JuMP.jl/issues/1825

Also related to https://github.com/jump-dev/JuMP.jl/pull/2449, which adds some other missing docstrings.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -23,8 +23,25 @@
 #############################################################################
 
 # Used for dispatching
+"""
+    PrintMode
+
+An abstract type used for dispatching printing.
+"""
 abstract type PrintMode end
+
+"""
+    REPLMode
+
+A type used for dispatching printing. Produces plain-text representations.
+"""
 abstract type REPLMode <: PrintMode end
+
+"""
+    IJuliaMode
+
+A type used for dispatching printing. Produces LaTeX representations.
+"""
 abstract type IJuliaMode <: PrintMode end
 
 # Whether something is zero or not for the purposes of printing it
@@ -634,6 +651,17 @@ function constraint_string(
         return prefix * constraint_str
     end
 end
+
+"""
+    constraint_string(
+        print_mode,
+        ref::ConstraintRef;
+        in_math_mode::Bool = false)
+
+Return a string representation of the constraint `ref`, given the `print_mode`.
+
+`print_mode` should be `IJuliaMode` or `REPLMode`.
+"""
 function constraint_string(print_mode, ref::ConstraintRef; in_math_mode = false)
     return constraint_string(print_mode, name(ref), constraint_object(ref), in_math_mode = in_math_mode)
 end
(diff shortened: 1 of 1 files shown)


src/print.jl
```diff
@@ -634,13 +651,32 @@ function constraint_string(
...
+    constraint_string(
+        print_mode,
+        ref::ConstraintRef;
+        in_math_mode::Bool = false)
+
+Return a string representation of the constraint `ref`, given the `print_mode`.
+
+`print_mode` should be `IJuliaMode` or `REPLMode`.
+"""
 function constraint_string(print_mode, ref::ConstraintRef; in_math_mode = false)
```
odow, line 665: Should these even be public methods? There are `Base.show(io::IO, ref::ConstraintRef)` and `Base.show(io::IO, ::MIME"text/latex", ref::ConstraintRef)`, which we should encourage the use-of instead.

src/print.jl
```diff
@@ -634,13 +651,32 @@ function constraint_string(
...
 #------------------------------------------------------------------------
+
+"""
+    nl_expr_string(model::Model, print_mode, c::_NonlinearExprData)
+
+Return a string representation of the nonlinear expression `c` belonging to
+`model`, given the `print_mode`. `print_mode` should be `IJuliaMode` or
+`REPLMode`.
+"""
 function nl_expr_string(model::Model, print_mode, c::_NonlinearExprData)
```
odow, line 680: Ditto on whether this should be exported.

src/print.jl
```diff
@@ -682,6 +718,13 @@ end
...
 # TODO: Printing is inconsistent between regular constraints and nonlinear
 # constraints because nonlinear constraints don't have names.
+"""
+    nl_constraint_string(model::Model, mode, c::_NonlinearConstraint)
+
+Return a string representation of the nonlinear constraint `c` belonging to
+`model`, given the `print_mode`. `print_mode` should be `IJuliaMode` or
+`REPLMode`.
+"""
 function nl_constraint_string(model::Model, mode, c::_NonlinearConstraint)
```
odow, line 728: Ditto on whether this should be exported.

src/print.jl
```diff
@@ -23,8 +23,25 @@
...
+An abstract type used for dispatching printing.
+"""
 abstract type PrintMode end
+
+"""
+    REPLMode
+
+A type used for dispatching printing. Produces plain-text representations.
+"""
 abstract type REPLMode <: PrintMode end
```
odow, line 38: A few questions on these. Why are the abstract? Is anyone subtyping them? Ideally, for JuMP 1.0 we would replace and use `MIME"text/plain"` and `MIME"text/latex"` instead.

src/print.jl
blegat: Yes, extension might add methods to this function

src/print.jl
blegat: We should probably revisit these questions once https://github.com/jump-dev/MathOptInterface.jl/pull/1180 is done. I'm waitinh

src/print.jl
odow: The could implement `show` directly though, and we could use that.

src/print.jl
odow: The printing PR is ready from my side: https://github.com/jump-dev/JuMP.jl/pull/2449

src/print.jl
blegat: I think the difference is that `is_math_mode` can be used to avoid `$$` to be added twice.

odow:
I'll merge the docstrings for now. We can consider a bigger printing refactor later.

PR #2478, opened 2021-02-17
Rewrite the Model documentation

odow:
Preview: https://jump.dev/JuMP.jl/previews/PR2478/manual/models/


docs/src/manual/models.md
```diff
@@ -0,0 +1,263 @@
...
+end
+DocTestFilters = [r"≤|<=", r"≥|>=", r" == | = ", r" ∈ | in ", r"MathOptInterface|MOI"]
+```
+
+# Models
+
+A JuMP model keeps a [MathOptInterface (MOI)](https://github.com/jump-dev/MathOptInterface.jl)
+*backend* of type `MOI.ModelLike` that stores the optimization
+problem and acts as the optimization solver. We call it an MOI *backend* and not
+optimizer as it can also be a wrapper around an optimization file format such as
```
blegat, line 14: `it can also for instance be`. It can be more than just either an optimizer of a MOI.Formats model, here it seems you say these are the only 2 possibilities.

docs/src/manual/models.md
```diff
@@ -0,0 +1,263 @@
...
+
+While this allows JuMP to be a thin wrapper on top of the solver API, as
+mentioned in the last point above, this seems rather demanding on the solver.
+Indeed, while some solvers support incremental building of the model and
+modifications before and after solve, other solvers only support the model being
+copied at once before solve. Moreover, it seems to require all solvers to
+implement all possible reformulations independently which seems both very
+ambitious and might generate a lot of duplicated code.
+
+These apparent limitations are addressed at level of MOI in a manner
```
blegat, line 39: at THE level

docs/src/manual/models.md
odow: Yeah I haven't touched this part. I'm going to move it to the end and simplify it a lot.

odow:
I moved the complexity of caching optimizers etc to the end, and greatly simplified what was there. In the long-term, the MOI docs are a better place for explanation material, since they can be in more depth than a general JuMP reader just looking to write models without worrying about the details.

blegat (approved):

PR #2481, opened 2021-02-18
Tidy various docstrings

odow:
Preview: https://jump.dev/JuMP.jl/previews/PR2481


src/JuMP.jl
```diff
@@ -298,7 +306,7 @@ end
 """
     mode(model::Model)
 
-Return mode (DIRECT, AUTOMATIC, MANUAL) of model.
+Return the `ModelMode` (`DIRECT`, `AUTOMATIC`, or `MANUAL`) of `model`.
```
blegat, line 309: I see that `ModelMode` has a docstring. Shouldn't it be in the doc so that this can reference it with `[@ref]` ?

PR #2482, opened 2021-02-18
Rewrite the solution documentation

odow:
Preview: https://jump.dev/JuMP.jl/previews/PR2482/manual/solutions/

### DIFF
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -208,13 +208,13 @@ And data, a 2-element Array{Float64,1}:
 
 ## Recommended workflow
 
-The following is a recommended workflow for solving a model and querying the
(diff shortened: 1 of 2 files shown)


docs/src/manual/solutions.md
```diff
@@ -1,127 +1,244 @@
...
+    Dimension 1, Symbol[:a, :b]
+And data, a 2-element Array{Float64,1}:
+  0.5714285714285694
+ -0.0
+```
+
+## Recommended workflow
+
+The following is a recommended workflow for solving a model and querying the
+solution is something like the following:
```
blegat, line 212: There is twice "the following"

blegat (approved):

Issue #2273, opened 2020-07-11
Address invalidations to speed up compile time


mlubin:
https://github.com/JuliaLang/www.julialang.org/pull/794

JuMP is mentioned as a package that's affected by this issue.

odow:
Here's some shiny new tools on this front: https://github.com/JuliaLang/www.julialang.org/pull/1093

odow:
On Julia 1.6-RC1 and current JuMP master:
```Julia
julia> length(invalidations)
2102

julia> trees[end-4]
inserting collect_to_with_first!(dest::Union{JuMP.Containers.DenseAxisArray, JuMP.Containers.SparseAxisArray}, first_value, iterator, state) in JuMP.Containers at /Users/oscar/.julia/packages/JuMP/W6eqk/src/Containers/Containers.jl:37 invalidated:
   backedges: 1: superseding collect_to_with_first!(dest, v1, itr, st) in Base at array.jl:705 with MethodInstance for collect_to_with_first!(::Any, ::Any, ::Base.Generator{_A, Test.var"#24#25"{Int64}} where _A, ::Any) (24 children)

julia> trees[end-7]
inserting println(io::IO, card::MathOptInterface.FileFormats.MPS.Card) in MathOptInterface.FileFormats.MPS at /Users/oscar/.julia/packages/MathOptInterface/ZJFKw/src/FileFormats/MPS/MPS.jl:115 invalidated:
   backedges: 1: superseding println(xs...) in Base at coreio.jl:4 with MethodInstance for println(::Any, ::Any) (1 children)
              2: superseding println(io::IO, xs...) in Base at strings/io.jl:73 with MethodInstance for println(::Base.TTY, ::Any) (2 children)
              3: superseding println(io::IO, xs...) in Base at strings/io.jl:73 with MethodInstance for println(::IO, ::Any) (7 children)

julia> trees[end-9]
inserting instantiate(bc::Base.Broadcast.Broadcasted{var"#s141", Axes, F, Args} where {var"#s141"<:JuMP.Containers.BroadcastStyle, Axes, F, Args<:Tuple}) in JuMP.Containers at /Users/oscar/.julia/packages/JuMP/W6eqk/src/Containers/SparseAxisArray.jl:194 invalidated:
[...]

Issue #1181, opened 2018-02-05
Precompile JuMP


mlubin:
When the API settles down, we should add `Base.precompile` statements to improve loading time.

blegat:
Isn't it already done ?

https://github.com/JuliaOpt/JuMP.jl/blob/0b849149bbc1866c652051311da8990969813efb/src/JuMP.jl#L11

mlubin:
I'm referring to https://github.com/JuliaOpt/JuMP.jl/blob/release-0.18/src/precompile.jl.

odow:
Started by #2484. We should continue to investigate this and track down causes of compilation latency, particularly after the next MOI release.

odow:
Closing in favor of #2735

PR #2484, opened 2021-02-19
Add precompilation directives

odow:
This PR is an attempt to get precompilation working throughout the JuMP ecosystem. There are a bunch of related PRs in other packages.
 * GLPK https://github.com/jump-dev/GLPK.jl/pull/173
 * Ipopt https://github.com/jump-dev/Ipopt.jl/pull/260
 * MathOptInterface https://github.com/jump-dev/MathOptInterface.jl/pull/1243
 * SCS https://github.com/jump-dev/SCS.jl/pull/204

Related issues: #2273, #1181

## Headline takeaways

* "Time-to-first-solve" is halved. The diet problem with GLPK:
```
# Before:
# 13.303004 seconds (43.19 M allocations: 2.462 GiB, 8.20% gc time, 22.66% compilation time)
# After:
# 7.823846 seconds (20.45 M allocations: 1.125 GiB, 4.87% gc time, 21.25% compilation time)
```
* There are a looooot of things that don't infer. (Details to follow.)
* Some precompile statements are invalid. 

Trying to isolate reasons, I find a lot of "Julia's specialization heuristics may be responsible," so it seems like even though we think Julia is specializing on the bridges, it actually isn't. (https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing)
```julia
using SnoopCompile
tinf = @snoopi_deep Foo.stress_precompile()
staleinstances(tinf)

itrigs = inference_triggers(tinf)
mtrigs = accumulate_by_source(Method, itrigs)
using AbstractTrees
itree = trigger_tree(mtrigs[end].itrigs)
suggest(itree.children[1])

julia> suggest(itree.children[1])
/Users/oscar/.julia/dev/MathOptInterface/src/Bridges/lazy_bridge_optimizer.jl:218: non-inferrable or unspecialized call, perhaps annotate node(b::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{SCS.Optimizer, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, F::Type{MathOptInterface.ScalarAffineFunction{Float64}}, S::Type{MathOptInterface.GreaterThan{Float64}}) at lazy_bridge_optimizer.jl:218 with type MethodInstance for supports_constraint(::Type{MathOptInterface.Bridges.Constraint.GreaterToIntervalBridge{Float64, F} where F<:MathOptInterface.AbstractScalarFunction}, ::Type{MathOptInterface.ScalarAffineFunction{Float64}}, ::Type{MathOptInterface.GreaterThan{Float64}})
If a noninferrable argument is a type or function, Julia's specialization heuristics may be responsible.
````

## Code

```Julia
module Foo

using JuMP
using Ipopt
using SCS
using GLPK

function stress_precompile()
    input_dir =
        joinpath(dirname(dirname(pathof(JuMP))), "docs", "src", "examples")
    for (root, dirs, files) in walkdir(input_dir)
        for file in files
            if !endswith(file, ".jl")
                continue
            end
            include(joinpath(root, file))
        end
    end
    return
end

function example_diet()
    categories = ["calories", "protein", "fat", "sodium"]
    category_data = Containers.DenseAxisArray([
        1800 2200;
        91   Inf;
        0    65;
        0    1779
        ], categories, ["min", "max"]
    )
    foods = [
        "hamburger", "chicken", "hot dog", "fries", "macaroni", "pizza",
        "salad", "milk", "ice cream",
    ]
    cost = Containers.DenseAxisArray(
        [2.49, 2.89, 1.50, 1.89, 2.09, 1.99, 2.49, 0.89, 1.59],
        foods
    )
    food_data = Containers.DenseAxisArray(
        [
            410 24 26 730;
            420 32 10 1190;
            560 20 32 1800;
            380  4 19 270;
            320 12 10 930;
            320 15 12 820;
            320 31 12 1230;
            100  8 2.5 125;
            330  8 10 180
        ], foods, categories
    )
    model = Model(GLPK.Optimizer)
    @variables(model, begin
        category_data[c, "min"] <= nutrition[c = categories] <= category_data[c, "max"]
        buy[foods] >= 0
    end)
    @objective(model, Min, sum(cost[f] * buy[f] for f in foods))
    @constraint(model, [c in categories],
        sum(food_data[f, c] * buy[f] for f in foods) == nutrition[c]
    )
    optimize!(model)
    term_status = termination_status(model)
    @assert term_status == MOI.OPTIMAL
    @constraint(model, buy["milk"] + buy["ice cream"] <= 6)
    optimize!(model)
    @assert termination_status(model) == MOI.INFEASIBLE
    @assert primal_status(model) == MOI.NO_SOLUTION
    return
end

end

# ------------------------------------------------------------------------------
# Uncomment this block to generate the precompile directives
#
# using SnoopCompile
# tinf = @snoopi_deep Foo.stress_precompile()
# ttot, pcs = SnoopCompile.parcel(tinf)
# SnoopCompile.write("precompiles", pcs)
# for file in readdir("precompiles")
#     if !endswith(file, ".jl")
#         continue
#     end
#     src = joinpath("precompiles", file)
#     m = match(r"precompile\_(.+)\.jl", file)
#     modules = split(m[1], ".")
#     modules = vcat(modules[1], "src", modules[2:end])
#     if !(modules[1] in ["GLPK", "Ipopt", "JuMP", "MathOptInterface", "SCS"])
#         continue
#     end
#     dest = joinpath("/Users/Oscar/.julia/dev", modules..., "precompile.jl")
#     @show dest
#     cp(src, dest; force = true)
# end
# ------------------------------------------------------------------------------

# ------------------------------------------------------------------------------
# A small example to see the effect of precompiling
#
# Before:
# 13.303004 seconds (43.19 M allocations: 2.462 GiB, 8.20% gc time, 22.66% compilation time)
# After:
# 7.823846 seconds (20.45 M allocations: 1.125 GiB, 4.87% gc time, 21.25% compilation time)
#
# @time Foo.example_diet()
# ------------------------------------------------------------------------------

# ------------------------------------------------------------------------------
# Check effect of precompiling on all examples
#
# Before:
# 83.043400 seconds (240.70 M allocations: 13.937 GiB, 6.46% gc time, 8.99% compilation time)
# After:
# 74.534196 seconds (190.81 M allocations: 11.039 GiB, 5.32% gc time, 7.50% compilation time)
#
# @time Foo.stress_precompile()
# ------------------------------------------------------------------------------
```

## SnoopCompile

https://timholy.github.io/SnoopCompile.jl/stable/snoopi_deep_parcel/


src/Containers/precompile.jl
```diff
@@ -0,0 +1,93 @@
...
+        if !ismissing(fbody)
+            precompile(fbody, (Function,typeof(nested),Function,Vararg{Function, N} where N,))
+        end
+    end   # time: 0.002819959
+    Base.precompile(Tuple{typeof(haskey),NoDuplicateDict{_A, _B} where {_A, _B},Tuple{Int64, Int64, Int64}})   # time: 0.002722087
+    Base.precompile(Tuple{typeof(vectorized_product),Base.OneTo{Int64},Vararg{Base.OneTo{Int64}, N} where N})   # time: 0.002584661
+    Base.precompile(Tuple{typeof(_to_index_tuple),Tuple{Int64, Function},Tuple{Dict{Int64, Int64}, Dict{Symbol, Int64}}})   # time: 0.002516737
+    Base.precompile(Tuple{typeof(_to_index_tuple),Tuple{String, Int64, Int64},Tuple{Dict{String, Int64}, Dict{Int64, Int64}, Dict{Int64, Int64}}})   # time: 0.002253078
+    Base.precompile(Tuple{typeof(has_dependent_sets),Vector{Any},Vector{Any}})   # time: 0.00217896
+    # TODO: Base.precompile(Tuple{Type{DenseAxisArray},Core.Array{T, N},Any,Tuple{Vararg{Dict, N}} where N})   # time: 0.001977609
```
odow, line 84: @timholy I found that `SnoopCompile.parcel(tinf)` created a lot of these invalid precompile statements. Is this a bug? Expected behavior? 

Here's an example of another one: https://github.com/jump-dev/MathOptInterface.jl/pull/1243#discussion_r578863220

(See the top of the PR for how I created these files.)

mlubin:
Nice! What about using the [bot](https://discourse.julialang.org/t/ann-snoopcompile-1-2-package-precompilation/33359) to keep this list up to date?

odow:
> What about using the bot to keep this list up to date?

:+1: but there are a few issues to sort first, like the invalid statements.

blegat:
It's a bit surprising that
```
# Before:
# 13.303004 seconds ... 22.66% compilation time
# After:
# 7.823846 seconds ... 21.25% compilation time
```
So we get 2x speedup by reducing compilation time that was only 22% :thinking: 

> it seems like even though we think Julia is specializing on the bridges, it actually isn't

Scary but interesting :)

src/Containers/precompile.jl
timholy: No, that's a bug. I haven't yet tried to reproduce this but this looks sufficiently detailed that I expect it to be straightforward. Thanks!!

timholy:
The bot doesn't use `@snoopi_deep` yet, I think. I am probably an outlier, but I tend to use the `parcel` to show me what benefits from precompilation but then rewrite the precompile file by hand. Example: https://github.com/JuliaImages/ImageCore.jl/blob/master/src/precompile.jl. I find that this avoids having to regenerate the precompile files almost daily (check the history of commits to Plots.jl if you want to see what I mean).

PR #2486, opened 2021-02-19
Add better error message when index is same as model

### LINKED ISSUE #858: Scope of index variables
```julia
m = Model()
@variable(m,x[1:5])
@constraint(m,[m=1:4],x[m]==x[m+1])
```
complains that there's no addconstraint method with an Int64 as the first argument. Should this work?
```julia
@constraint(m,[i=1:4],x[i]==x[i+1])
```
is fine.

odow:
Closes #858

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1726,7 +1726,7 @@ macro NLparameter(m, ex, extra...)
     idxvars, indices = Containers._build_ref_sets(_error, c)
     if m in idxvars
         _error(
-            "Index $(m) is the same symbol as the model. Use a different" *
+            "Index $(m) is the same symbol as the model. Use a different " *
             "name for the index.",
         )
     end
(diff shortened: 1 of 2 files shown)


test/macros.jl
```diff
@@ -752,6 +752,35 @@ end
         @variable(model, y; integer = true)
         @test is_integer(y)
     end
+
+    @testset "Model as index" begin
+        m = Model()
+        @variable(m, x)
+        @test_macro_throws(
+            ErrorException("In `@variable(m, y[m = 1:2] <= m)`: Index is the same name as the model."),
```
blegat, line 760: In Julia v1.0, it prints as `m=1:2` without spaces

src/macros.jl
```diff
@@ -434,7 +434,9 @@ function _constraint_macro(
     # Strategy: build up the code for add_constraint, and if needed we will wrap
     # in a function returning `ConstraintRef`s and give it to `Containers.container`.
     idxvars, indices = Containers._build_ref_sets(_error, c)
-
+    if args[1] in idxvars
+        _error("Index is the same name as the model.")
```
blegat, line 438: I would be more explicit, e.g. One index has the same name `...` than the JuMP model

PR #2494, opened 2021-02-21
Fix doctests

### LINKED ISSUE #2495: Documentation failing on master with MOI 0.9.20
@blegat see https://github.com/jump-dev/JuMP.jl/runs/1944623462#step:5:798

odow:
Seems like the recent change in MOI got more specific in what changes it supported: https://github.com/jump-dev/JuMP.jl/runs/1944623462#step:5:798

This is arguably incorrect on the JuMP side, rather than MOI.

Edit: I was responsible for this: https://github.com/jump-dev/MathOptInterface.jl/pull/1217

Closes #2495


blegat (approved):

odow:
Merged for now. We may have to make a new JuMP release if someone complains.

PR #2496, opened 2021-02-22
Adds yet another modelling tutorial from JuMPTutorials

ferrolho:
@odow, this is the _Space Shuttle Reentry Trajectory_ tutorial from https://github.com/jump-dev/JuMPTutorials.jl/pull/97


docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,312 @@
...
+# Finally, in order to approximate the derivatives of the problem dynamics, we will use either rectangular
+# or [trapezoidal](http://en.wikipedia.org/wiki/Trapezoidal_rule) integration.
+
+# > #### Disclaimer
+# > Do not try to actually land a Space Shuttle using this notebook! 😛
+# > There's no mesh refinement going on, which can lead to unrealistic trajectories having position
+# > and velocity errors with orders of magnitude $10^4$ ft and $10^2$ ft/sec, respectively.
+
+using Interpolations
+using JuMP, Ipopt #, KNITRO
```
odow, line 120: Please structure the imports as
```Julia
using JuMP
import Interpolations
import Ipopt
```
and preface the Interpolation calls with `Interpolations.`. It makes it clearer for readers where the interpolation code is coming from.

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,312 @@
...
+interp_linear = LinearInterpolation([1, n], [xₛ, xₜ])
+initial_guess = mapreduce(transpose, vcat, interp_linear.(1:n))
+set_start_value.(all_variables(model), vec(initial_guess))
+
+## Functions to restore `h` and `v` to their true scale
+@NLexpression(model, h[j = 1:n], scaled_h[j] * 1e5)
+@NLexpression(model, v[j = 1:n], scaled_v[j] * 1e4)
+
+## Helper functions
+@NLexpression(model, c_L[j = 1:n], a₀ + a₁ * rad2deg(α[j]))
```
odow, line 223: Fixed!

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,312 @@
...
+# > #### Disclaimer
+# > Do not try to actually land a Space Shuttle using this notebook! 😛
+# > There's no mesh refinement going on, which can lead to unrealistic trajectories having position
+# > and velocity errors with orders of magnitude $10^4$ ft and $10^2$ ft/sec, respectively.
+
+using Interpolations
+using JuMP, Ipopt #, KNITRO
+
+## Global variables
+w  = 203000.0  # weight (lb)
```
odow, line 123: Do you have the stack trace of the error from setting these to `const`? We should fix the root cause.

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,312 @@
...
+integration_rule = "rectangular"
+nothing #hide
+
+#-
+
+user_options_ipopt = (
+    "mu_strategy" => "monotone",
+    "linear_solver" => "ma27",  # For the best results, it is advised to experiment different linear solvers.
+                                ## If Ipopt is not compiled with MA27/MA57, it may fallback to 'MUMPS'.
+                                ## In general, the linear solver MA27 is much faster than MUMPS.
```
odow, line 172: Does Ipopt not throw an error if MA27 is not available? This comment may be better structured as a `!!! tip`.
```
!!! tip
    For best results, it is advised ...
```

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,312 @@
...
+)
+
+## user_options_knitro = (
+##     "algorithm" => KNITRO.KN_ALG_BAR_DIRECT,
+##     "bar_murule" => KNITRO.KN_BAR_MURULE_QUALITY,
+##     "linsolver" => KNITRO.KN_LINSOLVER_MA27,
+## )
```
odow, line 179: Please remove the KNITRO comments (here and throughout).

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,312 @@
...
+@variables(model, begin
+               0 ≤ scaled_h[1:n]                # altitude (ft) / 1e5
+                          ϕ[1:n]                # longitude (rad)
+    deg2rad(-89) ≤        θ[1:n] ≤ deg2rad(89)  # latitude (rad)
+            1e-4 ≤ scaled_v[1:n]                # velocity (ft/sec) / 1e4
+    deg2rad(-89) ≤        γ[1:n] ≤ deg2rad(89)  # flight path angle (rad)
+                          ψ[1:n]                # azimuth (rad)
+    deg2rad(-90) ≤        α[1:n] ≤ deg2rad(90)  # angle of attack (rad)
+    deg2rad(-89) ≤        β[1:n] ≤ deg2rad( 1)  # bank angle (rad)
+    ##        0.5 ≤       Δt[1:n] ≤ 1.5          # time step (sec)
```
odow, line 194: Unused?

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,312 @@
...
+
+## Solve for the control and state
+status = optimize!(model)
+
+## Show final crossrange of the solution
+println("Final latitude θ = ", round(objective_value(model) |> rad2deg, digits = 2), "°")
+
+# ### Plot the results
+
+## using Plots
```
odow, line 275: I'm enabling plotting here: https://github.com/jump-dev/JuMP.jl/pull/2497

So we might wait for that to merge, and then you can enable the plots.

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,312 @@
...
+# # Space Shuttle Reentry Trajectory
+
+# **Originally Contributed by**: Henrique Ferrolho
+
+# This tutorial demonstrates how to compute a reentry trajectory for the
+# [Space Shuttle](https://en.wikipedia.org/wiki/Space_Shuttle),
+# by formulating and solving a nonlinear programming problem.
+# The problem was drawn from Chapter 6 of 
+# ["Practical Methods for Optimal Control and Estimation Using Nonlinear Programming"](https://epubs.siam.org/doi/book/10.1137/1.9780898718577),
+# by John T. Betts.
```
odow, line 32: You could add something like
```
!!! tip
    This tutorial is a more-complicated version of the [Rocket Control](@ref) example.
    If you're new to solving nonlinear programs in JuMP, you may want to start there 
    instead.
```

odow:
Thanks for this. Good to see the `deg2rad` thing is fixed.

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
ferrolho: Can you have a look at this https://github.com/jump-dev/JuMPTutorials.jl/pull/97#discussion_r572812250? Let me know if you think I should remove the commented line, or if I should add more explanation to it.

odow: Perhaps add an `!!! info` block after the variable definitions with the explanation?

In general, we should remove commented code, because new users are just going to say "why is this here?"

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,331 @@
...
+# \begin{aligned}
+#      h & \quad \text{altitude (ft)},     \qquad & & \gamma \quad \text{flight path angle (rad)}, \\
+#   \phi & \quad \text{longitude (rad)},   \qquad & & \psi   \quad \text{azimuth (rad)},           \\
+# \theta & \quad \text{latitude (rad)},    \qquad & & \alpha \quad \text{angle of attack (rad)},   \\
+#      v & \quad \text{velocity (ft/sec)}, \qquad & & \beta  \quad \text{bank angle (rad)}.
+# \end{aligned}
+# ```
+
+# The aerodynamic and atmospheric forces on the vehicle are
+# specified by the following quantities (English units):
```
odow, line 65: Is "English unit" a standard term? I would call `ft` and `lb` etc "Imperial," but that may just be a Kiwi thing...

ferrolho: In Scotland we also call them imperial units. It just happens to be how the problem is originally stated in the book; see previous discussion about this https://github.com/jump-dev/JuMPTutorials.jl/pull/97#discussion_r572706571.

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,331 @@
...
+const b₀ =  0.07854
+const b₁ = -0.61592e-2
+const b₂ =  0.621408e-3
+const c₀ =  1.0672181
+const c₁ = -0.19213774e-1
+const c₂ =  0.21286289e-3
+const c₃ = -0.10117249e-5
+
+## Initial conditions
+const hₛ = 2.6          # altitude (ft) / 1e5
```
odow, line 149: The subscripts aren't displaying correctly 

image

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,331 @@
...
+    else
+        @error "Unexpected integration rule '$(integration_rule)'"
+    end
+end
+
+## Objective: Maximize crossrange
+@objective(model, Max, θ[n])
+
+## Solve for the control and state
+status = optimize!(model)
```
odow, line 288: Remove `status = `.

You could add after this `@assert termination_status(model) == MOI.LOCALLY_SOLVED`

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
```diff
@@ -0,0 +1,331 @@
...
+plt_bank_angle = plot(ts[1:end-1], rad2deg.(value.(β)[1:end-1]), legend=nothing, title="Bank Angle (deg)")
+plt_heating = plot(ts, q.(value.(scaled_h)*1e5, value.(scaled_v)*1e4, value.(α)), legend=nothing, title="Heating (BTU/ft/ft/sec)") 
+
+plt = plot(plt_attack_angle, plt_bank_angle, plt_heating, layout=grid(3, 1), linewidth=2, size=(700, 700))
+
+#-
+
+plt = plot(rad2deg.(value.(ϕ)), rad2deg.(value.(θ)), value.(scaled_h),
+           linewidth=2, legend=nothing, title="Space Shuttle Reentry Trajectory",
+           xlabel="Longitude (deg)", ylabel="Latitude (deg)", zlabel="Altitude (100,000 ft)")
```
odow, line 331: Nice
image

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
odow, line 288: Consider calling `set_silent(model)` before `optimize!`. It prints out quite a lot of output.

odow (commented):
One question: the solve time is ~30 seconds, which is not insignificant in comparison to the total doc build time. Is it possible to solve a coarser time discretization to shrink the solve time? (e.g., 2 seconds) Or will the trajectories diverge too much.

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
ferrolho: Oh, that is weird. In which app are they not displaying correctly? On my system they look fine on the notebook, code, and docs in Chrome.

ferrolho:
> Is it possible to solve a coarser time discretization to shrink the solve time?

Yes, it is possible. The trajectories shouldn't be affected as long as the integration time step and problem discretisation are changed accordingly.

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
ferrolho: image

odow: On Mac, in both chrome and safari: 
image
I guess I don't have a font with `_s` installed.

Can we not just write out `h_s`? I assume I won't be the only one with this issue, so it is only going to cause confusion.

Looks good otherwise though!

ferrolho: It seems you don't have the `\_s` or the `\_t`; I will replace those subscripts with plain `_s` and `_t`. It's weird because you do have others such as `\_r` and `\_e`. Did you notice any others not rendering?

odow (approved):
Thanks!

docs/src/tutorials/Nonlinear programs/space_shuttle_reentry_trajectory.jl
odow: Unicode sub/superscripts are weird. Here is what displays for me: `ₐₑₕᵢⱼₖₗₘₙₒₚᵣₛₜᵤᵥₓ`.

ferrolho:
Thank you for all the feedback!

Issue #2483, opened 2021-02-18
Migrate JuMPTutorials


odow:
This issue is to track the progress of migrating the [modeling tutorials from JuMPTutorials](https://github.com/jump-dev/JuMPTutorials.jl/tree/master/script/modelling).

* Tutorials should be added to some suitable subfolder of https://github.com/jump-dev/JuMP.jl/tree/master/docs/src/tutorials (make one if it doesn't fit. e.g., The examples have `MILP`, `Nonlinear`, `Quadratic`, and `SDP`). We could (and probably should) convert the existing examples into tutorials overtime.
* In addition, they need to be converted from Weave to Literate.
* Some tutorials require extra packages. We should try to minimize the number of new dependencies. <s>For example, we don't need Plots and Gadfly. If `Plots` adds too much to the CI runtime, we could consider using `GRUtils` instead.</s> Plots has been added.

## Tutorials

- [x] #2489 experiment_design
- [x] facility_location
    - Needs `Plots`
- [x] #2489 finance
- [x] #2489 geographic_clustering
    - Needs `Distances`
- [ ] geometric_problems
    - Needs `Gadfly` (Convert to `Plots` instead?)
- [ ] maxcut
    - Needs `LightGraphs`, `Graphplot`, `Colors`
    - These are quit a few dependencies. Do we need this tutorial?
- [x] #2489 n-queens
- [x] #2489 network_flows
- [x] #2489 power_systems
    - Original license:  Creative Commons Attribution-ShareAlike 4.0 International License.
    - https://github.com/JuliaOpt/juliaopt-notebooks/blob/master/notebooks/Dvorkin%20-%20Power%20systems%20-%20Economic%20dispatch%20and%20Unit%20commitment.ipynb
- [ ] problems_on_graphs
     - Needs `LightGraphs`, `Graphplot`, `Colors`
     - These are quit a few dependencies. Do we need this tutorial?
- [x] #2489 rocket_control
    - Original license:  Creative Commons Attribution-ShareAlike 4.0 International License.
    - https://github.com/JuliaOpt/juliaopt-notebooks/blob/master/notebooks/JuMP-Rocket.ipynb
- [x] #2489 sudoku
    - Delete existing [example](https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/examples/Mixed-integer%20linear%20programs/sudoku.jl).

## Unmerged PRs

[...]

PR #2497, opened 2021-02-22
Add plotting to tutorials

odow:
Part of #2483

We should check how long CI takes before merging.


odow:
ooof. CI time went from 12 minutes to 18.

It does print the plot: https://jump.dev/JuMP.jl/previews/PR2497/tutorials/Nonlinear%20programs/rocket_control/#Display-results
image
but it looks like I need to work on the margins.

ferrolho:
Hopefully, that time should come down with the upgrade to Julia 1.6.

odow:
Yes. Once 1.6 is released, I think it is reasonable to move the doc build from 1.0 to 1.6. That will allow us to use the jll builds for GLPK, Ipopt, and SCS as well, which should reduce the build/install times.

This is also only a price we pay once, so adding better plots to the rest of the tutorials should be cheap.

odow:
There you go. Please rebase your tutorial, or wait for #2498 if you want to re-add the `const` terms.

ferrolho:
I'll wait for #2498 to be merged. Thanks for this!

odow:
It's merged.

PR #2499, opened 2021-02-22
Add logistic regression tutorial from JuMPTutorials

frapac:
This PR adds the logistic regression tutorial from 
https://github.com/jump-dev/JuMPTutorials.jl/pull/71

The logistic problem is formulated using `MOI.ExponentialCone` (all mathematical details are given in the tutorial). However, I didn't know where to place the tutorial, as the problem is neither a non-linear program nor a semi-definite program. I have created a new subdirectory `Conic programs`, but feel free to tell me if that's not relevant.

### DIFF
--- a/docs/src/tutorials/Conic programs/logistic_regression.jl
+++ b/docs/src/tutorials/Conic programs/logistic_regression.jl
@@ -1,3 +1,25 @@
+# MIT License                                                                    #src
+#                                                                                #src
+# Copyright (c) 2020 François Pacaud                                             #src
+#                                                                                #src
+# Permission is hereby granted, free of charge, to any person obtaining a copy   #src
+# of this software and associated documentation files (the "Software"), to deal  #src
+# in the Software without restriction, including without limitation the rights   #src
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell      #src
+# copies of the Software, and to permit persons to whom the Software is          #src
+# furnished to do so, subject to the following conditions:                       #src
+#                                                                                #src
+# The above copyright notice and this permission notice shall be included in all #src
+# copies or substantial portions of the Software.                                #src
+#                                                                                #src
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR     #src
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,       #src
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE    #src
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER         #src
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,  #src
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  #src
+# SOFTWARE.                                                                      #src
+
 # # Logistic Regression
 
 # **Originally Contributed by**: François Pacaud
@@ -11,7 +33,7 @@
 # the resulting problem is convex and differentiable. On a modern optimization
 # glance, it is even conic representable.
 #
-# # Formulating the logistic regression problem
+# ## Formulating the logistic regression problem
 #
 # Suppose we have a set of training data-point $i = 1, \cdots, n$, where
 # for each $i$ we have a vector of features $x_i \in \mathbb{R}^p$ and a
@@ -104,8 +126,8 @@
 #
 # It is now time to pass to the implementation. We choose SCS as a conic solver.
 using JuMP
-using Random
-using SCS
+import Random
+import SCS
 
 Random.seed!(2713);
 
@@ -131,7 +153,7 @@ function softplus(model, t, u)
     @constraint(model, vec([-t, 1, z[2]]) in MOI.ExponentialCone())
 end
 
-# ## $\ell_2$ regularized logistic regression
+# ### $\ell_2$ regularized logistic regression
 # Then, with the help of the `softplus` function, we could write our
 # optimization model. In the $\ell_2$ regularization case, the constraint
 # $r \geq \|\theta\|_2$ rewrites as a second order cone constraint.
@@ -152,8 +174,11 @@ function build_logit_model(X, y, λ)
     return model
 end
 
-## Generate one dataset.
-## Be careful here, for large n and p SCS could fail to converge!
+# We generate the dataset.
+#
+# !!! Warning
+#     Be careful here, for large n and p SCS could fail to converge!
+#
 n, p = 200, 10
 X, y = generate_dataset(n, p, shift=10.0);
 
@@ -169,7 +194,7 @@ JuMP.optimize!(model)
 # of the dataset, nor by the penalty $\lambda$.
 
 
-# ### Sparse logistic regression
+# ### $\ell_1$ regularized logistic regression
 #
 # We now formulate the logistic problem with a $\ell_1$ regularization term.
 # The $\ell_1$ regularization ensures sparsity in the optimal
@@ -208,7 +233,7 @@ println("Number of non-zero components: ", count_nonzero(θ♯),
         " (out of ", p, " features)")
 
 
-# # Extensions
+# ### Extensions
 # A direct extension would be to consider the sparse logistic regression with
 # *hard* thresholding, which, on contrary to the *soft* version using a $\ell_1$ regularization,
 # adds an explicit cardinality constraint in its formulation:
@@ -231,5 +256,3 @@ println("Number of non-zero components: ", count_nonzero(θ♯),
 # binary variables. Thus the hard sparse regression problem could be solved
 # by any solver supporting mixed integer conic problems.
 #
-# ## References
-# 1. Logistic regression — MOSEK Fusion API. Available at: https://docs.mosek.com/9.2/pythonfusion/case-studies-logistic.html
--- a/docs/src/tutorials/Conic programs/logistic_regression.jl
+++ b/docs/src/tutorials/Conic programs/logistic_regression.jl
@@ -149,8 +149,8 @@ end
 function softplus(model, t, u)
     z = @variable(model, [1:2], lower_bound=0.0)
     @constraint(model, sum(z) <= 1.0)
-    @constraint(model, vec([u - t, 1, z[1]]) in MOI.ExponentialCone())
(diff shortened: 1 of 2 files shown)


docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,235 @@
+# # Logistic Regression
```
odow, line 1: Please add a license header

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,235 @@
...
+# is only a virtual constraint used to clarify the notation).
+# Thus, if $n \gg 1$, we get a large number of variables and constraints.
+
+
+# ## Fitting logistic regression with a conic solver
+#
+# It is now time to pass to the implementation. We choose SCS as a conic solver.
+using JuMP
+using Random
+using SCS
```
odow, line 108: Please use 
```Julia
using JuMP
import Random
import SCS
```
to make it clear which functions come from external packages.

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,235 @@
...
+    ## Add ℓ2 regularization
+    @variable(model, 0.0 <= reg)
+    @constraint(model, vec([reg; θ]) in MOI.SecondOrderCone(p+1))
+    ## Define objective
+    @objective(model, Min, sum(t) + λ * reg)
+    return model
+end
+
+## Generate one dataset.
+## Be careful here, for large n and p SCS could fail to converge!
```
odow, line 156: You can use documenter things like 
```Julia
# !!! warning
#     Be careful here, ...
n, p = 200, 10
```

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,235 @@
...
+# This tutorial shows how to solve a logistic regression problem
+# with JuMP. Logistic regression is a well known method in machine learning,
+# useful when we want to classify binary variables with the help of
+# a given set of features. To this goal,
+# we find the optimal combination of features maximizing
+# the (log)-likelihood onto a training set. From a modern optimization glance,
+# the resulting problem is convex and differentiable. On a modern optimization
+# glance, it is even conic representable.
+#
+# # Formulating the logistic regression problem
```
odow, line 14: Subsections should be heading 2 or more `# ## Formulating ...`

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,235 @@
...
+    end
+    ## Add ℓ2 regularization
+    @variable(model, 0.0 <= reg)
+    @constraint(model, vec([reg; θ]) in MOI.SecondOrderCone(p+1))
+    ## Define objective
+    @objective(model, Min, sum(t) + λ * reg)
+    return model
+end
+
+## Generate one dataset.
```
odow, line 155: Prefer things like this as `# Generate ...` so they get rendered as plain-text, not code comments.

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,235 @@
...
+JuMP.set_optimizer(model, SCS.Optimizer)
+JuMP.optimize!(model)
+
+θ♯ = JuMP.value.(model[:θ]);
+
+# It appears that the speed of convergence is not that impacted by the correlation
+# of the dataset, nor by the penalty $\lambda$.
+
+
+# ### Sparse logistic regression
```
odow, line 172: Perhaps `# ## $\ell_1$ regularized logistic regression` to match the previous heading?

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,235 @@
...
+sparse_model = build_sparse_logit_model(X, y, λ)
+JuMP.set_optimizer(sparse_model, SCS.Optimizer)
+JuMP.optimize!(sparse_model)
+
+θ♯ = JuMP.value.(sparse_model[:θ])
+println("Number of non-zero components: ", count_nonzero(θ♯),
+        " (out of ", p, " features)")
+
+
+# # Extensions
```
odow, line 211: `# ## Extensions` or perhaps `# ### Extensions`

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,235 @@
...
+# ```math
+# \| x\|_0 = \#\{i : \; x_i \neq 0\}
+# ```
+#
+# The cardinality constraint $\|\theta\|_0 \leq k$ could be reformulated with
+# binary variables. Thus the hard sparse regression problem could be solved
+# by any solver supporting mixed integer conic problems.
+#
+# ## References
+# 1. Logistic regression — MOSEK Fusion API. Available at: https://docs.mosek.com/9.2/pythonfusion/case-studies-logistic.html
```
odow, line 235: Is this referenced somewhere? If not. delete.

odow:
`Conic programs` is good.

docs/src/tutorials/Conic programs/logistic_regression.jl
frapac: You are right, I have updated the code accordingly

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,258 @@
...
+    model = Model()
+    @variable(model, θ[1:p])
+    @variable(model, t[1:n])
+    for i in 1:n
+        u = - (X[i, :]' * θ) * y[i]
+        softplus(model, t[i], u)
+    end
+    ## Add ℓ2 regularization
+    @variable(model, 0.0 <= reg)
+    @constraint(model, vec([reg; θ]) in MOI.SecondOrderCone(p+1))
```
odow, line 171: I think you just need `@constraint(model, [reg; θ] in SecondOrderCone())`

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,258 @@
...
+# !!! Warning
+#     Be careful here, for large n and p SCS could fail to converge!
+#
+n, p = 200, 10
+X, y = generate_dataset(n, p, shift=10.0);
+
+## We could now solve the logistic regression problem
+λ = 10.0
+model = build_logit_model(X, y, λ)
+JuMP.set_optimizer(model, SCS.Optimizer)
```
odow, line 188: Please remove `JuMP`.

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,258 @@
...
+    model = Model()
+    @variable(model, θ[1:p])
+    @variable(model, t[1:n])
+    for i in 1:n
+        u = - (X[i, :]' * θ) * y[i]
+        softplus(model, t[i], u)
+    end
+    ## Add ℓ1 regularization
+    @variable(model, 0.0 <= reg)
+    @constraint(model, vec([reg; θ]) in MOI.NormOneCone(p+1))
```
odow, line 216: Ditto on `vec`.

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,258 @@
...
+    X = hcat(X, ones(n_samples, 1))
+    return X, y
+end
+
+# We write a `softplus` function to formulate each constraint
+# $t \geq \log(1 + \exp(u))$ with two exponential cones.
+function softplus(model, t, u)
+    z = @variable(model, [1:2], lower_bound=0.0)
+    @constraint(model, sum(z) <= 1.0)
+    @constraint(model, vec([u - t, 1, z[1]]) in MOI.ExponentialCone())
```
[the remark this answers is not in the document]
frapac: For some reason, the syntax without `vec` was not working before. But I assume it works now.

docs/src/tutorials/Conic programs/logistic_regression.jl
```diff
@@ -0,0 +1,258 @@
...
+
+## We solve the sparse logistic regression problem on the same dataset as before.
+λ = 10.0
+sparse_model = build_sparse_logit_model(X, y, λ)
+JuMP.set_optimizer(sparse_model, SCS.Optimizer)
+JuMP.optimize!(sparse_model)
+
+θ♯ = JuMP.value.(sparse_model[:θ])
+println("Number of non-zero components: ", count_nonzero(θ♯),
+        " (out of ", p, " features)")
```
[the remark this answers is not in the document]
frapac: Indeed ... It did find a sparse solution on the larger problem `n = 2,000`, `m=100`. But SCS takes too long to solve the large instance, so I have decided to reduce the size of the problem to `n=200`, m=10`. Then I have tested for different values of `lambda`, to get a non-trivial solution, but the smaller the lambda, the longer is the convergence of SCS (by a non-negligible amount). 

A solution would be to use ECOS instead, but that would require to add another dependency to this project.

docs/src/tutorials/Conic programs/logistic_regression.jl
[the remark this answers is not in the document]
odow: Admittedly, I didn't check. `u` and `t` are both scalars, correct? This should work without `vec`.

docs/src/tutorials/Conic programs/logistic_regression.jl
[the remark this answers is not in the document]
odow: We should change the data to have more structure then so it can find a sparse solution.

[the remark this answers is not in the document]
frapac: Sounds reasonable. I will investigate that before updating this PR.

docs/src/tutorials/Conic programs/logistic_regression.jl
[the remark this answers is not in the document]
frapac: I confirm that it works without `vec`

docs/src/tutorials/Conic programs/logistic_regression.jl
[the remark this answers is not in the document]
frapac: Investigating it closer, it appears that we are limited by the precision of `SCS`: as it relies on ADMM, we cannot get high accuracy in the solution. I have loosen the tolerance from `1e-8` to `1e-6` in the function `count_nonzero`. With this value, we get a number of non-zero equal to: 
```
Number of non-zero components: 7 (out of 10 features) 
```
without changing the problem under consideration.

odow (approved):

odow:
Thanks @frapac https://jump.dev/JuMP.jl/dev/tutorials/Conic%20programs/logistic_regression/

PR #2503, opened 2021-02-23
DNMY: Switch to building documentation on Julia 1.5

odow:
With all of the tutorials and examples, the documentation now takes quite a while to build. This PR is just to see how long it would take if we switched to building on 1.5.

Switching to Julia 1.6 should bring many more benefits, because compilation happens in parallel, and a lot of work has gone into Plots.jl to reduce latency.


docs/src/manual/constraints.md
```diff
@@ -674,11 +674,10 @@ julia> a = 0
 0
 
 julia> @SDconstraint(model, [x x; x x] >= a)
-ERROR: Operation `-` between `Array{VariableRef,2}` and `Int64` is not allowed. You should use broadcast.
+ERROR: MethodError: promote_operation(::typeof(MutableArithmetics.sub_mul), ::Type{Array{VariableRef,2}}, ::Type{Int64}) is ambiguous. Candidates:
```
odow, line 677: @blegat is this a regression?

odow:
A couple of minutes quicker, but perhaps not enough to justify switching. We should try again once 1.6 is released.

PR #2512, opened 2021-03-01
Add `solution_summary` function

### LINKED ISSUE #2191: Feature Request: Report table for query solutions

AtsushiSakai:
Hi. First of all, thank you for developing great OSS!! JuMP is one of the greatest OSS I've ever used : ).

I am +1 for #2191, I think solution summary report is very helpful for checking optimization results briefly. 
So, I added `show_solution_summary` function to fix #2191 based on the comment( https://github.com/jump-dev/JuMP.jl/issues/2191#issuecomment-738339038). 
I also added tests and docs, and updated manual.

This is the first time for me to contribute JuMP, so any kind of feedback is welcome.


docs/src/manual/solutions.md
```diff
@@ -32,6 +32,29 @@ Subject to
...
+* Status
+  Termination status : OPTIMAL
+  Primal status      : FEASIBLE_POINT
+  Dual status        : FEASIBLE_POINT
+  Message from the solver:
+  "Solution is optimal"
+
+* Candidate solution
+  Objective value : -205.14285714285714
+  Objective bound : Inf
```
odow, line 51: We probably want to align these `:` with the line below.

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,29 @@ Subject to
  y[b] ≤ 1.0
 ```
 
+## Solutions summary  
+[`show_solution_summary`](@ref) can be used for checking the summary of the optimization solutions.
+
+```jldoctest solutions
+julia> show_solution_summary(model)
+* Solver:GLPK
```
odow, line 40: Add a space around `:`

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,29 @@ Subject to
...
+  Message from the solver:
+  "Solution is optimal"
+
+* Candidate solution
+  Objective value : -205.14285714285714
+  Objective bound : Inf
+  Dual objective value : -205.1428571428571
+
+* Work counters
+  Solve time (sec)   : 7.414817810058594e-5
```
odow, line 55: We need to write a doctest filter for this: https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Filtering-Doctests. Something like: `filter = r"[0-9]+.[0-9]+"`

(I'm going to suggest a related change below.)

src/print.jl
```diff
@@ -324,6 +324,94 @@ function objective_function_string(print_mode, model::Model)
...
+If `verbose` is true, additional information are written.
+"""
+function show_solution_summary(io::IO, model::Model; verbose::Bool = false)
+    print(io, "* Solver:", solver_name(model), "\n")
+    print(io, "\n")
+
+    _show_status_summary(io, model, verbose)
+    _show_candidate_solution_summary(io, model, verbose)
+    _show_work_counters_summary(io, model, verbose)
+end
```
odow, line 340: Add an explicit `return` at the end of this function.

src/print.jl
```diff
@@ -324,6 +324,94 @@ function objective_function_string(print_mode, model::Model)
...
 end
 
+"""
+    show_solution_summary([io::IO], model::Model; verbose::Bool = false)
+
+Write to `io` (or to the default output stream `stdout` if `io` is not given) a summary of the solution results.
+If `verbose` is true, additional information are written.
+"""
+function show_solution_summary(io::IO, model::Model; verbose::Bool = false)
+    print(io, "* Solver:", solver_name(model), "\n")
```
odow, line 334: You can use `println` instead of adding `\n`

src/print.jl
```diff
@@ -324,6 +324,94 @@ function objective_function_string(print_mode, model::Model)
...
+
+    _show_status_summary(io, model, verbose)
+    _show_candidate_solution_summary(io, model, verbose)
+    _show_work_counters_summary(io, model, verbose)
+end
+
+show_solution_summary(model::Model; verbose::Bool = false) = show_solution_summary(stdout, model; verbose = verbose)
+
+function _show_status_summary(io::IO, model::Model, verbose::Bool)
+    print(io, "* Status\n")
```
odow, line 345: `println` all through here

src/print.jl
```diff
@@ -324,6 +324,94 @@ function objective_function_string(print_mode, model::Model)
...
+        print(io, "  Dual objective value : ", dual_objective_value(model), "\n")
+    catch ArgumentError
+        # do not print
+    end
+
+    if verbose && has_values(model)
+        print(io, "  Primal solution : \n")
+        for (symbol, variable) in object_dictionary(model) 
+            print(io, "    ", string(symbol), ":", value(variable), "\n")
+        end
```
odow, line 375: Not sure if it should be that, or 
```Julia
for x in all_variables(model)
    println(io, "    ", name(x), " : ", value(x))
end
```

src/print.jl
```diff
@@ -324,6 +324,94 @@ function objective_function_string(print_mode, model::Model)
...
+                print(io, "    ", constraint_name, ":", dual(constraint), "\n")
+            end
+        end
+    end
+    print(io, "\n")
+end
+
+function _show_work_counters_summary(io::IO, model::Model, verbose::Bool)
+    print(io, "* Work counters\n")
+    print(io, "  Solve time (sec)   : ", solve_time(model), "\n")
```
odow, line 397: You may want to format `solve_time` as `@sprintf("%.5f", solve_time(model))`. Very small solve times aren't informative.

src/print.jl
```diff
@@ -324,6 +324,94 @@ function objective_function_string(print_mode, model::Model)
...
+    try
+        print(io, "  Simplex iterations : ", simplex_iterations(model), "\n")
+    catch ArgumentError
+        # do not print
+    end
```
odow, line 402: You could abstract this out into a function, since it comes up frequently
```Julia
function _try_print(io, key, f, model)
    try
        println(io, key, f(model))
    catch
        # Do not print
    end
end

_try_print(io, "  Simplex iterations : ", simplex_iterations, model)
```

Note that `catch ArgumentError` isn't doing what you think it is. It's assigning the error to the variable "ArgumentError". If you wanted to only catch argument errors, you should go
```Julia
try
    ...
catch err
    if !(err isa ArgumentError)
        rethrow(err)
    end
end
```

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,29 @@ Subject to
  y[b] ≤ 1.0
 ```
 
+## Solutions summary  
+[`show_solution_summary`](@ref) can be used for checking the summary of the optimization solutions.
```
odow, line 36: A very minor thing, but please add a blank line between the title and the first line.

src/print.jl
```diff
@@ -324,6 +324,94 @@ function objective_function_string(print_mode, model::Model)
    end
 end
 
+"""
+    show_solution_summary([io::IO], model::Model; verbose::Bool = false)
+
+Write to `io` (or to the default output stream `stdout` if `io` is not given) a summary of the solution results.
```
odow, line 330: Please wrap all lines at 80 characters.

odow (commented):
Thanks! A few comments, but this looks nice.

docs/src/manual/solutions.md
AtsushiSakai: I fixed

src/print.jl
AtsushiSakai: done.

src/print.jl
AtsushiSakai: I think your code is better. I fixed.

docs/src/manual/solutions.md
AtsushiSakai: I fixed.

docs/src/manual/solutions.md
AtsushiSakai: I fixed.

docs/src/manual/solutions.md
AtsushiSakai: I added the filter.

src/print.jl
AtsushiSakai: I used `println` instead of `print`.

src/print.jl
AtsushiSakai: Done.

src/print.jl
AtsushiSakai: Agree. I fixed.

src/print.jl
AtsushiSakai: Thank you. The helper function is useful. I used it. And thank you for letting me know the catch behavior. I was confused as other languages.

src/print.jl
AtsushiSakai: I fixed.

AtsushiSakai:
Thank you for reviewing. I think I addressed all of your comments and CI was fixed. PTAL.

docs/src/manual/solutions.md
```diff
@@ -3,7 +3,7 @@ CurrentModule = JuMP
 DocTestSetup = quote
     using JuMP, GLPK
 end
-DocTestFilters = [r"≤|<=", r"≥|>=", r" == | = ", r" ∈ | in ", r"MathOptInterface|MOI"]
+DocTestFilters = [r"≤|<=", r"≥|>=", r" == | = ", r" ∈ | in ", r"MathOptInterface|MOI", r"[0-9]+.[0-9]+"]
```
odow, line 6: We don't want to add this here, because it will now ignore _every_ number in this file. This could lead to incorrect results slipping past the tests. It should go
````
```jldoctest solutions; filter=r"[0-9]+.[0-9]+"
````

File: docs/src/manual/solutions.md [hunk 13]
```diff
@@ -32,6 +32,30 @@ Subject to
...
+  "Solution is optimal"
+
+* Candidate solution
+  Objective value      : -205.14285714285714
+  Objective bound      : Inf
+  Dual objective value : -205.1428571428571
+
+* Work counters
+  Solve time (sec)   : 0.00008
+```
```
odow, line 57: Perhaps we should also show the `verbose` mode as well?

src/print.jl
```diff
@@ -324,6 +326,84 @@ function objective_function_string(print_mode, model::Model)
...
+"""
+    show_solution_summary([io::IO], model::Model; verbose::Bool = false)
+
+Write to `io` (or to the default output stream `stdout` if `io` is not given)
+a summary of the solution results. 
+If `verbose` is true, additional information are written.
+"""
+function show_solution_summary(io::IO, model::Model; verbose::Bool = false)
+    println(io, "* Solver : ", solver_name(model))
+    println(io, "")
```
odow, line 338: You can go `println(io)`

src/print.jl
```diff
@@ -324,6 +326,84 @@ function objective_function_string(print_mode, model::Model)
...
+        end
+    end
+
+    if verbose && has_duals(model)
+        println(io, "  Dual solution : ")
+        constraint_id = 1
+        for (F, S) in list_of_constraint_types(model)
+            for constraint in all_constraints(model, F, S)
+                constraint_name = name(constraint)
+                if isempty(constraint_name)
```
odow, line 380: I wonder if we should skip constraints (and variables) that have empty names. If they aren't named, they probably aren't important to the user. And printing things with temporary names isn't very helpful because the user won't know what they refer to. What do you think?

src/print.jl
```diff
@@ -324,6 +326,84 @@ function objective_function_string(print_mode, model::Model)
    end
 end
 
+"""
+    show_solution_summary([io::IO], model::Model; verbose::Bool = false)
+
+Write to `io` (or to the default output stream `stdout` if `io` is not given)
+a summary of the solution results. 
+If `verbose` is true, additional information are written.
```
odow, line 334: Perhaps:

```
Write a summary of the solution results to `io` (or to `stdout` if `io` is not 
given).

If `verbose=true`, write out the primal solution for every variable and the
dual solution for every constraint, excluding those with empty names.
```

odow (commented):
Thanks for working on this. I think it will make a great addition.

A couple of very minor things, and one more substantial point about whether we should print values in `verbose` mode if they have empty names.

docs/src/manual/solutions.md
AtsushiSakai: I added.

docs/src/manual/solutions.md
AtsushiSakai: Oh. sorry. I should be. I fixed it.

src/print.jl
AtsushiSakai: Thank you. It is better. I used it.

src/print.jl
AtsushiSakai: I agree. Maybe the temporary name is not useful in usual cases. I fixed it.

src/print.jl
AtsushiSakai: I used your idea.

odow (approved):
Thanks! I added one thing to only print the variable if the name is non-empty. 

Assuming I didn't make a typo, I'll merge this once CI passes.

odow:
Actually, I have one more thing to bikeshed. I know we have `show_constraint_summary`, but should we call it `print_solution_summary`, or even just `solution_summary`?

AtsushiSakai:
I think many Julia codes use `print` for un-decorated representations and `show` for decorated representations. 
I feel the `show_solution_summary`, `show_constraints_summary `, `show_objective_function_summary` shows decorated representations, so `show_*_summary` make sense for me. But `print_*_summary` is also OK.
However, IMO, API name consistency is important. So, either using `show_*_summary` or deprecating current `show_*summary` functions and usinsg`print_*_summary` would be better.

blegat:
Maybe `solution_summary(model)` ? It prints if it's returned to the REPL but you can also print it with `print(solution_summary(model))`

AtsushiSakai:
I think @blegat idea is more Julia style. I think I can define SolutionSummary struct and change show_solution_summary to Base.show(summary::SolutionSummary).
However, if we do this, other summary struct should be defined. What do you think? @odow

test/print.jl
```diff
@@ -793,3 +793,132 @@ end
...
+
+    modelstring = """
+    variables: x, y
+    minobjective: -1.0*x
+    xub: x <= 2.0
+    ylb: y >= 0.0
+    noname: x + y <= 1.0
+    """
+
+    model = MOIU.Model{Float64}()
```
mlubin, line 816: It looks like this `model` variable is unused?

test/print.jl
```diff
@@ -793,3 +793,132 @@ end
     @variable(model, z, base_name = "z^1")
     io_test(IJuliaMode, z, "z\\^1")
 end
+
+@testset "Print solution summary" begin
+
+    m = Model()
```
mlubin, line 799: Nit: By convention we use `model` instead of `m`. It's confusing to have both `m` and `model` (looks like the latter isn't actually used).

mlubin (commented):
+1 to @blegat's struct proposal. Having a `solution_summary` doesn't force us to have other summary structs any more than `show_solution_summary` forces us to have other `show_XXX_summary`.

test/print.jl
AtsushiSakai: Thanks, I removed it.

test/print.jl
AtsushiSakai: I fixed.

AtsushiSakai:
@mlubin Thank you for your comment. I understood. I used struct. PTAL.

odow (commented):
@blegat, I don't really understand the motivation for returning a struct just to overload `show`. `solution_summary` just prints things. 

If we were to return a struct, it should have accessible fields with all the data.

```Julia
struct SolutionSummary
    solver::String
    # Status
    termination_status::MOI.TerminationStatusCode
    primal_status::MOI.ResultStatusCode
    dual_status::MOI.ResultStatusCode
    raw_status::String
    result_count::Int
    has_values::Bool
    has_duals::Bool
    # Candidate solution
    objective_value::Float64
    objective_bound::Float64
    dual_objective_value::Float64
    primal_solution::Dict{String,Float64}
    dual_solution::Dict{String,Float64}
    # Work counters
    solve_time::Float64
    barrier_iterations::Union{Missing,Int}
    simplex_iterations::Union{Missing,Int}
    node_count::Union{Missing,Int}
end
```
Then it makes sense to overload `show` on that.

blegat:
> If we were to return a struct, it should have accessible fields with all the data.

Agreed, no need to compute these only when printing so we might as well store them in a struct.

src/print.jl
```diff
@@ -324,6 +326,175 @@ function objective_function_string(print_mode, model::Model)
...
+    primal_status(model),
+    dual_status(model),
+    raw_status(model),
+    result_count(model),
+    has_values(model),
+    has_duals(model),
+    objective_value(model),
+    _try_get(objective_bound, model),
+    _try_get(dual_objective_value, model),
+    _get_solution_dict(model),
```
blegat, line 390: Maybe only do this in case `verbose` is `true` ?

src/print.jl
```diff
@@ -324,6 +326,175 @@ function objective_function_string(print_mode, model::Model)
...
+    termination_status(model),
+    primal_status(model),
+    dual_status(model),
+    raw_status(model),
+    result_count(model),
+    has_values(model),
+    has_duals(model),
+    objective_value(model),
+    _try_get(objective_bound, model),
+    _try_get(dual_objective_value, model),
```
blegat, line 389: We should have an array here with one entry for `i in 1:result_count` that contains a struct containing the objective value, solution and dual solution, dual objective value, ...

AtsushiSakai: Do you have an example code which has result_count is 2 or more? I would like to add a test using it.

src/print.jl
```diff
@@ -324,6 +326,100 @@ function objective_function_string(print_mode, model::Model)
    end
 end
 
+"""
+    SolutionSummary
+
+A struct to show a summary of the solution results.
+"""
+struct SolutionSummary;model::JuMP.Model;end
```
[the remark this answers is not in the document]
AtsushiSakai: Fixed.

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,57 @@ Subject to
  y[b] ≤ 1.0
 ```
 
+## Solutions summary
+
+[`solution_summary`](@ref) can be used for checking the summary of the optimization solutions.
+
+```jldoctest solutions; filter=r"[0-9]+.[0-9]+"
+julia> show(solution_summary(model))
```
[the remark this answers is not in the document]
AtsushiSakai: I removed it and below.

src/print.jl
```diff
@@ -324,6 +326,100 @@ function objective_function_string(print_mode, model::Model)
...
+"""
+    Base.show([io::IO], summary::SolutionSummary; verbose::Bool = false)
+
+Write a summary of the solution results to `io` (or to `stdout` if `io` is not 
+given).
+
+If `verbose=true`, write out the primal solution for every variable and the
+dual solution for every constraint, excluding those with empty names.
+"""
+function Base.show(io::IO, summary::SolutionSummary; verbose::Bool = false)
```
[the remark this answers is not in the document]
AtsushiSakai: Done

src/print.jl
```diff
@@ -324,6 +326,100 @@ function objective_function_string(print_mode, model::Model)
...
+A struct to show a summary of the solution results.
+"""
+struct SolutionSummary;model::JuMP.Model;end
+
+"""
+    solution_summary(model::Model)
+
+Create a summary of the solution results.
+"""
+solution_summary(model::Model) = SolutionSummary(model)
```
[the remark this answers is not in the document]
AtsushiSakai: Done

src/print.jl
AtsushiSakai: Done

test/print.jl
```diff
@@ -793,3 +793,121 @@ end
...
+
+    @test sprint((io, model) -> show(io, solution_summary(model, verbose=true)), model) == """
+* Solver : Mock
+
+* Status
+  Termination status : OPTIMAL
+  Primal status      : FEASIBLE_POINT
+  Dual status        : FEASIBLE_POINT
+  Result count       : 1
+  Has values         : true
```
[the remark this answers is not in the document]
AtsushiSakai: Removed.

src/print.jl
AtsushiSakai: I found this codes. Maybe I can use it. https://github.com/jump-dev/JuMP.jl/blob/90e63c47372e02b7093a9bf2fb4153dfa548c0a6/test/generate_and_solve.jl#L601-L638

odow: Let's just leave this as-is for now. It covers 99.9% of peoples use-cases, and I'm mindful that @AtsushiSakai has put a lot of effort in, and we keep asking for more and more changes.

In future, we could add a `alternative_solutions::Vector` that is backward compatible. The first result is special as it is the only solution guaranteed to be optimal if `TerminationStatus` is optimal.

odow:
Thanks @AtsushiSakai. It's a very nice contribution.

Issue #2509, opened 2021-02-28
Register non linear parameters in the model


davidanthoff:
There is already a TODO for this in the code [here](https://github.com/jump-dev/JuMP.jl/blob/749c828236dea3a065a428b93df7a60437d80052/src/macros.jl#L1750) :)

It would be awesome if NL parameters were registered in the model so that for example expression printing could use the proper name of the parameter in its output. Also, just seems in general that it should be that way?

odow:
As you've probably started to realize, nonlinear is a second-class citizen in JuMP at present. There are a number of features that are not implemented.

PR #2516, opened 2021-03-03
Register NLparameters and pretty print them

davidanthoff:
Fixes https://github.com/jump-dev/JuMP.jl/issues/2509.

I have to admit I'm not sure this makes sense or whether the implementation is a good idea. But it does start to print the parameter name in the `show` output rather than writing `parameter_1` etc :) But in any case, someone with a better understanding of the code base needs to weigh in and say whether this could potentially work or not.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1750,6 +1750,7 @@ macro NLparameter(m, ex, extra...)
     # TODO: NLparameters are not registered in the model because we don't yet
     # have an anonymous version.
     macro_code = _macro_assign_and_return(creation_code, variable,
-                                          Containers._get_name(c))
+                                          Containers._get_name(c),
+                                          model_for_registering = esc_m)
     return _finalize_macro(esc_m, macro_code, __source__)
 end
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1750,7 +1750,6 @@ macro NLparameter(m, ex, extra...)
     # TODO: NLparameters are not registered in the model because we don't yet
     # have an anonymous version.
     macro_code = _macro_assign_and_return(creation_code, variable,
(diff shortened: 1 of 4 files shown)


odow (commented):
One issue is that this is actually a breaking change because this will now error (where previously it worked):
```Julia
@variable(model, x)
@NLparameter(model, x == 1)
```

The code to look up a name isn't great (this only works for single parameters) but it is an improvement. It also only prints in subexpressions, not pure parameters
https://github.com/jump-dev/JuMP.jl/blob/442aee8efd5c7a433edffe49bc45b13230904b53/src/print.jl#L752-L758

Tests are failing here: https://github.com/jump-dev/JuMP.jl/blob/442aee8efd5c7a433edffe49bc45b13230904b53/test/print.jl#L250-L255

If you have code that would benefit from registering parameters, you can just go
```Julia
model[:p] = @NLparameter(model, p == 1)
```

davidanthoff:
I'm really just interested in the pretty printing at the moment. So maybe the way to change this PR is to _only_ merge the pretty printing side, which would be non breaking, and then I can do the registration step manually on my end?

Having said that, it might make sense to bring in the breaking change of not allowing variables and parameters to have the same name in some new major version? That does seem like a sensible default? But obviously no rush on that.

For now I'll change this PR to only do the pretty printing, and I'll also try to make that a bit more robust.

davidanthoff:
Alright, so this PR now has the registration of parameters removed, and all it does is use the name of a parameter in printing, _if_ that parameter was manually registered with the model.

odow (approved):
Yeah, we could definitely do better than this, but I think it will take a much bigger NLP redesign.

mlubin (approved):

PR #2520, opened 2021-03-09
Concrete optimizer type in CachingOptimizer

blegat:
Here are benchmarks (the timing vary quite a lot depending on the % of GC time) so it's more interesting to look at allocations:
```julia
using JuMP
function _speed(model, n)
    @time x = [@variable(model) for i in 1:n]
    @time for i in 1:n
        set_lower_bound(x[i], 0.0)
    end
end
function speed1(n)
    model = Model()
    _speed(model, n)
end
function speed2(n)
    model = Model(() -> MOIU.MockOptimizer(MOI.Utilities.Model{Float64}()))
    MOIU.attach_optimizer(model)
    _speed(model, n)
end
function speed3(n)
    model = Model(() -> MOIU.MockOptimizer(MOI.Utilities.Model{Float64}()))
    _speed(model, n)
end
speed1(100000)
speed2(100000)
speed3(100000)
```
Before:
```
julia> include("test/perf/many_calls.jl")
  0.016072 seconds (200.05 k allocations: 16.460 MiB, 45.11% gc time)
  0.014611 seconds (300.00 k allocations: 6.104 MiB)
  0.033083 seconds (300.17 k allocations: 29.605 MiB, 19.79% gc time)
  0.059006 seconds (600.07 k allocations: 22.019 MiB)
  0.014988 seconds (200.05 k allocations: 16.460 MiB, 40.04% gc time)
  0.014893 seconds (300.00 k allocations: 6.104 MiB)
```
After:
```
julia> include("test/perf/many_calls.jl")
  0.008772 seconds (200.05 k allocations: 16.460 MiB)
  0.015277 seconds (300.00 k allocations: 6.104 MiB)
  0.023118 seconds (200.17 k allocations: 28.080 MiB)
  0.061183 seconds (200.07 k allocations: 15.916 MiB)
  0.008718 seconds (200.05 k allocations: 16.460 MiB)
  0.013356 seconds (200.00 k allocations: 4.578 MiB)
```
With
```julia
function _speed(model, n)
    @time x = [@variable(model, lower_bound=0.0, upper_bound=1.0, integer=true) for i in 1:n]
    @time for i in 1:n
        fix(x[i], 0.5, force=true)
    end
end
```
Before:
```
  0.029207 seconds (400.05 k allocations: 19.511 MiB)
  0.041292 seconds (600.00 k allocations: 10.681 MiB)
  0.222799 seconds (1.30 M allocations: 78.879 MiB, 19.79% gc time)
  0.216714 seconds (1.30 M allocations: 32.701 MiB)
  0.033890 seconds (400.05 k allocations: 19.511 MiB)
  0.047814 seconds (600.00 k allocations: 10.681 MiB)
```
After:
```
  0.029957 seconds (400.05 k allocations: 19.511 MiB)
  0.044707 seconds (600.00 k allocations: 10.681 MiB)
  0.152942 seconds (200.38 k allocations: 62.094 MiB, 6.97% gc time)
  0.207885 seconds (800.07 k allocations: 25.071 MiB)
  0.030803 seconds (200.05 k allocations: 16.460 MiB)
  0.053592 seconds (600.00 k allocations: 10.681 MiB)
```
The big allocation drop here from `1.3 M` to `200 k` is due to https://github.com/jump-dev/JuMP.jl/blob/master/src/variables.jl#L882-L908 where we use the backend like it's type stable but before this PR, the caching optimizer is `MOIU.CachingOptimizer{MOI.AbstractOptimizer}` and after this PR, it is `MOI.CachingOptimizer{MOIU.MockOptimizer{MOIU.Model{Float64}}}`.

As the `backend` type is not concrete, it seems wasteful to also have a non-concrete optimizer type in the optimizer field of the backend.
We can benefit from the non-concretess of the backend in `set_optimizer` by changing its type.
Another option would be to parametrize the `JuMP.Model` type by the backend type and have a non-concrete optimizer type for the optimizer of the caching optimizer.
We need at least one to be an abstract type so that we can implement `set_optimizer` but it seems wasteful to have both of them.


odow:
Can we try this on some bigger workloads as well? https://github.com/jump-dev/MathOptInterface.jl/pull/1252#issuecomment-790216464

blegat:
The other allocations of `fix` should be fixed by https://github.com/jump-dev/MathOptInterface.jl/pull/1256, https://github.com/jump-dev/JuMP.jl/pull/2521 and https://github.com/jump-dev/MathOptInterface.jl/pull/1276

PR #2522, opened 2021-03-16
Embed the stable MathOptInterface docs into JuMP.

odow:
Not sure if we should do this (ever, or just yet), but I thought I would see what it looks like and get some feedback.

I'll post the preview link once the build finishes.

https://jump.dev/JuMP.jl/previews/PR2522/moi/

The idea is to make it easier for the JuMP docs to link to the MOI documentation,
and provide a single, unified set of documentation for JuMP/MOI. This should allow us to de-duplicate things like "What is a caching optimizer", and it provides a way for JuMP users to see things like "What sets are supported by MOI"

Things are controlled by pinning the MOI version in `docs/Project.toml`, so whenever we release MOI, we should update the pin and tag a new JuMP release to push the new docs.


odow:
Before merging this, we should update to the revised MOI docs (v0.9.21) and rewrite the introduction page to make clear that this is a copy of the MOI documentation.

odow:
So thoughts on this?

joaquimg:
I like this!
Brainstorming: should we add the warning explaining it is a copy of MOI docs in all pages? (I don't know, I liked to see it in the 1st page).

odow:
How's this:
image

blegat (approved):

odow:
Done in #2695

Issue #2424, opened 2021-01-05
Container with a dictionary as the dimension


odow:
@geoffleyland found this wonderfully confusing example:

```Julia

julia> d = Dict("a" => 1, "b" => 2)
Dict{String,Int64} with 2 entries:
  "b" => 2
  "a" => 1

julia> model = Model();

julia> @variable(model, x[(k, v) in d])
1-dimensional DenseAxisArray{VariableRef,1,...} with index sets:
    Dimension 1, Dict("b" => 2,"a" => 1)
And data, a 2-element Array{VariableRef,1}:
 x[("b", 2)]
 x[("a", 1)]

julia> x[("a", 1)]
ERROR: KeyError: key ("a", 1) not found
Stacktrace:
 [1] getindex at ./dict.jl:467 [inlined]
 [2] lookup_index at /Users/oscar/.julia/packages/JuMP/qhoVb/src/Containers/DenseAxisArray.jl:140 [inlined]
 [3] _to_index_tuple at /Users/oscar/.julia/packages/JuMP/qhoVb/src/Containers/DenseAxisArray.jl:149 [inlined]
 [4] to_index at /Users/oscar/.julia/packages/JuMP/qhoVb/src/Containers/DenseAxisArray.jl:167 [inlined]
 [5] getindex(::JuMP.Containers.DenseAxisArray{VariableRef,1,Tuple{Dict{String,Int64}},Tuple{Dict{Pair{String,Int64},Int64}}}, ::Tuple{String,Int64}) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/Containers/DenseAxisArray.jl:182
 [6] top-level scope at REPL[68]:1

julia> x["a" => 1]
x[("a", 1)]
```

I'm not sure if this is intended or not. It makes sense when you think through what is happening, but it isn't obvious! This should either be documented, or fixed so that the pair is added as a tuple.

fonterrageoff:
Thanks Oscar!

So this is not what I thought the problem was at all (I guessed tuples were being compared by reference).  But I'm curious:
```
julia> d = Dict("a" => 1, "b" => 2)
Dict{String,Int64} with 2 entries:
  "b" => 2
  "a" => 1

julia> [(k, v) for (k, v) in d]
2-element Array{Tuple{String,Int64},1}:
 ("b", 2)
 ("a", 1)

julia> [k=>v for (k, v) in d]
2-element Array{Pair{String,Int64},1}:
 "b" => 2
[...]

PR #2529, opened 2021-03-23
Collect axes in DenseAxisArray on construction

odow:
This partially resolves the non-obvious behavior in #2424, because now the dimension is a vector with `Pair` elements, but it doesn't address the printing issue.
```Julia
julia> @variable(model, x[(k, v) in d])
    Dimension 1, ["b" => 2, "a" => 1]
And data, a 2-element Array{VariableRef,1}:
 x[("b", 2)]
 x[("a", 1)]
```

It's more obvious what is going on with a second dimension:
```julia
julia> @variable(model, y[(k, v) in d, 1:2])
    Dimension 1, ["b" => 2, "a" => 1]
    Dimension 2, Base.OneTo(2)
And data, a 2×2 Array{VariableRef,2}:
 y[("b", 2),1]  y[("b", 2),2]
 y[("a", 1),1]  y[("a", 1),2]
```
We should print the key as it is, not how it appears in the macro constructor.

### DIFF
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -112,6 +112,9 @@ function Base.get(
     return i
 end
 
+_collect(x::AbstractVector) = x
+_collect(x) = collect(x)
+
 """
     DenseAxisArray(data::Array{T, N}, axes...) where {T, N}
 
@@ -133,9 +136,10 @@ julia> array[:b, 3]
 4
 ```
 """
-function DenseAxisArray(data::Array{T,N}, axs...) where {T,N}
-    @assert length(axs) == N
-    return DenseAxisArray(data, axs, build_lookup.(axs))
+function DenseAxisArray(data::Array{T,N}, axes...) where {T,N}
+    @assert length(axes) == N
+    new_axes = _collect.(axes)
+    return DenseAxisArray(data, new_axes, build_lookup.(new_axes))
 end
 
 """
--- a/test/Containers/DenseAxisArray.jl
+++ b/test/Containers/DenseAxisArray.jl
@@ -219,7 +219,7 @@ And data, a 0-dimensional $(Array{Int,0}):
         @test a isa Vector{Pair{Symbol,Int}}
         @test_throws KeyError A[(:a, 3), 1]
         @test_throws KeyError A[:a, 3]
-        @test_throws KeyError A[:b => 1, 2]
+        @test_throws KeyError A[:b=>1, 2]
     end
     @testset "AxisLookup" begin
         A = DenseAxisArray([5.0 6.0; 7.0 8.0], [:a, :b], [:a, :b])
--- a/test/Containers/DenseAxisArray.jl
+++ b/test/Containers/DenseAxisArray.jl
@@ -208,6 +208,19 @@ And data, a 0-dimensional $(Array{Int,0}):
             @test k[2] == Containers.DenseAxisArrayKey((2, :b))
         end
     end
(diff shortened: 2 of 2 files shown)


src/Containers/DenseAxisArray.jl
```diff
@@ -133,9 +136,10 @@ julia> array[:b, 3]
 4
 ```
 """
-function DenseAxisArray(data::Array{T,N}, axs...) where {T,N}
-    @assert length(axs) == N
-    return DenseAxisArray(data, axs, build_lookup.(axs))
+function DenseAxisArray(data::Array{T,N}, axes...) where {T,N}
+    @assert length(axes) == N
+    new_axes = _collect.(axes)
```
blegat, line 141: This should be in `build_lookup(::AbstractDict)` maybe

blegat:
So for Dict, `collect` gives you tuples but `iterate` gives you `Pair`s?

blegat:
I think the issue is that `Pair` implements `iterate` so you can do `k, v in pair`. We should disallow doing `(k,v) in... ` in the macros

odow:
The issue was that the axis was a dictionary. It printed
```Julia
julia> @variable(model, x[(k, v) in d])
    Dimension 1, Dict("b" => 2,"a" => 1)
And data, a 2-element Array{VariableRef,1}:
 x[("b", 2)]
 x[("a", 1)]
```
but you had to index
```Julia
x["b" => 2]
```
I argue that having a dictionary (or any iterable) as the axis is a bit weird. We really want an AbstractVector. So when it prints
```Julia
julia> @variable(model, x[(k, v) in d])
    Dimension 1, ["b" => 2, "a" => 1]
And data, a 2-element Array{VariableRef,1}:
 x[("b", 2)]
 x[("a", 1)]
```
It's more obvious what its going on. I haven't marked this as `closing #2424` because we still need to fix the naming in JuMP's macros.

odow:
@blegat bump.

blegat:
Some users might be using custom iterators for the indices and they probably don't want them to be `collect`ed.
The printing shows the iterators from which the indices are iterated over. The `iterate` on `Dict` returns `Pair`s so I don't see the issue.

odow:
I guess we should fix the printing then, so that it prints:
```Julia
julia> @variable(model, x[(k, v) in d])
    Dimension 1, Dict("b" => 2,"a" => 1)
And data, a 2-element Array{VariableRef,1}:
 x["b" => 2]
 x["a" => 1]
```

> I think the issue is that Pair implements iterate

The underlying issue is tuple unpacking:
```Julia
julia> (a, b) = 3 => 4
3 => 4

julia> a
3

julia> b
4
```

along with the fact that
```Julia
@variable(model, x[(k, v) in d])
```
is transformed into something like
```julia
foo = ((k, v),) -> @variable(model, base_name="x[($k,$v)]")
x = model[:x] = map(foo, (a for a in d))
```
It would be less surprising it it was transformed into 
```julia
foo = (kv,) -> @variable(model, base_name="x[$kv]")
x = model[:x] = map(foo, (a for a in d))
```
but you also need the independent `k` and `v` for things like variable bounds.

blegat:
> The underlying issue is tuple unpacking

I agree, as mentioned in https://github.com/jump-dev/JuMP.jl/pull/2529#issuecomment-808701450, I would suggest detecting tuple unpacking in the macro and transform the axes with tuple unpacking in this case. In other case, I don't think collecting would be helpful.

odow:
I'll close this for now and try again.

PR #2535, opened 2021-03-23
[docs] white background for logo

### LINKED ISSUE #2492: JuMP logo on night/dark theme
image

This is what I get on my phone's github app.
One possible fix is to use a picture with white background...

odow:
Not sure if we should merge this. Just checking what it looks like 

Closes #2492


odow:
Preview: https://jump.dev/JuMP.jl/previews/PR2535/index.html#
@joaquimg I don't think this improves anything
image

Other ideas? Documenter doesn't seem to support switching out an image based on the theme.

Issue #808, opened 2016-07-24
Make ``JuMP.register`` optional


joehuchette:
It would be a win from a usability perspective if you can omit registering a nonlinear function and have things work as expected, e.g.

``` jl
foo(x) = x^3
@NLConstraint(m, foo(x) <= 1)
```

mlubin:
This needs a bit of infrastructure work but I think it's feasible.

PR #2536, opened 2021-03-23
Improve the error message for unsupported functions

### LINKED ISSUE #1889: Does JuMP support max and min operator in the constraint
Say I have a problem like this:
`min f(x,y,z)
s.t. c = max(x,y,z)
`
In general, f is nonlinear, so I can't reduce the max function to a equivalent set of linear constraints. How can I solve it using JuMP?

odow:
Closes #1889

### DIFF
--- a/src/_Derivatives/forward.jl
+++ b/src/_Derivatives/forward.jl
@@ -187,7 +187,18 @@ function forward_eval(
                     r += 1
                 end
             else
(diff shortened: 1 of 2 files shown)


odow:
The second commit might close #808... It was easier than expected. Or am I missing something obvious? Not having to register functions would be a big usability win. I'll see if CI passes, and if so, I'll separate it into a different PR.

Issue #2526, opened 2021-03-22
Feasibility checker with function


blegat:
In the `primal_feasibility_report` just introduced by https://github.com/jump-dev/JuMP.jl/pull/2466, we expect the primal value to be given by a `Dict`. In the future, we might rely on `MOIU.eval_variables` to evaluate the function which requires a function mapping the variables to their values as arguments so we will have to create a lambda from the `Dict`. In view of this, we could also accept a function as argument to specify the value of the variables.
To allow the syntax:
```julia
primal_feasibility_report(model, tol) do var_ref
    ...
end
``` 
we should have the `point` as a first argument while it is currently the second argument.

odow:
We made it an `AbstractDict` and not a `Dict` so you can go:
```Julia
struct FunctionalDict{F} <: AbstractDict
    f::F
end
Base.getindex(f::FunctionalDict, key::VariableRef) = f.f(key)

FunctionalDict(x -> value(x))
```

Do we have a concrete use-case for this? I'm not in favor of adding complexity and multiple ways to do something before we really need to.

blegat:
Currently, we convert all function to JuMP before evaluating them. When we'll start looking at performance, we'll notice that we should directly evaluate them at the MOI level hence use:
https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Utilities/functions.jl#L6-L13
so passing a function is more efficient.
There are indeed two use cases. If the user has a costly way to get the value of the variables then he should first create a dict to avoid computing it twice. Otherwise, it's easier and more efficient to just pass a function.

odow:
We already evaluate the functions using a function: 
https://github.com/jump-dev/JuMP.jl/blob/e4320ea4aeb9ce5c155ecae905d7cd953016eb36/src/feasibility_checker.jl#L60-L69
https://github.com/jump-dev/JuMP.jl/blob/e4320ea4aeb9ce5c155ecae905d7cd953016eb36/src/feasibility_checker.jl#L108

odow:
To summarize the developer call:

* `primal_feasibility_report(model::Model, point::AbstractDict)` should stay.
* At some point in the future, we can add a non-breaking `primal_feasibility_report(point::Function model::Model)`
* Even if we add the functional form is added, we should not add `primal_feasibility_report(point::AbstractDict, model::Model)`.
* Before adding the functional form, we should wait until there is evidence that the functional form is more convenient than the dictionary.

[...]

PR #2538, opened 2021-03-24
[ci skip] update NEWS

odow:
There are quite a few new things building up, particularly the updated documentation, so I'll look to tag a new release once some of the open PRs are merged.


blegat:
We should reach a decision for https://github.com/jump-dev/JuMP.jl/issues/2526 before making a release. Maybe a good topic for JuMP-dev? It's best to double check as we won't be able to change this without being breaking

odow:
We can add the function as a new method without breaking anything.

PR #2543, opened 2021-03-26
Improve error message and add warning when re-registering a function

### LINKED ISSUE #2058: "reregister" a function?
Consider the following situation (with a non-sense MWE):

```julia
function optimisation_procedure(reoptimize::Bool)
    f = FMinimiser(2)
    
    model = Model(with_optimizer(Ipopt.Optimizer, print_level=0))
    register(model, :f, 1, f, autodiff=true)
    @variable(model, x)
    @NLobjective(model, Min, f(x))
    optimize!(model)
    
    if reoptimize
        # this could also be a while block
        f = FMinimiser(10)
        register(model, :f_new, 1, f2, autodiff=true)
        @NLobjective(model, Min, f(x))
        optimize!(model)
    end
    
    value(x)
end
```
I need to change the function to be optimised on the fly sometimes to reoptimise again. In these cases, the `register()` and the `@NLobjective` need to be reset, former with a new symbol identifier (`:f_new`). This works, but I'd like to have possibly multiple iterations and generating a new symbol every time seems awkward.

I assume there is a better way of changing the function on the fly?

odow:
Closes #2058

### DIFF
--- a/docs/src/manual/nlp.md
+++ b/docs/src/manual/nlp.md
@@ -360,9 +360,9 @@ The above code creates a JuMP model with the objective function
     expression algebraically in the macro.
 
 !!! warning
-    User-defined functions cannot be re-registered, and will not update if you
+    User-defined functions cannot be re-registered and will not update if you
     modify the underlying Julia function. If you want to change a user-defined
-    function between solves, rebuild the model, or use a different name. To use
+    function between solves, rebuild the model or use a different name. To use
     a different name programatically, see [Raw expression input](@ref).
 
 ### Register a function and gradient
--- a/docs/src/manual/nlp.md
+++ b/docs/src/manual/nlp.md
@@ -110,7 +110,7 @@ julia> @NLparameter(model, p[i = 1:2] == i)
 ```
 
 !!! info
-    A parameter is not an optimization variable. It must be fixed to a value with 
+    A parameter is not an optimization variable. It must be fixed to a value with
     `==`. If you want a parameter that is `<=` or `>=`, create a variable instead
     using [`@variable`](@ref).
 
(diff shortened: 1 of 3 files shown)


docs/src/manual/nlp.md
```diff
@@ -359,6 +359,12 @@ The above code creates a JuMP model with the objective function
     some cases. Only use a user-defined function if you cannot write out the
     expression algebraically in the macro.
 
+!!! warning
+    User-defined functions cannot be re-registered, and will not update if you
```
rtwalker, line 363: nit: no comma before "and"

docs/src/manual/nlp.md
```diff
@@ -359,6 +359,12 @@ The above code creates a JuMP model with the objective function
...
+    modify the underlying Julia function. If you want to change a user-defined
+    function between solves, rebuild the model, or use a different name. To use
```
rtwalker, line 365: nit: no comma before "or"

src/_Derivatives/types.jl
```diff
@@ -117,8 +122,13 @@ function register_univariate_operator!(
...
     fprimeprime,
 )
-    haskey(r.univariate_operator_to_id, s) &&
-        error("Operator $s has already been defined")
+    if haskey(r.univariate_operator_to_id, s)
+        error("""The univariate operator $s has already been defined.
+
+        User-defined functions can not be re-registered. If you want to
+        re-register a function between solves (e.g., to modify the user-defined
+        function), rebuild the model, or use a different function name.""")
```
rtwalker, line 130: same here

src/_Derivatives/types.jl
```diff
@@ -98,8 +98,13 @@ function register_multivariate_operator!(
...
     f::MOI.AbstractNLPEvaluator,
 )
-    haskey(r.multivariate_operator_to_id, s) &&
-        error("Operator $s has already been defined")
+    if haskey(r.multivariate_operator_to_id, s)
+        error("""The multivariate operator $s has already been defined.
+
+        User-defined functions can not be re-registered. If you want to
+        re-register a function between solves (e.g., to modify the user-defined
+        function), rebuild the model, or use a different function name.""")
```
rtwalker, line 106: same here

PR #2546, opened 2021-03-30
Remove un-used code

odow:
It seems we actually have
https://github.com/jump-dev/JuMP.jl/blob/3e517d7660131ef1ada47042182a192d611189a6/src/variables.jl#L166-L195


odow:
@blegat I guess this is accidental?

odow:
@blegat Bump.

blegat (approved):

PR #2548, opened 2021-03-30
Make some moi_xxx functions private

odow:
Do we have a policy on how we un-export functions that were mistakenly exported? (i.e., do they need deprecation warnings?)

Here we clearly want people to use `mode`, `add_constraint`, and `bridge_constraints`, not their `moi_xxx` equivalents.

Before 1.0, we should go through all exported functions and check whether we want to export them.


blegat (approved):

mlubin:
This is a breaking change, we should wait for the next 0.xx release:
https://github.com/zavalab/Plasmo.jl/blob/69f564f56068529502bee658cd9d8d096039143d/src/optimizer_interface.jl#L4

odow:
Closing for now because #2610 is changing some things. We can revisit.

PR #2550, opened 2021-03-31
Allow iterators of unknown size in macros

### LINKED ISSUE #1915: [Useability] `@constraint` and `@variable`: cannot use an iterator
I am working with graphs, basically generating one variable per edge and then adding constraints. However, when I cannot use the usual syntax `@variable(m, x[e in Iterators.filter(k -> true, edges(g))])`, due to `Iterators.filter(k -> true, edges(g))` returning an iterator whose size is unknown a priori: 

```
ERROR: MethodError: no method matching length(::Base.Iterators.Filter{getfield(Main, Symbol("##1670#1674")),LightGraphs.SimpleGraphs.SimpleEdgeIter{SimpleDiGraph{Int64}}})
Closest candidates are:
  length(::Core.SimpleVector) at essentials.jl:582
  length(::Base.MethodList) at reflection.jl:732
  length(::Core.MethodTable) at reflection.jl:806
  ...
```

An minimum "working" example: 

```julia
using JuMP, LightGraphs
g = PathGraph(3)
m = Model()
@variable(m, x[e in Iterators.filter(k -> true, edges(g))] = 0)
```

For anyone else facing this issue, the solution is simple: put a `collect` to transform the iterator into an array. 

Shouldn't JuMP be doing it automatically? A test could be implemented on `IteratorSize(IterType)`, to return more specific errors: if it is `IsInfinite`, it makes no sense; if it is `SizeUnknown`, JuMP should make an array out of the iterator first.

odow:
Closes #1915

### DIFF
--- a/src/Containers/vectorized_product_iterator.jl
+++ b/src/Containers/vectorized_product_iterator.jl
@@ -44,8 +44,12 @@ For instance:
 struct VectorizedProductIterator{T}
     prod::Iterators.ProductIterator{T}
 end
+function _collect(x)
+    # Collect iterators with unknown size so they can be used as axes.
+    return Base.IteratorSize(x) == Base.SizeUnknown() ? collect(x) : x
+end
 function vectorized_product(iterators...)
-    return VectorizedProductIterator(Iterators.product(iterators...))
+    return VectorizedProductIterator(Iterators.product(_collect.(iterators)...))
 end
 function Base.IteratorSize(
     ::Type{<:VectorizedProductIterator{<:Tuple{Vararg{Any,N}}}},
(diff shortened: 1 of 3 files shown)


src/Containers/vectorized_product_iterator.jl
```diff
@@ -44,8 +44,20 @@ For instance:
 struct VectorizedProductIterator{T}
     prod::Iterators.ProductIterator{T}
 end
+
+function _collect(x)
```
blegat, line 48: Maybe use a trait ? Call `_collect(x) = _collect(Base.IteratorSize(x), x)`

odow: Done.

blegat (approved):

Issue #2457, opened 2021-02-10
Inference failure of SparseAxisArray


odow:
Julia can't infer the concrete return type of a `SparseAxisArray` returned by `@container`.
```Julia
julia> f(n) = Containers.@container([i = 1:n, j = i+1:n], i + j)
f (generic function with 1 method)

julia> f(2)
JuMP.Containers.SparseAxisArray{Int64,2,Tuple{Int64,Int64}} with 1 entry:
  [1, 2]  =  3

julia> @code_warntype f(2)
Variables
  #self#::Core.Compiler.Const(f, false)
  n::Int64
  #1::var"#1#4"
  #2::var"#2#5"{Int64}
  #3::var"#3#6"{Int64}

Body::JuMP.Containers.SparseAxisArray
1 ─ %1  = JuMP.Containers.container::Core.Compiler.Const(JuMP.Containers.container, false)
│         (#1 = %new(Main.:(var"#1#4")))
│   %3  = #1::Core.Compiler.Const(var"#1#4"(), false)
│   %4  = JuMP.Containers.nested::Core.Compiler.Const(JuMP.Containers.nested, false)
│   %5  = Main.:(var"#2#5")::Core.Compiler.Const(var"#2#5", false)
│   %6  = Core.typeof(n)::Core.Compiler.Const(Int64, false)
│   %7  = Core.apply_type(%5, %6)::Core.Compiler.Const(var"#2#5"{Int64}, false)
│         (#2 = %new(%7, n))
│   %9  = #2::var"#2#5"{Int64}
│   %10 = Main.:(var"#3#6")::Core.Compiler.Const(var"#3#6", false)
│   %11 = Core.typeof(n)::Core.Compiler.Const(Int64, false)
│   %12 = Core.apply_type(%10, %11)::Core.Compiler.Const(var"#3#6"{Int64}, false)
│         (#3 = %new(%12, n))
│   %14 = #3::var"#3#6"{Int64}
│   %15 = (%4)(%9, %14)::JuMP.Containers.NestedIterator{Tuple{var"#2#5"{Int64},var"#3#6"{Int64}},JuMP.Containers.var"#21#23"}
│   %16 = (%1)(%3, %15)::JuMP.Containers.SparseAxisArray
└──       return %16
```

but it can with `DenseAxisArray`:
```Julia
julia> g(n) = Containers.@container([i = 2:n, j = 2:n], i + j)
g (generic function with 1 method)

julia> g(3)
[...]

PR #2553, opened 2021-04-06
Improve types of empty SparseAxisArrays

### LINKED ISSUE #2552: type unstability with container generation
```julia 
julia> model = Model(Gurobi.Optimizer)

julia> @variable(model, a[x in 1:10; x < 5], Bin)
JuMP.Containers.SparseAxisArray{VariableRef, 1, Tuple{Int64}} with 4 entries:

julia> @variable(model, b[x in 1:10; x > 15], Bin)
JuMP.Containers.SparseAxisArray{Any, 1, Tuple{Any}} with 0 entries
```
Here, I would expect the second example to return a `JuMP.Containers.SparseAxisArray{VariableRef, 1, Tuple{Int64}}`
I don't know if the current behaviour is expected, but it can cause type unstability if we don’t check the emptiness of the resulting container beforehand.

odow:
Closes #2552 

The call to `return_types` happens at runtime and is slow, but that's not a problem for now because things don't infer normally: #2457

### DIFF
--- a/src/Containers/container.jl
+++ b/src/Containers/container.jl
@@ -137,7 +137,9 @@ function _container_dict(
     return Dict{K,Any}()
 end
 
-_container_dict(::Any, ::Any, K::Type{<:NTuple{N,Any}}) where {N} = Dict{K,Any}()
+function _container_dict(::Any, ::Any, K::Type{<:NTuple{N,Any}}) where {N}
+    return Dict{K,Any}()
+end
 
 function _container_dict(
     ::Type{Union{}},
--- a/src/Containers/container.jl
+++ b/src/Containers/container.jl
@@ -113,13 +113,24 @@ function container(f::Function, indices, ::Type{SparseAxisArray})
     mappings = Base.Generator(I -> I => f(I...), indices)
     # Same as `Dict(mapping)` but it will error if two indices are the same.
     data = NoDuplicateDict(mappings)
-    dict = data.dict
-    if isempty(dict)
-        # If `dict` is empty, it was not able to determine the type
-        # of the key hence the type of `dict` is `Dict{Any, Any}`.
-        # This is an issue since `SparseAxisArray` needs the key
-        # type to be a tuple.
-        dict = Dict{_eltype_or_any(indices),Any}()
+    if !isempty(data.dict)
+        return SparseAxisArray(data.dict)
+    else
+        # If `dict` is empty, it was not able to determine the type of the key.
+        # To make a best-guess attemp, collect all of the keys excluding the
+        # conditional statement (these must be defined, because the conditional
+        # applies to the lowest-level of the index loops), then get the eltype
+        # of the result.
+        el = eltype(collect(nested(indices.iterators...)))
+        # If the eltype is a tuple, we can use that as the key for the
+        # SparseAxisArray. If it isn't, it may be something like `Any`, so we
+        # should just use an `NTuple{N,Any}` of appropriate length.
+        K = el <: Tuple ? el : _eltype_or_any(indices)
+        # For the key-type of the SparseAxisArray, use `Base.return_types`. But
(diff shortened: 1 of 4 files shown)


src/Containers/container.jl
```diff
@@ -113,13 +113,43 @@ function container(f::Function, indices, ::Type{SparseAxisArray})
...
+end
+_container_dict(::Any, ::Any, K) = Dict{K,Any}()
+
+# The NoDuplicateDict was not able to infer the element type. To make a
+# best-guess attempt, collect all of the keys excluding the conditional
+# statement (these must be defined, because the conditional applies to the
+# lowest-level of the index loops), then get the eltype of the result.
+function _sparseaxisarray(dict::Dict{Any,Any}, f, indices::NestedIterator)
+    @assert isempty(dict)
+    el = eltype(collect(nested(indices.iterators...)))
```
blegat, line 153: Why not `el = typeof(first(nested(indices.iterators...))))` ?
Instead of using `return_types` (which is discouraged), we could do `eltype(f(first(nested(indices.iterators...))))`

odow: Because the elements are not guaranteed to have the same type.

blegat: I think that's the advantage of `Base.@default_eltype`, see https://github.com/jump-dev/JuMP.jl/issues/2552#issuecomment-814717126
They call `return_types` on `_iterator_upper_bound` with the `Generator` directly.

src/Containers/container.jl
```diff
@@ -113,13 +113,51 @@ function container(f::Function, indices, ::Type{SparseAxisArray})
...
+
+# The NoDuplicateDict was able to infer the element type.
+_sparseaxisarray(dict::Dict, ::Any, ::Any) = SparseAxisArray(dict)
+
+function _container_dict(
+    K::Type{<:NTuple{N,Any}},
+    f::Function,
+    ::Type{<:NTuple{N,Any}},
+) where {N}
+    ret = Base.return_types(f, K)
```
blegat, line 127: Could we get rid of it if we do `Base.@default_eltype(Base.Generator(f, indices))` ?

src/Containers/container.jl
```diff
@@ -113,13 +113,51 @@ function container(f::Function, indices, ::Type{SparseAxisArray})
...
+
+function _container_dict(
+    ::Type{Union{}},
+    ::Function,
+    K::Type{<:NTuple{N,Any}},
+) where {N}
+    return Dict{K,Any}()
+end
+
+_default_eltype(x::NestedIterator) = Base.@default_eltype nested(x.iterators...)
```
blegat, line 152: Why do we need this ? Isn't the condition ignored by `Base.default_eltype` ?

odow: No, because the condition is part of the `iterate` interface.

src/Containers/container.jl
odow: At this point, we have the function `f` and we have argument type `K`, so there is no need for `@default_eltype`.  Calling `Base.@default_eltype(Base.Generator(f, indices))` earlier on doesn't work because it often returns something less-specific than the current approach.

src/Containers/container.jl
blegat: Yes but the compiler can't tell whether the condition is true or false so it assumes it might be true and add the returned type to the `Union`. At leas that's what I understand

odow: With the condition, it's more likely to return `Union{}` instead of inferring properly. They don't return the same thing.

odow:
@blegat can you take another look please?

blegat (approved):

blegat:
Maybe just add a few comments on the code clarifying why we deviate from what Julia does for `Dict` (as we discussed in the open comments)

PR #2555, opened 2021-04-10
Spell check

remi-garcia:
I am checking for misspellings, using tools obviously. It is still in progress but I have a few questions:
- Should we use British English or American English? (e.g. for modelled/modeled)
- I replaced "arguement" with "argument" because I understood that the second is correct in British English and American English, is that right?


docs/src/tutorials/Getting started/an_introduction_to_julia.jl
```diff
@@ -91,12 +91,12 @@ nothing #hide
...
 
 # !!! tip
-#     To make π (and most other greek letters), type `\pi` and then press
+#     To make π (and most other Greek letters), type `\pi` and then press
 #     `[TAB]`.
 
 typeof(π)
 
-# Athough if we do math with irrational numbers, they get converted to
+# Although if we do math with irrational numbers, they get converted to
```
remi-garcia, line 99: I am not sure this sentence should start with "Although". I would rather use "However".

mlubin:
This is very helpful, thanks!

> Should we use British English or American English?

American

docs/src/tutorials/Getting started/an_introduction_to_julia.jl
odow:
```suggestion
# However, if we do math with irrational numbers, they get converted to
```

remi-garcia:
Checking is done. I also replaced some http links with https ones.

odow (approved):
Very useful! What tools did you use for this?

remi-garcia:
Oh, nothing fancy. I just managed to have my spell check working in Atom (which was harder than it looks). Then it was mostly by hand.

Sorry if 
> using tools obviously

was misleading, it was to emphasize on the fact that despite my poor English corrections should be fine.

PR #2562, opened 2021-04-15
Fix handling of nonbasic free variables in lp_sensitivity_report

### LINKED ISSUE #2560: LP sensitivity summary doesn't handle nonbasic free variables correctly
x-ref https://github.com/jump-dev/JuMP.jl/issues/2244#issuecomment-820641815

```Julia
julia> model = Model(GLPK.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: EMPTY_OPTIMIZER
Solver name: GLPK

julia> @variable(model, x >= 1)
x

julia> @variable(model, y)
y

julia> @objective(model, Min, x)
x

julia> optimize!(model)

julia> lp_sensitivity_report(model)
ERROR: AssertionError: size(B, 1) == size(B, 2)
Stacktrace:
 [1] lp_sensitivity_report(model::Model; atol::Float64)
   @ JuMP ~/.julia/packages/JuMP/MIPb6/src/lp_sensitivity2.jl:78
 [2] lp_sensitivity_report(model::Model)
   @ JuMP ~/.julia/packages/JuMP/MIPb6/src/lp_sensitivity2.jl:64
 [3] top-level scope
   @ REPL[46]:1
```

cc @geoffleyland

odow:
Closes #2560 

Do we need to introduce a specific `VariableBasisStatus` to MOI?


src/lp_sensitivity2.jl
```diff
@@ -423,12 +429,25 @@ function _standard_form_basis(model::Model, std_form)
...
     end
+    for (x, col) in std_form.columns
+        if variable_seen[col]
+            continue
+        end
+        # MOI doesn't provide a way of distinguishing between basic and
+        # non-basic variables if they have no bounds. As a heuristic, assume the
+        # variable is non-basic if it is at 0.
+        #
+        # This doesn't cover degenerate problems (basic variable at 0), or
```
odow, line 447: @geoffleyland found a case where this happened. We could detect when the basis is not square and throw a better error message. The work-around is to add a bound so the variable isn't free.

odow:
Merging for now. we can revisit once we have `VariableBasisStatus`.

PR #2568, opened 2021-04-21
fix `Base.warn_once`

### LINKED ISSUE #2567: Deprecated `Base.warn_once`
Running the following example from the documentation of `register`
```
  model = Model()
  @variable(model, x[1:2])
  g(x::T, y::T) where {T<:Real} = x * y
  function ∇g(g::Vector{T}, x::T, y::T) where {T<:Real}
      g[1] = y
      g[2] = x
      return
  end
  register(model, :g, 2, g, ∇g; autodiff = true)
```
yields to an `warn_once` error. The function `Base.warn_once` is actually deprecated.

tmigot:
Fix #2567

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -1907,9 +1907,8 @@ function register(
             fprimeprime,
         )
     else
-        autodiff == false || Base.warn_once(
-            "autodiff=true ignored since gradient is already provided.",
-        )
+        autodiff == false ||
+            @warn("autodiff=true ignored since gradient is already provided.")
         m.nlp_data.largest_user_input_dimension =
             max(m.nlp_data.largest_user_input_dimension, dimension)
         d = _UserFunctionEvaluator(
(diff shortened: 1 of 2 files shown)


odow (changes requested):
Thanks! See the format check: 
https://github.com/jump-dev/JuMP.jl/pull/2568/checks?check_run_id=2396149942

You can manually make the change suggested in the patch (probably easiest since this is a small change), or install and run JuliaFormatter:
https://jump.dev/JuMP.jl/dev/developers/style/#JuliaFormatter

odow (approved):

tmigot:
> Thanks! See the format check:
> https://github.com/jump-dev/JuMP.jl/pull/2568/checks?check_run_id=2396149942
> 
> You can manually make the change suggested in the patch (probably easiest since this is a small change), or install and run JuliaFormatter:
> https://jump.dev/JuMP.jl/dev/developers/style/#JuliaFormatter

Thank you!

odow:
Ah. Turns out the line wasn't tested, which is how it escaped detection. I added a test. Thanks for finding it!

Issue #2523, opened 2021-03-16
Replace Containers with an external package


odow:
@oxinabox raised the possibility of replacing JuMP's `Containers` package with a different package.

More than 10% of our current open issues are [related to containers](https://github.com/jump-dev/JuMP.jl/issues?q=is%3Aissue+is%3Aopen+label%3A%22Category%3A+Containers%22), so it makes sense to offload this to a specialized package.

A wholesale replacement is probably too breaking for JuMP 1.0, but we should think about ways to make it easier to create containers of JuMP variables using external containers.

DenseAxisArray:
* https://github.com/mcabbott/AxisKeys.jl
* https://github.com/Tokazama/AbstractIndices.jl
* https://github.com/rafaqz/DimensionalData.jl

SparseAxisArray:
 * Just a standard `Dict`?
 * Or a `DenseAxisArray` with holes?

blegat:
+1 for moving it a separate package. https://github.com/mcabbott/AxisKeys.jl could not replace `DenseAxisArrays` as indexing with brackets expect the indexing of the underlying `Array`. If we have a separate package, we could define clearly the API needed (I think it's the `container` function) so that external packages can define it and be usable from JuMP without having to have JuMP in their dependencies.

odow:
I think I'm okay keeping the `Containers` submodule. It gives us greater control over what happens as the default for JuMP. But we should add documentation to make it easy to use different Containers.

At the moment, all we have is https://jump.dev/JuMP.jl/stable/variables/#User-defined-containers

odow:
If we remove https://github.com/jump-dev/JuMP.jl/blob/e2b6ee959c827d60f0c8178a813fddb6e21048b2/src/Containers/macro.jl#L188-L199, then we can dispatch on any container type. 

```Julia
julia> using JuMP

julia> function Containers.container(f::Function, indices, ::Type{Dict})
           return Dict(i => f(i...) for i in indices)
       end

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: NO_OPTIMIZER
Solver name: No optimizer attached.

julia> @variable(model, x[1:2, 1:2], container=Dict)
Dict{Tuple{Int64, Int64}, VariableRef} with 4 entries:
  (1, 2) => x[1,2]
  (1, 1) => x[1,1]
  (2, 2) => x[2,2]
  (2, 1) => x[2,1]
```

Issue #2176, opened 2020-02-24
Should SparseAxisArray be an AbstractDict?


matbesancon:
Example of surprising behaviour:
```julia
julia> m = Model()
julia> S = 1:10
julia> @variable(m, x[i=10:20,s=S; i == s], Bin)
JuMP.Containers.SparseAxisArray{VariableRef,2,Tuple{Int64,Int64}} with 1 entry:
  [10, 10]  =  x[10,10]

julia> collect(x)
ERROR: MethodError: no method matching size(::JuMP.Containers.SparseAxisArray{VariableRef,2,Tuple{Int64,Int64}})
Closest candidates are:
  size(::AbstractArray{T,N}, ::Any) where {T, N} at abstractarray.jl:38
  size(::BitArray{1}) at bitarray.jl:77
  size(::BitArray{1}, ::Integer) at bitarray.jl:81
```

`AbstractArray` must normally implement `size` and `getindex`. Given the functions implemented, it would probably make more sense for `SparseAxisArray` to subtype AbstractDict (with `getindex(s, k::K)`

odow:
This seems like a reasonable idea. There are a number of open issues relating to SparseAxisArrays, mainly relating to the fact that we call it an `Array`, but it doesn't implement the array interface.

matbesancon:
Looking at the code base, it also looks like we are not leveraging the current sub-typing anywhere, I'll try to put together a PR

blegat:
It's not really an `AbstractDict` either. An `AbstractDict` is like a vector of pairs mapping keys to values. It's length is the number of pairs and it has dimension 1.
The `SparseAxisArray` has dimension `N`, it has a size, which is the number of different key in each dimension but computing it is not O(1). We could precompute the size at construction since we go over all keys but it might not be the only think missing for `collect` to work.
What should be the result of calling `collect` on a `SparseAxisArray` ? It might be an `N`-dimensional array with missing entries that are filled with `missing` ?
If you want a dictionary interpretation of the `SparseAxisArray`, you can do `x.data`.

matbesancon:
> but it might not be the only think missing for collect to work

as long as we have getindex and size defined all other array-related functions have fallbacks.

Not sure it does have the array semantics given that elements are missing, one cannot colon-subindex `arr[:,1]` nor multiply with another array of values

blegat:
You could colon-index it, it would give a SparseAxisArray back.
[...]

PR #2570, opened 2021-04-21
Allow overloading of container types

odow:
There didn't seem to be a good reason to have the error.

This should make it easy to add `AxisKeys` as a backing type:
Closes https://github.com/jump-dev/JuMP.jl/issues/2523

And we now currently support `<:AbstractDict`, which avoids the need to make `SparseAxisArray` an `AbstractDict`:
Closes https://github.com/jump-dev/JuMP.jl/issues/2176

### DIFF
--- a/src/Containers/container.jl
+++ b/src/Containers/container.jl
@@ -163,3 +163,16 @@ function _sparseaxisarray(dict::Dict{Any,Any}, f, indices)
     d = _container_dict(_default_eltype(indices), f, _eltype_or_any(indices))
     return SparseAxisArray(d)
 end
+
+function container(f::Function, indices, D::Type{<:AbstractDict})
+    # Don't use length-1 tuples if there is only one index!
+    key(i) = length(i) == 1 ? i[1] : i
+    return D(key(i) => f(i...) for i in indices)
+end
+
+function container(::Function, ::Any, D::Type)
+    return error(
+        "Unable to build a container with the provided type $(D). Implement " *
+        "`Containers.container`.",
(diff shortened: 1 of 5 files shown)


test/Containers/macro.jl
```diff
@@ -94,4 +94,41 @@ using Test
...
+            z[i = 1:3, j = 1:3; isodd(i + j)],
+            i + j,
+            container = Dict
+        )
+        @test z isa Dict{Tuple{Int,Int},Int}
+        @test length(z) == 4
+        @test z[1, 2] == 3
+    end
+    @testset "Invalid container" begin
+        @test_throws ErrorException Containers.@container(
```
blegat, line 128: Maybe check the error message to make sure it triggers the right error ?

src/Containers/container.jl
```diff
@@ -163,3 +163,16 @@ function _sparseaxisarray(dict::Dict{Any,Any}, f, indices)
     d = _container_dict(_default_eltype(indices), f, _eltype_or_any(indices))
     return SparseAxisArray(d)
 end
+
+function container(f::Function, indices, D::Type{<:AbstractDict})
+    # Don't use length-1 tuples if there is only one index!
+    key(i) = length(i) == 1 ? i[1] : i
```
blegat, line 169: Is this type stable ? Why is this defined inside `container` ? We could have a function defined outside with two method and detect that it's a tuple with one element in the signature to ensure type stability.

odow:
@blegat how's this?

PR #2572, opened 2021-04-22
Add unsafe_backend

### LINKED ISSUE #2565: Add a function to get inner-most optimizer
Something we see frequently is people doing
```Julia
model = Model(Clp.Optimizer)
glpk = backend(model).optimizer.model.optimizer
```

This touches internal fields, and prevents us from changing the layers between `model` and `glpk` (https://github.com/jump-dev/JuMP.jl/pull/2513).

We should add a function 
```Julia
function backend_optimizer(model::Model)
```
that retrieves the inner optimizer.

odow:
Closes #2565 

This solves two pain points for users:
 * No longer need things like `backend(model).optimizer.model.optimizer` or `backend(model).optimizer.model`, and knowing which to use depends on the solver.
 * No longer have issues where they go `backend(model).optimizer.model.optimizer` and the solver isn't initialized because `attach_optimizer` hasn't been called yet.

### DIFF
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -241,8 +241,8 @@ with optimizer MOIB.LazyBridgeOptimizer{GLPK.Optimizer}
 The backend is a `MOIU.CachingOptimizer` in the state `EMPTY_OPTIMIZER` and mode
 `AUTOMATIC`.
 
-Alternatively, use [`unsafe_backend`](@ref) to access the `GLPK.Optimizer`
-object at the bottom of the stack.
+Alternatively, use [`unsafe_backend`](@ref) to access the innermost 
+`GLPK.Optimizer` object:
 ```jldoctest models_backends
 julia> unsafe_backend(model)
 A GLPK model
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -241,12 +241,17 @@ with optimizer MOIB.LazyBridgeOptimizer{GLPK.Optimizer}
 The backend is a `MOIU.CachingOptimizer` in the state `EMPTY_OPTIMIZER` and mode
 `AUTOMATIC`.
 
-Alternatively, pass `innermost = true` to access the `GLPK.Optimizer` object at 
-the bottom of the stack.
+Alternatively, use [`unsafe_backend`](@ref) to access the `GLPK.Optimizer`
+object at the bottom of the stack.
 ```jldoctest models_backends
-julia> backend(model; innermost = true)
+julia> unsafe_backend(model)
 A GLPK model
 ```
+
+!!! warning
+    [`backend`](@ref) and [`unsafe_backend`](@ref) are advanced routines. Read
+    their docstrings to understand the cavets of their usage. You should only
+    call them if you wish to access low-level solver-specific functions.
 
 ### CachingOptimizer
 
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -250,7 +250,7 @@ A GLPK model
 
 !!! warning
     [`backend`](@ref) and [`unsafe_backend`](@ref) are advanced routines. Read
-    their docstrings to understand the cavets of their usage. You should only
+    their docstrings to understand the caveats of their usage. You should only
     call them if you wish to access low-level solver-specific functions.
 
 ### CachingOptimizer
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -241,10 +241,10 @@ with optimizer MOIB.LazyBridgeOptimizer{GLPK.Optimizer}
 The backend is a `MOIU.CachingOptimizer` in the state `EMPTY_OPTIMIZER` and mode
 `AUTOMATIC`.
 
-Pass `innermost = true` to access the `GLPK.Optimizer` object at the bottom of
-the stack.
+Alternatively, pass `innermost = true` to access the `GLPK.Optimizer` object at 
+the bottom of the stack.
 ```jldoctest models_backends
-julia> b = backend(model; innermost = true)
+julia> backend(model; innermost = true)
 A GLPK model
 ```
 
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -241,6 +241,13 @@ with optimizer MOIB.LazyBridgeOptimizer{GLPK.Optimizer}
 The backend is a `MOIU.CachingOptimizer` in the state `EMPTY_OPTIMIZER` and mode
 `AUTOMATIC`.
 
+Pass `innermost = true` to access the `GLPK.Optimizer` object at the bottom of
+the stack.
+```jldoctest models_backends
+julia> b = backend(model; innermost = true)
+A GLPK model
+```
+
 ### CachingOptimizer
 
 A `MOIU.CachingOptimizer` is an MOI layer that abstracts the difference between
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -383,9 +383,8 @@ inner = unsafe_backend(model)
 ```
 
 Moreover, if you modify the JuMP model, the reference you have to the backend
-(i.e., `inner` in the example above) may be out-dated. Always get a new
-reference by calling `unsafe_backend` before calling any low-level
-functionality.
+(i.e., `inner` in the example above) may be out-dated, and you should call
+[`MOI.Utilities.attach_optimizer`](@ref) again.
 
 This function is also unsafe in the reverse direction: if you modify the unsafe
 backend, e.g., by adding a new constraint to `inner`, the changes may be
@@ -402,20 +401,12 @@ model = Model(GLPK.Optimizer)
 @variable(model, x >= 0)
 MOI.Utilities.attach_optimizer(model)
 glpk = unsafe_backend(model)
-# ... use GLPK ...
-@objective(model, Min, 2x + 1)
-MOI.Utilities.attach_optimizer(model)
-glpk = unsafe_backend(model)
-# ... use GLPK again...
 ```
 Use:
 ```julia
 model = direct_model(GLPK.Optimizer())
 @variable(model, x >= 0)
 glpk = backend(model)  # No need to call `attach_optimizer`.
-# ... use GLPK ...
-@objective(model, Min, 2x + 1)
-# ... use GLPK again. No need to call `backend` twice ...
 ```
 """
 unsafe_backend(model::Model) = unsafe_backend(backend(model))
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -374,9 +374,12 @@ See also: [`backend`](@ref).
 This function is unsafe for two main reasons.
 
 First, the formulation and order of variables and constraints in the unsafe
-backend may be different to the variables and constraints in `model` due to
-bridges. There is no solution to this. Use the alternative suggested below
-instead.
+backend may be different to the variables and constraints in `model`. This 
+can happen because of bridges, or because the solver requires the variables or 
+constraints in a specific order. In addition, the variable or constraint index 
+returned by [`index`](@ref) at the JuMP level may be different to the index of 
+the corresponding variable or constraint in the `unsafe_backed`. There is no 
+solution to this. Use the alternative suggested below instead.
 
 Second, the `unsafe_backend` may be empty, or lack some modifications made to
 the JuMP model. Thus, before calling `unsafe_backend` you should first call
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -332,23 +332,49 @@ end
 Base.broadcastable(model::Model) = Ref(model)
 
 """
-    backend(model::Model)
+    backend(model::Model; innermost::Bool = false)
 
 Return the lower-level MathOptInterface model that sits underneath JuMP. This
-model depends on which operating mode JuMP is in (see [`mode`](@ref)), and
-whether there are any bridges in the model.
+model depends on which operating mode JuMP is in (see [`mode`](@ref)).
 
-If JuMP is in `DIRECT` mode (i.e., the model was created using
-[`direct_model`](@ref)), the backend will be the optimizer passed to
-[`direct_model`](@ref).
+ * If JuMP is in `DIRECT` mode (i.e., the model was created using
+   [`direct_model`](@ref)), the backend will be the optimizer passed to
+   [`direct_model`](@ref).
+ * If JuMP is in `MANUAL` or `AUTOMATIC` mode, the backend is a
+   `MOI.Utilities.CachingOptimizer`.
 
-If JuMP is in `MANUAL` or `AUTOMATIC` mode, the backend is a
-`MOI.Utilities.CachingOptimizer`.
+For most solvers, JuMP will add additional `CachingOptimizer` and
+`LazyBridgeOptimizer` layers as needed. To skip these layers and return the
+innermost optimizer object, pass `innermost = true`.
 
-This function should only be used by advanced users looking to access low-level
-MathOptInterface or solver-specific functionality.
+**This function should only be used by advanced users looking to access
+low-level MathOptInterface or solver-specific functionality.**
 """
-backend(model::Model) = model.moi_backend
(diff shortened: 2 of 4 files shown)


src/JuMP.jl
```diff
@@ -332,23 +332,49 @@ end
...
+    end
+end
+
+function _innermost_backend(model::MOIU.CachingOptimizer)
+    if MOIU.state(model) == MOIU.EMPTY_OPTIMIZER
+        MOIU.attach_optimizer(model)
+    elseif MOIU.state(model) == MOIU.NO_OPTIMIZER
+        error(
+            "Unable to get innnermost optimizer because CachingOptimizer is " *
+            "in state NO_OPTIMIZER.",
```
blegat, line 367: We could be less cryptic and say that no optimizer was set at `Model` and `set_optimizer` was not called

src/JuMP.jl
```diff
@@ -332,23 +332,49 @@ end
...
+    if innermost
+        return _innermost_backend(model.moi_backend)
+    else
+        return model.moi_backend
+    end
+end
+
+function _innermost_backend(model::MOIU.CachingOptimizer)
+    if MOIU.state(model) == MOIU.EMPTY_OPTIMIZER
+        MOIU.attach_optimizer(model)
```
blegat, line 363: Not sure about this, why would the user need it to be attached ? Is it because for some solvers, before `copy_to`, the solver is not initialized at all and they can't even set parameters ? In that case, if the user make a modification to the model then the solver might be dropped and when we copy again, his changes might be lost.

src/JuMP.jl
```diff
@@ -332,23 +332,49 @@ end
...
-whether there are any bridges in the model.
+model depends on which operating mode JuMP is in (see [`mode`](@ref)).
 
-If JuMP is in `DIRECT` mode (i.e., the model was created using
-[`direct_model`](@ref)), the backend will be the optimizer passed to
-[`direct_model`](@ref).
+ * If JuMP is in `DIRECT` mode (i.e., the model was created using
+   [`direct_model`](@ref)), the backend will be the optimizer passed to
+   [`direct_model`](@ref).
+ * If JuMP is in `MANUAL` or `AUTOMATIC` mode, the backend is a
```
blegat, line 343: This discussion is only in the case `innermost` is `false`

odow:
How's this? `unsafe_backend`, with a sufficiently weighty "think about doing something else" docstring.

src/JuMP.jl
```diff
@@ -335,21 +335,123 @@ Base.broadcastable(model::Model) = Ref(model)
...
+model = direct_model(GLPK.Optimizer())
+@variable(model, x >= 0)
+glpk = backend(model)
+# ... use GLPK ...
+@objective(model, Min, 2x + 1)
+# ... use GLPK again. No need to call `backend` twice...
+```
+
+"""
+function unsafe_backend(model::Model; attach_caching_optimizers::Bool = true)
```
blegat, line 417: I would default to `false` for `attach_caching_optimizers`

odow: I think we want `true`. Surely if you want to access the low-level model, you want it attached? When would you want it explicitly not attached?

blegat: I understand that you want it attached when you want to print the model but in these cases, we now have writers in JuMP. In which other cases do you want it not attached ?
You don't need to have it attached when your just setting parameters, ...
If it's opt-in, it will force the user to read the docstring before using this option and understand that it can be automatically and silently detached. I would even prefer not having this option and have the user call `attach_optimizer` before calling `unsafe_backend`.
Having `true` by default can have 2 use cases: 1) saves some keystrokes and 2) allow it to just work for a user not understanding these `CachingOptimizer` layers.
I don't think 1) is so important and for 2), the same users will probably be even more confused when the model is detached and his changes are lost.

odow: I guess we could just not have a default...

blegat: I meant "In which other cases do you want it attached ?"

> I guess we could just not have a default...

That would make sense, we have to force the user to understand this before using `unsafe_backend`

odow: Calling low level functionality like presolves, fix-and-solves, conflict, etc. The only time you don't want it attached is setting parameters, which can be done via `set_optimizer_attribute`. People should only be accessing this if they want to do advanced stuff. Setting parameters doesn't really count.

blegat: I can imagine that most use cases need this but I'm worried that the changes will be dropped silently in `AUTOMATIC` mode. Maybe we could add a warning if the mode is not `MANUAL` ? I don't thing people realize that even setting the starting values of a variable can detach an optimizer. I still think requiring the user to write
```julia
MOIU.attach_optimizer(model)
JuMP.unsafe_backend(model)
```
is clean and prevents any confusion.

blegat: Also, with JuMP extensions that set optimize hook, attaching is not enough so attaching automatically wouldn't work in all use cases.

odow: > the changes will be dropped silently in AUTOMATIC mode

I'm confused as to what the problem is with this. Can you give a code example? The attach, modify the unsafe backend, and then modify the JuMP model and lose their changes to the unsafe backend? This is explicitly warned about in the docstring. Having a manual attach doesn't change this. (They make changes when it isn't attached? calling `optimize!` is going to clear them anyway.)

> Also, with JuMP extensions that set optimize hook

JuMP extensions shouldn't be using `unsafe_backend`. Can you give a code example of a use-case?

The main use case is for people wanting to call the solver functions like write or presolve on a model. In that situation, people would have to write `attach_optimizer`. The other use-case is setting some parameter, although they should use `set_optimizer_attribute`, so this isn't a good example. Until we actually have a use-case where this is a bad thing, I think we should keep it...

blegat: > The attach, modify the unsafe backend, and then modify the JuMP model and lose their changes to the unsafe backend?

Yes.

> Having a manual attach doesn't change this.

No but since it's the user that attached explicitly then he's responsible for making sure it's not detached.

> JuMP extensions shouldn't be using `unsafe_backend`.

No but the user might use `unsafe_backend` when using an extension, e.g.,
```julia
using SetProg
model = Model()
# ...
@constraint(model, ellipsoid ⊆ square)
unsafe_backend(model)
```
Since the inclusion constraints is transformed in an optimizer hook, it won't be done in `attach_optimizer` hence the backend won't have the constraints associated to this inclusion constraints.
Once the user does
```julia
optimize!(model)
```
then the constraints associated to the inclusion constraints are added which could detach the optimizer.

So by attaching automatically we wouldn't do something that we would recommend to the user. What we should recommend is to attach before calling `unsafe_backend` **and** use `MANUAL` mode so that an error is thrown if the backend is detached when they don't expect it to be. When you are using `unsafe_backend`, you need to control the attachment state of the backend and make sure it is not detached between `unsafe_backend` and `optimize!` so it does not make much sense to be in `AUTOMATIC` mode. Maybe we could even thrown a warning if it's on `AUTOMATIC` mode but that's maybe too much.

odow: I've reinforced the warning. 

The solution for things like adding constraints is for users to use `direct_model` instead. 

Users shouldn't be calling `unsafe_backend` with a JuMP extension that sets an optimizer hook prior to `optimize!`.

This use-case covers 99% of what people are going to want to do (get access to the solver. Call a summary C API). For anything else, they should use `direct_model`.

src/JuMP.jl
```diff
@@ -378,21 +379,25 @@ See also: [`backend`](@ref).
...
+If you modify the JuMP model, the reference you have to the backend may be
+invalid. Using an invalid reference may result in incorrect solutions, errors,
+or Julia crashing. Always get a new reference by calling `unsafe_backend` before
+calling any low-level functionality.
 
-Moreover, if you pass `attach_caching_optimizers = false`, you should assume
-that any `CachingOptimizer`s may be in the `EMPTY_OPTIMIZER` state.
+Moreover, if you modify the unsafe backend, e.g., by adding a new constraint,
+the changes may be silently discarded by JuMP when the model is modified or
+solved. Don't to this unless you understand when and how JuMP attaches and
```
blegat: to -> do

odow: Huh. Something is weird here. Looks like you couldn't see the latest changes. I already fixed this.

src/JuMP.jl
```diff
@@ -335,21 +335,126 @@ Base.broadcastable(model::Model) = Ref(model)
...
+low-level MathOptInterface or solver-specific functionality.**
 
-This function should only be used by advanced users looking to access low-level
-MathOptInterface or solver-specific functionality.
+## Notes
+
+If JuMP is not in `DIRECT` mode, the type returned by `backend` may change
+between any JuMP releases. Therefore, only use the public API exposed by
+MathOptInterface, and do not access internal fields such as
+`backend(model).optimizer.model`. If you require access to the
```
mlubin, line 354: This example of accessing internal fields isn't necessary, it only suggests users to try it out :)

src/JuMP.jl
```diff
@@ -335,21 +335,126 @@ Base.broadcastable(model::Model) = Ref(model)
...
 backend(model::Model) = model.moi_backend
 
+"""
+    unsafe_backend(model::Model; attach_caching_optimizers::Bool = true)
+
+Return the innermost optimizer associated with the JuMP model `model`, skipping
+`CachingOptimizer` and `LazyBridgeOptimizer` layers as needed.
+
+If `attach_caching_optimizers`, `MOIU.attach_optimizer(model)` will be called on
+all `MOIU.CachingOptimizer`s in the stack.
```
mlubin, line 369: before returning the backend.

src/JuMP.jl
```diff
@@ -335,21 +335,126 @@ Base.broadcastable(model::Model) = Ref(model)
...
+
+See also: [`backend`](@ref).
+
+## Unsafe behavior
+
+This function is unsafe because the innermost optimizer may not represent the
+same problem as `model`.
+
+If you modify the JuMP model, the reference you have to the backend may be
+invalid. Using an invalid reference may result in incorrect solutions, errors,
```
mlubin, line 383: What makes it actually invalid? It is just a dangling reference to an object that will be freed by the GC, or do we explicitly invalidate the models (e.g., free licenses)?

src/JuMP.jl
mlubin: I vote for no `attach_caching_optimizers` flag and instead forcing users to write:
```julia
MOIU.attach_optimizer(model)
JuMP.unsafe_backend(model)
```
First of all, `unsafe_backend` looks like a getter that doesn't change the state of the JuMP model. It's a bit surprising that it does by default. Second of all, suppose I'm reading this code that someone else wrote, and I haven't read the long docstring for `unsafe_backend`. The call to `MOIU.attach_optimizer` tells me that something state-changing is happening.

I don't really see the value in saving users one line of code. This is already an advanced and complex operation. And as @blegat mentioned, we can't even cover cases like optimizer hooks.

src/JuMP.jl
blegat: The Julia `Optimizer` object will still be the same, we don't invalidate the reference, we just call `MOI.empty!`

docs/src/manual/models.md
```diff
@@ -241,6 +241,18 @@ with optimizer MOIB.LazyBridgeOptimizer{GLPK.Optimizer}
 The backend is a `MOIU.CachingOptimizer` in the state `EMPTY_OPTIMIZER` and mode
 `AUTOMATIC`.
 
+Alternatively, use [`unsafe_backend`](@ref) to access the `GLPK.Optimizer`
+object at the bottom of the stack.
```
joaquimg, line 245: I would consider using the word innermost here, since its used in other descriptions.

src/JuMP.jl
```diff
@@ -335,21 +335,126 @@ Base.broadcastable(model::Model) = Ref(model)
...
+
+See also: [`unsafe_backend`](@ref).
 """
 backend(model::Model) = model.moi_backend
 
+"""
+    unsafe_backend(model::Model; attach_caching_optimizers::Bool = true)
+
+Return the innermost optimizer associated with the JuMP model `model`, skipping
+`CachingOptimizer` and `LazyBridgeOptimizer` layers as needed.
```
joaquimg, line 366: It will strictly stop after those two?
Or it will work recursively?

src/JuMP.jl
joaquimg: So the backend (the Optimizer) won't be invalid itself, but everything obtained from it might be, right?

src/JuMP.jl
joaquimg: And we can the two lines:
```julia
MOIU.attach_optimizer(model)
JuMP.unsafe_backend(model)
```
In the docstring.

src/JuMP.jl
```diff
@@ -335,21 +335,126 @@ Base.broadcastable(model::Model) = Ref(model)
...
+    model::MOIB.LazyBridgeOptimizer;
+    attach_caching_optimizers::Bool,
+)
+    return unsafe_backend(
+        model.model;
+        attach_caching_optimizers = attach_caching_optimizers,
+    )
+end
+
+unsafe_backend(model::MOI.ModelLike; kwargs...) = model
```
joaquimg, line 456: Are we recommending extensions and layers like Dualization.jl do something?

src/JuMP.jl
odow: Invalid might be the wrong word. I've changed to `out-dated`. 

The `Optimizer` object is still the same, but the inner pointer is probably changed or set to `C_NULL` on `MOI.empty!`, so calling low-level functions could do anything. 

We don't have any API guarantees on how solvers should handle the C models and when they are initialized and freed, only that accessing it through MOI is safe.

src/JuMP.jl
```diff
@@ -335,21 +335,104 @@ Base.broadcastable(model::Model) = Ref(model)
...
+**This function should only be used by advanced users looking to access
+low-level solver-specific functionality. It has a high-risk of incorrect usage.
+We strongly suggest you use the alternative suggested below.**
+
+See also: [`backend`](@ref).
+
+## Unsafe behavior
+
+This function is unsafe because the innermost optimizer may not represent the
+same problem as the JuMP `model`.
```
blegat, line 375: Give as example that the indices might be different than what `index` gives

src/JuMP.jl
```diff
@@ -335,21 +335,104 @@ Base.broadcastable(model::Model) = Ref(model)
...
+For example, instead of:
+```julia
+model = Model(GLPK.Optimizer)
+@variable(model, x >= 0)
+MOI.Utilities.attach_optimizer(model)
+glpk = unsafe_backend(model)
+# ... use GLPK ...
+@objective(model, Min, 2x + 1)
+MOI.Utilities.attach_optimizer(model)
+glpk = unsafe_backend(model)
```
blegat, line 408: Why does this need to be called again? In which case would this new `glpk` reference a different object?

odow: I guess I'm overthinking this. It's actually less unsafe than I thought. It's the inner pointer that is unsafe, not the `glpk` object.

src/JuMP.jl
joaquimg: So we might need to highlight that the advanced user should have some knowledge of the Solver's MOI wrapper.

src/JuMP.jl
odow: They could, but that would require them to have JuMP as a dependency. Getting the backend through a dualize model seems tricky.

src/JuMP.jl
```diff
@@ -335,21 +335,100 @@ Base.broadcastable(model::Model) = Ref(model)
...
+
+See also: [`backend`](@ref).
+
+## Unsafe behavior
+
+This function is unsafe for two main reasons.
+
+First, the formulation and order of variables and constraints in the unsafe
+backend may be different to the variables and constraints in `model` due to
+bridges. There is no solution to this. Use the alternative suggested below
```
blegat, line 378: Even with no bridges, there is no guarantee the indices of the cache are the same as those of the solver.

blegat (approved):

Issue #1166, opened 2018-01-27
set constraint names from @constraint macro


mlubin:
https://github.com/JuliaOpt/JuMP.jl/blob/c9200d7a108924bca2a703fcf122318675313933/src/macros.jl#L326

blegat:
What should be the names for vectorized constraints ?
For
```julia
@constraint(model, con[i=1:2], x .== i)
```
the names are `con[1]` and `con[2]`.

For
```julia
@constraint(model, con, [x, y] .== [0.0, 0.0])
```
should the names be `con` and `con` or `con[1]` and `con[2]` ?

For
```julia
@constraint(model, con[i=1:2], [x, y] .== [i, i])
```
should the names be `con[1]`, `con[1]`, `con[2]` and `con[2]` or `con[1][1]`, `con[1][2]`, `con[2][1]` and `con[2][2]` ?

If we choose to use different names accross the broadcast, we will need to somehow get the index in the broadcasted function (or stop relying on broadcast but then many exotic cases might stop working).
If we choose to use the same name then we remove the possibility to query constraints per name when they are set with the broadcast syntax but that does not seem to be a very common use case.

mlubin:
I don't have a clear answer for how to name the broadcasted constraints. I think using the same name for all of them is a reasonable first pass that we can refine later on if needed.

Removed from the 0.19 milestone. It would be nice to have, but this would be new functionality relative to 0.18 anyway.

danielolsen:
I just found this thread and would like to reiterate my support for this feature. As a user, I would want a set of unique names, because I am running into issues trying to do exactly what you describe
> query constraints per name when they are set with the broadcast syntax
to get the shadow prices of a set of constraints which are created via matrix operations inside `@constraint`. My workaround is to manually set the names in a loop after the constraint is generated, which works, but it's not intuitive that this is necessary from the documentation at http://www.juliaopt.org/JuMP.jl/dev/constraints/#Vectorized-constraints-1.

For @blegat's example, I would prefer names like `con[1,1]`, `con[1,2]`, `con[2,1]`, and `con[2,2]`.

odow:
Related issue: https://github.com/jump-dev/JuMP.jl/issues/1936

PR #2574, opened 2021-04-22
Enhance extendibility of constraint macros

### LINKED ISSUE #2573: Allow Extra Positional Arguments for @constraint
Following the discussion [here](https://github.com/pulsipher/InfiniteOpt.jl/issues/60), it would be nice for  `@constraint` to accept additional positional arguments that would then be passed on to `build_constraint` to better enable JuMP extensions. As Oscar pointed out, this would allow us to "tag" constraints like what can currently be done with `@variable`. Moreover, this change should be none breaking.

I believe this would require modification to `_constraint_macro` to allow for the following 4 cases:
1. `@constraint(model, constr_expr)`
2. `@constraint(model, container_expr, constr_expr)`
3. `@constraint(model, constr_expr, extra_args...)`
4. `@constraint(model, container_expr, constr_expr, extra_args...)`

Here the complicating factor would probably be to distinguish between cases 2-4 (i.e., recognizing when `container_expr` is given or not).

pulsipher:
This restructures `_constraint_macro` to handle extra positional arguments (closes #2573) and enables the use of `base_name` in like manner to `@variable`. 

For supporting `base_name` the syntax is:
```julia
julia> using JuMP; model = Model(); @variable(model, x);

julia> @constraint(model, x == 0, base_name = "name")
name : x = 0

julia> @constraint(model, c1, x == 0)
c1 : x = 0

julia> @constraint(model, c2, x == 0, base_name = "name")
name : x = 0
```

For supporting extra arguments to be passed on to `build_constraint` we have:
```julia
julia> using JuMP; model = Model(); @variable(model, x);

julia> struct MyConstrType end

julia> function JuMP.build_constraint(
    _error::Function,
    f::JuMP.AffExpr,
    set::MOI.EqualTo,
    extra::Type{MyConstrType};
    d = 0,
)
    new_set = MOI.EqualTo(set.value + d)
    return JuMP.build_constraint(_error, f, new_set)
end

julia> @constraint(model, x == 0, MyConstrType, d = 2, base_name = "test")
test : x == 2.0
```

Update:
This also fixes a bug with constraint macro error messages. Previously, error messages always erroneously excluded keyword arguments in the error message, but this behavior has been corrected.
```julia
julia> using JuMP; model = Model(); @variable(model, x);

julia> @constraint(model, kwarg = 2) # previous behavior
ERROR: LoadError: At REPL[2]:1: `@constraint(model)`: Not enough positional arguments

julia> @constraint(model, kwarg = 2) # new behavior
ERROR: LoadError: At REPL[2]:1: `@constraint(model, kwarg = 2)`: Not enough positional arguments
```

~~Update 2:
This adds `parse_extra_constraint_args` as an extendable function to allow extra positional arguments to be parsed prior to adding them to the call expression of `build_constraint`. This allows extensions to define symbolic parsing of arguments when these arguments can be quite cumbersome to define otherwise (i.e., take advantage of the macro to enable symbolic extension syntax).~~

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -39,8 +39,8 @@ end
     _add_positional_args(call, args)::Nothing
 
 Add the positional arguments `args` to the function call expression `call`, 
-escaping each argument expression. The elements of `args` should should be ones 
-that were extracted via [`Containers._extract_kw_args`](@ref) and had appropriate 
+escaping each argument expression. The elements of `args` should be ones that 
+were extracted via [`Containers._extract_kw_args`](@ref) and had appropriate 
 arguments filtered out (e.g., the model argument). This is able to incorporate 
 additional positional arguments to `call`s that already have keyword arguments.
 
@@ -561,8 +561,8 @@ end
 
 Returns the code for the macro `@constraint_like args...` of syntax
 ```julia
-@constraint_like model con extra_args...     # single constraint
-@constraint_like model ref con extra_args... # group of constraints
+@constraint_like(model, con, extra_arg, kw_args...)      # single constraint
+@constraint_like(model, ref, con, extra_arg, kw_args...) # group of constraints
 ```
 where `@constraint_like` is either `@constraint` or `@SDconstraint`.
 
@@ -570,7 +570,7 @@ The expression `con` is parsed by `parsefun` which returns a `build_constraint`
 call code that, when executed, returns an `AbstractConstraint`. The macro
 keyword arguments (except the `container` keyword argument which is used to
 determine the container type) are added to the `build_constraint` call. The 
-`extra_args` are added as terminal positional arguments to the `build_constraint` 
+`extra_arg` is added as terminal positional argument to the `build_constraint` 
 call along with any keyword arguments (apart from `container` and `base_name`). 
 The returned value of this call is passed to `add_constraint` which returns a 
 constraint reference.
@@ -621,6 +621,11 @@ function _constraint_macro(
         c = gensym()
         x = y
         anonvar = true
+    end
+
+    # Enforce that only one extra positional argument can be given 
+    if length(extra) > 1
+        _error("Cannot specify more than 1 additional positional argument.")
     end
 
     # Prepare the keyword arguments 
@@ -762,12 +767,11 @@ that either `func` or `set` will be some custom type, rather than e.g. a
 set appearing in the constraint.
 
 For extensions that need to create constraints with more information than just 
-`func` and `set`, additional positional arguments can be specified to 
+`func` and `set`, an additional positional argument can be specified to 
 `@constraint` that will then be passed on `build_constraint`. Hence, we can 
 enable this syntax by defining extensions of 
-`build_constraint(_error, func, set, my_args...; kw_args...)` (using explicit 
-typed arguments not splatted ones as shown). This produces the user syntax: 
-`@constraint(model, ref[...], expr, my_args..., kw_args...)`. 
+`build_constraint(_error, func, set, my_arg; kw_args...)`. This produces the 
+user syntax: `@constraint(model, ref[...], expr, my_arg, kw_args...)`. 
 """
 macro constraint(args...)
     return _constraint_macro(
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -59,7 +59,7 @@ julia> call
 function _add_positional_args(call, args)
     kw_args = filter(arg -> isexpr(arg, :kw), call.args)
     filter!(arg -> !isexpr(arg, :kw), call.args)
-    for arg in args 
+    for arg in args
         push!(call.args, esc(arg))
     end
     append!(call.args, kw_args)
@@ -398,16 +398,18 @@ function parse_constraint(_error::Function, args...)
 end
 
 # Generic fallback.
-function build_constraint(
-    _error::Function,
-    func,
-    set::Union{MOI.AbstractScalarSet,MOI.AbstractVectorSet},
-    args...;
-    kwargs...,
-)
+function build_constraint(_error::Function, func, set, args...; kwargs...)
+    arg_str = join(args, ", ")
+    arg_str = isempty(arg_str) ? "" : ", " * arg_str
+    kwarg_str = join(Tuple(string(k, " = ", v) for (k, v) in kwargs), ", ")
+    kwarg_str = isempty(kwarg_str) ? "" : "; " * kwarg_str
     return _error(
-        "Unrecognized arguments: $(join(args, ", ")). \n\nIf you're trying " * 
-        "to create a JuMP extension, you need to implement `build_constraint`.",
+        "Unrecognized constraint building format. Tried to invoke " *
+        "`build_constraint(error, $(func), $(set)$(arg_str)$(kwarg_str))`, " *
+        "but no such method exists. This is due to specifying an unrecognized " *
+        "function, constraint set, and/or extra positional/keyword arguments." *
+        "\n\nIf you're trying to create a JuMP extension, you need to " *
+        "implement `build_constraint` to accomodate these arguments.",
     )
 end
 
@@ -534,14 +536,6 @@ function build_constraint(
     return build_constraint(_error, 1.0func, lb, ub)
 end
 
-function build_constraint(_error::Function, expr, lb, ub)
-    lb isa Number || _error(string("Expected $lb to be a number."))
-    ub isa Number || _error(string("Expected $ub to be a number."))
-    if lb isa Number && ub isa Number
-        _error("Range constraint is not supported for $expr.")
-    end
-end
-
 function build_constraint(
     ::Function,
     x::AbstractVector{<:AbstractJuMPScalar},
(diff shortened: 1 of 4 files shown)


test/macros.jl
```diff
@@ -176,6 +182,29 @@ function build_constraint_keyword_test(ModelType::Type{<:JuMP.AbstractModel})
...
+    d = false,
+)
+    return JuMP.build_constraint(_error, f, set)
+end
+function build_constraint_extra_arg_test(ModelType::Type{<:JuMP.AbstractModel})
+    @testset "build_constraint with extra positional arguments" begin
+        model = ModelType()
+        @variable(model, x)
+        cref = @constraint(model, x == 0, MyConstrType)
+        @test JuMP.constraint_object(cref).set isa MOI.EqualTo{Float64}
```
odow, line 201: I wonder if you could change the test to do something like modify the set so it isn't `MOI.EqualTo(0.0)`. This doesn't test that the overload worked, only that it didn't error.

odow (changes requested):
This was easier than I expected! 

See the format check: https://jump.dev/JuMP.jl/stable/developers/style/#JuliaFormatter

odow:
See the failure on 1.0. I guess method specificity changed?

test/macros.jl
pulsipher: Ok, I modified this test to be more involved to verify it is working as expected.

pulsipher:
The 1.0 failure was indeed due to a method specificity change with the new generic fallback `build_constraint`. The new commit as an improved generic fallback which is now trigger by this test and should provide a much more useful error message to users then what was provided previously. 

I also found a bug with the way constraint `_error`s were built in that they removed the keyword arguments from `args` since it was overwritten later on in the macro. The new commit, fixes this issue as well.

odow:
I understand why @github made it so you have to approve new contributors, but it is annoying that I have to approve CI to run on every commit...

pulsipher:
Ya, it doesn't seem to make a lot of sense having to approve every commit.

I think the docs are failing because of MOI v0.9.21.

test/macros.jl
```diff
@@ -176,6 +183,75 @@ function build_constraint_keyword_test(ModelType::Type{<:JuMP.AbstractModel})
...
+    if isexpr(arg, :*=)
+        return [Expr(:call, :MyInfo, arg.args...)]
+    else
+        return [arg]
+    end
+end
+function constraint_with_symbolic_extra_args(
+    ModelType::Type{<:JuMP.AbstractModel},
+)
+    @testset "build constraint with extra symbilic arguments" begin
```
odow, line 243:
```suggestion
    @testset "build constraint with extra symbolic arguments" begin
```

test/macros.jl
```diff
@@ -176,6 +183,75 @@ function build_constraint_keyword_test(ModelType::Type{<:JuMP.AbstractModel})
...
+    ::Val{:constraint},
+    arg,
+)
+    if isexpr(arg, :*=)
+        return [Expr(:call, :MyInfo, arg.args...)]
+    else
+        return [arg]
+    end
+end
+function constraint_with_symbolic_extra_args(
```
odow, line 240: You need to call this function to test it.

test/macros.jl
```diff
@@ -176,6 +183,75 @@ function build_constraint_keyword_test(ModelType::Type{<:JuMP.AbstractModel})
...
+function JuMP.build_constraint(
+    _error::Function,
+    func::AffExpr,
+    set::MOI.AbstractScalarSet,
+    info::MyInfo,
+)
+    func.terms[info.var] *= info.value
+    return JuMP.build_constraint(_error, func, set)
+end
+function JuMP.parse_extra_constraint_args(
```
odow, line 229: Yeah, this is type-piracy. Please remove.

test/macros.jl
```diff
@@ -176,6 +183,75 @@ function build_constraint_keyword_test(ModelType::Type{<:JuMP.AbstractModel})
...
+    _error::Function,
+    f::AffExpr,
+    set::MOI.EqualTo,
+    extra::Type{MyConstrType};
+    d = 0,
+)
+    new_set = MOI.LessThan{Float64}(set.value + d)
+    return JuMP.build_constraint(_error, f, new_set)
+end
+function build_constraint_extra_arg_test(ModelType::Type{<:JuMP.AbstractModel})
```
odow, line 198: You need to call this function to test it.

odow (commented):
I see why you want `parse_extra_constraint_args`, but we can't include it because it's type-piracy. It's going to cause issues if you load two JuMP extensions at the same time.

odow:
> I think the docs are failing because of MOI v0.9.21.

Ah. I'll take a look.

test/macros.jl
pulsipher: Done!

src/macros.jl
```diff
@@ -366,14 +397,94 @@ function parse_constraint(_error::Function, args...)
...
+julia> @constraint(model, c1, x >= 0, MyInfo(x, 2.3))
+c1 : 2.3 x >= 0.0
+```
+For more complex use cases, these arguments might become complicated and verbose 
+to define explicitly and it may be more convenient to provide a symbolic syntax. 
+This can be accomplished by extending `parse_extra_constraint_args`.
+```jldoctest constr_ext
+julia> using Base.Meta;
+
+julia> function JuMP.parse_extra_constraint_args(_error::Function, ::Val{:constraint}, arg)
```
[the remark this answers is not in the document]
pulsipher: Ok, I have removed it.

test/macros.jl
pulsipher: Done

odow (commented):
I fixed the docs, so you might just need to rebase.

We should also add docs for this: 
https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/developers/extensions.md#extend-constraint
The `MyConstrType` example you have is good.

odow (approved):
Looks good, modulo the documentation.

odow:
@blegat any chance you can take a look?

src/macros.jl
```diff
@@ -35,6 +35,37 @@ function _add_kw_args(call, kw_args)
     end
 end
 
+"""
+    _add_positional_args(call, args)::Nothing
+
+Add the positional arguments `args` to the function call expression `call`, 
+escaping each argument expression. The elements of `args` should should be ones
```
joaquimg, line 42: Should should

src/macros.jl
```diff
@@ -35,6 +35,37 @@ function _add_kw_args(call, kw_args)
...
+julia> call = :(f(1, a=2))
+:(f(1, a = 2))
+
+julia> JuMP._add_positional_args(call, [:(x)])
+
+julia> call
+:(f(1, $(Expr(:escape, :x)), a = 2))
+```
+"""
+function _add_positional_args(call, args)
```
joaquimg, line 59: Can we force a single positional argument for now?
Multiple ones can be very hard to deal with.

joaquimg (changes requested):

src/macros.jl
pulsipher: Wouldn't that be inconsistent with `@variable' which allows multiple extra arguments?

joaquimg: Yes it would be inconsistent.
But I dont think that was a food decision.
I think we should remove that behaviour from the variable macro.
Note that there, many positional args are no actually positional because the are used in any order, almost as keywords.
We should allow a single user defined positional arg. Otherwise the combinatorics of methods that need to be implemented will lead to unnecessary errors.
If some interface really needs two, they could create an arg that is the combination of two.

joaquimg: ref this https://github.com/jump-dev/JuMP.jl/pull/2421#issuecomment-774062945
and all the above comments.

joaquimg: Moreover, allowing more than one later is non-breaking.
While constraining to one later is breaking.

joaquimg:
Related issue to the names keyword: #1166

src/macros.jl
joaquimg: opened an issue: https://github.com/jump-dev/JuMP.jl/issues/2583

src/macros.jl
```diff
@@ -490,14 +536,6 @@ function build_constraint(
     return build_constraint(_error, 1.0func, lb, ub)
 end
 
-function build_constraint(_error::Function, expr, lb, ub)
```
blegat, line 493: Where did this go ? This wasn't used before or it's not used now due to a change you made ?

odow: It became `function build_constraint(_error::Function, func, set, args...; kwargs...)`

odow: It's not sufficient to expect numbers and ranges anymore.

src/macros.jl
pulsipher: Good catch, this is now fixed

src/macros.jl
pulsipher: That makes sense, I have updated everything accordingly. This is enforced in `_constraint_macro` and not `_add_positional_args` to provide better error messages and leave `_add_positional_args` as a general method.

src/macros.jl
pulsipher: Oscar is right, it was removed in favor of the new fallback since it becomes ambiguous when an extra argument is given.

joaquimg (approved):

src/macros.jl
blegat: Thanks for the explanation

blegat (approved):

PR #2576, opened 2021-04-26
Added ∈ for @variable set notation

pulsipher:
This simply adds support for `∈` with `parse_one_operator_variable` instead of only allowing `in`. Hence, we can now have:
```julia
julia> using JuMP; model = Model();

julia> @variable(model, [1:2] in SecondOrderCone())
 noname
 noname

julia> @variable(model, [1:2] ∈ SecondOrderCone())
 noname
 noname
```


blegat (approved):

odow (approved):
LGTM, assuming CI passes.

PR #2578, opened 2021-04-29
[docs] remove solver-specific notes

odow:
I removed the "solver-specific notes" section because the list of solvers is growing, and this section is getting long and could get out-dated. in addition, we discuss some solvers in detail, so why not all? Details (like the paragraph on ProxSDP, to pick one example) are best left to the solver's README.

One option I have been considering is to figure out a way to include each solver's README as a page in the JuMP documentation. Then we could standardize that format to give users a consistent documentation for each solver in an accessible place.

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -159,11 +159,6 @@ See a complete list [here](https://www.gams.com/latest/docs/S_MAIN.html).
     [free community license](https://www.gams.com/latest/docs/UG_License.html#GAMS_Community_Licenses)
     exists.
     
-## NEOS-based solvers
-
-Use [NEOSServer.jl](https://github.com/odow/NEOSServer.jl) to access solvers
-available through the [NEOS Server](https://neos-server.org).
-
 ## Previously supported solvers
 
 The following solvers were compatible with JuMP up to release 0.18 but are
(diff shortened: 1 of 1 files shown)


mlubin:
We should check that the information we're deleting is present in the solver READMEs.

docs/src/installation.md
```diff
@@ -126,97 +126,6 @@ Where:
...
-
-Requires a license. A trial version is available for small problem instances.
-
-* CDD
-
-CDD can solve the problem both using `Float64` and `Rational{BigInt}`
-arithmetics. The arithmetic used the type `T` given in `CDDLib.Optimizer{T}`.
-Only `CDDLib.Optimizer{Float64}` can be used with JuMP as JuMP inputs the
-problem in `Float64` arithmetics. Use [MOI](https://github.com/jump-dev/MathOptInterface.jl)
-directly for `CDDLib.Optimizer{Rational{BigInt}}`.
```
odow, line 145: Moved to https://github.com/JuliaPolyhedra/CDDLib.jl/pull/64

docs/src/installation.md
```diff
@@ -126,97 +126,6 @@ Where:
...
-
-Requires a license (free for academic use). The [Mosek interface](https://github.com/MOSEK/Mosek.jl)
-is maintained by the Mosek team. (Thanks!) Note that even if the package
-implementing MathOptInterface is `MosekTools`, for consistency the MOI optimizer
-is called `Mosek.Optimizer` so do the following to create a model with the Mosek
-solver:
-```julia
-using MosekTools
-model = Model(Mosek.Optimizer)
-```
```
odow, line 193: Moved to https://github.com/MOSEK/Mosek.jl/pull/204

PR #2584, opened 2021-05-03
Update the JuMP logo to match modern Julia

### LINKED ISSUE #2581: Logo thoughts
Now that Dark Mode is everywhere, many logos are suddenly confronting their dark side - or lack of it. I noticed that the JuMP logo doesn't work too well in all contexts.

In the sidebar of the Documenter docs, there is a provision for both light-mode and dark-mode logos. So it would be possible to have two versions of the logo, one light, one dark, and Documenter can switch between them automatically (the dark one has to be called `logo-dark.svg`.

Screenshot 2021-05-02 at 16 08 03

It occurred to me that a square-er version of the logo might work better in some situations:

Screenshot 2021-05-02 at 16 09 11

When used elsewhere, there's an alternative to having a white box around the entire logo - you can make a version that has a small white outline around each shape, so that the same image can work against any background:

Screenshot 2021-05-02 at 16 13 34

It mostly works...

If you want any of these, let me know - I can make a PR with a suitable SVG (or you can have the code).

odow:
This PR updates the JuMP logo to match the current Julia logo.

## Current

image

## This PR

image

## Dark mode

This PR also adds logo intended for dark mode, as well as some CSS magic to make it work with Documenter themes.

image

## Julia

For perspective, here is the current Julia logo

image

https://jump.dev/JuMP.jl/previews/PR2584/

Closes #2581


odow:
@cormullion here's what it looks like https://jump.dev/JuMP.jl/previews/PR2584

A few comments:
 * We changed from the flat logo back to the more Julian one with lighter inner circles:
image
 * We should remove the background colors
 * We should remove the word-mark from the square logo. Or at least, upload additional copies without.
 * The switching is working well however!

image

image

joaquimg:
I don't like the yellowish background in light mode. I would keep transparent.

Dark theme is ok.

cormullion:
Nice job with the CSS!

There's a `sidebar_sitename=false` keyword for `format = Documenter.HTML(...` which lets you hide the site name - "useful when the logo already contains the name of the package".

I think you might need to keep a version with a background for the main README.md (unless you can do your CSS magic there as well?).

PS: I think I prefer the JuMP logo with the same stronger colors as the newer Julia logo (https://github.com/JuliaLang/www.julialang.org/pull/372) - it makes the JuMP logo look like it's from an earlier time - but I understand you might prefer it!

odow:
> I think I prefer the JuMP logo with the same stronger colors as the newer Julia logo 

At one point we had your exact logo, and we even printed stickers. (See the logo hiding in the JuMP-dev 2018 poster: https://jump.dev/meetings/bordeaux2018/. I think we can claim to be the "first" to use it.) But we changed back to the 6-color version to stick with Julia's then logo. I'd be happy to switch back to the future, but this might be a BDFL decision @mlubin.

> I think you might need to keep a version with a background for the main README.md (unless you can do your CSS magic there as well?).

TIL GitHub has a dark mode. I don't think we can do a dark mode. Documenter let's us embed CSS, which I'm not sure how to do in a normal README.md. I'll have a play tomorrow, but feel free to make PRs to this branch in the mean time (I'm in NZ, so I think we're completely opposite timezones).

cormullion:
Yes, Github's dark mode has caused a few problems (https://github.com/JuliaLang/julia/pull/39601) - I know lots of people love it, since some also use a browser plugin, which causes even more problems (https://github.com/JuliaLang/www.julialang.org/issues/1241).

odow:
How's this now?

cormullion:
Looks great!

mlubin:
I hadn't noticed that Julia's logo changed. It's reasonable to change JuMP's logo to follow it. Are we sure we're using the exact same colors as Julia?

Could you update the PR title and description? Changing JuMP's logo is more of a headline than enabling dark mode.

odow:
Updated. I think the colors are the same. @cormullion picked them, and he is the expert.

We could make some minor changes. For example, the gap between the circles is a bit smaller in the new version. Maybe we could make the circles smaller? The Julia logo has a nicer balance.

cormullion:
Colors are as specified in Colors.jl - `Colors.JULIA_LOGO_COLORS`, so should be identical.

Here is with the circles slightly smaller - compared with the current Julia logo.

Screenshot 2021-05-15 at 11 44 47

I updated the SVG files with the change here https://github.com/jump-dev/JuMP.jl/pull/2596.

mlubin:
Does the preview reflect the new version with the smaller circles?

odow:
Not yet. Documentation has to finish.

odow:
Up now. I think it looks better.

cormullion:
Cool. 

Just in case you need this again, the basic code to create these logos is [here](https://gist.github.com/cormullion/42977b9bca4c7d08a6455865297c6d9e). I run SVGs through [svg0](https://github.com/svg/svgo) after making them.

blegat (approved):

cormullion:
Fyi - Readme.md doesn't show the logo yet...

PR #2585, opened 2021-05-03
changed/added logos for darkmode

cormullion:
Added wordmarks (logo + text)  that might work better in dark mode, and square logos that would work in Documenter's navigation sidebar.

see #2581 #2584


odow:
Thanks! I'll merge this into my PR so we can see the preview. It doesn't run for external contributors.

PR #2590, opened 2021-05-12
Add support for Containers in VectorConstraints

### LINKED ISSUE #2563: Support Containers in `VectorConstraints`
Support containers in `VectorConstraints`: https://discourse.julialang.org/t/adding-a-sos2-constraint-for-a-variable-which-is-defined-over-a-denseaxisarray/59637/2

This means doing something other than `collect` here:
https://github.com/jump-dev/JuMP.jl/blob/e2b6ee959c827d60f0c8178a813fddb6e21048b2/src/constraints.jl#L481-L486

Perhaps something like
```julia
function VectorConstraint( 
     func::AbstractVector{<:AbstractJuMPScalar}, 
     set::MOI.AbstractVectorSet, 
 ) 
     f = [func[idx] for idx in eachindex(func)]
     return VectorConstraint(f, set) 
 end 
```

c-jg:
Fixes #2563

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -483,7 +483,7 @@ function VectorConstraint(
     set::MOI.AbstractVectorSet,
 )
     f = [func[idx] for idx in eachindex(func)]
-    return VectorConstraint(f, set) 
+    return VectorConstraint(f, set)
 end
 
 reshape_set(set::MOI.AbstractVectorSet, ::VectorShape) = set
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -482,7 +482,8 @@ function VectorConstraint(
     func::AbstractVector{<:AbstractJuMPScalar},
     set::MOI.AbstractVectorSet,
 )
-    return VectorConstraint(collect(func), set)
+    f = [func[idx] for idx in eachindex(func)]
+    return VectorConstraint(f, set) 
 end
 
 reshape_set(set::MOI.AbstractVectorSet, ::VectorShape) = set
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -482,6 +482,7 @@ function VectorConstraint(
     func::AbstractVector{<:AbstractJuMPScalar},
     set::MOI.AbstractVectorSet,
 )
+    # collect() is not used here so that DenseAxisArray will work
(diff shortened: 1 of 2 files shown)


odow (commented):
Thanks for the contribution!

Two things:
1. Format failed https://github.com/jump-dev/JuMP.jl/runs/2564556568?check_suite_focus=true because you have trailing whitespace. You can fix it manually following the diff linked above, or follow https://jump.dev/JuMP.jl/stable/developers/style/#JuliaFormatter
2. Can we add a test? Something like (I didn't test. Might be typos, etc.)
```Julia
function test_Container_constraints(ModelType, ::Any)
    m = ModelType()
    S = ["a", "b"]
    @variable(m, x[S], Bin)
    cref = @constraint(m, x in SOS1())
    c = JuMP.constraint_object(cref)
    @test c.func == [x["a"], x["b"]]
end
``` 
near https://github.com/jump-dev/JuMP.jl/blob/80626541cc19005d1cb48958531d2ec71e8b7bf7/test/constraint.jl#L49

c-jg:
Thanks @odow.  I assume you meant `SOS2()` instead of `SOS1()` in the test right?

odow:
No, I meant `SOS1()`, but I see now that it isn't well documented: https://github.com/jump-dev/JuMP.jl/issues/2591

An SOS1 constraint here means that at most one of the two variables can be non-zero. An SOS2 constraint here means that at most two variables can be non-zero, and if so, they must be adjacent... which is always satisfied, so the constraint can be trivially removed.

It doesn't matter for the purpose of this test though, we can keep SOS2.

odow (approved):

blegat (approved):

src/constraints.jl
```diff
@@ -482,7 +482,8 @@ function VectorConstraint(
     func::AbstractVector{<:AbstractJuMPScalar},
     set::MOI.AbstractVectorSet,
 )
-    return VectorConstraint(collect(func), set)
+    f = [func[idx] for idx in eachindex(func)]
```
blegat, line 485: Maybe just add a comment above this line saying that we don't use `collect` so that it work for DenseAxisArray

Issue #2355, opened 2020-10-18
Certain base functions on JuMP constraints not implemented for NLconstraints.


1ozturkbe:
Hey folks, I'm struggling a bit since there are certain basic methods on JuMP constraints that are not defined with NLconstraints, eg.:

``` julia
delete(::JuMP.Model, ::NonlinearConstraintIndex)
is_valid(::JuMP.Model, ::ConstraintRef{Model, NonlinearConstraintIndex, ScalarShape})
```
simply do not work. I was wondering what the design behind this is, and whether there are any workarounds. The lack of these functions makes dealing with nonlinear models cumbersome, even as I try to convert the problems into more tractable forms. Thanks in advance!

odow:
You can read more about the current design of JuMP, and why nonlinear is different, in this paper: https://arxiv.org/abs/2002.03447. There are no workarounds. If you need to delete a nonlinear constraint, rebuild the model instead.

We won't be supporting `delete` before https://github.com/jump-dev/MathOptInterface.jl/issues/846 is implemented. The current timeline for this is 2-3 years, so don't wait for it.

`is_valid` is perhaps easier to implement, so I'll leave this issue open for it. Are there any others?

TODOs
- [ ] Implement `is_valid(::JuMP.Model, ::ConstraintRef{Model, NonlinearConstraintIndex, ScalarShape})`

1ozturkbe:
I guess this is not implemented for either ConstraintRefs or NonlinearConstraintRefs, but efficient ways to extract variables from constraints/NLconstraints would be great. Also, if you ask me, the following  MWE should not return an error...
``` julia
using JuMP, Gurobi
m = Model(Gurobi.Optimizer)
@variable(m, x)
ex = @NLexpression(m, x^2)
@constraint(m, x >=1)
@objective(m, Min, x)
optimize!(m)
```
but it does (saying ERROR: The solver does not support nonlinear problems...) although there are truly no nonlinear constraints in the model. So somehow the behavior of expressions is also different.

odow:
> I guess the issues is that there is no equivalent of @build_constrant for nonlinear expressions, where the expressions are not attached to the model.

Correct.

1ozturkbe:
[...]

PR #2592, opened 2021-05-12
added support for is_valid for nonlinear constraints

hdavid16:
as suggested by @odow in 
https://github.com/jump-dev/JuMP.jl/issues/2355#issuecomment-840068136

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -305,6 +305,22 @@ function is_valid(model::Model, con_ref::ConstraintRef{<:AbstractModel})
     )
 end
 
+"""
+    is_valid(model::Model,
+             c::ConstraintRef{Model,NonlinearConstraintIndex})
+
+Return `true` if `constraint_ref` refers to a valid nonlinear constraint in
+    `model`.
+"""
+function is_valid(model::Model,
+                  c::ConstraintRef{Model,NonlinearConstraintIndex})
+    if model !== c.model
+        return false
+    end
+    _init_NLP(model)
+    return 1 <= c.index.value <= num_nl_constraints(model)
+end
+
 #############################################################################
 # AbstractConstraint
 """
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -306,14 +306,15 @@ function is_valid(model::Model, con_ref::ConstraintRef{<:AbstractModel})
 end
 
 """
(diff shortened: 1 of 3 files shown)


src/constraints.jl
```diff
@@ -305,6 +305,22 @@ function is_valid(model::Model, con_ref::ConstraintRef{<:AbstractModel})
...
 
+"""
+    is_valid(model::Model,
+             c::ConstraintRef{Model,NonlinearConstraintIndex})
+
+Return `true` if `constraint_ref` refers to a valid nonlinear constraint in
+    `model`.
+"""
+function is_valid(model::Model,
+                  c::ConstraintRef{Model,NonlinearConstraintIndex})
```
odow, line 316: This is going to fail the format check. Wait for CI to fail and follow the diff, or use https://jump.dev/JuMP.jl/stable/developers/style/#JuliaFormatter.

The formatter doesn't format docstrings, but you should update it to match, and fix the indent on line 313.

odow (commented):
Can you add some tests please?

Somewhere near here:
https://github.com/jump-dev/JuMP.jl/blob/80626541cc19005d1cb48958531d2ec71e8b7bf7/test/nlp.jl#L1044-L1051

src/constraints.jl
```diff
@@ -305,6 +305,23 @@ function is_valid(model::Model, con_ref::ConstraintRef{<:AbstractModel})
...
+function is_valid(
+    model::Model,
+    c::ConstraintRef{Model,NonlinearConstraintIndex},
+)
+    if model !== c.model
+        return false
+    end
+    _init_NLP(model)
+    return 1 <= c.index.value <= num_nl_constraints(model)
+end
```
odow, line 323: Ah. I missed this on first glance. It needs to go in the `src/nlp.jl` file after `NonlinearConstraintIndex` is defined.

hdavid16: yeah, I'm moving it over there. Noticed it after precompilation errors came up

src/constraints.jl
hdavid16: done

hdavid16:
added tests at the end of the nlp.jl tests file

odow (approved):
Looks good, assuming tests pass.

PR #2595, opened 2021-05-14
Generalize build_variable for Variable(s)ConstrainedOnCreation

### LINKED ISSUE #2594: Enhance Usability of Variable(s)ConstrainedOnCreation for Extensions
For extensions like InfiniteOpt.jl that introduce a number of new scalar like variable types (e.g., `InfiniteVariable`, `PointVariable`) in addition to `JuMP.ScalarVariable`, enabling the use of variable sets built into JuMP (e.g., `MOI.AbstractVectorSet`, `JuMP.SymMatrixSpace`, `JuMP.SecondOrderCone`) is a little tedious since all of their respective `JuMP.build_variable` methods are typed to `JuMP.ScalarVariable` or `Vector{<:JuMP.ScalarVariable}`. Thus, extensions like InfiniteOpt.jl need to extend `JuMP.build_variable` for all these cases to work with the extension variable types. 

One idea to alleviate this problem would be to introduce a new abstract variable type to JuMP for scalar variables:
```julia
abstract type AbstractScalarVariable <: AbstractVariable end
```
Then let 
```julia
struct ScalarVariable{S,T,U,V} <: AbstractScalarVariable
    info::VariableInfo{S,T,U,V}
end
```
With this, all the the appropriate `JuMP.build_variable` methods could be typed with `JuMP.AbstractScalarVariable` or `Vector{<:JuMP.AbstractScalarVariable}`. Then extensions like InfiniteOpt.jl could inherit these by simply having the extended variable types inherit `JuMP.AbstractScalarVariable` as appropriate.

If this sounds agreeable, I would be happy to make a pull request.

pulsipher:
Closes #2594.

### DIFF
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -93,7 +93,11 @@ end
         return NewVariable(info)
     end
     function JuMP.add_variable(model::Model, v::NewVariable, name::String = "")
-        return JuMP.add_variable(model, ScalarVariable(v.info), name)
+        return JuMP.add_variable(
+            model,
+            ScalarVariable(v.info),
+            name * "_normal_add",
+        )
     end
     function JuMP.add_variable(
         model::Model,
@@ -106,6 +110,7 @@ end
     )
         vs = map(i -> ScalarVariable(i.info), v.scalar_variables)
         new_v = VariablesConstrainedOnCreation(vs, v.set, v.shape)
+        names .*= "_constr_add"
         return JuMP.add_variable(model, new_v, names)
     end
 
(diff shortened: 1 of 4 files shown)


blegat (approved):

test/macros.jl
```diff
@@ -80,6 +80,47 @@ end
...
+    model = Model()
+    @variable(model, 0 <= x <= 1, NewVariable, Bin)
+    @test lower_bound(x) == 0
+    @test upper_bound(x) == 1
+    @test is_binary(x)
+    @test name(x) == "x"
+
+    @variable(model, y[1:3] in SecondOrderCone(), NewVariable)
+    @test name.(y) == ["y[$i]" for i in 1:3]
+    @test num_constraints(model, Vector{VariableRef}, MOI.SecondOrderCone) == 1
```
odow, line 121: This test doesn't test that `NewVariable` actually gets used.

If you did something like:
```Julia
function JuMP.add_variable(model::Model, v::NewVariable, name::String = "")
    return JuMP.add_variable(model, ScalarVariable(v.info), "NewVariable" * name)
end
```
then you could check the code-path actually got hit by looking at the variable names.

pulsipher: I have updated the tests accordingly.

odow (approved):

PR #2601, opened 2021-05-17
Clarify that delete does not call unregister

### LINKED ISSUE #2589: Document that delete does not call unregister
Found on [Discourse](https://discourse.julialang.org/t/disassociate-parameter-or-variable-from-a-model-in-jump/60851/5) by @path-doc 
https://github.com/jump-dev/JuMP.jl/blob/4f6abf80016fc5e07e4e85da84ae2f93396fa4a2/src/variables.jl#L229-L244

```Julia
julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x)
x

julia> model
A JuMP Model
Feasibility problem with:
Variable: 1
Model mode: AUTOMATIC
Names registered in the model: x

julia> delete(model, x)

julia> model[:x]
noname

julia> model
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Names registered in the model: x
```

odow:
Closes #2589

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -801,13 +801,12 @@ julia> normalized_coefficient(con, x[2])
     [`set_normalized_coefficient`](@ref) sets the coefficient of the
     normalized constraint.
 
-## Deleting constraints
+## Delete a constraint
 
-Constraints can be deleted from a model using [`delete`](@ref).
+Use [`delete`](@ref) to delete a constraint from a model. Use [`is_valid`](@ref)
+to check if a constraint belongs to a model and has not been deleted.
 
-Check if a constraint reference is valid using [`is_valid`](@ref).
-
-```jldoctest; setup = :(model=Model(); @variable(model, x))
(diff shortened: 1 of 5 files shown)


docs/src/manual/constraints.md
```diff
@@ -820,6 +819,38 @@ julia> is_valid(model, con)
...
+    `x = @variable(model, [1:N], ...)` where the name of the object does
+    not appear inside the macro.
+
+    Alternatively, use `unregister(model, :con)` to first unregister
+    the existing name from the model. Note that this will not delete the
+    object; it will just remove the reference at `model[:con]`.
+[...]
+```
+
+After calling [`delete`](@ref), call [`unregister`](@ref) to remove the symbolic
```
blegat, line 837: "After" -> "In addition to". It does not need to be done in a particular order

odow: It doesn't need the order, but I'd prefer it like this. `delete` is the main thing they're trying to do. `unregister` is an optional extra in some cases.

PR #2605, opened 2021-05-19
Example: how to list conflicting constraints

jd-foster:
This adds an example to the notes on how to get all conflicting constraints in an arbitrary way from a model.


docs/src/manual/solutions.md
```diff
@@ -450,6 +451,21 @@ MOI.get(model, MOI.ConstraintConflictStatus(), c2)
 new_model, reference_map = copy_conflict(model)
 ```
 
+Conflicting constraints can be collected in a list and printed 
+as follows:
+
+```julia
+Conflict_Constraint_List = []
```
odow, line 458:
```suggestion
conflict_constraint_list = ConstraintRef[]
```

odow:
@blegat did something change in the last couple of days???
https://github.com/jump-dev/JuMP.jl/pull/2605/checks?check_run_id=2624301638#step:5:622

@jd-foster the documentation failure doesn't look like it was caused by this PR.

odow:
Ah, it was this: https://github.com/JuliaLang/Compat.jl/pull/729

jd-foster:
I was messing with git commit messages; looks like a force-push updated this branch, and now requires re-doing actions

odow:
No problem. In future there's also no need to squash commits. I can do that from the web interface when merging :)

odow (approved):

jd-foster:
Ah, thanks. Good to know you can do it on your side in the future.

PR #2614, opened 2021-05-25
OptimizerWithAttributes direct model constructor

jd-lara:
Adds a way to create a direct model when the user passes an `OptimizerWithAttributes`. The rationale is to reuse solver definitions for different model options.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -335,17 +335,9 @@ end
 Return a new JuMP model that instantiates a [`OptimizerWithAttributes`](@ref) to store the
 model and solve it.
 """
-function direct_model(optimizer::MOI.OptimizerWithAttributes)
-    return Model(
-        MOI._instantiate_and_check(optimizer),
-        Dict{_MOICON,AbstractShape}(),
-        Set{Any}(),
-        nothing,
-        nothing,
-        Dict{Symbol,Any}(),
-        0,
-        Dict{Symbol,Any}(),
-    )
+function direct_model(factory::MOI.OptimizerWithAttributes)
+    optimizer = MOI.instantiate(factory)
+    return direct_model(optimizer)
 end
 
 Base.broadcastable(model::Model) = Ref(model)
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -348,8 +348,6 @@ function direct_model(optimizer::MOI.OptimizerWithAttributes)
     )
 end
 
-
-
 Base.broadcastable(model::Model) = Ref(model)
 
 """
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -330,10 +330,28 @@ function direct_model(backend::MOI.ModelLike)
 end
 
 """
-    direct_model(optimizer::MOI.OptimizerWithAttributes)
+    direct_model(factory::MOI.OptimizerWithAttributes)
 
-Return a new JuMP model that instantiates a [`OptimizerWithAttributes`](@ref) to store the
-model and solve it.
+Create a [`direct_model`](@ref) using `factory`, a `MOI.OptimizerWithAttributes`
+object created by [`optimizer_with_attributes`](@ref).
+
+## Example
+
+```julia
+model = direct_model(
+    optimizer_with_attributes(
+        Gurobi.Optimizer,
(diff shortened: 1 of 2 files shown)


src/JuMP.jl
```diff
@@ -329,6 +329,25 @@ function direct_model(backend::MOI.ModelLike)
     )
 end
 
+"""
+    direct_model(optimizer::MOI.OptimizerWithAttributes)
+
+Return a new JuMP model that instantiates a [`OptimizerWithAttributes`](@ref) to store the
+model and solve it.
+"""
+function direct_model(optimizer::MOI.OptimizerWithAttributes)
```
odow, line 338: Why not just 
```julia
function direct_model(factory::MOI.OptimizerWithAttributes)
    optimizer = MOI.instantiate(factory)
    return direct_model(optimizer)
end
```
This also needs a test.

jd-lara: I thought it would be unnecessary to call `MOI.instantiate` since it for direct mode bridges and the incremental interface isn't compatible. I can make the changes as suggested also.

odow: https://github.com/jump-dev/MathOptInterface.jl/blob/30506e3757e9d2d6c21f3cdafc207c9a2ae947b9/src/instantiate.jl#L117-L125

No bridges are added if you omit the `with_bridge_type` argument.

jd-lara: fixed in 764bed3, test added in 0110cac

test/model.jl
```diff
@@ -609,6 +609,20 @@ function test_copy_direct_mode()
...
+    function fake_optimizer()
+        return MOIU.MockOptimizer(
+            MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
+        )
+    end
+
+    optimizer = optimizer_with_attributes(fake_optimizer, "a" => 1, "b" => 2)
+    model = JuMP.direct_model(optimizer)
+    @test model.moi_backend isa MOIU.MockOptimizer
+    @test MOI.get(model.moi_backend, MOI.RawOptimizerAttribute("a")) == 1
```
odow, line 622:
```suggestion
    @test MOI.get(model.moi_backend, MOI.RawParameter("a")) == 1
```

test/model.jl
```diff
@@ -609,6 +609,20 @@ function test_copy_direct_mode()
...
+    @test MOI.get(model.moi_backend, MOI.RawOptimizerAttribute("b")) == 2
```
odow, line 623:
```suggestion
    @test MOI.get(model.moi_backend, MOI.RawParameter("b")) == 2
```

src/JuMP.jl
```diff
@@ -329,6 +329,17 @@ function direct_model(backend::MOI.ModelLike)
     )
 end
 
+"""
+    direct_model(optimizer::MOI.OptimizerWithAttributes)
+
+Return a new JuMP model that instantiates a [`OptimizerWithAttributes`](@ref) to store the
```
odow, line 335: Mention [`optimizer_with_attributes`](@ref) instead of [`OptimizerWithAttributes`](@ref).

Something like this?
````Julia
"""
    direct_model(factory::MOI.OptimizerWithAttributes)

Create a [`direct_model`](@ref) using `factory`, a `MOI.OptimizerWithAttributes`
object created by [`optimizer_with_attributes`](@ref).

## Example

```julia
model = direct_model(
    optimizer_with_attributes(
        Gurobi.Optimizer, 
        "Presolve" => 0, 
        "OutputFlag" => 1,
    )
)
```
is equivalent to:
```julia
set_optimizer_attribute(model, "Presolve", 0)
set_optimizer_attribute(model, "OutputFlag", 1)
```
"""
````

jd-lara: fixed via 1b8a6b1

test/model.jl
jd-lara: ha, it took me a second to realize the change in MOI version fixed in 82058d8

test/model.jl
jd-lara: fixed in 82058d8

odow (approved):

blegat (approved):

PR #2618, opened 2021-05-31
[breaking] Ignore infinite variable bounds

odow:
Introducing non-finite variable bounds is just a cause of trouble: https://github.com/jump-dev/MathOptInterface.jl/issues/1270

Setting a bound to `-Inf` or `Inf` after creation is just a sign that the user wants to delete the bound. So we should make them do it explicitly.

This is breaking because people may be relying on `has_lower_bound` returning `true` if the bound is `-Inf`.

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1270

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -121,15 +121,15 @@ struct VariableInfo{S,T,U,V}
         binary::Bool,
         integer::Bool,
     ) where {S,T,U,V}
-        if has_lb && lower_bound === -Inf
+        if has_lb && !isfinite(lower_bound)
             has_lb = false
             lower_bound = NaN
         end
-        if has_ub && upper_bound === Inf
+        if has_ub && !isfinite(upper_bound)
             has_ub = false
             upper_bound = NaN
         end
-        if has_fix && (fixed_value === -Inf || fixed_value === Inf)
+        if has_fix && !isfinite(fixed_value)
             error("Unable to fix variable to $(fixed_value)")
         end
         return new{S,T,U,V}(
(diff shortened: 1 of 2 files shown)


src/variables.jl
```diff
@@ -109,6 +109,42 @@ struct VariableInfo{S,T,U,V}
...
+        has_ub::Bool,
+        upper_bound::T,
+        has_fix::Bool,
+        fixed_value::U,
+        has_start::Bool,
+        start::V,
+        binary::Bool,
+        integer::Bool,
+    ) where {S,T,U,V}
+        if has_lb && lower_bound === -Inf
```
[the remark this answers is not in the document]
odow: Swapped for `isfinite`, which also checks for +/- inf and NaNs.

odow:
Bump. @blegat how's this?

blegat (approved):
Since JuMP uses MOI.LessThan and MOI.GreaterThan (instead of MOI.Interval), it indeed does should not accept infinite bounds. We should wait before v0.22 is approaching before merging though.

src/macros.jl
```diff
@@ -766,12 +766,12 @@ that either `func` or `set` will be some custom type, rather than e.g. a
 `Symbol`, since we will likely want to dispatch on the type of the function or
 set appearing in the constraint.
 
-For extensions that need to create constraints with more information than just 
-`func` and `set`, an additional positional argument can be specified to 
-`@constraint` that will then be passed on `build_constraint`. Hence, we can 
-enable this syntax by defining extensions of 
-`build_constraint(_error, func, set, my_arg; kw_args...)`. This produces the 
-user syntax: `@constraint(model, ref[...], expr, my_arg, kw_args...)`. 
+For extensions that need to create constraints with more information than just
```
blegat, line 769: Do this in a separate PR, this does not need to wait until v0.22

odow: oops. I think this just got caught up in the formatting.

Issue #2593, opened 2021-05-13
Pluralisation of @NLparameters


path-doc:
Hi following on from the discourse thread:
https://discourse.julialang.org/t/pluralisation-of-nlparameter/61060

I'd like to request pluralisation of `@NLparameter`.

Thanks!

odow:
It may be as simple as adding to this list:
https://github.com/jump-dev/JuMP.jl/blob/80626541cc19005d1cb48958531d2ec71e8b7bf7/src/macros.jl#L948-L956

Then it just requires docs and tests.

Pramodh-G:
Hi! I'm a beginner to the JuMP ecosystem (Julia in general too), would it be good if I could work on this with your help?

odow:
Sure. The steps are
- Develop the package `] dev JuMP`
- Go to `~/.Julia/dev/JuMP`
- Make a change to `src/macros.jl` by adding `(:NLparameters, Symbol("@NLparameter")),` to the list
- Open Julia and test that it works.
- Make a PR
- We also need to add some tests and documentation
  -  For the docs, copy https://github.com/jump-dev/JuMP.jl/blob/b3687f39f34420d8246318cb42a95a782b327f33/src/macros.jl#L1114-L1131 and add it to the API Reference https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/reference/nlp.md
  - For the tests, add a new test set here: https://github.com/jump-dev/JuMP.jl/blob/b3687f39f34420d8246318cb42a95a782b327f33/test/nlp.jl#L708-L712

Let me know if you have questions on specific steps.

path-doc:
I tried this, but I get the following error (using same format as for @variables, but the error is the same if I remover the base_name part):

julia> using JuMP
[ Info: Precompiling JuMP [4076af6c-e467-56ae-b986-b466b2749572]
julia> using Ipopt
julia> m = Model(Ipopt.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: EMPTY_OPTIMIZER
Solver name: Ipopt

julia> @NLparameter(m, a1 == 2)
"Reference to nonlinear parameter #1"

julia> @NLparameters(m,
                   begin
                   a == 1,
                   b == 2
                   end)
ERROR: LoadError: At REPL[29]:3: `@NLparameter(m, a == 1, b == 2)`: Too many arguments.
Stacktrace:
[...]

PR #2619, opened 2021-05-31
Add Pluralization of @NLparameters (#2593)

Pramodh-G:
Fixes Issue #2593 . 
Please suggest ways in which this PR can be improved :)

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1093,14 +1093,15 @@ end)
 ```
 """ :(@SDconstraints)
 
- @doc """
+@doc """
      @NLparameters(model, args...)
 
  Create and return multiple nonlinear parameters attached to model `model`, in the same fashion as
  [`@NLparameter`](@ref) macro.
 
- The model must be the first argument, and multiple variables can be added on
- multiple lines wrapped in a `begin ... end` block.
+ The model must be the first argument, and multiple parameters can be added on
+ multiple lines wrapped in a `begin ... end` block. Distinct parameters need to be placed
+ on separate lines without `base_names` as in the following example.
 
 # Example
 ```jldoctest; setup=:(using JuMP)
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1101,7 +1101,7 @@ end)
 
  The model must be the first argument, and multiple parameters can be added on
  multiple lines wrapped in a `begin ... end` block. Distinct parameters need to be placed
- on separate lines without `base_names` as in the following example.
+ on separate lines as in the following example.
 
 # Example
 ```jldoctest; setup=:(using JuMP)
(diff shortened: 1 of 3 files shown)


path-doc (commented):
May I suggest that you make clear the syntax for begin ... end blocks:

"Distinct parameters need to be placed on seperate lines without base_names as in the following example."

This is because other objects (such as variables and constraints) can be assigned names.

Pramodh-G:
Sure, I'll update it.

src/macros.jl
```diff
@@ -1092,6 +1093,30 @@ end)
...
 
+@doc """
+     @NLparameters(model, args...)
+
+ Create and return multiple nonlinear parameters attached to model `model`, in the same fashion as
+ [`@NLparameter`](@ref) macro.
+
+ The model must be the first argument, and multiple parameters can be added on
+ multiple lines wrapped in a `begin ... end` block. Distinct parameters need to be placed
+ on separate lines without `base_names` as in the following example.
```
blegat, line 1104: > without `base_names`

This is not clear what you mean by that, I would just remove it

odow, line 1104:
```suggestion
 on separate lines as in the following example.
```

odow (approved):

path-doc:
What was meant is that somewhere it should be clear that, unlike variables
and constraints, parameters cannot be given a ```name::String```.

odow:
> What was meant is that somewhere it should be clear that, unlike variables and constraints, parameters cannot be given a ```name::String```.
> […](#)

This is covered in the docstring of `@NLparameter`: https://jump.dev/JuMP.jl/v0.21.8/reference/nlp/#ref_nl_parameters. Adding extra arguments is a separate issue.

path-doc:
We wouldn't want any redundancy now would we :)

odow:
We link to the `@NLparameter` docstring and explain it follows the same syntax. Moreover, the general rule for docstrings is to explain the things that can be done,  rather than what can't be done.

Error messages should tell what people did wrong (this isn't the greatest example because it actually looks like a bug!:
```Julia
julia> @NLparameter(model, p == 1, base_name = "x")
ERROR: LoadError: At REPL[7]:1: `@NLparameter(model, p == 1)`: Too many arguments.
Stacktrace:
 [1] error(::String, ::String)
   @ Base ./error.jl:42
 [2] _macro_error(macroname::Symbol, args::Tuple{Symbol, Expr}, source::LineNumberNode, str::String)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:1365
 [3] (::JuMP.var"#_error#122"{LineNumberNode, Symbol, Expr})(str::String)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:2114
 [4] var"@NLparameter"(__source__::LineNumberNode, __module__::Module, m::Any, ex::Any, extra::Vararg{Any, N} where N)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:2118
in expression starting at REPL[7]:1
```

PR #2629, opened 2021-06-16
DNMY: Tests for performance improvements

### LINKED ISSUE #2628: Affine expressions are slow
@jd-lara has a problem where creating affine expressions is a bottleneck.

There are two main causes.

This double-hash in `_add_or_set!`
https://github.com/jump-dev/JuMP.jl/blob/2ff880d46161a8502d9ca378d098d60d1c9864c3/src/aff_expr.jl#L18-L28

This size hint:
https://github.com/jump-dev/JuMP.jl/blob/2ff880d46161a8502d9ca378d098d60d1c9864c3/src/aff_expr.jl#L390-L401
In particular, the size hint is pessimistic in that it suggests reserving enough space for the union of the elements assuming that there are no duplicates.

odow:
Opening this PR so @jd-lara has a target to benchmark against.

Closes #2628

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -15,15 +15,30 @@
 # Operator overloads in src/operators.jl
 #############################################################################
 
-# Utilities for OrderedDict
+import DataStructures
+
 function _add_or_set!(dict::OrderedDict{K,V}, k::K, v::V) where {K,V}
     # Adding zero terms to this dictionary leads to unacceptable performance
     # degradations. See, e.g., https://github.com/jump-dev/JuMP.jl/issues/1946.
     if iszero(v)
-        return dict  # No-op.
+        return
     end
-    # TODO: This unnecessarily requires two lookups for k.
-    dict[k] = get!(dict, k, zero(V)) + v
+    # A naive implementation of the following is
+    #     dict[k] = get!(dict, k, zero(V)) + v
+    # The problem with these is that there are two key lookups (one for
+    # `setindex!` and one to get the current element).
+    #
+    # The following is an optimization to reduce the cost to one lookup if the
+    # key exists. Using one lookup when the key doesn't exist requires some
+    # calls to `_setindex!` in OrderedCollections, which may be a step too far.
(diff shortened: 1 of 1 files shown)


jd-lara:
Some results of building the model: The Line Section is the one we are trying to improve. This is the result of a "sparse" expression array where there are 0's. If the ExpressionArray is filled with DoubleExpressions the problem never builds. 

**Julia tagged version***

image


**This Branch**

image

jd-lara:
@odow one of the things to note here is that when we do `constant*ParameterJuMP.DoubleGenericAffExpr{Float64, VariableRef, ParameterRef}` 

This goes into this function defined in ParameterJumP.jl

```julia
Base.:(*)(lhs::Number, rhs::PAE{C}) where C<:Number = PAE{C}(lhs*rhs.v, lhs*rhs.p)
```

Which in turn calls this code twice for the same lhs value. 

```julia
In *(lhs, rhs) at /Users/jdlara/.julia/packages/JuMP/7JcAM/src/operators.jl:50
 50  function Base.:*(lhs::_Constant, rhs::_GenericAffOrQuadExpr)
>51      if iszero(lhs)
 52          return zero(rhs)
 53      else
 54          α = _constant_to_number(lhs)
 55          return map_coefficients(c -> α * c, rhs)
```

odow:
Only if called outside the macro. JuMP rewrites `ret += lhs * rhs` to `add_to_expression!(ret, lhs, rhs)`

jd-lara:
I am running debugger, so this is what is being called in the @expression macro

odow:
Closing this for now: https://github.com/JuliaStochOpt/ParameterJuMP.jl/issues/85#issuecomment-862908846

I finally tracked down the likely culprit: Julia allocates `ScalarConstraint(::AffExpr,::EqualTo)` to the stack, but `ScalarConstraint(::ParameterJuMP.DoubleGenericAffExpr{Float64, VariableRef, ParameterRef}, ::EqualTo)` to the heap.

Cue more allocations and GC time. These allocations aren't coming from a mutable arithmetic failure.

PR #2630, opened 2021-06-17
Update printing of SparseAxisArray

### LINKED ISSUE #2624: stable printing for SparseAxisArray.jl
It might be better to show the array in a stable form via `sort`. This shouldn't make any performance difference, but look better and solve https://github.com/jump-dev/JuMP.jl/issues/2623.
https://github.com/jump-dev/JuMP.jl/blob/cde8bb54cf4f70841ca6c4a59ed6f4c848241ff0/src/Containers/SparseAxisArray.jl#L296
==>
```julia
 for (i, key) in enumerate(sort!(collect(keys(x.data))))
   value = x.data[key]
```

odow:
Closes #2624 

This PR simplifies the printing, makes it stable across Julia versions (now ordered alphabetically), and adds tests for the case where some rows are omitted.

### DIFF
--- a/src/Containers/SparseAxisArray.jl
+++ b/src/Containers/SparseAxisArray.jl
@@ -280,9 +280,9 @@ function Base.show(io::IOContext, x::SparseAxisArray)
     if isempty(x)
         return show(io, MIME("text/plain"), x)
     end
-    limit::Bool = get(io, :limit, false)
+    limit = get(io, :limit, false)::Bool
     half_screen_rows = limit ? div(displaysize(io)[1] - 8, 2) : typemax(Int)
-    print_entry(i) = i < half_screen_rows || i > length(x) - half_screen_row
+    print_entry(i) = i < half_screen_rows || i > length(x) - half_screen_rows
     # For stable printing, sort by the string value of the key.
     key_strings = [(join(key, ", "), value) for (key, value) in x.data]
     sort!(key_strings; by = x -> x[1])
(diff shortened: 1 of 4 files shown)


src/Containers/SparseAxisArray.jl
```diff
@@ -275,27 +275,26 @@ function Base.show(io::IO, ::MIME"text/plain", sa::SparseAxisArray)
...
         return show(io, MIME("text/plain"), x)
     end
-    limit::Bool = get(io, :limit, false)
+    limit = get(io, :limit, false)::Bool
     half_screen_rows = limit ? div(displaysize(io)[1] - 8, 2) : typemax(Int)
-    key_string(key::Tuple) = join(key, ", ")
     print_entry(i) = i < half_screen_rows || i > length(x) - half_screen_rows
+    # For stable printing, sort by the string value of the key.
+    key_strings = [(join(key, ", "), value) for (key, value) in x.data]
+    sort!(key_strings; by = x -> x[1])
```
blegat, line 288: What  if the key does not implement `isless` ? Probably using an `OrderedDict` is easier

odow: It does it on the string representation of the key, which we were already producing. This seemed much easier than an OrderedDict.

blegat: Indeed, it should work then. I'm just thinking that if we are using an ordered dict for affine expression to make sure things are always easily reproducible then we might as well do it for `SparseAxisArray`. Or we need to convince ourself that it will never affect any behaviors of JuMP.

odow: That requires a bigger change to how SparseAxisArray is constructed. This seemed like an easy win in the near-term.

blegat: Ok, let's keep that in mind

blegat (approved):

PR #2631, opened 2021-06-19
Apply the OutdatedWarning from DocumenterTools

odow:
<s>I'm hoping that the preview link should work.</s> Preview doesn't work because it's not on `master`. Because this is a bit of a mess at the moment.

I followed: https://discourse.julialang.org/t/ann-documenter-0-27/62726

@mortenpi is it meant to re-write such a large chunk of existing docs? I thought it just added the banner?

Here is what I see locally:

image


mortenpi:
I think it's because it uses an XML parser to read-modify-write the HTML and this leads to some formatting changes. It think it's all good if it all seems to work locally, but maybe @pfitzseb could also take a quick look.

odow:
Yeah I had a look around locally and everything seemed okay. It'd be hard to spot any mistakes though.

odow:
I'll merge for now. We can take a look once it's merged and revert this or make any necessary changes.

pfitzseb:
Yeah, getting proper diffs is hard/impossible. The good news is that we're using Gumbo.jl for parsing/writing, which should be guaranteed not to break anything.

odow:
I did break this https://github.com/jump-dev/JuMP.jl/pull/2633.

I guess because the most recent `stable` wasn't built with Documenter 0.27?

pfitzseb:
Kind of. You wouldn't get the warning if `stable` had been build with 0.27, yes. 
But when using `OutdatedWarning.generate` you need to figure out which changes to check in yourself -- that warning is completely unconditional.

odow:
> you need to figure out which changes to check in yourself

Yeah I missed this the first time around. (Although it was kinda obvious in retrospect.)

pfitzseb:
Added a warning to the docs: https://github.com/JuliaDocs/DocumenterTools.jl/pull/55

PR #2646, opened 2021-07-13
[docs] clarify when to use Symbols and Strings

odow:
I have people ask me this frequently, and it came up on Discourse here: https://discourse.julialang.org/t/passing-arguments-to-julia-functions/64335/3?u=odow

@rdeits I hope you don't mind me appropriating some of your wording. I thought it was well put.

### DIFF
--- a/docs/src/tutorials/Getting started/an_introduction_to_julia.jl
+++ b/docs/src/tutorials/Getting started/an_introduction_to_julia.jl
@@ -245,13 +245,19 @@ println("The value of x is: $(x)")
 
 # ### Symbols
 
-# Julia `Symbol`s provide a way to make human readable unique identifiers:
+# Julia `Symbol`s are a data structure from the compiler that represent Julia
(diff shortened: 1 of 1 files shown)


mlubin:
This seems to be taking a step back from the JuMP style guide (https://jump.dev/JuMP.jl/v0.21.8/developers/style/#@enum-vs.-Symbol): "Use of Symbol should typically be reserved for identifiers, e.g., for lookup in the JuMP model (model[:my_variable])."

odow:
The issue is that people coming from python etc have no idea what a symbol is, and saying "it's like an interned string" isn't helpful, neither are "identifiers".

Suggestions for a more beginner friendly explanation?

mlubin:
> Suggestions for a more beginner friendly explanation?

A data structure from the compiler that stores julia identifiers (i.e., variable names). It's often (ab)used to stand in for a string or an enum, when one of the former is likely a better choice. :)

PR #2650, opened 2021-07-26
[docs] add example of how to evaluate objective function at a point

jd-foster:
Adds in example as given here: https://discourse.julialang.org/t/how-to-evaluate-a-jump-model-given-a-candidate-solution/64555

Also adds missing docstring for `value` with change of function name consistent with other methods.


docs/src/manual/objective.md
```diff
@@ -61,6 +61,25 @@ julia> objective_function(model)
 2 x + 1
 ```
 
+## Evaluate the objective function at a point
+
+Use [`value`]($ref) to evaluate an objective function at a point specifying values for variables.
```
odow, line 66:
```suggestion
Use [`value`](@ref) to evaluate an objective function at a point specifying values for variables.
```

odow (approved):
LGTM, assuming that the doc failure was just the typo.

odow:
Hmm there must be something wrong with our Documenter setup for external contributors: https://github.com/jump-dev/JuMP.jl/pull/2652. I don't think the error is your fault. I'll look into it.

odow:
@mortenpi have you seen failures like this? the docs build, but when an outside contributor makes a PR it mistakenly tries to deploy to the push-preview.

image

mortenpi:
I thought we fixed it (https://github.com/JuliaDocs/Documenter.jl/pull/1567). Do you have Documenter 0.27? Otherwise, there is also [this related issue](https://github.com/JuliaDocs/Documenter.jl/issues/1572), but I don't think it's relevant here.

odow:
Yeah it's using 0.27.5

image

mortenpi:
Ah, it looks like the `curl` call fails somehow: https://github.com/jump-dev/JuMP.jl/pull/2650/checks?check_run_id=3235957904#step:5:971

So it behaves as intended: it can't detect that it _isn't_ the right repository, and so just attempts to push. As a curiosity, just to figure out how it failes, could you run it [with the `mp/preview-check-test` Documenter branch](https://github.com/JuliaDocs/Documenter.jl/commit/7378621142689b6292496c429b37be9d5c5934c1)? It adds some debug output.

mortenpi:
Ok, it looks like I'm relying on behaviour that was introduced after 1.0:

```
julia> run(pipeline(`ls -Alh`, stdout = IOBuffer()))
ERROR: MethodError: no method matching rawhandle(::Base.GenericIOBuffer{Array{UInt8,1}})
Closest candidates are:
  rawhandle(::RawFD) at process.jl:149
  rawhandle(::Base.DevNull) at process.jl:148
  rawhandle(::Base.Filesystem.File) at filesystem.jl:69
  ...
Stacktrace:
 [1] _jl_spawn(::String, ::Array{String,1}, ::Cmd, ::Tuple{RawFD,Base.GenericIOBuffer{Array{UInt8,1}},RawFD}) at ./process.jl:345
 [2] (::getfield(Base, Symbol("##495#496")){Cmd})(::Tuple{RawFD,Base.GenericIOBuffer{Array{UInt8,1}},RawFD}) at ./process.jl:509
 [3] setup_stdio(::getfield(Base, Symbol("##495#496")){Cmd}, ::Tuple{RawFD,Base.GenericIOBuffer{Array{UInt8,1}},RawFD}) at ./process.jl:490
 [4] #_spawn#494(::Nothing, ::Function, ::Cmd, ::Tuple{RawFD,Base.GenericIOBuffer{Array{UInt8,1}},RawFD}) at ./process.jl:508
 [5] (::getfield(Base, Symbol("#kw##_spawn")))(::NamedTuple{(:chain,),Tuple{Nothing}}, ::typeof(Base._spawn), ::Cmd, ::Tuple{RawFD,Base.GenericIOBuffer{Array{UInt8,1}},RawFD}) at ./none:0
 [6] #_spawn#491(::Nothing, ::Function, ::Base.CmdRedirect, ::Tuple{RawFD,RawFD,RawFD}) at ./process.jl:398
 [7] _spawn at ./process.jl:398 [inlined]
 [8] #run#505(::Bool, ::Function, ::Base.CmdRedirect) at ./process.jl:662
 [9] run(::Base.CmdRedirect) at ./process.jl:661
 [10] top-level scope at none:0
```

odow:
Yeah. I'm a bit confused at all the IO behind the scenes stuff. We should probably just switch to building the docs with Julia 1.6.

mortenpi:
> We should probably just switch to building the docs with Julia 1.6.

Yep, that should take care of this. But I think we want to fix this in Documenter as well, since we do support 1.0 there.

odow (approved):
I've tested this locally, so I'll merge once CI passes. It's just a documenter push-preview issue that won't occur when I push to `master`.

jd-foster:
This turned out to be a much bigger issue that I thought! :)

PR #2661, opened 2021-08-11
Add _get_try and fallbacks for some elements of solution summary

### LINKED ISSUE #2648: Bugs in solution summary when results not available
The `solution_summary` code doesn't handle infeasible solutions correctly:
https://github.com/jump-dev/JuMP.jl/blob/6901e8f98bf24242a141e003e42530fa90f33c3c/src/print.jl#L463-L482
`objective_value` and `solve_time` should probably be wrapped in `_try_get`, and default fallbacks provided for things like `solver_name` and `raw_status`.

AtsushiSakai:
To fix #2648

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -470,12 +470,12 @@ function solution_summary(model::Model; verbose::Bool = false)
         result_count(model),
         has_values(model),
         has_duals(model),
-        objective_value(model),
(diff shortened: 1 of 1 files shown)


odow (approved):
Thanks! The documentation failure isn't your fault. We're waiting for a new tag of Documenter.jl.

PR #2665, opened 2021-08-13
nothing comparison improvement in DenseAxisArray.jl

AtsushiSakai:
I'm not sure this is intended or not, but `DenseAxisArray.jl` is using equality comparison (==) for nothing, not identical comparison (===).
If this is not intended, I think later one should be used for small performance improvement.
If this is intended, I'm sorry for bothering you, please close this PR.


blegat (approved):

blegat:
Thanks! It should indeed be `===`

PR #2682, opened 2021-09-11
Style guide: `using Foo: Foo` is an acceptable substitute for `import Foo`

DilumAluthge:
Some people don't like using `import Foo`, so I want to clarify that `using Foo: Foo` is an acceptable substitute for `import Foo`.

The problem is specifically with `using Foo`.

Therefore, I think that adding this clarification may make this guideline more palatable.

### DIFF
--- a/docs/src/developers/style.md
+++ b/docs/src/developers/style.md
@@ -343,12 +343,13 @@ For the same reason that `from <module> import *` is not recommended in python
 statement makes it harder to track where symbols come from and exposes the code
 to ambiguities when two modules export the same symbol.
 
-`using Foo: Foo` is an acceptable substitute for `import Foo`.
-It is only `using Foo` that must be avoided.
-
 Prefer `using ModuleName: x, p` to `import ModuleName.x, ModuleName.p` and
 `import MyModule: x, p` because the `import` versions allow method extension
 without qualifying with the module name.
+
+Similarly, `using ModuleName: ModuleName` is an acceptable substitute for
+`import ModuleName`, because it does not bring all symbols exported by
+`ModuleName` into scope. However, we prefer `import ModuleName` for consistency.
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
cc: @kristofferc

docs/src/developers/style.md
```diff
@@ -343,6 +343,9 @@ For the same reason that `from <module> import *` is not recommended in python
 statement makes it harder to track where symbols come from and exposes the code
 to ambiguities when two modules export the same symbol.
 
+`using Foo: Foo` is an acceptable substitute for `import Foo`.
+It is only `using Foo` that must be avoided.
```
odow, line 347: What is the motivation for `using Foo: Foo` over `import Foo`? When would you use the former and not the latter?

DilumAluthge: There's no advantage. As I understand it, some people just have a personal preference to avoid `import`. I don't want those people to ignore this guideline, which is why I want to mention `using Foo: Foo`. (Otherwise, they might ignore the guideline and keep using `using Foo`.)

Personally, I use `import Foo`, because I think it looks cleaner.

odow: The JuMP style guide is reasonably opinionated, and writing `using JuMP: JuMP` or `using MathOptInterface: MathOptInterface` just seems more liable for confusion. If we all prefer `import Foo`, I don't see the need to tell people there is an alternative.

This section could be improved with an example perhaps?
````
Okay in throw-away scripts or at the REPL:
```julia
using JuMP
model = Model()
```

Preferred in packages, or in more substantial scripts:
```julia
import JuMP
model = JuMP.Model()
```
````

I'd even go as far as saying we should remove the sentence below. Public/private methods should be distinguished with `_`. And then either everything is imported with `using` or nothing is with `import`. That's an easy rule for new users to remember.

mlubin: I think it's fair to discuss that `using JuMP: JuMP` doesn't suffer from same issues as `using JuMP`, but that we prefer `import JuMP` for consistency. It's within the scope of the style guide to express a preference if there are two ways to do the same thing.

mlubin (approved):

Issue #2683, opened 2021-09-13
Large SparseAxisArrays get stuck on `show`


raphaelsaavedra:
Not sure if this is a known issue, I have found https://github.com/jump-dev/JuMP.jl/pull/2630 which seems somewhat related to this.

When I have a very large `SparseAxisArray` and I (accidentally or not) try to show it in the REPL, the following is printed in the REPL:

```
SparseAxisArray{Float64, 2, Tuple{String, Int64}} with 115134138 entries:

```

and it just takes forever and gets stuck. I end up having to kill the terminal.

odow:
Yes, presumably this is the cause:
https://github.com/jump-dev/JuMP.jl/blob/c9e7a894524fdbeecd98c3426d4ab4123dc07fb9/src/Containers/SparseAxisArray.jl#L289-L304

Something like figuring out the maximum length of a key when there are 115 million entries is a lot of work...

We should filter the rows that are printed up-front, and then only loop through those, rather than checking with the `print_entry` function.

blegat:
I guess we could improve things here:
https://github.com/jump-dev/JuMP.jl/blob/c9e7a894524fdbeecd98c3426d4ab4123dc07fb9/src/Containers/SparseAxisArray.jl#L284-L287
1) `typemax(Int)` by default seems to be a lot
2) Why do we create this filter `print_entry` and `key_strings` with all the entries and not just `key_strings` with only the entry we want to print directly ? It seems to be because we want stable selection of the entries we print that does not depend on the order of the dictionary. However,  `join(key, ", ")` might allocate a lot

PR #2686, opened 2021-09-16
Fix performance of show on SparseAxisArray

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/2683

@raphaelsaavedra try this branch out.

### DIFF
--- a/src/Containers/SparseAxisArray.jl
+++ b/src/Containers/SparseAxisArray.jl
@@ -287,8 +287,9 @@ function Base.show(io::IOContext, x::SparseAxisArray)
     end
     # For stable printing, sort by the string value of the key.
     key_strings = [
-        (join(key, ", "), value) for (i, (key, value)) in enumerate(x.data)
-        if i < half_screen_rows || i > length(x) - half_screen_rows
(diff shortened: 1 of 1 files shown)


raphaelsaavedra (commented):
Thanks, that completely solved the issue :)

src/Containers/SparseAxisArray.jl
```diff
@@ -282,24 +282,25 @@ function Base.show(io::IOContext, x::SparseAxisArray)
...
-    # For stable printing, sort by the string value of the key.
-    key_strings = [(join(key, ", "), value) for (key, value) in x.data]
-    sort!(key_strings; by = x -> x[1])
-    pad = maximum(
-        print_entry(i) ? length(x[1]) : 0 for (i, x) in enumerate(key_strings)
-    )
     if !haskey(io, :compact)
         io = IOContext(io, :compact => true)
     end
+    # For stable printing, sort by the string value of the key.
```
blegat, line 288: The printing is not stable anymore since the filter we apply depends of the order of the element in the dictionary iterators. We should comment about it to say that this is for performance reason

PR #2688, opened 2021-09-17
[docs] Clarify callback gotchas

### LINKED ISSUE #2642: Document callback gotchas
The documentation around callbacks can be improved.

x-ref https://github.com/gamma-opt/DecisionProgramming.jl/issues/20#issuecomment-874540048

cc @solliolli

## Only set one callback

The docstring of `AbstractCallback` contains a very important sentence that should be highlighted in the JuMP documentation:
> At most one callback of each type can be registered. If an optimizer already has a function for a callback type, and the user registers a new function, then the old one is replaced.

It's pretty easy to imagine users wanting to set multiple callbacks!

## LazyConstraint's may be added multiple times

Solvers (e.g., Gurobi) may remove previously added lazy constraints during a solve and end up revisiting a point that was previous cut off.

odow:
Closes #2642

### DIFF
--- a/docs/src/manual/callbacks.md
+++ b/docs/src/manual/callbacks.md
@@ -70,6 +70,11 @@ how to use it in the README of their Github repository.
 If you want to modify the problem in a callback, you _must_ use a lazy
 constraint.
 
+ !!! warning
+    You can only set each callback once. Calling `set` twice will over-write the
+    earlier callback. In addition, if you use a solver-independent callback, you
+    cannot set a solver-dependent callback.
+
 ## Lazy constraints
 
 Lazy constraints are useful when the full set of constraints is too large to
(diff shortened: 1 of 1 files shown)


docs/src/manual/callbacks.md
```diff
@@ -70,6 +70,11 @@ how to use it in the README of their Github repository.
 If you want to modify the problem in a callback, you _must_ use a lazy
 constraint.
 
+ !!! warning
+    You can only set each callback once. Calling `set` twice will over-write the
+    earlier callback. In addition, if you use a solver-independent callback, you
+    cannot set a solver-dependent callback.
```
blegat, line 76: This may be solver specific. Some solvers might support callback that are compatible with solver independent one, e.g. a callback that should return whether to stop

odow: This sentence is true for all current solvers implementing callbacks, so I'd prefer to leave as-is for now.

PR #2689, opened 2021-09-17
[Containers] add Array constructor for DenseAxisArray

### LINKED ISSUE #2644: Constructor of Array with DenseAxisArray
This was discussed in the JuliaLab meeting today. The current behavior of `JuMP.DenseAxisArray` is consistent with `OffsetArray` in the sense that `similar` converts it onto an `Array` but `Array` gives an error.
Making `convert` work would be too much as `convert` is called implicitly when in `setindex!` or when setting fields of `struct`s.
However, @stevengj suggested that the `Array` constructor should be allowed to coercing the `JuMP.DenseAxisArray` into dropping its dimensions would be reasonable.
We might want to follow the development on `OffsetArray` and follow what's decided there.

odow:
Closes #2644

### DIFF
--- a/docs/src/manual/containers.md
+++ b/docs/src/manual/containers.md
@@ -173,6 +173,16 @@ And data, a 2×2 Array{Tuple{Symbol,Int64},2}:
  (:A, 2)  (:B, 2)
 ```
 
+### Access internal data
+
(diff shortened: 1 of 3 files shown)


blegat (approved):
Would we want to say something in the doc?

PR #2690, opened 2021-09-17
Make NonlinearExpression <: AbstractJuMPScalar

### LINKED ISSUE #2635: Error message improvement for value(Array of NLexpression)
When calling `value(u)` where `u` is an array of variables the following error shows up
```julia
julia> value(u)
ERROR: `JuMP.value` is not defined for collections of JuMP types. Use Julia's broadcast syntax instead: `JuMP.value.(x)`.
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:33
 [2] value(#unused#::Matrix{VariableRef})
   @ JuMP C:\Users\user\.julia\packages\JuMP\Xrr7O\src\variables.jl:1204
```
This error is helpful.

On the other hand, when calling `value(u)` where `u` is an array of nonlinear expressions the error is
```julia
ERROR: LoadError: MethodError: no method matching value(::Matrix{NonlinearExpression})
```
This is not as helpful as it does not say that `value.` should be used instead.

odow:
Closes #2635

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -1364,7 +1364,7 @@ end
 
 # TODO: rename "m" field to "model" for style compliance
 """
-    NonlinearExpression
+    NonlinearExpression <: AbstractJuMPScalar
 
 A struct to represent a nonlinear expression.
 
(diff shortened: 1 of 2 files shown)


src/JuMP.jl
```diff
@@ -1370,7 +1370,7 @@ A struct to represent a nonlinear expression.
 
 Create an expression using [`@NLexpression`](@ref).
 """
-struct NonlinearExpression
+struct NonlinearExpression <: AbstractJuMPScalar
```
blegat, line 1373: Update docstring as well?

PR #2709, opened 2021-09-26
[breaking] Reset model to OPTIMIZE_NOT_CALLED if model is modified after optimize

### LINKED ISSUE #2566: Error on modify-then-query
If you solve a problem, modify it, then query a solution, the result is undefined behavior.

We've had a few bites at this:
- https://github.com/jump-dev/JuMP.jl/pull/2076
- https://github.com/jump-dev/JuMP.jl/issues/2400
- https://github.com/jump-dev/JuMP.jl/pull/2401

The docs currently say
image
but the warning is buried in the documentation and it keeps tripping people up.

I would propose we add a flag to check this at the JuMP level and throw `OptimizeNotCalled`. Asking every low-level solver to do so seems like too much work, and some solvers are fine with it.

The other option is to do it at the CachingOptimizer level.

odow:
This is a frequent confusion for users of all types, because the underlying
behavior at the solver level is undefined.

Closes #2566

### DIFF
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -284,24 +284,33 @@ Solution is optimal
   dual solution: c1 = 1.7142857142857142
 ```
 
-!!! warning
-    Querying solution information after modifying a solved model is undefined
-    behavior, and solvers may throw an error or return incorrect results.
-    Modifications include adding, deleting, or modifying any variable,
-    objective, or constraint. Instead of modify-then-query, query the results
-    first, then modify the problem. For example:
-    ```julia
-    model = Model(GLPK.Optimizer)
-    @variable(model, x >= 0)
-    optimize!(model)
-    # Bad:
-    set_lower_bound(x, 1)
-    @show value(x)
-    # Good:
-    x_val = value(x)
-    set_lower_bound(x, 1)
-    @show x_val
-    ```
+## OptimizeNotCalled errors
+
+Modifing a model after calling [`optimize!`](@ref) will reset the model into
+the `MOI.OPTIMIZE_NOT_CALLED` state. If you attempt to query solution
+information, an `OptimizeNotCalled` error will be thrown.
+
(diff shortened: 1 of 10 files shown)


odow:
This is worth a look @blegat and @mlubin (and anyone else interested). 

This is going to break some users code, but their code will only have worked for particular solvers, and if they swapped solvers, it's likely that bad things would have happened.

odow:
Yeah I guess allowing this in direct-mode is reasonable. One reason we never did this at the MOI level is that some solvers do support accessing solutions after modification.

odow:
Updated so that DIRECT mode ignores the dirty flag

blegat:
This look like a workaround to solvers bug. What don't we fix the solver wrappers instead ? We want a consistent behavior between MOI and JuMP.

odow:
It's not solver bug because it exposes the underlying behavior of the solver. 

We don't want to fix this in solver wrappers because: 
 * It would involve adding this flag to every solver that supports incremental modification
 * If you are using the solver directly, you may want/expect the underlying behavior

Some solvers (like Gurobi) already throw error messages for this. But they don't adequately explain why. So it says "Unable to retrieve attribute `Slack`" when "Unable to retrieve attribute `Slack` because the model has been modified since the last solve" would be better.


If you're in direct mode, or using MOI directly, caveat emptor.

blegat:
Ok, this is departing from "A JuMP model is just an MOI model with fancy macros" but we did the same with
https://github.com/jump-dev/JuMP.jl/blob/e5a4258615c68c4aeedc7c03d193df768aff0322/src/JuMP.jl#L1193-L1195
so it's consistent.

blegat (approved):

odow:
> this is departing from "A JuMP model is just an MOI model with fancy macros"

Yeah. This is a somewhat special case in that I see it trip up users of all experience levels frequently.

Issue #1151, opened 2017-12-18
Should we define zero(AffExpr) and one(AffExpr)?


mlubin:
The implementation of `zeros(AffExpr, n)` looks like `fill!(Array{AffExpr}(n), zero(AffExpr))`, which is problematic because the result is `n` entries that refer to the same `AffExpr` (https://github.com/JuliaOpt/JuMP.jl/issues/1113). Same for `ones`.

This leads me to question if we should define `zero(AffExpr)` or `one(AffExpr)` at all given that code may be written assuming that these return immutable objects. The replacement is `AffExpr(0.0)` and `AffExpr(1.0)`.

@daschw

blegat:
If Base is doing this, they may be implicitly assuming that only immutable objects should define zero and one.

Maybe we could have two version of `AffExpr` ? One immutable and one mutable.
The mutability of `AffExpr` is only used inside JuMP macros so we could take care of using the mutable version there.
If the user do operation on variables outside JuMP macros it will become the immutable version.

mlubin:
> The mutability of AffExpr is only used inside JuMP macros so we could take care of using the mutable version there.

Not true, we also expose `push!` and some other functions to append to `AffExpr`s. These are pretty useful in performance-critical situations.

joaquimg:
> The replacement is AffExpr(0.0) and AffExpr(1.0)

seems reasonable.

blegat:
> Not true, we also expose `push!` and some other functions to append to `AffExpr`s. These are pretty useful in performance-critical situations.

We could have a warning `push! not supported in AffExpr, please convert it to a MutableAffExpr`.

mlubin:
Having both a mutable and immutable AffExpr is worth considering, but I'd put it off until after 0.19.

joaquimg:
What is the advantage of having the immutable version?

mlubin:
This example by @jdlara-berkeley suggests a similar issue with `convert`:
```julia
m = Model()
x = @variable(m, x >= 0)
NetInjectionVar =  Array{JuMP.GenericAffExpr{Float64,JuMP.Variable},2}(5, 24)
NetInjectionVar[:]= 0.0 # Add this line to avoid #undef and enable addition of terms in the future without errors
append!(NetInjectionVar[1,1],x)
NetInjectionVar
5×24 Array{JuMP.GenericAffExpr{Float64,JuMP.Variable},2}:
[...]

PR #2711, opened 2021-09-27
WIP: remove zero() and one() for variables and expressions

odow:
See issue #1151 for background.

The problem is that calls like zero(AffExpr) lead to hard-to-diagnose
bugs.

However, much of Julia's LinearAlgebra routines assume that zero and
one are defined for the element types of arrays. Thus, if we do remove
these methods, we're likely to break a lot of user-code.

Perhaps we're best to just keep these definitions, and document their gotchas.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -1092,15 +1092,6 @@ include("constraints.jl")
 include("variables.jl")
 include("objective.jl")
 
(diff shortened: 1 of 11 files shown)


odow:
Closing because the conclusion in #1151 is that we should retain these for JuMP 1.0.

PR #2714, opened 2021-09-28
Clarify documentation of build_variable

### LINKED ISSUE #2583: Disallow multiple positional arguments in the `@variable` macro
The "positional" arguments `Bin`, `Int`, `SDP` etc are special and can be in any order.
However, the user (extension) defined arguments are ordered.

Hence,
`@variable(model, x[1:2,1:2], Int, PSD)`
and
`@variable(model, x[1:2,1:2], PSD, Int)`
work graciously.

`@variable(model, x[1:2,1:2], PSD, Foo, Int, Bar)`
requires that we implement:
`build_variable(... , ::Type{Foo}, ::Type{Bar})`
and
`@variable(model, x[1:2,1:2], PSD, Int, Bar, Foo)`
requires that we implement:
`build_variable(... , ::Type{Bar}, ::Type{Foo})`

if `build_variable(... , ::Type{Foo}, ::Type{Bar})` is implemented but `build_variable(... , ::Type{Bar}, ::Type{Foo})` is not, then
```julia
@variable(model, x, Bar, Foo)
ERROR: At REPL[58]:1: `@variable(model, sesx, Bar(), Foo())`: Unrecognized arguments: Bar, Foo.
(You may have passed these as positional arguments, or as a keyword value to `variable_type`.)

If you're trying to create a JuMP extension, you need to implement `build_variable`.
Stacktrace:
...
```
This is not exactly true also.

And even worse if you add more parameters.
The end-user might get a weird error if not all the dispatches are not implemented.

The main point is: too error-prone for extensions. And no clear use case.
Extensions can always define `FooBar` if they need some combination.

related to: https://github.com/jump-dev/JuMP.jl/pull/2421#issuecomment-774062945 (and  above)
and to: https://github.com/jump-dev/JuMP.jl/pull/2574#discussion_r624849218

odow:
Closes #2583

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1401,43 +1401,97 @@ _esc_non_constant(x::Expr) = isexpr(x, :quote) ? x : esc(x)
 _esc_non_constant(x) = esc(x)
 
 """
-    build_variable(_error::Function, info::VariableInfo; extra_kw_args...)
+    build_variable(
+        _error::Function,
+        info::VariableInfo,
+        args...;
+        kwargs...,
+    )
 
-Returns a new variable.
+Return a new [`AbstractVariable`](@ref) object.
 
-Extensions should define a method with additional positional arguments to
-dispatch the call to a different method.
+This method should only be implemented by developers creating JuMP extensions.
+It should never be called by users of JuMP.
 
-As an example, `@variable(model, x, foo)` foo will call
-`build_variable(_error, info, foo)`
+## Arguments
 
-See the [`@variable`](@ref) macro doc for more details.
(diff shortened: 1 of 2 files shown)


blegat (approved):
I down know how useful is this now that we have the `in` syntax though

odow:
I use it in SDDP.jl, so I'd like to keep it. It's very nice to be able to write
```julia
@variable(model, 0 <= x[a=1:4] <= a, Int, SDDP.State, initial_value = sqrt(a))
```

blegat:
Yes, I also use it in SetProg, PolyJuMP and SumOfSquares, we can keep the two options

PR #2716, opened 2021-09-29
[docs] Document zero and one gotchas for expressions

### LINKED ISSUE #1151: Should we define zero(AffExpr) and one(AffExpr)?

odow:
Closes #1151

### DIFF
--- a/docs/src/manual/expressions.md
+++ b/docs/src/manual/expressions.md
@@ -315,7 +315,8 @@ julia> one(AffExpr)
 ```
 
 However, this can result in a subtle bug if you call
-[`add_to_expression!`](@ref) on an element created by `zeros` or `ones`:
+[`add_to_expression!`](@ref) or the [MutableArithmetics API](https://github.com/jump-dev/MutableArithmetics.jl)
+on an element created by `zeros` or `ones`:
 ```jldoctest
 julia> x = zeros(AffExpr, 2)
 2-element Vector{AffExpr}:
@@ -393,3 +394,5 @@ julia> x
  1.1
  0
 ```
+Note that for large expressions this will be slower due to the allocation of
+additional temporary objects.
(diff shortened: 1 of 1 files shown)


docs/src/manual/expressions.md
```diff
@@ -292,3 +300,96 @@ macro and can be used only in [`@NLobjective`](@ref), [`@NLconstraint`](@ref),
...
+julia> add_to_expression!(x[1], 1.1)
+1.1
+
+julia> x
+2-element Vector{AffExpr}:
+ 1.1
+ 0
+```
+
+Alternatively, use non-mutating operation to avoid updating `x[1]` in-place:
```
blegat, line 381: Mention that this might be slow

docs/src/manual/expressions.md
```diff
@@ -292,3 +300,96 @@ macro and can be used only in [`@NLobjective`](@ref), [`@NLconstraint`](@ref),
...
+```jldoctest
+julia> zero(AffExpr)
+0
+
+julia> one(AffExpr)
+1
+```
+
+However, this can result in a subtle bug if you call
+[`add_to_expression!`](@ref) on an element created by `zeros` or `ones`:
```
blegat, line 318: Or the MA API

blegat (approved):

PR #2717, opened 2021-09-29
Fix bug when querying attributes from an optimizer

### LINKED ISSUE #2587: Custom attributes not queried correctly with caching optimizer
See https://stackoverflow.com/questions/61922425/how-can-i-determine-whether-a-jump-model-solved-by-gurobi-is-a-mip

Issue is in functions like this:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/3f872f9edb50b1a97cb65011430a0160fd56111e/src/Utilities/cachingoptimizer.jl#L539-L557
where we need a branch to see if it is possible to query directly from the optimizer.

odow:
In particular, we now wrap attributes in AttributeFromOptimizer.

Note that we do not need to do this for set, because the attributes
will be stored in the model cache, and copied to the optimizer when
it is attached.

Closes #2587

I've confirmed this works with Gurobi as well
```julia
julia> @testset "Gurobi" begin
       @testset "direct_model" begin
           @variable(model, x >= 1.5)
           @constraint(model, c, 2x + 1 == 0)

           @test get_optimizer_attribute(model, Gurobi.ModelAttribute("IsMIP")) == 0
           @test get_optimizer_attribute(model, Gurobi.VariableAttribute("LB"), x) == 1.5
           @test get_optimizer_attribute(model, Gurobi.ConstraintAttribute("RHS"), c) == -1.0
       end
       @testset "NO_OPTIMIZER" begin
           model = Model()
           @variable(model, x >= 1.5)
           @constraint(model, c, 2x + 1 == 0)
           @test_throws(
               KeyError,
               get_optimizer_attribute(model, Gurobi.ModelAttribute("IsMIP")),
           )
           @test get_optimizer_attribute(model, Gurobi.VariableAttribute("LB"), x) === nothing
           @test get_optimizer_attribute(model, Gurobi.ConstraintAttribute("RHS"), c) === nothing
       end
       @testset "EMPTY_OPTIMIZER" begin
           @variable(model, x >= 1.5)
           @constraint(model, c, 2x + 1 == 0)

           @test get_optimizer_attribute(model, Gurobi.ModelAttribute("IsMIP")) == 0
           @test get_optimizer_attribute(model, Gurobi.VariableAttribute("LB"), x) == 1.5
           @test get_optimizer_attribute(model, Gurobi.ConstraintAttribute("RHS"), c) == -1.0
       end
       @testset "ATTACHED_OPTIMIZER" begin
           @variable(model, x >= 1.5)
           @constraint(model, c, 2x + 1 == 0)
           MOI.Utilities.attach_optimizer(model)
           @test get_optimizer_attribute(model, Gurobi.ModelAttribute("IsMIP")) == 0
           @test get_optimizer_attribute(model, Gurobi.VariableAttribute("LB"), x) == 1.5
           @test get_optimizer_attribute(model, Gurobi.ConstraintAttribute("RHS"), c) == -1.0
       end
       end
Academic license - for non-commercial use only - expires 2021-11-06
Academic license - for non-commercial use only - expires 2021-11-06
Academic license - for non-commercial use only - expires 2021-11-06
Test Summary: | Pass  Total
Gurobi        |   12     12
Test.DefaultTestSet("Gurobi", Any[Test.DefaultTestSet("direct_model", Any[], 3, false, false), Test.DefaultTestSet("NO_OPTIMIZER", Any[], 3, false, false), Test.DefaultTestSet("EMPTY_OPTIMIZER", Any[], 3, false, false), Test.DefaultTestSet("ATTACHED_OPTIMIZER", Any[], 3, false, false)], 0, false, false)
```

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -954,20 +954,15 @@ function get_optimizer_attribute(model::Model, attr::MOI.AnyAttribute, args...)
     end
     # ... else we have a caching optimizer.
     state = MOI.Utilities.state(backend(model))
-    if state == MOI.Utilities.NO_OPTIMIZER
-        # There's no optimizer, so we can just query this from the model cache.
+    if state == MOI.Utilities.NO_OPTIMIZER ||
+       state == MOI.Utilities.EMPTY_OPTIMIZER
+        # Query from the model cache.
         return MOI.get(model, attr, args...)
-    elseif state == MOI.Utilities.EMPTY_OPTIMIZER
-        # There's an optimizer, but it's not attached. Since the user might be
-        # querying some low-level attribute of a variable or constraint, we need
-        # to attach the optimizer.
-        MOI.Utilities.attach_optimizer(model)
     else
-        # Nothing to do here.
         @assert state == MOI.Utilities.ATTACHED_OPTIMIZER
+        new_attr = MOI.Utilities.AttributeFromOptimizer(attr)
+        return MOI.get(backend(model), new_attr, index.(args)...)
     end
-    new_attr = MOI.Utilities.AttributeFromOptimizer(attr)
-    return MOI.get(backend(model), new_attr, index.(args)...)
 end
 
 # Special case to the method above: no need to wrap optimizer attributes in
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -933,6 +933,15 @@ end
 
 Return the value of the solver-specific attribute `attr` in `model`.
 
+!!! warning
+    If you are querying a solver-specific attribute such as
+    `Gurobi.ModelAttribute`, and you are not using [`direct_model`](@ref), and
+    you have not called [`optimize!`](@ref), you must first call
+    `MOI.Utilities.attach_optimizer(model)` in order to first copy the problem
+    into the solver.
+
+See also: [`set_optimizer_attribute`](@ref), [`set_optimizer_attributes`](@ref).
+
 ## Examples
 
 Query model or optimizer attributes:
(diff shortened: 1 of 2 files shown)


src/JuMP.jl
```diff
@@ -933,24 +933,57 @@ end
...
+    # ... else we have a caching optimizer.
+    state = MOI.Utilities.state(backend(model))
+    if state == MOI.Utilities.NO_OPTIMIZER
+        # There's no optimizer, so we can just query this from the model cache.
+        return MOI.get(model, attr, args...)
+    elseif state == MOI.Utilities.EMPTY_OPTIMIZER
+        # There's an optimizer, but it's not attached. Since the user might be
+        # querying some low-level attribute of a variable or constraint, we need
+        # to attach the optimizer.
+        MOI.Utilities.attach_optimizer(model)
```
blegat, line 971: This won't work for solvers only implementing `optimize!`

odow: The alternative is to throw a nice error and encourage users to call `optimize!` or `MOI.Utilities.attach_optimizer`. Perhaps that's better.

odow: How about this now?

src/JuMP.jl
```diff
@@ -938,24 +938,65 @@ end
...
+    end
+    # ... else we have a caching optimizer.
+    state = MOI.Utilities.state(backend(model))
+    if state == MOI.Utilities.NO_OPTIMIZER ||
+       state == MOI.Utilities.EMPTY_OPTIMIZER
+        # Query from the model cache.
+        return MOI.get(model, attr, args...)
+    else
+        @assert state == MOI.Utilities.ATTACHED_OPTIMIZER
+        new_attr = MOI.Utilities.AttributeFromOptimizer(attr)
```
blegat, line 988: `AttributeFromOptimizer` will allow you to go through one CachingOptimizer layer but if there is a second one then you will query from its cache.

blegat: It seems we need to do a recursion through the MOI layers at the JuMP level and not use `AttributeFromOptimizer`. Why not use `MOI.get(unsafe_backend(model), attr, ...)` ? By the way, there is the same discussion about having to call attach first there.

odow: How about `unsafe_get_optimizer_attribute` if you want to query solver-specific attributes that you haven't set yet?

blegat: Keeping also the method of this PR?

odow:
We concluded this could be fixed in Gurobi.jl:
https://github.com/jump-dev/JuMP.jl/pull/2784#issuecomment-956553900

PR #2725, opened 2021-10-01
Add better error messages for vectors passed as variable bounds

### LINKED ISSUE #2056: Better error messages for vector variable bounds
Hello,

As suggested by @odow in [this discussion](https://discourse.julialang.org/t/jump-glpk-with-equal-upper-and-lower-bounds/28469/4), I report a possible enhancement of error message when one inadvertidly uses broadcasted vector bounds when creating a vector of JuMP variables.

Considering the following preamble code,
```julia
using JuMP
model = Model()
xmax = [10., 0.]
```
the correct way to have a variable `x<=xmax` elementwise is by using an explicit indexing variable (which is [documented](http://www.juliaopt.org/JuMP.jl/stable/variables/#Variable-containers-1), but not exactly for this vector slicing usecase)

```julia
@variable(model, x[i=1:2] <= xmax[i])
```

However, there is a wrong approach that someone can easily follow (at least I did -;)), using  broadcasting on the comparison operator:
```julia
@variable(model, x[1:2] .<= xmax)
```
which yields the surprising error message: 
```julia
ERROR: LoadError: In `@variable(model, x[1:2] .<= xmax)`: Unknown sense .<=.
```

Also, in the case of two bounds, with the same mistake:
```julia
@variable(model, 0 <= x[1:2] .<= xmax)
```
the error message is different but also not helpful:
```julia
ERROR: LoadError: In `@variable(model, 0 <= x[1:2] .<= xmax)`: Use the form lb <= ... <= ub.
```

Expected behavior: in an ideal world, the error message would point out to the correct solution (ex : "broadcasted constraints are not yet supported, use ... instead"). Alternatively, it may be worth supporting the broadcasted syntax, but I don't have enough background on JuMP to judge the difficulty of the task. This last idea may relate to the closed issue #1455, although it was about applying broadcasting on the `set_lower_bound` function not within the `@variable` macro.

odow:
Closes #2056

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1479,6 +1479,36 @@ function build_variable(
             "`build_variable`. Read the docstring for more details.",
         )
     end
+    if info.lower_bound isa AbstractArray
+        _error(
+            "Passing arrays as variable bounds is not supported. Instead of " *
+            "`@variable(model, x[1:2] >= lb)`, do " *
+            "`@variable(model, x[i=1:2] >= lb[i])`. Alternatively, create " *
+            "the variable without bounds, then call `set_lower_bound.(x, lb)`",
+        )
+    elseif info.upper_bound isa AbstractArray
+        _error(
+            "Passing arrays as variable bounds is not supported. Instead of " *
+            "`@variable(model, x[1:2] <= ub)`, do " *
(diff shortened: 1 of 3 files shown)


blegat (approved):
LGTM when CI passes

PR #2729, opened 2021-10-03
Add error message for variable comparison with number

### LINKED ISSUE #2461: sum comparison parses wrongly without error
I'm currently thinking about allowing comparisons inside a sum for my ConstraintSolver.jl package. (See: https://github.com/Wikunia/ConstraintSolver.jl/issues/254)

Unfortunately I encountered the following problem: When combining comparisons inside the sum constraint with `&&` i.e 
```
model = Model()
@variable(model, x[1:5], Bin)
@constraint(model, sum(x[i] == 1 && x[i+1] == 1 for i=1:4) <= 1)
```
it currently gets parsed as `0 <= 1` instead of throwing an error as it does for things like:
```
@constraint(model, sum(x[i] == 1 for i=1:4) <= 1)
```

It's quite unlikely that someone encounters this but would be nice if it throws an error.

odow:
I see this on the forum a lot, so this error message should help. I'm a little wary of it breaking things though. There might be valid cases for comparing variables and numbers (e.g., in the keys of a dictionary).

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1445,7 +1445,7 @@ end
 ### Error messages for common incorrect usages
 ###
 
-for sym in (:(==), :(<=), :(>=))
+for sym in (:(<=), :(>=), :(<), :(>))
     msg = """Cannot evaluate `$(sym)` between a variable and a number.
 
     There are two common mistakes that lead to this.
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1440,3 +1440,56 @@ function relax_integrality(model::Model)
     end
     return unrelax
 end
+
+###
+### Error messages for common incorrect usages
+###
+
+for sym in (:(==), :(<=), :(>=))
+    msg = """Cannot evaluate `$(sym)` between a variable and a number.
+
+    There are two common mistakes that lead to this.
+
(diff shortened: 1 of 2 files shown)


odow:
The failure on 1.0 is one example of breakage.

Julia used to use `t != 0`
https://github.com/JuliaLang/julia/blob/3af96bcefcb0aec559d4652fd75e1bc3fcfe99da/stdlib/SparseArrays/src/sparsematrix.jl#L383

But it now uses `iszero`
https://github.com/JuliaLang/julia/blob/1b93d53fc4bb59350ada898038ed4de2994cce33/stdlib/SparseArrays/src/sparsematrix.jl#L627

Perhaps we shouldn't merge this. It gives a nice error message though.

blegat (approved):

blegat:
This is a frequent error so I would support this but I just read your last comment. It is indeed worrying.

odow:
Perhaps we should just do the inequalities, and still have `==(::VariableRef, ::Number) = false`.

blegat:
Yes, I don't imagine someone needing to use JuMP variables and numbers in a binary search tree. Comparison between JuMP variables don't work anyway

Issue #2318, opened 2020-08-21
Subtle trap with semicolon syntax


mlubin:
```julia
julia> m = Model();

julia> @variable(m, x[i=1:4; iseven(i)])
JuMP.Containers.SparseAxisArray{VariableRef,1,Tuple{Int64}} with 2 entries:
  [4]  =  x[4]
  [2]  =  x[2]

julia> @variable(m, y[i=1:4, iseven(i)])
JuMP.Containers.SparseAxisArray{VariableRef,2,Tuple{Int64,Bool}} with 4 entries:
  [1, false]  =  y[1,false]
  [3, false]  =  y[3,false]
  [4, true ]  =  y[4,true]
  [2, true ]  =  y[2,true]
```

I'm not sure how to fix this, but it's unfortunate that a one-character change that's hard to spot can have this subtle effect.

odow:
I didn't even realize this would work.

The only way I see to fix this is to revive our [old discussion](https://github.com/jump-dev/JuMP.jl/issues/692#issuecomment-251197721) with something along the lines of:
```julia
@variable(model, x[i=1:4])  # x[1]
@variable(model, x[i=1:4, j=i:5])  # x[1, 2]
@variable(model, x[i for i=1:4 if iseven(i)])  # x[2]

@variable(model, x[(i, j) for i=1:4, j=2:3 if iseven(i+ j)])  # x[(2, 2)] or x[2, 2]?
```
For backward compatibility, the last one could just create a `Dict{Tuple{Int, Int}, VariableRef}`, which would clarify slicing issues, etc.

mlubin:
> We can also replace the ; syntax by a if syntax without changing anything else.

How would we do that? `@variable(m, x[i=1:4 if iseven(i)])` doesn't parse.

Printing a warning if index sets are scalars might make sense. It probably will catch a number of subtle bugs. On the other hand, it may annoy people if we break from Julia's rules for what's iterable.

blegat:
The user could do `Model(disable_scalar_iterator_warning=false)` to allow the user to disable this. A common mistake I do is to write `length(a)` instead of `1:length(a)`

odow:
This came up on Discourse: https://discourse.julialang.org/t/error-syntax-function-argument-names-not-unique-around/53944

All options are pretty unhelpful, and could do with a better error message.
```Julia
model = Model()
@variable(model, x)

julia> @constraint(model, [s=1:2, s = 1], x <= s)
[...]

PR #2730, opened 2021-10-04
Warn if DenseAxisArray is passed length-1 axis

odow:
Part of #2318

This seems like it would highlight a common type of bug:

```julia
julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x[1:2, t=1])
┌ Warning: Axis contains one element: 1. If intended, you can safely ignore this warning. To explicitly pass the axis with one element, pass `[1]` instead of `1`.
└ @ JuMP.Containers ~/.julia/dev/JuMP/src/Containers/DenseAxisArray.jl:153
    Dimension 1, Base.OneTo(2)
    Dimension 2, fill(1)
And data, a 2×1 Matrix{VariableRef}:
 x[1,1]
 x[2,1]

julia> @variable(model, y[1:2, t=[1]])
    Dimension 1, Base.OneTo(2)
    Dimension 2, [1]
And data, a 2×1 Matrix{VariableRef}:
 y[1,1]
 y[2,1]
```

### DIFF
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -147,7 +147,17 @@ function Base.get(
 end
 
 _abstract_vector(x::AbstractVector) = x
-_abstract_vector(x) = [a for a in x]
+function _abstract_vector(x)
+    axis = [a for a in x]
+    if length(axis) == 1
+        @warn(
+            "Axis contains one element: $x. If intended, you can safely " *
+            "ignore this warning. To explicitly pass the axis with one " *
+            "element, pass `[$(axis[1])]` instead of `$(x)`.",
+        )
+    end
+    return axis
+end
 
 """
     DenseAxisArray(data::Array{T, N}, axes...) where {T, N}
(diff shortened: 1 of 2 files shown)


src/Containers/DenseAxisArray.jl
```diff
@@ -147,7 +147,17 @@ function Base.get(
...
 
 _abstract_vector(x::AbstractVector) = x
-_abstract_vector(x) = [a for a in x]
+function _abstract_vector(x)
+    axis = [a for a in x]
+    if length(axis) == 1
+        @warn(
+            "Axis contains one element: $x. If intended, you can safely " *
+            "ignore this warning. To explicitly pass the axis with one " *
+            "element, pass `[$(axis[1])]` instead of `$(x)`.",
```
blegat, line 156: `[...]` won't work if you pass an iterator of one element. Maybe you want to check that `x isa eltype(axis)`

PR #2732, opened 2021-10-04
[breaking] Deprecate SDconstraint and add new syntax for PSDCone constraints

### LINKED ISSUE #1798: inequality syntax for set constraints
As discussed on Gitter, for the sake of a minimal and consistent interface for generic set constraint modeling, let's remove the special case `@SDconstraint` and generalize it for other sets.

The current syntax `@constraint(model, X in SomeSet())` is great as a foundation, but some users will prefer to write some constraints with inequalities, like `@constraint(model, X >= Y, SomeSet())` (which specifies that the inequality is with respect to `SomeSet()` instead of the equivalent `@constraint(model, X - Y in SomeSet())`.

There are many directions we could go from there for further improving convenience and usability, and @blegat made some good suggestions - those are more complicated and require more discussion re: DCP modeling in JuMP so let's have a separate longer term issue for that. I'm certainly eager to help on the DCP modeling aspect down the track so I'm keen to discuss at JuMP-Dev.

odow:
This syntax is quite nice, and trivial to implement. I think it also allows us to deprecate `@SDconstraint`.

Closes #1798 

TODO
 - [x] Deprecate `@SDconstraint`
 - [x] Document


joaquimg:
Very nice indeed. Does it already work for any cone?

odow:
Not yet. I've only defined it for PSDCone for now. We could make a decision to extend it to any cone

src/sd.jl
```diff
@@ -94,6 +94,41 @@ triangular part of the matrix is constrained to belong to the
...
 
+function build_constraint(
+    _error::Function,
+    f::AbstractMatrix{<:AbstractJuMPScalar},
+    s::MOI.GreaterThan,
+    extra::PSDCone,
+)
+    @assert iszero(s.lower)
+    return build_constraint(_error, f, extra)
+end
```
odow, line 105: Adding a new cone is just a case of writing these `build_constraint` functions.

joaquimg: It might be possible (for vector inequalities) to use MOI sets and do it with one or two functions, right?
Might just a need a function: `_known_cone`

blegat: We should try to make sure this leads to a consistent interface.
It seems we would want to allow `@constraint(model, x >= y, cone)` to be synonymous to `@constraint(model, x - y in cone)`.

odow: Yes, it's trivial to implement this syntax for any AbstractVectorSet. I just haven't done more widely to check it makes sense to do it for PSDCone first.

blegat: I'm wondering if we could have a cone trait so that we don't need to have a new build_constraint for every set.
That would also allow doing `@constraint(model, p >= q, SOSCone()) ` with SumOfSquares

odow: You don't need a new `build_constraint`. You can just have 

```Julia
function build_constraint(
    _error::Function,
    f::AbstractMatrix{<:AbstractJuMPScalar},
    s::MOI.GreaterThan,
    extra::Union{MOI.AbstractVectorSet,JuMP.AbstractVectorSet},
)
    @assert iszero(s.lower)
    return build_constraint(_error, f, extra)
end
```

The `PSDCone` definition should probably be `struct PSDCone <: JuMP.AbstractVectorSet end`.

blegat: AbstractVectorSet means the elements of this sets are vectors. This is not the case for PSDCone, they are matrices, nor for SOSCone, they are polynomials (hence the need for reshape, etc... to vectorize into an MOI vector set). Moreover, some AbstractVectorSet would not be cone so it wouldn't make sense to compare them with an inequality.

odow: It's also pretty easy to just add a `Union` for the sets in JuMP.

blegat: Yes, we can start with that. But what about MOI sets ? Do we make a union of all MOI conic sets ? And update it after each release of MOI ?

odow: > so it wouldn't make sense to compare its elements with an inequality.

Okay. Sure, it doesn't make sense to do `X >= Y, MOI.Complments(2)`, but if the rule is just `X >= Y, Set()` is short-hand for `X - Y in Set()`, then it's up to the user to choose. We don't have to make the JuMP syntax mathematically meaningful.

odow: Regardless to this discussion, I think that's best left to a separate PR. This one is for deprecating `@SDconstraint` in favor of `@constraint(model, X >= Y, PSDCone())`

chriscoey:
Chiming in to say that it would be very handy in Hypatia if we could easily (little extra code) make this syntax available for all Hypatia's cones (which are unioned over here: https://github.com/chriscoey/Hypatia.jl/blob/6da5ec0d8d569bfabd3ead0c968da265ba239b87/src/MathOptInterface/cones.jl#L636).
Hypatia does not have JuMP as a dependency, only MOI, and we would like to keep it that way if possible.

odow:
Does this PR make sense for PSDCone? It seems like everyone is a "yes."

For this PR, let's focus on what is currently here. Then, in a new PR, we make any `X >= Y, Set()` short-hand for `X - Y in Set()` and document that.

blegat (approved):

PR #2733, opened 2021-10-04
Add set inequality syntax for any set

odow:
Follows on from #2732

The open question is how to type the `extra` argument to:
```
function build_constraint(
    _error::Function,
    f::AbstractArray{<:AbstractJuMPScalar},
    s::MOI.GreaterThan,
    extra,
)
    @assert iszero(s.lower)
    return build_constraint(_error, f, extra)
end
```

The options are
 * `extra::Any`
 * `extra::Union{MOI.AbstractVectorSet,JuMP.AbstractVectorSet}`
 * `extra::Union{... explicitly choose sets}`

The advantage of the first is that it lets any set work by default. The downside is that users might write syntax that doesn't make mathematical sense, like `@constraint(model, [2x + 1, x] >= 0, MOI.Complements(2))`. 

The second option adds a bit more type checking, and prevents people doing stuff like `, MOI.Integer()`. 

The third allows us to restrict the sets to proper cones mathematically, but it requires knowing about these in JuMP, which rules out Hypatia. We could add a new trait to MOI, `is_proper`, but that's also a lot of work.

My vote is the `::Any`, with the documentation that `X >= Y, Set()` is just syntactic sugar for `X - Y in Set()`, and it's up to the user to write meaningful constraints.

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -608,13 +608,19 @@ julia> @constraint(model, X >= Y, PSDCone())
  X[2,1] - 2  X[2,2] - 1] ∈ PSDCone()
 ```
 
-The following three syntax are equivalent:
- * `@constraint(model, X >= Y, PSDCone())`
- * `@constraint(model, Y <= X, PSDCone())`
- * `@constraint(model, X - Y in PSDCone())`
+!!! tip
+    `@constraint(model, X >= Y, Set())` is short-hand for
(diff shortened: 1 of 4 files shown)


chriscoey:
Thanks! I vote first or second option please!

joaquimg:
The second with docs seems very reasonable. Could be relaxed to Any later without breaking.

PR #2734, opened 2021-10-04
Don't return MA.Zero from expressions

### LINKED ISSUE #2677: Stop returning MA.Zero() from expressions
```julia
using JuMP, MutableArithmetics
m = Model()
@variable(m, x>=0)
@constraint(m,
     [x, MutableArithmetics.Zero()] in MOI.SOS1([1.0, 2.0])
 )
 
ERROR: At REPL[20]:1: `@constraint(m, [x, MutableArithmetics.Zero()] in MOI.SOS1([1.0, 2.0]))`: unable to add the constraint because we don't recognize Any[x, MutableArithmetics.Zero()] as a valid JuMP function.
Stacktrace:
 [1] error(::String, ::String)
   @ Base ./error.jl:42
 [2] _macro_error(macroname::Symbol, args::Vector{Any}, source::LineNumberNode, str::String)
   @ JuMP ~/.julia/packages/JuMP/MIPb6/src/macros.jl:1280
 [3] (::JuMP.var"#_error#83"{Symbol, LineNumberNode})(str::String)
   @ JuMP ~/.julia/packages/JuMP/MIPb6/src/macros.jl:548
 [4] build_constraint(_error::JuMP.var"#_error#83"{Symbol, LineNumberNode}, func::Vector{Any}, set::MathOptInterface.SOS1{Float64})
   @ JuMP ~/.julia/packages/JuMP/MIPb6/src/macros.jl:375
 [5] macro expansion
   @ ~/.julia/packages/JuMP/MIPb6/src/macros.jl:608 [inlined]
 [6] top-level scope
   @ REPL[20]:1

```

My use case is not actually putting `MutableArithmetics.Zero()` into the SOS constraint, but a variable is getting set to `MutableArithmetics.Zero()` by some other means (in JuMP I assume).

odow:
Closes #2677 

This should remove a lot of issues with people reporting `MA.Zero` related problems.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1257,6 +1257,14 @@ function _throw_error_for_invalid_sense(
 end
 
 """
+    _replace_zero(x)
+
+Replaces `_MA.Zero` with a floating point `0.0`.
+"""
+_replace_zero(::_MA.Zero) = 0.0
+_replace_zero(x) = x
+
+"""
     @objective(model::Model, sense, func)
 
 Set the objective sense to `sense` and objective function to `func`. The
@@ -1322,6 +1330,9 @@ macro objective(model, args...)
     newaff, parsecode = _MA.rewrite(x)
     code = quote
         $parsecode
+        # Don't leak a `_MA.Zero` if the objective expression is an empty
+        # summation, or other structure that returns `_MA.Zero()`.
+        $newaff = _replace_zero($newaff)
         set_objective($esc_model, $sense_expr, $newaff)
         $newaff
     end
@@ -1383,6 +1394,11 @@ macro expression(args...)
         )
     end
     code = _MA.rewrite_and_return(x)
+    code = quote
+        # Don't leak a `_MA.Zero` if the expression is an empty summation, or
+        # other structure that returns `_MA.Zero()`.
+        _replace_zero($code)
+    end
     code = Containers.container_code(idxvars, indices, code, requestedcontainer)
     # don't do anything with the model, but check that it's valid anyway
     if anonvar
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -1432,6 +1432,21 @@ function test_broadcasting_variable_in_set()
     return
 end
 
+function test_MA_Zero_objective()
+    model = Model()
+    @test @objective(model, Min, sum(i for i in 1:0)) === 0.0
+    return
+end
+
+function test_MA_Zero_expression()
+    model = Model()
+    @test @expression(model, sum(i for i in 1:0)) === 0.0
+    @expression(model, expr[j=1:2], sum(i for i in j:0))
+    @test expr == [0.0, 0.0]
+    @test expr isa Vector{Float64}
+    return
+end
+
 end  # module
 
 TestMacros.runtests()


src/macros.jl
```diff
@@ -1322,6 +1330,9 @@ macro objective(model, args...)
     newaff, parsecode = _MA.rewrite(x)
     code = quote
         $parsecode
+        # Don't leak a `_MA.Zero` if the objective expression is an empty
+        # summation, or other structure that returns `_MA.Zero()`.
+        $newaff = _replace_zero($newaff)
```
blegat, line 1335: Pourquoi ajouter de la complexité dans le macro code au lieu de garder la méthode pour `set_objective_function` ?

blegat:
Why is it fixing https://github.com/jump-dev/JuMP.jl/issues/2677 ? It does not change the behavior with `@constraint`

src/macros.jl
odow: Because it returns the `newaff` below. So `obj = @objective(model, Min, sum(i for i in 1:0))` leaks.

odow:
Things work as expected in `@constraint`:

@NLaws issue in #2677 was likely that he constructed an expression, then put the expression in the constraint:
```Julia
julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x)
x

julia> @constraint(model, x <= sum(i for i in 1:0))
x ≤ 0.0

julia> @constraint(model, [sum(i * x for i in 1:0), x] in SOS1())
ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
  [1] _empty_reduce_error()
    @ Base ./reduce.jl:299
  [2] mapreduce_empty(f::Function, op::Base.BottomRF{typeof(Base.add_sum)}, T::Type)
    @ Base ./reduce.jl:342
  [3] reduce_empty(op::Base.MappingRF{var"#1#2", Base.BottomRF{typeof(Base.add_sum)}}, #unused#::Type{Int64})
    @ Base ./reduce.jl:329
  [4] reduce_empty_iter
    @ ./reduce.jl:355 [inlined]
  [5] reduce_empty_iter
    @ ./reduce.jl:354 [inlined]
  [6] foldl_impl
    @ ./reduce.jl:49 [inlined]
  [7] mapfoldl_impl
    @ ./reduce.jl:44 [inlined]
  [8] #mapfoldl#214
    @ ./reduce.jl:160 [inlined]
  [9] mapfoldl
    @ ./reduce.jl:160 [inlined]
 [10] #mapreduce#218
    @ ./reduce.jl:287 [inlined]
 [11] mapreduce
    @ ./reduce.jl:287 [inlined]
 [12] #sum#221
    @ ./reduce.jl:501 [inlined]
 [13] sum
    @ ./reduce.jl:501 [inlined]
 [14] #sum#222
    @ ./reduce.jl:528 [inlined]
 [15] sum(a::Base.Generator{UnitRange{Int64}, var"#1#2"})
    @ Base ./reduce.jl:528
 [16] macro expansion
    @ ~/.julia/packages/MutableArithmetics/8xkW3/src/rewrite.jl:279 [inlined]
 [17] macro expansion
    @ ~/.julia/dev/JuMP/src/macros.jl:745 [inlined]
 [18] top-level scope
    @ REPL[12]:1
```

odow:
Although related: is this a bug in MA? Shouldn't we be rewriting the elements of the vector?
```
julia> @macroexpand @constraint(model, [x + sum(i * x for i in 1:0), x] in SOS1())
quote
    JuMP._valid_model(model, :model)
    begin
        var"#20###272" = begin
                let
                    begin
                        var"#21###274" = (JuMP.MutableArithmetics).Zero()
                        var"#22###273" = (JuMP.MutableArithmetics).operate!((JuMP.MutableArithmetics).add_mul, var"#21###274", [x + sum((i * x for i = 1:0)), x])
                    end
                    var"#22###273"
                end
            end
        [...]
    end
end
```

blegat:
Yes, it's an improvement for now, you can rebase and merge

Issue #2246, opened 2020-06-04
`parse_one_operator_constraint`: clash with `parse_ternary_constraint` when there are four arguments


dourouc05:
Following #2051, there is a new `parse_one_operator_constraint` extension point in JuMP. However, it is not called in a very specific situation: when the constraint looks like `binpacking(a, b, c, d)`, i.e. when there is just a function call and four arguments. For instance, the current implementation works correctly with `binpacking(a, b, collect(1:10), d)`. 

In that case, JuMP jumps to `parse_ternary_constraint(_error, vectorized, :binpacking, Val(a), b, Val(c), d)`. As neither `a` nor `b` are inequality signs, it directly generates an error. 

I only found a hacky solution for this: in `parse_ternary_constraint`, intercept all error calls, ensure that there is no `parse_one_operator_constraint` to call. https://github.com/dourouc05/JuCP.jl/commit/acb90e31e21098a4239bef89e6a427f445a0cb6f

An example to trigger this error: https://github.com/dourouc05/JuCP.jl/commit/acb90e31e21098a4239bef89e6a427f445a0cb6f#diff-fce720c43af3c52c862fd7451c7374b8R312-R327

I hope someone has more imagination that I do to fix this!

odow:
Closing in favor of #2236 we should come up with a proper design and thoroughly document it.

PR #2736, opened 2021-10-07
[breaking] refactor the parse_constraint methods

### LINKED ISSUE #2236: Clean up parse_constraint
We might consider renaming `parse_constraint` in `parse_constraint_call` and `parse_constraint_head` in `parse_constraint` but this might be breaking. We can consider this renaming just before we plan to release JuMP v0.22 or v1.0.
Another option is to just remove the current `parse_constraint` and rename `parse_constraint_head` in `parse_constraint`. Indeed, implementing
```julia
parse_constraint_call(_error::Function, [...])
```
or
```julia
parse_constraint(_error::Function, ::Val{:call}, [...])
```
is not too different and it might be confusing to have many different function names while only one would be enough.

See https://github.com/JuliaOpt/JuMP.jl/pull/2228

odow:
## Background

We previously had
 * `parse_constraint_expr`
 * `parse_constraint_head`
 * `parse_one_operator_constraint`
 * `parse_ternary_constraint`
with a mix of interceptions and reorganizations.

I've simplified this to:

 * A single `parse_constraint` entry point from the macro. This is not for extending.
 * `parse_constraint_head`, which is for intercepting `Expr(head, ...)` expressions.
 * `parse_constraint_call`, which is for intercepting `Expr(:call, op, ...)` expressions.

These are now better documented, and the documentation has examples of extending via `parse_constraint_`.

## Deprecations

 * `sense_to_set` is now `operator_to_set`
 * `parse_one_operator_constraint` is now `parse_constriant_call`
 * `parse_constraint_expr` is now `parse_constraint`

## Breaking changes

 * `parse_ternary_constraint` is removed

##  TODOs

 * [x] Documentation https://jump.dev/JuMP.jl/previews/PR2736/developers/extensions/#Extend-[@constraint](@ref)
 * [x] Tests
 * [x] Deprecations

Closes #2236

### DIFF
--- a/docs/src/developers/extensions.md
+++ b/docs/src/developers/extensions.md
@@ -160,7 +160,7 @@ following methods:
     publishing any code that implements these methods.
 
 [`parse_constraint_head`](@ref) should be implemented to intercept an expression
-based on the `.head` field. For example:
+based on the `.head` field of `Base.Expr`. For example:
 ```jldoctest
 julia> using JuMP
 


odow:
@dourouc05 Here's an example of the `binpacking` from #2246 

```Julia
julia> using JuMP

julia> const MutableArithmetics = JuMP._MA;

julia> function JuMP.parse_constraint_call(
           _error::Function,
           is_vectorized::Bool,
           ::Val{:binpacking},
           args...,
       )
           @assert !is_vectorized
           new_args = MutableArithmetics.rewrite.(args)
           parse_code = quote end
           new_lhs = Expr(:vect)
           for (lhs, code) in new_args
               push!(parse_code.args, code)
               push!(new_lhs.args, lhs)
           end
           N = length(args)
           x = gensym()
           push!(parse_code.args, Expr(:(=), x, new_lhs))
           build_code = :(build_constraint($(_error),  $new_lhs,  MOI.Zeros($N)))
           return parse_code, build_code
       end

julia> model = Model(); @variable(model, x[1:4]);

julia> @constraint(model, binpacking(x[1], x[2], x[3], x[4]))
[x[1], x[2], x[3], x[4]] ∈ MathOptInterface.Zeros(4)
```

odow:
@blegat can you take a look at this?

PR #2743, opened 2021-10-11
[docs] move contributing guide to docs and expand

### LINKED ISSUE #2675: Documentation improvements
The xarray docs are very good: http://xarray.pydata.org/en/stable/index.html

Collating some things here:

 - [x] "For users" and "For developers" sections
 - [x] <s>A "Community" section</s>
   - <s>Can we have external links in the navigation of Documenter?</s> No.
 - [ ] Comprehensive contributing guide http://xarray.pydata.org/en/stable/contributing.html
 - [x] NEWS.md is part of the docs http://xarray.pydata.org/en/stable/whats-new.html
 - [x] <s>How do I... is a table http://xarray.pydata.org/en/stable/howdoi.html</s> I think the existing manual is okay

odow:
Closes #2675 

<s>I'll add a preview link once the build finishes.</s>

Preview: https://jump.dev/JuMP.jl/previews/PR2743/developers/contributing/

### DIFF
--- a/docs/src/developers/contributing.md
+++ b/docs/src/developers/contributing.md
@@ -89,7 +89,14 @@ along the lines of the following:
 The first step is to find an [open issue](https://github.com/jump-dev/JuMP.jl/issues)
 (or open a new one) for the problem you want to solve. Then, _before_ spending
 too much time on it, discuss what you are planning to do in the issue to see if
-other contributors are fine with your proposed changes.
+other contributors are fine with your proposed changes. Getting feedback early can
+improve code quality, and avoid time spent writing code that does not get merged into
+JuMP.
+
+!!! tip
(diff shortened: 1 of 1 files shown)


pulsipher:
This is a nice improvement. One suggestion would be to consider adding in a guide for using a GUI interface (e.g., VSCode) to contribute. This removes the need to use the terminal to check out environments and use git which lowers the barrier for contributors that are comfortable with creating Julia scripts, but not running shell commands. Sadly, my current workload won't permit me to make such a contribution myself.

One other suggestion, might be to mention that the tests can be run via including the `runtests.jl` file directly in an active REPL, so that they can be run multiple times much faster. Naturally, this needs to be done cautiously, but it is a nice way to debug tests quickly.

PR #2746, opened 2021-10-11
[Containers] cleanup and document the macro API

### LINKED ISSUE #2604: Public API for Building Containers in Macros
For extensions, it would be nice to provide a public API for creating JuMP containers in extension defined macros. Currently, this is accomplished in JuMP's macros via the use of `Containers._build_ref_sets` and `Containers.container_code` which is only partially public (enabling flexibility in JuMP for future macro refactoring). However, this proposed addition would be nice as part of that planned refactoring effort to better enable extension macros that use JuMP containers, but do not wish to build upon `@variable`.  This can be the case for creating extension objects where one or more of the following hold:
- the objects are not decision variables (i.e., using something `@variable(m, p in Parameter)` is not intuitive nomenclature)
- the objects do not allow the use of variable information (e.g., `VariableInfo`)
- the objects are not well amendable to the scalar build paradigm (even with the use of sets)
- the objects can be more concisely/intuitively be defined with extension defined symbolic syntax (as enabled by a macro).
This motivation even presents itself within JuMP itself with `@NLparameter`. In InfiniteOpt, this has several use cases with infinite parameter, finite parameter, and parameter function objects (see the later comments on https://github.com/pulsipher/InfiniteOpt.jl/issues/60).

I know this might not be a high priority at the moment, but I just would like to make this feature request for consideration in future refactoring efforts.

odow:
This PR tidies up the `Containers.container_code` and `Containers.build_ref_sets` code, documents it, and makes `build_ref_sets` public.

The most important part to look at is the example in the docstring of `container_code` which shows how it can be called.

https://jump.dev/JuMP.jl/previews/PR2746/reference/containers/#JuMP.Containers.container_code

cc'ing people who have (ab)used the JuMP container macro internals:
@pulsipher 
@jalving
@martinbiel 

Closes #2604

### DIFF
--- a/test/Containers/macro.jl
+++ b/test/Containers/macro.jl
@@ -94,6 +94,26 @@ using Test
             throw(err)
         end,)
     end
+    @testset "double_filter_typed_vcat" begin
+        expr =
+            :(Containers.@container(x[i = 1:2; isodd(i); iseven(i + 1)], i + i))
+        @test_throws(LoadError, try
+            @eval $expr
+        catch err
+            throw(err)
+        end,)
+    end
+    @testset "double_filter_vect" begin
+        expr = :(Containers.@container(
+            x[i = 1:2, 1:2; isodd(i); iseven(i + 1)],
+            i + i,
+        ))
+        @test_throws(LoadError, try
+            @eval $expr
+        catch err
+            throw(err)
+        end,)
+    end
     @testset "Dict" begin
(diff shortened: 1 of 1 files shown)


odow:
Merging since I don't think this is controversial (It just changes `_build_ref_sets` to `build_ref_sets`). Once merged please try out `JuMP#master` and open an issue if you have suggestions for further improvements.

Issue #1939, opened 2019-04-09
Macros should use add_variables & add_constraints rather than singular methods


ndinsmore:
I would like to advocate that the `@constraint `and `@variable` macro should be updated to use `add_constraints` and `add_variables` rather than `add_constraints.(...)` and `add_variables.(...)`.

The issue here ends up being with solvers where adding constraints one by one is very slow, which is particularly true with direct_model, on a solver like Clp.  Which with the caching bridge & https://github.com/JuliaOpt/Clp.jl/pull/57 is 40x faster in benchmarking than the direct_model benchmark.

joaquimg:
How do you think this should work in JuMP?
That would only work when adding multiple variables at once like `@variable(model, x[1:10])`?

mlubin:
This is easier to do for `@variable` than `@constraint`. `MOI.add_constraints` assumes that the constraints are homogeneous (http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#MathOptInterface.add_constraints), while `@constraint` doesn't require that.

joaquimg:
At some point we discussed that his could be implemented in LQOI, JuMP would stay the way it is and LQOI could do some caching.

Some other interesting idea would be a mixed mode:
1) Build the problem with caching optimizer
2) push everything at once with `add_variables` and `add_constraints` with a specialized copy_to
3) the start using direct mode for modifications and so on

By the way,
if we used a specilized `copy_to` from caching to linquad you problem would be solved?

ndinsmore:
It looks likes constraints already branches for the "plural version" .  Besides the names I don't see why that is actually neccisary.  Even if you are only adding one contraint you can still use `add_constraints`
https://github.com/JuliaOpt/JuMP.jl/blob/903821b15f6bc5f69486a3eec70d0004d30f01bc/src/macros.jl#L593-L598

We could always add a non-homogenous fall back to add_constraints in MOI that just uses "add_constraint.(...)" .  That might actually still be faster because the optimizer could like deal with the type union better.

ndinsmore:
No need for the speciallized copy_to, with:
https://github.com/JuliaOpt/LinQuadOptInterface.jl/pull/101
https://github.com/JuliaOpt/Clp.jl/pull/57
is already very fast by using the plural add_ methods.

joaquimg:
The specialized `copy_to` I am talking about is the one copying data from the `CachingOptimizer` to the  LinQuadModel.

[...]

PR #2748, opened 2021-10-12
WIP: use add_constraints in macros

odow:
I took a look at #1939.

It's actually fairly trivial to implement this for constraints. It's harder for variables. The key is that we should just do it for broadcasted constraints, since these are much more likely to be of the same form. And we can limit ourselves to scalar constraints.

However...

I can't find a solver with a speed-up. It seems like we've solved much of the performance problems through the ecosystem.

Here's a benchmark which gets at the difference:
```Julia
using JuMP, Gurobi
const ENV = Gurobi.Env()

function main(N)
    GC.gc()
    model = direct_model(Gurobi.Optimizer(ENV))
    @variable(model, x[1:N])
    A = rand(N, N)
    f = @expression(model, A * x)
    s = MOI.GreaterThan(0.0)
    cons = ScalarConstraint.(f, s)
    @time add_constraints(model, cons, "")
end

function main2(N)
    GC.gc()
    model = direct_model(Gurobi.Optimizer(ENV))
    @variable(model, x[1:N])
    A = rand(N, N)
    f = @expression(model, A * x)
    s = MOI.GreaterThan(0.0)
    cons = ScalarConstraint.(f, s)
    @time add_constraint.(model, cons, "")
end
```
Which yields
```julia
julia> @time main(2000);
  0.164610 seconds (16.03 k allocations: 168.573 MiB)
  1.143480 seconds (90.12 k allocations: 432.391 MiB, 14.63% gc time)

julia> @time main(2000);
  0.161069 seconds (16.03 k allocations: 168.573 MiB)
  1.120660 seconds (90.12 k allocations: 432.391 MiB, 13.44% gc time)

julia> @time main2(2000);
  0.110007 seconds (18.02 k allocations: 107.786 MiB)
  1.073805 seconds (92.11 k allocations: 371.605 MiB, 14.25% gc time)

julia> @time main2(2000);
  0.108142 seconds (18.02 k allocations: 107.786 MiB)
  1.117617 seconds (92.11 k allocations: 371.605 MiB, 17.60% gc time)
```
So fewer allocations by count, but the constructions of `moi_function.(cons)` and `moi_set.(cons)` lead to more allocations my memory size which slow things down.

We've also jigged things so that the various `copy_to` are now much faster than they were when #1939 was opened.

We can consider adding this, but I would vote to close this and #1939 and leave as-is for now.

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -587,7 +587,7 @@ function _moi_add_constraints(
     return indices
 end
 
-function add_constraints(model::Model, cons, name::String = "")
+function add_constraints(model::AbstractModel, cons, name::String = "")
     return add_constraint.(model, cons, name)
 end
 


rafabench:
Yes, my Pkg status looks like this:
```julia
[a076750e] CPLEX v0.8.0 `https://github.com/jump-dev/CPLEX.jl.git#master`
[60bf3e95] GLPK v0.15.0
[2e9cd046] Gurobi v0.10.0 `https://github.com/jump-dev/Gurobi.jl.git#master`
[4076af6c] JuMP v0.22.0 `https://github.com/jump-dev/JuMP.jl.git#od/add_constraints`
[295af30f] Revise v3.1.20
[9e70acf3] Xpress v0.14.0 `https://github.com/jump-dev/Xpress.jl.git#moi_v0.10`
```
I will run again to check if the allocations stay the same

PR #2760, opened 2021-10-18
Build docs as a PDF

### LINKED ISSUE #1881: Generate a PDF version of the docs
This should be automatically deployed to gh-pages. I wonder if other packages have done the same.

https://juliadocs.github.io/Documenter.jl/stable/man/other-formats/#PDF-Output-via-LaTeX-1

odow:
Preview: https://jump.dev/JuMP.jl/previews/PR2760/JuMP.pdf

Closes #1881 

## Needs

- [x] https://github.com/JuliaDocs/Documenter.jl/pull/1709

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -309,7 +309,7 @@ if _PDF
     # Small inconsistency between HTML and LaTeX! We want the release notes to
     # be a separate chapter in LaTeX, but we don't want an annoying drop-down in
     # HTML.
-    _PAGES[7] = "Release Notes" => "release_notes.md"
+    _PAGES[7] = "Release Notes" => ["release_notes.md"]
     latex_platform = _IS_GITHUB_ACTIONS ? "docker" : "native"
     @time Documenter.makedocs(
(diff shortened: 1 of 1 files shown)


odow:
Still some problems building the LaTeX:

 - [ ] LaTeX error around NormSpectralCone https://github.com/jump-dev/MathOptInterface.jl/blob/master/docs/src/manual/standard_form.md#matrix-cones

image

- [ ] Portfolio example 

image


- [ ] Space shuttle

image

odow:
Finally we have success: https://jump.dev/JuMP.jl/previews/PR2760/JuMP.pdf

Since things are building, this isn't blocking for JuMP 0.22. It can wait until we release a new version of MOI. I also need to tidy up the fact that each page should have a level-1 heading for the chapter (e.g., `index.md` doesn't), and there are a couple of broken links to sort.

The bigger issue is that the doc run is now 30 minutes! We may want to consider only enabling this before each release, rather than on every PR.

PR #2761, opened 2021-10-18
[docs] reorganization of tutorials

odow:
Preview: https://jump.dev/JuMP.jl/previews/PR2761/

This PR manually orders the tutorial pages, along with some renaming and recategorization. 

We now have
 * Getting started
 * Linear programs
 * Nonlinear programs
 * Conic programs
 * Algorithms
 * Applications
 
Because there are a large number of files to keep track of, I added a validator that walks the directory to check we included every `.md` file.


odow:
Merging since I think this is an improvement and I want to make some other changes without conflict issues. If you have suggestions for other improvements, please open an issue.

PR #2782, opened 2021-10-27
Export Result and Termination status codes

odow:
This PR pulls a trick with `@eval` so that:
```julia
julia> OPTIMAL
OPTIMAL::TerminationStatusCode = 1

julia> MOI.OPTIMAL === JuMP.OPTIMAL === OPTIMAL
true
```
This makes it non-breaking, since all usages of `MOI.OPTIMAL` will continue to work.

So far, I've only done this for. `ResultStatusCode` and `TerminationStatusCode`. The reason being that these are the most commonly used. The others, like `CallbackStatusCode` and `BasisStatusCode` are a bit more expert-level and indicate that something else is going on.

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -166,8 +166,8 @@ DocTestSetup = quote
     @objective(model, Max, -2x);
     optimize!(model);
     mock = unsafe_backend(model);
-    MOI.set(mock, MOI.TerminationStatus(), MOI.OPTIMAL)
-    MOI.set(mock, MOI.DualStatus(), MOI.FEASIBLE_POINT)
+    MOI.set(mock, MOI.TerminationStatus(), OPTIMAL)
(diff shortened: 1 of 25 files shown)


joaquimg:
+1 for removing:
`set_objective_sense(model, MOI.MIN_SENSE)` from the tutorial. It should be considered "advanced".

joaquimg:
We could include an explicit:
```Julia 
using MathOptInterface
const MOI = MathOptInterface
```
in the tutorials that rely on MOI. This would allow getting rid of the export a bit easier in the future.
And would highlight it is another package and not just a mystical acronym inside JuMP.
Moreover, tutorials including `MOI.` could have a link to some page briefly explaining MOI.

joaquimg:
You will also remove `MOI` in front of sets that do not require it right?

odow:
Yes to all of the above. I'll merge this for now, and then make a separate PR tidying up the documentation.

PR #2784, opened 2021-10-29
[docs] Add note on solver-specific attributes

odow:
Okay, I've re-thought some things: 

The ask in https://github.com/jump-dev/JuMP.jl/pull/2587 is actually user-error. They should use `direct_model` and everything works as expected. Trying to map solver-specific attributes through layers like the caching optimizer is just asking for trouble.

Closes https://github.com/jump-dev/JuMP.jl/pull/2587
Closes https://github.com/jump-dev/JuMP.jl/pull/2717

### DIFF
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -512,3 +512,20 @@ If you expected the solver to support your problem, you may have an error in our
 The list of available solvers, along with the problem types they support, is available at https://jump.dev/JuMP.jl/stable/installation/#Supported-solvers.
 [...]
 ```
+
+## Solver-specific attributes
+
+Some solvers expose solver-specific attributes, such as `Gurobi.ModelAttribute`.
+To access these, use [`direct_model`](@ref) in order to expose the lower-level
+[MathOptInterface API](@ref moi_documentation).
+
+For example:
+```julia
+using JuMP, Gurobi
+model = direct_model(Gurobi.Optimizer())
+@variable(model, x >= 1.5)
+@constraint(model, c, 2x + 1 == 0)
+MOI.get(model, Gurobi.ModelAttribute("IsMIP"))       # Returns 0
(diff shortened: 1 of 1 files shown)


blegat:
I would prefer an actual solution like https://github.com/jump-dev/JuMP.jl/issues/2587#issuecomment-632914674

odow:
An actual solution seems like a feature addition that is non-breaking for JuMP and MOI 1.0.

Therefore, in the interest of releasing JuMP 1.0, I propose that we merge this as-is for JuMP 1.0.

It's going to affect a very small number of users.

blegat:
We're claiming that MOI was designed exactly to solve this kind of issues. The short term solution would be for Gurobi to implement `is_set_by_optimize`.

odow:
I don't think we ever claimed that getting solver-specific attributes through bridging and caching layers was doable.

This documentation change should appease most users. We can always add more features and remove this paragraph if requested in a JuMP 1.x release.

In addition, Gurobi can change the attributes to `is_set_by_optimize` so that it works with the CachingOptimizer, but that is independent of merging this PR.

blegat:
> I don't think we ever claimed that getting solver-specific attributes through bridging and caching layers was doable.

It is doable, I don't see anything blocking that.
If Gurobi fixes its attribute then what is this paragraph for ?

odow:
There paragraph is here so that we can release JuMP 1.0 as-is. If we want to fix things later, we can remove. In general, even if we support querying solver-specific attributes through a cache, we can still recommend using direct mode.

blegat:
There is nothing to fix in JuMP, Gurobi just needs to use `is_set_by_optimize`. The user shouldn't need to worry about it, the MOI wrapper should make it work through the layers.

odow:
Okay, I'll close all of these on the JuMP side and open an issue in Gurobi.jl

PR #2786, opened 2021-10-29
[docs] Propose new tutorial on multiple solutions.

jd-foster:
This is a tutorial on finding multiple feasible solutions for a MIP combinatorial-type problem. Welcome feedback.

### DIFF
--- a/docs/src/tutorials/linear/multiple_feasible_solutions.jl
+++ b/docs/src/tutorials/linear/multiple_feasible_solutions.jl
@@ -27,11 +27,13 @@
 
 # We'll look at some objects from recreational mathematics that could be called **"symmetric square sums"**.
 # Here are a few examples (inspired by [Number Squares](https://www.futilitycloset.com/2012/12/05/number-squares/)):
+"""
     1529      2318      5219
     5837      3790      2384
     2340      1956      1867
   + ____    + ____    + ____
     9706      8064      9470
+"""
 
 # Notice how all the digits 0 to 9 are used at least once,
 # the first three rows sum to the last row,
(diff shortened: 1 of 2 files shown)


docs/src/tutorials/linear/multiple_feasible_solutions.jl
```diff
@@ -0,0 +1,178 @@
...
+# the first three rows sum to the last row,
+# and the columns in each are the same as the corresponding rows (forming a symmetric matrix).
+
+# You might ask: how many such squares are there?
+# In this tutorial, we will find out
+# and show how to go about answering this and similar questions.
+
+# This tutorial uses the following packages:
+using JuMP
+import Gurobi
```
odow, line 48: Unfortunately we can't have any tutorials with Gurobi :( They all need to run on GitHub actions.

Does CPLEX return multiple solutions yet? We could _maybe_ run that in the JuMP CI.

odow (changes requested):

docs/src/tutorials/linear/multiple_feasible_solutions.jl
jd-foster: CPLEX itself has functionality to return multiple solutions, but it's not clear to me that it's accessible via the MOI wrapper.

odow:
Talked to @jd-foster. He's going to convert this into a guest blog post for https://github.com/jump-dev/jump-dev.github.io that we can then link to.

I think this is a good work-around for content that needs commercial solvers like Gurobi that we can't install into the JuMP CI.

PR #2794, opened 2021-11-06
Adding the LP sensitivity report print function for `solution_summary`

### LINKED ISSUE #2662: LP sensitivity reports
A common request when teaching LP101-type classes is for sensitivity reports similar to Excel solver.

Here is a suggestion from someone who teaches these types of classes:

image
image

image

We already have the information available in https://jump.dev/JuMP.jl/stable/manual/solutions/#Sensitivity-analysis-for-LP, so it's really a matter of presenting it appropriately. I don't know if we want a data frames dependency, but maybe we could use the Tables.jl interface (https://github.com/JuliaData/Tables.jl) or just print it for now, similar to `solution_summary`.

<s>In fact, maybe we could just add this as additional outputs to `solution_summary` if you pass a `sensitivity = true` keyword?</s> (This was a bad suggestion by me.)

Before working on this PR, please read the comments in https://github.com/jump-dev/JuMP.jl/pull/2794 to understand a previous attempt.

AtsushiSakai:
I added the LP sensitivity report print function for `solution_summary` to fix #2662.

Currently, just a printing function is added, not using Table.jl interfaces.
This is a print example:
```
* Solver : Mock
* Status
  Termination status : OPTIMAL
  Primal status      : FEASIBLE_POINT
  Dual status        : FEASIBLE_POINT
  Message from the solver:
  "solver specific string"
* Candidate solution
  Objective value      : 2.1
  Dual objective value : 2.1
* Work counters
  Solve time (sec)   : 5.00000
* LP sensitivity summary
- Variables sensitivity
  Variable name : x
    Objective value coefficient : 1.1
    Reduced cost : 0.1
    Lower bound : -1.0
    Upper bound : 1.0
    Optimal value : 1.0
    Allowable increase in objective coefficient : Inf
    Allowable decrease in objective coefficient : 0.1
  Variable name : y
    Objective value coefficient : 1.0
    Reduced cost : -0.0
    Lower bound : 0.0
    Upper bound : Inf
    Optimal value : 1.0
    Allowable increase in objective coefficient : 0.1
    Allowable decrease in objective coefficient : 1.0
- Constraints sensitivity
  Constraint name : c1
    Right hand side (RHS) coefficient : 1.0
    Shadow price : -0.0
    RHS optimal value : 1.0
    Allowable increase in RHS coefficient  : Inf
    Allowable decrease in RHS coefficient  : Inf
  Constraint name : c2
    Right hand side (RHS) coefficient : 2.0
    Shadow price : 1.0
    RHS optimal value : 2.0
    Allowable increase in RHS coefficient  : Inf
    Allowable decrease in RHS coefficient  : 1.0
```

I'm not sure how to get `slack or surplus value` for each constraint.
If someone can advise me how to do it, I'd like to add it.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -443,48 +443,58 @@ struct _LPConstraintSensitivitySummary
 end
 
 struct _LPSensitivitySummary
-    lp_variables_sensitivity_report::Dict{VariableRef, _LPVariableSensitivitySummary}
-    lp_constraints_sensitivity_report::Dict{ConstraintRef, _LPConstraintSensitivitySummary}
+    lp_variables_sensitivity_report::Dict{
+        VariableRef,
+        _LPVariableSensitivitySummary,
+    }
+    lp_constraints_sensitivity_report::Dict{
+        ConstraintRef,
+        _LPConstraintSensitivitySummary,
+    }
 end
 
 function _get_lp_sensitivity_summary(model)
-    report = lp_sensitivity_report(model);
+    report = lp_sensitivity_report(model)
 
-    obj_value_coeff_dict = Dict{VariableRef, Float64}()
+    obj_value_coeff_dict = Dict{VariableRef,Float64}()
     for (coeff, variable) in linear_terms(objective_function(model, AffExpr))
         obj_value_coeff_dict[variable] = coeff
     end
 
-    lp_variables_sensitivity_report = Dict{VariableRef, _LPVariableSensitivitySummary}()
+    lp_variables_sensitivity_report =
+        Dict{VariableRef,_LPVariableSensitivitySummary}()
     for variable in all_variables(model)
         v_lo, v_hi = report[variable]
         if haskey(obj_value_coeff_dict, variable)
-            lp_variables_sensitivity_report[variable] = _LPVariableSensitivitySummary(
-                name(variable),
-                obj_value_coeff_dict[variable],
-                reduced_cost.(variable),
-                has_lower_bound(variable) ? lower_bound(variable) : -Inf,
-                has_upper_bound(variable) ? upper_bound(variable) :  Inf,
-                value.(variable),
-                v_hi,
-                -v_lo,
-            )
+            lp_variables_sensitivity_report[variable] =
+                _LPVariableSensitivitySummary(
+                    name(variable),
+                    obj_value_coeff_dict[variable],
+                    reduced_cost.(variable),
+                    has_lower_bound(variable) ? lower_bound(variable) : -Inf,
+                    has_upper_bound(variable) ? upper_bound(variable) : Inf,
+                    value.(variable),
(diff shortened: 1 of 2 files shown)


odow:
Have you ever used Excel? 

The reason for a table is that Excel has a sensitivity report function that is often used to teach students in a LP-101 class. 

I expect this functionality to be used very rarely in JuMP outside the classroom, so we should put it in a separate function and reproduce exactly what teachers expect: https://www.solver.com/how-correctly-interpret-sensitivity-reports-premium-solver

image

It's also going to be slow, so we don't want to run this for large LPs, and it won't work for any other type of problem, which are other reasons for putting this in a separate function.

AtsushiSakai:
> Have you ever used Excel?
Unfortunately, I have not used Excel sensitivity report.

> so we should put it in a separate function and reproduce exactly what teachers expect:
OK. Do you mean a separate function means a separate print function?

odow:
We already have 
https://github.com/jump-dev/JuMP.jl/blob/744888cde7f98c7e567f470b4fffee5826b7ec3f/src/lp_sensitivity2.jl#L9-L63

So perhaps we should implement 
```Julia
Base.show(io::IO, report::SensitivityReport)
```
or 
```Julia
Base.print(io::IO, report::SenstivityReport)
```

I'm not sure whether we should print it to the screen, or print it as a CSV for people to open in Excel. 

I've used this before: https://github.com/ronisbr/PrettyTables.jl. Not sure if we want it as a JuMP dependency though.

src/print.jl
```diff
@@ -444,16 +526,22 @@ struct _SolutionSummary
...
     simplex_iterations::Union{Missing,Int}
     node_count::Union{Missing,Int}
+    # Sensitivity
+    sensitivity::Bool
+    lp_sensitivity_summary::Union{Missing,_LPSensitivitySummary}
 end
 
 """
-    solution_summary(model::Model; verbose::Bool = false)
+    solution_summary(model::Model; verbose::Bool = false, sensitivity = false)
```
joaquimg, line 535:
```suggestion
    solution_summary(model::Model; verbose::Bool = false, lp_sensitivity = false)
```
since its only for LP

odow:
I'm going to close this as stale. 

@AtsushiSakai thanks for taking a stab. It's always useful to see different design ideas, even if they don't end up being integrated. We're still interested in something like this, but it should be a separate print-out using the SensitivityReport, and be in the form of a table, not part of `solution_summary`.

odow:
Ah, @AtsushiSakai I see you were following my suggestion in https://github.com/jump-dev/JuMP.jl/issues/2662. I guess that was bad advice. Apologies. :smile:

AtsushiSakai:
I'm sorry I left this PR. It is OK to close this PR. I will try this again when I have time.

PR #2798, opened 2021-11-11
Rewrite manual/variables.md

odow:
I think there's a lot of improvements that could be made to the Variables page in the manual:
 * The introduction is confusing, and front-loads a lot of terminology
 * We take too long to get to different parts. We should start by example, then go into detail
 * The organization could be improved. More level-2 headers should be "How to" type queries so they are easier to navigate. Currently, we have "Variable names" as a `##`, and things like "How to query a variable by name" are buried unless you read the entire docs.


docs/src/manual/variables.md
```diff
@@ -142,135 +72,181 @@ In the above examples, `x_free` represents an unbounded optimization variable,
...
+    object; it will just remove the reference at `model[:x]`.
+[...]
 ```
 
-Fixing a variable with existing bounds will throw an error. To delete the bounds
-prior to fixing, use `fix(variable, value; force = true)`.
+A common reason for encountering this error is adding variables in a loop.
 
+As a work-around, JuMP provides *anonymous* variables. Create a scalar valued
+anonymous variable by ommitting the name argument:
```
blegat, line 217:
```suggestion
anonymous variable by omitting the name argument:
```

docs/src/manual/variables.md
```diff
@@ -327,40 +303,363 @@ Something like this will not work:
...
+julia> model[:x]
+2-element Vector{VariableRef}:
+ my_var[1]
+ my_var[2]
+```
+
+## String names, symbolic names, and bindings
+
+It's common for new users to experience confusion relating to JuMP variables.
+Part of the problem the overloaded use of "variable" in mathematical
```
blegat, line 333:
```suggestion
Part of the problem is the overloaded use of "variable" in mathematical
```

docs/src/manual/variables.md
```diff
@@ -589,178 +908,82 @@ model[:x]
...
+as an optional positional argument:
 ```jldoctest; setup=:(model=Model())
 julia> @variable(model, x[1:2, 1:2], PSD)
 2×2 LinearAlgebra.Symmetric{VariableRef, Matrix{VariableRef}}:
  x[1,1]  x[1,2]
  x[1,2]  x[2,2]
 ```
-or using the syntax for [Variables constrained on creation](@ref jump_variables_on_creation):
-```jldoctest; setup=:(model=Model())
-julia> @variable(model, x[1:2, 1:2] in PSDCone())
```
blegat, line 672: Why removing this ? It should now be the prefered way, it's quite consistent with the new syntax `@constraint(model, X >= Y, PSDCone())`

docs/src/manual/variables.md
```diff
@@ -589,178 +908,82 @@ model[:x]
...
+    DenseAxisArray or a SparseAxisArray.
+
+## Symmetric variables
 
-You can also impose a weaker constraint that the square matrix is only symmetric
-(instead of positive semidefinite) as follows:
+Declare a square matrix of JuMP variables to be symmetric by passing
+`Symmetric`  as an optional positional argument:
 ```jldoctest; setup=:(model=Model())
 julia> @variable(model, x[1:2, 1:2], Symmetric)
```
blegat, line 934: This can also be done with `@constraint(model, x[1:2, 1:2] in SymMatrixSpace())`

docs/src/manual/variables.md
odow: I've moved it to the constrained on creation section. I also don't know if it's the preferred way... `@variable(model, x[1:2, 1:2], PSD)` is shorter to type. I think the `x in Set` gets a bit confusing because it blurs variables and constraints.

docs/src/manual/variables.md
odow: Added an example.

docs/src/manual/variables.md
```diff
@@ -1017,13 +1006,49 @@ julia> @constraint(model, x in SecondOrderCone())
...
+x = @variable(model, [1:3], set = SecondOrderCone())
 ```
 
 !!! note
     You cannot delete the constraint associated with a variable constrained on
     creation.
 
+### Example: positive semidefinite variables
+
+Declare a matrix of JuMP variables to be positive semidefinite
```
blegat, line 1018: The reader might wonder what's the difference with the `@variable(model, x[1:2, 1:2], PSD)` above (same for symmetric)

odow: Added a comment.

blegat (approved):

PR #2811, opened 2021-11-22
adding Base.isempty(::Model)

### LINKED ISSUE #2810: define Base.isempty or is_empty for JuMP models
When passing a model as argument to solve a problem in-place, it would be handy to know if a model is empty or not, thus to implement MOI.is_empty for JuMP models or define a JuMP function if a different meaning must be attached.

Itt could make sense to have:
`MOI.is_empty(model)` <=> model is empty (some optimizer attributes can have been set)
`JuMP.is_empty(model)` <=> model is empty and no attribute has been set. Open question is whether the optimizer itself should be absent

matbesancon:
Closes #2810

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -680,7 +680,7 @@ function Base.isempty(model::Model)
     MOI.is_empty(model.moi_backend) || return false
     isempty(model.shapes) || return false
     model.nlp_data === nothing || return false
-    isempty(mode.obj_dict) && isempty(model.ext) || return false
+    isempty(model.obj_dict) && isempty(model.ext) || return false
     return !model.is_model_dirty
 end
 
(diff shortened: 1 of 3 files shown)


src/JuMP.jl
```diff
@@ -669,6 +669,21 @@ function Base.empty!(model::Model)::Model
...
+
+Verifies whether the model is empty, that is, whether the MOI backend
+is empty and whether the model is in the same state as at its creation
+apart from optimizer attributes.
+"""
+function Base.isempty(model::Model)
+    MOI.is_empty(model.moi_backend) || return false
+    isempty(model.shapes) || return false
+    model.nlp_data === nothing || return false
+    isempty(mode.obj_dict) && isempty(model.ext) || return false
```
odow, line 683:
```suggestion
    isempty(model.obj_dict) && isempty(model.ext) || return false
```

odow:
Thanks

PR #2812, opened 2021-11-25
[docs] add design patterns tutorial

odow:
Motivated by this discussion on Discourse: https://discourse.julialang.org/t/integrating-mathoptinterface-into-an-industry-scale-project-e-g-constraint-management/71943

I think this should be pretty helpful for people wanting to create larger JuMP models. It was definitely a gap in our documentation. 

Preview link: https://jump.dev/JuMP.jl/previews/PR2812/tutorials/getting_started/design_patterns_for_larger_models/

### DIFF
--- a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
+++ b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
@@ -349,9 +349,9 @@ function solve_knapsack_6(
     config::AbstractKnapsack,
 )
     model = Model(optimizer)
-    _add_knapsack_variables(model, data, config)
-    _add_knapsack_constraints(model, data, config)
-    _add_knapsack_objective(model, data, config)
+    add_knapsack_variables(model, data, config)
+    add_knapsack_constraints(model, data, config)
+    add_knapsack_objective(model, data, config)
     optimize!(model)
     if termination_status(model) != OPTIMAL
         @warn("Model not solved to optimality")
--- a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
+++ b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  #src
 # SOFTWARE.                                                                      #src
 
-# # Design partterns for larger models
+# # Design patterns for larger models
 
 # JuMP makes it easy to build and solve optimization models. However, once you
 # start to construct larger models, and especially ones that interact with
@@ -127,7 +127,10 @@ data = KnapsackData(objects, 10.0)
 function Base.show(io::IO, data::KnapsackData)
     println(io, "A knapsack with capacity $(data.capacity) and possible items:")
     for (k, v) in data.objects
-        println(io, "  $(rpad(k, 8)) : profit = $(v.profit), weight = $(v.weight)")
+        println(
+            io,
+            "  $(rpad(k, 8)) : profit = $(v.profit), weight = $(v.weight)",
+        )
     end
     return
 end
@@ -560,11 +563,11 @@ using Test
             joinpath(@__DIR__, "data", "knapsack.json"),
             KnapsackModel.BinaryKnapsack(),
         )
-        @test isapprox(x["apple"], 1, atol=1e-5)
-        @test isapprox(x["banana"], 0, atol=1e-5)
-        @test isapprox(x["cherry"], 0, atol=1e-5)
-        @test isapprox(x["date"], 1, atol=1e-5)
-        @test isapprox(x["eggplant"], 1, atol=1e-5)
+        @test isapprox(x["apple"], 1, atol = 1e-5)
+        @test isapprox(x["banana"], 0, atol = 1e-5)
+        @test isapprox(x["cherry"], 0, atol = 1e-5)
+        @test isapprox(x["date"], 1, atol = 1e-5)
+        @test isapprox(x["eggplant"], 1, atol = 1e-5)
     end
     @testset "feasible_integer_knapsack" begin
         x = KnapsackModel.solve_knapsack(
@@ -572,11 +575,11 @@ using Test
             joinpath(@__DIR__, "data", "knapsack.json"),
             KnapsackModel.IntegerKnapsack(),
         )
-        @test isapprox(x["apple"], 0, atol=1e-5)
-        @test isapprox(x["banana"], 0, atol=1e-5)
-        @test isapprox(x["cherry"], 0, atol=1e-5)
-        @test isapprox(x["date"], 5, atol=1e-5)
-        @test isapprox(x["eggplant"], 0, atol=1e-5)
+        @test isapprox(x["apple"], 0, atol = 1e-5)
+        @test isapprox(x["banana"], 0, atol = 1e-5)
+        @test isapprox(x["cherry"], 0, atol = 1e-5)
+        @test isapprox(x["date"], 5, atol = 1e-5)
+        @test isapprox(x["eggplant"], 0, atol = 1e-5)
     end
     @testset "feasible_binary_knapsack" begin
         x = KnapsackModel.solve_knapsack(
--- a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
+++ b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
@@ -37,7 +37,8 @@
 # ## Overview
 
 # This tutorial uses explanation-by-example. We're going to start with a simple
-# knapsack model, and then expand it to add various features and structure.
+# [knapsack model](https://en.wikipedia.org/wiki/Knapsack_problem), and then
+# expand it to add various features and structure.
 
 # ## A simple script
 
@@ -56,7 +57,7 @@ model = Model(GLPK.Optimizer)
 optimize!(model)
 value.(x)
 
-# The benefit of this approach are:
+# The benefits of this approach are:
 #  * it is quick to code
 #  * it is quick to make changes.
 
@@ -96,14 +97,21 @@ solve_knapsack_1([5, 3, 2, 7, 4], [2, 8, 4, 2, 5], 10)
 # indices is fragile and a common source of bugs. A good solution is to use
 # Julia's type system to create an abstraction over your data.
 
-# For example, we can create a struct that represents a single object:
+# For example, we can create a `struct` that represents a single object, with a
+# constructor that lets us validate assumptions on the input data:
 
 struct KnapsackObject
     profit::Float64
     weight::Float64
+    function KnapsackObject(profit::Float64, weight::Float64)
+        if weight < 0
+            throw(DomainError("Weight of object cannot be negative"))
+        end
+        return new(profit, weight)
+    end
 end
 
-# as well as a struct that holds a dictionary of objects and the knapsack's
+# as well as a `struct` that holds a dictionary of objects and the knapsack's
 # capacity:
 
 struct KnapsackData
@@ -156,9 +164,9 @@ solve_knapsack_2(data)
 
 # ## Read in data from files
 
-# Having a datastructure is a good step. But it is still annoying that we have
+# Having a data structure is a good step. But it is still annoying that we have
 # to hard-code the data into Julia. A good next step is to separate the data
-# into an external file format--JSON is a common choice.
+# into an external file format; JSON is a common choice.
 
 import JSON
 
--- a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
+++ b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
@@ -358,8 +358,7 @@ solve_knapsack_5(data, BinaryKnapsack())
 # Compared to where we started, our knapsack model is now significantly
 # different. We've wrapped it in a function, defined some data types, and
 # introduced configuration options to control the variables and constraints that
-# get added. There's a few other steps we can do to further improve things. They
-# are:
+# get added. There are a few other steps we can do to further improve things:
 #  * remove the dependence on `GLPK`
 #  * add checks that we found an optimal solution
 #  * add a helper function to avoid the need to explicitly construct the data.
@@ -394,11 +393,12 @@ solution = solve_knapsack_6(GLPK.Optimizer, data_filename, BinaryKnapsack())
 
 # ## Create a module
 
-# Now we're ready to expose our model to the wider-world. That might be as part
+# Now we're ready to expose our model to the wider world. That might be as part
 # of a larger Julia project that we're contributing to, or as a stand-alone
 # script that we can run on-demand. In either case, it's good practice to wrap
 # everything in a module. This further encapsulates our code into a single
-# namespace, and we can add documentation in the form of docstrings.
+# namespace, and we can add documentation in the form of
+# [docstrings](https://docs.julialang.org/en/v1/manual/documentation/).
 
 # Some good rules to follow when creating a module are:
 # * use `import` in a module instead of `using` to make it clear which functions
@@ -450,7 +450,7 @@ struct BinaryKnapsack <: _AbstractConfiguration end
 """
     IntegerKnapsack()
 
-Create an integer knapsack problem where each object can be taken any amount of
+Create an integer knapsack problem where each object can be taken any number of
 times.
 """
 struct IntegerKnapsack <: _AbstractConfiguration end
@@ -629,7 +629,8 @@ end
 
 # !!! tip
 #     Place these tests in a separate file `test_knapsack_model.jl` so that you
-#     can run the tests by `include`ing the file.
+#     can run the tests by adding `include("test_knapsack_model.jl")` to any
+#     file where needed.
 
 # ## Next steps
 
--- a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
+++ b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
@@ -168,6 +168,18 @@ solve_knapsack_2(data)
 # to hard-code the data into Julia. A good next step is to separate the data
 # into an external file format; JSON is a common choice.
 
+# The JuMP repository [has a file](https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/tutorials/getting_started/data)
+# we're going to use for this tutorial. It is found here:
+
+data_filename = joinpath(@__DIR__, "data", "knapsack.json")
+
+# with these contents:
+
+println(read(data_filename, String))
+
+# Now let's write a function that reads this file and builds a `KnapsackData`
+# object:
+
 import JSON
 
 function read_data(filename)
@@ -181,7 +193,7 @@ function read_data(filename)
     )
 end
 
-data = read_data(joinpath(@__DIR__, "data", "knapsack.json"))
+data = read_data(data_filename)
 
 # ## Add options via if-else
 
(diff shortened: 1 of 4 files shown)


docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,596 @@
...
+#     This tutorial is more advanced than the other "Getting started" tutorials.
+#     It's in the "Getting started" section to give you an early preview of how
+#     JuMP makes it easy to structure larger models. However, if you are new to
+#     JuMP you may want to briefly skim the tutorial, and come back to it once
+#     you have written a few JuMP models.
+
+# ## Overview
+
+# This tutorial uses explanation-by-example. We're going to start with a simple
+# knapsack model, and then expand it to add various features and structure.
```
jd-foster, line 40: Add link to Wikipedia article?
`[knapsack model](https://en.wikipedia.org/wiki/Knapsack_problem)

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,596 @@
...
+    return value.(x)
+end
+
+solve_knapsack_2(data)
+
+# ## Read in data from files
+
+# Having a datastructure is a good step. But it is still annoying that we have
+# to hard-code the data into Julia. A good next step is to separate the data
+# into an external file format--JSON is a common choice.
```
jd-foster, line 161: Just a thought: TOML is in Base Julia and I've found it to be more usable in some cases.

jd-foster:
```diff
diff --git a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
index 4c704538e..36063ece5 100644
--- a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
+++ b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
@@ -56,11 +56,11 @@ model = Model(GLPK.Optimizer)
 optimize!(model)
 value.(x)
 
-# The benefit of this approach are:
+# The benefits of this approach are:
 #  * it is quick to code
 #  * it is quick to make changes.
 
-# The downsides include:
+# Some downsides include:
 #  * all variables are global (read [Performance tips](https://docs.julialang.org/en/v1/manual/performance-tips/))
 #  * it is easy to introduce errors, e.g., having `profit` and `weight` be
 #    vectors of different lengths, or not match `N`
@@ -96,14 +96,14 @@ solve_knapsack_1([5, 3, 2, 7, 4], [2, 8, 4, 2, 5], 10)
 # indices is fragile and a common source of bugs. A good solution is to use
 # Julia's type system to create an abstraction over your data.
 
-# For example, we can create a struct that represents a single object:
+# For example, we can create a `struct` that represents a single object:
 
 struct KnapsackObject
     profit::Float64
     weight::Float64
 end
 
-# as well as a struct that holds a dictionary of objects and the knapsack's
+# as well as a `struct` that holds a dictionary of objects and the knapsack's
 # capacity:
 
 struct KnapsackData
@@ -156,9 +156,9 @@ solve_knapsack_2(data)
 
 # ## Read in data from files
 
-# Having a datastructure is a good step. But it is still annoying that we have
+# Having a data structure is a good step. But it is still annoying that we have
 # to hard-code the data into Julia. A good next step is to separate the data
-# into an external file format--JSON is a common choice.
+# into an external file format; JSON is a common choice.
 
 import JSON
 

```

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,596 @@
...
+    @constraint(
+        model,
+        sum(v.weight * x[k] for (k, v) in data.objects) <= data.capacity,
+    )
+    optimize!(model)
+    return value.(x)
+end
+
+solve_knapsack_2(data)
+
```
jd-foster, line 156: Could add something like:

This approach also allows us to write functions that do error checking
on the data:
```
function check_knapsack_data(data::KnapsackData)
    for (name,object) in data.objects
         @assert object.profit >= 0
         @assert object.weight >  0
    end
end
```

odow: I added a constructor for `KnapsackObject` with an error check

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
odow: True, although this tutorial is really targeting people from other language who are considering using JuMP for larger projects, so I'd prefer we kept generic. The number of people familiar with TOML is << number of people familiar with JSON. I'd guess that if you know what TOML is and you know there's a TOML standard library, you're probably comfortable writing the equivalent `_read_data`.

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,610 @@
...
+#     JuMP makes it easy to structure larger models. However, if you are new to
+#     JuMP you may want to briefly skim the tutorial, and come back to it once
+#     you have written a few JuMP models.
+
+# ## Overview
+
+# This tutorial uses explanation-by-example. We're going to start with a simple
+# [knapsack model](https://en.wikipedia.org/wiki/Knapsack_problem), and then
+# expand it to add various features and structure.
+
```
jd-foster, line 42: Add a note on how to of how to find the "data/" folder? 
https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/tutorials/getting_started/data

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,610 @@
...
+# First, start by defining a new abstract type, as well as new subtypes for each
+# of our options.
+
+abstract type AbstractKnapsack end
+
+struct BinaryKnapsack <: AbstractKnapsack end
+
+struct IntegerKnapsack <: AbstractKnapsack end
+
+# Then, we rewrite our `solve_knapsack` function to take a `config` argument,
```
jd-foster, line 238: Calling it `config` was a little unclear to me at first: `config_type`, `knapsack_type`, `typed_knapsack`...?

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,610 @@
...
+"""
+struct BinaryKnapsack <: _AbstractKnapsack end
+
+"""
+    IntegerKnapsack()
+
+Create an integer knapsack problem where each object can be taken any amount of
+times.
+"""
+struct IntegerKnapsack <: _AbstractKnapsack end
```
jd-foster, line 439: That's dedication to the `_x` is private rule by using on your module supertype!

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,610 @@
...
+    return _KnapsackData(
+        Dict(
+            k => _KnapsackObject(v["profit"], v["weight"]) for
+            (k, v) in d["objects"]
+        ),
+        d["capacity"],
+    )
+end
+
+abstract type _AbstractKnapsack end
```
jd-foster, line 424: That's dedication to the `_x` is private rule by using on your module supertype!

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,610 @@
...
+import .KnapsackModel
+
+KnapsackModel.solve_knapsack(
+    GLPK.Optimizer,
+    joinpath(@__DIR__, "data", "knapsack.json"),
+    KnapsackModel.BinaryKnapsack(),
+)
+
+# !!! note
+#     The `.` in `.KnapsackModel` denotes that it is a submodule and not a
```
jd-foster, line 556: Great tip. Did not know this and other's likely won't either.

odow: The wording here is slightly cheating. It means "Load `@__module__.KnapsackModel`, but I wasn't sure how to express this without getting too technical. 

Note that you can do things like `import ..Foo` as well for relative imports. We use that in MOI a fair bit.

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
odow: I changed `AbstractKnapsack` to `AbstractConfiguration`. Does that help?

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
odow: Should I not? My reasoning was that the functions like `add_knapsack_variables` are private, so even if an external user created a new subtype, it wouldn't be valid for them to extend the problem.

jd-foster:
```diff
diff --git a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
index 2f014654e..440fb1096 100644
--- a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
+++ b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
@@ -345,8 +345,7 @@ solve_knapsack_5(data, BinaryKnapsack())
 # Compared to where we started, our knapsack model is now significantly
 # different. We've wrapped it in a function, defined some data types, and
 # introduced configuration options to control the variables and constraints that
-# get added. There's a few other steps we can do to further improve things. They
-# are:
+# get added. There are a few other steps we can do to further improve things:
 #  * remove the dependence on `GLPK`
 #  * add checks that we found an optimal solution
 #  * add a helper function to avoid the need to explicitly construct the data.
@@ -377,17 +376,18 @@ solution = solve_knapsack_6(GLPK.Optimizer, data_filename, BinaryKnapsack())
 
 # ## Create a module
 
-# Now we're ready to expose our model to the wider-world. That might be as part
+# Now we're ready to expose our model to the wider world. That might be as part
 # of a larger Julia project that we're contributing to, or as a stand-alone
 # script that we can run on-demand. In either case, it's good practice to wrap
 # everything in a module. This further encapsulates our code into a single
-# namespace, and we can add documentation in the form of docstrings.
+# namespace, and we can add documentation in the form of 
+# [docstrings](https://docs.julialang.org/en/v1/manual/documentation/).
 
-# Some good rules to follow whe creating a module are:
-# * Use `import` in a module instead of `using` to make it clear which functions
+# Some good rules to follow when creating a module are:
+# * use `import` in a module instead of `using` to make it clear which functions
 #   are from which packages
-# * Use `_` to start function and type names that are considered private
-# * Add docstrings to all public variables and functions.
+# * use `_` to start function and type names that are considered private
+# * add docstrings to all public variables and functions.
 
 module KnapsackModel
 
@@ -433,7 +433,7 @@ struct BinaryKnapsack <: _AbstractKnapsack end
 """
     IntegerKnapsack()
 
-Create an integer knapsack problem where each object can be taken any amount of
+Create an integer knapsack problem where each object can be taken any number of
 times.
 """
 struct IntegerKnapsack <: _AbstractKnapsack end
@@ -607,4 +607,5 @@ end
 
 # !!! tip
 #     Place these tests in a separate file `test_knapsack_model.jl` so that you
-#     can run the tests by `include`ing the file.
+#     can run the tests by adding `include("test_knapsack_model.jl")` to 
+#     any file where needed.

```

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
jd-foster: It does. But I did like the term `AbstractKnapsack` quite a bit.

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
jd-foster: It's fine (tongue in check); just very hard-core :)

jd-foster:
This is great overall. I think it takes the documentation offering to the next level.

odow:
Yeah I'm massively in favor of more involved tutorials like this. I think we can slim-down the number of existing ones in favor of fewer, more in-depth tutorials.

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,645 @@
...
+# ## Read in data from files
+
+# Having a data structure is a good step. But it is still annoying that we have
+# to hard-code the data into Julia. A good next step is to separate the data
+# into an external file format; JSON is a common choice.
+
+# The JuMP repository [has a file](https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/tutorials/getting_started/data)
+# we're going to use for this tutorial. It is found here:
+
+data_filename = joinpath(@__DIR__, "data", "knapsack.json")
```
jd-foster, line 174: This returns the path of the CI on the web. Might be work suppressing output to avoid confusion, and/or clarify.

jd-foster:
```diff
diff --git a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
index 6625534ee..5580a72bc 100644
--- a/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
+++ b/docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
@@ -169,7 +169,8 @@ solve_knapsack_2(data)
 # into an external file format; JSON is a common choice.
 
 # The JuMP repository [has a file](https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/tutorials/getting_started/data)
-# we're going to use for this tutorial. It is found here:
+# `knapsack.json`. You may want to save the file in a directory named `data`
+# that is at the same directory as the file running our example code. Then it is found here:
 
 data_filename = joinpath(@__DIR__, "data", "knapsack.json")
 

```

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,645 @@
...
+            joinpath(@__DIR__, "data", "knapsack.json"),
+            KnapsackModel.IntegerKnapsack(),
+        )
+        @test isapprox(x["apple"], 0, atol = 1e-5)
+        @test isapprox(x["banana"], 0, atol = 1e-5)
+        @test isapprox(x["cherry"], 0, atol = 1e-5)
+        @test isapprox(x["date"], 5, atol = 1e-5)
+        @test isapprox(x["eggplant"], 0, atol = 1e-5)
+    end
+    @testset "feasible_binary_knapsack" begin
```
rschwarz, line 619: Should be `"infeasible_binary_knapsack"`?

pboes: yes I believe so

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,645 @@
...
+
+# First, start by defining a new abstract type, as well as new subtypes for each
+# of our options. These types are going to control the configuration of the
+# knapsack model.
+
+abstract type AbstractConfiguration end
+
+struct BinaryKnapsack <: AbstractConfiguration end
+
+struct IntegerKnapsack <: AbstractConfiguration end
```
pboes, line 249: Would it make sense to rename the configs into "BinaryKnapsackProblem", "IntegerKnapsackProblem", "UpperBoundedKnapsackProblem"? I feel this makes it clearer that this config completely changes the type of problem.

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,645 @@
...
+#     file where needed.
+
+# ## Next steps
+
+# We've only briefly scratched the surface of ways to create and structure large
+# JuMP models, so consider this tutorial a starting point, rather than a
+# comprehensive list of all the possible ways to structure JuMP models.  If you
+# are embarking on a large project that uses JuMP, a good next step is to
+# look at ways people have written large JuMP projects "in the wild". Here are
+# some examples:
```
pboes, line 642: NOt sure whether it's bad practice to link to youtube videos, but the [PowerModels-talk](https://youtu.be/POOt1FCA8LI?t=1200) contains some slides that fit the content of the tutorial so neatly that I personally would have found it much more helpful to be taken to that excerpt from the talk rather than the repo (where given the intimidating codebase I might not even attempt to find the design patterns)

pboes:
Overall this is really fantastic, I'm very glad I asked this innocuous question and am completely blown away by the fact that within two days this should have generated a tutorial on the docs that I personally have already found super helpful. So thanks @odow, @jd-foster, @rschwarz. 

One additional section that I would probably try to add is how to use variable and constraint names (in terms of the base_name/String representation) to generate useful printouts of models. I guess those names aren't used a lot in practice and helping get a quick overview of which constraints are in action in a given model would seem like a place for them to shine. I'm happy to try and provide an example of what I have in mind when I have more time over the weekend. But since there is an example of custom printing for the data already, seeing how a model could be printed informatively would seem to be on-topic as well...

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
```diff
@@ -0,0 +1,645 @@
...
+
+# A good next step is to wrap your model in a function. This is useful for a few
+# reasons:
+#  * it removes global variables
+#  * it encapsulates the JuMP model and forces you to clarify your inputs and
+#    outputs
+#  * we can add some error checking.
+
+function solve_knapsack_1(profit::Vector, weight::Vector, capacity::Real)
+    @assert length(profit) == length(weight)
```
ericphanson, line 81: I would make this a DimensionMismatch or such because asserts shouldn’t be used for error checking (they are documented that they could be disabled).

dourouc05:
That's a great addition to the docs! It's not the only way to implement large-scale models, but it's a very good way to do it.

> I think we can slim-down the number of existing ones in favor of fewer, more in-depth tutorials.

I wouldn't go into that direction, at least from the point of view of a user: I prefer to have specific tutorials instead of large ones that present many topics that are not necessarily that related and could be dealt with separately.

odow:
> I'm very glad I asked this innocuous question and am completely blown away by the fact that within two days this should have generated a tutorial on the docs that I personally have already found super helpful

Great! I've known that this has been a stumbling block for a while, so you finally motivated me to write it. 

If you have other suggestions or questions, please fire away. We're very receptive of constructive criticism.

> to generate useful printouts of models

This is covered in various parts:
 * https://jump.dev/JuMP.jl/dev/manual/variables/#Variable-names
 * https://jump.dev/JuMP.jl/dev/manual/constraints/#Constraint-names
 * https://jump.dev/JuMP.jl/dev/manual/models/#Print-the-model

For larger models, however, printing is almost uniformly a bad idea because the printouts become huge. (For various reasons, we _don't_ print the concise algebraic form of the model.)

You could potentially do something with the `Names registered in the model:` part of the print-out, which you can programmatically access with `keys(object_dictionary(model))`. I'm not sure there is a good way to provide general advice though. It's probably up to you to perform some book-keeping in an external data-structure about what constraints have been added to the model, rather than querying the JuMP model directly.

docs/src/tutorials/getting_started/design_patterns_for_larger_models.jl
odow: What about `BinaryConfig`, `IntegerConfig`, and UpperBoundedConfig`?

We're not really dispatching on the problem type. Dispatch is just a tool to configure the model in different ways. (e.g., you could use a config to set different solver arguments, or change the printing of names).

pboes: I agree. In this case I'd personally favor `BinaryKnapsackConfig`, `IntegerKnapsackConfig` and, `UpperBoundedKnapsackConfig` but your suggestion also does the job :)

odow:
Just re-firing the CI. Looks like it got caught in the recent outage of GitHub.

odow:
Thanks for the feedback.

Issue #2814, opened 2021-11-30
Refactor print to using printing implemented for MOI backend


blegat:
Adding it in the v1.0 milestone as it might change a bit the printing so it's best to do it now.

odow:
I considered this, but I decided there were enough issues to justify duplicating the code. Particularly around nonlinear

blegat:
For issues, 1, 4 and 5, this can be resolved by having a more flexible printing at the MOI level, I had this flexibility in https://github.com/jump-dev/MathOptInterface.jl/pull/1180.
For 3, I agree, if the MOI printing is flexible enough, the JuMP printing does not have to change too much.
For 2, the nonlinear part of printing can stay at the JuMP level for the moment.

odow:
The JuMP printing code _could_ be tidied up, and leverage a lot of the `MOI.Utilities._to_string` calls, but I don't think we should make major changes. Things are working well as-is. and we haven't needed to touch the code in ages. 

Duplicating code between JuMP and MOI isn't necessarily a bad thing.

blegat:
We would need https://github.com/jump-dev/MathOptInterface.jl/pull/1695 to be applied to JuMP as well for https://github.com/jump-dev/JuMP.jl/pull/2391.
So we might as well directly print the MOI function instead of converting it to an `AffExpr` and then print it, this would also be faster.

odow:
Yes, you could use MOI to print the MOI function. Except for the `noname` issue. I don't think we really care about speed. Printing the JuMP model will never be in a critical path.

joaquimg:
minor piece of this issue,
could we print: `noname[$(vi.value)]` instead of just `noname`?

blegat:
> could we print: `noname[$(vi.value)]` instead of just `noname`?

This of the features I would like added to the printing. One may would be to make it easy from MOI to determine how unnamed variable are printed, e.g., via a function as I did in https://github.com/jump-dev/MathOptInterface.jl/pull/1180. Then JuMP would give by default `_ -> noname` but the user could pass a function to JuMP that is passed to MOI to overwrite that. I usually want to give `vi -> "x[vi.value]"` except when I have variables `x` then I name it differently. This shows that the user needs to specify that since there is no good default that works for all use cases.

odow:
So I went through the printing file carefully in #2820.

[...]

PR #2815, opened 2021-12-01
WIP: refactor printing to use MOI.Utilities

odow:
Part of #2814 

Just to see what things look like. Tests are failing because I got up to the extension part.

Overall, I'm not convinced this is worth the breakage, particularly for extensions. If we were to make a change, we should get rid of `IJuliaMode` and `REPLMode` in favor of actual MIME types.

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -491,7 +491,7 @@ Alias for `GenericAffExpr{Float64,VariableRef}`, the specific
 const AffExpr = GenericAffExpr{Float64,VariableRef}
 
 # Check all coefficients are finite, i.e. not NaN, not Inf, not -Inf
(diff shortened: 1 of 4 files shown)


blegat:
I think it's nice to have a consistent printing between JuMP and MOI, it's consistent with the fact that JuMP is MOI + macros. If JuMP has a good reason to print something differently then it may be worth it to add an option to `MOI.Utilities._PrintOptions`.

> If we were to make a change, we should get rid of `IJuliaMode` and `REPLMode` in favor of actual MIME types.

I would support that

odow:
Discussed on monthly meeting. We should continue to explore ways to refactor the JuMP printing, but be mindful not to break printing of JuMP extensions.

odow:
Closing in favor of #2820

PR #2822, opened 2021-12-08
Rename printing of anonymous JuMP variables

odow:
Part of #2814 

It's always annoyed me that you can't tell anonymous variables apart in printing:
```Julia
julia> model = Model();

julia> x = @variable(model, [1:4])
4-element Vector{VariableRef}:
 noname
 noname
 noname
 noname

julia> @objective(model, Min, sum(i * x[i] for i in 1:4))
noname + 2 noname + 3 noname + 4 noname
```

This PR changes things to:
```julia
julia> model = Model();

julia> x = @variable(model, [1:4])
4-element Vector{VariableRef}:
 _[1]
 _[2]
 _[3]
 _[4]

julia> @objective(model, Min, sum(i * x[i] for i in 1:4))
_[1] + 2 _[2] + 3 _[3] + 4 _[4]
```

It's an open question why the prefix should be:
 * `noname[3]`: this would be most similar to what we already have, and no one should name a variable `noname`?
 * `anon[3]`: relates to the fact it is anonymous. But requires people to know the abbreviation. someone might name a variable `anon`?
 * `_[3]`: the shortest. But requires people to be familiar with `_` as a placeholder for a variable.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -436,6 +436,14 @@ function _nl_subexpression_string(print_mode, model::Model)
     return strings
 end
 
+anonymous_name(::Any, x::AbstractVariableRef) = "noname"
+
+anonymous_name(::Type{REPLMode}, x::VariableRef) = "_[$(index(x).value)]"
+
+function anonymous_name(::Type{IJuliaMode}, x::VariableRef)
+    return "\\_\\[$(index(x).value)\\]"
+end
+
 """
     function_string(
         print_mode::Type{<:JuMP.PrintMode},
@@ -448,7 +456,7 @@ Return a `String` representing the function `func` using print mode
 function function_string(::Type{REPLMode}, v::AbstractVariableRef)
     var_name = name(v)
     if isempty(var_name)
-        return "noname"
+        return anonymous_name(REPLMode, v)
     end
     return var_name
 end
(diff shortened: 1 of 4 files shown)


blegat:
What is used in LaTeX if we use `_`?

src/print.jl
```diff
@@ -436,6 +436,14 @@ function _nl_subexpression_string(print_mode, model::Model)
     return strings
 end
 
+anonymous_name(::Any, x::AbstractVariableRef) = "noname"
+
+anonymous_name(::Type{REPLMode}, x::VariableRef) = "_[$(index(x).value)]"
+
+function anonymous_name(::Type{IJuliaMode}, x::VariableRef)
+    return "\\_\\[$(index(x).value)\\]"
```
odow, line 444: @blegat we escape the underscore, so it'll get printed as `_[1]`

blegat: Probably it should be `{\\_}_{$(index(x).value)}` so that it prints as an index like other variables ?

odow:
@mlubin and @joaquimg thoughts?

joaquimg:
I am fine with all options: `noname[3]`, `anon[3]` and `_[3]`.
I usually like adding functionality like:
`set_anonimous_variable_string(model, "anon")`
Then I would go for `noname` by default since it's the clearest for the user.
If someone likes another string just use `set_anonimous_variable_string`.

odow:
> set_anonymous_variable_string(model, "anon")

I'm not in favor of this. It just adds complexity. 

`_` is probably the safest because it's a convention in Julia (and other languages) to indicate a variable that isn't used, like 

`_, x = foo()`. 

But yeah, no one has complained about `noname`, so we could stick with that.

mlubin:
If we keep `noname`, it should at least be `no_name` for better readability. I think `no_name`, `anon`, and `_` are all reasonable.

odow:
Okay, I've gone with `anon` for extensions, and left `_[i]` for `VariableRef`.

PR #2827, opened 2021-12-12
[Containers] disallow CartesianIndex as a key for DenseAxisArray

### LINKED ISSUE #2825: CartesianIndex as keys in a DenseAxisArray
Reported on SO: https://stackoverflow.com/questions/70284721/julia-jump-how-to-use-variable-with-cartesianindex/70296103

There's some weird things that happen:
```Julia
julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> S = CartesianIndex.([2, 4])
2-element Vector{CartesianIndex{1}}:
 CartesianIndex(2,)
 CartesianIndex(4,)

julia> @variable(model, x[S])
    Dimension 1, CartesianIndex{1}[CartesianIndex(2,), CartesianIndex(4,)]
And data, a 2-element Vector{VariableRef}:
 x[CartesianIndex(2,)]
 x[CartesianIndex(4,)]

julia> x[CartesianIndex(2,)]
x[CartesianIndex(4,)]

julia> x[CartesianIndex(1,)]
x[CartesianIndex(2,)]

julia> x[CartesianIndex(4,)]
ERROR: BoundsError: attempt to access 2-element Vector{VariableRef} at index [4]
Stacktrace:
 [1] getindex
   @ ./array.jl:801 [inlined]
 [2] getindex
   @ ./multidimensional.jl:637 [inlined]
 [3] getindex(A::JuMP.Containers.DenseAxisArray{VariableRef, 1, Tuple{Vector{CartesianIndex{1}}}, Tuple{JuMP.Containers._AxisLookup{Dict{CartesianIndex{1}, Int64}}}}, idx::CartesianIndex{1})
   @ JuMP.Containers ~/.julia/packages/JuMP/2IF9U/src/Containers/DenseAxisArray.jl:323
 [4] top-level scope
   @ REPL[42]:1
```

odow:
Technically breaking, but it has buggy behavior at present, so this is a bug-fix. 

We don't want cartesian indices as indices. They're not intended for that purpose. The original poster on stack overflow should use a different approach: https://stackoverflow.com/questions/70284721/julia-jump-how-to-use-variable-with-cartesianindex/70296103

Closes #2825

### DIFF
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -159,7 +159,7 @@ function _abstract_vector(x::AbstractVector)
     return x
 end
 
-_abstract_vector(x) = _abstract_vector([a for a in x])
+_abstract_vector(x) = _abstract_vector(vec([a for a in x]))
 
 function _abstract_vector(x::Number)
     @warn(
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -148,18 +148,17 @@ function Base.get(
     return get(x.data, key[1] => key[2], default)
 end
 
-function _abstract_vector(x::AbstractVector)
-    if eltype(x) <: CartesianIndex
-        error(
-            "Unsupported index type `CartesianIndex` in axis: $x. Cartesian " *
-            "indices are restricted for indexing into and iterating over " *
-            "multidimensional arrays.",
-        )
-    end
-    return x
+_abstract_vector(x::AbstractVector) = x
+
+function _abstract_vector(x::AbstractVector{<:CartesianIndex})
+    return error(
+        "Unsupported index type `CartesianIndex` in axis: $x. Cartesian " *
+        "indices are restricted for indexing into and iterating over " *
+        "multidimensional arrays.",
+    )
 end
 
-_abstract_vector(x) = _abstract_vector(vec([a for a in x]))
+_abstract_vector(x) = _abstract_vector([a for a in x])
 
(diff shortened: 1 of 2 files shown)


src/Containers/DenseAxisArray.jl
```diff
@@ -148,17 +148,26 @@ function Base.get(
     return get(x.data, key[1] => key[2], default)
 end
 
-_abstract_vector(x::AbstractVector) = x
+function _abstract_vector(x::AbstractVector)
```
blegat, line 151: Can't you do that with dispatch `_abstract_vector(::AbstractVector{<:CartesianIndex}) = error...`

src/Containers/DenseAxisArray.jl
```diff
@@ -148,17 +148,26 @@ function Base.get(
...
+            "Unsupported index type `CartesianIndex` in axis: $x. Cartesian " *
+            "indices are restricted for indexing into and iterating over " *
+            "multidimensional arrays.",
+        )
+    end
+    return x
+end
 
-_abstract_vector(x) = [a for a in x]
+_abstract_vector(x) = _abstract_vector(vec([a for a in x]))
```
blegat, line 162: What is `vec` here for ?

odow: I don't remember. I've removed it to see which test broke.

odow: I added a test. It was for when the index set is an AbstractArray, not an AbstractVector.

PR #2831, opened 2021-12-17
Add all_nl_constraints

odow:
Surprisingly missing. It was hard to get the NLP duals without this.

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -181,7 +181,8 @@ const NonlinearConstraintRef = ConstraintRef{Model,NonlinearConstraintIndex}
 """
     all_nl_constraints(model::Model)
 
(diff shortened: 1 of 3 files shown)


src/nlp.jl
```diff
@@ -178,6 +178,18 @@ end
 
 const NonlinearConstraintRef = ConstraintRef{Model,NonlinearConstraintIndex}
 
+"""
+    all_nl_constraints(model::Model)
+
+Return a vector of all nonlinear constraint references in the model in the
+correct order.
```
mlubin, line 185: "the correct order" is ambiguous. It's better to say, "in the order in which they were added to the model".

mlubin (approved):

PR #2843, opened 2022-01-04
Documentation improvement: Some suggestions for making it, IMHO, somewhat more clear.

ignace-computing:
This suggestion was inspired by Mathieu Tanneau's answer to one of my questions on Slack.

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -79,7 +79,7 @@ set_optimizer_attributes(model, "LogLevel" => 1, "PrimalTolerance" => 1e-7)
     repository for more details on how to use a particular solver, including any
     solver-specific options.
 
-### Supported solvers
+## Supported solvers (and applicability for problems)
 
 Most solvers are not written in Julia, and some require commercial licenses to
 use, so installation is often more complex.
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -79,7 +79,7 @@ set_optimizer_attributes(model, "LogLevel" => 1, "PrimalTolerance" => 1e-7)
     repository for more details on how to use a particular solver, including any
     solver-specific options.
 
-## Supported solvers (and applicability for problems)
+## Supported solvers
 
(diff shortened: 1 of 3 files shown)


docs/src/installation.md
```diff
@@ -79,7 +79,7 @@ set_optimizer_attributes(model, "LogLevel" => 1, "PrimalTolerance" => 1e-7)
     repository for more details on how to use a particular solver, including any
     solver-specific options.
 
-### Supported solvers
+## Supported solvers (and applicability for problems)
```
odow, line 82:
```suggestion
## Supported solvers
```
So we can make this a level-2 header, but changing the title name will break all links to the current header.

docs/src/tutorials/nonlinear/portfolio.jl
```diff
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  #src
 # SOFTWARE.                                                                      #src
 
-# # Portfolio optimization
+# # Portfolio optimization (Quadratic Program (QP))
```
odow, line 21: How about
```suggestion
# # Quadratic portfolio optimization
```

odow (commented):
Just a couple of things.

odow:
@ignace-computing what do you think of this now?

ignace-computing:
Yeah that looks fine to me!

odow:
Thanks for the contribution. "Supported solvers" now shows up in the left navigation bar:
image
https://jump.dev/JuMP.jl/dev/installation/#Supported-solvers

Please open an issue or a pull request if you have any other suggestions for improving the docs. I wonder if we could rethink the table to have more columns like LP, Integer, QP, NLP, Conic. But then it quickly gets out of hand and would become quite wide. (What about SOS, indicator, and complementarity constraints? SOC, RSOC, and EXP cones? Native support? Or any support via bridges?)

PR #2844, opened 2022-01-04
Pluralized macros return tuples

### LINKED ISSUE #2837: Make plural macros (e.g. @variables) return collections of the defined objects
When working on a large model it is extremely useful to be able to manipulate large groups of variables or constraints.

To avoid repeating long lists of variable names or equations, it would be very useful if the the ‘plural macros’ (e.g. @variables and @constraints), returned collections of the newly defined objects.

For example I would like for

```
vars = @variables model begin
  x
  y[1:2]
end
```

to be equivalent to writing

```
@variables model begin
  x
  y[1:2]
end
vars = [x, y]
```
and

```
eqs = @NLconstraints model begin
  E_x, x == 1
  E_y[i ∈ 1:2], y[i] == x
end
```
should be equivalent to writing

```
@NLconstraints model begin
  E_x, x == 1
  E_y[i ∈ 1:2], y[i] == x
end
eqs = [E_x, E_y]
```

Note that this is not just a matter of laziness, but that repeating long lists of items is prone to hard-to-debug errors. I.e. spotting which item is missing in a large collection of variables or constraints can take a long time.

MartinBonde:
Closes #2837
Pluralized macros (e.g. @variables) now return vectors of whatever the 'singular' macros return.

E.g. the following new tests pass:
    vars = @variables(model, begin
        x
        y[1:2]
    end)
    @test vars == [x, y]
    eqs = @constraints(model, begin
        E_x, x == 0
        E_y[i=1:2], y[i] == 0
    end)
    @test eqs == [E_x, E_y]

### DIFF
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -858,12 +858,12 @@ function test_Plural_returns()
         x
         y[1:2]
     end)
-    @test vars == [x, y]
+    @test vars == (x, y)
     eqs = @constraints(model, begin
         E_x, x == 0
         E_y[i=1:2], y[i] == 0
     end)
-    @test eqs == [E_x, E_y]
+    @test eqs == (E_x, E_y)
     return
 end
 
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -861,7 +861,7 @@ function test_Plural_returns()
     @test vars == (x, y)
     eqs = @constraints(model, begin
         E_x, x == 0
-        E_y[i=1:2], y[i] == 0
+        E_y[i = 1:2], y[i] == 0
     end)
     @test eqs == (E_x, E_y)
     return
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -852,6 +852,21 @@ function test_Plural_failures()
     return
 end
 
+function test_Plural_returns()
+    model = Model()
+    vars = @variables(model, begin
+        x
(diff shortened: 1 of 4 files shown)


mlubin:
Tuples seem like a better choice than vectors in this case. They'll give the compiler a chance at inferring the types of the elements if the situation permits. See https://discourse.julialang.org/t/when-to-use-tuples/18675.

MartinBonde:
I have no issue with using tuples instead.

test/macros.jl
```diff
@@ -852,6 +852,21 @@ function test_Plural_failures()
...
+function test_Plural_returns()
+    model = Model()
+    vars = @variables(model, begin
+        x
+        y[1:2]
+    end)
+    @test vars == (x, y)
+    eqs = @constraints(model, begin
+        E_x, x == 0
+        E_y[i=1:2], y[i] == 0
```
odow, line 864:
```suggestion
        E_y[i = 1:2], y[i] == 0
```

odow:
The documentation is failing because there are a lot of new printing changes now that `nothing` is no longer returned. 

The easiest step is probably to add a semi-colon to the plural usages that appear in the documentation.

odow:
Thanks @MartinBonde

PR #2847, opened 2022-01-05
Add start value for constraints

### LINKED ISSUE #2835: Implement set_start_value for constraint indices
Should it be `set_start_value` or `set_primal_start_value` given that for the dual it's `set_dual_start_value` and for variables it is `set_start_value`  ?

cc @kalmarek

odow:
Closes #2835

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -133,7 +133,7 @@ end
     set_start_value(con_ref::ConstraintRef, value)
 
 Set the primal start value (MOI attribute `ConstraintPrimalStart`) of the
-constraint `con_ref` to `value`. To remove a dual start value set it to
+constraint `con_ref` to `value`. To remove a primal start value set it to
 `nothing`.
 
 See also [`start_value`](@ref).
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -172,7 +172,7 @@ function set_start_value(
     },
     value,
 )
-    MOI.set(owner_model(con_ref), MOI.ConstraintDualStart(), con_ref, value)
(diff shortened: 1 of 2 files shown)


src/constraints.jl
```diff
@@ -129,6 +129,71 @@ function set_dual_start_value(
     return
 end
 
+"""
+    set_start_value(con_ref::ConstraintRef, value)
+
+Set the primal start value (MOI attribute `ConstraintPrimalStart`) of the
+constraint `con_ref` to `value`. To remove a dual start value set it to
```
joaquimg, line 136:
```suggestion
constraint `con_ref` to `value`. To remove a primal start value set it to
```

src/constraints.jl
```diff
@@ -129,6 +129,71 @@ function set_dual_start_value(
     return
 end
 
+"""
+    set_start_value(con_ref::ConstraintRef, value)
+
+Set the primal start value (MOI attribute `ConstraintPrimalStart`) of the
```
blegat, line 135: Can we reference to `ConstraintPrimalStart` since MOI is part of JuMP doc now ?

src/constraints.jl
```diff
@@ -129,6 +129,71 @@ function set_dual_start_value(
...
+        con_ref,
+        vectorize(value, con_ref.shape),
+    )
+    return
+end
+
+function set_start_value(
+    con_ref::ConstraintRef{
+        <:AbstractModel,
+        <:_MOICON{<:MOI.AbstractVectorFunction,<:MOI.AbstractVectorSet},
```
blegat, line 160: Why is this one only for vector constraints ? What happens if you set nothing for scalar constraints ?

PR #2858, opened 2022-01-31
Add warning and improve docs for modify-then-query

### LINKED ISSUE #2797: Remove dirty flag
The dirty flag added in https://github.com/jump-dev/JuMP.jl/pull/2709 removes some flexibility for the MOI backend as it assumes that setting an attributes always modifies the solution. However, for DiffOpt, you can set attributes that are the gradient and ask for forward and backward differentiation. And you can set a different gradient and do differentiation again, without the need to reoptimize.
I'm tempted to say that we should revert https://github.com/jump-dev/JuMP.jl/pull/2709 and rely on MOI model to implement `MOI.TerminationStatus` consistently. The consistency being check by MOI tests as we've done with the rest of the API.
Another way to make this work is to add a function in MOI `does_setting_this_attributes_requires_a_resolve` so that DiffOpt can implement a method that returns `false`.

odow:
Closes #2797

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -1234,22 +1234,30 @@ Return the value of the attribute `attr` from the model's MOI backend.
 function MOI.get(model::Model, attr::MOI.AbstractModelAttribute)
     if !MOI.is_set_by_optimize(attr)
         return MOI.get(backend(model), attr)
-    elseif attr isa MOI.TerminationStatus
-        if model.is_model_dirty && mode(model) != DIRECT
-            return MOI.OPTIMIZE_NOT_CALLED
-        end
-        return MOI.get(backend(model), attr)
-    elseif attr isa MOI.PrimalStatus || attr isa MOI.DualStatus
-        if model.is_model_dirty && mode(model) != DIRECT
-            return MOI.NO_SOLUTION
-        end
-        return MOI.get(backend(model), attr)
-    else
-        if model.is_model_dirty && mode(model) != DIRECT
-            throw(OptimizeNotCalled())
-        end
-        return _moi_get_result(backend(model), attr)
+    elseif model.is_model_dirty && mode(model) != DIRECT
+        @warn(
+            "The model has been modified since the last call to `optimize!`. " *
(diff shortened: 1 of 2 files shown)


odow:
Objections to merging?

src/JuMP.jl
```diff
@@ -1234,22 +1234,30 @@ Return the value of the attribute `attr` from the model's MOI backend.
...
-        end
-        return MOI.get(backend(model), attr)
-    else
-        if model.is_model_dirty && mode(model) != DIRECT
-            throw(OptimizeNotCalled())
-        end
-        return _moi_get_result(backend(model), attr)
+    elseif model.is_model_dirty && mode(model) != DIRECT
+        @warn(
+            "The model has been modified since the last call to `optimize!`. " *
```
joaquimg, line 1239: is it possible that optimize was never called? if so, the message would be slightly missleading

odow: Yes, this will also get called if you as for `value(x)` before calling `optimize!`. Should we amend to 
```
The model has been modified since the last call to `optimize!` (or `optimize!` has not been called yet).
```

joaquimg: good! +1 for the improved text.

src/JuMP.jl
```diff
@@ -1234,22 +1234,30 @@ Return the value of the attribute `attr` from the model's MOI backend.
...
-        return MOI.get(backend(model), attr)
-    else
-        if model.is_model_dirty && mode(model) != DIRECT
-            throw(OptimizeNotCalled())
-        end
-        return _moi_get_result(backend(model), attr)
+    elseif model.is_model_dirty && mode(model) != DIRECT
+        @warn(
+            "The model has been modified since the last call to `optimize!` (" *
+            "or `optimize!` has not been called yet. If you are iteratively " *
```
odow, line 1240:
```suggestion
            "or `optimize!` has not been called yet). If you are iteratively " *
```

joaquimg (approved):

Issue #2862, opened 2022-02-05
Debug TSP example on Julia 1.7


odow:
Per @jd-foster on Gitter:

<img width="913" alt="image" src="https://user-images.githubusercontent.com/8177701/152661740-e6e49098-d187-4384-8988-34086ebc2bc6.png">

danielschermer:
I believe that this might be related to the underlying Optimizer and how it handles lazy constraints (or the interface in between). Using Gurobi as an alternative Optimizer, no duplicate cycles are detected (and the lazy method scales much better). I checked which cycles are detected through:

`println("Node $(GLPK.glp_ios_curr_node(cb_data.tree)): Found cycle of length $(length(cycle)) $cycle")`

Using GLPK, for the value of `n=22` (for smaller values of `n` it seems to work fine, i.e., no duplicate cycles) the iterated model finds the following cycles:

```
Found cycle of length 3: [16, 10, 8]
Found cycle of length 4: [1, 14, 11, 2]
Found cycle of length 4: [7, 21, 13, 4]
```

The lazy constraint methods finds the following cycles. Note that identical cycles are detected at different nodes in the tree.

```
Node 5: Found cycle of length 3 [16, 10, 8] <- as in iterated
Node 18: Found cycle of length 3 [12, 22, 3]
Node 22: Found cycle of length 3 [1, 14, 2]
Node 22: Found cycle of length 4 [7, 21, 13, 4] <- as in iterated
Node 22: Found cycle of length 3 [17, 18, 9]
Node 22: Found cycle of length 4 [5, 20, 6, 15]
Node 22: Found cycle of length 5 [1, 14, 11, 19, 2]
Node 30: Found cycle of length 3 [1, 11, 2]
Node 30: Found cycle of length 3 [16, 10, 8] <- should not appear again
Node 30: Found cycle of length 4 [7, 21, 13, 4] <- should not appear again
Node 31: Found cycle of length 3 [16, 10, 8] <- should not appear again
Node 43: Found cycle of length 3 [12, 22, 3] <- should not appear again
Node 48: Found cycle of length 4 [7, 21, 13, 4] <- should not appear again
Node 27: Found cycle of length 3 [16, 10, 8] <- should not appear again
Node 48: Found cycle of length 4 [7, 21, 13, 4] <- should not appear again
Node 29: Found cycle of length 3 [16, 10, 8] <- should not appear again
[...]

PR #2863, opened 2022-02-07
Make Random seed more explicit

jd-foster:
Reproduce Julia v1.6 behaviour in v1.7+ (address #2862 )

### DIFF
--- a/docs/src/tutorials/algorithms/tsp_lazy_constraints.jl
+++ b/docs/src/tutorials/algorithms/tsp_lazy_constraints.jl
@@ -117,7 +117,7 @@ import Plots
 # thus, the weight (distance) of each edge is defined as follows.
 
 function generate_distance_matrix(n; random_seed = 1)
-    Random.seed!(Random.MersenneTwister(),random_seed)
+    Random.seed!(Random.MersenneTwister(), random_seed)
     X = 100 * rand(n)
     Y = 100 * rand(n)
     d = [sqrt((X[i] - X[j])^2 + (Y[i] - Y[j])^2) for i in 1:n, j in 1:n]
--- a/docs/src/tutorials/algorithms/tsp_lazy_constraints.jl
+++ b/docs/src/tutorials/algorithms/tsp_lazy_constraints.jl
@@ -117,7 +117,7 @@ import Plots
 # thus, the weight (distance) of each edge is defined as follows.
 
 function generate_distance_matrix(n; random_seed = 1)
-    Random.seed!(random_seed)
+    Random.seed!(Random.MersenneTwister(),random_seed)
     X = 100 * rand(n)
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/algorithms/tsp_lazy_constraints.jl
```diff
@@ -117,7 +117,7 @@ import Plots
 # thus, the weight (distance) of each edge is defined as follows.
 
 function generate_distance_matrix(n; random_seed = 1)
-    Random.seed!(random_seed)
+    Random.seed!(Random.MersenneTwister(),random_seed)
```
odow, line 120:
```suggestion
    Random.seed!(Random.MersenneTwister(), random_seed)
```

odow:
Did you try this on Julia 1.7? Did it work?

jd-foster:
Ah, a space was the issue for for formatter...
Yes, this seems to solve the issue in Julia 1.7.

jd-foster:
Actually, since it's random there is no reproducible behaviour here anyway regardless of Julia version. But at least it's explicit.

odow:
> Actually, since it's random there is no reproducible behaviour here anyway regardless of Julia version

What do you mean?

jd-foster:
Just that my "testing" by running the code for v1.6 and v1.7 seemed to come up with long-running examples either way.

docs/src/tutorials/algorithms/tsp_lazy_constraints.jl
```diff
@@ -117,7 +117,7 @@ import Plots
 # thus, the weight (distance) of each edge is defined as follows.
 
 function generate_distance_matrix(n; random_seed = 1)
-    Random.seed!(random_seed)
+    Random.seed!(Random.MersenneTwister(1), random_seed)
     X = 100 * rand(n)
     Y = 100 * rand(n)
```
odow, line 122: Doesn't this need to be:
```julia
    rng = Random.MersenneTwister(random_seed)
    X = 100 * rand(rng, n)
    Y = 100 * rand(rng, n)
```
Otherwise I don't think it's reproducible.

jd-foster: Yes, you're right.

jd-foster: Not sure how I broke the format-check though.

odow: I'm fixing that in #2865. Not sure why either. Must be a transitive dependency change?

odow: Seems to be fixed.

odow (approved):
LGTM. Assuming CI passes

PR #2872, opened 2022-02-14
[docs] add CasADi to Should I use JuMP?

### LINKED ISSUE #2871: [docs] Mention CasADi as an alternative to JuMP
Mention CasADi[.jl] as an alternative in https://jump.dev/JuMP.jl/dev/should_i_use/.

Context: https://discourse.julialang.org/t/jump-hessian-sparcity-issues-registered-functions-hessians-atan2-issue/76342/5

odow:
Closes #2871

### DIFF
--- a/docs/src/should_i_use.md
+++ b/docs/src/should_i_use.md
@@ -93,6 +93,18 @@ Alternatives to consider are:
  * [GalacticOptim.jl](https://github.com/SciML/GalacticOptim.jl)
  * [NLopt.jl](https://github.com/JuliaOpt/NLopt.jl)
 
+### Optimal control problems
+
+JuMP supports formulating optimal control problems as large nonlinear programs
+(see, for example, [Optimal control for a Space Shuttle reentry trajectory](@ref)).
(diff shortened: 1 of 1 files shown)


mlubin:
@ChrisRackauckas, are there any pointers you'd like us to add here?

pulsipher:
It might be worth mentioning InfiniteOpt.jl as an alternative for optimal control. Its nonlinear interface is more flexible than JuMP's and it can model optimal control problems in continuous time. Plus, some more optimal control specific features are currently under development and should be released in a month or two. I am also working on an interface to ModelingToolkit.jl to facilitate model transfer, automatic initialization/simulation, and model simplification (e.g., order reduction).

pulsipher:
Also, albeit with a more limited NLP interface, Pyomo.dae is probably worth mentioning.

odow:
I added InfiniteOpt and pyomo.DAE.

odow:
Merging for now. If @ChrisRackauckas has other suggestions, we can make a follow-up PR.

PR #2874, opened 2022-02-18
[breaking] remove PrintMode instead of explicit MIME types

### LINKED ISSUE #2814: Refactor print to using printing implemented for MOI backend

odow:
Closes #2814 

This is breaking for a few JuMP extensions. But it should just be a small find-and-replace of 
 * `::Type{REPLMode}` -> `::MIME"text/plain"`
 * `REPLMode` -> `MIME("text/plain")`
 *  `::Type{IJuliaMode}` -> `::MIME"text/latex"`
 * `IJuliaMode` -> `MIME("text/latex")`.

Using MIME types make it clearer what is happening (we use IJuliaMode in more than just IJulia), and it's also more extensible to other printing types in future.

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -1403,7 +1403,7 @@ function Base.show(io::IO, wrapper::_ParameterPrintWrapper)
         end
     end
     # No named parameter; use a generic name.
-    if wrapper.mode == IJuliaMode
+    if wrapper.mode == MIME("text/latex")
         print(io, "parameter_{$(p.index)}")
     else
         print(io, "parameter[$(p.index)]")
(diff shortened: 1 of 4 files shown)


odow:
There are quite a few current usages of this: https://juliahub.com/ui/Search?q=IJuliaMode&type=code

 * InfiniteOpt @pulsipher 
 * JuMPeR
 * MPSGE @davidanthoff 
 * ParameterJuMP
 * Plasmo @jalving
 * PolyJuMP
 * SetProg
 * StochasticPrograms @martinbiel 
 * SumOfSquares

This will be the only breaking change for JuMP 1.0. It should be pretty trivial to change. We decided against a more wholesale change of JuMP's printing machinery.

blegat (approved):

PR #2879, opened 2022-02-21
Add nl_dual_start_value

### LINKED ISSUE #2875: Expose NLPBlockDualStart
It doesn't seem like JuMP provides a nice way to set dual starting values for nonlinear constraints. See https://github.com/jump-dev/MathOptInterface.jl/issues/1746.

odow:
Closes #2875

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -236,9 +236,40 @@ end
 Set the value of the MOI attribute [`MOI.NLPBlockDualStart`](@ref).
 
 The start vector corresponds to the Lagrangian duals of the nonlinear
-constraints, in the order given by [`all_nl_constraints`](@ref).
+constraints, in the order given by [`all_nl_constraints`](@ref). That is, you
+must pass a single start vector corresponding to all of the nonlinear
+constraints in a single function call; you cannot set the dual start value of
+nonlinear constraints one-by-one. The example below demonstrates how to use
+[`all_nl_constraints`](@ref) to create a mapping between the nonlinear
+constraint references and the start vector.
 
 Pass `nothing` to unset a previous start.
+
+## Examples
+
+```jldoctest
+julia> model = Model();
+
+julia> @variable(model, x[1:2]);
+
+julia> nl1 = @NLconstraint(model, x[1] <= sqrt(x[2]));
+
+julia> nl2 = @NLconstraint(model, x[1] >= exp(x[2]));
+
+julia> start = Dict(nl1 => -1.0, nl2 => 1.0);
+
+julia> start_vector = [start[con] for con in all_nl_constraints(model)]
+2-element Vector{Float64}:
+ -1.0
+  1.0
+
+julia> set_nl_dual_start_value(model, start_vector)
+
+julia> nl_dual_start_value(model)
+2-element Vector{Float64}:
+ -1.0
+  1.0
+```
 """
 function set_nl_dual_start_value(model::Model, start::Vector{Float64})
     _init_NLP(model)
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -218,6 +218,44 @@ function dual(c::NonlinearConstraintRef)
     return nldata.nlconstr_duals[c.index.value]
 end
 
+"""
+    nlp_dual_start_value(model::Model)
+
+Return the current value of the MOI attribute [`MOI.NLPBlockDualStart`](@ref).
+"""
+function nlp_dual_start_value(model::Model)
+    return MOI.get(model, MOI.NLPBlockDualStart())
+end
+
+"""
+    set_nlp_dual_start_value(
+        model::Model,
+        start::Union{Nothing,Vector{Float64}},
+    )
+
+Set the value of the MOI attribute [`MOI.NLPBlockDualStart`](@ref). Pass
+`nothing` to unset a previous start.
+"""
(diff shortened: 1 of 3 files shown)


test/nlp.jl
```diff
@@ -1170,4 +1170,18 @@ end
...
+
+    @testset "dual_start_value" begin
+        model = Model()
+        @variable(model, x)
+        @variable(model, t >= 0)
+        @NLconstraint(model, 2x >= 1)
+        @NLconstraint(model, t <= sqrt(x))
+        @test nlp_dual_start_value(model) === nothing
+        set_nlp_dual_start_value(model, [1.0, -1.0])
+        @test nlp_dual_start_value(model) ==[1.0, -1.0]
```
odow, line 1182:
```suggestion
        @test nlp_dual_start_value(model) == [1.0, -1.0]
```

src/nlp.jl
```diff
@@ -218,6 +218,44 @@ function dual(c::NonlinearConstraintRef)
...
+end
+
+"""
+    set_nlp_dual_start_value(
+        model::Model,
+        start::Union{Nothing,Vector{Float64}},
+    )
+
+Set the value of the MOI attribute [`MOI.NLPBlockDualStart`](@ref). Pass
+`nothing` to unset a previous start.
```
mlubin, line 237: This could use an explanation of what the vector means and how to map it to nonlinear constraints that have been added to the model.

mlubin (commented):
Should these be named `nl_dual_start_value` and `set_nl_dual_start_value`? Do we use "nlp" in other names?

odow:
Hmm. We're slightly inconsistent in our {nl,NL}(p) naming scheme. 

We have:
 - num_nl_constraints
 - add_NL_constraint
 - NLPEvaluator

`nl_dual_start_value` would be fine by me. And we could consider renaming the `_NL_` functions to `_nl_`?

mlubin:
At the risk of bikeshedding...

"nl" is not a standard abbreviation. What if we replaced it with "nonlinear", e.g.,
`nonlinear_dual_start_value`, `num_nonlinear_constraints`, `add_nonlinear_constraint`?

`NLPEvaluator` is justified because of `MOI.AbstractNLPEvaluator`, and we don't want to tag MOI 2.0 because of this.
`@NLconstraint` is ok for legacy reasons and because macros are somewhat special.

odow:
It's not bikeshedding if the BDFL does it. 

I'm okay with being explicit. Does it warrant a deprecation release (i.e., JuMP 0.23), or a hard break clearly documented in the 1.0 upgrading guide?

mlubin:
> I'm okay with being explicit. Does it warrant a deprecation release (i.e., JuMP 0.23), or a hard break clearly documented in the 1.0 upgrading guide?

Depends on what we want the messaging to be for upgrading. Another option is add error messages from the old names pointing to the new names.

odow:
We're already going to have one hard break:  https://github.com/jump-dev/JuMP.jl/pull/2874

> Another option is add error messages from the old names pointing to the new names.

This might be the best option. We could remove the old names in a future 1.x release.

For this PR though, let's stick with `nl_dual_start` and we can rename things to be consistent in another PR.

mlubin (approved):

src/nlp.jl
```diff
@@ -218,6 +218,48 @@ function dual(c::NonlinearConstraintRef)
...
+"""
+    set_nl_dual_start_value(
+        model::Model,
+        start::Union{Nothing,Vector{Float64}},
+    )
+
+Set the value of the MOI attribute [`MOI.NLPBlockDualStart`](@ref).
+
+The start vector corresponds to the Lagrangian duals of the nonlinear
+constraints, in the order given by [`all_nl_constraints`](@ref).
```
mlubin, line 239: This description is a bit confusing. Let's say I added nonlinear constraints. After reading this I might think I have to query `all_nl_constraints` and construct some mapping to my constraints. It would be helpful to be more explicit about how to construct this vector given nonlinear constraint references.

odow: > After reading this I might think I have to query all_nl_constraints and construct some mapping to my constraints

You do have to do this. I've added an example to the docstring.

src/nlp.jl
```diff
@@ -236,9 +236,40 @@ end
...
+
+julia> @variable(model, x[1:2]);
+
+julia> nl1 = @NLconstraint(model, x[1] <= sqrt(x[2]));
+
+julia> nl2 = @NLconstraint(model, x[1] >= exp(x[2]));
+
+julia> start = Dict(nl1 => -1.0, nl2 => 1.0);
+
+julia> start_vector = [start[con] for con in all_nl_constraints(model)]
```
mlubin, line 261: Do you think it would be simpler to say: 
```
start_vector = zeros(num_nl_constraints(model))
start_vector[nl1.value] = -1
start_vector[nl2.value] = 1
```

odow: I generally shied away from theses `.value` accesses. I also don't know if they're part of the public API, which documenting like this would suggest.

In general, the user will need to maintain a separate mapping from constraint -> start value, so they'd probably write something like
```julia
for (con, start) in dict
    start_vector[con.value] = start
end
```
anyway. I don't see anyone hard-coding the start values like you have.

mlubin: Ok, sounds good.

PR #2881, opened 2022-02-21
[docs] switch from GLPK to HiGHS

### LINKED ISSUE #2878: [docs] update MILP examples to HiGHS
All MILP examples in the JuMP documentation use GLPK. The original motivation for this was that it was an open-source MILP solver that supported callbacks. However, it has the down-side of being the first solver that new users see when they read the docs.

We should switch all MILP solves except the callback examples to HiGHS because it's faster, more robust, and being actively developed. I've just tagged a 1.0 release, and the underlying solver is now v1.2, installable via JLLs and passing all tests in MOI. It also has QP support which is nice.

odow:
Closes #2878

TODOs
- [x] check the preview. https://jump.dev/JuMP.jl/previews/PR2881


odow:
@jajhall HiGHS is now the main solver we use in the JuMP documentation for MILPs.

PR #2883, opened 2022-02-21
[docs] update roadmap for post v1.0

odow:
Preliminary sketch. We should discuss in a monthly developer call.


odow:
@pulsipher suggests: refactoring the macro code

PR #2887, opened 2022-02-22
Fix passing OptimizerAttributes to set_optimizer

### LINKED ISSUE #2886: set_optimizer does not copy optimizer attributes
```Julia
julia> using JuMP, Clp

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x)
x

julia> set_silent(model)

julia> set_optimizer(model, Clp.Optimizer)

julia> optimize!(model)
Clp3002W Empty problem - 0 rows, 1 columns and 0 elements
Clp0000I Optimal - objective value 0
Clp0032I Optimal objective 0 - 0 iterations time 0.002
```

Found by @anavarro01 in https://github.com/odow/SDDP.jl/issues/495, but I initially mis-diagnosed it as the Gurobi printing issue (sorry).

`set_optimizer` constructs a new caching optimizer: https://github.com/jump-dev/JuMP.jl/blob/1e0b6ee82bfec5e4fe210b70caff47610607e81c/src/optimizer_interface.jl#L103-L121

but it is missing the logic we previous had to copy supported OptimizerAttributes:
https://github.com/jump-dev/MathOptInterface.jl/blob/5e4cd36b5477d7f2a241edfed1e186848992bd71/src/Utilities/cachingoptimizer.jl#L150-L164

odow:
Closes jump-dev/MathOptInterface.jl#1750

### DIFF
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -119,6 +119,9 @@ function set_optimizer(
     # using the existing `model_cache`.
     model.moi_backend =
         MOI.Utilities.CachingOptimizer(backend(model).model_cache, optimizer)
+    # reset_optimizer is called here so that we copy all copyable and supported
+    # optimizer attributes to the new optimizer.
+    MOI.Utilities.reset_optimizer(model.moi_backend, optimizer)
     return
 end
 
--- a/test/model.jl
+++ b/test/model.jl
@@ -813,6 +813,29 @@ function test_copy_conflict()
     @test "cref[1]" == @inferred JuMP.name(cref_1_new)
 end
 
+function test_copy_optimizer_attributes()
+    function optimizer()
+        return MOI.Utilities.MockOptimizer(
+            MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
+        )
+    end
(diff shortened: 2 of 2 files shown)


src/optimizer_interface.jl
```diff
@@ -119,6 +119,9 @@ function set_optimizer(
     # using the existing `model_cache`.
     model.moi_backend =
         MOI.Utilities.CachingOptimizer(backend(model).model_cache, optimizer)
```
blegat, line 121: The copy of the optimizer attributes should be done by the `CachingOptimizer` constructor. It is checking that the `optimizer` is empty to make sure that it is in sync:
https://github.com/jump-dev/MathOptInterface.jl/blob/5e4cd36b5477d7f2a241edfed1e186848992bd71/src/Utilities/cachingoptimizer.jl#L97
but since `attach_optimizer` does not copy the optimizer attributes, then the constructor should copy them otherwise they are not in sync.

odow:
Closing in favor of https://github.com/jump-dev/MathOptInterface.jl/pull/1749

PR #2891, opened 2022-02-24
Prep for v0.23

odow:
As discussed on the monthly call, v0.23 is being tagged as a RC for JuMP 1.0.

### DIFF
--- a/docs/src/release_notes.md
+++ b/docs/src/release_notes.md
@@ -1,6 +1,54 @@
 # Release notes
 
-## Version 0.22.3 (February 10, 2021)
+## Version 0.23.0 (February 25, 2022)
+
+**JuMP v0.23.0 is a breaking release. It is also a release-candidate for JuMP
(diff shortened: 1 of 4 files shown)


odow:
I've added a blog post here: https://github.com/jump-dev/jump-dev.github.io/pull/72, which we can announce via discourse.

docs/src/release_notes.md
```diff
@@ -1,6 +1,54 @@
 # Release notes
 
-## Version 0.22.3 (February 10, 2021)
+## Version 0.23.0 (February 25, 2022)
+
+**JuMP v0.23.0 is a breaking release. It is also a release-candidate for JuMP
+v1.0.0. That is, if no issues are found with the v0.23.0 release, then it will
+be re-tagged as v1.0.0.**
+
+- Breaking changes:
```
mlubin, line 9: Should we mention the switch to MOI 1.0?

odow: Done.

mlubin (approved):

PR #2893, opened 2022-02-27
Deprecate nl_expr_string and nl_constraint_string

### LINKED ISSUE #2892: Inconsistent Nonlinear Naming
With `v0.23` most public functions with `nl` or `NL` were renamed with `nonlinear`. However, `nl_expr_string` and `nl_constraint_string` remain. https://github.com/jump-dev/JuMP.jl/blob/9b4847a79323ff83797ca91ced2b6b03578c7d05/src/print.jl#L369-L390

I use these in my nonlinear interface tests in InfiniteOpt.jl as an easy way to check that I am building nonlinear expressions in JuMP as I expect: https://github.com/pulsipher/InfiniteOpt.jl/blob/e06a1ae57f34139593e4551a8d65e8ba7e99824a/test/TranscriptionOpt/model.jl#L510

odow:
Closes #2892

These methods shouldn't really be called by external users of JuMP because they rely on private methods and people hacking through the internals of `model.nlp_data`. I wonder if I should even make them `_nonlinear_expr_string`.

https://github.com/pulsipher/InfiniteOpt.jl/blob/797e6d931a02ee7808f2e636422e7ecb6a2baa1d/test/TranscriptionOpt/transcribe.jl#L327

I wonder if @pulsipher could instead use;
```Julia
julia> model = Model();

julia> @variable(model, x);

julia> @NLexpression(model, expr, sin(x));

julia> @NLconstraint(model, c, expr <= 1);

julia> nl_constraint_string(model, MIME("text/plain"), model.nlp_data.nlconstr[end])
"subexpression[1] - 1.0 ≤ 0"

julia> sprint(show, c)
"subexpression[1] - 1.0 ≤ 0"
```

And Plasmo is calling these with all manner of private functions (cc @jalving):
https://github.com/zavalab/Plasmo.jl/blob/c4f048b76fdb7230de728f7275f1f585a1c2945f/src/nlp_macro.jl#L82-L124

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -312,18 +312,26 @@ function objective_function_string(mode, model::Model)
     if nlobj === nothing
         return function_string(mode, objective_function(model))
     end
-    return nl_expr_string(model, mode, nlobj)
+    return nonlinear_expr_string(model, mode, nlobj)
 end
 
 """
-    nl_constraint_string(model::Model, mode::MIME, c::_NonlinearConstraint)
+    nonlinear_constraint_string(
+        model::Model,
+        mode::MIME,
+        c::_NonlinearConstraint,
+    )
 
 Return a string representation of the nonlinear constraint `c` belonging to
 `model`, given the `mode`.
 """
-function nl_constraint_string(model::Model, mode::MIME, c::_NonlinearConstraint)
+function nonlinear_constraint_string(
+    model::Model,
+    mode::MIME,
+    c::_NonlinearConstraint,
+)
     s = _sense(c)
-    nl = nl_expr_string(model, mode, c.terms)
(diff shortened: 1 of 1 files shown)


jalving:
The Plasmo `nlp_macro.jl` file is purely there to make `@NLconstraint` and `@NLobjective` work on a node (which wraps a `JuMP.Model`). It is very hacky, but I couldn't think of a better way to do it at the time. I haven't kept up with the NLP changes; is there a better way to do this now?

pulsipher:
Yes, I could get away with just using `sprint` for my use case.

odow:
Okay, I'll merge this for now, but please consider ways to avoid calling these functions. We'll keep them exported, and I'll add them to my list of things to consider in a NLP rewrite.

PR #2899, opened 2022-02-28
Fix conj and implement real/imag

blegat:


odow (commented):
Hmm. Now we're getting a bit out over our skis. How much of a complex interface are we committing to implementing?

blegat:
I don't think we need much more. @ccoffrin You can try out complex support with AC-OPF problem problems. Use ComplexOptInterface v0.1, MOI v1.0.1 and the JuMP branch of this PR.

ccoffrin:
Some preliminary feedback from a very quick test of AC-OPF in complex QCQP form.

Needs:
- `println(model)` currently broken when constraints include the type of `ComplexF64`, hard to tell if the model is building correctly.
- some solver for testing complex QCQP; bridges to Ipopt seems like a clear path.
- better error messages (two examples below)

Would be Helpful:
- support for complex equality constants (i.e. `@constraint(model, (xr + xi*im) == (dr + di*im)*(yr + yi*im))` where x, y are variables, d is data.  My current work around is to break into two constraints, one for real and one for imag. (this has a hard to follow error message)
- support for `abs`, `abs2`, and `angle` on complex numbers (https://docs.julialang.org/en/v1/manual/complex-and-rational-numbers/)
- JuMP variables representing complex numbers, so that I don't need to do `(xr + xi*im)` in many places.  Rectangular form is the natural choice, polar would also be useful.
- While writing the model I got this error often `ERROR: LoadError: syntax: missing comma or ) in argument list`, it seems to a macro parsing issue around the use of `im`.  Explicitly adding multiplication resolved the issue (e.g. `xi*im`) but the error message could be clearer.

blegat:
> `println(model)` currently broken when constraints include the type of `ComplexF64`, hard to tell if the model is building correctly.
> support for complex equality constants (i.e. `@constraint(model, (xr + xi*im) == (dr + di*im)*(yr + yi*im))` where x, y are variables, d is data.  My current work around is to break into two constraints, one for real and one for imag. (this has a hard to follow error message)

Both of these should be fixed by https://github.com/jump-dev/JuMP.jl/pull/2901

I would guess

ccoffrin:
`abs2` of affine expressions will be sufficient for the Complex QCQP of AC-OPF.

Re MWE.  Now that I look deeper it seems to be a Julia parsing issue, just my lack of understanding if `im` syntax, 
- `1 + 2im` -> works
- `1 + 2 im` -> `ERROR: syntax: extra token "im" after end of expression`
- `(1 + 2im)` works
- `(1 + 2 im)` -> `ERROR: syntax: missing comma or ) in argument list`
- `(1 + 2*im)` works

PR #2900, opened 2022-02-28
Fix solution_summary with empty model

odow:
Calling `solution_summary` on an empty model exposed a few issues.

Closes https://github.com/jump-dev/MathOptInterface.jl/pull/1747

### DIFF
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -202,5 +202,8 @@ Return the number of results available to query after a call to
 [`optimize!`](@ref).
 """
 function result_count(model::Model)::Int
+    if termination_status(model) == MOI.OPTIMIZE_NOT_CALLED
+        return 0
+    end
     return MOI.get(model, MOI.ResultCount())
 end
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -77,7 +77,7 @@ julia> solution_summary(model, verbose=true)
     y[a] : 1.00000e+00
     y[b] : 1.00000e+00
   Dual solution :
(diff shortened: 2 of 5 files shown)


src/optimizer_interface.jl
```diff
@@ -202,5 +202,8 @@ Return the number of results available to query after a call to
 [`optimize!`](@ref).
 """
 function result_count(model::Model)::Int
+    if termination_status(model) == MOI.OPTIMIZE_NOT_CALLED
+        return 0
```
blegat, line 206: At the MOI interface, I would still expect solvers to return 0 and not crash in case this is called before optimize

odow: `ResultCount` is set by optimize. Therefore, it naturally fits one of those attributes that may error if called before `optimize!`. At the MOI level, users should check `termination_status` to see if the solver has run. We can be more user-friendly at the JuMP level.

Also, in our test case, we don't have a solver, so we need this check here. Alternatively, we could do something in `solution_summary`.

blegat: Indeed, it's consistent with `_moi_get_result`.

blegat (approved):

odow:
That segfault is annoying. 

@kalmarek @bodono: SCS v3.2.0 was a breaking release compared to SCS v3.0.0, so we're getting a segfault. A temporary fix is to pin the version in SCS.jl: https://github.com/jump-dev/SCS.jl/pull/252

kalmarek:
@odow I think the vial solution for us is to keep SCS jlls pinned to the exact versions we know.

Issue #2910, opened 2022-03-05
NLP Interface Error Message with Registered Functions


ccoffrin:
This is a case where the user makes an implementation error and the message from JuMP could be improved.

In the following example the user means to put `@NLobjective(m, Min, f2(x[1],x[2]))` in the second model but uses `@NLobjective(m, Min, f(x[1],x[2]))` instead.  This results in a MethodError deep in JuMP's NLP implementation.  Oddly, if the function `f` is sufficiently simple the issue does not occur.

MWE
```
using JuMP
using Ipopt

# works without a problem
# function f(x...)
#     return x[1]^2 + x[2]^2 + 2
# end

# a function of this style of complexity is required to produce the issue
function f(x...)
    mf = Model(Ipopt.Optimizer)

    @variable(mf, y[1:2])
    @NLobjective(mf, Max, y[1]*x[1] + y[2]*x[2] - x[1]*y[1]^4 - 2*x[2]*y[2]^4)
    @constraint(mf, (y[1]-10)^2 + (y[2]-10)^2 <= 25)
    optimize!(mf)

    return objective_value(mf)
end

m = Model(Ipopt.Optimizer)
@variable(m, x[1:2]>=0)
JuMP.register(m, :f, 2, f, autodiff=true)
@NLobjective(m, Min, f(x[1],x[2]))
optimize!(m)


function f2(x...)
    return x[1]^2 + x[2]^2 + 2
end

m = Model(Ipopt.Optimizer)
@variable(m, x[1:2]>=0)
JuMP.register(m, :f2, 2, f2, autodiff=true)
# NOTE intentional bug to produce the error, user puts `f` instead of `f2`
@NLobjective(m, Min, f(x[1],x[2]))
optimize!(m)
```

In Julia v1.7, JuMP v0.23.1 I get,
```
RROR: LoadError: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{JuMP.var"#132#134"{typeof(f)}, Float64}, Float64, 2})
...
Stacktrace:
  [1] convert(#unused#::Type{Float64}, x::ForwardDiff.Dual{ForwardDiff.Tag{JuMP.var"#132#134"{typeof(f)}, Float64}, Float64, 2})
    @ Base ./number.jl:7
[...]

PR #2911, opened 2022-03-09
Check for ForwardDiff compatibility in register

odow:
Part of #2910


odow:
@ccoffrin how's this?

ccoffrin:
I did not have a chance to test it, but at a quick review looks good to me!

PR #2916, opened 2022-03-10
Fix scoping issue in user-defined Containers

odow:
We'd discussed this before, https://github.com/jump-dev/JuMP.jl/issues/2523, and I thought I fixed it, but I missed a scoping issue. 

Raised on discourse: https://discourse.julialang.org/t/using-alternative-array-container-types-in-jump/77707

### DIFF
--- a/test/Containers/macro.jl
+++ b/test/Containers/macro.jl
@@ -165,3 +165,42 @@ using Test
         @test x[(2, 2), 3] == 7
     end
 end
+
+struct _MyContainer end
+
+function Containers.container(f::Function, indices, ::Type{_MyContainer})
+    key(i::Tuple) = i
+    key(i::Tuple{T}) where {T} = i[1]
+    return Dict(key(i) => f(i...) for i in indices)
+end
+
+@testset "_MyContainer" begin
+    Containers.@container(v[i = 1:3], sin(i), container = _MyContainer)
+    @test v isa Dict{Int,Float64}
+    @test length(v) == 3
+    @test v[2] ≈ sin(2)
+    Containers.@container(w[i = 1:3, j = 1:3], i + j, container = _MyContainer)
+    @test w isa Dict{Tuple{Int,Int},Int}
+    @test length(w) == 9
+    @test w[2, 3] == 5
+    Containers.@container(
+        x[i = 1:3, j = [:a, :b]],
+        (j, i),
+        container = _MyContainer
+    )
+    @test x isa Dict{Tuple{Int,Symbol},Tuple{Symbol,Int}}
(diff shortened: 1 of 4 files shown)


mlubin:
Do we document the interface that external containers need to satisfy for this to work?

odow:
No, this is essentially undocumented because it's not something we really want to advertise: https://discourse.julialang.org/t/using-alternative-array-container-types-in-jump/77707/3. 

We should instead encourage people to create their own containers outside of JuMP, rather than extending things in this way.

odow:
For example, https://jump.dev/JuMP.jl/stable/manual/variables/#User-defined-containers

mlubin:
Why support it at all if we don't want to document it?

odow:
I'd be okay documenting this in the Developer/Extensions section. 

We introduced it with the view to unifying the named-array landscape in Julia, https://github.com/jump-dev/JuMP.jl/issues/2523, but we have some unique requirements that I don't know if it makes sense to do so.

The main distinction is that some package developers might want to extend things, but general users probably shouldn't.

mlubin:
Documenting it in the Developer/Extensions section would be fine. What error message will users get if they try a container that hasn't implemented the Containers method?

odow:
It's currently (after this PR)
```Julia
julia> struct Foo end;

julia> @variable(model, x[S], Bin, container = Foo)
ERROR: Unable to build a container with the provided type Foo. Implement `Containers.container`.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] container(#unused#::Function, #unused#::JuMP.Containers.VectorizedProductIterator{Tuple{Vector{String}}}, D::Type)
   @ JuMP.Containers ~/.julia/dev/JuMP/src/Containers/container.jl:176
 [3] macro expansion
   @ ~/.julia/dev/JuMP/src/macros.jl:142 [inlined]
 [4] top-level scope
   @ REPL[21]:1
```

TODOs
 - [ ] Add documentation
 - [ ] Improve error message

mlubin:
Ok, that's already a decent error message as is. I was worried about MethodErrors.

test/Containers/macro.jl
```diff
@@ -147,7 +147,7 @@ using Test
     @testset "Invalid container" begin
         err = ErrorException(
             "Unable to build a container with the provided type $(Int). " *
-            "Implement `Containers.container`.",
+            "Implement `Containers.container(::Function, indices, ::Type{Int})`.",
```
odow, line 150:
```suggestion
            "Implement `Containers.container(::Function, indices, ::Type{$Int})`.",
```

PR #2918, opened 2022-03-11
Add RelativeGap to solution summary

### LINKED ISSUE #2906: It could be helpful to have the `MOI.RelativeGap(model)` reported in the `solution_summary`
I understand that this only makes sense if there are discrete variables. However, it's an important parameter to have when solving discrete optimization problems.

odow:
Closes #2906

### DIFF
--- a/src/solution_summary.jl
+++ b/src/solution_summary.jl
@@ -120,11 +120,7 @@ function _show_candidate_solution_summary(io::IO, summary::_SolutionSummary)
         "  Objective bound      : ",
         summary.objective_bound,
     )
-    _print_if_not_missing(
-        io,
-        "  Objective gap        : ",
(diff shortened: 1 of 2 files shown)


src/solution_summary.jl
```diff
@@ -118,6 +120,7 @@ function _show_candidate_solution_summary(io::IO, summary::_SolutionSummary)
         "  Objective bound      : ",
         summary.objective_bound,
     )
+    _print_if_not_missing(io, "  Objective gap        : ", summary.relative_gap)
```
joaquimg, line 123: It should be clear that its relative and not absolute

odow: Yeah I renamed to "Relative gap"

Issue #2920, opened 2022-03-12
Consolidating logos and branding


mlubin:
Thanks to https://github.com/jump-dev/JuMP.jl/pull/2584 we have a "modern" Julia-inspired JuMP logo. We still have some traces of old logos and branding that should be updated:

- [x] https://numfocus.org/project/jump [Emailed NF]
- [x] NumFOCUS JuMP drive (not public) [Emailed NF]
- [x] https://twitter.com/jumpjl
- [x] JuMP.jl GitHub "social preview"
- [X] https://www.gurobi.com/resource/add-ons/ (JuliaOpt!) [Sent email to Gurobi]

Is https://github.com/jump-dev/JuMP.jl/tree/master/docs/src/assets the canonical source of the new logos?
Do we have a version without the "JuMP" text? (Needed for NumFOCUS and Twitter)

@cormullion

joaquimg:
The github logo is also a no text logo.

mlubin:
> The github logo is also a no text logo.

I see the JuMP-dev org has a "new" no-text logo, but not in a vector format. How was it generated?

cormullion:
This gist is the source - all you need is Julia.😀

mlubin:
Thanks, @cormullion. The script is helpful but it doesn't exactly reproduce [logo-with-text.svg](https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/assets/logo-with-text.svg).

Current logo:


![image](https://user-images.githubusercontent.com/1733683/158032931-c8abb774-2553-4663-a4d3-9343e34ed1f7.png)


Generated by https://gist.github.com/cormullion/42977b9bca4c7d08a6455865297c6d9e:

![image](https://user-images.githubusercontent.com/1733683/158032911-6d2313a1-fedd-40f5-be1b-db098cf5d0c5.png)

Differences I see are:
- Thickness of the line segments ("chopsticks") around the circles
- Different black for the text and line segments
- Size of the canvas and centering

The first two differences are most important in keeping the logo consistent.

Peeking into the svg, the code generating the line segments is different:
Current logo:
```
<path d="M41 75.5l100-40M25 127.5l116-116" fill="none" stroke-width="8" stroke-linecap="round" stroke="#333" stroke-miterlimit="10"/>
```

Generated by the gist:
```
[...]

PR #2924, opened 2022-03-14
[docs] Add script to reproduce logos

odow:
Part of #2920 

I think I manually cleaned up the current logos, but it's probably better to have them programmatically generated, even if it leaves the svgs slightly larger than they need to be.

Some of the numeric values changed because I adjusted the margins slightly to center the logos.

Here are the current ones:

| | Dark | Light |
| ----- | ------ | ------ |
| with-text-background | logo-dark-with-text-background | logo-with-text-background |
| with-text | logo-dark-with-text | logo-with-text |
| logo | logo-dark | logo |
| without-text | logo-dark-without-text | logo-without-text |


docs/src/assets/logo_constructor.jl
```diff
@@ -0,0 +1,310 @@
+using Luxor
```
joaquimg, line 1: Maybe annotate version since there ia no project toml for this?

odow:
I've updated the description with a table of the svgs.

mlubin (approved):

PR #2929, opened 2022-03-23
Fix bug in broadcasting SparseAxisArray

### LINKED ISSUE #2919: Bug in vectorized constraints with `SparseAxisArrays`
I think the following should work (but is not currently supported):

```julia
@variable(model, u[i = 1:2, j = i:3])
Containers.@container(c[i = 1:2, j = i:3],rand())
@constraint(model, u .<= c)

ERROR: MethodError: no method matching _indices()
Closest candidates are:
  _indices(::JuMP.Containers.SparseAxisArray, ::Any...) at C:\Users\HD\.julia\packages\JuMP\lnUbA\src\Containers\SparseAxisArray.jl:217
  _indices(::Any, ::Any...) at C:\Users\HD\.julia\packages\JuMP\lnUbA\src\Containers\SparseAxisArray.jl:215
```

If the indices match, shouldn't this be allowed?

odow:
Closes #2919

### DIFF
--- a/src/Containers/SparseAxisArray.jl
+++ b/src/Containers/SparseAxisArray.jl
@@ -179,10 +179,7 @@ end
 function Base.copy(
     bc::Base.Broadcast.Broadcasted{BroadcastStyle{N,K}},
 ) where {N,K}
-    dict = Dict(
-        index => bc.f(_get_arg(bc.args, index)...) for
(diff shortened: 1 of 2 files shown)


odow:
Nightly failure is #2930

PR #2937, opened 2022-03-31
DNMY: begin refactoring of nlp.jl

odow:
The nonlinear code in JuMP is spread across a range of files and submodules.
This is the first step towards documenting and refactoring (without breaking
changes) the NLP code in JuMP.

My goal is to move the NLPEvaluator code into _Derivatives and the
nonlinear expression graph stuff from _Derivatives into JuMP, so that we
separate the AD backend from the expression datastructure. That should
then let us swap in SymbolicAD without too much trouble.

I don't intend to merge this in a hurry, so I will probably create large PR to see
how things pan out, and then once we have something, I will split it into smaller,
more digestable PRs.

### DIFF
--- a/src/_Derivatives/_Derivatives.jl
+++ b/src/_Derivatives/_Derivatives.jl
@@ -43,4 +43,6 @@ include("forward.jl")
 include("reverse.jl")
 include("subexpressions.jl")
 
+include("nlp.jl")
+
 end # module
--- a/src/_Derivatives/nlp.jl
+++ b/src/_Derivatives/nlp.jl
@@ -1,0 +1,149 @@
+"""
+    _replace_moi_variables(
+        nd::Vector{NodeData},
+        moi_index_to_consecutive_index,
+    )
+
+This function returns a copy of `nd::Vector{NodeData}`, but with the
+`MOIVARIABLE` nodes replaced by a new `MOI.VariableIndex` in which the index of
+the variable is the 1-indexed column of the variable in the problem.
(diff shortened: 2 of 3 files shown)


odow:
Hmm. I went down a long path, but going any further by just moving some methods is actually pretty non-trivial, as there are a lot of inter-dependencies between JuMP objects, MOI, and the NodeData.

I think it's going to require standardizing a nonlinear data-structure at the JuMP level, independent of `MOI.AbstractNLPEvaluator`, which includes a way to register functions. It's mostly here:
https://github.com/jump-dev/JuMP.jl/blob/8ecd560daa56ab15626da8f9f48e94d68562be3a/src/nlp.jl#L69-L78
except that the node types and `UserOperatorRegistry` are in `_Derivatives`. 

Another consideration is that EAGO (@mewilhel) and Plasmo (@jalving) both use quite a bit of the forbidden `_Derivatives` submodule, so one thing to be aware of is that any changes are likely to break EAGO and Plasmo. 
https://github.com/PSORLab/EAGO.jl/blob/5251493476a85675310c679b3926c5bbbcd54800/src/EAGO.jl#L23-L28
https://github.com/zavalab/Plasmo.jl/blob/c4f048b76fdb7230de728f7275f1f585a1c2945f/src/nlp_macro.jl#L69-L89

odow:
Closing in favor of https://github.com/jump-dev/JuMP.jl/pull/2939

PR #2938, opened 2022-04-02
DNMY: Remove Calculus.jl as a dependency

odow:
Motivation is described by the comment in `univariate_expressions_generator.jl`:
```
# JuMP requires first and second-order symbolic derivatives for univariate
# functions. Calculus.jl is a package for computing them, and JuMP 1.0.0 used to
# dynamically compute them during precompilation. The benefit of this approach
# is that if the rules in Calculus are updated, then they will automatically flow
# into JuMP. However, the Calculus package has not seen any development for two
# years, and JuMP uses only a very small part of the total package. Therefore,
# this script statically builds the list of univariate functions supported by
# JuMP and writes them to univariate_expressions.jl. This has the benefit of
# removing a dependency, and saves the recalculation of these derivatives by
# every user.
```


odow:
Closing in favor of https://github.com/jump-dev/JuMP.jl/pull/2939

PR #2939, opened 2022-04-03
DNMY: exploration of a refactored Nonlinear API design

odow:
My new plan is to move all of the input nonlinear data structure into the `src/Nonlinear` submodule, with a view to eventually moving it to MOI. JuMP's NLP macros are then greatly simplified, and simply forward to the correct `set_objective`, `add_expression` etc. 

In addition, I'll move all of the current AD to `_Derivatives`, and this will consume a `NonlinearData` object and emit a `AbstractNLPEvaluator`.  That should let us swap in `SymbolicAD` which would consume the same `NonlinearData` object.

The design of `NonlinearData` and `parse.jl` is copied with some minor refactoring from the existing JuMP code. It doesn't address the question as to whether these are the best data structures etc.

My plan is to add new tests in `test/Nonlinear`, but leave all of the current tests in `test/nlp.jl`  so that we can replicate all of the error messages etc.

Once we have things working, we can decide to merge in one go, or break it down into smaller PRs.

The README serves as WIP documentation:
https://github.com/jump-dev/JuMP.jl/blob/od/nonlinear/src/Nonlinear/README.md

This is still missing some things for feature parity with JuMP:
 - [x] Splatting
 - [x] Various error messages
 - [x] sums and generator expressions


src/Nonlinear/univariate_expressions_generator.jl
```diff
@@ -0,0 +1,60 @@
...
+#! format:off
+#
+# This file is machine generated by running univariate_expressions_generator.jl
+
+const SYMBOLIC_UNIVARIATE_EXPRESSIONS = Tuple{Symbol,Expr,Any}[
+    (:+, :(one(x)), :(zero(x))),
+    (:-, :(-one(x)), :(zero(x))),
+    (:abs, :(ifelse(x >= 0, one(x), -one(x))), :(zero(x))),""",
+    )
+    for (op, deriv) in Calculus.symbolic_derivatives_1arg()
```
jd-foster, line 51: Is there scope for allow augmentation in the list that gets produced here: e.g. if there was a future need to add a univariate function, you would have somewhere in this file
```
custom_univariate_derivative_rules = [
( :sinc,      :( (cos(x) + sinc(x)) / x  )  )
]
```
then
```
my_symbolic_derivatives_1arg_list = append(Calculus.symbolic_derivatives_1arg(), custom_univariate_derivative_rules)
 ```
and this line is
```
for (op, deriv) in my_symbolic_derivatives_1arg_list
```

Comment (odow) [reply src/Nonlinear/univariate_expressions_generator.jl:51]: Yeah there's no restriction on adding new operators. I've just done the Calculus ones for now, which are the ones that JuMP previously supported.

If we have more in future, we can just modify the script to also print them.

File: src/Nonlinear/types.jl [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+end
+
+struct ConstraintIndex
+    value::Int
+end
+
+mutable struct NonlinearData
+    objective::Union{Nothing,NonlinearExpression}
+    expressions::Vector{NonlinearExpression}
+    constraints::Dict{ConstraintIndex,NonlinearConstraint}
```
blegat, line 65: Looks like a good place for a `MOI.Utilities.VectorOfConstraints`

src/Nonlinear/types.jl
```diff
@@ -0,0 +1,91 @@
...
+    NonlinearExpression() = new(Node[], Float64[])
+end
+
+function Base.:(==)(x::NonlinearExpression, y::NonlinearExpression)
+    return x.nodes == y.nodes && x.values == y.values
+end
+
+struct NonlinearConstraint
+    expression::NonlinearExpression
+    set::Union{
```
blegat, line 42: This could be a `MOI.Utilities.StructOfConstraint` with the 4 sets and then a concrete `MOI.Utilities.VectorOfConstraint{NonlinearExpression,S}` in each field of the `MOI.Utilities.StructOfConstraint`

src/Nonlinear/types.jl
```diff
@@ -0,0 +1,91 @@
...
+    NODE_SUBEXPRESSION,
+)
+
+struct Node
+    type::NodeType
+    index::Int
+    parent::Int
+end
+
+struct NonlinearExpression
```
blegat, line 30:
```suggestion
struct NonlinearExpression <: MOI.AbstractScalarFunction
```

src/Nonlinear/operators.jl
```diff
@@ -0,0 +1,287 @@
...
+    x::AbstractVector{T},
+) where {T}
+    if op == :+
+        fill!(g, 1.0)
+    elseif op == :-
+        fill!(g, -1.0)
+        g[1] = 1.0
+    elseif op == :*
+        total = *(x...)
+        for i in 1:length(x)
```
mlubin, line 228: This rule doesn't work if total == 0. See https://github.com/jump-dev/JuMP.jl/blob/8ecd560daa56ab15626da8f9f48e94d68562be3a/src/_Derivatives/forward.jl#L94-L116.

src/Nonlinear/operators.jl
```diff
@@ -0,0 +1,287 @@
...
+end
+
+function eval_multivariate_gradient(
+    registry::OperatorRegistry,
+    op::Symbol,
+    g::AbstractVector{T},
+    x::AbstractVector{T},
+) where {T}
+    if op == :+
+        fill!(g, 1.0)
```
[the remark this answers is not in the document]
odow: Agreed. Consider this very much a first draft until I untangle the input layer from `_Derivatives`. I'm open to changing these signatures.

src/Nonlinear/univariate_expressions_generator.jl
```diff
@@ -0,0 +1,60 @@
+#  Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+# JuMP requires first and second-order symbolic derivatives for univariate
```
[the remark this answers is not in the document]
odow: For this first pass, I was just focusing on replicating the current JuMP functionality. There's a clear path to updating this list in future.

src/Nonlinear/operators.jl
odow: Yeah there's going to be quite a few of these issues. I tried to do an in-place refactor in https://github.com/jump-dev/JuMP.jl/pull/2937 (+ a lot of uncommitted local changes), but it quickly got incredibly messy as there are too many linkages between `parse_nlp.jl`, `nlp.jl`, and `_Derivatives`. 

My plan is to be in a rush to merge this branch, and to keep all of the existing tests to catch any differences. But it's going to need some careful evaluation before we consider replacing the current AD without breakage.

src/Nonlinear/types.jl
odow: I considered this. But if it contains `PARAMETER` or `SUBEXPRESSION` nodes, or a user-defined function then we're in trouble.

I'm going to avoid this for now and see what happens later.

src/Nonlinear/types.jl
odow: This can be a performance optimization at a later date. Let's not get too complicated too soon.

src/Nonlinear/types.jl
odow: Ditto on premature optimization.

odow:
Closing in favor of https://github.com/jump-dev/MathOptInterface.jl/pull/1803

PR #2940, opened 2022-04-04
IO performance note

joaquimg:
x-ref: https://discourse.julialang.org/t/jump-moi-performance-overhead-vs-xpress-api/78994

### DIFF
--- a/docs/src/reference/models.md
+++ b/docs/src/reference/models.md
@@ -59,6 +59,11 @@ Base.copy(::AbstractModel)
 ```
 ## I/O
 
+!!! note
+    JuMP readers and writers are not optimized for performance. The most
+    efficient way to create a JuMP problem is, usually, incrementally building
+    the model with JuMP macros.
+
 ```@docs
 write_to_file
 Base.write(::IO, ::Model; ::MOI.FileFormats.FileFormat)


odow (commented):
This isn't strictly true. The problem is that they create an additional copy of the model, as we go `file -> FileFormats.X.Model -> JuMP.Model -> solver`.

joaquimg:
You thinks it is better to tell the entire story?
Like the MOI piece might be optimized but JuMP's is never optimized due to copy, and then this would be inefficient for jump end users that want to pass file data to solvers.

odow:
I commented one Discourse. It's not apparent where the slow-down is. I'd hold off adding this warning for now. And if we do decide to add it, it should probably go https://jump.dev/JuMP.jl/stable/manual/models/#Read-a-model-from-file?

odow:
> are users supposed to expect efficient model creation from files as they have in other software?

I think it's unreasonable to expect that reading a problem into JuMP and then into Xpress will be similar speed as reading the problem directly into Xpress. So I would expect a 2x slow-down at least (we have to read into JuMP, and then read into Xpress). The massive difference in the latest discourse post looks like compilation latency. I'd expect the second run time to be much closer in performance.

joaquimg:
I consider 2x os xpress reader a efficient MPS reader.

Since we agreed that readers should be efficient I will close this.

odow:
Yeah I don't think we're at the end of the story with the discourse post, but I don't have Xpress to test or benchmark. I guess `djunglas` is Daniel Junglas from FICO, so maybe they're providing official support?

odow:
Okay, I didn't have Xpress, but I do have HiGHS:
https://discourse.julialang.org/t/jump-moi-performance-overhead-vs-xpress-api/78994/10

The performance difference is just time-to-first-solve latency. We're actually really good on the second solve (0.0015 for JuMP vs 0.0011 for low-level HiGHS).

PR #2941, opened 2022-04-04
Account for diagonal quadratic coefficients

rjuly7:
Minor modification to this documentation: note that for diagonal quadratic coefficients of the form a*x[i]^2, the second derivative with respect to x[i] is 2*a. When adding the quadratic terms, we should check to see if the term is diagonal and compute the derivative accordingly. The reason this issue doesn't come up in the example model is that the optimal dual variable for the only quadratic constraint is 0, so the quadratic constraint doesn't show up in the optimal Hessian.


odow:
Oops. I always forget those multipliers.

PR #2942, opened 2022-04-05
DNMY: refactored _Derivatives on new .Nonlinear submodule

odow:
To keep things a little more comprehensible, I'm going to keep this
PR stacked on top of #2939, and I'll rebase it as needed.

<s>There's still quite a way to go before this is ready, but the objective
function and gradient evaluations are working as a proof-of-concept.</s>

### DIFF
--- a/docs/src/developers/nonlinear.md
+++ b/docs/src/developers/nonlinear.md
@@ -196,6 +196,7 @@ before we can call [`MOI.initialize`](@ref), we need to set an
 There are two to choose from within JuMP, although other packages may add more
 options by sub-typing [`Nonlinear.AbstractAutomaticDifferentiation`](@ref):
  * [`Nonlinear.Default`](@ref)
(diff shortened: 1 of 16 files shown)


odow:
This has been quite an effort, and #2939 is still missing some features, but things are now mostly plumbed together. 

We've effectively created a new DSL for nonlinear optimization based on a subset of pure-Julia expression trees as input, and an extensible AD backend system that defaults to JuMP's sparse reverse-mode AD:
https://github.com/jump-dev/JuMP.jl/tree/od/ReverseAD/src/Nonlinear

Because we don't depend on JuMP, only MathOptInterface, we can move `Nonlinear` to a submodule of MathOptInterface (or a separate package). That should greatly simplify the amount of code in JuMP related to nonlinear.

odow:
Closing in favor of https://github.com/jump-dev/MathOptInterface.jl/pull/1803

PR #2945, opened 2022-04-11
Improve minimal ellipses SDP demonstration

maxkapur:
This PR improves the "Minimum ellipses" demonstration in the documentation:
- Fixed the objective function to use the log determinant of the matrix instead of the trace.
   (Logdet is correct, but  the `MOI.RootDetConeTriangle` interface wasn't working at the time the previous version of this demo was created, so the author used trace as an approximation.)
- Allow input ellipses that aren't centered at the origin.
- Add section number to textbook reference. 
- Add graphical output.
- Change title from "Minimum ellipses" to "Minimal ellipses."

See [this discussion](https://discourse.julialang.org/t/jump-docs-on-sdp-replace-logdet-with-tr-is-this-correct/79324) on Discourse.

### DIFF
--- a/docs/src/tutorials/conic/min_ellipse.jl
+++ b/docs/src/tutorials/conic/min_ellipse.jl
@@ -77,7 +77,6 @@ function minimal_ellipse(As, bs, cs)
 
     @assert termination_status(mdl) == OPTIMAL
     @assert primal_status(mdl) == FEASIBLE_POINT
-    @assert isapprox(objective_value(mdl), 0.132413, atol=1e-4)
 
     # Restore original parameterization
     P = sqrt(value.(Psqr))
@@ -86,6 +85,8 @@ function minimal_ellipse(As, bs, cs)
 end
 
 P, q = minimal_ellipse(As, bs, cs)
+@assert isapprox(P, [0.423694 -0.039639; -0.039639 0.316316], atol=1e-4)
+@assert isapprox(q, [-0.396177, -0.021368], atol=1e-4)
 
 # Plot results
 pl = plot(size=(600,600))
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/conic/min_ellipse.jl
```diff
@@ -3,48 +3,101 @@
...
+            in PSDCone())
+    end
+
+    @objective(mdl, Max, logdetP)
+
+    optimize!(mdl)
+
+    @assert termination_status(mdl) == OPTIMAL
+    @assert primal_status(mdl) == FEASIBLE_POINT
+    @assert isapprox(objective_value(mdl), 0.132413, atol=1e-4)
```
jd-foster, line 80: I think this last `@assert` shouldn't be inside the function?

maxkapur: Nice catch, will fix shortly.

docs/src/tutorials/conic/min_ellipse.jl
```diff
@@ -3,48 +3,102 @@
...
-    optimize!(model)
-    Test.@test termination_status(model) == OPTIMAL
-    Test.@test primal_status(model) == FEASIBLE_POINT
-    Test.@test objective_value(model) ≈ 6.46236 atol = 1e-4
-    Test.@test value.(X) ≈ [3.1651 0.8022; 0.8022 3.2972] atol = 1e-4
-    return
+cs = [0.1831, 0.3295, 0.2077, 0.2362, 0.3284, 0.4931]
+
+function minimal_ellipse(As, bs, cs)
+    # Build the model under the change of variables
```
odow, line 52:
```suggestion
    ## Build the model under the change of variables
```

To make a comment inside a Julia block, you need double hash signs: `##`

docs/src/tutorials/conic/min_ellipse.jl
```diff
@@ -3,48 +3,102 @@
...
+    #       Psqr = P^2
+    #    q_tilde = Pq
+    mdl = Model(SCS.Optimizer)
+
+    m = length(As)
+    n, _ = size(first(As))
```
jd-foster, line 58: One idea is to add a validation step to the data:
```
if !(isreal(A) && transpose(A) == A)
    @error "Input matrices need to be real, symmetric matrices."
end
```

odow (approved):
Looks much better

PR #2949, opened 2022-04-11
Relax solution test tolerance

jd-foster:
These tolerances are a bit tight to pass on my machine due to natural variation.


odow (approved):
Sure. I also see that the plot didn't get displayed. Perhaps we need to get rid of the function?

PR #2950, opened 2022-04-12
Convert minimal ellipse tutorial to sections

jd-foster:
Re-formatting into sections and using LaTeX where appropriate.
Some other corrections and clarifications.

### DIFF
--- a/docs/src/tutorials/conic/min_ellipse.jl
+++ b/docs/src/tutorials/conic/min_ellipse.jl
@@ -23,7 +23,6 @@
 # ```math
 # \begin{aligned}
 # \text{maximize } & \quad \log(\det(P))  \\
-
 # \text{subject to } & \quad \tau_i \geq 0, & i = 1, \ldots, m, \\
 # & \quad\begin{bmatrix}
 #     P^2 - \tau_i A_i      &  P q - \tau_i b_i   &    0     \\
--- a/docs/src/tutorials/conic/min_ellipse.jl
+++ b/docs/src/tutorials/conic/min_ellipse.jl
@@ -5,110 +5,140 @@
 
 # # Minimal ellipses
 
-# This example comes from section 8.4.1 of the book *Convex Optimization* (Boyd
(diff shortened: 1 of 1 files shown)


jd-foster:
On my local machine, getting `Documenter.makedocs` failures due to doctests that don't align to types in newer versions of Julia and MOI.

docs/src/tutorials/conic/min_ellipse.jl
```diff
@@ -5,110 +5,140 @@
...
+# It is convenient to parameterize as the minimal enclosing ellipse as
+# ```math
+# \{ x : || Px + q || \leq 1 \}.
+# ```
 
-# Then the optimal `P` and `q` are given by the convex program
+# Then the optimal ``P`` and ``q`` are given by the convex semidefinite program
+# ```math
+# \begin{aligned}
+# \text{maximize } & \quad \log(\det(P))  \\
```
odow, line 26:
```suggestion
```

Okay. Debugging these latex failures is nigh-on impossible. But it's usually because of empty lines like this. there's an open issue to parse the latex log for errors, but yeah. not fun. Going from `.jl` to `.md` to `LaTeX` means it could be any number of reasons :(

jd-foster: Just pushed a formatted version. Fingers crossed.

odow:
@jd-foster happy? The latex issue was the extra line :(

jd-foster:
Yep, looks good.

odow:
Nice! https://jump.dev/JuMP.jl/dev/tutorials/conic/min_ellipse/

cc @maxkapur thanks for helping on this.

jd-foster:
Yes, thanks @maxkapur for improving this example. I thought the plots were really nice and needed to get onto the web page.

maxkapur:
Thank you both for all these improvements! The way you blocked out the code and plots looks great. This was my first PR *ever,* and first time using Literate/Documenter, so I really appreciated all your help.

PR #2954, opened 2022-04-15
Change "a" to "an"

CSWGTM-DAVE:
Just a small grammatical tweak.
Instead of "If you want to optimize a ordinary differential equation..." I changed to "If you want to optimize an ordinary differential equation" which is just changing the "a" to an "an" indefinite article.


mlubin:
Thanks!

PR #2955, opened 2022-04-18
Refactor to use MOI.Nonlinear

odow:
# ANN: upcoming refactoring of JuMP's nonlinear API

tl;dr: The next feature release of JuMP (probably v1.2.0) introduces a large
refactoring of JuMP's nonlinear API. If you have code that accessed private
features of the nonlinear API, such as `JuMP._Derivatives` or `model.nlp_data`,
your code will break. If you used only the public, documented API such as
`register`, `@NLconstraint` and `num_nonlinear_constraints`, this change does
not affect you. To try the uncoming release, use
`import Pkg; Pkg.pkg"add JuMP#od/moi-nonlinear"`.

The relevant pull request is https://github.com/jump-dev/JuMP.jl/pull/2955

## What are we doing?

Over the last few months, we have been refactoring how JuMP supports nonlinear
programs. This involved moving and re-organizing a large amount of code from
JuMP into MathOptInterface.

The result is the new `MOI.Nonlinear` submodule in MathOptInterface, with a
documented public API for creating and dealing with nonlinear programs. Read
more about it here: https://jump.dev/MathOptInterface.jl/stable/submodules/Nonlinear/overview/

However, as part of this work we are removing code from JuMP. This code was
internal, undocumented, and not intended for public use. Most of it was
contained in the `JuMP._Derivatives` submodule, but we also made changes such
as removing `model.nlp_data`.

## Why did we do this?

The nonlinear code in JuMP was a clear example of technical debt. It was
complicated, convoluted, and largely undocumented. People wanting to extend JuMP
for nonlinear programs were forced to use a range of hacks that relied on
undocumented internals.

The new `MOI.Nonlinear` submodule offers a stable, documented, and public API
for people to build JuMP extensions on. It also enables new features like
swappable automatic differentiation backends, and hessians of user-defined
functions.

We originally considered that any change in the nonlinear API would be a
breaking v2.0.0 release of JuMP and occur at least two years after the release
of JuMP 1.0. However, we implemented the changes quicker than expected, and were
able to do so in a way that does not break the public nonlinear API. Therefore,
we elected to classify this as a non-breaking feature release.

## Does it affect me?

If you have any code that called private features of the JuMP nonlinear API,
such as `JuMP._Derivatives` or `model.nlp_data`, your code will break.

If you used only the public, documented API such as `register`, `@NLconstraint`
and `num_nonlinear_constraints`, this does not affect you.

## What are the next steps

Try the uncoming release as follows: 
`import Pkg; Pkg.pkg"add JuMP#od/moi-nonlinear"`.

If you find any bugs or changes in performance, please post below, or open a
GitHub issue. Once we're happy that there are no issues with the changes, we
will release a new version of JuMP with the changes.


odow:
What are the concrete next steps for merging this into JuMP? It's a bit of a weird PR, because on one hand, it changes a lot of code, but on the other hand, very little actually changes at the user-facing level (just a couple of errors move from compile time to run time, and some constants change from `0` to `0.0`).

I think at minimum, we need the following reviews:

 - [ ] @blegat
 - [ ] @mlubin
 - [ ] @joaquimg 
 - [ ] @ccoffrin 

Optional
 
 - [ ] @frapac 
 - [ ] @matbesancon

src/operators.jl
```diff
@@ -397,7 +397,7 @@ end
 ###############################################################################
 
 const _OP_HINT = "Are you trying to build a nonlinear problem? Make sure you use @NLconstraint/@NLobjective."
-for (func, _) in Calculus.symbolic_derivatives_1arg(),
+for func in MOI.Nonlinear.DEFAULT_UNIVARIATE_OPERATORS[4:end],
```
frapac, line 400: I would avoid using a magic number here (why discarding the three first terms in the list?). How about implementing a proper getter in the MOI side to return the [following operators](https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Nonlinear/univariate_expressions.jl#L15-L83)?

src/optimizer_interface.jl
```diff
@@ -123,26 +123,44 @@ function set_optimizer(
...
+[`NoOptimizer`](@ref) error is thrown.
 
-Keyword arguments `kwargs` are passed to the `optimize_hook`. An error is
-thrown if `optimize_hook` is `nothing` and keyword arguments are provided.
+If `ignore_optimize_hook == true`, the optimize hook is ignored and the model is
+solved as if the hook was not set. Keyword arguments `kwargs` are passed to the
+`optimize_hook`. An error is thrown if `optimize_hook` is `nothing` and keyword
+arguments are provided.
+
+Pass `differentiation_backend` to set the `AbstractAutomaticDifferentiation`
```
frapac, line 144: Should we add a link to MOI's documentation there?

frapac (commented):
This PR is a net improvement imo. The code is a lot cleaner now the nonlinear backend is factorized on the MOI side. I have tested this branch on my custom nonlinear models, and JuMP works like a charm on them, returning the same results as before. 

I only have a few minor comments (I don't have the expertise to assess the correctness of the macro in JuMP). 

Also, one remaining TODO on the JuMP side was the support of incremental solves for nonlinear model:
https://github.com/jump-dev/JuMP.jl/blob/od/moi-nonlinear/src/optimizer_interface.jl#L156

I am wondering how difficult it would be to support incremental nonlinear solves with the new nonlinear interface? E.g. to add new nonlinear constraints `MOI.Nonlinear.Model` between two consecutive solves.

src/JuMP.jl
```diff
@@ -149,7 +143,7 @@ mutable struct Model <: AbstractModel
     # where kwargs get passed along to subsequent solve calls.
     optimize_hook::Any
     # TODO: Document.
-    nlp_data::Any
+    nlp_data::Union{Nothing,MOI.Nonlinear.Model}
```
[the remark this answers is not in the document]
matbesancon: +1

src/JuMP.jl
```diff
@@ -619,15 +613,6 @@ Returns number of variables in `model`.
 """
 num_variables(model::Model)::Int64 = MOI.get(model, MOI.NumberOfVariables())
 
-"""
-    num_nonlinear_constraints(model::Model)
-
-Returns the number of nonlinear constraints associated with the `model`.
```
mlubin, line 625: Is this method being deleted or moved somewhere else? I couldn't find easily find the new version if it moved.

mlubin (commented):
Have you run end-to-end benchmarks? Maybe this was already done, but I remember at some point there were performance TODOs in MOI.Nonlinear.

src/macros.jl
```diff
@@ -1797,6 +1797,168 @@ function _reorder_parameters(args)
...
+        )
+    end
+    if isexpr(x, :ref)
+        return esc(x)
+    elseif isexpr(x, :.)
+        return esc(x)
+    elseif _is_generator(x)
+        return _parse_generator_expression(code, x, operators)
+    elseif isexpr(x, Symbol("'"))
+        # Special-case the adjoint operator because it often happens with
```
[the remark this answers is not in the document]
odow: Done

src/optimizer_interface.jl
odow: Done

src/operators.jl
odow: I've refactored this loop. We ignored `+`, `-`, and `abs` because they were defined separately in `_Derivatives` and so weren't included here. Skipping `+` and `-` makes sense because they're implemented separately, but I don't know why `abs` was skipped.

odow:
> Also, one remaining TODO on the JuMP side was the support of incremental solves for nonlinear model:

That's my plan once this is merged, but it should be a separate PR.

src/JuMP.jl
odow: No public facing functions were deleted or changed. It's now here:
https://github.com/jump-dev/JuMP.jl/blob/bbd8d99a6cc639d928956994d70e4c3343d192bd/src/nlp.jl#L429-L436

The diff is a little hard to navigate. And it made sense to move this to the nonlinear code, rather than have it scattered in JuMP.jl.

odow:
> I remember at some point there were performance TODOs in MOI.Nonlinear.

I reverted the performance TODOs in MOI.Nonlinear, so now the only change is moving some of the expression generation from compile time in the macros to runtime. I can make some more benchmarks.

src/JuMP.jl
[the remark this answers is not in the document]
odow: I guess at the point, we're already breaking anything that touches `nlp_data`, so there's little harm in further breakage.

src/macros.jl
```diff
@@ -1797,6 +1797,168 @@ function _reorder_parameters(args)
...
 
+"""
+    _parse_nonlinear_expression(data, x::Expr)
+
+JuMP needs to build Nonlinear expression objects in macro scope. This has two
+main challenges:
+
+ 1. We need to evaluate local variables into the expressions. This is reasonably
+    easy, anywhere we see a symbol that is not a function call, replace it by
+    esc(x).
```
blegat, line 1808:
```suggestion
    `esc(x)`.
```

src/optimizer_interface.jl
```diff
@@ -123,26 +123,45 @@ function set_optimizer(
...
+Pass `differentiation_backend` to set the [`MOI.Nonlinear.AbstractAutomaticDifferentiation`](@ref)
+backend used to compute derivatives of nonlinear programs.
+
+If you require only `:ExprGraph`, it is more efficient to pass
+`differentiation_backend = MOI.Nonlinear.ExprGraphOnly()`.
 """
 function optimize!(
     model::Model;
     ignore_optimize_hook = (model.optimize_hook === nothing),
+    differentiation_backend::MOI.Nonlinear.AbstractAutomaticDifferentiation = MOI.Nonlinear.SparseReverseMode(),
```
blegat, line 153: I'm wondering whether this should be a keyword argument to `optimize!` or something set with a separate function.
For instance, if you call `optimize!` several times then you need to set it every time.
In the future, if the differentiation is done through bridges then this should be passed as an MOI attribute so having a separate function seems preferable in that regards.

ccoffrin:
I reviewed the code changes and nothing stood out to me as problematic.  I would suggest that any feature adds, which run the risk of generating a desire for possible future breaking changes be discussed at the JuMP developer call for concurrence across the wider group (if this has not already been done).

The only two points that I saw that would fall into this category were:
- API for setting AD backends
- API for setting user-defined Hessians

src/optimizer_interface.jl
mlubin: I don't have a strong preference either, but this seems like something that might change once we know all the use cases. Maybe we should keep this argument as an internal API for now (e.g., name it `_differentiation_backend` and document accordingly) so people have a chance to experiment?

odow: Okay, I've renamed to `_differentiation_backend`.

odow:
> API for setting user-defined Hessians

Yeah, let's leave this for the separate PR https://github.com/jump-dev/JuMP.jl/pull/2961

matbesancon (approved):
LGTM!

mlubin (approved):

ccoffrin (approved):

frapac (approved):

odow:
So with respect to benchmarks, this PR is a little slower at constructing the model because it double-hands the input expressions. The nonlinear macros used to expand into something that directly produced the nonlinear data structure, but now they expand into something that generates the input `Expr` for `MOI.Nonlinear`, which then creates the data structure at run-time.

For example, `master` creates:
```Julia
julia> @macroexpand @NLexpression(model, sin(x)^2)
quote
    #= REPL[13]:1 =#
    JuMP._valid_model(model, :model)
    JuMP.NonlinearExpression(model, begin
            #= /Users/oscar/.julia/dev/JuMP/src/parse_nlp.jl:542 =#
            var"#616#tape" = JuMP.NodeData[]
            #= /Users/oscar/.julia/dev/JuMP/src/parse_nlp.jl:543 =#
            var"#617#values" = JuMP.Float64[]
            #= /Users/oscar/.julia/dev/JuMP/src/parse_nlp.jl:544 =#
            let
                JuMP.push!(var"#616#tape", JuMP.NodeData(JuMP.CALL, 4, -1))
                var"#618###484" = JuMP.length(var"#616#tape")
                let
                    JuMP.push!(var"#616#tape", JuMP.NodeData(JuMP.CALLUNIVAR, 15, var"#618###484"))
                    var"#619###485" = JuMP.length(var"#616#tape")
                    JuMP._parse_NL_expr_runtime(model, x, var"#616#tape", var"#619###485", var"#617#values")
                end
                JuMP._parse_NL_expr_runtime(model, 2, var"#616#tape", var"#618###484", var"#617#values")
            end
            #= /Users/oscar/.julia/dev/JuMP/src/parse_nlp.jl:545 =#
            JuMP._NonlinearExprData(var"#616#tape", var"#617#values")
        end)
end
```
but this PR creates
```julia
julia> @macroexpand @NLexpression(model, sin(x)^2)
quote
    #= REPL[10]:1 =#
    JuMP._valid_model(model, :model)
    begin
        #= /Users/oscar/.julia/dev/JuMP/src/macros.jl:2409 =#
        begin
            #= /Users/oscar/.julia/dev/JuMP/src/macros.jl:1838 =#
            JuMP._init_NLP(model)
            var"#195###483" = JuMP.Expr(:call, :sin, x)
            var"#196###482" = JuMP.Expr(:call, :^, var"#195###483", 2)
        end
        #= /Users/oscar/.julia/dev/JuMP/src/macros.jl:2410 =#
        JuMP.add_nonlinear_expression(model, var"#196###482")
    end
end
```

Thus our worst-case benchmark is something like this one, which creates a lot of expressions:
```Julia
using JuMP
function main(n = 500)
    model = Model()
    γ, ρ, r, μR, ζ = 2, 0.05, 0.041, 0.051, 1.5
    σR = sqrt((ζ / γ + 1) / 2 * (μR - r)^2 / (ρ - r))
    amin, amax = -0.3, 1e4
    x = map(x -> x + 5 * x^10, range(0, stop = 1, length = n))
    a = (amax - amin) / (maximum(x) - minimum(x)) * x .+ amin
    z, λ = 3.0 * [0.01, 0.03], [0.5, 0.5]
    @variable(
        model,
        V[i = 1:n, j = 1:2],
        start = (z[j] + r * a[i])^(1 - γ) / (1 - γ) / ρ
    )
    @NLexpression(model, Vpb1[j = 1:2], (z[j] + a[1] * r)^(-γ))
    @NLexpression(
        model,
        Vpbo[i = 2:n, j = 1:2],
        (V[i, j] - V[i-1, j]) / (a[i] - a[i-1])
    )
    @NLexpression(
        model,
        Vpb[i = 1:n, j = 1:2],
        (i == 1) * Vpb1[j] + (i > 1) * Vpbo[max(i, 2), j]
    )
    @NLexpression(
        model,
        Vpfo[i = 1:(n-1), j = 1:2],
        (V[i+1, j] - V[i, j]) / (a[i+1] - a[i])
    )
    @NLexpression(
        model,
        Vpfn[j = 1:2],
        (z[j] + a[n] * r + (μR - r)^2 / (γ * σR^2) * a[n])^(-γ)
    )
    @NLexpression(
        model,
        Vpf[i = 1:n, j = 1:2],
        (i == n) * Vpfn[j] + (i < n) * Vpfo[min(i, n - 1), j]
    )
    @NLexpression(
        model,
        Vp[i = 1:n, j = 1:2, s = 1:2],
        (s == 1) * Vpb[i, j] + (s == 2) * Vpf[i, j]
    )
    @NLexpression(model, Vpp1[j = 1:2], 0.0)
    @NLexpression(model, Vppn[j = 1:2], -γ * Vp[n, j, 1] / a[n])
    @NLexpression(
        model,
        Vppo[i = 2:(n-1), j = 1:2],
        (
            (a[i] - a[i-1]) * V[i+1, j] + (a[i+1] - a[i]) * V[i-1, j] -
            (a[i+1] - a[i-1]) * V[i, j]
        ) / (0.5 * (a[i+1] - a[i-1]) * (a[i+1] - a[i]) * (a[i] - a[i-1]))
    )
    @NLexpression(
        model,
        Vpp[i = 1:n, j = 1:2],
        (i == 1) * Vpp1[j] +
        (i == n) * Vppn[j] +
        (i > 1) * (i < n) * Vppo[min(max(i, 2), n - 1), j]
    )
    @NLexpression(
        model,
        c[i = 1:n, j = 1:2, s = 1:2],
        abs(Vp[i, j, s])^(-1 / γ)
    )
    @NLexpression(model, k1[j = 1:2, s = 1:2], 0.0)
    @NLexpression(
        model,
        ko[i = 2:n, j = 1:2, s = 1:2],
        (-Vp[i, j, s] / Vpp[i, j]) * (μR - r) / σR^2
    )
    @NLexpression(
        model,
        kk[i = 1:n, j = 1:2, s = 1:2],
        (i == 1) * k1[j, s] + (i > 1) * ko[max(i, 2), j, s]
    )
    @NLexpression(
        model,
        k[i = 1:n, j = 1:2, s = 1:2],
        (kk[i, j, s] <= (a[i] - a[1])) * (kk[i, j, s] >= 0) * kk[i, j, s] +
        (kk[i, j, s] >= (a[i] - a[1])) * (a[i] - a[1])
    )
    @NLexpression(
        model,
        savings[i = 1:n, j = 1:2, s = 1:2],
        (z[j] + r * a[i] + (μR - r) * k[i, j, s] - c[i, j, s])
    )
    @NLexpression(
        model,
        UVp[i = 1:n, j = 1:2],
        (savings[i, j, 1] <= 0) * Vp[i, j, 1] +
        (savings[i, j, 1] > 0) * (savings[i, j, 2] >= 0) * Vp[i, j, 2]
    )
    @NLexpression(
        model,
        Uk[i = 1:n, j = 1:2],
        (savings[i, j, 1] <= 0) * k[i, j, 1] +
        (savings[i, j, 1] > 0) * (savings[i, j, 2] >= 0) * k[i, j, 2] +
        (savings[i, j, 1] > 0) *
        (savings[i, j, 2] < 0) *
        (k[i, j, 1] + k[i, j, 2]) / 2
    )
    @NLexpression(
        model,
        Uc[i = 1:n, j = 1:2],
        (savings[i, j, 1] <= 0) * c[i, j, 1] +
        (savings[i, j, 1] > 0) * (savings[i, j, 2] >= 0) * c[i, j, 2] +
        (savings[i, j, 1] > 0) *
        (savings[i, j, 2] < 0) *
        (z[j] + r * a[i] + (μR - r) * Uk[i, j])
    )
    @NLexpression(
        model,
        HJB[i = 1:n, j = 1:2],
        Uc[i, j]^(1 - γ) / (1 - γ) +
        UVp[i, j] * (z[j] + r * a[i] + (μR - r) * Uk[i, j] - Uc[i, j]) +
        Vpp[i, j] * 0.5 * Uk[i, j]^2 * σR^2 +
        λ[j] * (V[i, (j==1)*2+(j==2)*1] - V[i, j])
    )
    @NLconstraint(model, [i = 1:n, j = 1:2], ρ * V[i, j] == HJB[i, j])
    return model
end
```

`master`

```
julia> @time main(500);
  0.025588 seconds (189.71 k allocations: 28.809 MiB, 62.65% gc time)

julia> @time main(5000);
  0.252713 seconds (1.90 M allocations: 285.254 MiB, 37.24% gc time)

julia> @time main(50000);
  3.885934 seconds (19.00 M allocations: 2.791 GiB, 48.13% gc time)
```

`od/moi-nonlinear`

```
julia> @time main(500);
  0.059545 seconds (926.95 k allocations: 62.657 MiB, 33.84% gc time)

julia> @time main(5000);
  0.979608 seconds (9.42 M allocations: 623.331 MiB, 51.90% gc time)

julia> @time main(50000);
  8.716217 seconds (94.29 M allocations: 6.100 GiB, 36.49% gc time)
```

So this PR is slower (2-3x), but the slow-down is linear and unlikely to have performance cliffs.

Overall, I think the trade-off of refactoring the nonlinear into something more documented and maintainable is worth some performance hit on the front-end.

odow:
As another data point: the 10k bus system in PowerModels:
```Julia
function bench_opf_acp(; case::String)
    solver = optimizer_with_attributes(Ipopt.Optimizer, "tol" => 1e-6)
    data = parse_file(joinpath(@__DIR__, "data", case))
    result = run_opf(data, ACPPowerModel, solver)
    println()
    println(result["termination_status"])
    println(result["primal_status"])
    println(result["solve_time"])
    return
end
```

`master` is
```
julia> @time bench_opf_acp(case = "pglib_opf_case10000_goc.m")
[... omitted ... ]
LOCALLY_SOLVED
FEASIBLE_POINT
114.346195936203
122.032420 seconds (140.79 M allocations: 4.908 GiB, 2.65% gc time, 1.19% compilation time)
```
and this PR is
```
julia> @time bench_opf_acp(case = "pglib_opf_case10000_goc.m")
[... omitted ... ]
LOCALLY_SOLVED
FEASIBLE_POINT
110.53965616226196
118.962225 seconds (149.78 M allocations: 5.140 GiB, 2.31% gc time, 1.36% compilation time)
```
So they both take ~8 seconds of initialization and setup, and then the runtimes within Ipopt are pretty similar (I think the 4 second difference is just noise). Slightly more allocations in this PR from double handling the expressions.

odow:
I've tagged JuMP v1.1.1, so I'd propose that we merge this, and then invite people interested in nonlinear to try out `] add JuMP#master` for a while before we tag JuMP 1.2.0.

odow:
I've edited the PR description with a post I intend to post to discourse: https://github.com/jump-dev/JuMP.jl/pull/2955#issue-1207503496. Please review it and let me know if you have any suggestions. We should probably give the community a heads up before merging such a big change.

odow:
Posted here: https://discourse.julialang.org/t/ann-upcoming-refactoring-of-jumps-nonlinear-api/83052

odow:
I've opened issues in the packages that will break. I found them by searching Julia hub for things like `_Derivatives` and `.nlconstr` and `.nlexpr`.

A few, like NLPModelsJuMP, are easy fixes to use only the public API. But others like EAGO are pretty much dead in the water because they heavily used a lot of the stuff in `_Derivatives` and other `_private` methods.

Some of the others are just for printing (MPSGE), which I guess going without isn't too much of a penalty. Others like Complementarity and DisjunctiveProgramming (ab)use the API to do things like delete nonlinear constraints and modify nonlinear expressions, so I don't know if we have a safe work-around.

The main takeaway from this that I will push a commit to shortly, is to overload `model.nlp_data` with nice error message, so that anyone using a package which accesses the internal API can get something informative (like telling them to install JuMP v1.1).

odow:
I took a look at some of the packages:

 - Plasmo and EAGO are hard because they touch a lot of things. @jalving is taking a look at Plasmo.

 - Complementarity: The "simple" fix is just [deleting nonlinear constraints](https://github.com/chkwon/Complementarity.jl/blob/c1afb671907edbfb205300cf0fff46d695c2f4f0/src/mcp.jl#L235-L237), but a deeper look suggests we could do better by fixing the root cause. It does a number of questionable things, like calling `MOI.initialize` every time within the callback, mucking with macros, and adding new constraints.
 - MPSGE: depends on Complementarity. Hacks at Complementarity's expression system to print things nicely: https://github.com/anthofflab/MPSGE.jl/blob/83419a7060218b0255e7f8f8d372d407ba306c6b/src/algebraic_wrapper.jl#L96. A nicer fix of Complementarity could simplify this
 - NLPModelsJuMP: need constraint lower/upper bounds. Would be easy if you could call `index.(all_nonlinear_constraints(model))` and then `nonlinear_model(model)[c]` because this would return the expression/set form of the constraint. But they also do this highly questionable thing of copying the registered functions and they touch `.user_output_buffer` for some reason: https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/blob/2cbf7717c29ee7fc34d1f5a83f017d2cc6cb6058/src/utils.jl#L387-L409. Probably work taking a much larger look at how we could turn a JuMP model into an NLPModel, because it shouldn't require that sort of thing.
 - DisjunctiveProgramminng: modifies the expression of nonlinear constraints. Exposing access to the documented expression graph in MOI.Nonlinear is probably enough, but it might always be a little caveat utilitor.
 - BilevelJuMP is just some plumbing stuff of new `MOI.Nonlinear.parse_expression` overloads.

Action items:

- [ ] Implement `nonlinear_model` to avoid people writing `model.nlp_model`. 
- [ ] Implement `index` to return indices in MOI.Nonlinear given a nonlinear expression/constraint/parameter index from JuMP
- [ ] Add `delete(::Model, ::NonlinearConstraintRef)`

davidanthoff:
Cross posting a message I just left on discourse:

> I understand that technical making this a minor release is correct, but on the other hand my sense is that there are packages out there that depend on JuMP and that rely on internals that will break if this gets released as 1.2, right? Complementarity.jl is one that comes to mind, not sure whether there are others as well?
>
> In my mind, in such a situation it would be more helpful for users to just make this a 2.0 release. If I’m a user of a package that depends on JuMP internals, I a) might not even be aware of that, and b) can’t really do anything about it. In the end the user experience will just be that things don’t work, and I think it would be nicer to try to prevent that.
>
> Plus, I’m not aware how a depending package like Complementarity.jl could even signal that it doesn’t work with JuMP 1.2, but it does work with 1.1? Maybe I’m missing something there…
>
> Would there be any real downside of just making this a 2.0?
>
> And also: awesome work, very excited to try it!

davidanthoff:
Also, another question: I remember that there was discussion to move the functionality of Complementarity into JuMP eventually, and I think the idea was to first revamp the NLP stuff and then do that. Just wondering whether the current redesign is such that it would facilitate such a step?

matbesancon:
> I understand that technical making this a minor release is correct, but on the other hand my sense is that there are packages out there that depend on JuMP and that rely on internals that will break if this gets released as 1.2, right?

This is the same as for Julia itself, relying on internals and unexported modules makes it possible that future versions will break one's code.

> Plus, I’m not aware how a depending package like Complementarity.jl could even signal that it doesn’t work with JuMP 1.2, but it does work with 1.1?

Yes Pkg does that with Tilde or equality specifies https://pkgdocs.julialang.org/v1/compatibility/#Tilde-specifiers

> In my mind, in such a situation it would be more helpful for users to just make this a 2.0 release. If I’m a user of a package that depends on JuMP internals, I a) might not even be aware of that, and b) can’t really do anything about it. In the end the user experience will just be that things don’t work, and I think it would be nicer to try to prevent that.

This is a very valid point, that's why @odow opened PRs or at least issues on all packages depending on the nonlinear part and that will break with the upgrade, similar to what Julia does when something is "theoretically" not breaking but practically is because of reliance on internals or unspecified behavior

odow:
- [ ] We should update compat bounds and modify General

odow:
> I remember that there was discussion to move the functionality of Complementarity into JuMP eventually, and I think the idea was to first revamp the NLP stuff and then do that. Just wondering whether the current redesign is such that it would facilitate such a step?

No progress on this yet. The current change is very much a maintenance, documentation, and paying down technical debt change. There's no real new functionality. The nonlinear complementarity is something I'm thinking about though.

odow:
Opened PRs to fix compat bounds. EAGO, Complementarity, and MPSGE haven't released tagged versions that are compatible with JuMP 1.0 yet.

odow:
Hmm. @abelsiqueira found an unintentional breaking change we obviously don't have any tests for: https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/pull/118/files#r911475220

````julia
julia> using JuMP

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x)
x

julia> @NLconstraint(model, sin(x) ≤ 1)
ERROR: Unrecognized function "≤" used in nonlinear expression.

You must register it as a user-defined function before building
the model. For example, replacing `N` with the appropriate number
of arguments, do:
```julia
model = Model()
register(model, :≤, N, ≤, autodiff=true)
# ... variables and constraints ...
```

Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] assert_registered(registry::MathOptInterface.Nonlinear.OperatorRegistry, op::Symbol, nargs::Int64)
   @ MathOptInterface.Nonlinear ~/.julia/packages/MathOptInterface/kCmJV/src/Nonlinear/operators.jl:416
 [3] macro expansion
   @ ~/.julia/dev/JuMP/src/macros.jl:1874 [inlined]
 [4] top-level scope
   @ REPL[4]:1
````
This used to work:
```julia
julia> using JuMP

julia> model = Model(); @variable(model, x)
x

julia> @NLconstraint(model, sin(x) ≤ 1)
sin(x) - 1.0 ≤ 0
```

Fix and tests incoming.

odow:
Merging this now, but I'll wait a while longer before tagging v1.2.0 so we can have a go at updating some of the extensions.

PR #2957, opened 2022-04-20
[docs] clean and simplify docstring of variable macro

### LINKED ISSUE #2956: User of extensions creating anonynous variables
Although the keyword: `variable_type` of the `@variable` macro is documented in the macro. It refers to a "get more info below in the extensions session" but the variable extension session does not talk about this keyword again.

Also, the call:
```julia
@variable(m, y(3)) #or @variable(m, y())
```
currently triggers:
```julia
ERROR: LoadError: MethodError: no method matching parse_variable(::JuMP.var"#_error#110"{LineNumberNode}, ::JuMP._VariableInfoExpr, ::Symbol)
Closest candidates are:
  parse_variable(::Function, ::JuMP._VariableInfoExpr, ::Symbol, ::Number, ::Any) at C:\Users\joaquimgarcia\.julia\packages\JuMP\klrjG\src\macros.jl:1504
  parse_variable(::Function, ::JuMP._VariableInfoExpr, ::Symbol, ::Any, ::Any) at C:\Users\joaquimgarcia\.julia\packages\JuMP\klrjG\src\macros.jl:1486
  parse_variable(::Function, ::JuMP._VariableInfoExpr, ::Any, ::Symbol, ::Any, ::Symbol, ::Any) at C:\Users\joaquimgarcia\.julia\packages\JuMP\klrjG\src\macros.jl:1559
```
But if could suggest the `variable_type` key...

See the 3rd example in this comment: https://github.com/joaquimg/BilevelJuMP.jl/issues/158#issue-1090806911
(the first will be handled in the extension, and the 2nd is the user trying to guess so no special protection needed)

odow:
Closes #2956

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1736,102 +1736,99 @@ function _reorder_parameters(args)
 end
 
 """
-    @variable(model, kw_args...)
-
-Add an *anonymous* variable to the model `model` described by the keyword
-arguments `kw_args` and returns the variable.
-
     @variable(model, expr, args..., kw_args...)
 
 Add a variable to the model `model` described by the expression `expr`, the
-positional arguments `args` and the keyword arguments `kw_args`. The expression
-`expr` can either be (note that in the following the symbol `<=` can be used
-instead of `≤`, the symbol `>=`can be used instead of `≥`, the symbol `in` can be
-used instead of `∈`)
+positional arguments `args` and the keyword arguments `kw_args`.
 
-* of the form `varexpr` creating variables described by `varexpr`;
-* of the form `varexpr ≤ ub` (resp. `varexpr ≥ lb`) creating variables described by
-  `varexpr` with upper bounds given by `ub` (resp. lower bounds given by `lb`);
-* of the form `varexpr == value` creating variables described by `varexpr` with
-  fixed values given by `value`; or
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -1736,102 +1736,99 @@ function _reorder_parameters(args)
...
+ * `Symmetric`: Only available when creating a square matrix of variables, i.e.,
+   when `expr` is of the form `varname[1:n,1:n]` or `varname[i=1:n,j=1:n]`,
+   it creates a symmetric matrix of variables.
+ * `PSD`: A restrictive extension to `Symmetric` which constraints a square
+   matrix of variables to `Symmetric` and constrains to be positive
+   semidefinite.
+
+## Keyword arguments
 
 The recognized keyword arguments in `kw_args` are the following:
```
joaquimg, line 1794: Maybe add that the kwargs are especially useful for the anon variables case.

joaquimg (approved):

src/macros.jl
odow: How's this now?

joaquimg (approved):

src/macros.jl
joaquimg: great

PR #2953, opened 2022-04-13
WIP: explore API for multivariate hessians of user-defined functions

odow:
This is going to need some testing/performance evaluations.

But if we get get rid of the overhead, this is a much more extensible way of evaluating the hessians. It also doesn't rely on ForwardDiff, which is nice.

Would solve the magic issue: https://github.com/jump-dev/JuMP.jl/issues/1198

PR #2961, opened 2022-04-26
Add support for user-defined multivariate hessians

### LINKED ISSUE #1198: Feature Request: Second Derivatives for User Defined Functions
It would be extremely helpful for JuMP to support second derivatives for user-defined functions.  Ideally this could be done as efficiently as ReverseDiffSparse, but even just calling ForwardDiff.Hessian! would be a helpful option.  There are a broad class of problems that require optimizing functions that do not easily translate into the typical JuMP syntax (especially in ML/Statistics), so having JuMP able to handle such cases would be a huge benefit to people working on those problems and will greatly expand the number of potential JuMP users.

odow:
Refactor of #2953 

Requires https://github.com/jump-dev/MathOptInterface.jl/pull/1819

Closes #1198

### DIFF
--- a/docs/src/manual/nlp.md
+++ b/docs/src/manual/nlp.md
@@ -462,8 +462,9 @@ register(model, :rosenbrock, 2, f, ∇f, ∇²f)
     Make sure the first argument to `∇²f` supports an `AbstractMatrix`, and do
     not assume the input is `Float64`. You may assume the matrix is initialized
     with zeros, so you need only to fill in the non-zero terms. The matrix type
-    passed in as `H` is very limited. You may assume only that it supports
-    `size(H)` and `setindex!`.
+    passed in as `H` depends on the automatic differentiation system, so it may
+    be something other than `Matrix{Float64}`. You may assume only that it
+    supports `size(H)` and `setindex!`.
 
 ### User-defined functions with vector inputs
 
--- a/docs/src/manual/nlp.md
+++ b/docs/src/manual/nlp.md
@@ -459,12 +459,14 @@ register(model, :rosenbrock, 2, f, ∇f, ∇²f)
 ```
 
 !!! warning
-    Make sure the first argument to `∇²f` supports an `AbstractMatrix`, and do
-    not assume the input is `Float64`. You may assume the matrix is initialized
-    with zeros, so you need only to fill in the non-zero terms. The matrix type
-    passed in as `H` depends on the automatic differentiation system, so it may
-    be something other than `Matrix{Float64}`. You may assume only that it
-    supports `size(H)` and `setindex!`.
+    You may assume the Hessian matrix `H` is initialized with zeros, and because
+    `H` is symmetric, you need only to fill in the non-zero of the
+    lower-triangular terms. The matrix type passed in as `H` depends on the
+    automatic differentiation system, so make sure the first argument to the
+    Hessian function supports an `AbstractMatrix` (it may be something other
+    than `Matrix{Float64}`). However, you may assume only that `H` supports
+    `size(H)` and `setindex!`. Finally, the matrix is treated as dense, so the
+    performance will be poor on functions with high-dimensional input.
 
 ### User-defined functions with vector inputs
 
--- a/docs/src/manual/nlp.md
+++ b/docs/src/manual/nlp.md
@@ -357,12 +357,6 @@ The above code creates a JuMP model with the objective function
     and not `f` in the macros.
 
 !!! warning
-    If you use multi-variate user-defined functions, JuMP will disable
-    second-derivative information. This can lead to significant slow-downs in
-    some cases. Only use a user-defined function if you cannot write out the
-    expression algebraically in the macro.
-
-!!! warning
     User-defined functions cannot be re-registered and will not update if you
     modify the underlying Julia function. If you want to change a user-defined
     function between solves, rebuild the model or use a different name. To use
@@ -417,9 +411,11 @@ register(model, :my_square, 2, f, ∇f)
 
 ### Register a function, gradient, and hessian
 
-!!! warning
-    The ability to explicitly register a hessian is only available for
-    univariate functions.
+You can also register a function with the second-order derivative information,
+which is a scalar for univariate functions, and a symmetric matrix for
+multivariate functions.
+
+#### Univariate functions
 
 Instead of automatically differentiating the hessian, you can instead pass a
 function which returns a number representing the second-order derivative.
@@ -434,6 +430,40 @@ register(model, :my_square, 1, f, ∇f, ∇²f)
 @variable(model, x >= 0)
 @NLobjective(model, Min, my_square(x))
 ```
+
+#### Multivariate functions
+
+For multivariate functions, the hessian function `∇²f` must take an
+`AbstractMatrix` as the first argument, the lower-triangular of which is filled
+in-place:
+```@example
+using JuMP #hide
+f(x...) = (1 - x[1])^2 + 100 * (x[2] - x[1]^2)^2
+function ∇f(g, x...)
+    g[1] = 400 * x[1]^3 - 400 * x[1] * x[2] + 2 * x[1] - 2
+    g[2] = 200 * (x[2] - x[1]^2)
+    return
+end
+function ∇²f(H, x...)
+    H[1, 1] = 1200 * x[1]^2 - 400 * x[2] + 2
+    # H[1, 2] = -400 * x[1]  <-- Not needed. Fill the lower-triangular only.
+    H[2, 1] = -400 * x[1]
+    H[2, 2] = 200.0
+    return
+end
+
+model = Model()
+register(model, :rosenbrock, 2, f, ∇f, ∇²f)
+@variable(model, x[1:2])
+@NLobjective(model, Min, rosenbrock(x[1], x[2]))
+```
+
+!!! warning
+    Make sure the first argument to `∇²f` supports an `AbstractMatrix`, and do
+    not assume the input is `Float64`. You may assume the matrix is initialized
+    with zeros, so you need only to fill in the non-zero terms. The matrix type
+    passed in as `H` is very limited. You may assume only that it supports
+    `size(H)` and `setindex!`.
 
 ### User-defined functions with vector inputs
 
--- a/docs/src/tutorials/nonlinear/user_defined_hessians.jl
+++ b/docs/src/tutorials/nonlinear/user_defined_hessians.jl
@@ -20,17 +20,63 @@
 
 # # User-defined Hessians
 
-# In this tutorial, we explain how to write a user-defined function (see [User-defined Functions](@ref))
-# with a Hessian matrix explicitly provided by the user.
+# In this tutorial, we explain how to write a user-defined function (see
+# [User-defined Functions](@ref)) with a Hessian matrix explicitly provided by
+# the user.
 
 # This tutorial uses the following packages:
 
 using JuMP
 import Ipopt
 
-# ## Motivation
+## Rosenbrock example
 
-# Our goal for this tutorial is to solve the bilevel optimization problem:
+# As a simple example, we first consider the Rosenbrock function:
+
+rosenbrock(x...) = (1 - x[1])^2 + 100 * (x[2] - x[1]^2)^2
+
+# which has the gradient vector:
+
+function ∇rosenbrock(g::AbstractVector, x...)
+    g[1] = 400 * x[1]^3 - 400 * x[1] * x[2] + 2 * x[1] - 2
+    g[2] = 200 * (x[2] - x[1]^2)
+    return
+end
+
+# and the Hessian matrix:
+
+function ∇²rosenbrock(H::AbstractMatrix, x...)
+    H[1, 1] = 1200 * x[1]^2 - 400 * x[2] + 2
+    # H[1, 2] = -400 * x[1] <-- not needed because Hessian is symmetric
+    H[2, 1] = -400 * x[1]
+    H[2, 2] = 200.0
+    return
+end
+
+# You may assume the Hessian matrix `H` is initialized with zeros, and
+# because it is symmetric you need only to fill in the non-zero of the
+# lower-triangular terms.
+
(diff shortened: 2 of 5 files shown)


odow:
I added a tutorial. Preview is here: https://jump.dev/JuMP.jl/previews/PR2961/tutorials/nonlinear/user_defined_hessians/

I think this is quite nice.

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+
+# ## Decomposition
+
+# There are a few ways to solve this problem, but we are going to use a
+# nonlinear decomposition method. The first step is to write a function to
+# compute:
+
+# ```math
+# \begin{array}{r l}
+#   V(x_1, x_z) = \max\limits_{y} & x_1^2 y_1 + x_2^2 y_2  - x_1 y_1^4 - 2 x_2 y_2^4 \\
```
noahrhodes, line 61: x_z seems like a typo, should it be x_2?

odow, line 61:
```suggestion
#   V(x_1, x_2) = \max\limits_{y} & x_1^2 y_1 + x_2^2 y_2  - x_1 y_1^4 - 2 x_2 y_2^4 \\
```

odow: I must be going blind. That's a nasty one to spot.

docs/src/manual/nlp.md
```diff
@@ -435,6 +431,40 @@ register(model, :my_square, 1, f, ∇f, ∇²f)
...
+register(model, :rosenbrock, 2, f, ∇f, ∇²f)
+@variable(model, x[1:2])
+@NLobjective(model, Min, rosenbrock(x[1], x[2]))
+```
+
+!!! warning
+    Make sure the first argument to `∇²f` supports an `AbstractMatrix`, and do
+    not assume the input is `Float64`. You may assume the matrix is initialized
+    with zeros, so you need only to fill in the non-zero terms. The matrix type
+    passed in as `H` is very limited. You may assume only that it supports
```
frapac, line 465: Suggestion: rephrase 
```
The matrix type passed in as `H` is very limited.
```
with 
```
By default, the Hessian function supports a broad range of `AbstractMatrix` format, as long as it supports `size(H)` and `setindex!`.
```

test/nlp.jl
```diff
@@ -1095,12 +1095,16 @@ function test_hessians_user_defined_multivariate_functions()
...
-        register(model, :my_f, 2, my_f, my_∇f, my_∇²f),
-    )
+    register(model, :my_f, 2, my_f, my_∇f, my_∇²f)
+    @variable(model, x[1:2])
+    @NLobjective(model, Min, my_f(x[1], x[2]))
+    evaluator = NLPEvaluator(model)
+    MOI.initialize(evaluator, [:Hess])
+    H = MOI.hessian_lagrangian_structure(evaluator)
+    @test H == [(1, 1), (2, 2), (2, 1)]
+    H_nz = fill(NaN, 3)
```
frapac, line 1105: Nice to test with an initial `H_nz` filled with `NaN` :+1:

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE    #src
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER         #src
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,  #src
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  #src
+# SOFTWARE.                                                                      #src
+
+# # User-defined Hessians
+
+# In this tutorial, we explain how to write a user-defined function (see [User-defined Functions](@ref))
+# with an explicit Hessian matrix.
```
frapac, line 24: Suggestion: `with a Hessian matrix explicitly provided by the user`

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+using JuMP
+import Ipopt
+
+# ## Motivation
+
+# Our goal for this tutorial is to solve the bilevel optimization problem:
+
+# ```math
+# \begin{array}{r l}
+# \min\limits_{x} & x_1^2 + x_2^2 + z \\
```
frapac, line 37: Aren't we minimizing wrt `z`? 
```
\min\limits_{x, z}
```

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+# involving variables ``y``. From the perspective of the lower-level, the
+# values of ``x`` are fixed parameters, and so the model optimizes ``y`` given
+# those fixed parameters. Simultaneously, the upper level is optimizing ``x``
+# given the response of ``y``.
+
+# ## Decomposition
+
+# There are a few ways to solve this problem, but we are going to use a
+# nonlinear decomposition method. The first step is to write a function to
+# compute:
```
frapac, line 57: Suggestion:
```
to compute the lower level problem:
```

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+        Max,
+        x[1]^2 * y[1] + x[2]^2 * y[2] - x[1] * y[1]^4 - 2 * x[2] * y[2]^4,
+    )
+    @constraint(model, (y[1] - 10)^2 + (y[2] - 10)^2 <= 25)
+    optimize!(model)
+    @assert termination_status(model) == LOCALLY_SOLVED
+    return objective_value(model), value.(y)
+end
+
+# This function takes a guess of ``x`` and returns the optimal lower-level
```
frapac, line 81: Suggestion: `a guess` -> `an initial value`

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+# \min\limits_{x} & x_1^2 + x_2^2 + V(x_1, x_2) \\
+# s.t.            & x \ge 0.
+# \end{array}
+# ```
+
+# This looks like a nonlinear optimization problem with a user-defined function
+# ``V``! However, because ``V`` solves an optimization problem internally, we
+# can't use automatic differentiation to compute the first and second
+# derivatives. Instead, we can use JuMP's ability to pass callback functions
+# for the gradient and hessian instead.
```
frapac, line 104: Suggestion: `hessian` -> `Hessian`

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+# to ``x``, giving:
+
+function ∇V(g::AbstractVector, x...)
+    _, y = solve_lower_level(x...)
+    g[1] = 2 * x[1] * y[1] - y[1]^4
+    g[2] = 2 * x[2] * y[2] - 2 * y[2]^4
+    return
+end
+
+# Second, we need to define the hessian of ``V`` with respect to ``x``. This is
```
frapac, line 118: Suggestion: `hessian` -> `Hessian`

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+# To compute the optimal lower-level decision variables, we need to call
+# `solve_lower_level` with the optimal upper-level decision variables:
+
+_, y = solve_lower_level(value.(x)...)
+y
+
+# ## Memoization
+
+# Our solution approach works, but it has a performance problem: every time
+# we need to compute the value, gradient, or hessian of ``V``, we have to
```
frapac, line 155: Suggestion: `hessian` -> `Hessian`

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+# re-solve the lower-level optimization problem! This is wasteful, because we
+# will often call the gradient and hessian at the same point, and so solving the
+# problem twice with the same input repeats work unnecessarily.
+
+# We can work around this by using memoization:
+
+function memoized_solve_lower_level()
+    last_x, f, y = nothing, NaN, [NaN, NaN]
+    function _update_if_needed(x...)
+        if last_x != x
```
frapac, line 165: Would it be better to use `last_x !== x` as we store the reference to `last_x`?

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+        return
+    end
+    return memoized_f, memoized_∇f, memoized_∇²f
+end
+
+f, ∇f, ∇²f = memoized_solve_lower_level()
+
+# The function above is a little confusing, but it returns three new functions
+# `f`, `∇f`, and `∇²f`, each of which call `_update_if_needed(x...)`. This
+# function only updates the cached values of `f` and `y` if the input `x` is
```
frapac, line 194: Suggestion: `the cached values of the objective `f` and the primal variables `y` if ...`

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+# different to what is last saw.
```
frapac, line 195: Suggestion: `to what is last saw` -> `to its previous value`

frapac (approved):
It overall looks good to me, apart of a few minor comments. I must say that the tutorial is very nice. 

Also: would it be possible to add a warning telling explicitly that sparse Hessian are not supported in user-registered function?

odow:
I made your changes.

> Also: would it be possible to add a warning telling explicitly that sparse Hessian are not supported in user-registered function?

I don't know if we need to warn that something doesn't exist. There's no way for the user to specify a sparse hessian. I say we let this go out into the wild with little fanfare and see how people end up using it, and whether it makes a performance difference.

docs/src/manual/nlp.md
odow: I changed the wording slightly, it's not that the Hessian function supports a broad range, it's that the AD system passes in something that acts like a matrix but isn't really, so user functions are limited in what they can do.

odow:
I'll plan to merge this in a few days unless anyone wants to review beforehand. At the JuMP side, this is a pretty minor change. It's also opt-in, so it shouldn't affect anyone.

mlubin:
I'd like to take a look at this, but it may be a couple days before I get a chance.

docs/src/manual/nlp.md
```diff
@@ -435,6 +431,41 @@ register(model, :my_square, 1, f, ∇f, ∇²f)
...
+```
+
+!!! warning
+    Make sure the first argument to `∇²f` supports an `AbstractMatrix`, and do
+    not assume the input is `Float64`. You may assume the matrix is initialized
+    with zeros, so you need only to fill in the non-zero terms. The matrix type
+    passed in as `H` depends on the automatic differentiation system, so it may
+    be something other than `Matrix{Float64}`. You may assume only that it
+    supports `size(H)` and `setindex!`.
+
```
mlubin, line 468: Add a warning that the hessian matrix is treated as dense, so the performance will be poor on functions with high-dimensional input.

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+        H[1, 1] = 2 * y[1]
+        H[2, 2] = 2 * y[2]
+        return
+    end
+    return memoized_f, memoized_∇f, memoized_∇²f
+end
+
+f, ∇f, ∇²f = memoized_solve_lower_level()
+
+# The function above is a little confusing, but it returns three new functions
```
mlubin, line 192: To make the function less confusing, what about storing the state in a struct instead of a closure?

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,210 @@
...
+# with a Hessian matrix explicitly provided by the user.
+
+# This tutorial uses the following packages:
+
+using JuMP
+import Ipopt
+
+# ## Motivation
+
+# Our goal for this tutorial is to solve the bilevel optimization problem:
```
mlubin, line 33: It might help to present a simpler case first, like one that doesn't require understanding bilevel optimization.

odow:
How's this now?

docs/src/tutorials/nonlinear/user_defined_hessians.jl
```diff
@@ -0,0 +1,264 @@
...
+# We can work around this by using [memoization](https://en.wikipedia.org/wiki/Memoization).
+# First, we create a struct to cache the results:
+
+mutable struct Cache
+    x::Any
+    f::Float64
+    y::Vector{Float64}
+end
+
+cache = Cache(Float64[], NaN, Float64[])
```
mlubin, line 215: Untyped global variables :(
We don't want people following this example.
How about
```jl
function cached_f(cache, x...)
   ...
end
...
cache = Cache(...)
register(model, :V, 2, (x...) -> cached_f(cache, x...), ...)
```

odow: I considered adding a `const`, but this would also work.

mlubin: The global var version is very limiting, I'd rather push users in a better direction.

mlubin (approved):

PR #2962, opened 2022-04-26
Update installation.md

braisgonzalezusc:
Add RAPOSa solver to the list of solvers:

RAPOSa solver is compatible with JuMP, as it can be seen on its website https://raposa.usc.es/requirements.html, using AmplNLWriter. Regarding the license, RAPOSa is free to use but not open source, so it has a specific license, which can be seen when downloading the executable from its website.


odow (approved):
Nice!

PR #2964, opened 2022-04-26
[docs] clarify primal_feasibility_report for infeasible models

### LINKED ISSUE #2960: Primal Feasibility Report for Infeasible Models
I would like to be able to call `primal_feasibility_report` on an infeasible model and get a primal feasibility report. 

The docs show calling it on a feasible model....and it returns nothing, which isn't useful. But it would be useful to call it on an infeasible solution and figure out which constraints need to be changed and by how much. 

It might be possible to compute this feasibility report using `compute_conflict!` and some manual reasoning about the constraints, but it seems to me the functionality already exists inside `primal_feasibility_report`. Is there a way to use the two calls together to get what I want?

odow:
Closes #2960

@chelseas how is this?

### DIFF
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -625,9 +625,19 @@ To use the primal solution from a solve, omit the `point` argument:
 ```jldoctest feasibility
 julia> optimize!(model)
 
-julia> primal_feasibility_report(model)
+julia> primal_feasibility_report(model; atol = 0.0)
 Dict{Any, Float64}()
 ```
+Calling [`primal_feasibility_report`](@ref) without the `point` argument is most
+useful when the solver reports that the solutionn is feasible, but you want to
+check for violated constraints at a different tolerance level.
+
+!!! warning
+    If the problem is infeasible, calling [`primal_feasibility_report`](@ref) is
+    usually the wrong thing to do, because most solvers will not return an
+    infeasible point. To diagnose the source of infeasibility, consider using
+    [Conflicts](@ref) (for supported solvers), or comment out constraints
+    one-by-one until the problem becomes feasible.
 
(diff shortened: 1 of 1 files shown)


docs/src/manual/solutions.md
```diff
@@ -625,9 +625,19 @@ To use the primal solution from a solve, omit the `point` argument:
 ```jldoctest feasibility
 julia> optimize!(model)
 
-julia> primal_feasibility_report(model)
+julia> primal_feasibility_report(model; atol = 0.0)
 Dict{Any, Float64}()
 ```
+Calling [`primal_feasibility_report`](@ref) without the `point` argument is most
+useful when the solver reports that the solutionn is feasible, but you want to
```
mlubin, line 632: Typo: solution

mlubin, line 632: I'm not sure this is true. It could also be used for first-order solvers that return approximate solutions or stopped early with unknown solution quality.

docs/src/manual/solutions.md
```diff
@@ -625,9 +625,19 @@ To use the primal solution from a solve, omit the `point` argument:
...
+check for violated constraints at a different tolerance level.
+
+!!! warning
+    If the problem is infeasible, calling [`primal_feasibility_report`](@ref) is
+    usually the wrong thing to do, because most solvers will not return an
```
mlubin, line 637: "wrong thing to do" is too opinionated. There are plenty of scenarios where it makes sense, e.g., if you have a point in mind that you think should be feasible calling `primal_feasibility_report` is the right thing to do to debug the infeasibility.

 I'd say: To apply `primal_feasibility_report` to infeasible models, you must also provide a candidate point (solvers generally do not provide one). For diagnosing the cause of an infeasible model, see also Conflicts.

mlubin (approved):

PR #2965, opened 2022-04-28
[docs] expand example of PackageCompiler.jl

odow:
This section is taken from a paper we're writing on JuMP 1.0. I thought it was nice.

### DIFF
--- a/docs/src/tutorials/getting_started/performance_tips.jl
+++ b/docs/src/tutorials/getting_started/performance_tips.jl
@@ -76,6 +76,7 @@ model = Model(HiGHS.Optimizer)
 optimize!(model)
 open("model.log", "w") do io
     print(io, solution_summary(model; verbose = true))
+    return
 end
 
 # Saving the problem in `model.jl` and calling from the command line results in:
--- a/docs/src/tutorials/getting_started/performance_tips.jl
+++ b/docs/src/tutorials/getting_started/performance_tips.jl
@@ -63,12 +63,57 @@ model = Model(HiGHS.Optimizer; add_bridges = false)
 
 # ### Use PackageCompiler
 
-# As a final option, consider using [PackageCompiler.jl](https://julialang.github.io/PackageCompiler.jl/dev/)
-# to [create a custom sysimage](https://julialang.github.io/PackageCompiler.jl/dev/examples/plots/).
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/getting_started/performance_tips.jl
```diff
@@ -63,12 +63,58 @@ model = Model(HiGHS.Optimizer; add_bridges = false)
...
+@constraint(model, c2, 7x + 12y >= 120)
+optimize!(model)
+open("model.log", "w") do io
+    print(io, solution_summary(model; verbose = true))
+    return
+end
+
+# Saving the problem in `model.jl` and calling from the command line results in:
+# ```
+# \$ time julia model.jl
```
odow, line 84:
```suggestion
# $ time julia model.jl
```

docs/src/tutorials/getting_started/performance_tips.jl
```diff
@@ -63,12 +63,58 @@ model = Model(HiGHS.Optimizer; add_bridges = false)
...
+# PackageCompiler.create_sysimage(
+#     ["JuMP", "HiGHS"],
+#     sysimage_path = "customimage." * Libdl.dlext,
+#     precompile_execution_file = "model.jl",
+# )
+# ```
+# When Julia is run with the custom image, the run time is now 0.7 seconds
+# instead of 16:
+# ```
+# \$ time julia --sysimage customimage model.jl
```
odow, line 108:
```suggestion
# $ time julia --sysimage customimage model.jl
```

docs/src/tutorials/getting_started/performance_tips.jl
```diff
@@ -63,12 +63,58 @@ model = Model(HiGHS.Optimizer; add_bridges = false)
...
+# ```
+# Other performance tweaks, such as disabling bridges or using direct mode can
+# reduce this time futher.
 
-# This is a good option if you have finished prototyping a model, and you now
-# want to call it frequently from the command line without paying the
-# compilation price.
+# !!! note
+#     `create_sysimage` only needs to be run once, and the same sysimage can be
+#     used---to a slight detriment of performance---even if we modify
```
odow, line 116:
```suggestion
#     used--to a slight detriment of performance--even if we modify
```

PR #2966, opened 2022-04-28
Fix error message when broadcasting bounds

### LINKED ISSUE #2944: Vectorized bounds on variables
It would be nice to be able to do the following to set bounds on variable arrays (or other variable containers):

```julia
@variable(m, [1,2] .<= x[1:2] .<= [3,4])
```

This currently throws a `LoadError` indicating that `<=` should be used instead. Using `<=` instead of the vectorized inequality instructs the user with "Instead of `@variable(model, x[1:2] >= lb)`, do `@variable(model, x[i=1:2] >= lb[i])`. Alternatively, create the variable without bounds, then call `set_lower_bound.(x, lb)`." This makes perfect sense. Would it make sense to add the capability of passing arrays as variable bounds to generate bounded variables on the fly?

odow:
Closes #2944

cc @hdavid16

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1439,34 +1439,77 @@ function build_variable(
         )
     end
     if info.lower_bound isa AbstractArray
-        _error(
-            "Passing arrays as variable bounds is not supported. Instead of " *
-            "`@variable(model, x[1:2] >= lb)`, do " *
-            "`@variable(model, x[i=1:2] >= lb[i])`. Alternatively, create " *
-            "the variable without bounds, then call `set_lower_bound.(x, lb)`",
-        )
+        _error("""
+               Passing arrays as variable bounds is not supported.
+
+               Instead of:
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -1439,34 +1439,77 @@ function build_variable(
...
-            "`@variable(model, x[1:2] >= lb)`, do " *
-            "`@variable(model, x[i=1:2] >= lb[i])`. Alternatively, create " *
-            "the variable without bounds, then call `set_lower_bound.(x, lb)`",
-        )
+        _error("""
+               Passing arrays as variable bounds is not supported.
+
+               Instead of:
+               ```julia
+               @variable(model, x[1:2] >= lb)
```
joaquimg, line 1447:
```suggestion
               @variable(model, x[1:2] .>= lb)
```

PR #2972, opened 2022-05-05
[docs] clarify modify-then-query in direct mode

### LINKED ISSUE #2970: [Docs, Direct mode] Solver status not updated after problem solved-then-modified
I don't know whether this is intended behavior; at least ~I could not find it documented~ (see edit below).
It seems that, when using direct mode, certain attributes such as `termination_status` and `objective_value` return erroneous values in the example below.

The overall workflow is:
1. Create a JuMP direct model, add some variables, constraints, etc.
1. Solve the model via `optimize!`
1. Add a new constraint to the model
1. **do not** call `optimize!`
1. Call `terminations_status`, `objective_value`, and `value` of some variables

Step 4 above is stupid, but may happen if, e.g., someone forgets to re-optimize their model.
Then, you would expect that step 5. would return `OPTIMIZE_NOT_CALLED` and, accordingly, that `objective_value` and `value` would throw and `OptimizeNotCalled` error. This is what happens in normal mode.

In direct mode, however, the termination status at step 5 is `OPTIMAL`, calling `objective_value` does not throw any error and returns a different value between solvers (HiGHS returns `0.0`, Gurobi and GPLK return the previous value).

___

[EDIT] after I wrote all of this, I found the [last paragraph of `OptimizeNoCalled` docs](https://jump.dev/JuMP.jl/stable/manual/solutions/#OptimizeNotCalled-errors) that appears to describe the behavior I experienced below. There remains the odd behavior of HiGHS who returns `0.0` as objective value after modifying the problem.
Would it be acceptable to add a pointer to the present behavior in the [direct mode docs](https://jump.dev/JuMP.jl/stable/manual/models/#Direct-mode)?

___

Normal mode:
```julia
using JuMP, HiGHS

# Create a simple model
set_silent(model)
@variable(model, x)
@constraint(model, x >= 1)
@objective(model, Min, 1*x)

# Solve and grab the solution
optimize!(model)
@show termination_status(model)  # outputs MOI.OPTIMAL
@show objective_value(model)     # outputs 1.0
@show value(x)                   # outputs 1.0

# Modify the problem
@constraint(model, x >= 2)
# ... and query the result immediately after
@show termination_status(model)  # outputs MOI.OPTIMIZE_NOT_CALLED
@show objective_value(model)     # throws ERROR: OptimizeNotCalled()
@show value(x)                   # throws ERROR: OptimizeNotCalled()
```

Direct mode (here uses HiGHS; I got similar behavior with Gurobi and GLPK)
```julia
using JuMP, HiGHS

# Create a simple model
model = JuMP.direct_model(HiGHS.Optimizer())
set_silent(model)
@variable(model, x)
@constraint(model, x >= 1)
@objective(model, Min, 1*x)

# Solve and grab the solution
optimize!(model)
@show termination_status(model)  # outputs MOI.OPTIMAL
@show objective_value(model)     # outputs 1.0
@show value(x)                   # outputs 1.0

# Modify the problem
@constraint(model, x >= 2)
# ... and query the result immediately after
@show termination_status(model)  # outputs MOI.OPTIMAL
@show objective_value(model)     # outputs 0.0 (Gurobi and GLPK: 1.0)
@show value(x)                   # outputs 1.0
```

package versions:
```
  [4076af6c] JuMP v1.0.0
  [60bf3e95] GLPK v1.0.1
  [2e9cd046] Gurobi v0.10.1
  [87dc4568] HiGHS v1.1.3
```

odow:
Closes #2970

cc @mtanneau

### DIFF
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -552,3 +552,9 @@ If you expected the solver to support your problem, you may have an error in you
 The list of available solvers, along with the problem types they support, is available at https://jump.dev/JuMP.jl/stable/installation/#Supported-solvers.
 [...]
 ```
+
+!!! warning
+    Another downside of direct mode is that the behavior of querying solution
+    information after modifying the problem is solver-specific. This can lead to
+    errors, or the solver silently returning an incorrect value. See
+    [OptimizeNotCalled errors](@ref) for more information.
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -372,6 +372,11 @@ julia> termination_status(model)
 OPTIMAL::TerminationStatusCode = 1
 ```
 
+!!! warning
+    Be careful doing this! If your particular solver does not support
+    querying solution information after modification, it may silently return
+    incorrect solutions or throw an error.
+
 ```@meta
 # TODO: How to accurately measure the solve time.
 ```


mtanneau:
Wow, that was fast! Thanks!

PR #2978, opened 2022-05-08
Add set_string_names_on_creation

### LINKED ISSUE #2973: Document cost of named variables and constraints in Performance Tips
When building a model with millions of variables and constraints, I see model build time roughly half when switching to using anonymous variables and constraints. 

An example model has size
```jl
Variables: 2574376 
`AffExpr`-in-`EqualTo{Float64}`: 1097232 constraints
`AffExpr`-in-`GreaterThan{Float64}`: 315984 constraints
`AffExpr`-in-`LessThan{Float64}`: 1445558 constraints
`VariableRef`-in-`GreaterThan{Float64}`: 1721368 constraints
`VariableRef`-in-`LessThan{Float64}`: 40032 constraints
`VariableRef`-in-`ZeroOne`: 40032 constraints 
```

And an example code change would be:
```diff
-        @variable(model, foo[z in zones, t in times] >= 0)
+        model[:foo] = foo = @variable(model, [z in zones, t in times], lower_bound=0)
```

Changing all `@variable` and `@constraint` calls similarly, I see differences in model build time like below:

|  | time | allocs | mem |
| ------ | ------ | ------ | ------ |
| names | 36.214 s | 131.82 M | 7.19 GiB |
|      no names | 16.652 s | 67.09 M | 4.05 GiB |
| relative |  2.175x | 1.965x | 1.775x |

Depending on the model we're building, the time speed up varies (~1.3 - 2x) but the reduction of allocations is pretty consistently about a ~2x reduction.

The different types of "names" and the behaviour of the macros regarding names is documented
- for variables: https://jump.dev/JuMP.jl/stable/manual/variables/#String-names,-symbolic-names,-and-bindings
- for constraints: https://jump.dev/JuMP.jl/stable/manual/constraints/#String-names,-symbolic-names,-and-bindings

So actually making this change was fairly easy. And this discovery is a very welcome performance improvement!

However, i don't think this significant performance consideration is mentioned anywhere in the docs. And in particular it does't seem to be mentioned in [the Performance Tips](https://jump.dev/JuMP.jl/stable/tutorials/getting_started/performance_tips/)

I think the overhead of named variables/constraints is known (xref https://github.com/jump-dev/JuMP.jl/issues/2817#issuecomment-993085257), so I think worth documenting given it can be a significant build-time cost and is avoidable with relatively small user code changes.

(p.s. i'm opening this just before some time off so may not be quick to respond)

odow:
Closes #2973 

TODO
- [x] Document
- [x] Consider doing same to `@constraint`

### DIFF
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -294,6 +294,18 @@ julia> x
  my_var[2]
 ```
 
+!!! tip
+    For models with millions of variables, setting the string name of each
+    variable can take a significant portion of the total time required to build
+    the model. Turn off `String` names by passing `set_string_name = false` to
+    [`@variable`](@ref):
+    ```jldoctest
+    julia> model = Model();
+
+    julia> @variable(model, x, set_string_name = false)
+    _[1]
+    ```
+
 ### Retrieve a variable by name
 
 Retrieve a variable from a model using [`variable_by_name`](@ref):
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -295,16 +295,16 @@ julia> x
 ```
 
 !!! tip
-    For models with millions of variables, setting the string name of each
-    variable can take a significant portion of the total time required to build
-    the model. Turn off `String` names by passing `set_string_name = false` to
-    [`@variable`](@ref):
+    For some models, setting the string name of each variable can take a
+    non-trivial portion of the total time required to build the model. Turn off
+    `String` names by passing `set_string_name = false` to [`@variable`](@ref):
     ```jldoctest
     julia> model = Model();
 
     julia> @variable(model, x, set_string_name = false)
     _[1]
     ```
+    See [Disable string names](@ref) for more information.
 
 ### Retrieve a variable by name
 
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -347,7 +347,7 @@ julia> model[:x]
  my_var[2]
 ```
 
-## String names, symbolic names, and bindings
+## [String names, symbolic names, and bindings](@id variable_names_and_bindings)
 
 It's common for new users to experience confusion relating to JuMP variables.
 Part of the problem is the overloaded use of "variable" in mathematical
--- a/docs/src/tutorials/getting_started/performance_tips.jl
+++ b/docs/src/tutorials/getting_started/performance_tips.jl
@@ -179,7 +179,7 @@ model = Model();
 
 @constraint(model, c, 2x <= 1, set_string_name = false)
 
-# Note that this doesn't change now symbolic names and bindings are stored:
+# Note that this doesn't change how symbolic names and bindings are stored:
 
 x
 
@@ -192,6 +192,6 @@ model[:x]
 x === model[:x]
 
 # !!! info
-#     For more information on the difference betweene string names, symbolic
+#     For more information on the difference between string names, symbolic
 #     names, and bindings, see
 #     [String names, symbolic names, and bindings](@ref variable_names_and_bindings).
--- a/docs/src/tutorials/getting_started/performance_tips.jl
+++ b/docs/src/tutorials/getting_started/performance_tips.jl
@@ -159,3 +159,39 @@ model = Model()
 # If we use the [`@expression`](@ref) macro, we get many fewer allocations:
 
 @allocated @expression(model, x[1] + x[2] + x[3])
+
+# ## Disable string names
+
+# By default, JuMP creates `String` names for variables and constraints and
+# passes these to the solver. The benefit of passing names is that it improves
+# the readability of log messages from the solver (e.g., "variable x has invalid
+# bounds" instead of "variable v1203 has invalid bounds"), but for larger models
+# the overhead of passing names can be non-trivial.
+
+# Disable the passing of names from JuMP to the solver by setting
+# `set_string_name = false` in the [`@variable`](@ref) and [`@constraint`](@ref)
+# macros.
+
+model = Model();
+@variable(model, x, set_string_name = false)
+
+#-
+
+@constraint(model, c, 2x <= 1, set_string_name = false)
+
+# Note that this doesn't change now symbolic names and bindings are stored:
+
+x
+
+#-
+
+model[:x]
+
+#-
+
+x === model[:x]
+
+# !!! info
+#     For more information on the difference betweene string names, symbolic
+#     names, and bindings, see
+#     [String names, symbolic names, and bindings](@ref variable_names_and_bindings).
--- a/docs/src/tutorials/getting_started/performance_tips.jl
+++ b/docs/src/tutorials/getting_started/performance_tips.jl
@@ -170,10 +170,11 @@ model = Model()
 
 # Disable the creation of `String` names by setting `set_string_name = false` in
 # the [`@variable`](@ref) and [`@constraint`](@ref) macros, or by calling
-# [`set_string_names`](@ref) to disable all names for a particular model:
+# [`set_string_names_on_creation`](@ref) to disable all names for a particular
+# model:
 
 model = Model()
-set_string_names(model, false)
+set_string_names_on_creation(model, false)
 @variable(model, x)
 
 #-
--- a/docs/src/tutorials/getting_started/performance_tips.jl
+++ b/docs/src/tutorials/getting_started/performance_tips.jl
@@ -168,16 +168,17 @@ model = Model()
 # bounds" instead of "variable v1203 has invalid bounds"), but for larger models
 # the overhead of passing names can be non-trivial.
 
-# Disable the passing of names from JuMP to the solver by setting
-# `set_string_name = false` in the [`@variable`](@ref) and [`@constraint`](@ref)
-# macros.
+# Disable the creation of `String` names by setting `set_string_name = false` in
+# the [`@variable`](@ref) and [`@constraint`](@ref) macros, or by calling
+# [`set_string_names`](@ref) to disable all names for a particular model:
 
-model = Model();
-@variable(model, x, set_string_name = false)
+model = Model()
+set_string_names(model, false)
+@variable(model, x)
 
 #-
 
-@constraint(model, c, 2x <= 1, set_string_name = false)
+@constraint(model, c, 2x <= 1)
 
 # Note that this doesn't change how symbolic names and bindings are stored:
 
--- a/docs/src/tutorials/getting_started/performance_tips.jl
+++ b/docs/src/tutorials/getting_started/performance_tips.jl
@@ -193,6 +193,10 @@ model[:x]
 
 x === model[:x]
 
+# But you can no longer look up the variable by the string name:
+
+variable_by_name(model, "x") === nothing
+
 # !!! info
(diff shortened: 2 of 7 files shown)


nickrobinson251:
Thanks! This looks great, and is a much simipler option than the user code changes i mentioned having to make in #2973  :)

I think to close #2973 (from my perspective), this same keyword would need to be available in `@constraint`, so the code change is the same for `@variable` and `@constraint`. And for this option to be mentioned on the Performance Tips page specifically; having this info in the relevant section of the docs is also very helpful, but personally i think this can make big enough performance difference to be worthy of mention on the dedicated Performance Tips page too.

docs/src/tutorials/getting_started/performance_tips.jl
```diff
@@ -159,3 +159,39 @@ model = Model()
...
+# macros.
+
+model = Model();
+@variable(model, x, set_string_name = false)
+
+#-
+
+@constraint(model, c, 2x <= 1, set_string_name = false)
+
+# Note that this doesn't change now symbolic names and bindings are stored:
```
nickrobinson251, line 182: typo?
```suggestion
# Note that this doesn't change how symbolic names and bindings are stored:
```

src/macros.jl
```diff
@@ -874,6 +882,9 @@ The recognized keyword arguments in `kw_args` are the following:
   constraint names are set to `base_name[...]` for each index `...` of the axes
   `axes`.
 * `container`: Specify the container type.
+* `set_string_name::Bool = true`: control whether to set the
+  [`MOI.Constraint`](@ref) attribute. Passing `set_string_name = false` can
```
nickrobinson251, line 886:
```suggestion
  [`MOI.ConstraintName`](@ref) attribute. Passing `set_string_name = false` can
```

docs/src/tutorials/getting_started/performance_tips.jl
```diff
@@ -159,3 +159,39 @@ model = Model()
...
+#-
+
+model[:x]
+
+#-
+
+x === model[:x]
+
+# !!! info
+#     For more information on the difference betweene string names, symbolic
```
nickrobinson251, line 195:
```suggestion
#     For more information on the difference between string names, symbolic
```

odow:
Seems like https://github.com/jump-dev/MutableArithmetics.jl/pull/153 broke our tests.

odow:
Fixing the tests here: https://github.com/jump-dev/MutableArithmetics.jl/pull/156

odow:
@nickrobinson251 how's this now?

nickrobinson251:
Yeah, this looks good. 

I've updated the model building code mentioned in #2973. I don't see the same performance gains using this new syntax, but it's most of the way there so it's almost certainly human-error on my part (i'm looking into it), but just in case: is there anyway using this keyword  like `@constraint(model, con[...], x[...] <= y[...], set_string_names-=false)` wouldn't save as much work as using `model[:con] = @constraint(model, [...], x[...] <= y[...])`?

One thought does come to mind when updating all that code: 
in that use-case, "set_string_names" is something we actually want to set _at the `Model` level_ (not at a per-`@variable`/`@constraint` level) i.e. what would be optimal for that code is if we could write e.g. `model = Model(; set_string_names=false)` and then, for that `model`, `@variable(model, x)` would be equivalent to `@variable(model, x, set_string_name=false)`, and avoids adding `set_string_name=false` to tens of `@variable`/`@constraint` calls in the code-base (and reduces the chance someone forgets to add `set_string_name=false` when adding a new `@variable`/`@constraint`).

so the macro logic would be something like 
```diff
new_name_code = if isempty(set_string_name_kw_args)
-        name_code   
+        model.set_string_names ? name_code : ""
    else
        Expr(:if, esc(set_string_name_kw_args[1].args[2]), name_code, "")
    end
```
(assuming `model` (somehow) stores a `set_string_names::Bool`)

odow:
Hmm. Let me take a look. Admittedly, I didn't benchmark this. 

And yes, as I was writing this PR I considered a model-level flag. But we haven't added any others, so this would be precedent setting. I don't think we need both the model and macro level flags.

odow:
Benchmarking seems to work for me?
```Julia
julia> using JuMP

julia> using BenchmarkTools

julia> function bar(flag::Bool)
           model = Model()
           @variable(model, x[1:1_000_000], set_string_name = flag)
           return model
       end
bar (generic function with 2 methods)

julia> @benchmark foo(true)
BenchmarkTools.Trial: 8 samples with 1 evaluation.
 Range (min … max):  555.151 ms … 813.241 ms  ┊ GC (min … max): 25.19% … 47.18%
 Time  (median):     688.693 ms               ┊ GC (median):    27.57%
 Time  (mean ± σ):   683.801 ms ±  99.854 ms  ┊ GC (mean ± σ):  35.44% ±  9.90%

  █       █ █       █                         ██        █     █  
  █▁▁▁▁▁▁▁█▁█▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁██▁▁▁▁▁▁▁▁█▁▁▁▁▁█ ▁
  555 ms           Histogram: frequency by time          813 ms <

 Memory estimate: 351.84 MiB, allocs estimate: 7000197.

julia> @benchmark foo(false)
BenchmarkTools.Trial: 38 samples with 1 evaluation.
 Range (min … max):  127.087 ms … 140.850 ms  ┊ GC (min … max): 15.42% … 16.45%
 Time  (median):     135.225 ms               ┊ GC (median):    14.96%
 Time  (mean ± σ):   134.352 ms ±   4.170 ms  ┊ GC (mean ± σ):  14.28% ±  2.31%

           █▃       ▃                             ▃ ▃ ▃ ▃        
  ▇▇▁▁▁▁▁▁▇██▁▁▁▇▇▇▇█▁▇▇▁▁▁▇▁▁▁▁▁▁▁▇▇▁▁▇▇▇▁▁▇▇▇▁▁▇█▇█▁█▇█▁▁▇▁▁▇ ▁
  127 ms           Histogram: frequency by time          141 ms <

 Memory estimate: 143.08 MiB, allocs estimate: 2000147.

julia> function bar(flag::Bool)
           model = Model()
           @variable(model, x[1:1_000_000], set_string_name = false)
           @constraint(model, con[i=1:1_000_000], i * x[i] >= i, set_string_name = flag)
           return model
       end
bar (generic function with 2 methods)

julia> @benchmark bar(true)
BenchmarkTools.Trial: 2 samples with 1 evaluation.
 Range (min … max):  2.734 s …   2.826 s  ┊ GC (min … max): 39.93% … 44.46%
 Time  (median):     2.780 s              ┊ GC (median):    42.23%
 Time  (mean ± σ):   2.780 s ± 64.850 ms  ┊ GC (mean ± σ):  42.23% ±  3.21%

  █                                                       █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  2.73 s         Histogram: frequency by time        2.83 s <

 Memory estimate: 1.29 GiB, allocs estimate: 22000227.

julia> @benchmark bar(false)
BenchmarkTools.Trial: 4 samples with 1 evaluation.
 Range (min … max):  1.268 s …    1.518 s  ┊ GC (min … max): 38.18% … 48.50%
 Time  (median):     1.406 s               ┊ GC (median):    45.42%
 Time  (mean ± σ):   1.399 s ± 110.421 ms  ┊ GC (mean ± σ):  44.62% ±  4.63%

  █                   █                      █             █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  1.27 s         Histogram: frequency by time         1.52 s <

 Memory estimate: 1.07 GiB, allocs estimate: 16000177.
```

odow:
How's this?

nickrobinson251 (approved):

docs/src/manual/variables.md
```diff
@@ -294,6 +294,18 @@ julia> x
  my_var[2]
 ```
 
+!!! tip
+    For models with millions of variables, setting the string name of each
```
mlubin, line 298: Why does the number of variables affect the proportion of the time spent in generating the names? Shouldn't the proportion remain constant as the model size increases, all else equal?

docs/src/tutorials/getting_started/performance_tips.jl
```diff
@@ -159,3 +159,40 @@ model = Model()
...
+
+model = Model()
+set_string_names(model, false)
+@variable(model, x)
+
+#-
+
+@constraint(model, c, 2x <= 1)
+
+# Note that this doesn't change how symbolic names and bindings are stored:
```
mlubin, line 183: Perhaps also show that looking up the variable by string name will now give an error

src/JuMP.jl
```diff
@@ -161,6 +161,10 @@ mutable struct Model <: AbstractModel
     # Enable extensions to attach arbitrary information to a JuMP model by
     # using an extension-specific symbol as a key.
     ext::Dict{Symbol,Any}
+    # A model-level option that is used as the default for the set_string_name
+    # keyword to @variable and @constraint. This is a special case because it
```
mlubin, line 165: Unclear what "This" is referring to. Special case of what?

src/JuMP.jl
```diff
@@ -409,6 +414,27 @@ function _try_get_solver_name(model_like)
     end
 end
 
+"""
+    set_string_names(model::Model, value::Bool)
```
mlubin, line 418: I'd recommend `set_string_names_on_creation` for more verbosity. This will appear in advanced code and only in few spots, so it doesn't hurt to be clearer. Same for the name of the field of `Model`.

docs/src/manual/variables.md
odow: I guess so. But it's really only noticeable as you get to larger models.

src/JuMP.jl
odow: Done

src/JuMP.jl
odow: I removed that part of the sentence. I was aiming to describe that this is the first model-level option we've introduced, but perhaps it's not important.

odow:
@nickrobinson251 I think we'd prefer to keep the function approach. It matches other JuMP settings like `set_silent` and `set_time_limit_sec`.

src/JuMP.jl
```diff
@@ -409,6 +413,28 @@ function _try_get_solver_name(model_like)
...
+of reducing the readability of printing and solver log messages.
+"""
+function set_string_names_on_creation(model::Model, value::Bool)
+    model.set_string_names_on_creation = value
+    return
+end
+
+set_string_names_on_creation(model::Model) = model.set_string_names_on_creation
+
+set_string_names_on_creation(::AbstractModel, ::Bool) = nothing
```
odow, line 435:
```suggestion
```

mlubin (approved):

PR #2980, opened 2022-05-12
Update release_notes.md with explicit info

artkuo:
Added explicit info for new names of the deprecated functions. Saves user from searching for this info. Also added which version instituted the deprecation (v0.23.0 and v0.23.1). Again, saves user from searching.


odow (approved):
Thanks! Yes, we could get better at our release notes.

PR #2987, opened 2022-05-17
[docs] add webapp tutorial

odow:
I've had this question a few times, and I think it's useful to show people a mechanism for deploying JuMP models in the documentation. Could still use being fleshed out  a little more, but I just want to see if it runs okay in Documenter.jl.

Preview: https://jump.dev/JuMP.jl/previews/PR2987/tutorials/applications/web_app/


docs/make.jl
```diff
@@ -158,7 +158,10 @@ const _PAGES = [
             "tutorials/algorithms/cutting_stock_column_generation.md",
             "tutorials/algorithms/tsp_lazy_constraints.md",
         ],
-        "Applications" => ["tutorials/applications/power_systems.md"],
+        "Applications" =>
```
odow, line 161:
```suggestion
        "Applications" => [
```

docs/src/tutorials/applications/web_app.jl
```diff
@@ -0,0 +1,142 @@
...
+        ["Content-Type" => "application/json"],
+        JSON.json(data),
+    )
+    if ret.status != 200
+        ## This could happen if there are time-outs, network errors, etc.
+        return Dict(
+            "status" => "failure",
+            "code" => ret.status,
+            "body" => String(ret.body),
+
```
odow, line 114:
```suggestion
```

blegat (approved):
Good idea

docs/src/tutorials/applications/web_app.jl
```diff
@@ -0,0 +1,141 @@
+# Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors    #src
+# This Source Code Form is subject to the terms of the Mozilla Public License   #src
+# v.2.0. If a copy of the MPL was not distributed with this file, You can       #src
+# obtain one at https://mozilla.org/MPL/2.0/.                                   #src
+
+# # Serving web apps
+
+# This tutorial demonstrates how to setup and serve JuMP models via a REST API.
+
+# First, we need the standard JuMP packages:
```
dourouc05, line 10: Is HiGHS a standard JuMP package? I also believe that beginners may be a bit puzzled. I'd rather say "we need JuMP and a solver".

docs/src/tutorials/applications/web_app.jl
```diff
@@ -0,0 +1,141 @@
+# Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors    #src
+# This Source Code Form is subject to the terms of the Mozilla Public License   #src
+# v.2.0. If a copy of the MPL was not distributed with this file, You can       #src
+# obtain one at https://mozilla.org/MPL/2.0/.                                   #src
+
+# # Serving web apps
+
+# This tutorial demonstrates how to setup and serve JuMP models via a REST API.
```
dourouc05, line 8: Maybe this could be made slightly clearer for beginners by saying how the parameters and solutions are transferred. I would add the following: 

> In the example app we are building, we are interested in solving a trivial mixed-integer program with one parameter, a lower bound. To call the service, the users are expected to send an HTTP POST request with JSON contents indicating the lower bound. The returned value is the solution of the MIP as JSON.

dourouc05 (approved):

mlubin:
I want to take a look, wait for my review

docs/src/tutorials/applications/web_app.jl
```diff
@@ -0,0 +1,141 @@
...
+    return HTTP.Response(200, JSON.json(solution))
+end
+
+# Finally, we need an HTTP server. There are a variety of ways you can do this
+# in HTTP.jl. We use an explicit `Sockets.listen` so we have manual control of
+# when we shutdown the server.
+
+function setup_server(host, port)
+    server = HTTP.Sockets.listen(host, port)
+    @async HTTP.serve(host, port; server = server) do request
```
mlubin, line 75: What happens when you get multiple requests in parallel? Does the server run multiple threads?

odow: No. It looks like you need to write the thread handling yourself: https://github.com/JuliaWeb/HTTP.jl/issues/798

docs/src/tutorials/applications/web_app.jl
```diff
@@ -0,0 +1,165 @@
...
+    return server
+end
+
+# !!! info
+#     `@async` runs the server in a background process. If you omit `@async`,
+#     `HTTP.serve` will block the current Julia process.
+
+# !!! warning
+#     HTTP.jl does not serve requests on a separate thread. Therefore, a
+#     long-running will block the main thread, preventing concurrent users from
```
odow, line 109:
```suggestion
#     long-running job will block the main thread, preventing concurrent users from
```

matbesancon (approved):

docs/src/tutorials/applications/web_app.jl
```diff
@@ -0,0 +1,165 @@
...
+    end
+    model = Model(HiGHS.Optimizer)
+    set_silent(model)
+    @variable(model, x >= params["lower_bound"], Int)
+    optimize!(model)
+    return Dict{String,Any}(
+        "status" => "okay",
+        "terminaton_status" => termination_status(model),
+        "primal_status" => primal_status(model),
+        "x" => has_values(model) ? value(x) : NaN,
```
mlubin, line 58: JSON doesn't officially support NaN. This works when reading and writing from Julia, but could be a problem across languages. I think the cleanest workaround is to only set the "x" key in the dictionary if it has a value.

matbesancon: I guess it's an okay solution, the user gets a type error if they haven't checked the status before

odow: JSON.jl writes NaN out as `"null"`:
```Julia
julia> using JSON

julia> JSON.json(NaN)
"null"
```
but yes, I'll change to only include x if it has a value.

mlubin (approved):

Issue #2990, opened 2022-05-25
JuliaFormatter 1.0


odow:
JuliaFormatter went 1.0, which should simplify our formatting actions: https://github.com/domluna/JuliaFormatter.jl/releases/tag/v1.0.0

There are a couple of options we could consider turning on:
 * https://github.com/domluna/JuliaFormatter.jl#separate_kwargs_with_semicolon
   Because it actually makes a difference in places:
   ```Julia
   julia> sum(i for i in 1:0; init = 0)
   0
   
   julia> sum(i for i in 1:0, init = 0)
   ERROR: ArgumentError: reducing over an empty collection is not allowed
   Stacktrace:
   ```
 * https://github.com/domluna/JuliaFormatter.jl#format_docstrings: blocked by https://github.com/domluna/JuliaFormatter.jl/issues/597

odow:
Currently blocked by https://github.com/domluna/JuliaFormatter.jl/issues/603 and https://github.com/MichaelHatherly/CommonMark.jl/issues/43.

odow:
Closing since we've updated to JuliaFormatter v1.0. Formatting the docstrings is a low priority, it's blocked upstream with no ETA, and no one has ever requested it.

PR #2992, opened 2022-05-25
JuliaFormatter: separate_kwargs_with_semicolon

odow:
Part of #2990 

I think this is a reasonable change, because it can only ever reduce confusion.

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -5,4 +5,5 @@ always_for_in = true
 always_use_return = true
 margin = 80
 remove_extra_newlines = true
+separate_kwargs_with_semicolon = true
 short_to_long_function_def = true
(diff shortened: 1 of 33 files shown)


mlubin:
It looks a bit uglier to me, but I guess the disambiguity argument with `sum(i for i in 1:0; init = 0)` is more important.

odow:
The original motivation was https://discourse.julialang.org/t/keyword-behaves-differently-using-or/78233. @ccoffrin and I hit it in some NLP/PowerModels stuff.

odow:
Hmm. I guess I didn't try:
```Julia
julia> using JuliaFormatter

julia> format_text("sum(i for i in 1:0, init = 0)"; separate_kwargs_with_semicolon = true)
"sum(i for i = 1:0, init in 0)"

julia> format_text("sum(i for i in [], init = 0)"; separate_kwargs_with_semicolon = true)
"sum(i for i in [], init in 0)"
```
So it won't fix the `sum` problem, but it will make clear that something is wrong.

odow:
Okay for now. I'll look into whether we can exclude `;` if they are in a function call with kwargs only.

PR #2993, opened 2022-05-26
RFC: JuliaFormatter: use format_docstrings = true

odow:
I think this is probably a win in the long term. It found a few cases where we missed a closing ` ``` `, and there were a few places with incorrect spacing after a list of bullets. 

The downside is the change from `*` to `-` and indentation change, which makes the diff a lot noisier than it would otherwise be. But that's just a penalty to pay once.

I think the additional blank lines in doctests if the line returns `nothing` are a bug?
```julia
julia> model = Model();

julia> unregister(model, :x)

julia> 
```
In my REPL there is only a single blank line, not two.

cc @domluna, that 1.0.2 patch worked.

### DIFF
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -179,6 +179,7 @@ and the given axes. Exactly `N` axes must be provided, and their lengths must
 match `size(data)` in the corresponding dimensions.
 
 # Example
+
 ```jldoctest; setup=:(using JuMP)
 julia> array = JuMP.Containers.DenseAxisArray([1 2; 3 4], [:a, :b], 2:3)
 2-dimensional DenseAxisArray{Int64,2,...} with index sets:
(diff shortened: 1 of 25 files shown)


src/Containers/DenseAxisArray.jl
```diff
@@ -210,9 +211,11 @@ Construct an uninitialized DenseAxisArray with element-type `T` indexed over the
 given axes.
 
 # Example
+
 ```jldoctest; setup=:(using JuMP)
 julia> array = JuMP.Containers.DenseAxisArray{Float64}(undef, [:a, :b], 1:2);
 
+
```
odow, line 218: Here's an example where I don't thing it's needed. Locally, I get:
```julia
julia> array = JuMP.Containers.DenseAxisArray{Float64}(undef, [:a, :b], 1:2);

julia> 
```

odow:
This is blocked by a couple of upstream bugs in JuliaFormatter.

odow:
Closing because this isn't ready yet. Blocked by upstream.

PR #2995, opened 2022-05-27
[docs] add style guide on abstract types

### LINKED ISSUE #2988: Discussion/style guide updates on abstraction/correctness/composability
https://yuri.is/not-julia/ and the corresponding [HN discussion](https://news.ycombinator.com/item?id=31396861) is worth a read. Overall I think JuMP is already doing quite well with respect to the cultural issues raised (none of the bugs involved JuMP); however, JuMP is part of the Julia community so we should reflect on what we can do to improve. 

IMO the post claims the following issues in the Julia ecosystem:
- Code tends to not document assumptions on abstract types (assuming AbstractArrays are 1-based)
- Undocumented assumptions are not checked (e.g., [`copyto!` and aliasing](https://github.com/JuliaLang/julia/issues/39460))
- Composition of packages is untested (e.g., OffsetArrays and Distributions)

JuMP's style guide doesn't really discuss these issues, except for user-facing method errors (which are a result of unchecked assumptions on abstract types).

odow:
## What is this PR?

This PR adds some recommendations to the JuMP style guide that we think you should follow when working with abstract types and method arguments in Julia.

## Why

The  [not-julia](https://yuri.is/not-julia/) article made some very good points.

There was a long discussion on Discourse:
 * https://discourse.julialang.org/t/discussion-on-why-i-no-longer-recommend-julia-by-yuri-vishnevsky/81151
and also one on HackerNews. Here's the relevant JuMP issue: https://github.com/jump-dev/JuMP.jl/issues/2988

## Feedback

We're in no rush to merge this, so please chime in with feedback. This will likely take a few rounds of edits to get right.

Closes #2988

### DIFF
--- a/docs/src/developers/style.md
+++ b/docs/src/developers/style.md
@@ -51,18 +51,20 @@ format("test")
 The following sections outline extra style guide points that are not fixed
 automatically by JuliaFormatter.
 
-### Dealing with abstract types and composition
+### Abstract types and composition
 
-Specifying argument types for methods is mostly optional in Julia. The benefit
-of this is that it enables functions and types from one package to be used with
-functions and types from another package via multiple dispatch.
+Specifying types for method arguments and struct fields is mostly optional in
+Julia. The benefit of this is that it enables functions and types from one
+package to be used with functions and types from another package via multiple
+dispatch.
 
 However, abstractly typed methods have two main drawbacks:
 
  1. It's possible to find out that you are working with unexpected types deep
-    in the call chain, potentially leading to hard-to-diagnose `MethodError`s.
- 2. Untyped function arguments can lead to correctness problems if the input
-    type does not satisfy assumptions made by the author of the function.
+    in the call chain, potentially leading to hard-to-diagnose [`MethodError`s](https://docs.julialang.org/en/v1/manual/methods/#Defining-Methods).
+ 2. Untyped function arguments can lead to correctness problems if the user's
+    choice of input type does not satisfy the assumptions made by the author of
+    the function.
 
 As a motivating example, consider the following function:
 ```jldoctest my_sum
@@ -76,13 +78,14 @@ julia> function my_sum(x)
 my_sum (generic function with 1 method)
 ```
 This function contains a number of implicit assumptions about the type of `x`:
- * `x` supports 1-based indexing and implements `length`
+ * `x` supports 1-based `getindex` and implements `length`
  * The element type of `x` supports addition with `0.0`, and then with the
    result of `x + 0.0`.
 
-(As a motivating example for the second point, [`VariableRef`](@ref) plus
-`Float64` produces an [`AffExpr`](@ref). Do not assume that `+(::T, ::T)`
-produces the type `T`.)
+!!! info
+    As a motivating example for the second point, [`VariableRef`](@ref) plus
+    `Float64` produces an [`AffExpr`](@ref). Do not assume that `+(::T, ::T)`
+    produces the type `T`.
 
 `my_sum` works as expected if the user passes in `Vector{Float64}`:
 ```jldoctest my_sum
@@ -146,8 +149,8 @@ public_function(x) = _internal_function(x)
 #### Dealing with correctness
 
 Dealing with correctness is harder, because Julia has no way of formally
-specifying interfaces that abstract types must implement. Innstead, here are two
-main options that you can use when writing and interacting with generic code:
+specifying interfaces that abstract types must implement. Instead, here are two
+options that you can use when writing and interacting with generic code:
 
 **Option 1: Use concrete types and let users extend new methods.**
 
@@ -174,7 +177,7 @@ ERROR: MethodError: no method matching my_sum_option_1(::String)
 ```
 and it allows other types to be supported in future by defining new methods:
 ```jldoctest my_sum
-julia> function my_sum_option_1(x::Array{T,N}) where {T,N}
+julia> function my_sum_option_1(x::Array{T,N}) where {T<:Number,N}
            y = zero(T)
            for i in eachindex(x)
                y += x[i]
@@ -185,6 +188,12 @@ my_sum_option_1 (generic function with 2 methods)
 ```
 Importantly, these methods do not have to be defined in the original package.
 
+!!! info
+    Some usage of abstract types is okay. For example, in `my_sum_option_1`, we
+    allowed the element type, `T`, to be a subtype of `Number`. This is fairly
+    safe, but it still has an implicit assumption that `T` supports `zero(T)`
+    and `+(::T, ::T)`.
+
 **Option 2: Program defensively, and validate all assumptions.**
 
 An alternative is to program defensively, and to rigorously document and
@@ -192,9 +201,10 @@ validate all assumptions that the code makes. In particular:
 
  1. All assumptions on abstract types that aren't guaranteed by the definition
     of the abstract type (for example, optional methods without a fallback)
-    should be documented
+    should be documented.
  2. If practical, the assumptions should be checked in code, and informative
-    error messages should be provided to the user if the assumptions are not met
+    error messages should be provided to the user if the assumptions are not
+    met.
  3. Tests should cover for a range of corner cases and argument types.
 
 For example:
--- a/docs/src/developers/style.md
+++ b/docs/src/developers/style.md
@@ -216,7 +216,12 @@ validate all assumptions that the code makes. In particular:
 
 For example:
 ```jldoctest my_sum
-function _validate_my_sum_defensive_assumptions(x::AbstractArray{T}) where {T}
+"""
+    test_my_sum_defensive_assumptions(x::AbstractArray{T}) where {T}
+
+Test the assumptions made by `my_sum_defensive`.
+"""
+function test_my_sum_defensive_assumptions(x::AbstractArray{T}) where {T}
     try
         # Some types may not define zero.
         @assert zero(T) isa T
@@ -247,7 +252,7 @@ This function makes the following assumptions:
  * That  `+(::T, ::T)` is defined
 """
 function my_sum_defensive(x::AbstractArray{T}) where {T}
-    _validate_my_sum_defensive_assumptions(x)
+    test_my_sum_defensive_assumptions(x)
     y = zero(T)
     for xi in x
         y += xi
@@ -276,6 +281,10 @@ julia> my_sum_defensive(['a', 'b', 'c'])
 ERROR: Unable to call my_sum_defensive(::Vector{Char}) because it failed an internal assumption
 [...]
 ```
+
+As an alternative, you may choose not to call `test_my_sum_defensive_assumptions`
+within `my_sum_defensive`, and instead ask users of `my_sum_defensive` to call
+it in their tests.
 
 ### Juxtaposed multiplication
 
--- a/docs/src/developers/style.md
+++ b/docs/src/developers/style.md
@@ -210,11 +210,29 @@ validate all assumptions that the code makes. In particular:
     should be documented.
  2. If practical, the assumptions should be checked in code, and informative
     error messages should be provided to the user if the assumptions are not
-    met.
+    met. In general, these checks may be expensive, so you should prefer to do
+    this once, at the highest level of the call-chain.
  3. Tests should cover for a range of corner cases and argument types.
 
 For example:
 ```jldoctest my_sum
+function _validate_my_sum_defensive_assumptions(x::AbstractArray{T}) where {T}
+    try
+        # Some types may not define zero.
+        @assert zero(T) isa T
+        # Check iteration supported
+        @assert iterate(x) isa Union{Nothing,Tuple{T,Int}}
+        # Check that + is defined
+        @assert +(zero(T), zero(T)) isa Any
+    catch err
+        error(
+            "Unable to call my_sum_defensive(::$(typeof(x))) because " *
+            "it failed an internal assumption",
+        )
+    end
+    return
+end
+
 """
     my_sum_defensive(x::AbstractArray{T}) where {T}
 
@@ -229,19 +247,7 @@ This function makes the following assumptions:
  * That  `+(::T, ::T)` is defined
 """
 function my_sum_defensive(x::AbstractArray{T}) where {T}
-    try
-        # Some types may not define zero.
-        @assert zero(T) isa T
-        # Check iteration supported
-        @assert iterate(x) isa Union{Nothing,Tuple{T,Int}}
-        # Check that + is defined
-        @assert +(zero(T), zero(T)) isa Any
-    catch err
-        error(
-            "Unable to call my_sum_defensive(::$(typeof(x))) because " *
-            "it failed an internal assumption",
-        )
-    end
+    _validate_my_sum_defensive_assumptions(x)
     y = zero(T)
     for xi in x
         y += xi
--- a/docs/src/developers/style.md
+++ b/docs/src/developers/style.md
@@ -51,6 +51,162 @@ format("test")
 The following sections outline extra style guide points that are not fixed
 automatically by JuliaFormatter.
 
+### [The `MethodError` principle](@id method_error_principle)
+
+Specifying argument types for methods is mostly optional in Julia, which means
+that it's possible to find out that you are working with unexpected types deep in
+the call chain. Avoid this situation or handle it with a helpful error message.
+
+*A user should see a `MethodError` only for methods that they called directly.*
+
+Bad:
+```julia
+_internal_function(x::Integer) = x + 1
+# The user sees a MethodError for _internal_function when calling
+# public_function("a string"). This is not very helpful.
+public_function(x) = _internal_function(x)
+```
+
+Good:
+```julia
+_internal_function(x::Integer) = x + 1
+# The user sees a MethodError for public_function when calling
+# public_function("a string"). This is easy to understand.
+public_function(x::Integer) = _internal_function(x)
+```
+
+If it is hard to provide an error message at the top of the call chain,
+then the following pattern is also ok:
+```julia
+_internal_function(x::Integer) = x + 1
+function _internal_function(x)
+    error(
+        "Internal error. This probably means that you called " *
+        "public_function() with the wrong type.",
+    )
+end
+public_function(x) = _internal_function(x)
+```
+
+### Dealing with abstract types and composition
+
+In addition to `MethodError`s, untyped function arguments can lead to
+correctness problems if the input type does not satisfy assumptions made by the
+author of the function.
+
+As a motivating example, consider the following function:
+```jldoctest my_sum
+julia> function my_sum(x)
+           y = 0.0
+           for i in 1:length(x)
+               y += x[i]
+           end
+           return y
+       end
+my_sum (generic function with 1 method)
+```
+This function contains a number of implicit assumptions about the type of `x`:
+ * `x` supports 1-based indexing and implements `length`
+ * The element type of `x` supports addition with `0.0`, and then with the
+   result of `x + 0.0`.
+
+(As a motivating example for the second point, [`VariableRef`](@ref) plus
+`Float64` produces an [`AffExpr`](@ref). Do not assume that `+(::T, ::T)`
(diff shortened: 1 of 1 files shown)


docs/src/developers/style.md
```diff
@@ -51,6 +51,162 @@ format("test")
 The following sections outline extra style guide points that are not fixed
 automatically by JuliaFormatter.
 
+### [The `MethodError` principle](@id method_error_principle)
```
mlubin, line 54: I think this can be rewritten more clearly as a corollary of the guidelines for dealing the abstraction and composition, rather than a standalone principle.

docs/src/developers/style.md
```diff
@@ -51,6 +51,162 @@ format("test")
...
+```jldoctest my_sum
+julia> my_sum("abc")
+ERROR: MethodError: no method matching +(::Float64, ::Char)
+[...]
+```
+
+The JuMP style guide has two main suggestions for writing and interacting with
+generic code.
+
+**Option 1: Program restrictively and let users extend new methods.**
```
mlubin, line 138: Maybe "Use concrete types" instead of "Program restrictively"?

docs/src/developers/style.md
```diff
@@ -51,6 +51,162 @@ format("test")
...
+
+**Option 2: Program defensively, and validate all assumptions.**
+
+An alternative is to program defensively, and to rigorously list and validate
+all assumptions that the code makes. For example:
+```jldoctest my_sum
+julia> function my_sum_defensive(x::AbstractArray{T}) where {T}
+           try
+               # Some types may not define zero.
+               @assert zero(T) isa T
```
mlubin, line 183: I'd reorder the priorities: 1) All assumptions on abstract types (that aren't guaranteed by the definition of the abstract type) should be documented. 2) If practical, also check the assumptions in code. 3) It's good practice to add test coverage for a range of corner cases.

docs/src/developers/style.md
```diff
@@ -51,6 +51,162 @@ format("test")
...
+generic code.
+
+**Option 1: Program restrictively and let users extend new methods.**
+
+In this scenario, we explictly restrict input arguments to types that are tested
+and have been validated for correctness. For example:
+```jldoctest my_sum
+julia> function my_sum_restrictive(x::Vector{Float64})
+           y = 0.0
+           for i in 1:length(x)
```
mlubin, line 145: What do say about preferring generic operations like `zero` or `eachindex` when it's easy to do so?

docs/src/developers/style.md
```diff
@@ -51,6 +51,210 @@ format("test")
...
+Importantly, these methods do not have to be defined in the original package.
+
+**Option 2: Program defensively, and validate all assumptions.**
+
+An alternative is to program defensively, and to rigorously document and
+validate all assumptions that the code makes. In particular:
+
+ 1. All assumptions on abstract types that aren't guaranteed by the definition
+    of the abstract type (for example, optional methods without a fallback)
+    should be documented
```
mlubin, line 195: Nit: missing period for bullets 1 and 2.

docs/src/developers/style.md
```diff
@@ -51,6 +51,220 @@ format("test")
 The following sections outline extra style guide points that are not fixed
 automatically by JuliaFormatter.
 
+### Abstract types and composition
+
+Specifying types for method arguments and struct fields is mostly optional in
+Julia. The benefit of this is that it enables functions and types from one
```
mlubin, line 57: The second sentence "The benefit of this ..." applies clearly for method arguments but is a bit confusing for struct fields.

mlubin (approved):
This looks good as a starting point for discussions.

docs/src/developers/style.md
```diff
@@ -51,6 +51,220 @@ format("test")
...
+```
+This function contains a number of implicit assumptions about the type of `x`:
+ * `x` supports 1-based `getindex` and implements `length`
+ * The element type of `x` supports addition with `0.0`, and then with the
+   result of `x + 0.0`.
+
+!!! info
+    As a motivating example for the second point, [`VariableRef`](@ref) plus
+    `Float64` produces an [`AffExpr`](@ref). Do not assume that `+(::T, ::T)`
+    produces the type `T`.
```
matbesancon, line 88:
```suggestion
    `Float64` produces an [`AffExpr`](@ref). Do not assume that `+(::A, ::B)`
    produces the type `A` or `B`.
```
If you want to prove the point for a same type `T`, another example would be more appropriate

nickrobinson251:
In case it's of interest, the recently released [SciML style](https://github.com/SciML/SciMLStyle) also has a section on recommendations e.g. https://github.com/SciML/SciMLStyle#errors-should-be-caught-as-high-as-possible-and-error-messages-should-be-contextualized-for-newcommers

I'm not sure what the name is for these kind of guidelines (principles? / philosophies? / dogmas?!), different from the syntactic style suggestions (things JuliaFormatter could in principle automate, "put space around binary operators" etc), but i think it's interesting that these things are being made more explicit. I suppose, like with syntactic style, it may be interesting in time to see where different large projects agree/disagree on various points.

odow:
> it may be interesting in time to see where different large projects agree/disagree on various points.

The tension is definitely the extent to which abstract interfaces are supported. 

I'm also not a major fan of things like https://github.com/SciML/SciMLStyle#closures-should-be-avoided-whenever-possible 

We spent a lot of effort in the development of MathOptInterface chasing performance, but readability and maintainability are equally important, if not more so.

docs/src/developers/style.md
```diff
@@ -51,6 +51,220 @@ format("test")
...
+ERROR: MethodError: no method matching +(::Float64, ::Char)
+[...]
+```
+This particular `MethodError` is hard to debug, particularly for new users,
+because it mentions `+`, `Float64`, and `Char`, none of which were called or
+passed by the user.
+
+#### Dealing with `MethodError`s
+
+All code must follow the `MethodError` principle:
```
matbesancon, line 113: All code seems a bit extreme. I can understand providing an opinion and the advantage it can provide in some cases, but overgeneralizing this can lead to a lot of code to produce errors, and may require overly specific methods to avoid ambiguities with the error fallback.

@oxinabox detailed it more here: https://www.oxinabox.net/2020/04/19/Julia-Antipatterns.html#notimplemented-exceptions

odow: The experience is that beginner Julia users find MethodErrors hard to deal with. They might not even have a good concept of what a type is, let alone specifics of why `Vector{Float64} <: Vector{Real}` is `false`.

Is "All _public_ code" better? Admittedly, we don't do this perfectly, so it's a standard to strive for, not a fixed rule.

mlubin: I'm not sure I agree with distinguishing between public and private code. Private code should still follow the style guide. Instead of "All code must [...]", what about "Code should [...]"? The style guide is a set of recommendations, so "must" is out of character.

docs/src/developers/style.md
```diff
@@ -51,6 +51,220 @@ format("test")
 The following sections outline extra style guide points that are not fixed
 automatically by JuliaFormatter.
 
+### Abstract types and composition
+
+Specifying types for method arguments and struct fields is mostly optional in
```
blegat, line 56: I wouldn't throw struct field types into the same baskets, it is a common misunderstanding for new users that typing arguments is optional but typing struct fields is important for performance

docs/src/developers/style.md
```diff
@@ -51,6 +51,220 @@ format("test")
...
+```
+This function contains a number of implicit assumptions about the type of `x`:
+ * `x` supports 1-based `getindex` and implements `length`
+ * The element type of `x` supports addition with `0.0`, and then with the
+   result of `x + 0.0`.
+
+!!! info
+    As a motivating example for the second point, [`VariableRef`](@ref) plus
+    `Float64` produces an [`AffExpr`](@ref). Do not assume that `+(::A, ::B)`
+    produces the type `A` or `B`.
```
blegat, line 88:
```suggestion
    produces an instance of the type `A` or `B`.
```

docs/src/developers/style.md
```diff
@@ -51,6 +51,220 @@ format("test")
...
+```
+
+If it is hard to provide an error message at the top of the call chain,
+then the following pattern is also ok:
+```julia
+_internal_function(x::Integer) = x + 1
+function _internal_function(x)
+    error(
+        "Internal error. This probably means that you called " *
+        "public_function() with the wrong type.",
```
blegat, line 143:
```suggestion
        "`public_function()` with the wrong type.",
```

docs/src/developers/style.md
```diff
@@ -51,6 +51,220 @@ format("test")
...
+## Assumptions
+
+This function makes the following assumptions:
+
+ * That `zero(T)` is defined
+ * That `x` supports the iteration interface
+ * That  `+(::T, ::T)` is defined
+"""
+function my_sum_defensive(x::AbstractArray{T}) where {T}
+    try
```
blegat, line 226: We don't want to start doing this in JuMP and MOI right ? Documenting it is enough I think

docs/src/developers/style.md
oxinabox: I didn't have any comment on this when tagged because, I think JuMP has its own philosphy that is strong and well-rounded and targetting a different audience to my blog post.

But actually now I do have an opinion because JuMP's current doing of this is annoying to me.
because I was like `shadow_price(my_constraint)` just  now
and I got 
```julia
julia> shadow_price.(cons)
ERROR: The shadow price is not defined or not implemented for this type of constraint.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] shadow_price(con_ref::ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.VectorAffineFunction{Float64}, MathOptInterface.NormOneCone}, VectorShape})
 ```
 
 and my initial instinct was: _"Ok well lets see what types it is defined for"_.
and so I used `methods(shadow_price)`.
```julia
julia> methods(shadow_price)
# 4 methods for generic function "shadow_price":
[1] shadow_price(con_ref::ConstraintRef{<:AbstractModel, MathOptInterface.ConstraintIndex{F, S}}) where {S<:MathOptInterface.LessThan, F} in JuMP at /home/oxinabox/.julia/packages/JuMP/Psd1J/src/constraints.jl:1056
[2] shadow_price(con_ref::ConstraintRef{<:AbstractModel, MathOptInterface.ConstraintIndex{F, S}}) where {S<:MathOptInterface.GreaterThan, F} in JuMP at /home/oxinabox/.julia/packages/JuMP/Psd1J/src/constraints.jl:1069
[3] shadow_price(con_ref::ConstraintRef{<:AbstractModel, MathOptInterface.ConstraintIndex{F, S}}) where {S<:MathOptInterface.EqualTo, F} in JuMP at /home/oxinabox/.julia/packages/JuMP/Psd1J/src/constraints.jl:1082
[4] shadow_price(con_ref::ConstraintRef{<:AbstractModel, <:MathOptInterface.ConstraintIndex}) in JuMP at /home/oxinabox/.julia/packages/JuMP/Psd1J/src/constraints.jl:1017
```
Which of-course lists there as being a method for this.
The one that manually threw the error.

So while i do think JuMP is probably right in having a strong philosphy around these things, 
I will just say it is not without cost in terms of usability to experienced julia programmers.
It is probably worth it to be more friendly to less experienced programmers.
That is a call that should be conciously made however.

odow: > I will just say it is not without cost in terms of usability to experienced julia programmers.
> It is probably worth it to be more friendly to less experienced programmers.

Yes, I think we went for the latter.

docs/src/developers/style.md
odow: No. At the JuMP level, we mostly go with the concrete type suggestion.

But here's a case where we (I) didn't check assumptions: https://github.com/jump-dev/JuMP.jl/issues/2999

ChrisRackauckas:
> I'm also not a major fan of things like https://github.com/SciML/SciMLStyle#closures-should-be-avoided-whenever-possible

> We spent a lot of effort in the development of MathOptInterface chasing performance, but readability and maintainability are equally important, if not more so.

The suggestion to not do that is a suggestion against maintainability and readable error messages. Anonymous functions and closures give unnamed functions in stack traces and give lots of hidden compile time issues, along with some correctness issues that can come from unintentional captured variables. So that's a direct contradiction of the proposed values.

odow:
> Anonymous functions and closures give unnamed functions in stack traces

So the issue is the `#5` instead of `Fix2` here?
```Julia
julia> function foo(vector_of_vectors, i)
           return map(Base.Fix2(getindex, i), vector_of_vectors)
       end
foo (generic function with 1 method)

julia> function bar(vector_of_vectors, i)
           return map(v -> v[i], vector_of_vectors)
       end
bar (generic function with 1 method)

julia> vector_of_vectors = [rand(3) for _ in 1:4]
4-element Vector{Vector{Float64}}:
 [0.2944113044040204, 0.1502238100477964, 0.7719060173197017]
 [0.5503852948486079, 0.12060690924862927, 0.9112641636545447]
 [0.8292675030520171, 0.8359221806182853, 0.04041374985594448]
 [0.08277416126462644, 0.14552667363773697, 0.7186541169561522]

julia> foo(vector_of_vectors, 4)
ERROR: BoundsError: attempt to access 3-element Vector{Float64} at index [4]
Stacktrace:
 [1] getindex
   @ ./array.jl:805 [inlined]
 [2] Fix2
   @ ./operators.jl:1002 [inlined]
 [3] iterate
   @ ./generator.jl:47 [inlined]
 [4] _collect
   @ ./array.jl:695 [inlined]
 [5] collect_similar
   @ ./array.jl:606 [inlined]
 [6] map
   @ ./abstractarray.jl:2294 [inlined]
 [7] foo(vector_of_vectors::Vector{Vector{Float64}}, i::Int64)
   @ Main ./REPL[6]:2
 [8] top-level scope
   @ REPL[9]:1

julia> bar(vector_of_vectors, 4)
ERROR: BoundsError: attempt to access 3-element Vector{Float64} at index [4]
Stacktrace:
 [1] getindex
   @ ./array.jl:805 [inlined]
 [2] #5
   @ ./REPL[7]:2 [inlined]
 [3] iterate
   @ ./generator.jl:47 [inlined]
 [4] _collect
   @ ./array.jl:695 [inlined]
 [5] collect_similar
   @ ./array.jl:606 [inlined]
 [6] map
   @ ./abstractarray.jl:2294 [inlined]
 [7] bar(vector_of_vectors::Vector{Vector{Float64}}, i::Int64)
   @ Main ./REPL[7]:2
 [8] top-level scope
   @ REPL[10]:1
```
I guess that's a trade-off, but I'd rather code that was easier to read than stack traces that were easier to read.

> along with some correctness issues

The compile issues are known and also part of the trade-off, but what are the correctness issues?

ChrisRackauckas:
> So the issue is the #5 instead of Fix2 here?
> I guess that's a trade-off, but I'd rather code that was easier to read than stack traces that were easier to read.

Yes, you get names for everything and they can be documented. Though if you do use a anonymous function, one thing you could do is used a named anonymous function. This is part of the ChainRules style rules, as seen here:

https://github.com/JuliaDiff/ChainRules.jl/blob/v1.35.3/src/rulesets/LinearAlgebra/uniformscaling.jl

where if this is seen in a stacktrace, it will use the name `plus_back` instead of an obscure `#5`. That's one way to make things a bit nicer (and is very crucial for debuggable AD stacktraces).

>The compile issues are known and also part of the trade-off, but what are the correctness issues?

@chriselrod had a bunch of examples of some difficult to debug cases which prompted him to write that rule into the style guide, so I'll let him share those.

chriselrod:
> The compile issues are known and also part of the trade-off, but what are the correctness issues?

Here is a minimal example:
```julia
julia> let
         s = Threads.Atomic{Int}(0);
         Threads.@threads for i = 1:10000
           x = i
           sleep(1e-3);
           Threads.atomic_add!(s,x)
         end
         s[]
       end
50005000

julia> sum(1:10_000)
50005000

julia> let
         s = Threads.Atomic{Int}(0);
         Threads.@threads for i = 1:10000
           x = i
           sleep(1e-3);
           Threads.atomic_add!(s,x)
         end
         x = 0
         s[]
       end
50046631
```

chriselrod:
> but readability and maintainability are equally important, if not more so.

These are precisely why I don't want closures in a code base.
I want to keep spooky action at a distance to a minimum.

odow:
Okay, so to check my understanding, the issue is that `x = 0` introduces `x` into the scope of the let-block, so the for-loop captures `x` and modifies it in the threading which is not thread-safe. But if you went `local x = i`, presumably that would work.

It also seems like something that captures `x` is a closure, but which doesn't lead to the inner `x` getting confused for it will also work:
```Julia
julia> function foo(f)
           s = Threads.Atomic{Int}(0)
           Threads.@threads for i = 1:10000
               x = f(i)
               sleep(1e-3)
               Threads.atomic_add!(s, x)
           end
           return s[]
       end
foo (generic function with 1 method)

julia> let
           x = 0
           foo(i -> i + x)
       end
50005000
```
So the problem isn't really closures, it's dealing with scoping issues and thread-safety in threading code? That seems like a tricky footgun to get right, but not a reason to avoid closures in all cases.

Anyway, we're derailing the conversation from the actual PR, which was about `MethodErrors`. There's no suggestion in the JuMP style guide at present to use or avoid closures. It was just an off-hand comment about some differences.

chriselrod:
> if you went `local x = i`, presumably that would work.

Yes. That's what I do in the proprietary code. It's of course easier to add `local` in front of every line of a big closure than to rewrite the code not to use closures.
Perhaps a softer guideline is "closures are okay so long as every assignment is prefaced by `local`".
I would still strongly discourage "global", preferring `Ref` if that's the actually intended behavior (which it of course never is).
That would give the code the property of being able to move it around without changing its meaning, which is IMO a desirable property / makes it much easier to understand by mitigating the significance of context.

> So the problem isn't really closures

I would say the problem is that closures capture and mutate bindings, creating global state in code where if you don't look at the entire program, it looks like things are local.

Mutable global state is widely considered bad.
We generally discourage people from defining
```julia
const MYOPTIONS = Ref{MyOptionType}()
```
as means of passing around information / controlling the behavior of their code, suggesting they instead pass their options around as a (keyword) argument to all their functions that need it.
This would especially be the recommendation if it is a variable being actively mutated as part of their computations.

The problem with closures are that they randomly make people's code follow really bad/strongly discouraged practices without anyone even realizing it, and hence introduce the associated bugs that people weren't even looking out for.

odow:
One place JuMP uses mutating closures is in callbacks https://jump.dev/JuMP.jl/stable/manual/callbacks/#Lazy-constraints. Most of the JuMP API is mutating-by-default, so we already don't work with threading.

The bigger problem I see is that to the general user, it isn't actually obvious that this _is_ a closure, or that it leads to non-local behavior. 
```Julia
julia> let
         s = Threads.Atomic{Int}(0);
         Threads.@threads for i = 1:10000
           x = i
           sleep(1e-3);
           Threads.atomic_add!(s,x)
         end
         x = 0
         s[]
       end
```
You almost need a Julia flag (or a linter) `julia --error-on-modify-captured-variables` that catches this at the compiler level. Leaving it to the user to notice ain't going to work.

chriselrod:
> You almost need a Julia flag (or a linter) `julia --error-on-modify-captured-variables` that catches this at the compiler level. 

If the LSP would warn about it reliably, that'd solve most of my griping.
I'm not sure how difficult that'd be. On the one hand, you don't even need type information to figure that out, but on the other you need to be able to expand macros. `Threads.@threads` as you noted, as well as DSLs introduce a lot of closures.

If it were always obvious what was happening, there wouldn't be much of a problem.

Maybe I should file an issue/feature request at `LanguageServer.jl`?

odow:
@matbesancon, I tweaked the wording of the assumption checks.

odow:
What about instead of validating assumptions, we say you should provide a function that the user can call in their tests which checks assumptions?

matbesancon:
that sounds like a good alternative to suggest

odow:
Okay, I think we're at a point where we can merge this. We can always continue to tweak it going forward.

matbesancon (approved):

PR #3000, opened 2022-06-08
Fix SparseAxisArray and MutableArithmetics

### LINKED ISSUE #2999: MutableArithmetics 1.0.3 broke tests
https://github.com/jump-dev/JuMP.jl/runs/6800887087?check_suite_focus=true

Problem was https://github.com/jump-dev/MutableArithmetics.jl/pull/159 because it calls `size` on `SparseAxisArray` which doesn't support it. (Ironically, the tests broke in https://github.com/jump-dev/JuMP.jl/pull/2995.)

We should fix this in JuMP, but we should also add a CI check in MutableArithmetics to stop us from breaking JuMP in future, particularly now that we're both at 1.0 and the API is stable.

odow:
Quick fix to see if this fixes tests.

Closes #2999

### DIFF
--- a/src/Containers/SparseAxisArray.jl
+++ b/src/Containers/SparseAxisArray.jl
@@ -58,12 +58,17 @@ function Base.size(::SparseAxisArray)
     )
 end
 
+# Needed to work-around JuMP issue #2999
 function MutableArithmetics.broadcast_mutability(
     x::SparseAxisArray,
     op,
     args::Vararg{Any,N},
 ) where {N}
-    return MutableArithmetics.broadcast_mutability(typeof(x), op, typeof.(args)...)
+    return MutableArithmetics.broadcast_mutability(
+        typeof(x),
+        op,
+        typeof.(args)...,
+    )
 end
 
 # A `length` argument can be given because `IteratorSize` is `HasLength`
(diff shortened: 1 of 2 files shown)


src/Containers/SparseAxisArray.jl
```diff
@@ -58,6 +58,14 @@ function Base.size(::SparseAxisArray)
     )
 end
 
+function MutableArithmetics.broadcast_mutability(
```
odow, line 61:
```suggestion
# Needed to work-around JuMP issue #2999
function MutableArithmetics.broadcast_mutability(
```

src/Containers/SparseAxisArray.jl
```diff
@@ -58,6 +58,14 @@ function Base.size(::SparseAxisArray)
...
+    x::SparseAxisArray,
+    op,
+    args::Vararg{Any,N},
+) where {N}
+    return MutableArithmetics.broadcast_mutability(typeof(x), op, typeof.(args)...)
```
odow, line 66:
```suggestion
    return MutableArithmetics.broadcast_mutability(
        typeof(x),
        op,
        typeof.(args)...,
    )
```

odow:
I'll aim to fix this in https://github.com/jump-dev/MutableArithmetics.jl/pull/160 with a quick release of MA@1.0.4

odow:
Closing in favor of https://github.com/jump-dev/MutableArithmetics.jl/pull/160

PR #2936, opened 2022-03-29
Add num_constraints for counting all constraints in a model

odow:
As requested by @ccoffrin on Gitter.

PR #3002, opened 2022-06-10
Add all_constraints for all constraint types

### LINKED ISSUE #3001: all_constraints(model)
I wish to list all constraints so i can then check all `shadow_prices` so I can find which as nonzero to determine which constraints are binding.
I have also wanted to list all constraints for check feasibility of a potential primal solution in the past, but that is now a inbuilt feature of JuMP so i don't need to.

define 
```julia
all_constraints(model) = mapreduce((f,s)->all_constraints(model, f,s), vcat, list_of_constraint_types(model))
```
in JuMP and export it

Not doing this, and making the user do this

odow:
Closes #3001 

cc @oxinabox

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -1280,3 +1280,51 @@ function num_constraints(model::Model; count_variable_in_set_constraints::Bool)
     end
     return ret
 end
+
+"""
+    all_constraints(model::Model; include_variable_in_set_constraints::Bool)
+
+Return a list of all constraints in `model`.
+
+If `include_variable_in_set_constraints == true`, then `VariableRef` constraints
+such as `VariableRef`-in-`Integer` are included. To return only the structural
+constraints (e.g., the rows in the constraint matrix of a linear program), pass
+`include_variable_in_set_constraints = false`.
+
+## Examples
+
+```jldoctest; setup=:(using JuMP)
+julia> model = Model();
+
+julia> @variable(model, x >= 0, Int);
+
+julia> @constraint(model, 2x <= 1);
+
+julia> @NLconstraint(model, x^2 <= 1);
+
+julia> all_constraints(model; include_variable_in_set_constraints = true)
+4-element Vector{ConstraintRef}:
+ 2 x ≤ 1.0
+ x ≥ 0.0
+ x integer
+ x ^ 2.0 - 1.0 ≤ 0
+
+julia> all_constraints(model; include_variable_in_set_constraints = false)
+2-element Vector{ConstraintRef}:
+ 2 x ≤ 1.0
+ x ^ 2.0 - 1.0 ≤ 0
+```
+"""
+function all_constraints(
+    model::Model;
+    include_variable_in_set_constraints::Bool,
+)
+    ret = ConstraintRef[]
+    for (F, S) in list_of_constraint_types(model)
+        if F != VariableRef || include_variable_in_set_constraints
+            append!(ret, all_constraints(model, F, S))
+        end
+    end
+    append!(ret, all_nonlinear_constraints(model))
+    return ret
+end
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -1282,7 +1282,10 @@ function num_constraints(model::Model; count_variable_in_set_constraints::Bool)
 end
 
 """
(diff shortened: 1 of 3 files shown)


src/constraints.jl
```diff
@@ -1280,3 +1280,51 @@ function num_constraints(model::Model; count_variable_in_set_constraints::Bool)
...
+
+"""
+    all_constraints(model::Model; include_variable_in_set_constraints::Bool)
+
+Return a list of all constraints in `model`.
+
+If `include_variable_in_set_constraints == true`, then `VariableRef` constraints
+such as `VariableRef`-in-`Integer` are included. To return only the structural
+constraints (e.g., the rows in the constraint matrix of a linear program), pass
+`include_variable_in_set_constraints = false`.
```
blegat, line 1292: I would add a warning saying that this is not the best approach if performance is a concern

odow: I don't think we need to say this. It's unlikely that getting a list of all the constraints will be a performance bottleneck, and we shouldn't be scaring people into more complicated code just for the sake of performance.

blegat: Why unlikely ? We've had issues with MOI packages not doing this function barrier and getting performance issue

odow: Doing it on a critical path within a package is different to a user of JuMP.

blegat: Users of JuMP might be writing packages as well

matbesancon: yes there has been a whole phase of encouraging people to just write libraries on top of JuMP

src/constraints.jl
```diff
@@ -1280,3 +1280,51 @@ function num_constraints(model::Model; count_variable_in_set_constraints::Bool)
...
+
+julia> all_constraints(model; include_variable_in_set_constraints = false)
+2-element Vector{ConstraintRef}:
+ 2 x ≤ 1.0
+ x ^ 2.0 - 1.0 ≤ 0
+```
+"""
+function all_constraints(
+    model::Model;
+    include_variable_in_set_constraints::Bool,
```
matbesancon, line 1320:
```suggestion
    include_variable_in_set_constraints::Bool=true,
```
why not including a default?

odow: https://github.com/jump-dev/JuMP.jl/pull/2936#discussion_r837959751

src/constraints.jl
odow: How's this now?

src/constraints.jl
matbesancon: yes ok I understand being consistent with the other function, I still think that requiring the keyword is not ideal, defaulting to true makes sense here since we ask for `all_`, and the keyword is fairly long

src/constraints.jl
```diff
@@ -1280,3 +1280,57 @@ function num_constraints(model::Model; count_variable_in_set_constraints::Bool)
...
+
+Return a list of all constraints in `model`.
+
+If `include_variable_in_set_constraints == true`, then `VariableRef` constraints
+such as `VariableRef`-in-`Integer` are included. To return only the structural
+constraints (e.g., the rows in the constraint matrix of a linear program), pass
+`include_variable_in_set_constraints = false`.
+
+Note that this function is type-unstable because it returns an abstractly typed
+vector.
```
blegat, line 1298: We should at least give a pointer to somewhere where we show how to do this efficiently

matbesancon: Pointing to here:
https://github.com/jump-dev/JuMP.jl/pull/3002/files#diff-374232ee51673809a2a65674f659b6dda32ecec8854a91876b2ff2566e1f07d2R957
maybe?

odow: I just think this docstring is the wrong place for an extended note on performance and how to do this differently. If someone is writing on top of JuMP, then can profile their code, or already know about function barriers.

src/constraints.jl
odow: This function should be used pretty infrequently, so I don't think the long keyword is a problem. Forcing the user to pick also helps clarify whether they understand that `x in Integer` will be returned as a constraint.

PR #3004, opened 2022-06-11
Use MathJax to render formula in doc

metab0t:
MOI uses MathJax to render formula correctly.
https://github.com/jump-dev/MathOptInterface.jl/blob/fd3d4aa5c8fe0fb2ad412ffcf1825e2ec914a99b/docs/make.jl#L99


odow:
Thanks! Looks better now.

PR #3009, opened 2022-06-24
[docs] add section on extension compatibility

odow:
As discussed on today's developer call, extensions using the private API must use appropriate compat bounds.

x-ref: https://github.com/jump-dev/JuMP.jl/pull/2955


mlubin:
Longer term it would be nice to have automatic issues/PRs to update extensions' compatibility with JuMP. Bumping the compat for every patch release is a lot of manual work.

odow:
I guess that's what https://github.com/JuliaRegistries/CompatHelper.jl is?

odow:
Merging for now. Let's see how we go, and if it is a problem we can add the suggestion for CompatHelper.jl, or consider other options.

PR #3016, opened 2022-07-11
Add Clarabel to solver list

blegat:
@goulart-paul Any objection to add it ?


goulart-paul:
None at all.   We would be very pleased to have it listed there.   Thanks.

PR #3017, opened 2022-07-13
Add fallback for unsupported Name attribute

odow:
Related to https://github.com/jump-dev/Ipopt.jl/issues/317. We can fix Ipopt, but there might be other solvers.  The `Name` attribute is optional, in general.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -131,12 +131,11 @@ default if empty.
 name(model::AbstractModel) = "An Abstract JuMP Model"
 
 function name(model::Model)
-    try
+    if MOI.supports(backend(model), MOI.Name())
         ret = MOI.get(model, MOI.Name())
         return isempty(ret) ? "A JuMP Model" : ret
-    catch
-        return "A JuMP Model"
     end
+    return "A JuMP Model"
 end
 
 """
--- a/src/print.jl
+++ b/src/print.jl
@@ -131,8 +131,12 @@ default if empty.
 name(model::AbstractModel) = "An Abstract JuMP Model"
 
 function name(model::Model)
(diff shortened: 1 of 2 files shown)


src/print.jl
```diff
@@ -131,8 +131,11 @@ default if empty.
 name(model::AbstractModel) = "An Abstract JuMP Model"
 
 function name(model::Model)
-    ret = MOI.get(model, MOI.Name())
-    return isempty(ret) ? "A JuMP Model" : ret
+    if MOI.supports(backend(model), MOI.Name())
```
blegat, line 134: What about backends that support getting but not setting a name ?

odow: What solver is that? And if they support getting but not setting, then JuMP will just print the default.

blegat: File readers ? This is just not respecting the MOI interface so it means that a valid MOI model will not work with JuMP because it assumes things outside the interface

odow: I'm not sure I understand/see the problem. This fixes a clear bug: https://github.com/jump-dev/Ipopt.jl/issues/317

blegat: > I'm not sure I understand/see the problem. This fixes a clear bug: jump-dev/Ipopt.jl#317

Yes, but the fixes doesn't respect the MOI interface. `supports` is about `MOI.set`, not about `MOI.get`.

Correct fixes could be:
* have a fallback in MOI that returns `""` if the attribute is not implemented or
* do a `try`-`catch`

blegat: See https://github.com/jump-dev/MathOptInterface.jl/pull/1726

Issue #1185, opened 2018-02-05
Fast resolves for NLP


mlubin:
https://github.com/JuliaOpt/MathOptInterface.jl/pull/202. Needs more infrastructure outside JuMP first (i.e. MOIT, MOIU, solver wrappers). We'll have to think about how we want to test this on the JuMP side.

mlubin:
This is mostly covered by https://github.com/JuliaOpt/JuMP.jl/pull/1223 except for duals and deciding what to do about incremental solves.

mlubin:
Duals are working. The only remaining issue is re-implementing fast resolves which I'll drop from the 0.19 milestone.

mlubin:
Dropping this from the 1.0 milestone. I've yet to see anyone who has said that they care and have benefited from fast NLP resolves.

odow:
I took a look at this and started hacking some things. But it's actually a little tricky, because now that we abstract the AD from the `Nonlinear.Model` object, there's no guarantee that updating the parameter values in the model will be reflected in the NLP evaluator. 

And even if JuMP doesn't set a new NLPBlockData, solvers like Ipopt call `MOI.initialize` in their `optimize!` call anyway, https://github.com/jump-dev/Ipopt.jl/blob/a1502e6bc410ee4531e07cc792932a7b0453e73b/src/MOI_wrapper.jl#L1085, which is the cause of the initial overhead.

I'll also note that this has had no comments since 2019, and only one thumbs up, so 

> I've yet to see anyone who has said that they care and have benefited from fast NLP resolves.

is probably still true. The overhead of setting up the AD system obviously isn't that big of a deal.

An alternative approach moving forward (that would require a fair bit more work) is probably to allow solvers like Ipopt to accept a `Nonlinear.Model` object directly, and then they can just check if the nonlinear model object is `===` to the last one they received.

I vote we close this. If someone asks in future, we can revisit.

blegat:
Maybe now with https://github.com/odow/SymbolicAD.jl it makes more sense to care about it ?

odow:
Not sure. It'd still take some changes in Ipopt.jl to implement properly. I think we should wait to see someone complain that it is a bottleneck. Premature optimization and all that.

odow:
We discussed this on the monthly nonlinear call.

odow:
@ccoffrin wants to properly benchmark on an OPF problem.

odow:
[...]

PR #3018, opened 2022-07-14
DNMY: experimental testing for fast resolves of NLP

odow:
This is NOT safe to merge because doesn't update the expression
graphs and so will break AmplNLWriter etc.

Part of #1185 

This needs some changes to Ipopt to see any potential benefits.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -152,8 +152,6 @@ mutable struct Model <: AbstractModel
     # A flag to track whether we have modified the model after calling
     # optimize!.
     is_model_dirty::Bool
-    # A flag to track whether we need to rebuild the NLP model before optimize!
-    is_nlp_model_dirty::Bool
     # Enable extensions to attach arbitrary information to a JuMP model by
     # using an extension-specific symbol as a key.
     ext::Dict{Symbol,Any}
@@ -258,7 +256,6 @@ function direct_model(backend::MOI.ModelLike)
         Dict{Symbol,Any}(),
         0,
         false,
-        true,
         Dict{Symbol,Any}(),
         true,
     )
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -152,6 +152,8 @@ mutable struct Model <: AbstractModel
     # A flag to track whether we have modified the model after calling
     # optimize!.
     is_model_dirty::Bool
+    # A flag to track whether we need to rebuild the NLP model before optimize!
+    is_nlp_model_dirty::Bool
     # Enable extensions to attach arbitrary information to a JuMP model by
     # using an extension-specific symbol as a key.
     ext::Dict{Symbol,Any}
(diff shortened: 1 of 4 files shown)


odow:
So the problem is that some AD backends might not update their expressions (or AD calls) if a parameter value is updated after `initialize`:
```Julia
julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x)
x

julia> @NLparameter(model, p == 2)
p == 2.0

julia> @NLexpression(model, ex, p)
subexpression[1]: p

julia> @NLobjective(model, Min, x^ex)

julia> evaluator = NLPEvaluator(model)
Nonlinear.Evaluator with available features:
  * :Grad
  * :Jac
  * :JacVec
  * :Hess
  * :HessVec
  * :ExprGraph

julia> MOI.initialize(evaluator, [:ExprGraph])

julia> MOI.objective_expr(evaluator)
:(x[MathOptInterface.VariableIndex(1)] ^ 2.0)

julia> set_value(p, 3)
3

julia> MOI.objective_expr(evaluator)
:(x[MathOptInterface.VariableIndex(1)] ^ 2.0)
```
But adding a way to update a parameter in `AbstractNLPEvaluator` breaks the abstraction that you can create a `Nonlinear.Model`, and then convert it into an `AbstractNLPEvaluator`. Now you'll have to keep the model around and check for updated parameter values.

odow:
`initialize` is essentially like `final_touch`. It says "I've finished building the model, and I'm ready to set everything up."

But the problem is that we want people to be able to modify parameter values and not have to rebuild everything, but the `NLPEvaluator`s don't have the concept of a parameter.

odow:
So one way to move forward with this is to update MOI.Nonlinear so that `initialize` doesn't store expression graphs. Then we can special-case JuMP to avoid setting the evaluator if the backend is `ReverseSparseAD`, and the previous backed was also `ReverseSparseAD`.

I don't think there's a generic solution, short of completely changing how the nonlinear interface is passed from JuMP to the solver.

odow:
Part of the problem is that we have an `Evaluator` which requires a fixed model. (You can't add new variables after creating the `Evaluator` object.)

We could switch to some mechanism where we pass the `Nonlinear.Model` object instead, but I wonder if it's simpler to just have a kwarg in `optimize!` that opt-in skips the NLP update. Then people solving PF in a loop can explicitly decide that things still work if they skip updating the Evaluator every time.

odow:
This allows a loop like:
```Julia
    for i in 1:n
        for (i,load) in data["load"]
            set_value(pd_parameter[load["index"]], (1.0+(rand()-0.5)/10)*load["pd_base"])
            set_value(qd_parameter[load["index"]], (1.0+(rand()-0.5)/10)*load["qd_base"])
        end
        if i > 1
            x = all_variables(model)
            x0 = value.(x)
            set_start_value.(x, x0)
        end
        optimize!(model; _skip_nonlinear_update = i > 1)

        @assert(termination_status(model) == LOCALLY_SOLVED)
        @assert(primal_status(model) == FEASIBLE_POINT)
    end
```

The problem with the previous "is the nonlinear model dirty" approach is that we'd also need to set the evaluator if the backend changed, even if the model didn't.

PR #3024, opened 2022-07-23
Tweak wording in getting started tutorial

mlubin:
"Cartesian product" sounds more standard than "rectangular set"


blegat (approved):
I agree

PR #3031, opened 2022-08-02
[Containers] add slicing for SparseAxisArray

### LINKED ISSUE #3029: Support slicing in SparseAxisArray
@hellemo gave a talk at JuMP-dev on sparse variables:

 * https://github.com/hellemo/SparseVariables.jl
 * https://www.youtube.com/watch?v=YuDvfZo9W5A

We should look through for ideas, and support slicing.

odow:
Closes #3029

This creates a copy, so it doesn't address the performance issues in @hellemo's talk, but it does address part of the usability.

This PR doesn't support slicing on `setindex!`. We'd have to play with the broadcasting issues.

### DIFF
--- a/src/Containers/SparseAxisArray.jl
+++ b/src/Containers/SparseAxisArray.jl
@@ -99,6 +99,10 @@ function _colon_error(::Colon, args...)
     )
 end
 
+_has_colon() = false
+_has_colon(::Any, args...) = _has_colon(args...)
+_has_colon(::Colon, args...) = true
+
 function Base.setindex!(
     d::SparseAxisArray{T,N,K},
     value,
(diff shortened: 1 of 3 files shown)


src/Containers/SparseAxisArray.jl
```diff
@@ -122,14 +122,49 @@ function Base.getindex(
     return getindex(d, idx...)
 end
 
-function Base.getindex(d::SparseAxisArray{T,N}, idx...) where {T,N}
+function Base.getindex(d::SparseAxisArray{T,N,K}, idx...) where {T,N,K}
     if length(idx) < N
         throw(BoundsError(d, idx))
     end
-    _colon_error(idx...)
+    if _is_slice(first(keys(d.data)), idx)
```
blegat, line 129: `first(keys(d.data))` will fail if the dictionary is empty

odow: I've updated the code to use a generated function, which has the nice side-effect of ensuring type stability of the result if the filter is empty.

PR #3037, opened 2022-08-12
Fix bug with set_string_name and VariablesConstrainedOnCreation

### LINKED ISSUE #3035: Error creating Symmetric variables without string-names in 1.1.1.
I am trying to create a symmetric variable (or matrix of variables) with JuMP 1.1.1 using the option to not create string names.

The following two work for me

```
model = Model(Hypatia.Optimizer)
@variable(model, [1:2, 1:2], Symmetric)
```

```
model = Model(Hypatia.Optimizer)
@variable(model, [1:2, 1:2])
```

but when combine the two options (symmetric and no string names) 
```
model = Model(Hypatia.Optimizer)
@variable(model, [1:2, 1:2], Symmetric)
```
it seems to try to create a matrix from a string
`MethodError: no method matching (Matrix)(::String)`
with the stacktrace containing
```
Stacktrace:
 [1] vectorize(matrix::String, shape::SymmetricMatrixShape)
   @ JuMP ~/.julia/packages/JuMP/Y4piv/src/sd.jl:216
 [2] add_variable(model::Model, variable::VariablesConstrainedOnCreation{MathOptInterface.Reals, SymmetricMatrixShape, ScalarVariable{Float64, Float64, Float64, Float64}}, names::String)
   @ JuMP ~/.julia/packages/JuMP/Y4piv/src/variables.jl:1220
```

My `Pkg.status()` used for testing
```
      Status `~/.julia/environments/v1.7/Project.toml`
  [b99e6be6] Hypatia v0.7.0
  [7073ff75] IJulia v1.23.3
  [4076af6c] JuMP v1.1.1
  [0f1e0344] WebIO v0.8.18
```

odow:
Closes #3035

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1212,6 +1212,15 @@ function VariablesConstrainedOnCreation(
     return VariablesConstrainedOnCreation(variables, set, VectorShape())
 end
 
+_vectorize_names(names, shape) = vectorize(names, shape)
+
+# In some cases `names` may be a "" passed in from the macros. For now, this is
+# the only case we support, so throw an assertion error if not empty.
+function _vectorize_names(name::String, ::Any)
+    @assert isempty(name)
+    return nothing
+end
+
 function add_variable(
     model::Model,
     variable::VariablesConstrainedOnCreation,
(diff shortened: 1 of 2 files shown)


src/variables.jl
```diff
@@ -1227,6 +1227,21 @@ function add_variable(
...
+function add_variable(
+    model::Model,
+    variable::VariablesConstrainedOnCreation,
+    name::String,
+)
+    var_indices = _moi_add_constrained_variables(
+        backend(model),
+        variable.scalar_variables,
+        variable.set,
+        fill(name, length(variable.scalar_variables)),
```
blegat, line 1239: If the purpose of https://github.com/jump-dev/JuMP.jl/pull/2978/ was performance, maybe we don't want to create this dummy array identical strings ? Either using FillArray or adding support for it in `_moi_add_constrained_variables`.

src/variables.jl
```diff
@@ -1227,6 +1227,21 @@ function add_variable(
     return reshape_vector(var_refs, variable.shape)
 end
 
+function add_variable(
+    model::Model,
+    variable::VariablesConstrainedOnCreation,
+    name::String,
```
blegat, line 1233: Is it allowed to call this function with a non-empty name ?

src/variables.jl
odow: The performance problem was passing the names to solvers, not this allocation.

odow: Okay, I got rid of this.

PR #3044, opened 2022-08-17
[docs] clarify singleton set confusions

### LINKED ISSUE #3042: [docs] Clarify `String` singleton as an iterable object for `@variable`
See: https://discourse.julialang.org/t/jump-variable/85853

This seems like it shouldn't happen from a user perspective without at least a warning.

odow:
Closes #3042 

cc @jd-foster

### DIFF
--- a/docs/src/tutorials/getting_started/getting_started_with_sets_and_indexing.jl
+++ b/docs/src/tutorials/getting_started/getting_started_with_sets_and_indexing.jl
@@ -118,14 +118,17 @@ x["A", "D"]
 
 # ## Sets to watch out for
 
-# JuMP supports any sets which are iterable, that is, the set `set` supports
-# a for-loop like: `[i for i in set]`. This causes a few common errors.
+# JuMP supports any sets which are [iterable] (https://docs.julialang.org/en/v1/base/collections/),
+# that is, the set `set` supports a for-loop like: `[i for i in set]`. This
+# causes a few common errors.
 
-# First, if `T = 3`, you may pass the integer `T` instead of the range `1:T`:
+# First, if `T = 3`, you may pass the integer `T` by mistake instead of a range
+# like `1:T`:
 model = Model()
 T = 3
 @variable(model, x[T])
 
+# This results in a single variable being created, instead of three as desired.
 # Because this is a common error, a warning is printed, advising you to pass a
 # `Vector{Int}` instead:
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/getting_started/getting_started_with_sets_and_indexing.jl
```diff
@@ -116,6 +116,36 @@ x["A", "D"]
...
+
+@variable(model, x_fixed[[T]])
+
+# Second, because `String`s are iterable, passing a `"abc"` as a singleton index
+# is the same as passing `['a', 'b', 'c']`:
+
+@variable(model, y["abc"])
+
+# This time, a warning is not printed (because there are occasions when this
+# behavior is desired), but the work-around is similar, pass a `Vector{String}`
```
mlubin, line 140: It seems like a stretch to say there are occasions when this behavior is desired. At best it's bad style.

odow: Then should we add a warning for `String` like we do for `Int`?

mlubin: The warning "Axis contains one element" doesn't apply for most `String`s. Adding warnings to steer people away from a certain syntax is arguably a breaking change as well. I don't think this corner case is worth releasing JuMP 2.0 for.

odow: Removed the parenthetic statement.

mlubin (approved):

docs/src/tutorials/getting_started/getting_started_with_sets_and_indexing.jl
```diff
@@ -116,6 +116,35 @@ x["A", "D"]
 # !!! info
 #     Note how we indexed `x["A", "D"]` instead of `x[("A", "D")]` as above.
 
+# ## Sets to watch out for
+
+# JuMP supports any sets which are iterable, that is, the set `set` supports
```
jd-foster, line 121: [iterable](https://docs.julialang.org/en/v1/base/collections/)

docs/src/tutorials/getting_started/getting_started_with_sets_and_indexing.jl
```diff
@@ -116,6 +116,35 @@ x["A", "D"]
...
+# a for-loop like: `[i for i in set]`. This causes a few common errors.
+
+# First, if `T = 3`, you may pass the integer `T` instead of the range `1:T`:
```
jd-foster, line 124: "instead of range" -> "and index a single variable instead of indexing over the intended range `1:T`:"

PR #3061, opened 2022-08-30
[docs] reduce some unnecessary outputs

odow:
Some of the LP tutorials contain a lot of unnecessary output (pages of printed variables, or multiple plots in sequence).

They currently finish at page 179, so lets see how many pages we claw back. Update: 164, so we saved 15 pages!

Preview: https://jump.dev/JuMP.jl/previews/PR3061/JuMP.pdf

### DIFF
--- a/docs/src/tutorials/linear/facility_location.jl
+++ b/docs/src/tutorials/linear/facility_location.jl
@@ -72,12 +72,10 @@ m = 12
 n = 5
 
 ## Clients' locations
-Xc = rand(m)
-Yc = rand(m)
+Xc, Yc = rand(m), rand(m)
 
 ## Facilities' potential locations
-Xf = rand(n)
-Yf = rand(n)
+Xf, Yf = rand(n), rand(n)
 
 ## Fixed costs
 f = ones(n);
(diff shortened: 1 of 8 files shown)


docs/src/tutorials/linear/geographic_clustering.jl
```diff
@@ -33,96 +33,30 @@ import HiGHS
 import LinearAlgebra
 
 # For this example, we'll use the 20 most populous cities in the United States.
-
 cities = DataFrames.DataFrame(;
```
odow, line 36:
```suggestion
cities = DataFrames.DataFrame(
```

docs/src/tutorials/linear/facility_location.jl
```diff
@@ -282,30 +271,27 @@ p = Plots.scatter(
...
-        end
+for i in 1:m, j in 1:n
+    if x_[i, j] == 1
+        Plots.plot!(
+            [Xc[i], Xf[j]],
+            [Yc[i], Yf[j]];
+            color = :black,
+            label = nothing,
+        )
+        break
```
odow, line 293:
```suggestion
```

PR #3068, opened 2022-09-06
[docs] add more examples to nlp.md

### LINKED ISSUE #3067: Request for the capability to obtain the start value of the NLexpressions
1. The problem and necessity

I am using JuMP to develop a computable general equilibrium model which is important for climate change economic studies. In general, the CGE model is developed by the GAMS but is not open-source, since the PATH solver is not freely accessible. 

JuMP maybe an useful tool to build the open-source CGE model because of its packages: PATHSolver.jl and Complementarity.jl. However, I feel that those packages are still not perfect to build the model. One necessary function is to evaluate whether the nonlinear equations are balanced in the initial value of variables. So it would be quite useful to have an API to **directly obtain the initial value of the @NLexpression.**

2. Possible solution
Currently, I found an API named **JuMP.start_value()**, which could obtain the initial value of the variables. So I am asking for help to extend the function of this API to obtain the initial value of expressions. This function seems to be easy to handle: **1. initializing the value of variables in the expression, 2. calculating the initial value of expression, 3. return the value.** 

This is one expression printed out: subexpression[58]: ff[16] * pff[16] ^ 0.3 - 0.0 * prod[16] * pprod[16] ^ 0.

If it is possible, when I code **start_value(expr_name)**, I can obtain the value of the expression.

Thanks for your help!

odow:
Closes #3067

### DIFF
--- a/docs/src/manual/nlp.md
+++ b/docs/src/manual/nlp.md
@@ -97,7 +97,7 @@ julia> @NLexpression(model, nested[i = 1:2], sin(my_expr[i]))
 
 Use [`value`](@ref) to query the value of a nonlinear expression:
 ```jldoctest nl_expression
-julia> set_start_value.(x, [1.0, 2.0])
+julia> set_start_value(x[1], 1.0)
 
 julia> value(start_value, nested[1])
 0.7456241416655579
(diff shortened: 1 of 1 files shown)


docs/src/manual/nlp.md
```diff
@@ -90,6 +95,17 @@ julia> @NLexpression(model, nested[i = 1:2], sin(my_expr[i]))
  subexpression[7]: sin(subexpression[5])
 ```
 
+Use [`value`](@ref) to query the value of a nonlinear expression:
+```jldoctest nl_expression
+julia> set_start_value.(x, [1.0, 2.0])
```
odow, line 100:
```suggestion
julia> set_start_value(x[1], 1.0)
```

PR #3075, opened 2022-09-15
[docs] override style in PDF output

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/pull/1999

https://jump.dev/JuMP.jl/previews/PR3075/JuMP.pdf


odow:
It may take a while for Documenter v0.28 to be released with some changes, but in the mean time, this is muuuuch nicer than what we had before.

PR #3080, opened 2022-09-18
[docs] add Vale as a documentation linter

odow:
Experimenting again with https://vale.sh.

It found a few typos.


mlubin:
That's a fair number of typos.

odow:
Maintaining the accept list seems like it could be a pain. But the real issue is that GitHub doesn't have any linting in the web editor, so it's hard to catch these in review.

Anyway. More tooling seems like positive progress if it leads to a higher quality product.

blegat (approved):
This is nice, I didn't know this existed

odow:
This needs a new tag of MOI so we don't fail on the errors in MOI's docs.

docs/styles/Vocab/JuMP-Vocab/accept.txt
```diff
@@ -0,0 +1,149 @@
...
+MOI
+Mosek
+MOSEK
+Pajarito
+PATHSolver
+Penbmi
+puzzlor
+RAPOSa
+Xpress
+
```
joaquimg, line 105: empty?

PR #3087, opened 2022-09-23
Improve performance of `relax_integrality`

### LINKED ISSUE #3086: `relax_integrality` takes really long for large problems even if there aren't that many integer variables
Example:

```
julia> model
A JuMP Model
Minimization problem with:
Variables: 1193206
`AffExpr`-in-`MathOptInterface.EqualTo{Float64}`: 399576 constraints
`AffExpr`-in-`MathOptInterface.GreaterThan{Float64}`: 43031 constraints
`AffExpr`-in-`MathOptInterface.LessThan{Float64}`: 904612 constraints
`VariableRef`-in-`MathOptInterface.GreaterThan{Float64}`: 1144294 constraints
`VariableRef`-in-`MathOptInterface.LessThan{Float64}`: 48912 constraints
`VariableRef`-in-`MathOptInterface.ZeroOne`: 48912 constraints
Model mode: AUTOMATIC

julia> @time JuMP.relax_integrality(model)
 44.611944 seconds (33.02 M allocations: 1.550 GiB, 1.33% gc time, 0.82% compilation time)
```

Note that although the problem is large (>1m constraints and >2.5m constraints), there are just around 50k binaries.
I suspect the reason for the poor performance is that in [relax_integrality](https://github.com/jump-dev/JuMP.jl/blob/08dae8502aa01fd22e965a31cbe2fbcc1804bc77/src/variables.jl#L1421) we iterate over `all_variables(model)`.
Not sure if there's a straightforward improvement that still provides informative error messages, but my guess is that iterating over just the integer/binary ones would reduce this runtime.

raphaelsaavedra:
Attempt to close #3086 

This makes relaxing my problem with the following size:

```julia
julia> model
A JuMP Model
Minimization problem with:
Variables: 1193207
`AffExpr`-in-`MathOptInterface.EqualTo{Float64}`: 399576 constraints
`AffExpr`-in-`MathOptInterface.GreaterThan{Float64}`: 43031 constraints
`AffExpr`-in-`MathOptInterface.LessThan{Float64}`: 904612 constraints
`VariableRef`-in-`MathOptInterface.GreaterThan{Float64}`: 1144294 constraints
`VariableRef`-in-`MathOptInterface.LessThan{Float64}`: 48912 constraints
`VariableRef`-in-`MathOptInterface.Integer`: 1 constraint
`VariableRef`-in-`MathOptInterface.ZeroOne`: 48912 constraints
Model mode: AUTOMATIC
```

go from 

```julia
julia> @time JuMP.relax_integrality(model)
 44.611944 seconds (33.02 M allocations: 1.550 GiB, 1.33% gc time, 0.82% compilation time)
```

to 

```julia
julia> @time JuMP.relax_integrality(model)
  1.818470 seconds (2.73 M allocations: 105.338 MiB, 6.19% compilation time: 66% of which was recompilation)
```

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1419,25 +1419,17 @@ Subject to
 ```
 """
 function relax_integrality(model::Model)
-    semicont_type = _MOICON{MOI.VariableIndex,MOI.Semicontinuous{Float64}}
-    semiint_type = _MOICON{MOI.VariableIndex,MOI.Semiinteger{Float64}}
-    semicont_semiint_constraints = vcat(
-        all_constraints(model, VariableRef, MOI.ZeroOne),
-        all_constraints(model, VariableRef, MOI.Integer),
-    )
-    semicont_semiint_variables = VariableRef.(semicont_semiint_constraints)
-    for v in semicont_semiint_variables
-        if MOI.is_valid(backend(model), semicont_type(index(v).value))
-            error(
-                "Support for relaxing semicontinuous constraints is not " *
-                "yet implemented.",
-            )
-        elseif MOI.is_valid(backend(model), semiint_type(index(v).value))
-            error(
-                "Support for relaxing semi-integer constraints is not " *
-                "yet implemented.",
-            )
-        end
+    if num_constraints(model, VariableRef, MOI.Semicontinuous{Float64}) > 0
+        error(
+            "Support for relaxing semicontinuous constraints is not " *
+            "yet implemented.",
+        )
+    end
+    if num_constraints(model, VariableRef, MOI.Semiinteger{Float64}) > 0
+        error(
+            "Support for relaxing semi-integer constraints is not " *
+            "yet implemented.",
+        )
     end
 
     bin_int_constraints = vcat(
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1421,13 +1421,12 @@ Subject to
 function relax_integrality(model::Model)
     semicont_type = _MOICON{MOI.VariableIndex,MOI.Semicontinuous{Float64}}
     semiint_type = _MOICON{MOI.VariableIndex,MOI.Semiinteger{Float64}}
-    # Get only the binary and integer variables, as the others don't need to be modified.
-    bin_int_constraints = vcat(
+    semicont_semiint_constraints = vcat(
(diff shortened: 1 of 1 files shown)


src/variables.jl
```diff
@@ -1421,7 +1421,13 @@ Subject to
...
     semicont_type = _MOICON{MOI.VariableIndex,MOI.Semicontinuous{Float64}}
     semiint_type = _MOICON{MOI.VariableIndex,MOI.Semiinteger{Float64}}
-    for v in all_variables(model)
+    # Get only the binary and integer variables, as the others don't need to be modified.
+    bin_int_constraints = vcat(
+        all_constraints(model, VariableRef, MOI.ZeroOne),
+        all_constraints(model, VariableRef, MOI.Integer),
+    )
+    bin_int_variables = VariableRef.(bin_int_constraints)
+    for v in bin_int_constraints
```
mlubin, line 1430: Please continue to check for the presence of semicont and semiint variables among all variables. The current logic is nonsense because a variable can't be both binary/integer and semicont/semiint. You should be able to do this without an expensive loop by asking for the number of semicontinuous or semiinteger constraints.

raphaelsaavedra: Ah I see, I misunderstood what was happening in this loop. Please take a look if it's correct now.

src/variables.jl
```diff
@@ -1421,7 +1421,12 @@ Subject to
 function relax_integrality(model::Model)
     semicont_type = _MOICON{MOI.VariableIndex,MOI.Semicontinuous{Float64}}
     semiint_type = _MOICON{MOI.VariableIndex,MOI.Semiinteger{Float64}}
-    for v in all_variables(model)
+    semicont_semiint_constraints = vcat(
+        all_constraints(model, VariableRef, MOI.ZeroOne),
+        all_constraints(model, VariableRef, MOI.Integer),
+    )
+    semicont_semiint_variables = VariableRef.(semicont_semiint_constraints)
+    for v in semicont_semiint_variables
```
mlubin, line 1429: This code is more complex than needed. You could do something like:
```
if num_constraints(model, VariableRef, MOI.Semicontinuous{Float64}) > 0:
    error("Support for relaxing semicontinuous ...")
end
# Same for Semiinteger
```

mlubin (approved):

PR #3088, opened 2022-09-25
Pass names (index_vars) to container extensions.

hellemo:
Our container in `SparseVariables` supports storing the names of the subindices (`index_vars`), but these are not currently passed to containers.

See https://github.com/hellemo/SparseVariables.jl/issues/19

### DIFF
--- a/src/Containers/container.jl
+++ b/src/Containers/container.jl
@@ -15,7 +15,7 @@ a `VectorizedProductIterator` and the function returns
 function default_container end
 
 """
-    container(f::Function, names, indices, c::Type{C})
+    container(f::Function, indices, c::Type{C}, names)
 
 Create a container of type `C` with index names `names`, indices `indices` and
 values at given indices given by `f`. If this method is not specialized on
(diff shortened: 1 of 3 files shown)


src/Containers/container.jl
```diff
@@ -178,3 +178,8 @@ function container(::Function, ::Any, D::Type)
         "`Containers.container(::Function, indices, ::Type{$(D)})`.",
     )
 end
+
+# Default to dropping names (as before)
+function container(f::Function, names, indices, D)
```
odow, line 183: This method can probably be moved to line 63, and the docstring updated to reflect the additional argument.

odow (commented):
One thought I had: it might make sense to make `names` the last argument? Then we'd have a 4, 3, 2 argument methods, each dropping the last argument.

hellemo:
Yeah, I guess so. My reasoning was that names go naturally together with the indices they refer to, but there is a certain appeal to dropping (or adding) the last argument and keep the first arguments the same. No strong opinions from my side, can update if you prefer.

odow:
Yeah, I think please swap. Let's have consistency in the dropping argument order.

odow (approved):

blegat (approved):

PR #3102, opened 2022-10-01
[docs] tweak the order and content of should_i_use

odow:
A couple of tweaks to the order of `When should I use JuMP?` to make "Speed" less important. 

Put another way, I think our selling point is "nice syntax and by the way, it's pretty fast" instead of "it's fast, and by the way, we have with nice syntax."

### DIFF
--- a/docs/src/should_i_use.md
+++ b/docs/src/should_i_use.md
@@ -9,7 +9,8 @@ you should _not_ use JuMP.
 ## When should I use JuMP?
 
 You should use JuMP if you have a constrained optimization problem for which you
-can formulate:
+can formulate using the language of mathematical programming, that is:
+
  * a set of decision variables
  * a scalar objective function
  * a set of constraints.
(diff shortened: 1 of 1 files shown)


mlubin (approved):

docs/src/should_i_use.md
```diff
@@ -102,25 +104,30 @@ need to write out the dynamics in algebraic form) that mean JuMP might not be
...
  * [vOptGeneric.jl](https://github.com/vOptSolver/vOptGeneric.jl)
 
 ### Disciplined convex programming
 
 JuMP does not support [disciplined convex programming (DCP)](https://dcp.stanford.edu).
 
 Alternatives to consider are:
+
  * [Convex.jl](https://github.com/jump-dev/Convex.jl)
+ * [cvxpy [Python]](https://github.com/cvxpy/cvxpy)
```
odow, line 129:
```suggestion
 * [CVXPY [Python]](https://github.com/cvxpy/cvxpy)
```

Issue #3096, opened 2022-09-28
Tables.jl support for results


hellemo:
We are considering moving the Tables.jl support for results from SparseVariables.jl to another package. (https://github.com/hellemo/SparseVariables.jl/issues/20). One issue then is whether support for the default containers from JuMP should be in a separate package from the SparseVariables type(s) or not. 

One alternative would be to add `Tables.jl` support for `DenseAxisArrays` and `SparseAxisArrays` to JuMP, as Tables is designed to be a light dependency (and keep the Tables support for our own types in our package). What do you think?

Some other things to consider (that might go together with this or belong in separate issues) are:

* It would be nice if `DenseAxisArrays` and `SparseAxisArrays` would also store the subindex names/index_vars, in our current implementation these have to be added explicitly when generating e.g. a `DataFrame` with the variable values from the solver (kind of a follow-up to https://github.com/jump-dev/JuMP.jl/pull/3088). With names stored with the container, one could generate result tables with good default column names automatically.
* Performance of reading results from solvers. I haven't looked into this yet, but getting large results feels kind of slow. Fast read of results that can easily be dropped into dataframes, plots, csv etc would be a nice usability improvement.

odow:
Related https://discourse.julialang.org/t/efficient-serialization-of-jump-results/87896

cc @jd-foster 

I haven't really looked at the Tables.jl stuff, but what would the output be? Columns with the keys and then one value column? That could be quite nice, and would work for both Dense and Sparse

hellemo:
Yes, exactly. Tables support for Dense is already implemented in SparseVariables, and I think Sparse is in the works (@trulsf ?)

For the example in the readme (after solving):
```
julia> # Read solution values for u (DenseAxisArray)
       DataFrame(table(u, :u, :car, :year))
12×3 DataFrame
│ Row │ car    │ year  │ u       │
│     │ String │ Int64 │ Float64 │
├─────┼────────┼───────┼─────────┤
│ 1   │ ford   │ 2000  │ 1.0     │
│ 2   │ bmw    │ 2000  │ 1.0     │
│ 3   │ opel   │ 2000  │ 1.0     │
│ 4   │ ford   │ 2001  │ 1.0     │
│ 5   │ bmw    │ 2001  │ 1.0     │
│ 6   │ opel   │ 2001  │ 1.0     │
│ 7   │ ford   │ 2002  │ 1.0     │
│ 8   │ bmw    │ 2002  │ 1.0     │
[...]

PR #3104, opened 2022-10-04
Tables.jl support

trulsf:
Support for solution tables supporting the `Tables.jl` interface for JuMP variable containers. Currently has support for `Array`, `DenseAxisArray`, `SparseAxisArray`. 

Closes https://github.com/jump-dev/JuMP.jl/issues/3096

### DIFF
--- a/src/Containers/tables.jl
+++ b/src/Containers/tables.jl
@@ -7,11 +7,11 @@ function _row_iterator(x::Array)
     return zip(eachindex(x), Iterators.product(axes(x)...))
 end
 
-function _row_iterator(x::Containers.DenseAxisArray)
+function _row_iterator(x::DenseAxisArray)
     return zip(vec(eachindex(x)), Iterators.product(axes(x)...))
 end
 
-function _row_iterator(x::Containers.SparseAxisArray)
+function _row_iterator(x::SparseAxisArray)
     return zip(eachindex(x.data), keys(x.data))
 end
 
@@ -35,7 +35,7 @@ julia> model = Model();
 
 julia> @variable(model, x[i=1:2, j=i:2] >= 0, start = i+j);
 
-julia> table(start_value, x, :start, :I, :J)
+julia> Containers.table(start_value, x, :start, :I, :J)
 3-element Vector{NamedTuple{(:I, :J, :start), Tuple{Int64, Int64, Float64}}}:
  (I = 1, J = 2, start = 3.0)
  (I = 1, J = 1, start = 2.0)
@@ -44,7 +44,7 @@ julia> table(start_value, x, :start, :I, :J)
 """
 function table(
     f::Function,
-    x::Union{Array,Containers.DenseAxisArray,Containers.SparseAxisArray},
+    x::Union{Array,DenseAxisArray,SparseAxisArray},
     value_name::Symbol,
     col_names::Symbol...,
 )
--- a/src/Containers/tables.jl
+++ b/src/Containers/tables.jl
@@ -10,14 +10,16 @@ _rows(x::DenseAxisArray) = zip(vec(eachindex(x)), Iterators.product(axes(x)...))
 _rows(x::SparseAxisArray) = zip(eachindex(x.data), keys(x.data))
 
 """
-    rowtable([f::Function=identity,] x, [names::Symbol...])
+    rowtable([f::Function=identity,] x; [header::Vector{Symbol} = Symbol[]])
 
 Applies the function `f` to all elements of the variable container `x`,
-returning the result as a `Vector` of `NamedTuple`s, where `names` are used for
-the corresponding axis names. If `x` is an `N`-dimensional array, there must be
-`N+1` names, so that the last name corresponds to the result of `f(x[i])`.
+returning the result as a `Vector` of `NamedTuple`s, where `header` is a vector
+containing the corresponding axis names.
 
-If `names` are omitted, then the default names are `(:x1, :x2, ..., :xN, :y)`.
+If `x` is an `N`-dimensional array, there must be `N+1` names, so that the last
+name corresponds to the result of `f(x[i])`.
+
+If `header` is left empty, then the default header is `[:x1, :x2, ..., :xN, :y]`.
 
 !!! info
     A `Vector` of `NamedTuple`s implements the [Tables.jl](https://github.com/JuliaData/Tables.jl)
@@ -31,7 +33,7 @@ julia> model = Model();
 
 julia> @variable(model, x[i=1:2, j=i:2] >= 0, start = i+j);
 
-julia> Containers.rowtable(start_value, x, :i, :j, :start)
+julia> Containers.rowtable(start_value, x; header = [:i, :j, :start])
 3-element Vector{NamedTuple{(:i, :j, :start), Tuple{Int64, Int64, Float64}}}:
  (i = 1, j = 2, start = 3.0)
  (i = 1, j = 1, start = 2.0)
@@ -46,19 +48,21 @@ julia> Containers.rowtable(x)
 """
 function rowtable(
     f::Function,
-    x::Union{Array,DenseAxisArray,SparseAxisArray},
-    names::Symbol...,
+    x::Union{Array,DenseAxisArray,SparseAxisArray};
+    header::Vector{Symbol} = Symbol[],
 )
-    if length(names) == 0
-        return rowtable(f, x, [Symbol("x$i") for i in 1:ndims(x)]..., :y)
+    if isempty(header)
+        header = Symbol[Symbol("x$i") for i in 1:ndims(x)]
+        push!(header, :y)
     end
-    got, want = length(names), ndims(x) + 1
+    got, want = length(header), ndims(x) + 1
     if got != want
         error(
             "Invalid number of column names provided: Got $got, expected $want.",
         )
     end
+    names = tuple(header...)
     return [NamedTuple{names}((args..., f(x[i]))) for (i, args) in _rows(x)]
 end
 
-rowtable(x, names::Symbol...) = rowtable(identity, x, names...)
+rowtable(x; kwargs...) = rowtable(identity, x; kwargs...)
--- a/src/Containers/tables.jl
+++ b/src/Containers/tables.jl
@@ -43,7 +43,7 @@ function table(
 )
     got, want = length(names), ndims(x) + 1
     if got != want
-        error("Invalid number column names provided: Got $got, expected $want.")
+        error("Invalid number of column names provided: Got $got, expected $want.")
     end
     return [NamedTuple{names}((args..., f(x[i]))) for (i, args) in _rows(x)]
 end
--- a/src/Containers/tables.jl
+++ b/src/Containers/tables.jl
@@ -10,12 +10,14 @@ _rows(x::DenseAxisArray) = zip(vec(eachindex(x)), Iterators.product(axes(x)...))
 _rows(x::SparseAxisArray) = zip(eachindex(x.data), keys(x.data))
 
 """
-    table([f::Function=identity,] x, names::Symbol...)
+    table([f::Function=identity,] x, [names::Symbol...])
 
 Applies the function `f` to all elements of the variable container `x`,
 returning the result as a `Vector` of `NamedTuple`s, where `names` are used for
 the corresponding axis names. If `x` is an `N`-dimensional array, there must be
 `N+1` names, so that the last name corresponds to the result of `f(x[i])`.
+
+If `names` are omitted, then the default names are `(:x1, :x2, ..., :xN, :y)`.
 
 !!! info
     A `Vector` of `NamedTuple`s implements the [Tables.jl](https://github.com/JuliaData/Tables.jl)
@@ -34,6 +36,12 @@ julia> Containers.table(start_value, x, :i, :j, :start)
  (i = 1, j = 2, start = 3.0)
  (i = 1, j = 1, start = 2.0)
  (i = 2, j = 2, start = 4.0)
+
+julia> Containers.table(x)
+3-element Vector{NamedTuple{(:x1, :x2, :y), Tuple{Int64, Int64, VariableRef}}}:
+ (x1 = 1, x2 = 2, y = x[1,2])
+ (x1 = 1, x2 = 1, y = x[1,1])
+ (x1 = 2, x2 = 2, y = x[2,2])
 ```
 """
 function table(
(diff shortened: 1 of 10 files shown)


src/JuMP.jl
```diff
@@ -1360,6 +1360,9 @@ include("callbacks.jl")
 include("file_formats.jl")
 include("feasibility_checker.jl")
 
+using Tables
```
blegat, line 1363: Use `import`; see https://jump.dev/JuMP.jl/dev/developers/style/#using-vs.-import

odow:
So I took a look at this. Any reason that we can't just return a `Vector{NamedTuple}`? This satisfies the tables interface without us needing the dependency, and the code is pretty trivial:

```julia
using JuMP
model = Model()
@variable(model, x[1:2, 1:2])
@variable(model, y[1:2, 1:2], container = DenseAxisArray)
@variable(model, z[i=1:3, j=1:3; isodd(i+j)], container = SparseAxisArray)

function _row_iterator(x::Union{Array,Containers.DenseAxisArray})
    return zip(eachindex(x), Iterators.product(axes(x)...))
end

function _row_iterator(x::Containers.SparseAxisArray)
    return zip(eachindex(x.data), keys(x.data))
end

function table(x, name::Symbol, col_names::Symbol...)
    return table(identity, x, name, col_names...)
end

function table(
    f::Function, 
    x::Union{Array,Containers.DenseAxisArray,Containers.SparseAxisArray}, 
    name::Symbol, 
    col_names::Symbol...,
)
    C = (col_names..., name)
    return [NamedTuple{C}((args..., f(x[i]))) for (i, args) in _row_iterator(x)]
end

julia> for row in table(JuMP.name, x, :value, :A, :B)
           @show row
       end
row = (A = 1, B = 1, value = "x[1,1]")
row = (A = 2, B = 1, value = "x[2,1]")
row = (A = 1, B = 2, value = "x[1,2]")
row = (A = 2, B = 2, value = "x[2,2]")

julia> for row in table(JuMP.name, y, :value, :A, :B)
           @show row
       end
row = (A = 1, B = 1, value = "y[1,1]")
row = (A = 2, B = 1, value = "y[2,1]")
row = (A = 1, B = 2, value = "y[1,2]")
row = (A = 2, B = 2, value = "y[2,2]")

julia> for row in table(z, :value, :A, :B)
           @show row
       end
row = (A = 3, B = 2, value = z[3,2])
row = (A = 1, B = 2, value = z[1,2])
row = (A = 2, B = 1, value = z[2,1])
row = (A = 2, B = 3, value = z[2,3])

julia> import PrettyTables

julia> PrettyTables.pretty_table(table(z, :value, :A, :B))
┌───────┬───────┬─────────────┐
│     A │     B │       value │
│ Int64 │ Int64 │ VariableRef │
├───────┼───────┼─────────────┤
│     3 │     2 │      z[3,2] │
│     1 │     2 │      z[1,2] │
│     2 │     1 │      z[2,1] │
│     2 │     3 │      z[2,3] │
└───────┴───────┴─────────────┘
```

trulsf:
I was not aware that a vector of NamedTuple implemented the Tables.jl interface, but it makes sense. I should read the docs more carefully next time :)

The only problem with your suggested implementation is that you return an array of NamedTuples instead of a a vector for Array and DenseAxisArray, but it is fixed by e.g. 
```julia
return vec([NamedTuple{C}((args..., f(x[i]))) for (i, args) in _row_iterator(x)])
```
(Maybe more efficient to fix it in the iterator?)

I can update the PR with your code or is it more efficient that you take this further if it is to be added to JuMP?

odow:
You can steal my code and update the example :)

odow (commented):
@trulsf I tidied things up a little.

Questions:

 * Should this go in `JuMP.table` or `JuMP.Containers.table`? It seems like JuMP automatically exporting `table` could lead to clashes with other packages?
 * Is `value_name, col_names` the correct order? It seems to lead to a counter-intuitive ordering of the resulting named tuple. What if we just had `names...`?
 
We should also think about what documentation this needs:

```Julia
julia> using JuMP, DataFrames

julia> x = Containers.@container([i = 1:2, j = [:A, :B]], (i, j))
2-dimensional DenseAxisArray{Tuple{Int64, Symbol},2,...} with index sets:
    Dimension 1, Base.OneTo(2)
    Dimension 2, [:A, :B]
And data, a 2×2 Matrix{Tuple{Int64, Symbol}}:
 (1, :A)  (1, :B)
 (2, :A)  (2, :B)

julia> DataFrames.DataFrame(table(x, :x, :i, :j))  # <- here's the ordering that feels wrong
4×3 DataFrame
 Row │ i      j       x       
     │ Int64  Symbol  Tuple…  
─────┼────────────────────────
   1 │     1  A       (1, :A)
   2 │     2  A       (2, :A)
   3 │     1  B       (1, :B)
   4 │     2  B       (2, :B)
```

odow:
After merging, we should update the relevant discourse posts:

 * https://discourse.julialang.org/t/extracting-jump-results/51429
 * https://discourse.julialang.org/t/converting-dense-axis-array-results-into-dataframe/83780

src/Containers/tables.jl
```diff
@@ -0,0 +1,51 @@
...
+```
+"""
+function table(
+    f::Function,
+    x::Union{Array,DenseAxisArray,SparseAxisArray},
+    names::Symbol...,
+)
+    got, want = length(names), ndims(x) + 1
+    if got != want
+        error("Invalid number column names provided: Got $got, expected $want.")
```
odow, line 46:
```suggestion
        error("Invalid number of column names provided: Got $got, expected $want.")
```

odow (approved):
I think I answered my two questions, and I added some docs. I think this will be quite useful.

src/Containers/tables.jl
```diff
@@ -0,0 +1,51 @@
...
+```
+"""
+function table(
+    f::Function,
+    x::Union{Array,DenseAxisArray,SparseAxisArray},
+    names::Symbol...,
+)
+    got, want = length(names), ndims(x) + 1
+    if got != want
+        error("Invalid number of column names provided: Got $got, expected $want.")
```
odow, line 46:
```suggestion
        error(
            "Invalid number of column names provided: Got $got, expected $want.",
        )
```

mlubin:
I'd like to take a look before we merge this. Please bug me if I don't review it quickly.

odow:
Sure. I guess one other potential change is to call it `Containers.rowtable` to match https://tables.juliadata.org/stable/#Tables.rowtable and to reflect the fact that we're returning a one-dimensional object.

hellemo:
Very nice!

Did you consider storing the names in the containers, which would simplify the use of tables further (as mentioned in #3096 and demonstrated in the last example), following up on #3088 ?
```
julia> using PrettyTables

julia> pretty_table(table(y)) # SparseVarArray
┌──────┬──────┬─────┐
│  car │ year │   y │
├──────┼──────┼─────┤
│  bmw │ 2001 │ 0.0 │
│ ford │ 2001 │ 0.0 │
│ ford │ 2000 │ 0.0 │
│  bmw │ 2002 │ 0.0 │
└──────┴──────┴─────┘```

**Comment by odow** [comment]
> Did you consider storing the names in the containers

Not yet. That's a larger, more orthogonal change. We'd have to think about how names etc are preserved during broadcasting etc. Otherwise someone would say "why does `table(x)` work, but `table(value.(x))` not?

**Comment by hellemo** [comment]
> Not yet. That's a larger, more orthogonal change. We'd have to think about how names etc are preserved during broadcasting etc. Otherwise someone would say "why does `table(x)` work, but `table(value.(x))` not?

I guess I imagined the scope of `table` to be narrower and more of an alternative way to get (all) the values. So I don't quite understand why you would want to do `table(value.(x))` when `table(x)`  is shorter and hopefully optimized for fetching all values at once. A follow-up question is of course your thoughts on how to make this possible/easy (assuming it is supported by solver APIs, which I haven't checked). Subsetting or slicing might complicate this approach a bit, I imagine.

I was kind of hoping to get rid of the call to tables where you have to specify column names (you already specified them once), and rather provide defaults if they are not specified to the macro on container creation, like e.g. DataFrames.

Just picking up on these in case they are important to consider at this point. (In terms of designing API for table and considering possibilities to optimize fetching of solution values)

**Comment by odow** [comment]
> So I don't quite understand why you would want to do table(value.(x)) when table(x) is shorter and hopefully optimized for fetching all values at once

The current design doesn't assume `value` is used. You can either go `table(value, x, names...)` or `table(value.(x), names...)` (the default map function is `identity`).

The problem with names is this:
```julia
julia> using JuMP.Containers

julia> Containers.@container(x[a=1:2, b=1:2], a + b, container = DenseAxisArray)
2-dimensional DenseAxisArray{Int64,2,...} with index sets:
    Dimension 1, Base.OneTo(2)
    Dimension 2, Base.OneTo(2)
And data, a 2×2 Matrix{Int64}:
 2  3
 3  4

julia> Containers.@container(y[i=1:2, j=1:2], i + j, container = DenseAxisArray)
2-dimensional DenseAxisArray{Int64,2,...} with index sets:
    Dimension 1, Base.OneTo(2)
    Dimension 2, Base.OneTo(2)
And data, a 2×2 Matrix{Int64}:
 2  3
 3  4

julia> x .+ y
2-dimensional DenseAxisArray{Int64,2,...} with index sets:
    Dimension 1, Base.OneTo(2)
    Dimension 2, Base.OneTo(2)
And data, a 2×2 Matrix{Int64}:
 4  6
 6  8
```
Currently, we consider two arrays to be alike if the have the same axes. The "name" associated with each axis is ignored. 

> I was kind of hoping to get rid of the call to tables where you have to specify column names (you already specified them once), and rather provide defaults if they are not specified to the macro on container creation, like e.g. DataFrames.

A reasonable compromise might be `Symbol("x$I")` for the axes, and `:y` for the result, if no names are specified.

odow:
This is now:

```julia
julia> model = Model();

julia> @variable(model, x[i=1:2, j=i:2] >= 0, start = i+j);

julia> Containers.table(start_value, x, :i, :j, :start)
3-element Vector{NamedTuple{(:i, :j, :start), Tuple{Int64, Int64, Float64}}}:
 (i = 1, j = 2, start = 3.0)
 (i = 1, j = 1, start = 2.0)
 (i = 2, j = 2, start = 4.0)

julia> Containers.table(x)
3-element Vector{NamedTuple{(:x1, :x2, :y), Tuple{Int64, Int64, VariableRef}}}:
 (x1 = 1, x2 = 2, y = x[1,2])
 (x1 = 1, x2 = 1, y = x[1,1])
 (x1 = 2, x2 = 2, y = x[2,2])
```

mlubin:
Agreed with the `rowtable` naming. `table` is a bit confusing and seems to "underdeliver" by returning a `Vector{NamedTuple}` that you need additional packages to work with nicely.

odow:
Updated to `rowtable`. We can't add a method to `Tables.rowtable` because that would be piracy, and we return a slightly different notion of table:
```Julia
julia> using Tables

julia> x = rand(2, 2)
2×2 Matrix{Float64}:
 0.262588  0.086919
 0.199496  0.542848

julia> Tables.rowtable(x)
ERROR: ArgumentError: a 'Matrix{Float64}' is not a table; see `?Tables.table` for ways to treat an AbstractVecOrMat as a table
Stacktrace:
 [1] rows(m::Matrix{Float64})
   @ Tables ~/.julia/packages/Tables/T7rHm/src/matrix.jl:3
 [2] rowtable(itr::Matrix{Float64})
   @ Tables ~/.julia/packages/Tables/T7rHm/src/namedtuples.jl:105
 [3] top-level scope
   @ REPL[4]:1

julia> Tables.table(x)
Tables.MatrixTable{Matrix{Float64}} with 2 rows, 2 columns, and schema:
 :Column1  Float64
 :Column2  Float64

julia> Tables.rowtable(Tables.table(x))
2-element Vector{NamedTuple{(:Column1, :Column2), Tuple{Float64, Float64}}}:
 (Column1 = 0.262588126998895, Column2 = 0.08691902246653505)
 (Column1 = 0.19949566550192044, Column2 = 0.5428477642995122)
```

I wonder if we should follow the `; header::Vector{Symbol}` argument for the names. It'd be simpler.
```Julia
help?> Tables.table
  Tables.table(m::AbstractVecOrMat; [header])

  Wrap an AbstractVecOrMat (Matrix, Vector, Adjoint, etc.) in a MatrixTable, which satisfies the
  Tables.jl interface. (An AbstractVector is treated as a 1-column matrix.) This allows accessing
  the matrix via Tables.rows and Tables.columns. An optional keyword argument iterator header can
  be passed which will be converted to a Vector{Symbol} to be used as the column names. Note that
  no copy of the AbstractVecOrMat is made.

julia> Tables.table(x, header = [:x1, :x2])
Tables.MatrixTable{Matrix{Float64}} with 2 rows, 2 columns, and schema:
 :x1  Float64
 :x2  Float64

julia> Tables.rowtable(Tables.table(x, header = [:x1, :x2]))
2-element Vector{NamedTuple{(:x1, :x2), Tuple{Float64, Float64}}}:
 (x1 = 0.262588126998895, x2 = 0.08691902246653505)
 (x1 = 0.19949566550192044, x2 = 0.5428477642995122)
```

hellemo:
> The current design doesn't assume `value` is used. You can either go `table(value, x, names...)` or `table(value.(x), names...)` (the default map function is `identity`).

While I appreciate the elegance of this solution, doesn't this make it harder to specialize on a `table` method for `value` that is optimized for fetching all values at once? (or is there a problem with this idea more generally? - would appreciate some pointers on where to start looking into that)

> Currently, we consider two arrays to be alike if the have the same axes. The "name" associated with each axis is ignored.

You could still keep that behaviour, no? I mean, you already disregard the names initially. The suggested change would just be to keep them around for when they would be of use, such as when getting results via table.

> A reasonable compromise might be `Symbol("x$I")` for the axes, and `:y` for the result, if no names are specified.

Thanks.

odow:
> While I appreciate the elegance of this solution, doesn't this make it harder to specialize on a table method for value that is optimized for fetching all values at once? (or is there a problem with this idea more generally? - would appreciate some pointers on where to start looking into that)

You could write a specific `rowtable(::typeof(JuMP.value), x::SparseVarArray)` method to make it faster.

I would expect that `rowtable(value, x)` takes about the same time as `value.(x)`; it has to call `value(x[i])` for every element. But I think the utility of `rowtable` is not speed, but that it simplifies converting into tables.

odow:
Thanks @trulsf, I think this is a nice addition.

trulsf:
Thanks @odow for your work on incorporating this in JuMP. We are looking forward to using the new features!

odow:
I'll make a new release: https://github.com/jump-dev/JuMP.jl/pull/3117

PR #3107, opened 2022-10-14
[docs] remove incorrect statement about conflicting objectives

odow:
This statement is incorrect. Any calls to `@objective` or `@NLobjective` overwrite existing objective functions.


blegat:
I am not sure to agree, whenever writing a solver wrapper, I always give precedence to the NL objective function if any

odow:
> I always give precedence to the NL objective function if any

There's a test for that at the MOI level. 

At the JuMP level it's a different story. That sentence didn't reflect actual behavior. I think I just wrote it with MOI behavior in mind, without actually testing.

PR #3109, opened 2022-10-20
Add HermitianPSDCone

blegat:
This PR makes it possible to create Hermitian PSD matrix of variable.

The point of argument is going to be how to handle the keyword arguments of `@variable`. This PR does the following:
* If `binary` is true for any entry, throw an error.
* `fixed_value`, `start`, `lower_bound` and `upper_bound` should correspond to entries of hermitian matrices. We interpret the real (resp. imaginary) part as the corresponding value for the real (resp. imaginary) variable created. IIRC @mlubin was worried that it would be confusing, e.g., that `lower_bound = 1 + 2im` would be interpreted as a lower bound of `1` for the real part and a lower bound of `2` for the imaginary part. As a related note, @ccoffrin was asking for `@constraint(model, x + y * im .<= 1 + 2im)` to be interpreted as `@constraint(model, x <= 1)` and `@constraint(model, y <= 2)`.
* `integer` means that both the real and imaginary parts are `integer`. Maybe we should just error here ?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-MathOptInterface = "1.3.0"
+MathOptInterface = "1.7"
 MutableArithmetics = "1"
 OrderedCollections = "1"
 julia = "1.6"
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -445,7 +445,7 @@ struct HermitianPSDCone end
 
 Shape object for a hermitian square matrix of `side_dimension` rows and
 columns. The vectorized form corresponds to
-[`MathOptInterface.HermitianPositiveSemidefiniteConeTriangle`](@ref).
+[`MOI.HermitianPositiveSemidefiniteConeTriangle`](@ref).
 """
 struct HermitianMatrixShape <: AbstractShape
     side_dimension::Int
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -415,3 +415,141 @@ function build_constraint(
         shape,
     )
 end
+
+"""
+    HermitianPSDCone
+
+Hermitian positive semidefinite cone object that can be used to create a
+hermitian positive semidefinite square matrix in the [`@variable`](@ref)
+macro.
+
+## Examples
+
+Consider the following example:
+```jldoctest PSDCone; setup = :(using JuMP)
+julia> model = Model();
+
+julia> @variable(model, H[1:3, 1:3] in HermitianPSDCone())
+3×3 Matrix{GenericAffExpr{ComplexF64, VariableRef}}:
+ real(H[1,1])                                real(H[1,2]) + (0.0 + 1.0im) imag(H[1,2])   real(H[1,3]) + (0.0 + 1.0im) imag(H[1,3])
+ real(H[1,2]) + (-0.0 - 1.0im) imag(H[1,2])  real(H[2,2])                                real(H[2,3]) + (0.0 + 1.0im) imag(H[2,3])
+ real(H[1,3]) + (-0.0 - 1.0im) imag(H[1,3])  real(H[2,3]) + (-0.0 - 1.0im) imag(H[2,3])  real(H[3,3])
+```
+We see in the output of the last command that the matrix the vectorization of the
+matrix is constrained to belong to the `PositiveSemidefiniteConeSquare`.
+"""
+struct HermitianPSDCone end
+
+"""
+    SymmetricMatrixShape
+
+Shape object for a hermitian square matrix of `side_dimension` rows and
+columns. The vectorized form corresponds to
+[`MathOptInterface.HermitianPositiveSemidefiniteConeTriangle`](@ref).
+"""
+struct HermitianMatrixShape <: AbstractShape
+    side_dimension::Int
+end
+
+function vectorize(matrix::Matrix, ::HermitianMatrixShape)
+    n = LinearAlgebra.checksquare(matrix)
+    return vcat(
+        vectorize(_real.(matrix), SymmetricMatrixShape(n)),
+        vectorize(
+            _imag.(matrix[1:(end-1), 2:end]),
+            SymmetricMatrixShape(n - 1),
+        ),
+    )
+end
+
+function reshape_vector(v::Vector{T}, shape::HermitianMatrixShape) where {T}
+    NewType = _MA.promote_operation(_MA.add_mul, T, Complex{Bool}, T)
+    n = shape.side_dimension
+    matrix = Matrix{NewType}(undef, n, n)
+    real_k = 0
+    imag_k = MOI.dimension(MOI.PositiveSemidefiniteConeTriangle(n))
+    for j in 1:n
+        for i in 1:(j-1)
+            real_k += 1
+            imag_k += 1
+            matrix[i, j] = v[real_k] + im * v[imag_k]
+            matrix[j, i] = v[real_k] - im * v[imag_k]
+        end
+        real_k += 1
+        matrix[j, j] = v[real_k]
+    end
+    return matrix
+end
+
+function _mapinfo(f::Function, v::JuMP.ScalarVariable)
+    info = v.info
+    return ScalarVariable(
+        VariableInfo(
+            info.has_lb,
+            f(info.lower_bound),
+            info.has_ub,
+            f(info.upper_bound),
+            info.has_fix,
+            f(info.fixed_value),
+            info.has_start,
+            f(info.start),
+            info.binary,
+            info.integer,
+        ),
+    )
+end
+
+_real(s::String) = string("real(", s, ")")
+_imag(s::String) = string("imag(", s, ")")
+
+_real(v::ScalarVariable) = _mapinfo(real, v)
+_imag(v::ScalarVariable) = _mapinfo(imag, v)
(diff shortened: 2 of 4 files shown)


src/sd.jl
```diff
@@ -415,3 +415,141 @@ function build_constraint(
...
+julia> model = Model();
+
+julia> @variable(model, H[1:3, 1:3] in HermitianPSDCone())
+3×3 Matrix{GenericAffExpr{ComplexF64, VariableRef}}:
+ real(H[1,1])                                real(H[1,2]) + (0.0 + 1.0im) imag(H[1,2])   real(H[1,3]) + (0.0 + 1.0im) imag(H[1,3])
+ real(H[1,2]) + (-0.0 - 1.0im) imag(H[1,2])  real(H[2,2])                                real(H[2,3]) + (0.0 + 1.0im) imag(H[2,3])
+ real(H[1,3]) + (-0.0 - 1.0im) imag(H[1,3])  real(H[2,3]) + (-0.0 - 1.0im) imag(H[2,3])  real(H[3,3])
+```
+We see in the output of the last command that the matrix the vectorization of the
+matrix is constrained to belong to the `PositiveSemidefiniteConeSquare`.
```
odow, line 439: This sentence does't follow what I see. We should explain that `H` is an affine expression, but it is backed by a PSD matrix.

src/sd.jl
```diff
@@ -415,3 +415,141 @@ function build_constraint(
...
+"""
+struct HermitianPSDCone end
+
+"""
+    SymmetricMatrixShape
```
odow, line 444:
```suggestion
    HermitianMatrixShape
```

src/sd.jl
```diff
@@ -415,3 +415,141 @@ function build_constraint(
         shape,
     )
 end
+
+"""
+    HermitianPSDCone
+
+Hermitian positive semidefinite cone object that can be used to create a
+hermitian positive semidefinite square matrix in the [`@variable`](@ref)
```
odow, line 423:
```suggestion
Hermitian positive semidefinite square matrix in the [`@variable`](@ref)
```

I think all usages of `hermitian` should be capitalized?

src/sd.jl
```diff
@@ -415,3 +415,141 @@ function build_constraint(
...
+matrix is constrained to belong to the `PositiveSemidefiniteConeSquare`.
+"""
+struct HermitianPSDCone end
+
+"""
+    SymmetricMatrixShape
+
+Shape object for a hermitian square matrix of `side_dimension` rows and
+columns. The vectorized form corresponds to
+[`MathOptInterface.HermitianPositiveSemidefiniteConeTriangle`](@ref).
```
odow, line 448:
```suggestion
[`MOI.HermitianPositiveSemidefiniteConeTriangle`](@ref).
```

test/variable.jl
```diff
@@ -1077,6 +1077,80 @@ function test_Model_VariableIndex_VariableRef_fix_with_upper_bound(::Any, ::Any)
...
+    @test num_variables(model) == 4
+    v = all_variables(model)
+    _test_variable_name_util(v[1], "real(Q[1,1])")
+    @test Q[1, 1] == 1v[1]
+    _test_variable_name_util(v[2], "real(Q[1,2])")
+    _test_variable_name_util(v[4], "imag(Q[1,2])")
+    @test Q[1, 2] == v[2] + v[4] * im
+    _test_variable_name_util(v[3], "real(Q[2,2])")
+    @test Q[2, 2] == 1v[3]
+    @test Q[2, 1] == conj(Q[1, 2])
```
odow, line 1092:
```suggestion
    @test Q[2, 1] == conj(Q[1, 2])
    return
```

test/variable.jl
```diff
@@ -1077,6 +1077,80 @@ function test_Model_VariableIndex_VariableRef_fix_with_upper_bound(::Any, ::Any)
...
+    @test_throws ErrorException @variable(
+        model,
+        H[i = 1:2, j = 1:2] in HermitianPSDCone(),
+        integer = i > j
+    )
+    @test_throws ErrorException @variable(
+        model,
+        H[i = 1:2, j = 1:2] in HermitianPSDCone(),
+        Bool
+    )
```
odow, line 1121:
```suggestion
    )
    return
```

test/variable.jl
```diff
@@ -1077,6 +1077,80 @@ function test_Model_VariableIndex_VariableRef_fix_with_upper_bound(::Any, ::Any)
...
+        @test value(upper_bound, H[i, j]) == (i * j) + (j - i) * im
+    end
+    @variable(
+        model,
+        T[i = 1:2, j = 1:2] in HermitianPSDCone(),
+        start = (i + j) + (j - i) * im
+    )
+    for i in 1:2, j in 1:2
+        @test value(start_value, T[i, j]) == (i + j) + (j - i) * im
+    end
```
odow, line 1151:
```suggestion
    end
    return
```

test/variable.jl
```diff
@@ -1077,6 +1077,80 @@ function test_Model_VariableIndex_VariableRef_fix_with_upper_bound(::Any, ::Any)
...
+    )
+    @test_throws ErrorException @variable(
+        model,
+        H[i = 1:2, j = 1:2] in HermitianPSDCone(),
+        integer = i > j
+    )
+    @test_throws ErrorException @variable(
+        model,
+        H[i = 1:2, j = 1:2] in HermitianPSDCone(),
+        Bool
```
odow, line 1120:
```suggestion
        Bin
```

Not sure what `Bool` was meant to do here?

blegat: Oops

src/sd.jl
blegat: Indeed, you are right

ccoffrin:
`HermitianPSDCone` is natural for use in the SDP relaxation of AC-OPF.  Shall we test that model?

blegat:
Is that the model that was tested with ComplexOptInterface ?

ccoffrin:
Almost, that model was a non-convex QCQP.  To make it an SDP you need to lift the voltage variable products into a matrix put PSD on that.

odow (approved):
Nice enough for now. Users might be a bit confused with the `real(H[1, 1])` name. It looks like a function call, not the string name. Potentially `real_H[1, 1]` and `imag_H[1, 1]`? Or `re_H[1, 1]` and `im_H[1, 1]`?

blegat:
The advantage with this name is that it corresponds exactly to what you would get by typing `real(H[1, 1])`. Currently, you can get the value of the function by typing what is printed (unless you use anonymous variables).

odow:
As discussed on the monthly meeting, we should add a page to the manual in the docs explaining all quirks of the complex support in JuMP. 

 * Behavior of complex bounds
 * How to access real and imaginary parts of `H`
 * Why the naming is what it is
 * How to access things like `value(lower_bound, H[1, 2])` instead of `lower_bound(H[1, 2])`

Potentially also deal with complex equality constraints, but that might be a separate PR, or in the AC-OPF case.

src/sd.jl
```diff
@@ -415,3 +415,158 @@ function build_constraint(
...
+            info.has_start,
+            f(info.start),
+            info.binary,
+            info.integer,
+        ),
+    )
+end
+
+_real(s::String) = string("real(", s, ")")
+_imag(s::String) = string("imag(", s, ")")
```
odow, line 520: If the variable is anonymous, then the names will become `real()` and `imag()`. I'm not sure if `_[1]` or `real(_[1])` is better.

Project.toml
```diff
@@ -12,7 +12,7 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-MathOptInterface = "1.3.0"
+MathOptInterface = "1.7"
```
odow, line 15: @blegat we need to remember to update this when we rely on new MOI features

test/variable.jl
```diff
@@ -1077,6 +1077,93 @@ function test_Model_VariableIndex_VariableRef_fix_with_upper_bound(::Any, ::Any)
...
+    return
+end
+
+function test_Hermitian_PSD_anon(::Any, ::Any)
+    model = Model()
+    x = @variable(model, [1:2, 1:2] in HermitianPSDCone())
+    @test sprint(show, x[1, 1]) ==  "_[1]"
+    @test sprint(show, x[1, 2]) ==  "_[2] + (0.0 + 1.0im) _[4]"
+    @test sprint(show, x[2, 1]) ==  "_[2] + (-0.0 - 1.0im) _[4]"
+    @test sprint(show, x[2, 2]) ==  "_[3]"
```
odow, line 1163: I updated the printing. How's this?

odow (approved):
Probably easiest if we leave the docs for a separate PR.

odow:
> The advantage with this name is that it corresponds exactly to what you would get by typing real(H[1, 1])

This is also not quite right, because it returns an affine expression, not the variable.

Moreover
```julia
julia> imag(H[1, 2])
0 real(H[1,2]) + imag(H[1,2])

julia> real(H[1, 2])
real(H[1,2]) + 0 imag(H[1,2])
```

test/variable.jl
blegat: Thanks, that's better

Project.toml
blegat: Indeed, good catch

blegat:
It now throws errors in case the `integer` option is used as discussed in the jump-dev call. I'll merge tomorrow unless there is any more objection

odow (approved):

Issue #3034, opened 2022-08-03
Debuggability of JuMP models


odow:
One area where I think we can do much more to help users is to provide tooling and documentation to help them debug and test the optimization models that they build.

I had this as one of the items in https://github.com/jump-dev/JuMP.jl/issues/2348, but I'm lifting it to a stand-alone issue.

I think what this needs is a new tutorial that goes over some common issues, and provides links to other resources.

If anyone has ideas/links, please post them below.

## Motivation

@mtanneau's [JuMP-dev talk](https://www.youtube.com/watch?v=61fnHmri7_Q
) discussed a lot of this, but it keeps coming up. I see a lot of common issues on Discourse where people are missing some strategies to debug things.

## Prior art

It's also a topic that is surprisingly lacking in other modeling libraries.
  * GAMS has a nice section: https://www.gams.com/latest/docs/UG_ExecErrPerformance.html. It's probably the most complete I could find.
  * Pyomo doesn't have much https://pyomo.readthedocs.io/en/stable/model_debugging/index.html
  * AMPL has a short section in Chapter 1 on "Catching errors." But it basically advises people to add `check`s to their code to validate data, and that "you may have to spend some time reviewing your formulation before you discover what is wrong." The tooling suggestion is to print out a constraint or objective to look at the data. But that doesn't scale to large problems very well.
  * Gurobi has `feasrelax` and `IIS` support: https://www.gurobi.com/documentation/9.5/examples/diagnose_and_cope_with_inf.html
  * Localsolver has a short section on debugging: https://www.localsolver.com/docs/last/guidelines/debuggingamodel.html

I found this excellent paper that should be required reading for anyone who encounters a bug modelling:

* Pannell, Kingwell, and Schilizzi (1996). Debugging mathematical programming models: principles and practical strategies. _Review of Marketing and Agricultural Economics_, 64(1), 86--100. [pdf](https://github.com/jump-dev/JuMP.jl/files/9255639/64010086.pdf). 

It does have one great paragraph: "Just like science, debugging cannot be a cold, calculating and linear process. Both involve essential elements of inspired guesswork, hutches and sudden flashes of insight which cut through the mist."

They lay out a good framework for debugging as a table/flow-chart. We could aim to replicate that.

Other links

 * https://medium.com/bcggamma/five-strategies-for-debugging-mip-models-4f715809f9f0
[...]

PR #3113, opened 2022-10-24
Added `relax_constraint`, similar to `relax_integrality`

this-josh:
I've added a method to relax the specified constraints, here is how it works:
```Julia
    model = Model()
    @variable(model, x)
    @constraint(model, c1, x≤1)
    @test is_valid(model, c1)
    unrelax = relax_constraint(model, c1)
    @test !is_valid(model, c1)
    unrelax()
    @test is_valid(model, model[:c1])
```

Currently I haven't documented it, if you're interesting in merging this I can add some docs and more thorough testing (e.g. with Non linear constraints).


odow:
Hi @this-josh, a good first step, before opening a PR, is to open an issue to discuss a proposed change (see step 1: https://jump.dev/JuMP.jl/stable/developers/contributing/#Contribute-code-to-JuMP).

What is the motivation for this? Why is something like `delete` not sufficient?

The `unregister` logic is a bit tricky to get right. It won't work for something like 
```julia
model = Model()
@variable(model, x)
@constraint(model, c[i=1:2], x <= i)
@test is_valid(model, c[1])
unrelax = relax_constraint(model, c[1])
@test !is_valid(model, c[1])
unrelax()
@test is_valid(model, model[:c][1])
```

Instead of deleting the constraint, and then re-adding, you might be interested in relaxing the feasibility of the problem: https://github.com/jump-dev/MathOptInterface.jl/pull/1995 and https://github.com/jump-dev/JuMP.jl/issues/3034.

this-josh:
HI, I appreciate that but in this case I had to write the code for my own purposes. With the code written it seemed clearer to open this PR and combine discussion into the PR. In hindsight I could have just included the code snippet in an Issue. 

This was useful for me when doing a feasibility pump. I have some non-linear constraints, and my own linear relaxation. Using this I can selectively choose between them. 

You're right, I would prefer to avoid deleting constraints. After looking at your links I expect [#1995](https://github.com/jump-dev/MathOptInterface.jl/pull/1995) should cover this in a more elegant way.

PR #3118, opened 2022-10-29
Add set_time_limit for setting time limits with Dates.Period

### LINKED ISSUE #3115: set_time_limit_sec: take Dates.Period as argument?
Currently, `set_time_limit_sec` only takes a real number of seconds as arguments. What about taking the Julia standard for durations, `Dates.Period` (standard library)? The only goal would be to make this function easier to use from Julia if you already use standard-library features for time.

Its implementation could look like this: 

```
function set_time_limit_sec(model, time::Dates.Period)
  set_time_limit_sec(model, floor(rd.timeout, Nanosecond).value * 1.0e-9)
  return
end
```

odow:
Closes #3115

I'm not convinced that it should be `set_time_limit` instead of `set_time_limit_sec`. I get that semantically it's a little clearer, but it introduces a new function, and stops authors from (easily) writing generic code that allows users to pass either a `Float64` or a `Dates.Period` object.

cc @dourouc05

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ repo = "https://github.com/jump-dev/JuMP.jl.git"
 version = "1.4.0"
 
 [deps]
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 MutableArithmetics = "d8a4904e-b15c-11e9-3269-09a3773c0cb0"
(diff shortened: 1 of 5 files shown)


dourouc05 (approved):

dourouc05:
I have no strong opinion on the name, for the exact same reasons you mentioned. What might be useful would be to have a `time_limit` (without suffix) to get the time limit as a `Period` object, but that would make one more symbol exported from JuMP (and risk conflicts with other packages).

odow:
> Given this incompatibility, I'm worried we're just going to create more potential for confusion by starting to accept Periods. Maybe we should leave it at documenting how to convert a Dates.Period object to seconds for set_time_limit_sec?

This sounds okay. It's really just a one-liner.

odow:
Here's a PR with just the documentation change: https://github.com/jump-dev/JuMP.jl/pull/3121

PR #3121, opened 2022-10-30
[docs] explain Dates.period in set_time_limit_sec

odow:
Closes https://github.com/jump-dev/JuMP.jl/pull/3118
Closes https://github.com/jump-dev/JuMP.jl/pull/3115

A simpler alternative to #3118

### DIFF
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -253,7 +253,7 @@ consider converting a `Dates.Period` object to seconds:
 ```jldoctest time_limit
 julia> import Dates
 
-julia> seconds(x::Dates.Period) = Dates.value(round(x, Dates.Second))
+julia> seconds(x::Dates.Period) = 1e-3 * Dates.value(round(x, Dates.Millisecond))
 seconds (generic function with 1 method)
 
 julia> set_time_limit_sec(model, seconds(Dates.Hour(1)))
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -232,7 +232,7 @@ julia> unset_silent(model)
 
 Use [`set_time_limit_sec`](@ref), [`unset_time_limit_sec`](@ref), and
 [`time_limit_sec`](@ref) to manage time limits.
-```jldoctest
+```jldoctest time_limit
 julia> model = Model(HiGHS.Optimizer);
 
 julia> set_time_limit_sec(model, 60.0)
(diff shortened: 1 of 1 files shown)


docs/src/manual/models.md
```diff
@@ -247,6 +247,21 @@ julia> time_limit_sec(model)
 Inf
 ```
 
+To avoid making unit-conversion mistakes with [`set_time_limit_sec`](@ref),
+consider converting a `Dates.Period` object to seconds:
+
+```jldoctest time_limit
+julia> import Dates
+
+julia> seconds(x::Dates.Period) = 1e-9 * Dates.value(floor(x, Dates.Nanosecond))
```
dourouc05, line 256: Instead of using a magic constant (which is a reason why I brought the topic to the table), you could do something like `convert(Second, Hour(72))` or `Second(Hour(72))`. But it may have problems if the number is not an integer number of seconds.

```julia
julia> convert(Dates.Minute, Dates.Millisecond(75631000))
ERROR: InexactError: divexact(Int64, 1260.5166666666667)
Stacktrace:
 [1] divexact
   @ C:\Users\Thibaut\AppData\Local\Programs\Julia-1.8.2\share\julia\stdlib\v1.8\Dates\src\periods.jl:414 [inlined]
 [2] convert(#unused#::Type{Minute}, x::Millisecond)
   @ Dates C:\Users\Thibaut\AppData\Local\Programs\Julia-1.8.2\share\julia\stdlib\v1.8\Dates\src\periods.jl:441
 [3] top-level scope
   @ REPL[61]:1
```

Does it make sense to support sub-second resolution? Are solvers precise enough for this?

odow: I guess you could use `seconds(x::Dates.Period) = Dates.value(ceil(x, Dates.Second))`

mlubin: > Does it make sense to support sub-second resolution? Are solvers precise enough for this?

Absolutely, we shouldn't arbitrarily round to seconds. I'd guess milliseconds would be fine if we don't want to do nanoseconds. We regularly get questions about solving problems in a millisecond time frame.

odow: Updated. I've never seen anyone measure a time in in microseconds.

docs/src/manual/models.md
```diff
@@ -247,6 +247,21 @@ julia> time_limit_sec(model)
 Inf
 ```
 
+To avoid making unit-conversion mistakes with [`set_time_limit_sec`](@ref),
```
mlubin, line 250: I'd frame it more positively, e.g., if your time limit is encoded as a `Dates.Period` object, you can use the following code to convert it to `Float64` for `set_time_limit_sec`.

mlubin (approved):

PR #3123, opened 2022-11-01
Fix error message for vectorized interval constraints

odow:
Raised on Discourse: https://discourse.julialang.org/t/jump-quadratic-problem/89475

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -617,6 +617,18 @@ end
 
 function build_constraint(
     _error::Function,
+    ::AbstractArray,
+    ::AbstractVector,
+    ::AbstractVector
+)
+    return _error(
+        "Unexpected vectors in scalar constraint. Did you mean to use the dot ",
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -615,6 +615,18 @@ function build_constraint(
     )
 end
 
+function build_constraint(
+    _error::Function,
+    ::AbstractArray,
+    ::AbstractVector,
+    ::AbstractVector
```
odow, line 622:
```suggestion
    ::AbstractVector,
```

src/macros.jl
```diff
@@ -731,7 +743,6 @@ function _constraint_macro(
     source::LineNumberNode,
 )
     _error(str...) = _macro_error(macro_name, args, source, str...)
```
odow, line 745:
```suggestion
    _error(str...) = _macro_error(macro_name, args, source, str...)

```

PR #3106, opened 2022-10-10
[Nonlinear] add new nonlinear interface

odow:
## Documentation

https://jump.dev/JuMP.jl/previews/PR3106/manual/nonlinear/

## About

This PR is new implementation of nonlinear programming for JuMP.

It introduces a new `GenericNonlinearExpr` data structure, which replicates the structure of basic Julia Expr objects. The motivation for this is that we already create these in the macros, and it seems to work okay, so why not make it official!

## Demos

Cribbing @pulsipher's demos in https://github.com/infiniteopt/InfiniteOpt.jl/pull/161, we have:

### Operator overloading

Nonlinear expressions can be defined outside of macros like their quad/affine
counterparts:

```julia
julia> using JuMP; m = Model(); @variable(m, x);

julia> x^2.3 + sin(x)
(x ^ 2.3) + sin(x)
```

### Macro definitions

By making the necessary extensions to MutableArithmetics, we can define
nonlinear expressions inside of `@expression`, `@objective`, and `@constraint` (i.e.,
no need for the special NL macros):

```julia
julia> @expression(m, x + x^3 * sin(x)^x)
x + ((x ^ 3) * (sin(x) ^ x))
```

### LinearAlgebra support

We can also do linear algebra operations:

```julia
julia> @variable(m, X[1:3, 1:2]); @variable(m, v[1:3]); @variable(m, w[1:2]);

julia> @expression(m, v' * X * w)
0.0 + ((X[1,1]*v[1] + X[2,1]*v[2] + X[3,1]*v[3]) * w[1]) + ((X[1,2]*v[1] + X[2,2]*v[2] + X[3,2]*v[3]) * w[2])
```

### Function tracing

Because we support operator definition, we can trace functions in like manner to
Symbolics.jl:

```julia
julia> my_func(y) = 2^y + exp(y^-2.3);

julia> my_func(x)
(2.0 ^ x) + exp(x ^ -2.3)
```

### User-defined operators

For user-defined operators that are not compatible with tracing, we allow them
to be registered like in JuMP and Symbolics.jl.

```julia
julia> my_func2(a) = sin(a) * eta(a);

julia> @operator(m, op_my_func, 1, my_func2)
NonlinearOperator(:op_my_func, my_func2)

julia> @expression(m, op_my_func(x)^2)
op_my_func(x) ^ 2.0
```

You can also implement this by defining a new overload:
```julia
julia> add_nonlinear_operator(m, 1, my_func2; name = :op_my_func2)
[...]

PR #3125, opened 2022-11-03
Change MutableArithmetics.rewrite to move_factors_into_sums=false

odow:
Needs https://github.com/jump-dev/MutableArithmetics.jl/pull/170

I've been experimenting with a drop-in replacement for `MA.@rewrite` and `MA.rewrite`. JuMP throws up some interesting challenges, so it took me a while to get this correct. The tests are now passing with a couple of minimal changes.

1. A change from compile-time to run-time error for `@expression(model, x <= 1)`
2. A correction of the error in `@constraint(m, x[i = 1] <= 1)` so that it matches the non-macro error obtained from `x[i = 1]`
3. A printing change from `y * x` to `x * y`. This is just cosmetic, and depends on the order that we encounter the variables, but the new behavior now more closely resembles what the user wrote.

The biggest surprises were some of our design choices:

 * Rewriting `a + b` to `operate!!(add_mul, a, b)` instead of `operate!!(+, a, b)` 
 * Rewriting `a - b` to `operate!!(sub_mul, a, b)` instead of `operate!!(-, a, b)`
 * We rewrite only `+` and `-` to their mutable `add_mul` and `sub_mul` equivalents. We don't rewrite something like `operate!!(/, a, b)`, even if the operation is supported.
 * Supporting the invalid syntax `sum(i+j for i=1:2, j=i:2)`:
   ```julia
   julia> sum(i+j for i=1:2, j=i:2)
   ERROR: UndefVarError: i not defined

   julia> JuMP.MutableArithmetics2.@rewrite(sum(i+j for i=1:2, j=i:2))
   9
   ```
   The last one took me a while to debug...

There are a couple of options for merging this.

 1. <s>keep it in JuMP</s>
 2. Add it to MutableArithmetics as a replacement for the existing code
 3. Add it to MutableArithmetics as a new feature, perhaps renaming to `rewrite_expr`?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,7 +14,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
 MathOptInterface = "1.14"
-MutableArithmetics = "1"
+MutableArithmetics = "1.1"
 OrderedCollections = "1"
 SnoopPrecompile = "1"
 julia = "1.6"
--- a/docs/src/manual/expressions.md
+++ b/docs/src/manual/expressions.md
@@ -195,7 +195,7 @@ julia> @variable(model, y)
 y
 
 julia> ex = @expression(model, x^2 + 2 * x * y + y^2 + x + y - 1)
-x² + 2 y*x + y² + x + y - 1
+x² + 2 x*y + y² + x + y - 1
 ```
 
 ### Operator overloading
--- a/src/complement.jl
+++ b/src/complement.jl
@@ -70,6 +70,6 @@ function parse_constraint_call(
     F,
     x,
 )
-    f, parse_code = _MA.rewrite(F)
+    f, parse_code = _MA.rewrite(F; move_factors_into_sums = false)
     return parse_code, :(_build_complements_constraint($errorf, $f, $(esc(x))))
 end
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -500,9 +500,9 @@ function parse_constraint_head(
             "`$ub >= ... >= $lb`.",
         )
     end
-    new_aff, parse_aff = _MA.rewrite(aff)
-    new_lb, parse_lb = _MA.rewrite(lb)
-    new_ub, parse_ub = _MA.rewrite(ub)
+    new_aff, parse_aff = _MA.rewrite(aff; move_factors_into_sums = false)
+    new_lb, parse_lb = _MA.rewrite(lb; move_factors_into_sums = false)
+    new_ub, parse_ub = _MA.rewrite(ub; move_factors_into_sums = false)
     parse_code = quote
         $parse_aff
         $parse_lb
@@ -583,7 +583,7 @@ function parse_constraint_call(
     func,
     set,
 )
-    f, parse_code = _MA.rewrite(func)
+    f, parse_code = _MA.rewrite(func; move_factors_into_sums = false)
     build_call = if vectorized
         :(build_constraint.($_error, _desparsify($f), Ref($(esc(set)))))
     else
@@ -617,7 +617,7 @@ function parse_constraint_call(
     rhs,
 )
     func = vectorized ? :($lhs .- $rhs) : :($lhs - $rhs)
-    f, parse_code = _MA.rewrite(func)
+    f, parse_code = _MA.rewrite(func; move_factors_into_sums = false)
     set = operator_to_set(_error, operator)
     # `_functionize` deals with the pathological case where the `lhs` is a
     # `VariableRef` and the `rhs` is a summation with no terms.
@@ -1496,7 +1496,7 @@ macro objective(model, args...)
     end
     sense, x = args
     sense_expr = _moi_sense(_error, sense)
-    newaff, parsecode = _MA.rewrite(x)
+    newaff, parsecode = _MA.rewrite(x; move_factors_into_sums = false)
     code = quote
         $parsecode
         # Don't leak a `_MA.Zero` if the objective expression is an empty
@@ -1585,7 +1585,7 @@ macro expression(args...)
             "different name for the index.",
         )
     end
-    code = _MA.rewrite_and_return(x)
+    code = _MA.rewrite_and_return(x; move_factors_into_sums = false)
     code = quote
         # Don't leak a `_MA.Zero` if the expression is an empty summation, or
         # other structure that returns `_MA.Zero()`.
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -603,7 +603,7 @@ end
 function test_Error_on_unexpected_comparison()
     m = Model()
     @variable(m, x)
-    @test_macro_throws ErrorException @expression(m, x <= 1)
+    @test_throws ErrorException @expression(m, x <= 1)
     return
 end
 
--- a/test/test_print.jl
+++ b/test/test_print.jl
@@ -180,7 +180,7 @@ function test_printing_expressions()
         "x_{1}\\times y_{2,2} + x_{2}\\times y_{2,2} + z$ijulia_sq + 3 x_{1} + 3 x_{2} - 1",
     )
 
-    ex = @expression(mod, -z * x[1] - x[1] * z + x[1] * x[2] + 0 * z^2)
+    ex = @expression(mod, -z * x[1] - z * x[1] + x[1] * x[2] + 0 * z^2)
     io_test(MIME("text/plain"), ex, "-2 z*x[1] + x[1]*x[2]")
     io_test(MIME("text/latex"), ex, "-2 z\\times x_{1} + x_{1}\\times x_{2}")
 


odow:
Urgh. We're back to the `Union` issue in MA, this time in another place.
```julia
julia> import MutableArithmetics

julia> const MA = MutableArithmetics
MutableArithmetics

julia> MA.operate(*, [1, 2]', Vector{Union{String,Float64}}([1.0, 2.0]))
5.0

julia> MA.operate(*, [1 2; 3 4], Vector{Union{String,Float64}}([1.0, 2.0]))
ERROR: MethodError: no method matching zero(::Type{Union{Float64, String}})
Closest candidates are:
  zero(::Union{Type{P}, P}) where P<:Dates.Period at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Dates/src/periods.jl:53
  zero(::MutableArithmetics.AbstractMutable) at /Users/oscar/.julia/dev/MutableArithmetics/src/dispatch.jl:784
  zero(::SparseArrays.AbstractSparseArray) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/SparseArrays/src/SparseArrays.jl:55
  ...
Stacktrace:
 [1] _instantiate_zero(#unused#::Type{Union{Float64, String}})
   @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:28
 [2] promote_operation_fallback(op::typeof(*), #unused#::Type{Int64}, #unused#::Type{Union{Float64, String}})
   @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:51
 [3] promote_operation(::typeof(*), ::Type, ::Type)
   @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:113
 [4] promote_sum_mul(T::Type, S::Type)
   @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/implementations/LinearAlgebra.jl:186
 [5] promote_array_mul(#unused#::Type{Matrix{Int64}}, #unused#::Type{Vector{Union{Float64, String}}})
   @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/implementations/LinearAlgebra.jl:191
 [6] operate(#unused#::typeof(*), A::Matrix{Int64}, B::Vector{Union{Float64, String}})
   @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/implementations/LinearAlgebra.jl:388
 [7] top-level scope
   @ REPL[5]:1
```

.github/workflows/aqua.yml
```diff
@@ -17,6 +17,7 @@ jobs:
         shell: julia --color=yes {0}
         run: |
           using Pkg
+          Pkg.pkg"add MutableArithmetics#od/rewrite-2"
```
odow, line 20:
```suggestion
          Pkg.pkg"add MutableArithmetics#od/master"
```

test/runtests.jl
```diff
@@ -10,6 +10,9 @@
 # test/runtests.jl
 #############################################################################
 
+import Pkg
+Pkg.pkg"add MutableArithmetics#od/rewrite-2"
```
odow, line 14:
```suggestion
Pkg.pkg"add MutableArithmetics#od/master"
```

docs/make.jl
```diff
@@ -1,3 +1,6 @@
+import Pkg
+Pkg.pkg"add MutableArithmetics#od/rewrite-2"
```
odow, line 2:
```suggestion
Pkg.pkg"add MutableArithmetics#master"
```

odow:
Run the larger JuMP benchmark suite

odow:
I ran the benchmarks here: https://github.com/jump-dev/JuMP.jl/blob/master/test/perf/JuMPBenchmarks.jl. 

No major changes. 

 * The linear examples use a lot more memory, but are 30% faster.
 * The quadratic summations are slightly slower.
 * Everything else is invariant.

The linear examples use more memory because we're no longer collapsing sums into each other. If you write
```Julia
model = Model()
@variable(model, x[1:3])
@expression(model, sum(x[i] for i in 1:3) + sum(x[i] for i in 1:3))
```
the old MA produced 
```Julia
quote
    var"#11###286" = (JuMP.MutableArithmetics).Zero()
    for i = 1:3
        var"#11###286" = (JuMP.MutableArithmetics).operate!!((JuMP.MutableArithmetics).add_mul, var"#11###286", x[i])
    end
    var"#12###285" = var"#11###286"
    for i = 1:3
        var"#12###285" = (JuMP.MutableArithmetics).operate!!((JuMP.MutableArithmetics).add_mul, var"#12###285", x[i])
    end
    var"#13###284" = var"#12###285"
end
```
whereas now we do
```julia
quote
    var"#11###284" = MutableArithmetics.Zero()
    for i = 1:3
        var"#11###284" = (MutableArithmetics.operate!!)(MutableArithmetics.add_mul, var"#11###284", x[i])
    end
    var"#12###285" = MutableArithmetics.Zero()
    for i = 1:3
        var"#12###285" = (MutableArithmetics.operate!!)(MutableArithmetics.add_mul, var"#12###285", x[i])
    end
    var"#13###286" = (MutableArithmetics.operate!!)(MutableArithmetics.add_mul, var"#11###284", var"#12###285")
end
```
The new rewrite follows the syntax (there are two sums), at the cost of creating an additional `AffExpr` object.
The same thing happens if you have nested sums like `sum(j * sum(x[i] for i in 1:N) for j in 1:N)` (which is what the benchmarks are testing).

Here's the output of the benchmarks:

```Julia
(base) oscar@Oscars-MBP JuMP % git checkout master
Switched to branch 'master'
Your branch is up to date with 'master'.
(base) oscar@Oscars-MBP JuMP % julia --project=. test/perf/JuMPBenchmarks.jl -f expr.txt
... lines omitted ...
(base) oscar@Oscars-MBP JuMP % git checkout od/ma2
Switched to branch 'od/ma2'
Your branch is up to date with 'origin/od/ma2'.
(base) oscar@Oscars-MBP JuMP % julia --project=. test/perf/JuMPBenchmarks.jl -f expr.txt --compare

========== Results ==========

bench_macro_linear_010
BenchmarkTools.TrialJudgement: 
  time:   -31.57% => improvement (5.00% tolerance)
  memory: +191.15% => regression (1.00% tolerance)

bench_dense_axis_constraints_index
BenchmarkTools.TrialJudgement: 
  time:   -0.68% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_array_sum_050
BenchmarkTools.TrialJudgement: 
  time:   -2.69% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_vector_sum_050
BenchmarkTools.TrialJudgement: 
  time:   -0.01% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_sparse_axis_constraints_iterate
BenchmarkTools.TrialJudgement: 
  time:   -4.00% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_matrix_quadratic_product_010
BenchmarkTools.TrialJudgement: 
  time:   -1.49% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_matrix_affine_product_010
BenchmarkTools.TrialJudgement: 
  time:   -0.45% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_sparse_axis_constraints_index
BenchmarkTools.TrialJudgement: 
  time:   +2.37% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_vector_sum_010
BenchmarkTools.TrialJudgement: 
  time:   -0.43% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_dense_axis_constraints_iterate
BenchmarkTools.TrialJudgement: 
  time:   -0.04% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_matrix_sum_050
BenchmarkTools.TrialJudgement: 
  time:   +4.17% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_macro_quad_050
BenchmarkTools.TrialJudgement: 
  time:   +12.98% => regression (5.00% tolerance)
  memory: -15.85% => improvement (1.00% tolerance)

bench_matrix_affine_product_050
BenchmarkTools.TrialJudgement: 
  time:   -1.21% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_matrix_quadratic_product_050
BenchmarkTools.TrialJudgement: 
  time:   +0.96% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_macro_linear_050
BenchmarkTools.TrialJudgement: 
  time:   -39.13% => improvement (5.00% tolerance)
  memory: +341.27% => regression (1.00% tolerance)

bench_array_sum_010
BenchmarkTools.TrialJudgement: 
  time:   -2.52% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_matrix_sum_010
BenchmarkTools.TrialJudgement: 
  time:   -1.35% => invariant (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)

bench_macro_quad_010
BenchmarkTools.TrialJudgement: 
  time:   +5.64% => regression (5.00% tolerance)
  memory: +7.79% => regression (1.00% tolerance)
```

odow:
Running `extension-tests` on this: https://github.com/jump-dev/JuMP.jl/actions/runs/5216388377

Now that I've used the nonlinear branch a bunch, I'm pretty confident that this change is both safe to do. I'd like to get it in separately to the nonlinear PR so we're not changing too much in a single go.

odow:
@pulsipher this breaks the tests of InfininteOpt: https://github.com/jump-dev/JuMP.jl/actions/runs/5216388377/jobs/9415059599

image

It's the same set of changes here:

https://github.com/infiniteopt/InfiniteOpt.jl/pull/308/files#diff-6f28d0077bbc402cee4e236a0ca51e04d420de58e2a9c5897868ece68f2de188L82

We now build expressions that match what you type, instead of these somewhat arbitrary `x / y --> x * (1 / y)` reformulations.

Thoughts on merging this separately to the nonlinear stuff?

odow:
This also seemed to fix a series of test failures I'd seen on Alpine, likely because we're passing more sensible formulations to the solver.

pulsipher:
> @pulsipher this breaks the tests of InfininteOpt: https://github.com/jump-dev/JuMP.jl/actions/runs/5216388377/jobs/9415059599
> 
> image
> It's the same set of changes here:
> 
> https://github.com/infiniteopt/InfiniteOpt.jl/pull/308/files#diff-6f28d0077bbc402cee4e236a0ca51e04d420de58e2a9c5897868ece68f2de188L82
> 
> We now build expressions that match what you type, instead of these somewhat arbitrary `x / y --> x * (1 / y)` reformulations.
> 
> Thoughts on merging this separately to the nonlinear stuff?

I am definitely in favor of this change to MutableArithmetics. Naturally, it would be easiest for me to have this merged at about the same time as #3106, but it shouldn't be that much work to breakup my existing pull request in InfiniteOpt accordingly. So, I guess in short I have no problem with this being merged.

odow:
If it didn't break your tests I'd merge this ASAP, but perhaps it's okay as a PR immediately before #3106, so that they both come out in the same JuMP release.

odow:
I guess the other point to readers is that this isn't a breaking change per se. It just breaks the tests in which @pulsipher compares two expressions; the expressions are equivalent mathematically, but they have a different structure in code.

PR #3129, opened 2022-11-14
Add support for get/set of attributes in optimizer_with_attributes

### LINKED ISSUE #3128: Setting Optimizer Attributes on an OptimizerWithAttributes
Thought,  should `set_optimizer_attribiute` be defined for objects of type `OptimizerWithAttributes`, similarly to how it works for the type `Model`?

For example,
```
solver = JuMP.optimizer_with_attributes(Ipopt.Optimizer)
if <some condition>
    JuMP.set_optimizer_attribiute(solver, "tol", 1e-6)
end
```

my current work around, which is not too bad,
```
options = Dict{String,Any}()
if <some condition>
    options["tol"] = 1e-6
end
solver = JuMP.optimizer_with_attributes(Ipopt.Optimizer, options...)
```

odow:
Closes #3128 

There's also https://github.com/jump-dev/MathOptInterface.jl/pull/2049 on the MOI side, so we could either wait for a new tag, or merge this PR now and update it later once MOI is released.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-MathOptInterface = "1.7"
+MathOptInterface = "1.10"
 MutableArithmetics = "1"
 OrderedCollections = "1"
 julia = "1.6"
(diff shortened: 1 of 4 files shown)


odow:
https://github.com/jump-dev/MathOptInterface.jl/pull/2049 is merged, so I guess we're happy to add this to JuMP.

blegat:
Maybe we can wait for an MOI tag, this isn't the most pressing feature

PR #3134, opened 2022-11-23
Add support for creation of complex variables

blegat:
This PR makes it possible to create complex variables. The main point of arguments were discussed part of https://github.com/jump-dev/JuMP.jl/pull/3109 so this PR should be more straightforward.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -11,7 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


odow (approved):

odow:
Merge, and then work on some comprehensive docs for `Complex`? Then the NumFOCUS small development grant is doooone.

blegat:
Yes, we still needs docs but also PSD constraints in `@constraint` as requested in https://github.com/jump-dev/ComplexOptInterface.jl/issues/17

PR #3138, opened 2022-11-30
Report result_count in solution_summary by default

### LINKED ISSUE #3136: solution_summary should report multiple solutions
It'd be helpful if `solution_summary` gave the number of solutions, as well as their statuses. This would have helped diagnose https://discourse.julialang.org/t/mosek-does-not-give-the-optimal-value-derived-by-simplex-methods/90727

odow:
Closes #3136

The backstory to this is that https://discourse.julialang.org/t/mosek-does-not-give-the-optimal-value-derived-by-simplex-methods/90727 would have been a lot more obvious to debug if `result_count` was reported by default in `solution_summary`.

I'm open to other suggestions about how we could communicate this. The `Solution #1 (default)` comes from #3136, but it doesn't really apply when `result_count = 0`. Should we even report primal/dual status then?

### DIFF
--- a/test/solution_summary.jl
+++ b/test/solution_summary.jl
@@ -26,8 +26,10 @@ function test_empty_model()
 
 * Status
   Termination status : OPTIMIZE_NOT_CALLED
-  Primal status      : NO_SOLUTION
-  Dual status        : NO_SOLUTION
+  Result count       : 0
+  Solution #1 (default)
+    Primal status    : NO_SOLUTION
+    Dual status      : NO_SOLUTION
   Message from the solver:
   "optimize not called"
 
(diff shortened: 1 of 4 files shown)


odow:
@blegat suggests `solution_summary(model; result::Int = 1)`

odow:
How's this? https://jump.dev/JuMP.jl/previews/PR3138/manual/solutions/#Solutions-summary

test/solution_summary.jl
```diff
@@ -25,6 +25,7 @@ function test_empty_model()
 * Solver : No optimizer attached.
 
 * Status
+  Result             : 1 of 0
```
mlubin, line 28: 1 of 0?

odow: By default, `result = 1`, and the `result_count` is `0`, so that's literally what your asking for with something like `primal_status(model)`. 

I guess we could make this `0 of 0`?

mlubin: Can we move `Primal status` and `Dual status` under `* Candidate solution`?
E.g.,
```
* Status
Result count: X
Termination status: ...

* Candidate solution (result #Y)
Primal status: ...
Dual status: ...
```

odow: Oh yeah, that's a better idea.

odow: Done

PR #3140, opened 2022-12-02
Add relax_with_penalty

### LINKED ISSUE #3034: Debuggability of JuMP models

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/pull/1995
Closes #3034 

Preview: https://jump.dev/JuMP.jl/previews/PR3140/tutorials/getting_started/debugging/

### DIFF
--- a/docs/src/tutorials/getting_started/debugging.jl
+++ b/docs/src/tutorials/getting_started/debugging.jl
@@ -168,7 +168,7 @@ model = Model(HiGHS.Optimizer)
 set_silent(model)
 @variable(model, x >= 0)
 @objective(model, Max, 2x + 1)
-@constraint(model, 2x - 1 <= -2)
+@constraint(model, con, 2x - 1 <= -2)
 
 # because the bound says that `x >= 0`, but we can rewrite the constraint to be
 # `x <= -1/2`. When the problem is infeasible, JuMP may return one of a number
@@ -229,6 +229,51 @@ termination_status(model)
 #     In this case, HiGHS does not support computing conflicts, but other
 #     solvers such as Gurobi and CPLEX do. If the solver does support computing
 #     conflicts, read [Conflicts](@ref) for more details.
+
+# ### PenaltyRelaxation
+
+# Another strategy to debug sources of infeasibility is the
+# [`penalty_relaxation`](@ref) function.
+#
+# The penalty relaxation modifies constraints of the form ``f(x) \\in S`` into
+# ``f(x) + y - z \\in S``, where ``y, z \\ge 0``, and then it introduces a
+# penalty term into the objective of ``a \\times (y + z)`` (if minimizing, else
+# ``-a``), where ``a`` is a penalty.
+
+map = penalty_relaxation(model)
+
+# Here `map` is a dictionary which maps constraint indices to an affine
+# expression representing ``(y + z)``.
+
+# If we optimize the relaxed model, this time we get a feasible solution:
+
+optimize!(model)
+termination_status(model)
+
+# Iterate over the contents of `map` to see which connstraints are violated:
+
+for (con, penalty) in map
+    violation = value(penalty)
+    if violation > 0
+        println("Constraint `$(name(con))` is violated by $violation")
+    end
+end
+
+# Once you find a violated constraint in the relaxed problem, take a look to see
+# if there is a typo or other common mistake in that particular constraint.
+
+# Consult the docstring [`penalty_relaxation`](@ref) for information on how to
+# modify the penalty cost term `a`, either for every constraint in the model or
+# a particular subset of the constraints.
+
+# When using [`penalty_relaxation`](@ref), you should be aware that:
+#
+#  * Variable bounds and integrality restrictions are not relaxed. If the
+#    problem is still infeasible after calling [`penalty_relaxation`](@ref),
+#    check the variable bounds.
+#  * You cannot undo the penalty relaxation. If you need an unmodified model,
+#    rebuild the problem, or call [`copy_model`](@ref) before calling
+#    [`penalty_relaxation`](@ref).
 
 # ## Debugging an unbounded model
 
--- a/docs/src/tutorials/getting_started/debugging.jl
+++ b/docs/src/tutorials/getting_started/debugging.jl
@@ -233,14 +233,14 @@ termination_status(model)
 # ### PenaltyRelaxation
 
 # Another strategy to debug sources of infeasibility is the
-# [`penalty_relaxation`](@ref) function.
+# [`penalty_relaxation!`](@ref) function.
 #
 # The penalty relaxation modifies constraints of the form ``f(x) \in S`` into
 # ``f(x) + y - z \in S``, where ``y, z \ge 0``, and then it introduces a
 # penalty term into the objective of ``a \times (y + z)`` (if minimizing, else
 # ``-a``), where ``a`` is a penalty.
 
-map = penalty_relaxation(model)
+map = penalty_relaxation!(model)
 
 # Here `map` is a dictionary which maps constraint indices to an affine
 # expression representing ``(y + z)``.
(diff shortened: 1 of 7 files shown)


docs/src/tutorials/getting_started/debugging.jl
```diff
@@ -230,6 +230,51 @@ termination_status(model)
...
+
+# Here `map` is a dictionary which maps constraint indices to an affine
+# expression representing ``(y + z)``.
+
+# If we optimize the relaxed model, this time we get a feasible solution:
+
+optimize!(model)
+termination_status(model)
+
+# Iterate over the contents of `map` to see which connstraints are violated:
```
odow, line 253:
```suggestion
# Iterate over the contents of `map` to see which constraints are violated:
```

src/constraints.jl
```diff
@@ -1344,3 +1344,107 @@ function all_constraints(
...
+    before calling [`penalty_relaxation`](@ref).
+
+## Reformulation
+
+See [`MOI.Utilities.ScalarPenaltyRelaxation`](@ref) for details of the
+reformulation.
+
+For each constraint `ci`, the penalty passed to
+[`MOI.Utilities.ScalarPenaltyRelaxation`](@ref) is `get(penalties, ci, default)`.
+If the value is `nothing`, because `ci` does not exist in `penalties` and
```
blegat, line 1370: What if the user puts `nothing` as a value in the dictionary ?

blegat:
According to the JuMP style guide:

> Omit `!` when the name itself makes it clear that modification is taking place, e.g., `add_constraint` and `set_name`

I don't think that we are in the case where "the name itself makes it clear that modification is taking place"

src/constraints.jl
odow: They can't. I'll update the type signature.

odow:
Updated to `penalty_relaxation!`

odow:
Bike-shedding, how about one of:

 * `make_penalty_relaxation(model)`
 * `create_penalty_relaxation(model)`
 * `add_penalty_relaxation(model)`
 * `reformulate_as_penalty_relaxation(model)`
 * `modify_model_for_penalty_relaxation(model)`

odow:
Following today's call, I've renamed to `relax_with_penalty!`.

blegat (approved):

odow:
I'll wait a couple more days for comments, and then I'll merge.

PR #3143, opened 2022-12-07
[docs] remove the workforce and steelT3 tutorials

odow:
This PR removes the workforce scheduling problem tutorial: https://jump.dev/JuMP.jl/stable/tutorials/linear/prod/

The tutorial is a bit of a mess. It's based on a model from AMPL, all of the data is hard-coded, and the variables and parameters use cryptic variable names. There's nothing in this tutorial that isn't better covered by other tutorials that we have. I looked into tidying it up, but it became a bit convoluted so I opted to delete instead.

I also looked at the analytics data.

It only had 46 views in the last 30 days:

image

Compare that to the 416 views for the diet problem:

image

Or even the 83 views for a tutorial like finance:

image


odow:
I also removed the steelT3 tutorial:
image

We should focus on fewer, higher quality tutorials instead of a bunch of examples that don't offer new content.

mlubin:
Just for reference of anyone wondering why we deleted these, which tutorials cover the content in workforce and steelT3?

mlubin (approved):

Issue #2348, opened 2020-10-03
Add documentation on parallel solves of a JuMP model


odow:
I get this question quite frequently. A section should be added to the docs explaining the different options.

Here is a great answer from SO: https://stackoverflow.com/questions/64177415/create-jump-model-with-multi-threading-in-julia

We should add a section on using `Distributed`. Here is an old answer from Discourse: https://discourse.julialang.org/t/parallel-solves-in-gurobi-jl/9908/8?u=odow

odow:
@jd-lara asks: Add a tutorial on how to get the "duals" of a MIP (i.e., solve, fix the integers, convert to LP, and re-solve). 

We had some code to do so in PowerSimulations (and gitter), but writing a nice tutorial would be good.

odow:
We should also add in-place solves for the subproblems of the column generation and Benders decomposition tutorials: 
https://jump.dev/JuMP.jl/dev/tutorials/algorithms/benders_decomposition/

odow:
- [ ] add a tutorial (e.g., AC-OPF) using complex number support: https://github.com/jump-dev/JuMP.jl/issues/3114.

odow:
- [ ] add an introduction to optimization tutorial. See the discussion in https://github.com/jump-dev/JuMP.jl/issues/2859

odow:
I added a bunch of examples for constraint programming: https://jump.dev/JuMP.jl/dev/tutorials/linear/constraint_programming/

We should do the same for conic. A lot of those sets aren't used very much because they're not easily visible from JuMP.

odow:
We're at a point where I've done most of the documentation changes I wanted to do. We still need to add a tutorial/example on the complex-valued stuff, and potentially a general introduction to optimization.

I'm not sure if the JuMP docs are the right place for a general introduction. Does anyone have a list of teaching materials that we could steer people towards instead?

jd-foster:
I think [Linear Programming with MATLAB](https://epubs.siam.org/doi/book/10.1137/1.9780898718775) is brilliant and only wish (1) it was open and (2) it had a Julia version. But that doesn't answer your question.

odow:
- [ ] Multiple people have been confused about setting warm starts for Ipopt because they follow the conic tutorial: 
      -  https://stackoverflow.com/questions/75543689/primal-and-dual-warm-starts-jump-error-message
      - https://discourse.julialang.org/t/jump-model-warm-start-using-ipopt/92660

[...]

PR #3148, opened 2022-12-08
[docs] clarify you can use index sets when setting the start value

### LINKED ISSUE #3147: [docs] clarify you can use index sets when setting the start value
Question from SO: https://stackoverflow.com/questions/74732967/how-does-one-elegantly-initialize-start-values-or-parameter-values-when-creating/74735001

odow:
Closes #3147

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Added Benders tutorial with in-place resolves (#3145)
  - Added more [Tips and tricks](@id linear_tips_and_tricks) for linear programs
    (#3144)
+ - Clarified documentation that `start` can depend on the indices of a
+   variable container (#3148)
 
 ## Version 1.5.0 (December 8, 2022)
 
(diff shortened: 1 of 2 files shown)


odow:
@ajcost I saw your comment too late 😆  Was this sufficient?

From your SO post:

> apologies as I am new to the language and tried all resources but cannot find a solution to this problem

Hi! Feedback on our documentation from newcomers is _super_ helpful. If you get stuck at any point, please open a PR or leave a comment here: https://github.com/jump-dev/JuMP.jl/issues/2348. 

You can make a PR to the docs by clicking the pencil icon in the top-right corner of any page of the documentation:

image

ajcost:
@odow no worries, yeah super helfpul only other suggestion is to maybe have the same doc level for @NLparameter as you do for @variable with the examples etc. but other than that it was super understandable. 

Great project!

odow:
Do you mean here https://jump.dev/JuMP.jl/stable/reference/nlp/#JuMP.@NLparameter?

Or here https://jump.dev/JuMP.jl/stable/manual/nlp/#Create-a-nonlinear-parameter?

ajcost:
yep nevermind that's what i was thinking of

PR #3150, opened 2022-12-13
Fix promotion and adjoint for complex expressions

blegat:
Some of the matrices constructed in the tests had eltype Any or GenericAffExpr with no type argument. With this PR, the eltype is now promoted properly.
Secondly, adjoint should redirect to conj in case coefficients are complex.

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -244,11 +244,7 @@ coefficient(::GenericAffExpr{C,V}, ::V, ::V) where {C,V} = zero(C)
 Remove terms in the affine expression with `0` coefficients.
 """
 function drop_zeros!(expr::GenericAffExpr)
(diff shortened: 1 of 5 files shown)


src/aff_expr.jl
```diff
@@ -496,16 +492,25 @@ end
...
+end
+
+function SparseArrays.dropzeros(aff::GenericAffExpr)
+    result = copy(aff)
+    _drop_zeros!(result.terms)
     if iszero(result.constant)
         # This is to work around isequal(0.0, -0.0) == false.
         result.constant = zero(typeof(result.constant))
+    elseif result.constant isa Complex && iszero(imag(result.constant))
+        result.constant = real(result.constant)
```
odow, line 513: This line needs a test

odow (approved):

Issue #2998, opened 2022-06-07
Containers do not support view


odow:
From Pierre Pasquet on slack:

```julia
julia> x = Containers.@container([i=[:a, :b]], 0)
1-dimensional DenseAxisArray{Int64,1,...} with index sets:
    Dimension 1, [:a, :b]
And data, a 2-element Vector{Int64}:
 0
 0

julia> view(x, :)
ERROR: MethodError: no method matching Base.Slice(::Vector{Symbol})
Closest candidates are:
  Base.Slice(::Base.Slice) at indices.jl:353
  Base.Slice(::T) where T<:AbstractUnitRange at indices.jl:351
Stacktrace:
 [1] uncolon(inds::Tuple{Vector{Symbol}}, I::Tuple{Colon})
   @ Base ./multidimensional.jl:827
 [2] to_indices
   @ ./multidimensional.jl:822 [inlined]
 [3] to_indices
   @ ./indices.jl:325 [inlined]
 [4] view(A::JuMP.Containers.DenseAxisArray{Int64, 1, Tuple{Vector{Symbol}}, Tuple{JuMP.Containers._AxisLookup{Dict{Symbol, Int64}}}}, I::Function)
   @ Base ./subarray.jl:176
 [5] top-level scope
   @ REPL[7]:1
```
We should throw a nicer error message, since this might be something that people do more often in future.

odow:
Annoyingly, it works if the indices are `AbstractUnitRange`:
```Julia
julia> x = Containers.@container([i=1:2], 0)
2-element Vector{Int64}:
 0
 0

julia> view(x, :)
2-element view(::Vector{Int64}, :) with eltype Int64:
 0
 0

julia> x = Containers.@container([i=2:3], 0)
1-dimensional DenseAxisArray{Int64,1,...} with index sets:
    Dimension 1, 2:3
And data, a 2-element Vector{Int64}:
 0
 0

julia> v = view(x, :)
2-element view(::JuMP.Containers.DenseAxisArray{Int64, 1, Tuple{UnitRange{Int64}}, Tuple{JuMP.Containers._AxisLookup{Tuple{Int64, Int64}}}}, :) with eltype Int64 with indices 2:3:
 0
 0

julia> v[1] = 1
[...]

PR #3152, opened 2022-12-15
[Containers] add support for view of DenseAxisArray

### LINKED ISSUE #3151: DenseAxisArray with Symbol key does not support setting
Reported on discourse: https://discourse.julialang.org/t/denseaxisarray-not-working-as-expected-with-single-dimensional/91646

```julia
julia> D = Containers.DenseAxisArray(zeros(3), [:a, :b, :c])
1-dimensional DenseAxisArray{Float64,1,...} with index sets:
    Dimension 1, [:a, :b, :c]
And data, a 3-element Vector{Float64}:
 0.0
 0.0
 0.0

julia> E = Containers.DenseAxisArray(ones(3), [:a, :b, :c])
1-dimensional DenseAxisArray{Float64,1,...} with index sets:
    Dimension 1, [:a, :b, :c]
And data, a 3-element Vector{Float64}:
 1.0
 1.0
 1.0

julia> I = [:a, :b]
2-element Vector{Symbol}:
 :a
 :b

julia> D[I] = E[I]
ERROR: MethodError: no method matching Int64(::Symbol)
Closest candidates are:
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at char.jl:50
  (::Type{T})(::BigInt) where T<:Union{Int128, Int16, Int32, Int64, Int8} at gmp.jl:356
  (::Type{T})(::Base.TwicePrecision) where T<:Number at twiceprecision.jl:243
  ...
Stacktrace:
 [1] (::Base.var"#105#106")(x::Vector{Symbol})
   @ Base ./abstractarray.jl:99
 [2] _tuple_any(::Base.var"#105#106", ::Bool, ::Vector{Symbol})
   @ Base ./tuple.jl:497
 [3] _tuple_any(f::Function, t::Tuple{Vector{Symbol}})
   @ Base ./tuple.jl:494
 [4] has_offset_axes(A::JuMP.Containers.DenseAxisArray{Float64, 1, Tuple{Vector{Symbol}}, Tuple{JuMP.Containers._AxisLookup{Dict{Symbol, Int64}}}})
   @ Base ./abstractarray.jl:99
 [5] require_one_based_indexing(A::JuMP.Containers.DenseAxisArray{Float64, 1, Tuple{Vector{Symbol}}, Tuple{JuMP.Containers._AxisLookup{Dict{Symbol, Int64}}}})
   @ Base ./abstractarray.jl:103
 [6] setindex!
   @ ./array.jl:851 [inlined]
 [7] setindex!(A::JuMP.Containers.DenseAxisArray{Float64, 1, Tuple{Vector{Symbol}}, Tuple{JuMP.Containers._AxisLookup{Dict{Symbol, Int64}}}}, v::JuMP.Containers.DenseAxisArray{Float64, 1, Tuple{Vector{Symbol}}, Tuple{JuMP.Containers._AxisLookup{Dict{Symbol, Int64}}}}, idx::Vector{Symbol})
   @ JuMP.Containers ~/.julia/packages/JuMP/puvTM/src/Containers/DenseAxisArray.jl:344
 [8] top-level scope
   @ REPL[13]:1

julia> D[I] = 0.0
ERROR: ArgumentError: indexed assignment with a single value to many locations is not supported; perhaps use broadcasting `.=` instead?
Stacktrace:
 [1] setindex_shape_check(::Float64, ::Int64)
   @ Base ./indices.jl:261
 [2] _unsafe_setindex!(#unused#::IndexLinear, A::Vector{Float64}, x::Float64, I::Vector{Int64})
   @ Base ./multidimensional.jl:902
 [3] _setindex!
   @ ./multidimensional.jl:893 [inlined]
 [4] setindex!
   @ ./abstractarray.jl:1267 [inlined]
 [5] setindex!(A::JuMP.Containers.DenseAxisArray{Float64, 1, Tuple{Vector{Symbol}}, Tuple{JuMP.Containers._AxisLookup{Dict{Symbol, Int64}}}}, v::Float64, idx::Vector{Symbol})
   @ JuMP.Containers ~/.julia/packages/JuMP/puvTM/src/Containers/DenseAxisArray.jl:344
 [6] top-level scope
   @ REPL[14]:1

julia> D[I] .= 0.0
ERROR: MethodError: no method matching one(::Symbol)
Closest candidates are:
  one(::Union{Type{T}, T}) where T<:AbstractString at strings/basic.jl:262
  one(::Union{Type{P}, P}) where P<:Dates.Period at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Dates/src/periods.jl:54
  one(::LinearAlgebra.Tridiagonal{T, V} where V<:AbstractVector{T}) where T at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/special.jl:307
  ...
Stacktrace:
  [1] isone(x::Symbol)
    @ Base ./number.jl:60
  [2] (::JuMP.Containers.var"#9#11"{Vector{Symbol}})(i::Int64)
    @ JuMP.Containers ~/.julia/packages/JuMP/puvTM/src/Containers/DenseAxisArray.jl:323
  [3] _any
    @ ./reduce.jl:876 [inlined]
  [4] #any#702
    @ ./reducedim.jl:883 [inlined]
  [5] any
    @ ./reducedim.jl:883 [inlined]
  [6] to_index(A::JuMP.Containers.DenseAxisArray{Float64, 1, Tuple{Vector{Symbol}}, Tuple{JuMP.Containers._AxisLookup{Dict{Symbol, Int64}}}}, idx::Vector{Symbol})
    @ JuMP.Containers ~/.julia/packages/JuMP/puvTM/src/Containers/DenseAxisArray.jl:323
  [7] to_indices
    @ ./indices.jl:333 [inlined]
  [8] to_indices
    @ ./indices.jl:325 [inlined]
  [9] view
    @ ./subarray.jl:176 [inlined]
 [10] maybeview
    @ ./views.jl:133 [inlined]
 [11] dotview(::JuMP.Containers.DenseAxisArray{Float64, 1, Tuple{Vector{Symbol}}, Tuple{JuMP.Containers._AxisLookup{Dict{Symbol, Int64}}}}, ::Vector{Symbol})
    @ Base.Broadcast ./broadcast.jl:1212
 [12] top-level scope
    @ REPL[15]:1
```

odow:
Closes #3151
Part of #2998 (still to do SparseAxisArray?)

I'm not overly convinced of this approach. I felt I was just poking methods until things were working. For something like `D[I] .= 1.0`, Base uses a weird dispatch down `dotview` and `maybeview`.

I should add more tests before merging, and consider also SparseAxisArray.

There's also a question: should you be able to access a value that isn't in the view, but is in the underlying array?

### DIFF
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -353,8 +353,6 @@ function Base.setindex!(
     value::DenseAxisArray{T,N},
     args...,
 ) where {T,N}
-    @show args
-    @show Base.to_index(A, args)
     for key in Base.product(args...)
         A[key...] = value[key...]
     end
@@ -583,7 +581,7 @@ Base.repeat(x::DenseAxisArray; kwargs...) = repeat(x.data; kwargs...)
 
 _get_subaxis(::Colon, b) = b
 
-function _get_subaxis(a, b)
+function _get_subaxis(a::AbstractVector, b)
     for ai in a
         if !(ai in b)
             throw(KeyError(ai))
@@ -591,6 +589,14 @@ function _get_subaxis(a, b)
     end
     return a
 end
+
+function _get_subaxis(a::T, b::AbstractVector{T}) where {T}
+    if !(a in b)
+        throw(KeyError(a))
+    end
+    return a
+end
+
 struct DenseAxisArrayView{T,N,D,A} <: AbstractArray{T,N}
     data::D
     axes::A
@@ -598,7 +604,7 @@ struct DenseAxisArrayView{T,N,D,A} <: AbstractArray{T,N}
         x::Containers.DenseAxisArray{T,N},
         args...,
     ) where {T,N}
-        axis = tuple([_get_subaxis(a, b) for (a, b) in zip(args, axes(x))]...)
+        axis = _get_subaxis.(args, axes(x))
         return new{T,N,typeof(x),typeof(axis)}(x, axis)
     end
 end
@@ -612,15 +618,20 @@ Base.size(x::DenseAxisArrayView) = length.(x.axes)
 Base.axes(x::DenseAxisArrayView) = x.axes
 
 function Base.getindex(x::DenseAxisArrayView, args...)
-    return getindex(x.data, args...)
+    y = _get_subaxis.(args, x.axes)
+    return getindex(x.data, y...)
 end
+
+Base.getindex(a::DenseAxisArrayView, k::DenseAxisArrayKey) = a[k.I...]
 
 function Base.setindex!(x::DenseAxisArrayView, args...)
     return setindex!(x.data, args...)
 end
 
 function Base.eachindex(A::DenseAxisArrayView)
-    return DenseAxisArrayKey.(Base.product(A.axes...))
+    # Return a generator so that we lazily evaluate the product instead of
+    # collecting into a vector.
+    return (DenseAxisArrayKey(k) for k in Base.product(A.axes...))
 end
 
 Base.show(io::IO, x::DenseAxisArrayView) = print(io, x.data)
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -631,6 +631,10 @@ end
 function Base.eachindex(A::DenseAxisArrayView)
     # Return a generator so that we lazily evaluate the product instead of
     # collecting into a vector.
+    #
+    # In future, we might want to return the appropriate matrix of
+    # `CartesianIndex` to avoid having to do the lookups with
+    # `DenseAxisArrayKey`.
     return (DenseAxisArrayKey(k) for k in Base.product(A.axes...))
 end
 
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -348,6 +348,19 @@ function Base.IndexStyle(::Type{DenseAxisArray{T,N,Ax}}) where {T,N,Ax}
     return IndexAnyCartesian()
 end
 
+function Base.setindex!(
+    A::DenseAxisArray{T,N},
+    value::DenseAxisArray{T,N},
+    args...,
+) where {T,N}
+    @show args
+    @show Base.to_index(A, args)
+    for key in Base.product(args...)
+        A[key...] = value[key...]
+    end
+    return A
+end
+
 ########
 # Keys #
 ########
@@ -362,6 +375,10 @@ struct DenseAxisArrayKey{T<:Tuple}
 end
 Base.getindex(k::DenseAxisArrayKey, args...) = getindex(k.I, args...)
 Base.getindex(a::DenseAxisArray, k::DenseAxisArrayKey) = a[k.I...]
+
+function Base.setindex!(A::DenseAxisArray, value, key::DenseAxisArrayKey)
+    return setindex!(A, value, key.I...)
+end
 
 struct DenseAxisArrayKeys{T<:Tuple,S<:DenseAxisArrayKey,N} <: AbstractArray{S,N}
     product_iter::Base.Iterators.ProductIterator{T}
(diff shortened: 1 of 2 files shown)


src/Containers/DenseAxisArray.jl
```diff
@@ -348,6 +348,19 @@ function Base.IndexStyle(::Type{DenseAxisArray{T,N,Ax}}) where {T,N,Ax}
     return IndexAnyCartesian()
 end
 
+function Base.setindex!(
+    A::DenseAxisArray{T,N},
+    value::DenseAxisArray{T,N},
+    args...,
+) where {T,N}
+    @show args
+    @show Base.to_index(A, args)
```
blegat, line 357: Remove

odow: Oops. Yeah. I guess my bigger question was;

> I'm not overly convinced of this approach.
> should you be able to access a value that isn't in the view, but is in the underlying array?

src/Containers/DenseAxisArray.jl
```diff
@@ -559,3 +576,57 @@ end
...
+    return a
+end
+struct DenseAxisArrayView{T,N,D,A} <: AbstractArray{T,N}
+    data::D
+    axes::A
+    function DenseAxisArrayView(
+        x::Containers.DenseAxisArray{T,N},
+        args...,
+    ) where {T,N}
+        axis = tuple([_get_subaxis(a, b) for (a, b) in zip(args, axes(x))]...)
```
blegat, line 601:
```suggestion
        axis = _get_subaxis.(args, axes(x))
```

src/Containers/DenseAxisArray.jl
```diff
@@ -559,3 +576,57 @@ end
...
+function Base.getindex(x::DenseAxisArrayView, args...)
+    return getindex(x.data, args...)
+end
+
+function Base.setindex!(x::DenseAxisArrayView, args...)
+    return setindex!(x.data, args...)
+end
+
+function Base.eachindex(A::DenseAxisArrayView)
+    return DenseAxisArrayKey.(Base.product(A.axes...))
```
blegat, line 623: For `DenseAxisArray` we can provide the fast access with `CartesianIndex`, maybe leave a comment here saying that this is a bit suboptimal but fine for now

blegat: The whole point of views is about performance so this might give surprising user experience

src/Containers/DenseAxisArray.jl
blegat: I don't think we should give access to it

odow:
Okay. I fixed the `getindex` problem to respect the new axes, changed `eachindex` to be lazy, and added a couple more tests.

src/Containers/DenseAxisArray.jl
```diff
@@ -559,3 +574,70 @@ end
...
+Base.getindex(a::DenseAxisArrayView, k::DenseAxisArrayKey) = a[k.I...]
+
+function Base.setindex!(x::DenseAxisArrayView, args...)
+    return setindex!(x.data, args...)
+end
+
+function Base.eachindex(A::DenseAxisArrayView)
+    # Return a generator so that we lazily evaluate the product instead of
+    # collecting into a vector.
+    return (DenseAxisArrayKey(k) for k in Base.product(A.axes...))
```
blegat, line 634: The access provided is still slower since we need to do the dictionary `getindex` instead of getting directly the elements of the underlying array.
It would be more efficient to return an iterator over the `CartesianIndex` directly by looking at the subset of each axes of `x.data` that we have.

odow: > It would be more efficient to return an iterator over the CartesianIndex directly by looking at the subset of each axes of x.data that we have.

But that would require similar lookups, just in `eachindex` instead of `getindex`? The view does not need to be contiguous. Although I guess it'd only need `sum(length.(A.axes))` instead of `length(A)` lookups.

I think this is a premature optimization. Let's see if anyone complains first.

blegat: > Although I guess it'd only need `sum(length.(A.axes))` instead of `length(A)` lookups.

Yes exactly

> I think this is a premature optimization. Let's see if anyone complains first.

I agree but I would leave a comment in the code

blegat (approved):

PR #3156, opened 2022-12-19
Add explanatory warning before OptimizeNotCalled error

### LINKED ISSUE #3155: Improve Error Messages for Querying Modified Models
With the following simple (yet a little contrived example) I get a relatively unintuitive error:
```julia
using JuMP, Gurobi
@variable(model, x[1:2] >= 0)
@objective(model, Min, x[1])
optimize!(model)

for i in 1:2
    set_start_value(x[i], value(x[i]))
end
```
```julia
ERROR: OptimizeNotCalled()
Stacktrace:
 [1] get(model::Model, attr::MathOptInterface.VariablePrimal, v::VariableRef)
   @ JuMP C:\Users\bbgui\.julia\packages\JuMP\puvTM\src\JuMP.jl:1249
 [2] value(v::VariableRef; result::Int64)
   @ JuMP C:\Users\bbgui\.julia\packages\JuMP\puvTM\src\variables.jl:1032
 [3] value(v::VariableRef)
   @ JuMP C:\Users\bbgui\.julia\packages\JuMP\puvTM\src\variables.jl:1032
 [4] top-level scope
   @ .\REPL[19]:2
```
The warning provided in https://github.com/jump-dev/JuMP.jl/blob/9b4c88021178f6b1245a78cb1a23852bc1778eeb/src/JuMP.jl#L1206-L1211 would be more helpful, but this is only implemented for `MOI.get` with `attr::MOI.AbstractModelAttribute`, not the other dispatches. It would be great to add this to the other methods.

In this case and the larger less contrived model this was inspired from, I can use direct mode to get around this and avoid unnecessary loops. It would be nice to have a flag when initializing the model to opt out of using `is_model_dirty`. I am working on a benchmarking paper right now on iterative algorithms using JuMP and Pyomo where this capability would allow for more of an apples to apples comparison with Pyomo such that I avoid using direct mode and/or using a different update strategy.

odow:
Closes #3155

### DIFF
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -305,7 +305,7 @@ If you are iteratively querying solution information and modifying a model,
 query all the results first, then modify the problem.
 
 For example, instead of:
-```jldoctest
+```jldoctest; filter = r"\@ JuMP.+/src/JuMP.jl"
 julia> model = Model(HiGHS.Optimizer);
 
 julia> set_silent(model)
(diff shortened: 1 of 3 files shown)


odow:
@pulsipher are you happy if I merge this? Or do you still want something else?

PR #3158, opened 2022-12-20
Refactor test/Containers into functional form

odow:


odow:
These test failures are precisely the reason to refactor the tests. Changing the scope of files in `test/Containers` breaks an unrelated test in `test/containers.jl` (because `test/Containers` no longer includes a global `using JuMP` statement).

PR #3168, opened 2023-01-03
Refactor runtests.jl to use a single run_tests function

odow:
This PR continues the refactoring I have been doing of JuMP's tests.

 * Tests are now contained in a module which lives in test/test_xxx.jl file
 * Within a module, tests are functions beginning with test_
 * JuMPExtensions are tested by functions with the prefix test_extension_ and take two optional positional argumennts for the model type and variable type
 * Test individual files via the command line \`julia runtests.jl [\<file\> [\<file\> ...]]\`

This design is flexible, and separates the runninng of the tests from their definition. This should let us swap in other drivers instead of the one currently in runtests.jl.

For example, we might want to experiment with:

 * Running all test functions within a module in parallel
 * Running each test file in parallel
 * Providing a more flexible command line interface

If we get something nice, we could package it into Pukeko.jl for wider use in the Julia ecosystem.

Relatedly, this closes https://github.com/jump-dev/JuMP.jl/issues/1711, because it should enable external packages to run:
```Julia
import JuMP
include(joinpath(dirname(dirname(pathof(JuMP))), "test", "Kokako.jl"))
const MODULES_TO_TEST = Kokako.include_modules_to_test(JuMP)
Kokako.run_tests(
    MODULES_TO_TEST,
    JuMPExtension.MyModel,
    JuMPExtension.MyVariableRef;
    test_prefix = "test_extension_",
)
```


odow:
Okay. I like where this is at. We have a decision on where the `Kokako.jl` file should live. 

[Pukeko.jl](https://github.com/IainNZ/Pukeko.jl) is a registered package that I was [shamelessly inspired by](https://www.doc.govt.nz/nature/native-animals/birds/birds-a-z/kokako/).

Using the `Base.Test` macros makes things a lot simpler, which is one point in favor of this system. Perhaps I'll change the license to MIT for that file so that we can leave it here for now, and anyone is free to rip it off later. The functions are trivial.

blegat (approved):

blegat:
We can leave it in JuMP to play around it for a bit but I think it would be nice as an external package since we are using this pattern in many packages now

PR #3173, opened 2023-01-04
Fix method ambiguities in JuMP.Containers

odow:


src/Containers/DenseAxisArray.jl
```diff
@@ -253,12 +253,13 @@ end
...
 
-# We specify `Ax` for the type of `axes` to avoid conflict where `axes` has type `Tuple{Vararg{Int,N}}`.
+# We specify `Ax` for the type of `axes` to avoid conflict where `axes` has type
+# `Tuple{Vararg{Int,N}}`.
 function Base.similar(
     A::DenseAxisArray{T,N,Ax},
     ::Type{S},
     axes::Ax,
-) where {T,N,Ax,S}
+) where {T,N,Ax<:Tuple{<:AbstractVector},S}
```
blegat, line 262: We also support axes that are not `AbstractVector` at the moment so that would be breaking

odow: The type allows it, but we convert everything to `AbstractVector` on creation: https://github.com/jump-dev/JuMP.jl/pull/2713.

https://github.com/jump-dev/JuMP.jl/blob/9f61e95c3ef9db87d3de24b2bea85c7752b6277d/src/Containers/DenseAxisArray.jl#L158-L206

odow:
Urgh. https://github.com/jump-dev/JuMP.jl/pull/3152 just added new ambiguities. I think that shows we need to get these tests in ASAP to keep a lid on things.

blegat (approved):

PR #3176, opened 2023-01-06
Add support for vector-valued objectives

### LINKED ISSUE #2099: Multiobjective support in JuMP

odow:
Closes #2099 

## Overview

The issue https://github.com/jump-dev/JuMP.jl/issues/2099 discussed two approaches for implementing MO in JuMP.

 1. Treat multicriteria as a vector of scalar objectives and a vector of scalar senses. This would let someone write [Min, Max], [f(x), g(x)]. The main reason for this approach is that it matches what users want to do.
 2. Treat multicriteria as an optimization problem with a vector-valued objective function. Users could write only Min, f(x) where f(x) is a vector. The main reason for this approach is that it matches what MathOptInterface wants.

This PR implements option 2. The strongest reason in support of option 2 is that it requires very little code to implement, suggesting that it is a natural extension of MOI.

There are also solvers, like Gurobi, that support 2, but not 1. That is, they have a single objective sense but support multiple objectives: https://www.gurobi.com/documentation/10.0/refman/c_setobjectiven.html. (You could work-around that by setting the weight of the `Max` objective to `-1` instead of `1`, but that's getting a bit finicky.)

The biggest downside is that it doesn't overcome the Min-Max issue; but I think we can work around this with user-facing cosmetic tooling in JuMP; solvers would be forced to accept a single sense.

## Documentation

See https://jump.dev/JuMP.jl/previews/PR3176/manual/objective/#Set-a-vector-valued-objective

## Testing it out

The MOI PR is:

 * https://github.com/jump-dev/MathOptInterface.jl/pull/2070

The supported solvers are:

 * https://github.com/odow/MOO.jl (Follow https://github.com/odow/MOO.jl#installation)
 * https://github.com/jump-dev/Gurobi.jl/pull/497 (`] add Gurobi#od/vector-optimization`)

Examples to follow are:

 * https://github.com/odow/MOO.jl/blob/master/test/mo_models.jl

but this explains the basic syntax. Currently you need `add_bridges = false` because of an issue in MOI.

```julia
function example_jump_bolp_1_min_max(f)
    model = Model(f; add_bridges = false)
    set_silent(model)
    @variable(model, x >= 0.0)
    @variable(model, y >= 0.25)
    @constraint(model, x + y >= 1.0)
    @constraint(model, 0.5 * x + y >= 0.75)
    @expression(model, obj1, 2 * x + y)
    @expression(model, obj2, -x - 3 * y)
    # We want
    #   min obj1
    #   max obj2
    # But you must pick a single sense.
    @objective(model, Min, [obj1, -obj2])
    optimize!(model)
    @test termination_status(model) == OPTIMAL
    @test result_count(model) == 3
    X = [[0.0, 1.0], [0.5, 0.5], [1.0, 0.25]]
    Y = [[1.0, 3.0], [1.5, 2.0], [2.25, 1.75]]
    for i in 1:3
        @test primal_status(model; result = i) == MOI.FEASIBLE_POINT
        @test dual_status(model; result = i) == MOI.NO_SOLUTION
        @test objective_value(model; result = i) == Y[i]
        @test value(x; result = i) == X[i][1]
        @test value(y; result = i) == X[i][2]
    end
    @test objective_bound(model) == [1.0, 1.75]
    return
end
```


odow:
Picking on the main sticking points from #2099

> we have chosen to handle the objectives separately. Adding an index which denote a specific objective as suggested by @odow is our preferred option because we would like in a near future to remove one objective without redefining all the model.

You can re-define the objective without rebuilding the model. This is working with MOO.jl

```Julia
model = Model()
@variable(model, x >= 0.0)
@variable(model, y >= 0.25)
@constraint(model, x + y >= 1.0)
@constraint(model, 0.5 * x + y >= 0.75)
@expression(model, obj1, 2 * x + y)
@expression(model, obj2, -x - 3 * y)
# Optimize two objectives
@objective(model, Min, [obj1, -obj2])
optimize!(model)
# ... do something with solution
# Drop objective 2
@objective(model, Min, [obj1])
optimize!(model)
# ... do something with solution
```

> a discussion about how to pass the resolution method to invoke (and possibly the required parameters) is needed.

These are just normal solver parameters
```julia
using JuMP
import HiGHS, MOO
model = JuMP.Model(() -> MOO.Optimizer(HiGHS.Optimizer))
set_optimizer_attribute(model, "algorithm", MOO.NISE())
```

> a discussion about how to retrieve Y_N and X_E is needed as soon as the problem is not discrete. Indeed, for the MO-MILP case, Y_N may be composed of points, segments (open or closed), facets (full or not; open or closed). We have not yet stated about a protocol in the vOpt team. One option may be to return each object separately, with its proprieties, and a description of their adjacency relations.

This is tricky. For now, `value(x; result = i)` and `objective_value(model, ; result = i)` return points. Presumably, we'd need some solver-specific attributes here for the facets. For example:
```julia
Y_N = MOI.get(model, vOpt.EfficientFacets())
```

> In writing `@objective(model, Max, [dot(x, p1), dot(x, p2)])`, the user cannot
> (1) blend functions to minimize and to maximize in the model, and

This is still a problem.

> (2) point out a function to delete without rebuilding all the model.

You can set a new objective, but we can't say "delete the 2nd objective from this vector." Would solvers benefit from the ability to do this?

docs/make.jl
```diff
@@ -1,5 +1,6 @@
 import Pkg
 Pkg.pkg"add Documenter#71e9f40"
+Pkg.pkg"add MathOptInterface#od/vector-optimization"
```
odow, line 3:
```suggestion
Pkg.pkg"add MathOptInterface#master"
```

docs/src/manual/objective.md
```diff
@@ -157,3 +157,80 @@ julia> @objective(model, Min, 2x)
...
+2-element Vector{AffExpr}:
+ x[1] + 1
+ 2 x[2]
+```
+
+In most cases, multi-objective optimization solvers will return multiple
+solutions, corresponding to points on the Pareto frontier. See [Multiple solutions](@ref)
+for information on how to query and work with multiple solutions.
+
+Note that you can must set a single objective sense, that is, you cannot have
```
odow, line 183:
```suggestion
Note that you must set a single objective sense, that is, you cannot have
```

docs/src/manual/objective.md
```diff
@@ -157,3 +157,80 @@ julia> @objective(model, Min, 2x)
...
+
+julia> @objective(model, Min, [obj1, obj2, obj3])  # Three-objective problem
+3-element Vector{AffExpr}:
+ x[1] + 1
+ 2 x[1]
+ x[1] + x[2]
+
+julia> # optimize!(model), look at the solution, talk to stakeholders, then
+       # decide you want to solve a new problem where the third objective is
+       # removed and constrained to be better than 2.0.
```
odow, line 226:
```suggestion
       # removed and constrained to be no worse than 2.0.
```

test/runtests.jl
```diff
@@ -8,6 +8,9 @@
 # See https://github.com/jump-dev/JuMP.jl
 #############################################################################
 
+import Pkg
+Pkg.pkg"add MathOptInterface#od/vector-optimization"
```
odow, line 12:
```suggestion
Pkg.pkg"add MathOptInterface#master"
```

test/test_generate_and_solve.jl
```diff
@@ -552,4 +552,42 @@ c2: x + y <= 1.0
...
+    @variable(model, y >= 1.0)
+    @objective(model, Min, [x, y])
+    optimize!(model)
+    mock = unsafe_backend(model)
+    MOI.set(mock, MOI.TerminationStatus(), MOI.OPTIMAL)
+    MOI.set(mock, MOI.RawStatusString(), "Optimal")
+    MOI.set(mock, MOI.ResultCount(), 1)
+    MOI.set(mock, MOI.PrimalStatus(), MOI.FEASIBLE_POINT)
+    MOI.set(mock, MOI.DualStatus(), MOI.NO_SOLUTION)
+    # MOI.set(mock, MOI.ObjectiveValue(), [0.0, 1.0])
```
odow, line 569:
```suggestion
```

odow:
Before merging, I'd like to tag https://github.com/jump-dev/MultiObjectiveAlgorithms.jl as a public package and add a tutorial to the documentation. This would let us document both vector-valued objectives, and solvers which support multiple solutions.

docs/make.jl
```diff
@@ -1,5 +1,6 @@
 import Pkg
 Pkg.pkg"add Documenter#71e9f40"
+Pkg.pkg"add https://github.com/jump-dev/MultiObjectiveAlgorithms.jl"
```
odow, line 3: TODO: replace this with tagged version before merging

odow:
I've added a tutorial using MultiObjectiveAlgorithms. I think it's pretty cool. The ease to which we can solve multi-objective problems is going to be a nice boost to the popularity of JuMP. 

I'll post a link here once CI finishes.

odow:
Here's the tutorial: https://jump.dev/JuMP.jl/previews/PR3176/tutorials/linear/multi_objective_knapsack

mlubin (approved):

docs/src/tutorials/linear/multi_objective_knapsack.jl
```diff
@@ -0,0 +1,197 @@
...
+
+# Note that because we set a vector of two objective functions, the objective
+# value is a vector with two elements. We can also query the value of each
+# objective separately:
+
+value(profit_expr; result = 5)
+
+# ## Visualizing objective space
+
+# Unlike single-objective optimization problem, multi-objective optimization
```
odow, line 168:
```suggestion
# Unlike single-objective optimization problems, multi-objective optimization
```

docs/src/manual/solutions.md
```diff
@@ -573,6 +573,10 @@ for i in 2:result_count(model)
 end
 ```
 
+!!! tip
+    The [Multi-objective knapsack](@ref) tutorial includes an example of
```
odow, line 577:
```suggestion
    The [Multi-objective knapsack](@ref) tutorial provides an example of
```

docs/src/tutorials/linear/multi_objective_knapsack.jl
```diff
@@ -0,0 +1,180 @@
...
+#      & \sum\limits_{i \in I} r_i x_i \\
+# \text{s.t.}\ \ & \sum\limits_{i \in I} w_i x_i \le c\\
+# & x_i \in \{0, 1\} && \forall i \in I
+# \end{aligned}
+# ```
+
+# ## Data
+
+# The data for this example was taken from [vOptGeneric](https://github.com/vOptSolver/vOptGeneric.jl),
+# and the original author was [@xgandibleux].
```
odow, line 63:
```suggestion
# and the original author was [@xgandibleux](https://github.com/xgandibleux).
```

odow:
@xgandibleux: I've updated the tutorial to use vectors instead of the `Item` struct I had before:
https://jump.dev/JuMP.jl/previews/PR3176/tutorials/linear/multi_objective_knapsack/#Data

xgandibleux:
Great!

odow:
After 3+ years (#2099), this is finally ready to merge!

docs/Project.toml
```diff
@@ -33,7 +34,8 @@ Ipopt = "=1.1.0"
 JSON = "0.21"
 JSONSchema = "1"
 Literate = "2.8"
-MathOptInterface = "=1.11.5"
+MathOptInterface = "=1.12.0"
+MultiObjectiveAlgorithms = "=0.1.0"
```
odow, line 38:
```suggestion
MultiObjectiveAlgorithms = "=0.1.1"
```

src/objective.jl
```diff
@@ -25,8 +25,14 @@ end
 
 Return the best known bound on the optimal objective value after a call to
 `optimize!(model)`.
+
+For scalar-valued objectives, this function returns a `Float64`. For
+vector-valued objectives, it returns a `Vector{Float64}`.
+
+In the case of a vector-valued objective, this returns the ideal point, that
```
blegat, line 32: What is an "ideal point" ? Is this a technical term from the multi-objective field ?

odow: Yes. See https://en.wikipedia.org/wiki/Multi-objective_optimization ("ideal objective vector").

It's also plotted in the tutorial:

image

blegat: Ok, maybe write it in italic to signify its the first occurence of this technical word and maybe add a href to the wikipedia page

odow, line 32:
```suggestion
In the case of a vector-valued objective, this returns the _ideal point_, that
```

odow: Added italics. I don't think the href is useful. We don't link to other technical terms, and the docstring includes a `that is, ...` comment in non-jargon.

blegat (approved):

xgandibleux:
Ideed, Ideal point is a common reference point in MOO, and I agree with Oscar, extra explanations are not required here.

PR #3181, opened 2023-01-09
Improve error message for common case of using Bool instead of Bin

odow:
This PR was motivated by the discourse post:
https://discourse.julialang.org/t/how-to-declare-a-boolean-variable/41279

It is one of the top-10 most read posts on the JuMP section of the discourse forum, so it is obviously a common stumbling block, particularly because we support Int for integer variables.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1454,6 +1454,13 @@ function build_variable(
         )
     end
     for (key, _) in kwargs
+        if key == :Bool
+            _error(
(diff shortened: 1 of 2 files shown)


odow:
This is definitely a JuMP 2.0 consideration. It's weird that we have `@variable(model, x, Int)` that relies on parsing the symbol `:Int`, and the fact that `Bin` isn't a variable in scope. `@variable(model, x, Bool)` would make more sense.

PR #3182, opened 2023-01-09
Switch to LU factorization in lp_sensitivity_report

odow:
@adow031 has a case where this factorization is failing, saying:

> If this shouldn't happen, I can give you the code to take a look at it (I couldn't find a simple case where this would happen) - but it's a bit of an edge case, so it's not a big deal for me.

```julia
Presolving model
0 rows, 0 cols, 0 nonzeros
0 rows, 0 cols, 0 nonzeros
Presolve : Reductions: rows 0(-3); columns 0(-5); elements 0(-8) - Reduced to empty
Solving the original LP from the solution after postsolve
Model   status      : Optimal
Objective value     : -1.0000000000e+00
HiGHS run time      :          0.03
Solving LP without presolve or with basis
Using EKK dual simplex solver - serial
  Iteration        Objective     Infeasibilities num(sum)
          0     0.0000000000e+00 Ph1: 0(0) 0s
          1     0.0000000000e+00 Pr: 0(0) 0s
Model   status      : Optimal
Simplex   iterations: 1
Objective value     :  0.0000000000e+00
HiGHS run time      :          0.04

ERROR: ZeroPivotException: factorization encountered one or more zero pivots. Consider switching to a pivoted LU factorization.
Stacktrace:
  [1] #ldlt!#10
    @ C:\Users\adow031\AppData\Local\Programs\Julia-1.8.3\share\julia\stdlib\v1.8\SuiteSparse\src\cholmod.jl:1308 [inlined]
  [2] ldlt!(F::SuiteSparse.CHOLMOD.Factor{Float64}, A::LinearAlgebra.Hermitian{Float64, SparseArrays.SparseMatrixCSC{Float64, Int64}}; shift::Float64, check::Bool)
    @ SuiteSparse.CHOLMOD C:\Users\adow031\AppData\Local\Programs\Julia-1.8.3\share\julia\stdlib\v1.8\SuiteSparse\src\cholmod.jl:1333
  [3] ldlt!
    @ C:\Users\adow031\AppData\Local\Programs\Julia-1.8.3\share\julia\stdlib\v1.8\SuiteSparse\src\cholmod.jl:1333 [inlined]
  [4] factorize
    @ C:\Users\adow031\AppData\Local\Programs\Julia-1.8.3\share\julia\stdlib\v1.8\SparseArrays\src\linalg.jl:1631 [inlined]
  [5] factorize(A::SparseArrays.SparseMatrixCSC{Float64, Int64})
    @ SparseArrays C:\Users\adow031\AppData\Local\Programs\Julia-1.8.3\share\julia\stdlib\v1.8\SparseArrays\src\linalg.jl:1609
  [6] lp_sensitivity_report(model::Model; atol::Float64)
    @ JuMP C:\Users\adow031\.julia\packages\JuMP\vuP7I\src\lp_sensitivity2.jl:93
  [7] lp_sensitivity_report
    @ C:\Users\adow031\.julia\packages\JuMP\vuP7I\src\lp_sensitivity2.jl:63 [inlined
```

I think we should be fine forcing `lu` here; we don't need the special casing for structured matrices. Not sure how to test a case where the `ldlt` was chosen and failed.

### DIFF
--- a/src/lp_sensitivity2.jl
+++ b/src/lp_sensitivity2.jl
@@ -90,17 +90,17 @@ function lp_sensitivity_report(model::Model; atol::Float64 = 1e-8)
     l_B = @view std_form.lower[basis.basic_cols]
     u_B = @view std_form.upper[basis.basic_cols]
 
-    d = Dict{Int,Vector{Float64}}()
-    if size(B, 1) > 0
-        B_fact = LinearAlgebra.lu(B)
-        for j in 1:length(basis.basic_cols)
-            if basis.basic_cols[j] == false
-                # We call `collect` here because some Julia versions are missing
-                # sparse matrix \ sparse vector fallbacks.
-                d[j] = B_fact \ collect(std_form.A[:, j])
-            end
-        end
+    B_fact = if size(B, 1) > 0
+        LinearAlgebra.lu(B)
+    else
+        zeros(Float64, (0, 0))
     end
+    d = Dict{Int,Vector{Float64}}(
+        # We call `collect` here because some Julia versions are missing sparse
+        # matrix \ sparse vector fallbacks.
+        j => B_fact \ collect(std_form.A[:, j]) for
+        j in 1:length(basis.basic_cols) if basis.basic_cols[j] == false
+    )
 
     report = SensitivityReport(
         Dict{ConstraintRef,Tuple{Float64,Float64}}(),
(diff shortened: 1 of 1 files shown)


odow:
> I think we should be fine forcing lu here

I guess I was wrong

odow:
@adow031 do you want to try this branch? `] add JuMP#odow-patch-1`

odow:
@adow031 reports that this is working correctly on the problems that were previously failing.

blegat:
> > I think we should be fine forcing lu here
>
> I guess I was wrong

What's the conclusion ?

odow:
> What's the conclusion ?

`lu` threw an error when the matrix if `0x0`: https://github.com/jump-dev/JuMP.jl/pull/3182/commits/8d1bc98e0f33fc50d9b88c8d0fd48be19b6e96ea

PR #3188, opened 2023-01-12
Refactor JuMP.jl, moving methods into their related files

odow:
The diff is fairly large, but this has always annoyed me: The previous content of JuMP.jl was confusing. It wasn't ordered, and it contained a mix of methods that are better off living in other files.

For example, why was ` index(cr::ConstraintRef)` in JuMP.jl, but ` index(v::VariableRef)` in variables.jl?

There were also things like `result_count(model)` was in `optimizer_interface.jl`, but `termination_status(model)` was in JuMP.jl.

### DIFF
--- a/docs/src/manual/solutions.md
+++ b/docs/src/manual/solutions.md
@@ -305,7 +305,7 @@ If you are iteratively querying solution information and modifying a model,
 query all the results first, then modify the problem.
 
 For example, instead of:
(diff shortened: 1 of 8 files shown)


mlubin (approved):

blegat (approved):
I agree, there were many things in JuMP.jl that should have been at other places

PR #3189, opened 2023-01-12
Adding generator min/max support for NLP

michel2323:
`minimum/maximum`, and a multivariate `min/max` do not work for `NLExpressions`. This fixed that for us. Maybe I am also missing something.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -5,6 +5,8 @@
 
 _is_sum(s::Symbol) = (s == :sum) || (s == :∑) || (s == :Σ)
 _is_prod(s::Symbol) = (s == :prod) || (s == :∏)
+_is_max(s::Symbol) = (s == :max)
+_is_min(s::Symbol) = (s == :min)
 
 """
     _add_kw_args(call, kw_args)
@@ -1993,9 +1995,14 @@ function _parse_generator_expression(code, x, operators)
     y = gensym()
     y_expr, default = if _is_sum(x.args[1])
         :($y = Expr(:call, :+)), 0
-    else
-        @assert _is_prod(x.args[1])
+    elseif _is_prod(x.args[1])
         :($y = Expr(:call, :*)), 1
+    elseif _is_max(x.args[1])
+        :($y = Expr(:call, :max)), -Inf
+    elseif _is_min(x.args[1])
+        :($y = Expr(:call, :max)), Inf
+    else
+        error("Unsupported generator $(x.args[1]).")
     end
     block = _MA.rewrite_generator(
         x.args[2],
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1996,9 +1996,9 @@ function _parse_generator_expression(code, x, operators)
     elseif _is_prod(x.args[1])
         :($y = Expr(:call, :*)), 1
     elseif x.args[1] == :maximum
-        :($y = Expr(:call, :max)), -Inf
+        :($y = Expr(:call, :max)), nothing
     elseif x.args[1] == :minimum
-        :($y = Expr(:call, :min)), Inf
+        :($y = Expr(:call, :min)), nothing
     else
         error("Unsupported generator `:$(x.args[1])`")
     end
@@ -2012,11 +2012,16 @@ function _parse_generator_expression(code, x, operators)
         end,
     )
     # Special case that was handled by JuMP in the past.
+    error_string = "reducing over an empty collection in `$(x.args[1])` is not allowed"
     push!(code.args, quote
         $y_expr
         $block
         if length($y.args) == 1
-            $y = $default
+            if $default === nothing
+                throw(ArgumentError($error_string))
+            else
+                $y = $default
+            end
         end
     end)
     return y
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -5,8 +5,6 @@
 
 _is_sum(s::Symbol) = (s == :sum) || (s == :∑) || (s == :Σ)
 _is_prod(s::Symbol) = (s == :prod) || (s == :∏)
-_is_max(s::Symbol) = (s == :max)
-_is_min(s::Symbol) = (s == :min)
 
 """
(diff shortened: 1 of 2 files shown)


odow:
So just to clarify first, what is the problem, and what do you want to happen?

Note that multivariate `min` is supported:
```julia
julia> using JuMP

julia> model = Model();
A JuMP Model

julia> @variable(model, x[1:4])
4-element Vector{VariableRef}:
 x[1]
 x[2]
 x[3]
 x[4]

julia> @NLexpression(model, min(x...))
subexpression[1]: min(x[1], x[2], x[3], x[4])
```

but `min` of a generator isn't:
```julia
julia> @NLexpression(model, min(x[i] for i in 1:4))
ERROR: LoadError: AssertionError: _is_prod(x.args[1])
Stacktrace:
 [1] _parse_generator_expression(code::Expr, x::Expr, operators::Set{Tuple{Symbol, Int64}})
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:1997
 [2] _parse_nonlinear_expression_inner(code::Expr, x::Expr, operators::Set{Tuple{Symbol, Int64}})
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:1963
 [3] _parse_nonlinear_expression(model::Expr, x::Expr)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:1871
 [4] var"@NLexpression"(__source__::LineNumberNode, __module__::Module, args::Vararg{Any, N} where N)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:2465
in expression starting at REPL[5]:1
```

Although neither is it for regular arrays:
```julia
julia> y = rand(4)
4-element Vector{Float64}:
 0.25753161724346896
 0.6917793041257281
 0.222468836962886
 0.5311831189890737

julia> min(y[i] for i in 1:4)
ERROR: MethodError: no method matching min(::Base.Generator{UnitRange{Int64}, var"#3#4"})
Closest candidates are:
  min(::Any, ::Missing) at missing.jl:127
  min(::Any, ::Any) at operators.jl:433
  min(::Any, ::Any, ::Any, ::Any...) at operators.jl:560
```

We'd need to support `minimum` and `maximum`:
```julia
julia> @NLexpression(model, minimum(x[i] for i in 1:4))
ERROR: LoadError: AssertionError: _is_prod(x.args[1])
Stacktrace:
 [1] _parse_generator_expression(code::Expr, x::Expr, operators::Set{Tuple{Symbol, Int64}})
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:1997
 [2] _parse_nonlinear_expression_inner(code::Expr, x::Expr, operators::Set{Tuple{Symbol, Int64}})
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:1963
 [3] _parse_nonlinear_expression(model::Expr, x::Expr)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:1871
 [4] var"@NLexpression"(__source__::LineNumberNode, __module__::Module, args::Vararg{Any, N} where N)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:2465
in expression starting at REPL[6]:1

julia> minimum(y[i] for i in 1:4)
0.222468836962886
```

These weren't support previously, because I only recently added support for multivariate min and max: https://github.com/jump-dev/MathOptInterface.jl/pull/2023.

src/macros.jl
```diff
@@ -5,6 +5,8 @@
 
 _is_sum(s::Symbol) = (s == :sum) || (s == :∑) || (s == :Σ)
 _is_prod(s::Symbol) = (s == :prod) || (s == :∏)
+_is_max(s::Symbol) = (s == :max)
+_is_min(s::Symbol) = (s == :min)
```
odow, line 9: These should probably be `:minimum` and `:maximum`

src/macros.jl
```diff
@@ -1993,9 +1995,14 @@ function _parse_generator_expression(code, x, operators)
...
-    else
-        @assert _is_prod(x.args[1])
+    elseif _is_prod(x.args[1])
         :($y = Expr(:call, :*)), 1
+    elseif _is_max(x.args[1])
+        :($y = Expr(:call, :max)), -Inf
+    elseif _is_min(x.args[1])
+        :($y = Expr(:call, :max)), Inf
+    else
+        error("Unsupported generator $(x.args[1]).")
```
odow, line 2005: Thanks for fixing the assertion error :smile:

odow (commented):
This needs tests. 

Add new test functions for `:minimum`, `:maximum` and the unsupported generator near here:

https://github.com/jump-dev/JuMP.jl/blob/67ed28062e1c13bc398724d6c4a1ad6bec9c71cd/test/test_nlp.jl#L359-L373

test/test_nlp.jl
```diff
@@ -372,6 +372,40 @@ function test_parse_prod()
     return
 end
 
+function test_parse_minimum()
+    model = Model()
+    @variable(model, x[1:2])
+    ex = @NLexpression(model, minimum(x[i] for i in 1:2))
+    @test sprint(show, ex) == "subexpression[1]: min(x[1], x[2])"
+    ex2 = @NLexpression(model, minimum(x[i] for i in 1:0))
+    @test sprint(show, ex2) == "subexpression[2]: Inf"
```
odow, line 381: I'm open to other suggestions for what we should return here instead of `Inf`. Not sure whether it's better to throw an error here instead?

odow: Thinking on this, I don't think there's ever a case in which falling back to `Inf` can lead to a good model. It's always a modeling error, so I've changed to throwing a similar error to the one you get if you call `minimum(x for x in 1:0)`.

michel2323:
Right. I misused multivariate here. I meant generator `minimum` and `maximum`. Looks great to me. I think this is more than I wanted to happen 🙂.

odow (approved):

PR #3191, opened 2023-01-16
Add support for generic number type

blegat:
A first approach could be to change `Model` into `Model{T}` with `Model() = Model{Float64}()` and `VariableRef` into `VariableRef{T}` with `VariableRef() = VariableRef{Float64}()`.
However, this would be breaking with respect to performance. Users having a `Model` or `VariableRef` as a field would not have a concrete field anymore.
Moreover, even if it is nonbreaking, users only caring about `Float64` will have non-concrete fields when using `Model` or `VariableRef` even if these types seem concrete from the name.

For these reasons, this PR is considering a different approach.
We define `GenericModel{T}` and `const Model = GenericModel{Float64}` and `GenericVariableRef{T}` with `const VariableRef = GenericVariableRef{Float64}`.
We now have a few advantages:

* This is non-breaking
* When the users use `GenericModel` and `GenericVariableRef` as a `struct` field, it's clear from the name that it is not concrete
* It is consistent with `AffExpr`/`GenericAffExpr` and `QuadExpr`/`GenericQuadExpr`

This PR also adds the `value_type` function to the JuMP API. This function gives the return type of the `JuMP.value` function on variables of the JuMP model. It can be called on a variable or a model. Calling it on an affine or quadratic may be confusing since the coefficient of the expression might be different from the `value_type` of the variables or even the return type of `value` on the expression. It is therefore not defined for affine and quadratic expressions.

- [x] Make current tests pass
- [x] Add tests
- [x] Add docs

Note that incorporates the changes of

- [x] https://github.com/jump-dev/JuMP.jl/pull/3332

because otherwise, I would have had to complicate the tests because of this:
```julia
julia> string(Float32(1) * im)
"0.0f0 + 1.0f0im"
```

Here if how it looks like
```julia
julia> model = GenericModel{Rational{BigInt}}(CDDLib.Optimizer{Rational{BigInt}})
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: EMPTY_OPTIMIZER
Solver name: CDD

julia> @variable(model, 0 <= x[1:2])
2-element Vector{GenericVariableRef{Rational{BigInt}}}:
 x[1]
 x[2]

julia> @constraint(model, c1, 2x[1] + x[2] <= 1)
c1 : 2//1 x[1] + x[2] ≤ 1//1

julia> @constraint(model, c2, x[1] + 3x[2] <= 2)
c2 : x[1] + 3//1 x[2] ≤ 2//1

julia> @objective(model, Max, sum(x))
x[1] + x[2]

julia> optimize!(model)

julia> solution_summary(model)
* Solver : CDD

* Status
  Result count       : 1
  Termination status : OPTIMAL
  Message from the solver:
  "Optimal"

* Candidate solution (result #1)
  Primal status      : FEASIBLE_POINT
  Dual status        : NO_SOLUTION
  Objective value    : 4//5

* Work counters


julia> value.(x)
2-element Vector{Rational{BigInt}}:
 1//5
 3//5

julia> value(c1)
1//1

julia> value(c2)
2//1
```

Closes https://github.com/jump-dev/JuMP.jl/issues/2025

### DIFF
--- a/docs/src/developers/extensions.md
+++ b/docs/src/developers/extensions.md
@@ -339,14 +339,6 @@ julia> function JuMP.add_constraint(
             )
        end
 
-julia> function JuMP.model_convert(model::AbstractModel, rhs::MyConstraint)
-           return MyConstraint(
-                rhs.name,
-                model_convert(model, rhs.f),
-                model_convert(model, rhs.s),
-           )
-       end
-
 julia> @constraint(model, my_con, 2x <= 1, MyTag("my_prefix"))
 my_prefix[my_con] : 2 x - 1 ≤ 0
 ```
--- a/docs/src/developers/extensions.md
+++ b/docs/src/developers/extensions.md
@@ -339,8 +339,16 @@ julia> function JuMP.add_constraint(
             )
        end
 
+julia> function JuMP.model_convert(model::AbstractModel, rhs::MyConstraint)
+           return MyConstraint(
+                rhs.name,
+                model_convert(model, rhs.f),
+                model_convert(model, rhs.s),
+           )
+       end
+
 julia> @constraint(model, my_con, 2x <= 1, MyTag("my_prefix"))
-my_prefix[my_con] : 2 x - 1 ≤ false
+my_prefix[my_con] : 2 x - 1 ≤ 0
 ```
 
 ## The extension dictionary
--- a/docs/src/developers/extensions.md
+++ b/docs/src/developers/extensions.md
@@ -340,7 +340,7 @@ julia> function JuMP.add_constraint(
        end
 
 julia> @constraint(model, my_con, 2x <= 1, MyTag("my_prefix"))
-my_prefix[my_con] : 2 x - 1 ≤ 0
+my_prefix[my_con] : 2 x - 1 ≤ false
 ```
 
 ## The extension dictionary
--- a/docs/src/tutorials/getting_started/getting_started_with_JuMP.jl
+++ b/docs/src/tutorials/getting_started/getting_started_with_JuMP.jl
@@ -135,7 +135,7 @@ using JuMP
 using HiGHS
 
 # JuMP builds problems incrementally in a `Model` object. Create a model by
-# passing an optimizer to the [`Model`](@ref) function:
+# passing an optimizer to the [`Model`](@ref GenericModel) function:
 
 model = Model(HiGHS.Optimizer)
 
--- a/docs/src/tutorials/getting_started/performance_tips.jl
+++ b/docs/src/tutorials/getting_started/performance_tips.jl
@@ -57,7 +57,7 @@ using HiGHS  # hide
 # At present, the majority of the latency problems are caused by JuMP's bridging
 # mechanism. If you only use constraints that are natively supported by the
 # solver, you can disable bridges by passing `add_bridges = false` to
-# [`Model`](@ref).
+# [`Model`](@ref GenericModel).
 
 model = Model(HiGHS.Optimizer; add_bridges = false)
 
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -73,7 +73,17 @@ An abstract type that should be subtyped for users creating JuMP extensions.
 """
 abstract type AbstractModel end
 
-mutable struct Model <: AbstractModel
+"""
+    value_type(::Type{<:Union{AbstractModel,AbstractVariableRef}})
+
+Return the return type of [`value`](@ref) for variables of that model. It
+defaults to `Float64` if it is not implemented.
+"""
+function value_type end
+
+value_type(::Type{<:AbstractModel}) = Float64
+
+mutable struct GenericModel{T} <: AbstractModel
     # In MANUAL and AUTOMATIC modes, CachingOptimizer.
     # In DIRECT mode, will hold an AbstractOptimizer.
     moi_backend::MOI.AbstractOptimizer
@@ -84,7 +94,7 @@ mutable struct Model <: AbstractModel
     # same bridge may be added many times so we store them in a `Set` instead
     # of, e.g., a `Vector`.
     bridge_types::Set{Any}
-    # Hook into a solve call...function of the form f(m::Model; kwargs...),
+    # Hook into a solve call...function of the form f(m::GenericModel; kwargs...),
     # where kwargs get passed along to subsequent solve calls.
     optimize_hook::Any
     # TODO: Document.
@@ -105,7 +115,9 @@ mutable struct Model <: AbstractModel
     set_string_names_on_creation::Bool
 end
 
-function Base.getproperty(model::Model, name::Symbol)
+value_type(::Type{GenericModel{T}}) where {T} = T
+
+function Base.getproperty(model::GenericModel, name::Symbol)
     if name == :nlp_data
         error(
             "The internal field `.nlp_data` was removed from `Model` in JuMP " *
@@ -123,26 +135,29 @@ function Base.getproperty(model::Model, name::Symbol)
 end
 
 """
-    Model()
+    GenericModel{T}()
 
-Return a new JuMP model without any optimizer; the model is stored in
-a cache.
+Return a new JuMP model of value type `T` without any optimizer; the model is
+stored in a cache.
 
 Use [`set_optimizer`](@ref) to set the optimizer before calling
 [`optimize!`](@ref).
 """
-function Model()
+function GenericModel{T}() where {T}
     caching_opt = MOIU.CachingOptimizer(
-        MOIU.UniversalFallback(MOIU.Model{Float64}()),
+        MOIU.UniversalFallback(MOIU.Model{T}()),
         MOIU.AUTOMATIC,
     )
-    return direct_model(caching_opt)
+    return direct_model(caching_opt, T)
 end
 
-"""
-    Model(optimizer_factory; add_bridges::Bool = true)
+const Model = GenericModel{Float64}
 
-Return a new JuMP model with the provided optimizer and bridge settings.
+"""
+    GenericModel{T}(optimizer_factory; add_bridges::Bool = true) where {T}
+
+Return a new JuMP model with the provided optimizer, bridge settings and
+[`value_type`](@ref) `T`.
 
 See [`set_optimizer`](@ref) for the description of the `optimizer_factory` and
 `add_bridges` arguments.
@@ -166,18 +181,21 @@ julia> import MultiObjectiveAlgorithms as MOA
 julia> model = Model(() -> MOA.Optimizer(HiGHS.Optimizer); add_bridges = false);
 ```
 """
-function Model((@nospecialize optimizer_factory); add_bridges::Bool = true)
-    model = Model()
+function GenericModel{T}(
+    (@nospecialize optimizer_factory);
+    add_bridges::Bool = true,
+) where {T}
+    model = GenericModel{T}()
     set_optimizer(model, optimizer_factory; add_bridges = add_bridges)
     return model
 end
 
 """
-    direct_model(backend::MOI.ModelLike)
+    direct_model(backend::MOI.ModelLike, T::Type = Float64)
 
 Return a new JuMP model using [`backend`](@ref) to store the model and solve it.
 
-As opposed to the [`Model`](@ref) constructor, no cache of the model is stored
+As opposed to the [`GenericModel`](@ref) constructor, no cache of the model is stored
 outside of [`backend`](@ref) and no bridges are automatically applied to
 [`backend`](@ref).
 
@@ -188,16 +206,16 @@ in mind the following implications of creating models using this *direct* mode:
 
 * When [`backend`](@ref) does not support an operation, such as modifying
   constraints or adding variables/constraints after solving, an error is
-  thrown. For models created using the [`Model`](@ref) constructor, such
+  thrown. For models created using the [`GenericModel`](@ref) constructor, such
   situations can be dealt with by storing the modifications in a cache and
   loading them into the optimizer when `optimize!` is called.
 * No constraint bridging is supported by default.
 * The optimizer used cannot be changed the model is constructed.
 * The model created cannot be copied.
 """
-function direct_model(backend::MOI.ModelLike)
+function direct_model(backend::MOI.ModelLike, T::Type = Float64)
     @assert MOI.is_empty(backend)
-    return Model(
+    return GenericModel{T}(
         backend,
         Dict{MOI.ConstraintIndex,AbstractShape}(),
         Set{Any}(),
@@ -254,10 +272,10 @@ function direct_model(factory::MOI.OptimizerWithAttributes)
     return direct_model(optimizer)
 end
 
-Base.broadcastable(model::Model) = Ref(model)
+Base.broadcastable(model::GenericModel) = Ref(model)
 
 """
-    backend(model::Model)
+    backend(model::GenericModel)
 
 Return the lower-level MathOptInterface model that sits underneath JuMP. This
 model depends on which operating mode JuMP is in (see [`mode`](@ref)).
@@ -281,10 +299,10 @@ the innermost optimizer, see [`unsafe_backend`](@ref). Alternatively, use
 
 See also: [`unsafe_backend`](@ref).
 """
-backend(model::Model) = model.moi_backend
+backend(model::GenericModel) = model.moi_backend
 
 """
-    unsafe_backend(model::Model)
+    unsafe_backend(model::GenericModel)
 
 Return the innermost optimizer associated with the JuMP model `model`.
 
@@ -375,7 +393,7 @@ julia> highs = backend(model)  # No need to call `attach_optimizer`.
 A HiGHS model with 1 columns and 0 rows.
 ```
 """
-unsafe_backend(model::Model) = unsafe_backend(backend(model))
+unsafe_backend(model::GenericModel) = unsafe_backend(backend(model))
 
 function unsafe_backend(model::MOIU.CachingOptimizer)
     if MOIU.state(model) == MOIU.NO_OPTIMIZER
@@ -400,19 +418,19 @@ function _moi_mode(model::MOIU.CachingOptimizer)
 end
 
 """
-    mode(model::Model)
+    mode(model::GenericModel)
 
 Return the [`ModelMode`](@ref) ([`DIRECT`](@ref), [`AUTOMATIC`](@ref), or
 [`MANUAL`](@ref)) of `model`.
 """
-function mode(model::Model)
+function mode(model::GenericModel)
     # The type of `backend(model)` is not type-stable, so we use a function
     # barrier (`_moi_mode`) to improve performance.
     return _moi_mode(backend(model))
 end
 
 """
-    set_string_names_on_creation(model::Model, value::Bool)
+    set_string_names_on_creation(model::GenericModel, value::Bool)
 
 Set the default argument of the `set_string_name` keyword in the
 [`@variable`](@ref) and [`@constraint`](@ref) macros to `value`. This is used to
@@ -423,12 +441,14 @@ By default, `value` is `true`. However, for larger models calling
 `set_string_names_on_creation(model, false)` can improve performance at the cost
 of reducing the readability of printing and solver log messages.
 """
-function set_string_names_on_creation(model::Model, value::Bool)
+function set_string_names_on_creation(model::GenericModel, value::Bool)
     model.set_string_names_on_creation = value
     return
 end
 
-set_string_names_on_creation(model::Model) = model.set_string_names_on_creation
+function set_string_names_on_creation(model::GenericModel)
+    return model.set_string_names_on_creation
+end
 
 set_string_names_on_creation(::AbstractModel) = true
 
@@ -439,7 +459,7 @@ function _moi_bridge_constraints(model::MOIU.CachingOptimizer)
 end
 
 """
-    bridge_constraints(model::Model)
+    bridge_constraints(model::GenericModel)
 
 When in direct mode, return `false`.
 
@@ -448,7 +468,7 @@ optimizer is set and unsupported constraints are automatically bridged
 to equivalent supported constraints when an appropriate transformation is
 available.
 """
-function bridge_constraints(model::Model)
+function bridge_constraints(model::GenericModel)
     # The type of `backend(model)` is not type-stable, so we use a function
     # barrier (`_moi_bridge_constraints`) to improve performance.
     return _moi_bridge_constraints(backend(model))
@@ -481,7 +501,7 @@ function _moi_call_bridge_function(
 end
 
 """
-     add_bridge(model::Model, BT::Type{<:MOI.Bridges.AbstractBridge})
+     add_bridge(model::AbstractModel, BT::Type{<:MOI.Bridges.AbstractBridge})
 
 Add `BT` to the list of bridges that can be used to transform unsupported
 constraints into an equivalent formulation using only constraints supported by
@@ -489,7 +509,7 @@ the optimizer.
 
 See also: [`remove_bridge`](@ref).
 """
-function add_bridge(model::Model, BT::Type{<:MOI.Bridges.AbstractBridge})
+function add_bridge(model::GenericModel, BT::Type{<:MOI.Bridges.AbstractBridge})
     push!(model.bridge_types, BT)
     _moi_call_bridge_function(
         MOI.Bridges.add_bridge,
@@ -500,7 +520,7 @@ function add_bridge(model::Model, BT::Type{<:MOI.Bridges.AbstractBridge})
 end
 
 """
-     remove_bridge(model::Model, BT::Type{<:MOI.Bridges.AbstractBridge})
+     remove_bridge(model::GenericModel, BT::Type{<:MOI.Bridges.AbstractBridge})
 
 Remove `BT` to the list of bridges that can be used to transform unsupported
 constraints into an equivalent formulation using only constraints supported by
@@ -508,13 +528,12 @@ the optimizer.
 
 See also: [`add_bridge`](@ref).
 """
-function remove_bridge(model::Model, BT::Type{<:MOI.Bridges.AbstractBridge})
+function remove_bridge(
+    model::GenericModel{T},
+    BT::Type{<:MOI.Bridges.AbstractBridge},
+) where {T}
     delete!(model.bridge_types, BT)
-    _moi_call_bridge_function(
-        MOI.Bridges.remove_bridge,
-        backend(model),
-        BT{Float64},
-    )
+    _moi_call_bridge_function(MOI.Bridges.remove_bridge, backend(model), BT{T})
     if mode(model) != DIRECT
         MOI.Utilities.reset_optimizer(model)
     end
@@ -522,7 +541,7 @@ function remove_bridge(model::Model, BT::Type{<:MOI.Bridges.AbstractBridge})
 end
 
 """
-     print_bridge_graph([io::IO,] model::Model)
+     print_bridge_graph([io::IO,] model::GenericModel)
 
 Print the hyper-graph containing all variable, constraint, and objective types
 that could be obtained by bridging the variables, constraints, and objectives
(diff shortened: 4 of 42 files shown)


src/callbacks.jl
```diff
@@ -76,15 +76,15 @@ function MOI.submit(model::Model, cb::MOI.UserCut, con::ScalarConstraint)
 end
 
 function MOI.submit(
-    model::Model,
+    model::GenericModel{T},
     cb::MOI.HeuristicSolution,
     variables::Vector{VariableRef},
     values::Vector{<:Real},
-)
+) where {T}
```
odow, line 83: So we might have to back up here and have a bigger discussion. I don't think we ever said that that variable values were something other than `Float64`. If the coefficients are `Complex`, that doesn't mean the variables are?

odow (commented):
I think we're going to need to see a bunch of tests and examples. I don't know if I understand all the implications.

src/callbacks.jl
blegat: `T` is still a subtype of `Real`. It's typically either `Float64` or `Rational{BigInt}`. The value of the variables have type `T` and the value of the coefficients can either have type `T` or `Complex{T}`.

blegat:
> I think we're going to need to see a bunch of tests and examples. I don't know if I understand all the implications.

Yes, it's still a WIP but I opened it so that we can see that there is already one PR open with many changes all around so that we avoid making other big refactoring that will create conflicts

test/test_expr.jl
```diff
@@ -268,8 +268,8 @@ function test_extension_MA_add_mul(
     @variable(model, x)
     @variable(model, y)
     # MA.add_mul!!(ex::Number, c::Number, x::GenericAffExpr)
-    aff = MA.add_mul!!(1.0, 2.0, JuMP.GenericAffExpr(1.0, :a => 1.0))
-    @test JuMP.isequal_canonical(aff, JuMP.GenericAffExpr(3.0, :a => 2.0))
+    aff = MA.add_mul!!(1.0, 2.0, JuMP.GenericAffExpr(1.0, x => 1.0))
```
blegat, line 271: @odow Was that intentional ? I could also define `value_type(::Symbol)` to make it work

src/JuMP.jl
```diff
@@ -77,7 +77,22 @@ An abstract type that should be subtyped for users creating JuMP extensions.
...
+"""
+function value_type end
+
+value_type(::Type{<:AbstractModel}) = Float64
+
+"""
+    Model
+
+A mathematical model of an optimization problem.
+"""
```
odow, line 94: Please remove the docstring: https://github.com/jump-dev/JuMP.jl/pull/3198

src/copy.jl
```diff
@@ -36,11 +36,11 @@ reference of the copied model can be obtained by indexing the map with the
 reference of the corresponding reference of the original model.
 """
 struct ReferenceMap
-    model::Model
+    model::GenericModel
```
odow, line 39:
```suggestion
    model::GenericModel{T}
```
Should this be a concrete type?

src/feasibility_checker.jl
```diff
@@ -168,48 +168,52 @@ function _add_infeasible_nonlinear_constraints(
...
-function _distance_to_set(::Any, set::MOI.AbstractSet)
+function _distance_to_set(::Any, set::MOI.AbstractSet, ::Type)
     return error(
         "Feasibility checker for set type $(typeof(set)) has not been " *
         "implemented yet.",
     )
 end
 
-_distance_to_set(::Missing, ::MOI.AbstractSet) = 0.0
+_distance_to_set(::Missing, ::MOI.AbstractSet, ::Type{T}) where {T} = zero(T)
```
odow, line 178: So the type argument is only needed because of `missing`? Perhaps we could just return `missing` here and then catch it above.

src/macros.jl
```diff
@@ -195,13 +195,13 @@ moved to the set with `MOIU.shift_constant`.
 """
 function operator_to_set end
 
-operator_to_set(::Function, ::Union{Val{:(<=)},Val{:(≤)}}) = MOI.LessThan(0.0)
+operator_to_set(::Function, ::Union{Val{:(<=)},Val{:(≤)}}) = MOI.LessThan(false)
```
odow, line 198: Ouch

test/test_expr.jl
odow: I didn't change the content of the tests. So I assumed it was intentional. I didn't track the original introduction.

odow (commented):
I haven't reviewed fully; this is a big PR. I'm a little nervous about making such a big change to JuMP at this point.

One intermediate option that I think we've discussed before is to take data as input in Float64 and to convert it to the type `T` before passing to the solver. Then we get the benefit of solving in higher fidelity, whilst keeping JuMP simple.

blegat:
> One intermediate option that I think we've discussed before is to take data as input in Float64 and to convert it to the type `T` before passing to the solver. Then we get the benefit of solving in higher fidelity, whilst keeping JuMP simple.

There are a few advantages of supporting arbitrary types at the JuMP level:

* For callbacks, both for MIP and NLP solvers, having to convert to `Float64` will give rounding errors in `BigFloat` and will be a complete blocker for `Rational{BigInt}` (rational does not make sense for NLP callbacks but it does for MIP callbacks).
* Sometimes, you data are known to a very high precision and you would like to describe it precisely as well: this is an example https://github.com/blegat/SwitchOnSafety.jl/blob/master/examples/Finiteness_conjecture_counterexample.ipynb
* Sometimes, you use JuMP in a loop where you get solution, then you do something with it, then modify the model (or create a new one based on the solution) and then solve again. Having to convert to `Float64` in between is problematic.
* When you solve a reational LP, it would be nice to directly get the rational solution without having to call `rationalize` and to be able to communicate rational coefficients without having the solver to rationalize.
* Some bridges multiply by irrational numbers etc..., if these are done in `Float64` instead of `BigFloat`, you already loose precision.

I'd also argue that this PR keeps JuMP simple. There are a lot of changes but it's mostly renamings from `Model` to `GenericModel` and `VariableRef` to `GenericVariableRef`.  Users and even JuMP extension don't even have to do this if they don't care about non-`Float64`. So basically if you only care about `Float64`, you won't see a change.

src/macros.jl
blegat: It's similar to what `LinearAlgebra` is doing with `UniformScaling`. Using `Bool` at first and them promoting is a smooth path.
Whenever an expression sees a variable, it automatically promotes to `value_type` though.
So `x + 1` is still `AffExpr`, it does not build `GenericAffExpr{Int}` since otherwise we could loose a lot of time converting and hence allocating.

src/JuMP.jl
blegat: Oops, the rebases were a bit hard

src/copy.jl
blegat: Good catch

blegat: Fixed

test/test_constraint.jl
```diff
@@ -539,17 +555,18 @@ function test_extension_SDP_errors(
...
     @variable(model, z)
     @variable(model, w)
-    aff_str = "$(GenericAffExpr{Float64,VariableRefType})"
+    aff_str = "$AffExprType"
     err = ErrorException(
         "In `@constraint(model, [x 1; 1 -y] >= [1 x; x -2], PSDCone(), unknown_kw = 1)`:" *
         " Unrecognized constraint building format. Tried to invoke " *
         "`build_constraint(error, $(aff_str)[x - " *
-        "1 -x + 1; -x + 1 -y + 2], $(MOI.GreaterThan(0.0)), $(PSDCone()); unknown_kw = 1)`, but no " *
+        "1 -x + 1; -x + 1 -y + 2], $(MOI.GreaterThan(false)), $(PSDCone()); unknown_kw = 1)`, but no " *
```
odow, line 569: What's going on here? `GreaterThan(false)`?

test/test_macros.jl
```diff
@@ -1516,10 +1524,10 @@ end
 
 function test_MA_Zero_expression()
     model = Model()
-    @test @expression(model, sum(i for i in 1:0)) === 0.0
+    @test @expression(model, sum(i for i in 1:0)) === false
```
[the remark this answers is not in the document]
odow: Ooof. I'm not sure what the right return is here. I think we should be able to see `model` in the macro and return the right `zero(value_type(model))`? There's a `_replace_zero(::MA.Zero) = 0.0` function somewhere.

test/test_macros.jl
```diff
@@ -1510,16 +1520,16 @@ end
...
     @expression(model, expr[j = 1:2], sum(i for i in j:0))
-    @test expr == [0.0, 0.0]
-    @test expr isa Vector{Float64}
+    @test expr == [false, false]
+    @test expr isa Vector{Bool}
```
odow, line 1532: Yeah I think this is wrong, and arguably breaking. I know a bunch of people use this behavior (e.g., PowerModels and PowerSimulations).

test/test_feasibility_checker.jl
```diff
@@ -11,92 +11,92 @@ using Test
 function test_unsupported()
     @test_throws(
         ErrorException,
-        JuMP._distance_to_set([1.0, 1.0], MOI.Complements(2)),
+        JuMP._distance_to_set([1.0, 1.0], MOI.Complements(2), Float64),
```
odow, line 14: This shouldn't need the type. We can infer it from the type of the input point.

src/variables.jl
```diff
@@ -4,11 +4,12 @@
...
 
 """
-    num_variables(model::Model)::Int64
+    num_variables(model::GenericModel)::Int64
 
 Returns number of variables in `model`.
 """
-num_variables(model::Model)::Int64 = MOI.get(model, MOI.NumberOfVariables())
+num_variables(model::GenericModel)::Int64 =
+    MOI.get(model, MOI.NumberOfVariables())
```
odow, line 12: Weird that JuliaFormatter didn't rewrite this as a proper function?

src/variables.jl
```diff
@@ -195,22 +196,35 @@ with variables of type `V<:AbstractVariableRef` and coefficients of type `T`
...
+value_type(::Type{<:AbstractVariableRef}) = Float64
+
 Base.conj(v::AbstractVariableRef) = v
 Base.real(v::AbstractVariableRef) = v
 Base.imag(v::AbstractVariableRef) = zero(v)
 Base.abs2(v::AbstractVariableRef) = v^2
 
 """
-    VariableRef <: AbstractVariableRef
+    GenericVariableRef <: AbstractVariableRef
```
odow, line 210:
```suggestion
    GenericVariableRef{T} <: AbstractVariableRef
```

src/sets.jl
```diff
@@ -220,17 +222,19 @@ See [here](http://lpsolve.sourceforge.net/5.5/SOS.htm) for a description of SOS
...
-    function SOS2(weights::AbstractVector = Float64[])
-        return new(convert(Vector{Float64}, weights))
+struct SOS2{T} <: AbstractVectorSet
+    weights::Vector{T}
+    function SOS2{T}(weights::AbstractVector = T[]) where {T}
+        return new{T}(convert(Vector{T}, weights))
     end
 end
-function moi_set(set::SOS2, dim::Int)
+SOS2() = SOS2{Int}()
```
odow, line 231: I think this is breaking. Because solvers support `MOI.SOS2{Float64}`, not `MOI.SOS2{Int}`.

src/macros.jl
```diff
@@ -1222,11 +1278,11 @@ end
 
 Replaces `_MA.Zero` with a floating point `0.0`.
 """
-_replace_zero(::_MA.Zero) = 0.0
+_replace_zero(::_MA.Zero) = false
```
odow, line 1281: I think we can fix this part to support returning the right `zero(value_type(model))`.

src/macros.jl
```diff
@@ -689,25 +685,80 @@ function build_constraint(
...
     ::Function,
-    x::AbstractVector{<:AbstractJuMPScalar},
+    x::AbstractVector{T},
     set::MOI.SOS2,
-)
-    return VectorConstraint(x, MOI.SOS2{Float64}(set.weights))
+) where {T<:AbstractJuMPScalar}
+    return VectorConstraint(
+        x,
+        MOI.SOS2{value_type(variable_ref_type(T))}(set.weights),
```
odow, line 714: Oh. Messy. The `SOS2{Int}` gets converted here? That's confusing.

odow (commented):
Heading in the right direction. I don't really know how we're going to test this in its entirety.

src/sets.jl
blegat: This is going to get converted in `model_convert` in the macro

blegat:
@odow Any ideas where the
```
 Error: reference for '`GenericModel`' could not be found in src/tutorials/getting_started/performance_tips.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/wMjrb/src/utilities/utilities.jl:44
┌ Error: no doc found for reference '[`ntoh`](@ref)' in src/reference/models.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/wMjrb/src/utilities/utilities.jl:44
┌ Error: no doc found for reference '[`ltoh`](@ref)' in src/reference/models.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/wMjrb/src/utilities/utilities.jl:44
┌ Error: no doc found for reference '[`readchomp`](@ref)' in src/reference/models.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/wMjrb/src/utilities/utilities.jl:44
┌ Error: reference for '`GenericModel`' could not be found in src/tutorials/getting_started/getting_started_with_JuMP.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/wMjrb/src/utilities/utilities.jl:44
```
are coming from ?

docs/src/tutorials/getting_started/performance_tips.jl
```diff
@@ -57,7 +57,7 @@ using HiGHS  # hide
 # At present, the majority of the latency problems are caused by JuMP's bridging
 # mechanism. If you only use constraints that are natively supported by the
 # solver, you can disable bridges by passing `add_bridges = false` to
-# [`Model`](@ref).
+# [`Model`](@ref `GenericModel`).
```
odow, line 60:
```suggestion
# [`Model`](@ref GenericModel).
```

@blegat: haven't tested, but I think one error is this.

docs/src/tutorials/getting_started/getting_started_with_JuMP.jl
```diff
@@ -135,7 +135,7 @@ using JuMP
 using HiGHS
 
 # JuMP builds problems incrementally in a `Model` object. Create a model by
-# passing an optimizer to the [`Model`](@ref) function:
+# passing an optimizer to the [`Model`](@ref `GenericModel`) function:
```
odow, line 138:
```suggestion
# passing an optimizer to the [`Model`](@ref GenericModel) function:
```

odow:
The other errors are because of 
```
Base.write(::IO, ::Model; ::MOI.FileFormats.FileFormat)
Base.read(::IO, ::Type{Model}; ::MOI.FileFormats.FileFormat)
```
in `docs/src/reference.models.md`. These aren't the type signatures anymore, so Documenter tries to bring in the generic `Base.copy` docstring.

src/macros.jl
```diff
@@ -706,55 +712,48 @@ function build_constraint(
...
 function build_constraint(
-    _error::Function,
-    v::AbstractJuMPScalar,
+    ::Function,
+    v::Union{AbstractJuMPScalar},
     set::MOI.AbstractScalarSet,
 )
     return ScalarConstraint(v, set)
 end
+function _clear_constant!(expr::Union{GenericAffExpr,GenericQuadExpr})
```
odow, line 721:
```suggestion

function _clear_constant!(expr::Union{GenericAffExpr,GenericQuadExpr})
```

src/macros.jl
```diff
@@ -706,55 +712,48 @@ function build_constraint(
...
+    offset = constant(expr)
+    add_to_expression!(expr, -offset)
+    return expr, offset
+end
+function _clear_constant!(α::Number)
```
odow, line 726:
```suggestion

function _clear_constant!(α::Number)
```

src/macros.jl
```diff
@@ -706,55 +712,48 @@ function build_constraint(
...
+    return zero(α), α
+end
 function build_constraint(
```
odow, line 729:
```suggestion

function build_constraint(
```

src/macros.jl
```diff
@@ -706,55 +712,48 @@ function build_constraint(
...
-)
-    return build_constraint(_error, convert(AffExpr, α), set)
-end
 function build_constraint(
     _error::Function,
     ::_MA.Zero,
     set::MOI.AbstractScalarSet,
 )
-    return build_constraint(_error, zero(AffExpr), set)
+    return build_constraint(_error, false, set)
```
odow, line 747: Here's the culprit! I don't know why this method exists. But it seems horrific.

src/macros.jl
```diff
@@ -843,25 +845,80 @@ function build_constraint(
...
+
+function model_convert(model, α::Number)
+    T = value_type(typeof(model))
+    V = variable_ref_type(model)
+    C = _complex_convert_type(T, typeof(α))
+    return convert(GenericAffExpr{C,V}, α)
+end
+
+function model_convert(model, con::BridgeableConstraint)
+    return BridgeableConstraint(
```
odow, line 899: This method isn't tested.

src/macros.jl
```diff
@@ -843,25 +845,80 @@ function build_constraint(
...
+
+function model_convert(model, con::VectorConstraint)
+    return VectorConstraint(
+        model_convert.(model, con.func),
+        model_convert(model, con.set),
+        con.shape,
+    )
+end
+
+model_convert(model, con) = con
```
odow, line 920: This method isn't tested.

src/macros.jl
odow: See `_functionize` comment.

src/nlp.jl
```diff
@@ -61,16 +61,16 @@ end
...
-)
+) where {T}
     sum_id = model.operators.multivariate_operator_to_id[:+]
     prod_id = model.operators.multivariate_operator_to_id[:*]
     call_multi = MOI.Nonlinear.NODE_CALL_MULTIVARIATE
     n_terms = length(x.terms) + !iszero(x.constant)
     if n_terms == 0
         # If `x` is empty, then substitute `0.0` as the expression.
-        MOI.Nonlinear.parse_expression(model, expr, 0.0, parent)
+        MOI.Nonlinear.parse_expression(model, expr, zero(T), parent)
```
odow, line 73: So far, `MOI.Nonlinear` is hard-coded to assume `Float64`.

src/nlp.jl
```diff
@@ -207,9 +211,9 @@ function MOI.Nonlinear.parse_expression(
     return MOI.Nonlinear.parse_expression(model, expr, index, parent)
 end
 
-function add_nonlinear_parameter(model::Model, value::Real)
+function add_nonlinear_parameter(model::GenericModel{T}, value::Real) where {T}
     _init_NLP(model)
-    p = MOI.Nonlinear.add_parameter(model.nlp_model, Float64(value))
+    p = MOI.Nonlinear.add_parameter(model.nlp_model, convert(T, value))
```
odow, line 216: Ditto. parameters support only `Float64`.

src/sets.jl
```diff
@@ -193,17 +193,19 @@ they should be unique values. The *k*th element in the set corresponds to the
...
-    function SOS1(weights::AbstractVector = Float64[])
-        return new(convert(Vector{Float64}, weights))
+struct SOS1{T} <: AbstractVectorSet
+    weights::Vector{T}
+    function SOS1{T}(weights::AbstractVector = T[]) where {T}
+        return new{T}(convert(Vector{T}, weights))
     end
 end
-function moi_set(set::SOS1, dim::Int)
+SOS1() = SOS1{Int}()
```
odow, line 202: Add a comment that `Int` is chosen as a placeholder, and the true type is converted when adding to the model.

odow:
Here's the `extension-tests.yml` job: https://github.com/jump-dev/JuMP.jl/actions/runs/4941377839

odow:
The PowerModels failure seems unrelated and just a numerical issue. But this is breaking for InfiniteOpt.

odow:
Seems like InfiniteOpt also broken with https://github.com/jump-dev/JuMP.jl/pull/3350, so this might not be the cause.

odow:
We're not the culprit: https://github.com/infiniteopt/InfiniteOpt.jl/issues/315. But we should hold off at minimum until InfiniteOpt is back passing tests to see if anything else is hidden in the test failures.

docs/src/developers/extensions.md
```diff
@@ -340,7 +340,7 @@ julia> function JuMP.add_constraint(
        end
 
 julia> @constraint(model, my_con, 2x <= 1, MyTag("my_prefix"))
-my_prefix[my_con] : 2 x - 1 ≤ 0
+my_prefix[my_con] : 2 x - 1 ≤ false
```
joaquimg, line 343: why `false`?

src/lp_sensitivity2.jl
```diff
@@ -8,24 +8,24 @@
...
+    objective::Dict{GenericVariableRef{T},Tuple{T,T}}
 end
 
 Base.getindex(s::SensitivityReport, c::ConstraintRef) = s.rhs[c]
-Base.getindex(s::SensitivityReport, x::VariableRef) = s.objective[x]
+Base.getindex(s::SensitivityReport, x::GenericVariableRef) = s.objective[x]
 
 """
-    lp_sensitivity_report(model::Model; atol::Float64 = 1e-8)::SensitivityReport
+    lp_sensitivity_report(model::GenericModel{T}; atol::T = Base.rtoldefault(T))::SensitivityReport{T} where {T}
```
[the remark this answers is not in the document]
blegat: It is almost the same for `Float64`
```
julia> Base.rtoldefault(Float64)
1.4901161193847656e-8
```

src/macros.jl
blegat: I guess it get's called when you have an empty sum or something like that. It will be converted by `model_convert` before getting passed to `JuMP.add_constraint`

src/lp_sensitivity2.jl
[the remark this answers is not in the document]
odow: :+1: Oh I thought it was much smaller

src/macros.jl
```diff
@@ -373,7 +373,7 @@ function _functionize(
 end
 
 _functionize(x) = x
-_functionize(::_MA.Zero) = 0.0
+_functionize(::_MA.Zero) = false
```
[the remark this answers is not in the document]
blegat: There is no way to make this work without breaking the `build_constraint` API. I think it's find, `build_constraint` should only depends on the expression, not on the model so it's ok for it to not be `Float64`-specific before it goes through `model_convert`

src/constraints.jl
```diff
@@ -577,8 +577,10 @@ representing the function and the `set` field contains the MOI set.
 See also the [documentation](@ref Constraints) on JuMP's representation of
 constraints for more background.
 """
-struct ScalarConstraint{F<:AbstractJuMPScalar,S<:MOI.AbstractScalarSet} <:
-       AbstractConstraint
+struct ScalarConstraint{
+    F<:Union{Number,AbstractJuMPScalar},
```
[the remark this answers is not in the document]
blegat: Ok so this is needed for this test for instance:
https://github.com/jump-dev/JuMP.jl/blob/2d666cb01857701e7442cee844ba00ee83b4dca9/test/test_constraint.jl#L103
Before, we would just create an `AffExpr` but in `build_constraint`, we cannot build an `AffExpr` since we don't know what the variable type is since it depends on the value type and we don't have any variable nor model.
Actually, building an `AffExpr` is not so extension friendly either, e.g., the `JuMPExtension` also has a different variable type so I think it does make sense to allow `Number` here.

test/test_macros.jl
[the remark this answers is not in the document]
odow: @blegat this is still a breaking change.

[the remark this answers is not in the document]
odow: It's an easy fix. Incoming once I fix the test changes

test/test_constraint.jl
```diff
@@ -1597,6 +1620,9 @@ function test_semiinteger()
 end
 
 function test_symmetric_vectorize_allocations()
+    if VERSION < v"1.8"
+        return
+    end
```
odow, line 1625: Intended?

src/variables.jl
```diff
@@ -377,36 +413,40 @@ function delete(model::Model, variable_refs::Vector{VariableRef})
...
-function VariableRef(m::Model)
-    index = MOI.add_variable(backend(m))
-    return VariableRef(m, index)
+function GenericVariableRef{T}(model::GenericModel{T}) where {T}
+    index = MOI.add_variable(backend(model))
+    return GenericVariableRef{T}(model, index)
+end
+
+function GenericVariableRef(model::GenericModel{T}) where {T}
+    return GenericVariableRef{T}(model)
```
odow, line 449: Why is this method needed? Not covered by tests.

odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/5041405190

test/test_constraint.jl
blegat: Yes, this allocated more for Julia v1.6 but couldn't figure out why.

blegat:
SDDP.jl failure is due to https://github.com/jump-dev/JuMP.jl/pull/3341

blegat:
> Perhaps we can just break extensions with constraints with no variables...

We can transform a `ScalarConstraint{<:Number}` into a `ScalarConstraint{AffExpr}` in `model_convert` if we're worried about that

blegat:
I would still do it on the output of `build_constraint`. If we move it to `add_constraint` it would be quite breaking for JuMP extensions defining their own `add_constraint`.

blegat:
Thinking more about it, I'd be fine with both options. Simpler macro code is good

src/constraints.jl
```diff
@@ -682,15 +684,16 @@ function _moi_add_constraint(
...
 
 Add a constraint `con` to `Model model` and sets its name.
 """
 function add_constraint(
-    model::Model,
+    model::GenericModel,
     con::AbstractConstraint,
     name::String = "",
 )
+    con = model_convert(model, con)
```
odow, line 696: @blegat: it was sufficient to fix the extension in the docs by adding this here. This means some extra allocations though, so we should probably figure out when we don't need to `model_convert` objects?

odow:
New run with more packages: https://github.com/jump-dev/JuMP.jl/actions/runs/5051033339

src/feasibility_checker.jl
```diff
@@ -74,10 +74,10 @@ end
 """
     primal_feasibility_report(
         point::Function,
-        model::Model;
-        atol::Float64 = 0.0,
+        model::GenericModel{t};
```
odow, line 77:
```suggestion
        model::GenericModel{T};
```

odow:
So what if we split this up into a few different PRs? 

 1. Add `GenericModel` and all of the simple replacements. But don't actually test or declare support for anything other than `GenericModel{Float64}`
 2. Add `GenericVariableRef`
 3. Add `value_type(::Type{AbstractModel})` and update the coefficients in tests
 4. Add `model_convert` and add tests for arbitrary number types

Then 1, 2, and 3 are all definitely non-breaking. And we can focus on 4 as a source of the issue.

joaquimg:
>What other packages should we test?

solvers like Alpine

joaquimg:
Since it is a big change, we can even consider releasing a beta version and asking in discourse for people to report.

odow:
Am I brave enough to try rebasing this on top of #3378?

odow:
See https://github.com/jump-dev/JuMP.jl/pull/3385 for a replacement.

pulsipher:
> New run with more packages: https://github.com/jump-dev/JuMP.jl/actions/runs/5051033339

InfiniteOpt's master is now up to date and passing tests, should you want to retest.

odow:
Closing in favor of #3385

Issue #3192, opened 2023-01-19
Precompilation in the SnoopPrecompile/Julia 1.9+ world


timholy:
Because of its popularity, I'd like to use JuMP as one of several "showcases" for the impact of pkgimages in Julia 1.9 (CC @vchuravy, @vtjnash, @KristofferC). It turns out that to really showcase this work, JuMP might need a few tweaks. Normally I just submit PRs, but due to the "orthogonality" of solvers, JuMP presents some interesting challenges, and so I decided to instead open this issue.

JuMP and its ecosystem have had a lot of nice work done on invalidations and precompilation already, and these lay the foundation and make everything I'm about to show much easier. (Thanks!) The main remaining gap is due, I think, to the fact that the precompilation work occurred before the arrival of SnoopPrecompile and pkgimages in Julia 1.9. 

Let me begin by showing that there's opportunity for substantial further improvement. All tests were conducted on a reasonably up-to-date Julia `master`:
```julia
julia> @time using JuMP, GLPK
  7.286941 seconds (9.35 M allocations: 604.153 MiB, 4.49% gc time, 0.79% compilation time)

julia> @time @eval begin
               let
                   model = Model(GLPK.Optimizer)
                   @variable(model, x >= 0)
                   @variable(model, 0 <= y <= 3)
                   @objective(model, Min, 12x + 20y)
                   @constraint(model, c1, 6x + 8y >= 100)
                   @constraint(model, c2, 7x + 12y >= 120)
                   optimize!(model)
               end
           end;
  8.465601 seconds (10.10 M allocations: 1.385 GiB, 4.05% gc time, 99.78% compilation time)
```
Now, let me create a new package, `StartupJuMP`, purely for the purpose of extra precompilation. (That's a viable strategy on Julia 1.8 and higher, with the big impact arriving in Julia 1.9.) The source code is a single file, `src/StartupJuMP.jl`, with contents:
```julia
module StartupJuMP

using GLPK
using JuMP
using SnoopPrecompile

@precompile_all_calls begin
    # Because lots of the work is done by macros, and macros are expanded
    # at lowering time, not much of this would get precompiled without `@eval`
    @eval begin
        let
            model = Model(GLPK.Optimizer)
            @variable(model, x >= 0)
            @variable(model, 0 <= y <= 3)
            @objective(model, Min, 12x + 20y)
            @constraint(model, c1, 6x + 8y >= 100)
            @constraint(model, c2, 7x + 12y >= 120)
[...]

PR #3193, opened 2023-01-19
Add SnoopPrecompile workload

timholy:
On my machine, this cuts the TTFX of a simple workload (using the
GLPK solver, which is not used here) from about 8.5s to about 3s.

The improvement was measured on Julia `master`; there may be some (much smaller) improvement for Julia 1.8 as well, but I haven't tested. The workload is the one in #3192, identical to the one in the `@precompile_all_calls` block except using GLPK as the solver.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,12 +9,14 @@ MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 MutableArithmetics = "d8a4904e-b15c-11e9-3269-09a3773c0cb0"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+SnoopPrecompile = "66db9d55-30c0-4569-8b51-7e840670fc0c"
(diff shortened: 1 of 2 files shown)


timholy:
I'm not certain what you want to do about your current `_precompile_` workload. Perhaps the best option might be to expand this simple workload to accomodate those cases?

timholy:
> Perhaps the best option might be to expand this simple workload to accommodate those cases?

To clarify, I'm not familiar with what one needs to do to set up a problem with, e.g., a `MOI.RotatedSecondOrderCone` constraint:
https://github.com/jump-dev/JuMP.jl/blob/a978323d95799ee04aa1eeb9bd5d5ce663a07215/src/precompile.jl#L29

My "showcase" for Julia 1.9 will not need breadth of coverage, but ideally someone who knows JuMP will eventually expand this to cover more cases and force "better" precompilation of all the things covered in the current `precompile.jl` infrastructure. At which point you could remove `precompile.jl` entirely.

timholy:
I'm not aware of anything wrong with merging this. The other things are "nice to have" but someone could tackle those later.

odow (approved):
We can add this approach to the solver packages, using only the MOI interface. It'll still miss some of the JuMP -> Solver methods, but it should give a nice reduction. I'll have a play.

PR #3198, opened 2023-01-24
Fix docstring of Model not appearing in docs

odow:
x-ref https://discourse.julialang.org/t/add-bridges/93478/2

Upstream issue is https://github.com/JuliaDocs/Documenter.jl/issues/396. Easiest just to remove the docstring on the struct.

- [x] Check preview before merging


odow:
Fixed

image

Issue #2227, opened 2020-04-21
Improve support for constraint programming


dourouc05:
I'm making this new issue to track the progress on CP support in JuMP. (Brief reminder: started with https://github.com/JuliaOpt/JuMP.jl/issues/2014; then, a few changes in https://github.com/JuliaOpt/JuMP.jl/pull/2051 to make more constraints parse.) I'm opening this issue to discuss the progress on this side (maybe this should be marked as 1.0?). 

For now, these things have been implemented: 

- Several CP sets, an extension to MOI: [ConstraintProgrammingExtensions.jl](https://github.com/dourouc05/ConstraintProgrammingExtensions.jl)
- New constraints in JuMP to create CP constraints easily: [JuCP.jl](https://github.com/dourouc05/JuCP.jl/); this absolutely requires https://github.com/JuliaOpt/JuMP.jl/pull/2051 to be merged
- A first solver is partially wrapped: [CPLEXCP.jl](https://github.com/dourouc05/CPLEXCP.jl)

There are many more details about all this in [JuCP](https://github.com/dourouc05/JuCP.jl/blob/master/README.md), including design considerations and next steps. Some of these steps could be made into GSoC projects (dealing with tests, writing new wrappers, mostly), if it's not too late [for inclusion](https://github.com/jump-dev/GSOC2020/blob/master/ideas-list.md). 

-----------------

More complete list of things: 

- On MOI's side: 
  - [x] Discuss the best way to deal with functions: https://github.com/jump-dev/MathOptInterface.jl/issues/846#issuecomment-816252103
- On JuMP's side: 
  - [x] Strict inequality: already possible. https://github.com/dourouc05/JuCP.jl/blob/0c2d0f359e882baf4ff06d404e47d9ee542b05e0/src/sets.jl#L51-L53
  - [x] New syntax (functions as constraints): `@constraint(m, constraint(x, y))`.  PR #2051. 
  - [x] New syntax (reification): `@constraint(m, x := { constraint(x, y) })`, `@constraint(m, x := { x < y })`.  PR #2228 (split from #2051). 
  - [ ] New syntax (new expressions in very limited cases): `@constraint(m, x == count(y .== 1))`. PR #2229 or PR #2241 (split from #2051). 
  - [ ] New syntax (new expressions in less limited cases): `@constraint(m, x == y[z])` with either `y` or `z` variables. No PR for now. Is an overload of `getindex` enough? 
[...]

PR #3206, opened 2023-01-30
Add <--> syntax support for reified constraints

odow:
x-ref #2227

So far, MiniZinc.jl is the only solver supporting `MOI.Reified`, but we've discussed adding this syntax in #2227.


odow:
Hmm. This passes tests locally:
```raw
(base) oscar@Oscars-MBP JuMP % julia --project=. test/runtests.jl test_reified.jl
[ Info: Loading test_reified.jl
Test Summary:         | Pass  Total
test_reified.jl:test_ |    8      8
Test Summary:                   |
test_reified.jl:test_extension_ | No tests
```

blegat:
I like the idea of `{ ... }`, it's consistent with indicator constraints. How standard is `:=` though ? Could we make `<=>` and `⇔` work as well ?

odow:
We can't make `<=>` work because it doesn't parse:
```julia
julia> :(z <=> {a in b})
ERROR: syntax: space before "{" not allowed in "> {" at REPL[1]:1
Stacktrace:
 [1] top-level scope
   @ none:1
```

`\iff` works, but that's not very helpful without an ascii fallback.
```julia
julia> :(z ⟺ {a in b})
:(z ⟺ {a in b})
```

odow:
Ahh. But this is available:
```Julia
julia> :(a <--> b)
:(a <--> b)

julia> dump(:(a <--> b))
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol <-->
    2: Symbol a
    3: Symbol b
```

chriscoey:
`<-->` looks OK. Out of curiosity, is there a list of available Julia operator symbols somewhere? Where did you find `<-->`?

odow:
This isn't a great answer, but the list of symbols is in the parser:

https://github.com/JuliaLang/julia/blob/9c563ad6a4b5568e07d88776c42dd5c47fabc45b/src/julia-parser.scm#L13

I originally found it via trial-and-error...

chriscoey:
oo, so many arrows. would be nice if the implication constraint and reified constraint arrows are consistent e.g. `-->` and `<-->`

odow:
> would be nice if the implication constraint and reified constraint arrows are consistent e.g. --> and <-->

Yes, we could add `-->` support.

mlubin:
Very cool, I didn't expect this would be so easy!

odow:
> Very cool, I didn't expect this would be so easy!

Yes! https://github.com/jump-dev/JuMP.jl/issues/2227 has spurred quite a few changes that made this easy.

blegat (approved):

PR #3208, opened 2023-02-01
Add fix_discrete_variables for computing the dual of a MIP

odow:
This comes up quite often on discourse etc, and @jd-lara has written custom code to do this in PowerSimulations etc.

x-ref: https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-1029439667

Open to bike-shed. I considered making this a separate `relax_integrality_and_fix` function instead of the keyword argument, but the overlap between the two methods is quite large, so it seemed to make more sense as an option.

Docs: https://jump.dev/JuMP.jl/previews/PR3208/tutorials/linear/mip_duality/

### DIFF
--- a/docs/src/tutorials/linear/mip_duality.jl
+++ b/docs/src/tutorials/linear/mip_duality.jl
@@ -85,26 +85,26 @@ unfix.(dispatch)
 set_binary.(dispatch)
 print(model)
 
-# ## Use `relax_integrality`
+# ## Use `fix_discrete_variables`
 
 # Manually choosing the variables to relax and fix works for our small example,
 # but it becomes more difficult in problems with a larger number of binary and
 # integer variables. To automate the process we just did manually, JuMP provides
-# a `fix` keyword argument to [`relax_integrality`](@ref) function:
+# the [`fix_discrete_variables`](@ref) function:
 
 optimize!(model)
 dual_status(model)
 
 #-
 
-undo = relax_integrality(model; fix = value);
+undo = fix_discrete_variables(model);
 
 # Here `undo` is a function that, when called with no arguments, returns the
 # model to the original mixed-integer formulation.
 
 # !!! tip
-#     Afer calling [`relax_integrality`](@ref), you can set a new solver with
-#     [`set_optimizer`](@ref) if your mixed-integer solver does not support
+#     Afer calling [`fix_discrete_variables`](@ref), you can set a new solver
+#     with [`set_optimizer`](@ref) if your mixed-integer solver does not support
 #     computing a dual solutio.
 
 print(model)
--- a/docs/src/tutorials/linear/mip_duality.jl
+++ b/docs/src/tutorials/linear/mip_duality.jl
@@ -67,6 +67,7 @@ dual_status(model)
 discrete_values = value.(dispatch)
 fix.(dispatch, discrete_values; force = true)
 unset_binary.(dispatch)
+print(model)
 
 # Now if we re-solve the problem, we obtain a `FEASIBLE_POINT` for the dual:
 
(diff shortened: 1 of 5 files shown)


docs/src/tutorials/linear/mip_duality.jl
```diff
@@ -0,0 +1,124 @@
...
+
+undo = relax_integrality(model; fix = value);
+
+# Here `undo` is a function that, when called with no arguments, returns the
+# model to the original mixed-integer formulation.
+
+# !!! tip
+#     Afer calling [`relax_integrality`](@ref), you can set a new solver with
+#     [`set_optimizer`](@ref) if your mixed-integer solver does not support
+#     computing a dual solutio.
```
odow, line 108:
```suggestion
#     computing a dual solution.
```

odow:
What about a new function like `fix_discrete_variables`?

Gurobi calls this `fixmodel`: https://www.gurobi.com/documentation/9.5/refman/c_fixmodel.html

mlubin:
`fix_discrete_variables` sgtm

src/variables.jl
```diff
@@ -1543,10 +1543,22 @@ function _info_from_variable(v::VariableRef)
 end
 
 """
-    relax_integrality(model::Model)
+    relax_integrality(model::Model; fix::Union{Nothing,Function} = nothing)
 
 Modifies `model` to "relax" all binary and integrality constraints on
-variables. Specifically,
+variables. In addition, if `fix` is passed a function, then the discrete
+variables are fixed to the value of `fix(x)`.
```
blegat, line 1550: If `fix` returns `nothing` then the variable is not fixed or it throws an error ?

odow: Rewritten as two separate functions.

src/variables.jl
```diff
@@ -1591,7 +1591,67 @@ Subject to
  x binary
 ```
 """
-function relax_integrality(model::Model)
+relax_integrality(model::Model) = _relax_or_fix_integrality(nothing, model)
+
+"""
+    fix_discrete_variables([var_value::Function = value,] model::Mode)
```
mlubin, line 1597:
```suggestion
    fix_discrete_variables([var_value::Function = value,] model::Model)
```

mlubin (approved):

src/variables.jl
```diff
@@ -1591,7 +1591,67 @@ Subject to
...
+julia> print(model)
+Min x + y
+Subject to
+ y ≥ 1.0
+ y ≤ 10.0
+ y integer
+ x binary
+```
+"""
+function fix_discrete_variables(var_value::Function, model::Model)
```
joaquimg, line 1645: dont know if the `::Function` was on purpose, but it will force nonspecialize: https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing
maybe:
```julia
function fix_discrete_variables(var_value::F, model::Model) where {F<:Function}
```

src/variables.jl
```diff
@@ -1591,7 +1591,67 @@ Subject to
...
+    return _relax_or_fix_integrality(var_value, model)
+end
+
+fix_discrete_variables(model::Model) = fix_discrete_variables(value, model)
+
+function _relax_or_fix_integrality(
+    var_value::Union{Nothing,Function},
```
joaquimg, line 1652: we might want to specialize here at least since the function is called a lot

odow: Have you seen this be a performance issue on large models?

odow:
> This comes up quite often on discourse etc

Speak of the devil: https://discourse.julialang.org/t/jump-support-for-solver-specific-methods-to-create-fixed-mip/94775

PR #3209, opened 2023-02-02
Use MOI.Utilities.distance_to_set in primal_feasibility_report

odow:
Added in MOI v1.11.0


blegat (approved):
Nice, that helps for https://github.com/jump-dev/JuMP.jl/pull/3191#discussion_r1091181594

PR #3216, opened 2023-02-13
Update to knapsack tutorial

jd-foster:
Updating the knapsack problem tutorial from just a function to the "newer" format.
Tracking: #2348.


docs/src/tutorials/linear/knapsack.jl
```diff
@@ -3,41 +3,134 @@
...
+# selected items.
+@objective(model, Max, sum(profit[i] * x[i] for i in 1:5))
+
+# Let's print a human-readable description of the model and 
+# check that the model looks as expected:
+print(model)
+
+# We can now solve the optimization problem and inspect the results.
+optimize!(model)
+solution_summary(model)
```
odow, line 89:
```suggestion
solution_summary(model)

# The items chosen are

items_chosen = [i for i in 1:5 if value(x[i]) > 0.5]
```

PR #3218, opened 2023-02-14
New conic example: ellipsoid approximation

jd-foster:
Adding this, partly inspired by this question on discourse:
https://discourse.julialang.org/t/optimization-jl-and-constraints/94519/8

Still needs linking within the navigation.


blegat (approved):
Nice, thanks!

odow:
This file needs to be added to `make.jl`

PR #3219, opened 2023-02-14
Add get_attribute and set_attribute

joaquimg:
We currently have `get_optimizer_attribute` and `set_optimizer_attribute`, which are specialized for `MOI.AbstractOptimizerAttribute`. However, there is no JuMP-only way to get or set `MOI.AbstractModelAttribute`, `MOI.AbstractVariableAttribute`, or `MOI.AbstractConstraintAttribute`.

This PR adds `get_attribute` and `set_attribute` to provide a JuMP-level API for all `MOI.get` and `MOI.set` (including `MOI.AbstractOptimizerAttribute`). 

Because we can't break existing code, we keep `[gs]et_optimizer_attribute`, but the docstrings have been modified to include a warning that users should prefer `[gs]et_attribute`, and the `_optimizer_attribute` version may be removed in a future breaking release of JuMP.


src/optimizer_interface.jl
```diff
@@ -720,6 +720,13 @@ function MOI.get(
     return _moi_get_result(backend(model), attr, index(v))
 end
 
+function get_variable_attribute(
```
odow, line 723: I think we either need `get_variable_attribute(model, attr, x)` or `get_variable_attribute(x, attr)`.

There's also an argument that we should just overload `get_optimizer_attribute(x, attr)` and  `set_optimizer_attribute(x, attr, value)`.

src/optimizer_interface.jl
```diff
@@ -720,6 +720,13 @@ function MOI.get(
...
+    attr::MOI.AbstractVariableAttribute,
+    v::VariableRef,
+)
+    MOI.get(owner_model(v), attr)
```
odow:
```suggestion
    return MOI.get(owner_model(v), attr)
```

src/optimizer_interface.jl
```diff
@@ -740,6 +747,13 @@ function MOI.get(
     return _moi_get_result(backend(model), attr, index(cr))
 end
 
+function get_constraint_attribute(
+    attr::MOI.AbstractConstraintAttribute,
+    cr::ConstraintRef,
+)
+    MOI.get(owner_model(cr), attr)
```
odow:
```suggestion
    return MOI.get(owner_model(cr), attr)
```

src/optimizer_interface.jl
```diff
@@ -772,6 +794,14 @@ function MOI.set(
     return MOI.set(backend(model), attr, index(cr), value)
 end
 
+function set_constraint_attribute(
+    attr::MOI.AbstractConstraintAttribute,
+    cr::ConstraintRef,
+    value,
+)
+    MOI.set(owner_model(cr), attr, value)
```
odow:
```suggestion
    MOI.set(owner_model(cr), attr, value)
    return
```

src/optimizer_interface.jl
```diff
@@ -761,6 +775,14 @@ function MOI.set(
     return MOI.set(backend(model), attr, index(v), value)
 end
 
+function set_variable_attribute(
+    attr::MOI.AbstractVariableAttribute,
+    v::VariableRef,
+    value,
+)
+    MOI.set(owner_model(v), attr, value)
```
odow:
```suggestion
    MOI.set(owner_model(v), attr, value)
    return
```

src/optimizer_interface.jl
joaquimg: > I think we either need get_variable_attribute(model, attr, x) or get_variable_attribute(x, attr).

omitting the model was suggested in: https://github.com/jump-dev/ParametricOptInterface.jl/issues/121#issuecomment-1430395250

will switch to `get_variable_attribute(x, attr)`

> There's also an argument that we should just overload get_optimizer_attribute(x, attr) and set_optimizer_attribute(x, attr, value).

I understand, I am not sure which is better. Totaly to open to change,
Since we have: `MOI.AbstractOptimizerAttribute` we could reserve `get_optimizer_attribute` for it.
And  `MOI.AbstractModelAttribute` would match with `get_model_attribute`.

odow: I don't think users really care about the distinction of optimizer/model/variable/constraint. 

At the JuMP level, `get_optimizer_attribute` could just mean "query something from the bottom-most solver"

joaquimg: I'd be fine with that.

joaquimg: maybe `get_attribute` for all? since `get_optimizer_attribute(x, attr)` without the model parameters might be slightly weird?

odow: This feels like a naming question for @mlubin

blegat: I prefer not adding in the name something that is in the signature so I would prefer `get_attribute` and `set_attribute`

docs/src/manual/solutions.md
```diff
@@ -524,7 +524,7 @@ conflict, iterate over the list of constraints and query the
 list_of_conflicting_constraints = ConstraintRef[]
 for (F, S) in list_of_constraint_types(model)
     for con in all_constraints(model, F, S)
-        if MOI.get(model, MOI.ConstraintConflictStatus(), con) == MOI.IN_CONFLICT
+        if get_attribute(model, MOI.ConstraintConflictStatus(), con) == MOI.IN_CONFLICT
```
odow, line 527:
```suggestion
        if get_attribute(con, MOI.ConstraintConflictStatus()) == MOI.IN_CONFLICT
```

src/optimizer_interface.jl
```diff
@@ -772,6 +774,61 @@ function MOI.set(
...
+function get_attribute(cr::ConstraintRef, attr::MOI.AbstractConstraintAttribute)
+    return MOI.get(owner_model(cr), attr, cr)
+end
+
+"""
+    set_attribute(model::Model, attr::MOI.AbstractModelAttribute, value)
+    set_attribute(x::VariableRef, attr::MOI.AbstractVariableAttribute, value)
+    set_attribute(cr::ConstraintRef, attr::MOI.AbstractConstraintAttribute, value)
+
+Set an MOI attribute from in the model.
```
odow, line 804:
```suggestion
Set an MOI attribute in the model.
```

src/optimizer_interface.jl
odow: Updated to use `get/set_attribute`. I think the `getindex` feels a bit too cute? `model[MOI.Silent()] = true`? `x[MOI.VariableName]`? I guess that makes sense, but it doesn't feel very JuMP-like. (More Pyomo-like.)

joaquimg: I like `get/set_attribute`. It is very consistent with everything in JuMP.

>  I think the getindex feels a bit too cute? model[MOI.Silent()] = true? x[MOI.VariableName]? I guess that makes sense, but it doesn't feel very JuMP-like. (More Pyomo-like.)

I agree and also have mixed feelings. We can consider that and bikeshed, I would just leave it for another PR/issue and merge this one as it seems a nice improvement.

odow:
One current issue is what to do with `get_optimizer_attribute`. We can't delete it, but we could remove it from the documentation in favor of `get_attribute` instead. Conceptually, having a single interface for attributes is useful, but changing the API is also a drag on users.

joaquimg:
Yeah, we are in 1.x so I agree `get_optimizer_attribute` has to stay.

I am not sure we should remove it from docs since someone might be using it, and trying to find it in the docs. We could update its docs and **strongly recommend** `get_attribute` instead, maybe even say its legacy and should be removed in the future 2.x version.

docs/src/reference/models.md
```diff
@@ -40,16 +40,16 @@ set_string_names_on_creation
...
 set_attribute
 set_silent
 unset_silent
 set_time_limit_sec
 unset_time_limit_sec
 time_limit_sec
+get_optimizer_attribute
+set_optimizer_attribute
+set_optimizer_attributes
 ```
```
joaquimg, line 53: maybe even:

```suggestion
` ` `

### Legacy
` ` `@docs
get_optimizer_attribute
set_optimizer_attribute
set_optimizer_attributes
` ` `
```
(added the space to that preview compiles properly...)

odow: I added a `!!! compat` warning to the docstrings. I don't know if we need to make an explicit `Legacy` section.

mlubin:
The current proposal looks reasonable. Could you add a short PR description?

odow:
> Could you add a short PR description?

Done.

src/optimizer_interface.jl
blegat: We already have that for getting your variables from they name with `model[:x]`. We're not aiming for cute syntax at the MOI level but we can't say that it's not the purpose of JuMP. `getindex`, `setindex!` is a good way to have a nice and consistent syntax for these kind of things.

odow: I guess there's a difference between things that JuMP/the model owns, like `:x`, and attributes like `MOI.Silent`. Regardless, can we leave this for a later discussion? See how people use this first?

blegat: But then we might get the same issue that `get_optimizer_attribute` that we now want to remove but it's breaking.
I guess here we could support both syntax

odow: > But then we might get the same issue

I don't think we're ever going to want people to set options using `model["output_flag"] = "off"`?

blegat:
I think there are a few options to consider:

1. It seems we allowed `set_optimizer_attributes(model, attr1 => value1, attr2 => value2)` but we don't want to do it for `set_attributes(model, attr1 => value1, attr2 => value2)`, why not ?
2. `set_attribute(model, ::String, value)` is the reason we had `set_optimizer_attributes` I think. Maybe keep `set_optimizer_attributes` for strings ?

odow:
1. I figured `set_attribute.(model, [attr1 => value1, attr2 => value2])` is just as easy, and leverages existing Julia syntax. Can add if we think `set_attributes` is useful.
2. `set_attribute(model, ::String, value)` has an unambiguous interpretation, so I don't think we need `set_optimizer_attribute`. There's no equivalent `RawVariableAttribute`, and even if there was, we can distinguish based on the type of the first argument.

blegat:
> I figured `set_attribute.(model, [attr1 => value1, attr2 => value2])` is just as easy, and leverages existing Julia syntax. Can add if we think `set_attributes` is useful.

Yes, but we agreed on the oppose before JuMP v1.0 so why would we come to a different conclusion now ?

odow:
I will add

blegat (approved):

odow:
@joaquimg are you happy to merge?

PR #3222, opened 2023-02-17
Use and document preference for import MathOptInterface as MOI

odow:
Related to https://github.com/jump-dev/MathOptInterface.jl/pull/2096.

When I tried `import MutableArithmetics as MA`, it exposed a hygiene issue that I'm currently working on.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -19,30 +19,20 @@ module JuMP
 
 import Base.Meta: isexpr, quot
 import LinearAlgebra
-import MathOptInterface
+import MathOptInterface as MOI
 import MutableArithmetics
 import OrderedCollections
 import OrderedCollections: OrderedDict
 import Printf
 import SparseArrays
 
+# We can't use import MutableArithmetics as _MA because of a bug in MA.
 const _MA = MutableArithmetics
 
-"""
-    MOI
+export MOI  # Explicitly export this
 
-Shorthand for the MathOptInterface package.
-"""
-const MOI = MathOptInterface
-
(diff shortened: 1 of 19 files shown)


src/JuMP.jl
```diff
@@ -19,30 +19,20 @@ module JuMP
...
 import LinearAlgebra
-import MathOptInterface
+import MathOptInterface as MOI
 import MutableArithmetics
 import OrderedCollections
 import OrderedCollections: OrderedDict
 import Printf
 import SparseArrays
 
+# We can't use import MutableArithmetics as _MA because of a bug in MA.
```
odow, line 29: This will be fixed by https://github.com/jump-dev/MutableArithmetics.jl/pull/199

test/nlp_solver.jl
```diff
@@ -23,8 +23,7 @@
 
 using Ipopt, JuMP
 using Test
-using MathOptInterface
-const MOI = MathOptInterface
+using MathOptInterface as MOI
```
odow, line 26:
```suggestion
import MathOptInterface as MOI
```

src/JuMP.jl
```diff
@@ -19,22 +19,16 @@ module JuMP
...
```
odow, line 29:
```suggestion
# We can't use import MutableArithmetics as _MA because of a bug in MA.
# Fixed in MutableArithmetics v1.2.3, but would require bumping the compat
# bound so we can keep it as this until necessary.
```

blegat (approved):

PR #3226, opened 2023-02-20
Fix init kwarg in nonlinear generator expressions

### LINKED ISSUE #3224: Error with semicolon/sum in nonlinear constraint
I recently hit the error mentioned [here](https://discourse.julialang.org/t/error-reducing-over-an-empty-collection-is-not-allowed/47410) and have been updating all my sums to use `; init = 0` instead of `, init = 0` to guard against empty generators. When I make this change to sums in `@NLconstraint`s, I get "UndefVarError: init not defined". Is this intended behavior? If so, I think the error message could be more clear.

Interestingly, `sum(generator, init = 0)` seems to work in NLconstraint even when the generator is empty.

I'm using JuMP 1.8.0. Code to reproduce:
```julia
using JuMP
m = Model()
@variable(m, v[1:3])
@NLconstraint(m, nlcon, sum(v[i] for i in 1:3; init = 0) == 1)
```

odow:
Closes #3224

### DIFF
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -1797,6 +1797,12 @@ function test_nonlinear_generator_init_sum()
     y = 3
     a = @NLexpression(model, sum(x for i in 1:0; init = y))
     @test string(a) == "subexpression[4]: 3.0"
+    a = @NLexpression(model, sum(x for i in 1:0, j in 1:0; init = 1))
+    @test string(a) == "subexpression[5]: 1.0"
+    a = @NLexpression(model, sum(x for i in 1:2, j in 1:2; init = 3))
+    @test string(a) == "subexpression[6]: x + x + x + x"
+    a = @NLexpression(model, sum(x for i in 1:2, j in 1:0; init = 3))
+    @test string(a) == "subexpression[7]: 3.0"
     return
 end
 
@@ -1830,6 +1836,17 @@ function test_nonlinear_generator_init_min()
     return
 end
 
+function test_nonlinear_generator_bad_init()
+    model = Model()
+    @variable(model, x)
+    expr = :(sum((x for i = 1:1); bad_init = 3))
+    @test_macro_throws(
+        ErrorException("Unsupported nonlinear expression: $expr"),
+        @NLexpression(model, sum(x for i in 1:1; bad_init = 3))
+    )
+    return
+end
+
 #!format: off
 
 function test_nonlinear_generator_pos_init_sum()
@@ -1844,12 +1861,6 @@ function test_nonlinear_generator_pos_init_sum()
     y = 3
     a = @NLexpression(model, sum(x for i in 1:0, init = y))
     @test string(a) == "subexpression[4]: 3.0"
-    a = @NLexpression(model, sum(x for i in 1:0, j in 1:0; init = 1))
-    @test string(a) == "subexpression[5]: 1.0"
-    a = @NLexpression(model, sum(x for i in 1:2, j in 1:2; init = 3))
-    @test string(a) == "subexpression[6]: x + x + x + x"
-    a = @NLexpression(model, sum(x for i in 1:2, j in 1:0; init = 3))
-    @test string(a) == "subexpression[7]: 3.0"
     return
 end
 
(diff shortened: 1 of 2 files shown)


test/test_macros.jl
```diff
@@ -1785,4 +1785,115 @@ function test_variable_Bool_argument()
...
+    @test string(a) == "subexpression[3]: 3.0"
+    a = @NLexpression(model, minimum(x for i in 1:0; init = 2))
+    @test string(a) == "subexpression[4]: 2.0"
+    return
+end
+
+function test_nonlinear_generator_bad_init()
+    model = Model()
+    @variable(model, x)
+    expr = :(sum((x for i = 1:1); bad_init = 3))
```
odow, line 1842:
```suggestion
    expr = :(sum((x for i in 1:1); bad_init = 3))
```

test/test_macros.jl
```diff
@@ -1785,4 +1785,115 @@ function test_variable_Bool_argument()
...
+    @test string(a) == "subexpression[2]: 0.0"
+    a = @NLexpression(model, sum(x for i in 1:0; init = 1))
+    @test string(a) == "subexpression[3]: 1.0"
+    y = 3
+    a = @NLexpression(model, sum(x for i in 1:0; init = y))
+    @test string(a) == "subexpression[4]: 3.0"
+    a = @NLexpression(model, sum(x for i in 1:0, j in 1:0; init = 1))
+    @test string(a) == "subexpression[5]: 1.0"
+    a = @NLexpression(model, sum(x for i in 1:2, j in 1:2; init = 3))
+    @test string(a) == "subexpression[6]: x + x + x + x"
```
odow, line 1803: So one issue here: it isn't obvious what to do with a non-empty expression and an `init`. The Julia docs say:
```julia
    sum(itr; [init])

  Returns the sum of all elements in a collection.

  The return type is Int for signed integers of less than system word size, and UInt for unsigned integers of less than system word size. For all other arguments, a common return type is found to which
  all arguments are promoted.

  The value returned for empty itr can be specified by init. It must be the additive identity (i.e. zero) as it is unspecified whether init is used for non-empty collections.
```
so it's undefined behavior.

At the REPL, it currently does:
```Julia
julia> sum(i for i in 1:3; init = 2)
8

julia> minimum(i for i in 1:3; init = 2)
1

julia> minimum(i for i in 1:3; init = -2)
-2
```

odow: cc @robbybp

test/test_macros.jl
```diff
@@ -1785,4 +1785,121 @@ function test_variable_Bool_argument()
...
+    @test string(a) == "subexpression[2]: 0.0"
+    a = @NLexpression(model, sum(x for i in 1:0; init = 1))
+    @test string(a) == "subexpression[3]: 1.0"
+    y = 3
+    a = @NLexpression(model, sum(x for i in 1:0; init = y))
+    @test string(a) == "subexpression[4]: 3.0"
+    a = @NLexpression(model, sum(x for i in 1:0, j in 1:0; init = 1))
+    @test string(a) == "subexpression[5]: 1.0"
+    a = @NLexpression(model, sum(x for i in 1:2, j in 1:2; init = 3))
+    @test string(a) == "subexpression[6]: 3.0 + x + x + x + x"
```
odow, line 1803: I've updated, so now it is this. I don't know which is better. My guess is that people would probably expect this behavior, even if the docstring says otherwise?

blegat (approved):

test/test_macros.jl
Robbybp: If, according to the docs, `init` must be the additive identify, maybe we should raise an error otherwise.

odow: The annoying part is that `Base` currently works:
```julia
julia> sum(i for i in 1:3; init = 2)
8
```

Robbybp: The undefined behavior is unfortunate. This behavior of `init` is what I would expect, however.

odow: So happy to merge this for now? We can always add a warning later if the undefined behavior in `Base` changes or is confusing.

Robbybp: Yes, happy to merge. Thank for fixing.

Robbybp (approved):

PR #3227, opened 2023-02-21
[docs] add multiobjective portfolio example

odow:
x-ref: https://discourse.julialang.org/t/jump-non-linear-optimization/94020/10

Another example on discourse that multi-objective is going to get used more than I perhaps expected.

Preview: http://jump.dev/JuMP.jl/previews/PR3227/tutorials/nonlinear/portfolio/


odow:
I think this looks good

image

blegat:
As a follow up from https://github.com/jump-dev/MathOptInterface.jl/pull/2070#issuecomment-1404755386, it's nice that it works so smoothly.
I believe this is called the Markowitz efficient frontier: https://en.wikipedia.org/wiki/Efficient_frontier, that's how the Convex.jl example calls it, we could add that.
It might be nice adding a comment why MOA only returns a finite number of solutions while we expect an infinite number.

PR #3228, opened 2023-02-21
Prep for v1.8.1

odow:
Wait for 

 - [x] #3227 
 - [x] #3229 
 - [x] #3231


odow:
Merged because this is a trivial release. It's just a single fix plus some documentation.

PR #3231, opened 2023-02-22
[docs] add tutorial on parallelism

odow:
x-ref #2348

Not really sure how this will work in CI...

https://jump.dev/JuMP.jl/previews/PR3231/tutorials/algorithms/parallelism/


odow:
So I think the issue is https://github.com/JuliaDocs/IOCapture.jl/issues/14. I need to look into this more

blegat (approved):
Nice addition

odow:
I've set this up so that it doesn't run in CI. Instead, I ran locally and then copy-pasted the results. I'll take a look into the interaction between Literate, Documenter, and multithreading, but it seems like a hard problem to solve.

PR #2524, opened 2021-03-22
Allow slicing of DenseAxisArray

odow:
DenseAxisArrays with `OneTo` axes used a dictionary that mapped `i => i`. This PR replaces that with a `Base.OneTo` for improved efficiency. But in doing so, I realized we can now define `get` for vectors, and support proper slicing of `DenseAxisArray`.

Closes #287

PR #3237, opened 2023-02-23
[Containers] add support for Base.getindex(::Container; kwargs...)

### LINKED ISSUE #1291: Strongly typed indices
This is more of a Julia ecosystem issue than a JuMP issue, but it would be nice to have better syntax to do the following:

```julia
item_count = 3
factory_count = 3
struct ItemIndex
    value::Int
end

struct FactoryIndex
    value::Int
end

item_set = [ItemIndex(i) for i in 1:item_count]
factory_set = [FactoryIndex(i) for i in 1:factory_count]

m = Model()
@variable(m, item_is_produced_at_factory[item_set, factory_set], Bin)
item_is_produced_at_factory[ItemIndex(1), FactoryIndex(1)] # Ok
item_is_produced_at_factory[FactoryIndex(1), ItemIndex(1)] # Error
```

odow:
Closes #1291

There are still a few issues to work through:

 - [x] documentation
 - [x] DenseAxisView
 - [x] Slicing needs to preserve names
 - [x] What if broadcasting mixes names? Answer: we take the first set of names. We can't force equal names because that would be breaking.

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -1057,7 +1057,7 @@ JuMP.Containers.SparseAxisArray{ConstraintRef{Model, MathOptInterface.Constraint
     If you have many index dimensions and a large amount of sparsity, read
     [Performance considerations](@ref).
 
-### Forcing the container type
+### [Forcing the container type](@id constraint_forcing)
 
 When creating a container of constraints, JuMP will attempt to choose the
 tightest container type that can store the constraints. However, because this
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -1057,7 +1057,7 @@ JuMP.Containers.SparseAxisArray{ConstraintRef{Model, MathOptInterface.Constraint
     If you have many index dimensions and a large amount of sparsity, read
     [Performance considerations](@ref).
 
-### [Forcing the container type](@id constraint_forcing)
+### Forcing the container type
 
 When creating a container of constraints, JuMP will attempt to choose the
 tightest container type that can store the constraints. However, because this
--- a/docs/src/manual/containers.md
+++ b/docs/src/manual/containers.md
@@ -518,7 +518,7 @@ julia> Containers.@container(
            i + j,
            enable_keyword_indexing = true,
        )
-SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 6 entries:
+JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 6 entries:
   [2, 1]  =  3
   [3, 1]  =  4
   [3, 2]  =  5
@@ -527,7 +527,7 @@ SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 6 entries:
   [4, 3]  =  7
 
 julia> y[i = 2:4, j = 1]
-SparseAxisArray{Int64, 1, Tuple{Int64}} with 3 entries:
+JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 3 entries:
   [2]  =  3
   [3]  =  4
   [4]  =  5
--- a/docs/src/manual/containers.md
+++ b/docs/src/manual/containers.md
@@ -488,22 +488,20 @@ JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 2 entries:
 
 ## Keyword indexing
 
-Opt-in to the experimental keyword indexing syntax by setting
-`Containers.ENABLE_KEYWORD_INDEXING` to true:
+Opt-in to the experimental keyword indexing syntax by passing
+`enable_keyword_indexing = true` to the [`Containers.@container`](@ref) macro,
+or call [`enable_container_keyword_indexing`](@ref).
+
+Keyword indexing lets you uses the named indices of [`Containers.DenseAxisArray`](@ref)
+and [`Containers.SparseAxisArray`](@ref) as keyword arguments when indexing a
+container:
 
 ```jldoctest container_kwarg_indexing
-julia> using JuMP.Containers
-
-julia> Containers.ENABLE_KEYWORD_INDEXING[] = true
-true
-```
-
-Keyword indexing lets you uses the named indices of
-[`Containers.DenseAxisArray`](@ref) and [`Containers.SparseAxisArray`](@ref)
-as keyword arguments when indexing a container:
-
-```jldoctest container_kwarg_indexing
-julia> Containers.@container(x[i=2:4, j=1:3], i + j)
+julia> Containers.@container(
+           x[i=2:4, j=1:3],
+           i + j,
+           enable_keyword_indexing = true,
+       )
 2-dimensional DenseAxisArray{Int64,2,...} with index sets:
     Dimension 1, 2:4
     Dimension 2, Base.OneTo(3)
@@ -515,7 +513,11 @@ And data, a 3×3 Matrix{Int64}:
 julia> x[i = 2, j = 1]
 3
 
-julia> Containers.@container(y[i=2:4, j=1:3; i > j], i + j)
+julia> Containers.@container(
+           y[i=2:4, j=1:3; i > j],
+           i + j,
+           enable_keyword_indexing = true,
+       )
 SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 6 entries:
   [2, 1]  =  3
   [3, 1]  =  4
@@ -529,13 +531,47 @@ SparseAxisArray{Int64, 1, Tuple{Int64}} with 3 entries:
   [2]  =  3
   [3]  =  4
   [4]  =  5
+
+julia> using JuMP
+
+julia> model = Model();
+
+julia> @variable(model, x[i = 2:3])
+1-dimensional DenseAxisArray{VariableRef,1,...} with index sets:
+    Dimension 1, 2:3
+And data, a 2-element Vector{VariableRef}:
+ x[2]
+ x[3]
+
+julia> x[i = 2]
+ERROR: Keyword indexing is disabled. To enable, pass `enable_keyword_indexing = true` to the `Containers.@container` macro, or call `JuMP.enable_container_keyword_indexing(model, true)` before calling any JuMP macros like `@variable`.
+Stacktrace:
+[...]
+
+julia> model = Model();
+
+julia> enable_container_keyword_indexing(model, true)
+
+julia> @variable(model, x[i = 2:3])
+1-dimensional DenseAxisArray{VariableRef,1,...} with index sets:
+    Dimension 1, 2:3
+And data, a 2-element Vector{VariableRef}:
+ x[2]
+ x[3]
+
+julia> x[i = 2]
+x[2]
 ```
 
-The main reason this syntax is currently experimental is that it does not work
-for `Array`
+The main reason this syntax is currently opt-in is that it does not work for
+`Array`:
 
 ```jldoctest container_kwarg_indexing
-julia> Containers.@container(z[i=1:4, j=1:3], i + j)
+julia> Containers.@container(
+           z[i=1:4, j=1:3],
+           i + j,
+           enable_keyword_indexing = true,
+       )
 4×3 Matrix{Int64}:
  2  3  4
  3  4  5
--- a/docs/src/manual/containers.md
+++ b/docs/src/manual/containers.md
@@ -259,6 +259,22 @@ julia> DataFrames.DataFrame(table)
    4 │     2  B       (2, :B)
 ```
 
+### Keyword indexing
+
+Keyword indexing can be used if all of the indices are named:
+
+```jldoctest containers_dense
+julia> x[i = 1, j = :A]
+(1, :A)
+
+julia> x[i = 1:2, j = :A]
+1-dimensional DenseAxisArray{Tuple{Int64, Symbol},1,...} with index sets:
+    Dimension 1, [1, 2]
+And data, a 2-element Vector{Tuple{Int64, Symbol}}:
+ (1, :A)
+ (2, :A)
+```
+
 ## SparseAxisArray
 
 A [`Containers.SparseAxisArray`](@ref) is created when the index sets are
@@ -350,6 +366,20 @@ julia> DataFrames.DataFrame(table)
    2 │     2  A       (2, :A)
    3 │     2  B       (2, :B)
    4 │     3  A       (3, :A)
+```
+
+### Keyword indexing
+
+Keyword indexing can be used if all of the indices are named:
+
+```jldoctest containers_sparse
+julia> x[i = 2, j = :A]
+(2, :A)
+
+julia> x[i = 2:3, j = :A]
+SparseAxisArray{Tuple{Int64, Symbol}, 1, Tuple{Int64}} with 2 entries:
+  [2]  =  (2, :A)
+  [3]  =  (3, :A)
 ```
 
 ## Forcing the container type
--- a/docs/src/manual/containers.md
+++ b/docs/src/manual/containers.md
@@ -488,15 +488,19 @@ JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 2 entries:
 
 ## Keyword indexing
 
-Opt-in to the experimental keyword indexing syntax by passing
-`enable_keyword_indexing = true` to the [`Containers.@container`](@ref) macro,
-or call [`enable_container_keyword_indexing`](@ref).
-
+JuMP v1.10.0 added experimental support for keyword indexing of
+[`Containers.DenseAxisArray`](@ref) and [`Containers.SparseAxisArray`](@ref).
 Keyword indexing lets you uses the named indices of [`Containers.DenseAxisArray`](@ref)
 and [`Containers.SparseAxisArray`](@ref) as keyword arguments when indexing a
-container:
+container.
 
-```jldoctest container_kwarg_indexing
+### The container macro
+
+For containers constructed using the [`Containers.@container`](@ref) macro,
+opt-in to the keyword indexing syntax by passing
+`enable_keyword_indexing = true`:
+
+```jldoctest
 julia> Containers.@container(
            x[i=2:4, j=1:3],
            i + j,
@@ -531,7 +535,14 @@ JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 3 entries:
   [2]  =  3
   [3]  =  4
   [4]  =  5
+```
 
+### The JuMP macros
+
+For containers constructed using the JuMP macros like [`@variable`](@ref),
+opt-in using [`enable_container_keyword_indexing`](@ref):
+
+```jldoctest
 julia> using JuMP
 
 julia> model = Model();
@@ -563,12 +574,14 @@ julia> x[i = 2]
 x[2]
 ```
 
-The main reason this syntax is currently opt-in is that it does not work for
+### Limitations
+
+The keyword indexing syntax is currently opt-in because it does not work for
 `Array`:
 
-```jldoctest container_kwarg_indexing
+```jldoctest
 julia> Containers.@container(
-           z[i=1:4, j=1:3],
+           x[i=1:4, j=1:3],
            i + j,
            enable_keyword_indexing = true,
        )
@@ -578,7 +591,36 @@ julia> Containers.@container(
  4  5  6
  5  6  7
 
-julia> z[i = 2, j = 1]
+julia> x[i = 2, j = 1]
 ERROR: MethodError: no method matching getindex(::Matrix{Int64}; i=2, j=1)
 [...]
+```
+
+Work-around this limitation by forcing the container type:
+
+```jldoctest
+julia> model = Model();
+
+julia> enable_container_keyword_indexing(model, true)
+
+julia> @variable(model, x[i = 1:3])
+3-element Vector{VariableRef}:
+ x[1]
+ x[2]
+ x[3]
+
+julia> x[i = 2]
+ERROR: MethodError: no method matching getindex(::Vector{VariableRef}; i=2)
+[...]
+
+julia> @variable(model, y[i = 1:3], container = DenseAxisArray)
+1-dimensional DenseAxisArray{VariableRef,1,...} with index sets:
+    Dimension 1, Base.OneTo(3)
+And data, a 3-element Vector{VariableRef}:
+ y[1]
+ y[2]
+ y[3]
+
+julia> y[i = 2]
+y[2]
 ```
--- a/docs/src/manual/containers.md
+++ b/docs/src/manual/containers.md
@@ -540,7 +540,7 @@ JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 3 entries:
 ### The JuMP macros
 
 For containers constructed using the JuMP macros like [`@variable`](@ref),
-opt-in using [`enable_container_keyword_indexing`](@ref):
+opt-in using [`enable_keyword_indexing`](@ref):
 
 ```jldoctest
 julia> using JuMP
@@ -555,13 +555,13 @@ And data, a 2-element Vector{VariableRef}:
  x[3]
 
 julia> x[i = 2]
-ERROR: Keyword indexing is disabled. To enable, pass `enable_keyword_indexing = true` to the `Containers.@container` macro, or call `JuMP.enable_container_keyword_indexing(model, true)` before calling any JuMP macros like `@variable`.
+ERROR: Keyword indexing is disabled. To enable, pass `enable_keyword_indexing = true` to the `Containers.@container` macro, or call `JuMP.enable_keyword_indexing(model, true)` before calling any JuMP macros like `@variable`.
 Stacktrace:
 [...]
 
 julia> model = Model();
 
-julia> enable_container_keyword_indexing(model, true)
+julia> enable_keyword_indexing(model, true)
 
 julia> @variable(model, x[i = 2:3])
 1-dimensional DenseAxisArray{VariableRef,1,...} with index sets:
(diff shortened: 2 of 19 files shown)


test/Containers/test_SparseAxisArray.jl
```diff
@@ -247,4 +247,61 @@ function test_ambuguity_BroadcastStyleUnknown()
     return
 end
 
+function test_containers_sparseaxisarray_kwarg_indexing()
+    Containers.@container(
+        x[i = 2:3, j = 1:2],
+        i + j,
+        container = SparseAxisArray
+    )
+    for i in (2, 3, 2:2, 2:3, :), j in (1, 2, 1:2, 1:1, 2:2, :)
```
joaquimg, line 256: Any reason why `1:2` is not here?

metab0t:
Maybe `DenseAxisView` should also be adapted to support `@view A[i=1, j=:]`?

Furthermore, what do you think about the following convenient features:
- Remove the restrictions for orders of kwargs `A[i=1, j=2] == A[j=2, i=1]`
- Automatic inserting `:` for missing dimensions `A[i=2:3] == A[i=2:3, j=:]`

odow:
> what do you think about the following convenient features:

I think they're a little too magical. This syntax is opt-in and more verbose, so it will be used by people looking for correctness and to make the code more readable. Enforcing an order of arguments helps clarify that that, and so does forcing arguments like `j = :`.

docs/src/manual/containers.md
```diff
@@ -352,6 +368,20 @@ julia> DataFrames.DataFrame(table)
...
+### Keyword indexing
+
+Keyword indexing can be used if all of the indices are named:
+
+```jldoctest containers_sparse
+julia> x[i = 2, j = :A]
+(2, :A)
+
+julia> x[i = 2:3, j = :A]
+SparseAxisArray{Tuple{Int64, Symbol}, 1, Tuple{Int64}} with 2 entries:
```
odow, line 380:
```suggestion
JuMP.Containers.SparseAxisArray{Tuple{Int64, Symbol}, 1, Tuple{Int64}} with 2 entries:
```

blegat:
We don't have a good answer for fixing the usability issue of a user doing
```julia
model = Model()
@variable(model, x[i=1:2, j = 1:3])
x[i = 1, j = 3] # errors as `x` is a `Matrix`.
```
We have many users that don't care about which type of container is returned by macro and it this kind of erros would be very confusing.

What we need to do is to return a `DenseAxisArray` if there are names in the macro.
If a user really need to have an `Array` and not a `DenseAxisArray`, this user cares about container so we can ask him to do
```julia
@variable(model, x[i=1:2, j = 1:3], lower_bound = i + j, requested_container = Array)
```
However, that is breaking so maybe we need to wait for JuMP v2.0 for this PR ?

odow:
It's confusing, but because it's opt-in, the user can also opt-in to `container = DenseAxisArray`.

I think changing the default is a bad trade-off. Giving everyone a heavier, slower, more complicated data structure just so some people don't get confused when using kwarg indexing?

blegat:
> It's confusing, but because it's opt-in, the user can also opt-in to `container = DenseAxisArray`.

What is opt-in ? If the user uses keyword indexing with an array and it gets an error, he didn't opt-in for anything and we won't understand a thing

> I think changing the default is a bad trade-off. Giving everyone a heavier, slower, more complicated data structure just so some people don't get confused when using kwarg indexing

It shouldn't be heavier if every axis is `Base.OneTo` thanks to https://github.com/jump-dev/JuMP.jl/pull/2524

odow:
> What is opt-in ?

Choosing to use `kwargs` instead of `args`. `args` works the same everywhere. `kwargs` doesn't. That's for the user to figure out.

We're also not going to change the default before JuMP 2.0, so this is a moot discussion.

blegat:
> We're also not going to change the default before JuMP 2.0, so this is a moot discussion.

Why do you mean by the default ? The one we advertise in the docs ?

> Choosing to use `kwargs` instead of `args`. `args` works the same everywhere. `kwargs` doesn't. That's for the user to figure out.

If keyword argument indexing works, the user will just use it without realizing he has opt-in to anything. I don't see how that fixes the usability issue.

odow:
> Why do you mean by the default ? 

The type created by `@variable(model, x[i=1:2, j=1:2])`

> I don't see how that fixes the usability issue.

It doesn't. But it means they can choose a trade-off. Positional indexing works the same everywhere. Kwarg indexing works for Dense and SparseAxisArray.

It's not like you can only use positional with Array and kwarg with {Dense,Sparse}AxisArray.

I also think this discussion is somewhat academic. Let's get it out there and see if people complain.

odow:
@mlubin you're going to need to take a look at this one.

src/Containers/DenseAxisArray.jl
```diff
@@ -674,7 +734,34 @@ end
...
+                "indexing, the indices must match the exact name and order " *
+                "used when creating the container.",
+            )
+        end
+        return kwargs[i]
+    end
+end
+
+function Base.getindex(x::DenseAxisArrayView, args...; kwargs...)
+    if !isempty(kwargs)
```
mlubin, line 759: `getindex` is performance sensitive. Is there any impact from this extra `!isempty(kwargs)` in the case of all positional arguments?

mlubin (commented):
My proposal is to make this explicitly opt in, i.e., disabled unless you pass `keyword_indexing=True` to `@variable` or `@constraint`. This way we can provide a useful error message when `keyword_indexing=True` and the container is `Array`.
Let's see if the feature gets used and if anyone likes it, then we can decide what to break for the next major release.

blegat:
> My proposal is to make this explicitly opt in, i.e., disabled unless you pass `keyword_indexing=True` to `@variable` or `@constraint`. This way we can provide a useful error message when `keyword_indexing=True` and the container is `Array`.
Let's see if the feature gets used and if anyone likes it, then we can decide what to break for the next major release.

That would resolve the usability issue. Maybe this option can be at the level of the model ?

mlubin:
> Maybe this option can be at the level of the model ?

If it's at the level of the model, then does this mean we print a warning or error for every `Array` container in the model?

odow:
I looked into the flag option, and it's actually pretty complicated to do with a model-level flag, because the value of the flag needs to be available at macro-expansion time. so `@variable(model, x[I=1:2], enable_keyword_indexing = true)` works, but some model-level thing doesn't.

One option could be some global constant and require users to opt-in with `JuMP.Containers.enable_keyword_indexing() = true`. 

But in general, I think the option is over-kill. Is it really too much to ask a user who chooses to use keyword indexing to understand the difference between an `Array` and a `DenseAxisArray`? It's a shame that we can't provide a nice error message, but the default isn't that bad, and I can explain in the intro documentation:

```julia
julia> model = Model();

julia> @variable(model, x[i=1:2, j=1:2])
2×2 Matrix{VariableRef}:
 x[1,1]  x[1,2]
 x[2,1]  x[2,2]

julia> @variable(model, y[i=1:2, j=1:2], container = DenseAxisArray)
    Dimension 1, Base.OneTo(2)
    Dimension 2, Base.OneTo(2)
And data, a 2×2 Matrix{VariableRef}:
 y[1,1]  y[1,2]
 y[2,1]  y[2,2]

julia> x[i=1, j=1]
ERROR: MethodError: no method matching getindex(::Matrix{VariableRef}; i=1, j=1)
Closest candidates are:
  getindex(::Array, ::Int64) at array.jl:805 got unsupported keyword arguments "i", "j"
  getindex(::Array, ::Int64, ::Int64, ::Int64...) at array.jl:806 got unsupported keyword arguments "i", "j"
  getindex(::Array, ::UnitRange{Int64}) at array.jl:809 got unsupported keyword arguments "i", "j"
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[8]:1

julia> y[i=1, j=1]
y[1,1]
```

src/Containers/DenseAxisArray.jl
odow: Nope. It gets compiled away:

```julia
shell> git status
On branch od/container-kwarg-getindex
Your branch is up to date with 'origin/od/container-kwarg-getindex'.

nothing to commit, working tree clean

julia> using JuMP

julia> using BenchmarkTools

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x[i=2:3, j=4:5])
    Dimension 1, 2:3
    Dimension 2, 4:5
And data, a 2×2 Matrix{VariableRef}:
 x[2,4]  x[2,5]
 x[3,4]  x[3,5]

julia> @btime getindex($x, 2, 4)
  3.466 ns (0 allocations: 0 bytes)
x[2,4]

julia> exit()
(base) oscar@Oscars-MBP JuMP % git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
(base) oscar@Oscars-MBP JuMP % julia --project=.  
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.7 (2022-07-19)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using JuMP

julia> using BenchmarkTools

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x[i=2:3, j=4:5])
    Dimension 1, 2:3
    Dimension 2, 4:5
And data, a 2×2 Matrix{VariableRef}:
 x[2,4]  x[2,5]
 x[3,4]  x[3,5]

julia> @btime getindex($x, 2, 4)
  3.725 ns (0 allocations: 0 bytes)
x[2,4]
```

There's still room to improve the `kwarg` option, but let's settle on the syntax/opt-in stuff first:
```julia
julia> @btime getindex($x; i = 2, j = 4)
  33.646 ns (1 allocation: 32 bytes)
x[2,4]
```

mlubin:
`@variable(model, x[i=1:2, j=1:2])`
What about generating a warning here because the index variables are unused and have no effect?

src/Containers/DenseAxisArray.jl
mlubin: Is kwarg indexing fundamentally 10x slower than positional indexing? If that's true we might want to reconsider the syntax.

odow:
It seems weird to throw an error there, but not if `@variable(model, x[i=1:2, j=1:2] <= i)`. And it doesn't cover containers of expressions or constraints, which are also liable for errors.

And this is still premised on the fact that users will be confused, which I'm not sure they will be.

_If_ we get complaints, then I think it's something that can be resolved with documentation. Not with warnings or `enable_keyword_indexing_and_I_understand_the_caveats(model)` functions.

## Spitballing other ideas:

We could make something like `x[NamedIndex(i = 2, j = 2)]` work. It's explicitly opt-in, and we could give a nice error for `Array`.

src/Containers/DenseAxisArray.jl
odow: No, if you change the method to a singular `getindex(x; kwargs...)` it comes down to 12ns and 0 allocations, but then I needed to add a bunch of additional methods to avoid ambiguities. Left as-is for now while we sort the syntax.

Keyword indexing in general does have a slight performance hit though.

metab0t: > Keyword indexing in general does have a slight performance hit though.

In fact, keyword indexing can be as fast as Cartesian indexing if the tuple of names is encoded into the customized array type, but it is obviously not an option for JuMP now.

odow:
We seem to be stalemating. To summarize where we're at:

 * We want `x[; kwargs...]` syntax
 * The syntax doesn't work for `Array`, and we can't provide a nice error message
 * We can't change the default container for `x[i=1:2]` to `DenseAxisArray` because that is breaking

The options are:

 1. Release as-is and wait to see if users find it confusing that they can use keyword indexing only with `DenseAxisArray` and `SparseAxisArray`
 2. Have some `enable_keyword_indexing(model)` opt-in
 3. Throw some warning
 4. Don't merge

The warning is hard, because it'd be hard to distinguish `x[i=1:2] >= i` from `x[i=1:2] >= 0`, and we might end up throwing a warning when the user never intended to use kwarg indexing, or not throw a warning when they did.

The opt-in is also hard because it would require the macro code to depend on the run-time setting in `model`. Although we've done this before with `set_string_names_on_creation`, this is a little more complicated to implement because it's in the `Container` submodule. I'll take another stab at implementing.

If that doesn't work, we're either left with 1 or 4, which probably means 4 if we don't agree.

mlubin:
> It seems weird to throw an error there, but not if `@variable(model, x[i=1:2, j=1:2] <= i)`

IMO it's not at all weird to warn when there are unused variables.

> And it doesn't cover containers of expressions or constraints, which are also liable for errors.

Agreed.

> If a user really need to have an Array and not a DenseAxisArray, this user cares about container so we can ask him to do `@variable(model, x[i=1:2, j = 1:3], lower_bound = i + j, requested_container = Array)` However, that is breaking so maybe we need to wait for JuMP v2.0 for this PR ?

This change seems very disruptive for people who are already happily using Array containers. If we do this we should be confident about how useful the keyword indexing is. It's hard to have this confidence before there are people using it in the wild and giving us feedback. So, I'm in favor of the more explicit opt-in version. It's ok with me if it's a flag to the macro. We don't need a complicated model-level flag. If the feedback is, "this feature is great but it's too annoying to opt in", then maybe we'd make it the default and deal with the fallout.

odow:
>  It's ok with me if it's a flag to the macro. We don't need a complicated model-level flag. 

What about a global flag via `Containers.enable_keyword_indexing() = true`? That's a lot simpler to implement than a model- or macro-specific flag. (And a lot less tedious for people to try out.)

odow:
I guess I was imagining that this flag was going to be used in user-level code only

Okay. Let me do the macro- and model-level flag...

odow:
The demo is
```Julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, x[i = 2:3])
    Dimension 1, 2:3
And data, a 2-element Vector{VariableRef}:
 x[2]
 x[3]

julia> x[i = 2]
ERROR: Keyword indexing is disabled. To enable, pass `enable_keyword_indexing = true` to the `Containers.@container` macro, or call `JuMP.enable_container_keyword_indexing(model, true)` before calling any JuMP macros like `@variable`.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] _check_keyword_indexing_allowed(#unused#::Nothing)
   @ JuMP.Containers ~/.julia/dev/JuMP/src/Containers/Containers.jl:31
 [3] #_kwargs_to_args#17
   @ ~/.julia/dev/JuMP/src/Containers/DenseAxisArray.jl:356 [inlined]
 [4] #getindex#20
   @ ~/.julia/dev/JuMP/src/Containers/DenseAxisArray.jl:375 [inlined]
 [5] top-level scope
   @ REPL[5]:1

julia> model = Model();

julia> enable_container_keyword_indexing(model, true)

julia> @variable(model, x[i = 2:3])
    Dimension 1, 2:3
And data, a 2-element Vector{VariableRef}:
 x[2]
 x[3]

julia> x[i = 2]
x[2]
```

docs/src/manual/containers.md
```diff
@@ -485,3 +485,100 @@ JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 2 entries:
...
+
+julia> x[i = 2, j = 1]
+3
+
+julia> Containers.@container(
+           y[i=2:4, j=1:3; i > j],
+           i + j,
+           enable_keyword_indexing = true,
+       )
+SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 6 entries:
```
odow, line 521:
```suggestion
JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 6 entries:
```

docs/src/manual/containers.md
```diff
@@ -485,3 +485,100 @@ JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 2 entries:
...
+  [2, 1]  =  3
+  [3, 1]  =  4
+  [3, 2]  =  5
+  [4, 1]  =  5
+  [4, 2]  =  6
+  [4, 3]  =  7
+
+julia> y[i = 2:4, j = 1]
+SparseAxisArray{Int64, 1, Tuple{Int64}} with 3 entries:
```
odow, line 530:
```suggestion
JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 3 entries:
```

docs/src/manual/containers.md
```diff
@@ -352,7 +352,7 @@ julia> DataFrames.DataFrame(table)
    4 │     3  A       (3, :A)
 ```
 
-## Forcing the container type
+## [Forcing the container type](@id container_forcing)
```
odow, line 355:
```suggestion
## Forcing the container type
```

docs/src/manual/constraints.md
```diff
@@ -870,7 +870,7 @@ JuMP.Containers.SparseAxisArray{ConstraintRef{Model, MathOptInterface.Constraint
     If you have many index dimensions and a large amount of sparsity, read
     [Performance considerations](@ref).
 
-### Forcing the container type
+### [Forcing the container type](@id constraint_forcing)
```
odow, line 873:
```suggestion
### Forcing the container type
```

src/Containers/SparseAxisArray.jl
```diff
@@ -226,7 +284,7 @@ function Base.copy(
         # type information to call SparseAxisArray(dict). As a work-around, we
         # explicitly construct the type of the resulting SparseAxisArray.
         # For more, see JuMP issue #2867.
-        return SparseAxisArray{Any,N,K}(dict)
+        return SparseAxisArray{Any,N,K}(dict, ntuple(i -> Symbol("#$i"), N))
```
odow, line 287:
```suggestion
        return SparseAxisArray{Any,N,K}(dict, nothing)
```

docs/src/manual/containers.md
```diff
@@ -485,3 +485,100 @@ JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 2 entries:
...
+```
+
+The main reason this syntax is currently opt-in is that it does not work for
+`Array`:
+
+```jldoctest container_kwarg_indexing
+julia> Containers.@container(
+           z[i=1:4, j=1:3],
+           i + j,
+           enable_keyword_indexing = true,
```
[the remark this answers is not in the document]
odow: I don't know if we can do this without throwing misleading errors if you create an array using `@variable` after calling `enable_container_keyword_indexing(model, true)`. 

Few people actually use `Containers.@container`, and setting this option requires reading the docstring and the limitations are documented.

docs/src/manual/containers.md
```diff
@@ -485,3 +485,100 @@ JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 2 entries:
   [1, 2]  =  3
   [1, 4]  =  5
 ```
+
+## Keyword indexing
+
+Opt-in to the experimental keyword indexing syntax by passing
+`enable_keyword_indexing = true` to the [`Containers.@container`](@ref) macro,
+or call [`enable_container_keyword_indexing`](@ref).
```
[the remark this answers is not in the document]
odow: I've updated the docs with more detail

docs/src/manual/containers.md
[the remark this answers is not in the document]
blegat: Could we make it so that if `enable_container_keyword_indexing(model, true)` is called or if `enable_keyword_indexing = true` and there are keywords then we create a `DenseAxisArray` instead of an `Array`. Then we never have to throw any error.

docs/src/manual/containers.md
```diff
@@ -485,3 +485,142 @@ JuMP.Containers.SparseAxisArray{Int64, 2, Tuple{Int64, Int64}} with 2 entries:
...
+JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 3 entries:
+  [2]  =  3
+  [3]  =  4
+  [4]  =  5
+```
+
+### The JuMP macros
+
+For containers constructed using the JuMP macros like [`@variable`](@ref),
+opt-in using [`enable_container_keyword_indexing`](@ref):
```
blegat, line 543: Why not `enable_keyword_indexing` ? It's clear we refer to containers when we say `indexing` isn't it ?

docs/src/manual/containers.md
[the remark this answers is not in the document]
odow: Because I don't think keyword indexing should take priority over returning an Array.

I still think we're over thinking this. People don't get confused that `x[1:2]` returns an `Array` but `x[S]` where `S=1:2` returns a `DenseAxisArray`. 

I'd still prefer that this was turned on by default, and we deal with any confusion with better documentation _once we have feedback that it is indeed an issue_.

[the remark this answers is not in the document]
blegat: > Because I don't think keyword indexing should take priority over returning an Array.

If you pass the keyword argument to a macro and then you return an `Array`, it doesn't make any sense.
The user explictly said he wanted keyword indexing so you can't return a container that does not.
For the model level flag, I would also return a `DenseAxisArray` when names indices are used at creation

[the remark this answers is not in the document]
odow: This now triggers an error.

odow:
Okay. I've tried yet again. Now with three-valued logic to tell the difference between the model flag, `enable_keyword_indexing = true`, and `enable_keyword_indexing = false`.

If you set `true` AND the type is an `Array`, then an error is thrown.

The model flag sets `enable_keyword_indexing = missing`, which silently skips names with `Array`.

> For the model level flag, I would also return a DenseAxisArray when names indices are used at creation

I don't agree. This opt-in should enable new features. It shouldn't break existing code.

I'm still going to argue that we now have a complicated opt-in feature that most people won't use (because it's opt-in) and that could just be turned on by default and any confusion addressed with better documentation...

blegat:
> I don't agree. This opt-in should enable new features. It shouldn't break existing code.

I feel like having the following error:
```julia
@constraint(model, x[i = 1:2], enable_keyword_indexing = true)
```
is quite confusing. The users is telling explicitly that he wants to be create a container that he can index with `i`.
Why don't we just return a `DenseAxisArray` ? It's not going to break existing code since existing code don't use this keyword in macros.
For the model-level flag, maybe we can error since we don't want the model-level flag to break existing code but for the keyword to a macro, it's weird to error and force the user to do
```julia
@constraint(model, x[i = 1:2], enable_keyword_indexing = true, container_type = DenseAxisArray)
```
Even more so if we agree that the the long term goal is to return a `DenseAxisArray` in these cases.

odow:
> `@constraint(model, x[i = 1:2], enable_keyword_indexing = true)`

The JuMP macros do not have this keyword. Only `Containers.@container`.

> Even more so if we agree that the the long term goal is to return a DenseAxisArray in these cases.

I _don't_ think this is the long-term goal. My long-term goal is to make it on by default without `enable_keyword_indexing(model, true)`, leaving all other behavior the same.

blegat:
> I _don't_ think this is the long-term goal. My long-term goal is to make it on by default without `enable_keyword_indexing(model, true)`, leaving all other behavior the same.

I don't think we can settle on this PR without agreeing on the long-term goal ^^

odow:
To be honest, I'm beginning to regret opening this PR...

mlubin (approved):

src/JuMP.jl
```diff
@@ -774,6 +774,39 @@ work for JuMP types) and exploits the mutability of `AffExpr` and `QuadExpr`.
...
 
+function _get_index_keyword_indexing_error()
+    return ErrorException(
+        """
+        Indexing an `Array` with keyword arguments is not supported.
+
+        Indexing with keyword arguments _is_ supported for `Container.DenseAxisArray`
+        (and `Container.SparseAxisArray`).
+
+        Force the container type by passing `container = DenseAxisArray` or to any of
```
odow, line 785:
```suggestion
        Force the container type by passing `container = DenseAxisArray` to any of
```

src/Containers/SparseAxisArray.jl
```diff
@@ -140,6 +184,19 @@ end
     return length(expr.args) == 1 ? :(nothing) : expr
 end
 
+_sliced_key_name(::Type{K}, ::Nothing, args...) where {K<:Tuple} = nothing
```
odow, line 187:
```suggestion
```

src/Containers/SparseAxisArray.jl
```diff
@@ -140,6 +184,19 @@ end
...
+
```
odow, line 188:
```suggestion
```

odow:
Performance issue is still unresolved, https://github.com/jump-dev/JuMP.jl/pull/3237#discussion_r1125818709, but I'll fix that in a separate PR. This PR is already quite complicated.

src/JuMP.jl
```diff
@@ -774,6 +774,39 @@ work for JuMP types) and exploits the mutability of `AffExpr` and `QuadExpr`.
...
+        x[3]
+
+        julia> x[i=2]
+        x[2]
+        ```
+        """,
+    )
+end
+
+function Base.getindex(x::Array{<:AbstractJuMPScalar}; kwargs...)
```
blegat, line 806: Isn't that called when you do `getindex(x)` without any keyword argument ? There should be at least one

odow: What does it mean to call `getindex(x)`?

Even without JuMP:
```Julia
julia> x = rand(3)
3-element Vector{Float64}:
 0.8542832661650948
 0.36511705612410084
 0.12183318054742132

julia> x[]
ERROR: BoundsError: attempt to access 3-element Vector{Float64} at index []
Stacktrace:
 [1] throw_boundserror(A::Vector{Float64}, I::Tuple{})
   @ Base ./abstractarray.jl:651
 [2] checkbounds
   @ ./abstractarray.jl:616 [inlined]
 [3] _getindex
   @ ./abstractarray.jl:1196 [inlined]
 [4] getindex(::Vector{Float64})
   @ Base ./abstractarray.jl:1170
 [5] top-level scope
   @ REPL[451]:1
```
But yeah, I guess we could throw a bounds error if there are no kwargs and we get dispatched here.

odow: Updated

blegat (approved):

PR #3238, opened 2023-02-24
Add set_start_values to simplify setting all starting values in a model

odow:
x-ref: https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-1442189960

This is a fairly common request. And "copy this large function from the docs but change it" isn't a great answer to have.

Needs:

 - [x] docs
 - [x] tests

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -781,7 +781,8 @@ julia> dual_start_value(con)
 
 !!! tip
     To simplify setting start values for all variables and constraints in a
-    model, see [`set_start_values`](@ref). The [Primal and dual warm-starts](@ref) also gives a detailed description of how to iterate over constraints
+    model, see [`set_start_values`](@ref). The [Primal and dual warm-starts](@ref)
+    tutorial also gives a detailed description of how to iterate over constraints
     in the model to set custom start values.
 
 ## Constraint containers
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -780,8 +780,9 @@ julia> dual_start_value(con)
 ```
 
 !!! tip
-    For more information, check out the [Primal and dual warm-starts](@ref)
-    tutorial.
+    To simplify setting start values for all variables and constraints in a
+    model, see [`set_start_values`](@ref). The [Primal and dual warm-starts](@ref) also gives a detailed description of how to iterate over constraints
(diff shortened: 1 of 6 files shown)


src/optimizer_interface.jl
```diff
@@ -820,3 +820,127 @@ function optimizer_index(x::Union{VariableRef,ConstraintRef{Model}})
...
+)
+    ## Store a mapping of the variable primal solution
+    variable_primal = Dict()
+    support_variable_primal = MOI.supports(
+        backend(model),
+        MOI.VariablePrimalStart(),
+        MOI.VariableIndex,
+    )
+    if support_variable_primal && variable_primal_start !== nothing
+        for x in all_variables(model)
```
blegat, line 898: This is going to be slow because of the nonconcrete type of `moi_backend`, we need to do `_moi_...`

blegat: Ah no I see, this is because `value` needs a `VariableRef`. Let's keep it simple then

src/optimizer_interface.jl
```diff
@@ -820,3 +820,127 @@ function optimizer_index(x::Union{VariableRef,ConstraintRef{Model}})
...
+            variable_primal[x] = variable_primal_start(x)
+        end
+    end
+    # In the following, we loop through every constraint and store a mapping
+    # from the constraint index to a tuple containing the primal and dual
+    # solutions.
+    constraint_primal, constraint_dual = Dict(), Dict()
+    for (F, S) in list_of_constraint_types(model)
```
blegat, line 906: We can at least have a function barrier for this

docs/src/manual/constraints.md
```diff
@@ -780,8 +780,9 @@ julia> dual_start_value(con)
 ```
 
 !!! tip
-    For more information, check out the [Primal and dual warm-starts](@ref)
-    tutorial.
+    To simplify setting start values for all variables and constraints in a
+    model, see [`set_start_values`](@ref). The [Primal and dual warm-starts](@ref) also gives a detailed description of how to iterate over constraints
```
odow, line 784:
```suggestion
    model, see [`set_start_values`](@ref). The [Primal and dual warm-starts](@ref)
    tutorial also gives a detailed description of how to iterate over constraints
```

blegat (approved):
Nice addition, I often need to do that

PR #3240, opened 2023-02-26
[docs] fix thread safefy issue in parallelism tutorial

odow:
As pointed out on Discourse, https://discourse.julialang.org/t/jump-multi-threading-example/95210/2, I didn't check carefully enough and I assumed that `push!` was an atomic operation in Julia.

### DIFF
--- a/docs/src/tutorials/algorithms/parallelism.md
+++ b/docs/src/tutorials/algorithms/parallelism.md
@@ -46,10 +46,10 @@ The easiest way to use multi-threading in Julia is by placing the
 
 ````julia
 julia> @time begin
-           ids = Int[]
+           ids = zeros(Threads.nthreads())
            Threads.@threads for i in 1:Threads.nthreads()
                global ids
-               push!(ids, Threads.threadid())
+               ids[i] = Threads.threadid()
                sleep(1.0)
            end
        end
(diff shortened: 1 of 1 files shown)


ericphanson:
I don’t think this is safe either, xref https://github.com/JuliaLang/julia/pull/48542

odow:
Urgh. Okay. I've updated to use a lock, which is what I probably should have used right from the start.

docs/src/tutorials/algorithms/parallelism.md
```diff
@@ -205,10 +211,13 @@ set_silent(model)
 @variable(model, x)
 @objective(model, Min, x)
 solutions = Float64[]
+my_lock = Threads.ReentrantLock()
 Threads.@threads for i in 1:10
     set_lower_bound(x, i)
     optimize!(model)
-    push!(solutions, objective_value(model))
+    Threads.lock(my_lock) do
+        push!(solutions, objective_value(model))
```
ericphanson, line 219: For this one, maybe it makes sense to push `(i, objective_value(model))` or something? Since they might not be in order.

odow: :+1: updated

PR #3248, opened 2023-02-27
Backport dot product fix with complex numbers

blegat:
We cannot make a patch release with the master branch because of https://github.com/jump-dev/JuMP.jl/pull/3219 so let's do a patch release on the release-1.8 branch.


mlubin (approved):

odow:
Now `master` doesn't have the v1.8.2 commit. We can easily fix that, but my preference would have been to release v1.9.0. We don't want to have to start messing with backported branches.

PR #3256, opened 2023-02-28
Support inequality syntax for HermitianPSDCone

odow:
This also adds support for general matrix types for which `LinearAlgebra.ishermitian(X) == true`, and it throws a nicer error if not.


blegat:
It's inconsistent with the `Symmetric` case, we should have a square set at the MOI level.
We probably don't need to add a new square bridge though, the same bridge could do both and ensure we have a consistent experience.

odow:
I don't know if we want the complexity of adding a square set to MOI. Isn't it okay just to error? We can always replace the error with a square set later.

blegat:
Yes, erroring is best until we have the square set in MOI.

blegat:
I don't think so

blegat:
Wait, I meant no change in master, not no change in this PR

PR #3260, opened 2023-03-02
Tidy tests by removing inconsistent and unneeded JuMP. prefixes

odow:


test/nlp_solver.jl
```diff
@@ -5,7 +5,7 @@
 #############################################################################
 # JuMP
 # An algebraic modelling language for Julia
-# See https://github.com/jump-dev/JuMP.jl
+# See https://github.com/jump-dev/jl
```
odow, line 8:
```suggestion
# See https://github.com/jump-dev/JuMP.jl
```

PR #3263, opened 2023-03-03
[docs] modify the TokenIgnores in vale for fewer false negatives

odow:
How vale parse the TokenIgnores is a little mysterious. We were previously attempting to skip anything latex related like `\foo`, but it seems to have done more than that. I've swapped to checking `$ foo $`, and vale will skip fenced code blocks.


docs/.vale.ini
```diff
@@ -7,4 +7,4 @@ Vocab = JuMP-Vocab
 BasedOnStyles = Vale
 
 # Skip \\token and \token LaTeX commands
```
odow, line 9:
```suggestion
```

PR #3264, opened 2023-03-03
Revert "Support inequality syntax for HermitianPSDCone (#3256)"

blegat:
This reverts commit ac20706eabd7be48a9a92a443226aea0c7b2c6e6.

As mentioned in https://github.com/jump-dev/JuMP.jl/pull/3256#issuecomment-1448185501, it's an inconsistent behavior with the `Symmetric` case because when you get `dual`, you will get a `Hermitian` matrix while it's not the case for `Symmetric`.


odow:
I misunderstood your comment :smile:

Issue #3267, opened 2023-03-06
[docs] Power systems


jd-foster:
1) The timing comparison between adjustable costs has `elapsed time: 0.1309... seconds` while the _in-place_ version (where the claim is "faster than re-building the entire model") has `elapsed time: 0.2225... seconds`?

2) In the unit commitment formulation we have both the generator simple bounds and the UC bounds:
```julia
function solve_unit_commitment(generators::Vector, wind, scenario)
...
    @variable(model, generators[i].min <= g[i = 1:N] <= generators[i].max)
...
    # !!! New: add binary on-off variables for each generator
    @variable(model, u[i = 1:N], Bin)
    @constraint(model, [i = 1:N], g[i] <= generators[i].max * u[i])
    @constraint(model, [i = 1:N], g[i] >= generators[i].min * u[i])
...
```
which would imply that, since `generators[2].min = 300` then `u[2] == 1` regardless?

odow:
1. Oh dear. This might just be variance in CI. If you run locally, it should be faster?
 2. Bug. Variables should not have bounds

odow:
Huh. It seems that 1 isn't a fluke. It actually goes way back. Let me try again locally. This might be compilation latency

jd-foster:
Yes, that was my thought too.

jd-foster:
For 2)
## Before:
<img width="665" alt="image" src="https://user-images.githubusercontent.com/38274066/223000542-d9b3b2fc-9667-4b23-bc0a-cfca5ac9f0d5.png">

## After:
<img width="650" alt="image" src="https://user-images.githubusercontent.com/38274066/223000705-878ef351-6c7e-4399-b0ef-ab026ae22d1a.png">

odow:
That new graph looks better/ You can see G2 switch on at the right time, and we're not spilling wind anymore.

jd-foster:
Yes, that exact observation led me to look into the formulation.

PR #3269, opened 2023-03-06
[docs] clarify in-place can be slower for small models

odow:
x-ref: #3267 

I think this was actually compilation latency in the reverse direction.

On my first run, I got
```
elapsed time: 0.6746890544891357 seconds
elapsed time: 0.32920384407043457 seconds
```
and on my second I got
```
elapsed time: 0.09308099746704102 seconds
elapsed time: 0.18986296653747559 seconds
```

I'm not sure if it's because we've changed things in HiGHS and JuMP since I wrote the tutorial, or if it is to do with recent changes in compilation latency. But I've definitely given tutorials and talks where in-place is faster... (Perhaps it was also back when we used GLPK in the documentation?).

It seems like this could change in the future, so let's be more circumspect in how we word things going forward.

### DIFF
--- a/docs/src/tutorials/applications/power_systems.jl
+++ b/docs/src/tutorials/applications/power_systems.jl
@@ -241,8 +241,10 @@ inplace_df = solve_economic_dispatch_inplace(
 )
 print(string("elapsed time: ", time() - start, " seconds"))
 
-# Adjusting specific constraints or the objective function is faster than
-# re-building the entire model.
+# For small models, adjusting specific constraints or the objective function is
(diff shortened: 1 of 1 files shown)


odow:
Thoughts @jd-foster?

https://jump.dev/JuMP.jl/previews/PR3269/tutorials/applications/power_systems/#Modifying-the-JuMP-model-in-place

jd-foster:
Looks like the best you can do without adding a big model to the example, but the docs build time is lengthy enough as is.

PR #3273, opened 2023-03-08
Add JuMP-level Nonnegatives, Nonpositives, and Zeros sets

### LINKED ISSUE #3265: Add JuMP.Zeros set to replace Function-in-EqualTo(0)
x-ref: https://github.com/jump-dev/JuMP.jl/pull/3257#issuecomment-1453197428

odow:
This change enables `Ax >= b` to be lowered to `Ax - b in Nonnegatives()`, and removes the confusing Vector-in-GreaterThan(0) method in build_constraint that asserted the right-hand side was zero.

Closes #3265 

I'm not sure why it didn't occur to me to do this ages ago.

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -134,7 +134,7 @@ the constraint object `con_vector` or `con_scalar`.
    the broadcast `.==`.
 
 JuMP reformulates both constraints into the other form if needed by the solver,
-but choosing right format for a particular solver is more efficient.
+but choosing the right format for a particular solver is more efficient.
 
 You can also use `<=`, `.<=` , `>=`, and `.>=` as comparison operators in the
 constraint.
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -108,16 +108,36 @@ julia> b = [5, 6]
  5
  6
 
-julia> @constraint(model, con, A * x .== b)
+julia> @constraint(model, con_vector, A * x == b)
+con_vector : [x[1] + 2 x[2] - 5, 3 x[1] + 4 x[2] - 6] ∈ MathOptInterface.Zeros(2)
+
+julia> @constraint(model, con_scalar, A * x .== b)
 2-element Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.EqualTo{Float64}}, ScalarShape}}:
- con : x[1] + 2 x[2] = 5.0
- con : 3 x[1] + 4 x[2] = 6.0
+ con_scalar : x[1] + 2 x[2] = 5.0
+ con_scalar : 3 x[1] + 4 x[2] = 6.0
 ```
 
-!!! note
-    Make sure to use [Julia's dot syntax](https://docs.julialang.org/en/v1/manual/functions/index.html#man-vectorized-1)
-    in front of the comparison operators (for example, `.==`, `.>=`, and `.<=`). If you
-    use a comparison without the dot, an error will be thrown.
+The two constraints, `==` and `.==` are similar, but subtly different. The first
+creates a single constraint that is a [`MOI.VectorAffineFunction`](@ref) in
+[`MOI.Zeros`](@ref) constraint. The second creates a vector of
+[`MOI.ScalarAffineFunction`](@ref) in [`MOI.EqualTo`](@ref) constraints.
+
+Which formulation to choose depends on the solver, and what you want to do with
+the constraint object `con_vector` or `con_scalar`.
+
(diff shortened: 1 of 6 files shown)


docs/src/manual/constraints.md
```diff
@@ -108,16 +108,36 @@ julia> b = [5, 6]
...
+ * If you are using a conic solver, expect the dual of `con_vector` to be a
+   `Vector{Float64}`, and do not intend to delete a row in the constraint,
+   choose the `==` formulation.
+ * If you are using a solver that expects a list of scalar constraints, for
+   example HiGHS, or you wish to delete part of the constraint or access a
+   single row of the constraint, for example, `dual(con_scalar[2])`, then use
+   the broadcast `.==`.
+
+JuMP reformulates both constraints into the other form if needed by the solver,
+but choosing right format for a particular solver is more efficient.
```
jd-foster, line 137:
```suggestion
but choosing the right format for a particular solver is more efficient.
```

src/macros.jl
```diff
@@ -523,6 +604,50 @@ end
...
+function build_constraint(_error::Function, f, set::Zeros, args...; kwargs...)
+    return build_constraint(_error, f, MOI.EqualTo(0.0), args...; kwargs...)
+end
+
+function build_constraint(
+    _error::Function,
+    f::AbstractVector,
+    set::Nonnegatives,
+)
+    return build_constraint(_error, f, MOI.Nonnegatives(length(f)))
```
blegat, line 636: This is equivalent to assuming that `@constraint(model, a >= b)` is equivalent to `@constraint(model, a >= b, Nonnegatives())` right ?
I think we explicitly decided not to have these cones by default but I would be ok to change it but I think we need to make sure everybody agrees on it.
It also kind of makes sense since we assumes `Nonnegatives` by default to compare vector of objective in multiobjective optimization.

odow: > This is equivalent to assuming that @constraint(model, a >= b) is equivalent to @constraint(model, a >= b, Nonnegatives()) right?

Yes, for vector-valued `a - b`. I think this is pretty standard interpretation. Do you remember why we decided not to?

It makes it a lot easier to write models that conic solvers support without bridges. Currently they'll bridge _every_ scalar linear constraint to a vector-valued one of dimension 1.

blegat: I don't remember, @mlubin do you ?
It's also the convention Boyd's uses in his book.

mlubin:
The trade-off is that the solvers where we care about an efficient direct mode are unlikely to natively support `Nonnegatives`. Isn't this a potentially breaking change for anyone using direct mode?

blegat:
> Isn't this a potentially breaking change for anyone using direct mode?

No because the only cases that will change are cases that were previously erroring. The scalar `>=` and `.>=` still create `GreaterThan` constraints but the vector `>=` used to error and it now creates a `Nonnegatives` constraint.

test/test_constraint.jl
```diff
@@ -113,9 +113,27 @@ function test_extension_AffExpr_vectorized_constraints(
     model = ModelType()
     @variable(model, x)
     err = ErrorException(
-        "In `@constraint(model, [x, 2x] == [1 - x, 3])`: Unexpected vector in scalar constraint. Did you mean to use the dot comparison operators like .==, .<=, and .>= instead?",
+        "In `@constraint(model, [x, 2x] in MOI.EqualTo(1.0))`: " *
+        "Unexpected vector in scalar constraint. The left- and right-hand " *
+        "sides of the constraint must have the same dimension.",
     )
-    @test_throws_strip err @constraint(model, [x, 2x] == [1 - x, 3])
```
odow, line 118: Here's an example of code that previously errored

mlubin:
I guess it's ok then. It's a bit confusing that we'll allow both `A*x >= b` and `A*x .>= b` which are the same constraints with subtle differences.

docs/src/manual/constraints.md
```diff
@@ -108,16 +108,36 @@ julia> b = [5, 6]
...
- con : 3 x[1] + 4 x[2] = 6.0
+ con_scalar : x[1] + 2 x[2] = 5.0
+ con_scalar : 3 x[1] + 4 x[2] = 6.0
 ```
 
-!!! note
-    Make sure to use [Julia's dot syntax](https://docs.julialang.org/en/v1/manual/functions/index.html#man-vectorized-1)
-    in front of the comparison operators (for example, `.==`, `.>=`, and `.<=`). If you
-    use a comparison without the dot, an error will be thrown.
+The two constraints, `==` and `.==` are similar, but subtly different. The first
```
odow, line 120: Is this bit of the documentation enough to describe the differences?

mlubin (approved):

PR #3274, opened 2023-03-08
[docs] split user-defined Hessian tutorial in two

jd-foster:
Removed the Rosenbrock example (it's covered in the manual section) to get to the bilevel optimization as the meat of the tutorial, which is not in the title too. What do you think?


odow:
If I recall correctly, I think @ccoffrin and I discussed that having a simple example to start had benefits. Perhaps we could just split in two?

odow:
Perhaps let's leave the user-defined hessians as just Rosenbrock, and make "Nested optimization" a separate tutorial?

jd-foster:
Ok, I'll split it in two.

docs/src/tutorials/nonlinear/nested_problems.jl
```diff
@@ -0,0 +1,226 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,  #src
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  #src
+# SOFTWARE.                                                                      #src
+
+# # Nested optimization problems
+
+# In this tutorial, we explain how to write a user-defined function (see
+# [User-defined Functions](@ref)) with a Hessian matrix explicitly provided by
+# the user. User-defined Hessian functions can be useful when solving more complicated
+# problems, such as bilevel or nested optimization problems.
```
odow, line 26: I think the purpose of this tutorial is now to show how to write a nested optimization problem, and we should clarify below that the Hessian is optional if they just have the first derivatives.

docs/src/tutorials/nonlinear/nested_problems.jl
```diff
@@ -0,0 +1,226 @@
...
+# Second, we need to define the Hessian of ``V`` with respect to ``x``. This is
+# a symmetric matrix, but in our example only the diagonal elements are
+# non-zero:
+
+function ∇²V(H::AbstractMatrix, x...)
+    _, y = solve_lower_level(x...)
+    H[1, 1] = 2 * y[1]
+    H[2, 2] = 2 * y[2]
+    return
+end
```
odow, line 130: You could say here

```
!!! info
    Providing an explicit Hessian function is optional.
```

docs/src/tutorials/nonlinear/nested_problems.jl
```diff
@@ -0,0 +1,233 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,  #src
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  #src
+# SOFTWARE.                                                                      #src
+
+# # Nested optimization problems
+
+# In this tutorial we explain how to write a *nested* optimization problem,
+# where an *outer* problem uses the results from the optimization of an *inner*
+# subproblem.
+# We describe how to define a user-defined function to handle the decomposition
```
odow, line 26:
```suggestion
#
# To model the problem, we define a user-defined function to handle the decomposition
```

PR #3276, opened 2023-03-09
[docs] Improve docstrings and convert julia blocks to jldoctest

odow:
Follow-up to #3275 

No code changes, only docstrings.

### DIFF
--- a/src/Containers/nested_iterator.jl
+++ b/src/Containers/nested_iterator.jl
@@ -47,8 +47,18 @@ Create a [`NestedIterator`](@ref).
 
 ## Example
 
-```julia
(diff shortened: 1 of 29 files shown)


src/Containers/nested_iterator.jl
```diff
@@ -47,8 +55,18 @@ Create a [`NestedIterator`](@ref).
 
 ## Example
 
-```julia
-nested(1:2, ["A", "B"]; condition = (i, j) -> isodd(i) || j == "B")
```
odow, line 51: Here's one example where changing to a doctest caught some old syntax that hadn't been updated.

PR #3277, opened 2023-03-09
Update[typo] should_i_use.md

aafsar:
Changed "... JuMP can [create -> solve] problems at similar speeds"

### DIFF
--- a/docs/src/should_i_use.md
+++ b/docs/src/should_i_use.md
@@ -42,7 +42,7 @@ Key reasons to use JuMP include:
      can be embedded in commercial software that complies with the terms of the
      license.
  - Speed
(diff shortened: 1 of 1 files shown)


blegat:
I think this is intentional. The job of JuMP and AMPL is to transform the problem from the algebraic form given given by the user to the format required by the solver. The solving part wouldn't depend much on JuMP vs AMPL. Except for nonlinear solver which call AD oracle during solve. Maybe `create` is not the right work though.

mlubin:
Right, JuMP does not _solve_ optimization problems

aafsar:
Ah makes sense, thank you! Closing the PR then.

PR #3280, opened 2023-03-12
[Containers] fix when axis is Vector{Any}

### LINKED ISSUE #3279: Containers can't index when key is Vector{Any}
Reported on SO: https://stackoverflow.com/questions/75692147/why-is-my-program-searching-for-this-key

```julia
julia> using JuMP

julia> key = Any[Any["macros", 1], "Outsized_Equipment"]
2-element Vector{Any}:
 Any["macros", 1]
 "Outsized_Equipment"

julia> K = Any[key]
1-element Vector{Any}:
 Any[Any["macros", 1], "Outsized_Equipment"]

julia> model = Model();

julia> @variable(model, x[K])
    Dimension 1, Any[Any[Any["macros", 1], "Outsized_Equipment"]]
And data, a 1-element Vector{VariableRef}:
 x[Any[Any["macros", 1], "Outsized_Equipment"]]

julia> x[key]
ERROR: KeyError: key "macros" not found
Stacktrace:
```

odow:
Closes #3279


src/Containers/DenseAxisArray.jl
```diff
@@ -60,9 +60,15 @@ end
 
 function Base.getindex(
     x::_AxisLookup{Dict{K,Int}},
-    key::K,
-) where {K<:AbstractVector}
-    return x.data[key]
+    key::T,
+) where {T<:AbstractVector,K>:T}
```
blegat, line 64: Can't we keep also the efficient method in case `T` is `K` ?

src/Containers/DenseAxisArray.jl
```diff
@@ -60,9 +60,15 @@ end
...
+    # In this method, we can't tell if `keys` is an actual key, or a vector of
+    # keys. One common cause happens when `K` is `Any`. To ensure correctness,
+    # we need to slightly expensive check if `keys` is a key.
+    if haskey(x.data, key)
+        return x.data[key]
+    end
+    return [x[k] for k in key]
```
blegat, line 71: Can't there be some edge cases here in case the vector of keys is like
`[1, [1], [[1]]]`, Look's like Russel's paradox :laughing:

src/Containers/DenseAxisArray.jl
odow: I don't think so, because the case `T = K = Vector{Any}` is ambiguous. There's probably a bunch of similar types.

src/Containers/DenseAxisArray.jl
odow: Let's hope that no one builds the indices of their variables using set theory...

I think this is still safe to merge. It's a strict improvement that fixes buggy behavior. No one has every complained about not being able to write a key like `[1, [1], [[1]]]`...

odow:
@blegat bump

blegat:
I'm still not convinced that there is no edge cases like https://github.com/jump-dev/JuMP.jl/pull/3280#discussion_r1133578235
I feel this is getting into dark territory

src/Containers/DenseAxisArray.jl
odow: `Any[Any["a"], Any["b"], Any[Any["a"], Any["b"]]]`

odow: Call is not sure what to do here. Look at `Any[1, 2]`. Disallow slicing if case is ambiguous. Provide alternate slicing syntax.

src/Containers/DenseAxisArray.jl
```diff
@@ -60,9 +60,21 @@ end
...
+    key::T,
+) where {T<:AbstractVector,K>:T}
+    # In this method, we can't tell if `keys` is an actual key, or a vector of
+    # keys. One common cause happens when `K` is `Any`.
+    if haskey(x.data, key)
+        if all(haskey(x.data, k) for k in key)
+            error(
+                "ambiguous use of getindex with key $key. We cannot tell if " *
+                "you meant to return the single element corresponding to the " *
+                "key, or a slice for each element in the key.",
```
odow, line 72: So I've opted just to error in the strict case that you index an ambiguous case. This is kinda similar to method ambiguities. You can define them, but you only error if there is a confusion at runtime. You could imagine someone defining an ambiguous set an then converting to `Array(x)` and being fine.

blegat (approved):

PR #3281, opened 2023-03-12
Add support for {Hermitian,Symmetric} in Zeros constraint

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3257#issuecomment-1465135635

### DIFF
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -634,15 +634,19 @@ function build_constraint(_error::Function, H::LinearAlgebra.Hermitian, ::Zeros)
     n = LinearAlgebra.checksquare(H)
     shape = HermitianMatrixShape(n)
     x = vectorize(H, shape)
-    return build_constraint(error, x, MOI.Zeros(length(x)))
+    return VectorConstraint(x, MOI.Zeros(length(x)), shape)
 end
+
+reshape_set(s::MOI.Zeros, ::HermitianMatrixShape) = Zeros()
 
 function build_constraint(_error::Function, f::LinearAlgebra.Symmetric, ::Zeros)
     n = LinearAlgebra.checksquare(f)
     shape = SymmetricMatrixShape(n)
     x = vectorize(f, shape)
-    return build_constraint(error, x, MOI.Zeros(length(x)))
+    return VectorConstraint(x, MOI.Zeros(length(x)), shape)
 end
+
+reshape_set(::MOI.Zeros, ::SymmetricMatrixShape) = Zeros()
 
 function build_constraint(_error::Function, ::AbstractMatrix, ::Nonnegatives)
     return _error(
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -647,14 +647,18 @@ end
 function build_constraint(_error::Function, ::AbstractMatrix, ::Nonnegatives)
     return _error(
         "Unsupported matrix in vector-valued set. Did you mean to use the " *
-        "broadcasting syntax `.>=` instead?",
+        "broadcasting syntax `.>=` instead? Alternatively, perhaps you are " *
+        "missing a set argument like `@constraint(model, X >= 0, PSDCone())` " *
+        "or `@constraint(model, X >= 0, HermmitianPSDCone())`.",
     )
 end
 
 function build_constraint(_error::Function, ::AbstractMatrix, ::Nonpositives)
     return _error(
         "Unsupported matrix in vector-valued set. Did you mean to use the " *
-        "broadcasting syntax `.<=` instead?",
+        "broadcasting syntax `.<=` instead? Alternatively, perhaps you are " *
+        "missing a set argument like `@constraint(model, X <= 0, PSDCone())` " *
+        "or `@constraint(model, X <= 0, HermmitianPSDCone())`.",
     )
 end
 
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -629,3 +629,46 @@ function build_constraint(
         shape,
     )
 end
+
+function build_constraint(
+    _error::Function,
+    H::LinearAlgebra.Hermitian,
+    ::Zeros,
+)
+    n = LinearAlgebra.checksquare(H)
+    shape = HermitianMatrixShape(n)
+    x = vectorize(H, shape)
+    return build_constraint(error, x, MOI.Zeros(length(x)))
+end
+
+function build_constraint(
+    _error::Function,
+    f::LinearAlgebra.Symmetric,
+    ::Zeros,
+)
+    n = LinearAlgebra.checksquare(f)
+    shape = SymmetricMatrixShape(n)
+    x = vectorize(f, shape)
+    return build_constraint(error, x, MOI.Zeros(length(x)))
+end
+
+function build_constraint(_error::Function, ::AbstractMatrix, ::Nonnegatives)
+    return _error(
+        "Unsupported matrix in vector-valued set. Did you mean to use the " *
+        "broadcasting syntax `.>=` instead?",
+    )
+end
+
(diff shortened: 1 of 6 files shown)


test/test_macros.jl
```diff
@@ -1894,4 +1894,35 @@ end
...
+        ),
+        @constraint(model, X <= A),
+    )
+    @test_throws_strip(
+        ErrorException(
+            "In `@constraint(model, X == A)`: Unsupported matrix in " *
+            "vector-valued set. Did you mean to use the broadcasting syntax " *
+            "`.==` instead?",
+        ),
+        @constraint(model, X == A),
```
odow, line 1923: We recently added (https://github.com/jump-dev/JuMP.jl/pull/3273) support for `vector == vector`. But it is less clear what to do for `matrix == matrix`. 

It gets lowered to `X - A in Zeros()`, but `X - A` is a matrix, and `Zeros()` is a vector-valued set, which triggers this error.

src/sd.jl
```diff
@@ -629,3 +629,38 @@ function build_constraint(
...
+    n = LinearAlgebra.checksquare(f)
+    shape = SymmetricMatrixShape(n)
+    x = vectorize(f, shape)
+    return build_constraint(error, x, MOI.Zeros(length(x)))
+end
+
+function build_constraint(_error::Function, ::AbstractMatrix, ::Nonnegatives)
+    return _error(
+        "Unsupported matrix in vector-valued set. Did you mean to use the " *
+        "broadcasting syntax `.>=` instead?",
```
blegat, line 650: The user could also have forgotter to use `PSDCone` or `HermitianPSDCone`

test/test_macros.jl
blegat: We could make the equality case work.
The inequality, it's better that it's an error since the user most probably forgot `PSDCone()`

araujoms: I think it wouldn't even be possible to support inequalities, as JuMP cannot know whether the user wants PSDCone() or HermitianPSDCone().

In any case, I concur, allowing inequalities leads to ambiguities. Maybe the user has a symmetric matrix, but wants to constrain it to be elementwise nonnegative.

There's also a bug that always happens with YALMIP: the user has a matrix they think is Hermitian but it's actually not because of numerical noise. Then the user writes >= 0 thinking it's constraining it to be PSD, but YALMIP instead interprets that as elementwise non-negativity.

blegat: > There's also a bug that always happens with YALMIP: the user has a matrix they think is Hermitian but it's actually not because of numerical noise. Then the user writes >= 0 thinking it's constraining it to be PSD, but YALMIP instead interprets that as elementwise non-negativity.

Indeed, we want to avoid that, an error asking the user to be explicit about it is better than trying to save a few keystrokes

odow: Let's merge this and see how we go re-feedback. The non-symmetric case `X .== Y` is pretty easy to type, and there's a nice error message.

The specialized support for symmetric and hermitian is more useful because of the redundant constraints.

Agree on inequalities.

blegat: > The case of equalities is not so clear-cut because the user intention is not ambiguous. If JuMP interprets == as .== in the non-symmetric case the result will be a slowdown, not a bug (assuming the user thought the matrix was Hermitian).

> I would just change the error message to say "Please declare your matrix as Symmetric or Hermitian if it is the case. Otherwise use .=="

We can't really do that since JuMP is a general purpose solver so having equality between matrices is completely fine and it's weird to talk about `Symmetric` or `Hermitian` matrices outside the context of SDP.

src/sd.jl
```diff
@@ -629,3 +629,42 @@ function build_constraint(
         shape,
     )
 end
+
+function build_constraint(_error::Function, H::LinearAlgebra.Hermitian, ::Zeros)
+    n = LinearAlgebra.checksquare(H)
+    shape = HermitianMatrixShape(n)
+    x = vectorize(H, shape)
+    return build_constraint(error, x, MOI.Zeros(length(x)))
```
blegat, line 637:
```suggestion
    return VectorConstraint(x, MOI.Zeros(length(x)), shape)
```

src/sd.jl
```diff
@@ -629,3 +629,42 @@ function build_constraint(
...
+end
+
+function build_constraint(_error::Function, f::LinearAlgebra.Symmetric, ::Zeros)
+    n = LinearAlgebra.checksquare(f)
+    shape = SymmetricMatrixShape(n)
+    x = vectorize(f, shape)
+    return build_constraint(error, x, MOI.Zeros(length(x)))
```
blegat, line 644:
```suggestion
    return return VectorConstraint(x, MOI.Zeros(length(x)), shape)
```

test/test_constraint.jl
```diff
@@ -1544,4 +1544,29 @@ function test_extension_Zeros(ModelType = Model, VariableRefType = VariableRef)
...
+
+function test_symmetric_in_zeros()
+    model = Model()
+    @variable(model, H[1:2, 1:2], Symmetric)
+    c = @constraint(model, H == LinearAlgebra.I)
+    con = constraint_object(c)
+    @test isequal_canonical(con.func[1], H[1, 1] - 1)
+    @test isequal_canonical(con.func[2], H[1, 2] - 0)
+    @test isequal_canonical(con.func[3], H[2, 2] - 1)
+    @test con.set == MOI.Zeros(3)
```
blegat, line 1568: Test that the value and dual are `Symmetric` matrices and not a vector

test/test_constraint.jl
```diff
@@ -1544,4 +1544,29 @@ function test_extension_Zeros(ModelType = Model, VariableRefType = VariableRef)
...
+function test_hermitian_in_zeros()
+    model = Model()
+    @variable(model, H[1:2, 1:2] in HermitianPSDCone())
+    c = @constraint(model, H == LinearAlgebra.I)
+    con = constraint_object(c)
+    @test isequal_canonical(con.func[1], real(H[1, 1]) - 1)
+    @test isequal_canonical(con.func[2], real(H[1, 2]))
+    @test isequal_canonical(con.func[3], real(H[2, 2]) - 1)
+    @test isequal_canonical(con.func[4], imag(H[1, 2]))
+    @test con.set == MOI.Zeros(4)
```
blegat, line 1556: Test that the value and dual are `Symmetric` matrices and not a vector

test/test_macros.jl
araujoms: Fair enough, but the error message seems to imply that == is never supported for matrices, which is not the case.

blegat: Good point,  `JuMP.Zeros` is indeed not a vector-valued set (since it works with scalars, Symmetric and Hermitian). "Equality between `$(typeof(matrix))`. Use `.==` for elementwise inequality or wrap the matrix in  `Symmetric` or `Hermitian`) for equality between matrices of such special structure"

blegat:
Let's also document it, e.g., in https://jump.dev/JuMP.jl/stable/manual/constraints/#Symmetry

test/test_macros.jl
araujoms: Perfect, thanks.

araujoms:
One should also update the [complex tutorial](https://jump.dev/JuMP.jl/stable/tutorials/conic/quantum_discrimination/) to use the new syntax and mention that it's more efficient.

test/test_macros.jl
odow: Updated the docs and the error message.

blegat (approved):

PR #3282, opened 2023-03-13
[docs] add p-norm example

odow:
Came up on Discourse: https://discourse.julialang.org/t/nonlinear-optimization-using-jump-the-solver-does-not-support-nonlinear-problems-i-e-nlobjective-and-nlconstraint/95978

Preview: https://jump.dev/JuMP.jl/previews/PR3282/tutorials/conic/tips_and_tricks/#P-Norm

### DIFF
--- a/docs/src/tutorials/conic/tips_and_tricks.jl
+++ b/docs/src/tutorials/conic/tips_and_tricks.jl
@@ -223,11 +223,10 @@ value(t), value(x)
 # can be modeled using a set of power cones. See the [Mosek Modeling Cookbook](https://docs.mosek.com/modeling-cookbook/powo.html#p-norm-cones)
 # for the derivation.
 
-function p_norm(x0::Vector, p)
-    N = length(x0)
+function p_norm(x::Vector, p)
+    N = length(x)
     model = Model(SCS.Optimizer)
     set_silent(model)
-    @variable(model, x[i = 1:N] == x0[i])
     @variable(model, r[1:N])
     @variable(model, t)
     @constraint(model, [i = 1:N], [r[i], t, x[i]] in MOI.PowerCone(1 / p))
@@ -237,8 +236,8 @@ function p_norm(x0::Vector, p)
     return value(t)
 end
 
-x0 = rand(5);
-LinearAlgebra.norm(x0, 4), p_norm(x0, 4)
+x = rand(5);
+LinearAlgebra.norm(x, 4), p_norm(x, 4)
 
 # ## Positive Semidefinite Cone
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/conic/tips_and_tricks.jl
```diff
@@ -217,6 +217,28 @@ value(t), value(x)
 # that offers an alternative formulation that can be more efficient for some
 # formulations.
 
+# ## P-Norm
+
+# The p-norm ``||x||_p = \left(\sum\limits_{i} |x_i|^p\right)^{\frac{1}{p})``
```
odow, line 222:
```suggestion
# The p-norm ``||x||_p = \left(\sum\limits_{i} |x_i|^p\right)^{\frac{1}{p}}``
```

odow:
This probably comes up often enough that I wonder if it's worth a dedicated set and bridge?

docs/src/tutorials/conic/tips_and_tricks.jl
```diff
@@ -217,6 +217,28 @@ value(t), value(x)
 # that offers an alternative formulation that can be more efficient for some
 # formulations.
 
+# ## P-Norm
+
+# The p-norm ``||x||_p = \left(\sum\limits_{i} |x_i|^p\right)^{\frac{1}{p}}``
+# can be modeled using a set of power cones. See the [Mosek Modeling Cookbook](https://docs.mosek.com/modeling-cookbook/powo.html#p-norm-cones)
```
blegat, line 223: What do you mean by "set of power cones"

blegat:
> This probably comes up often enough that I wonder if it's worth a dedicated set and bridge?

Yes, @chriscoey do you have something like that in Hypatia ?

odow:
I've opened an issue: https://github.com/jump-dev/MathOptInterface.jl/issues/2118.

This doc PR can be merged without waiting for the resolution.

chriscoey:
> > This probably comes up often enough that I wonder if it's worth a dedicated set and bridge?
> 
> Yes, @chriscoey do you have something like that in Hypatia ?

No, we didn't know a nice barrier for p-norm so we didn't try implementing anything.

docs/src/tutorials/conic/tips_and_tricks.jl
odow, line 223:
```suggestion
# can be modeled using [`MOI.PowerCone`](@ref)s. See the [Mosek Modeling Cookbook](https://docs.mosek.com/modeling-cookbook/powo.html#p-norm-cones)
```

PR #3285, opened 2023-03-16
[docs] add Google style rules to Vale and fix warnings

odow:
x-ref: https://github.com/jump-dev/MathOptInterface.jl/pull/2110

Needs 

 - [ ] MOI v1.13.2


odow:
While we wait for v1.13.2, I've disabled the Google checks. We could still merge this as an improvement, and it would simplify the diff when we go to actually enable the rules.

Issue #3291, opened 2023-03-17
Create Hermitian variables that are not PSD


araujoms:
I need to create a Hermitian variable that is not PSD. JuMP doesn't provide any simple way of doing that. It would be great if I could just `@variable(model, x[1:d,1:d], Hermitian)` as I can do for Symmetric variables.

blegat:
https://github.com/jump-dev/JuMP.jl/pull/3292 provides such feature. We keep `@variable(model, x[1:d,1:d], Symmetric)` more as compatibility with older releases. We prefer extending with `@variable(model, ... in ...)` because that type of extension can easily be done independently of the macro code while `@variable(model, x[1:d,1:d], Hermitian)` would need us to update the macro code so it cannot be done by JuMP extensions for instance.

araujoms:
Isn't it easy to just create an alias? I think it's rather inelegant to have conflicting syntax between the Symmetric and Hermitian cases.

blegat:
As you can see below:
```julia
julia> @variable(model, [1:2, 1:2], Hermitian)
ERROR: UndefVarError: Hermitian not defined
Stacktrace:
 [1] (::var"#5#6")(317::Int64, 318::Int64)
   @ Main ~/.julia/dev/JuMP/src/Containers/macro.jl:297
 [2] (::JuMP.Containers.var"#44#45"{var"#5#6"})(I::Tuple{Int64, Int64})
   @ JuMP.Containers ~/.julia/dev/JuMP/src/Containers/container.jl:81
 [3] iterate
   @ ./generator.jl:47 [inlined]
 [4] collect(itr::Base.Generator{JuMP.Containers.VectorizedProductIterator{Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}, JuMP.Containers.var"#44#45"{var"#5#6"}})
   @ Base ./array.jl:787
 [5] map(f::Function, A::JuMP.Containers.VectorizedProductIterator{Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}})
   @ Base ./abstractarray.jl:2961
 [6] container
   @ ~/.julia/dev/JuMP/src/Containers/container.jl:81 [inlined]
 [7] container(f::Function, indices::JuMP.Containers.VectorizedProductIterator{Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}})
   @ JuMP.Containers ~/.julia/dev/JuMP/src/Containers/container.jl:75
[...]

PR #3292, opened 2023-03-17
Add HermitianMatrixSpace

blegat:
Closes https://github.com/jump-dev/JuMP.jl/issues/3291

- [x] Add example in manual like skew symmetric
- [x] Tests (including edge cases with variable bounds, etc...)

PR #3293, opened 2023-03-17
Add Hermitian tag in variable macro

blegat:
Requires https://github.com/jump-dev/JuMP.jl/pull/3292 to be merged first.

Another option is to reexport the symbol of `LinearAlgebra.Hermitian` and `LinearAlgebra.Symmetric` and to overload `build_variable`. I always felt it a bit annoying to ask to user to load `LinearAlgebra` intself to use the `Symmetric` symbol which is part of the JuMP syntax.
The issue is that it won't work if the user `import` JuMP because then `Symmetric` won't be in the scope so it would be breaking.

Closes https://github.com/jump-dev/JuMP.jl/issues/3291

### DIFF
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -1260,8 +1260,17 @@ julia> first(all_constraints(model, Vector{VariableRef}, MOI.HermitianPositiveSe
 
 ### Example: Hermitian variables
 
-Declare a matrix of JuMP variables to be Hermitian using
-[`HermitianMatrixSpace`](@ref):
+Declare a matrix of JuMP variables to be Hermitian using the `Hermitian` tag:
+```jldoctest hermitian
+julia> model = Model();
+
+julia> @variable(model, x[1:2, 1:2], Hermitian)
+2×2 LinearAlgebra.Hermitian{GenericAffExpr{ComplexF64, VariableRef}, Matrix{GenericAffExpr{ComplexF64, VariableRef}}}:
+ real(x[1,1])                               …  real(x[1,2]) + (0.0 + 1.0im) imag(x[1,2])
+ real(x[1,2]) + (0.0 - 1.0im) imag(x[1,2])     real(x[2,2])
+```
+
+This is equivalent to declaring the variable in [`HermitianMatrixSpace`](@ref):
 ```jldoctest hermitian
 julia> model = Model();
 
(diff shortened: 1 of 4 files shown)


odow:
Just merge this into one PR? Doesn't need to be separate.

blegat:
Yes, but I was worried this one not reach consensus directly

odow:
Thinking on this more, I understand arguments both for and against adding this. You could ask why `Hermitian` but not `HermitianPSD` or `SkewSymmetric`. Also `H[1:2, 1:2] in HermitianMatrixSpace()` isn't that much more complicated than `H[1:2, 1:2], Hermitian`, especially if people would try `H[1:2, 1:2], LinearAlgebraHermitian`.

You're right about the breaking stuff though. In JuMP 2.0, I'd make it `Bool`, `Integer`, `LinearAlgebra.Symmetric`, etc. Use the Julia types, don't detect symbols in the macro.

blegat:
Yes, Especially `SkewSymmetric`. If the argument is that this is types that exists in Julia then the fact that we use the symbols makes it inconsistent. So if we add this, we would detect it by the symbol for consistency but the rule is still that we only detect symbols that correspond to actual types because in JuMP v2 we will not detect them as symbols anymore and
```julia
import JuMP
model = Model()
@variable(model, x[1:2, 1:2], Hermitian)
```
won't work anymore, but
```julia
using JuMP
model = Model()
h = Hermitian
@variable(model, x[1:2, 1:2], h)
```
will work in JuMP v2. We would make it work in JuMP v1 but in a separate PR that does it for all of them

odow:
Agree. I'm ambivalent on adding this. How common are hermitians matrices? Not very? If so, asking for `in HermitianMatrixSpace()` isn't that much of an ask.

araujoms:
Besides introducing an inconsistency in the syntax, the problem with `HermitianMatrixSpace()` is not that it's long, is that it's non-standard. Both `Symmetric` and `Hermitian` are not only the obvious thing to write, but also what the user already knows from LinearAlgebra. `HermitianMatrixSpace()` forces the uses to learn JuMP-only syntax, and it's not particularly easy to remember.

For comparison, YALMIP's syntax is `x = sdpvar(d,d,'hermitian','complex')` and CVX's is `variable x(d,d) hermitian`.

odow:
Monthly call says go for this.

odow:
@blegat how's this?

blegat:
Looks good to me, we can merge

mlubin (approved):

Issue #3287, opened 2023-03-16
Cannot multiply I by a variable and add it to a matrix.


araujoms:
If I create a variable `@variable(model, x)` and a numerical matrix `a = randn(Float64,(2,2))`, the expression `a+x*I` gives me the error `MethodError: no method matching +(::Matrix{Float64}, ::AffExpr)`.

Of course, it works without any problem if instead of `I` one uses an explicit identity matrix like `id = Symmetric(Matrix{Float64}(I,2,2))`

odow:
> We should probably create a Diagonal matrix here.

How would we know the dimension of `x * I`?

odow:
There's a few things here to fix:

```Julia
julia> using JuMP, LinearAlgebra

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: NO_OPTIMIZER
Solver name: No optimizer attached.

julia> @variable(model, x)
x

julia> a = rand(2, 2)
2×2 Matrix{Float64}:
 0.191126  0.178202
 0.71827   0.63977

julia> @expression(model, a + x * I)
ERROR: MethodError: no method matching zero(::Type{Matrix{Float64}})
Closest candidates are:
  zero(::Union{Type{P}, P}) where P<:Dates.Period at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Dates/src/periods.jl:53
  zero(::GenericQuadExpr) at /Users/oscar/.julia/dev/JuMP/src/quad_expr.jl:113
  zero(::GenericAffExpr) at /Users/oscar/.julia/dev/JuMP/src/aff_expr.jl:200
  ...
Stacktrace:
  [1] _instantiate_zero(#unused#::Type{Matrix{Float64}})
    @ MutableArithmetics ~/.julia/packages/MutableArithmetics/geMUn/src/interface.jl:28
  [2] promote_operation_fallback(op::typeof(+), #unused#::Type{Matrix{Float64}}, #unused#::Type{AffExpr})
    @ MutableArithmetics ~/.julia/packages/MutableArithmetics/geMUn/src/interface.jl:51
  [3] promote_operation(::typeof(+), ::Type, ::Type)
    @ MutableArithmetics ~/.julia/packages/MutableArithmetics/geMUn/src/interface.jl:113
[...]

PR #3295, opened 2023-03-19
Fix missing MA.promote_operation method with LinearAlgebra.I

odow:
Part of #3287

### DIFF
--- a/test/test_expr.jl
+++ b/test/test_expr.jl
@@ -441,4 +441,12 @@ function test_expression_ambiguities()
     return
 end
 
+function test_linear_algebra_scaling_matrix()
+    model = Model()
+    @variable(model, x)
+    A = [1 2; 3 4]
+    @test @expression(model, A + x * I(2)) == A + x * I(2) == A + [1 0; 0 1] * x
+    return
+end
+
 end  # TestExpr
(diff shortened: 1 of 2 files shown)


test/test_expr.jl
```diff
@@ -441,4 +441,12 @@ function test_expression_ambiguities()
     return
 end
 
+function test_linear_algebra_scaling_matrix()
+    model = Model()
+    @variable(model, x)
+    A = [1 2; 3 4]
+    @test @expression(model, A + x * I(2)) == A + x * I(2) == A + [1 0; 0 1] * x
```
odow, line 448:
```suggestion
    I = LinearAlgebra.I(2)
    @test @expression(model, A + x * I) == A + x * I == A + [1 0; 0 1] * x
```

blegat:
The right fix is to define `similar_array_type` for `Diagonal` like
https://github.com/jump-dev/MutableArithmetics.jl/blob/master/src/implementations/LinearAlgebra.jl#L151-L156
https://github.com/jump-dev/MutableArithmetics.jl/blob/master/src/implementations/LinearAlgebra.jl#L158
https://github.com/jump-dev/MutableArithmetics.jl/blob/master/src/implementations/LinearAlgebra.jl#L442-L447

odow:
Closing in favor of https://github.com/jump-dev/MutableArithmetics.jl/pull/205

PR #3296, opened 2023-03-22
Small refactoring for hermitian check

blegat:
Less duplicate code is better

### DIFF
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -372,13 +372,6 @@ function build_variable(
     variables::Matrix{<:AbstractVariable},
     ::HermitianMatrixSpace,
 )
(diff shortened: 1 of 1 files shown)


odow:
> Less duplicate code is better

I don't agree. It moves the error into a utility function, when it should really be in `build_variable`. I think that makes it less readable. It's not obvious that `_vectorize_complex_variables` should throw an error if one of them is discrete.

blegat:
If you have discrete constraints then in the vectorization itself, you should have to handle them, e.g., decide the value of the imag part of a boolean variable. Since the utility function does not handle that yet, it makes sense that it errors in that case.

PR #3332, opened 2023-04-27
Improve printing of complex numbers

blegat:
I tried to make it consistent with both how `Base` prints complex numbers (`im` is just on the right of the imaginary part without any `*` in between) and how JuMP prints real coefficients.

### DIFF
--- a/docs/src/manual/complex.md
+++ b/docs/src/manual/complex.md
@@ -21,7 +21,7 @@ Create a complex-valued variable using [`ComplexPlane`](@ref):
 julia> model = Model();
 
 julia> @variable(model, x in ComplexPlane())
-real(x) + im imag(x)
+real(x) + imag(x) im
 ```
 
 Note that `x` is not a [`VariableRef`](@ref); instead, it is an affine
@@ -64,7 +64,7 @@ To create an anonymous variable, use the `set` keyword argument:
 julia> model = Model();
 
 julia> x = @variable(model, set = ComplexPlane())
-_[1] + im _[2]
+_[1] + _[2] im
 ```
 
 ## Complex-valued variable bounds
@@ -85,7 +85,7 @@ julia> @variable(
            upper_bound = 2.0 + 3.0im,
            start = 4im,
        )
-real(x) + im imag(x)
+real(x) + imag(x) im
 
 julia> vars = all_variables(model)
 2-element Vector{VariableRef}:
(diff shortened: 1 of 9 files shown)


docs/src/manual/complex.md
```diff
@@ -21,7 +21,7 @@ Create a complex-valued variable using [`ComplexPlane`](@ref):
 julia> model = Model();
 
 julia> @variable(model, x in ComplexPlane())
-real(x) + (0.0 + 1.0im) imag(x)
+real(x) + im imag(x)
```
odow, line 24: I don't know if I like this. Wouldn't `real(x) + imag(x) im` make more sense?

docs/src/manual/complex.md
```diff
@@ -125,7 +125,7 @@ julia> set_silent(model)
 julia> @variable(model, x[1:2]);
 
 julia> @constraint(model, (1 + 2im) * x[1] + 3 * x[2] == 4 + 5im)
-(1.0 + 2.0im) x[1] + (3.0 + 0.0im) x[2] = 4.0 + 5.0im
+(1 + 2im) x[1] + 3 x[2] = 4.0 + 5.0im
```
odow, line 128: I went for `1.0 + 2.0im` so that we didn't have differences depending on the left-hand and right-hand sides of expressions.

blegat: That applies for `Float64` also then. We should apply our pretty printing to the rhs as well then

docs/src/manual/complex.md
blegat: Yes, that would look better, good point

docs/src/manual/complex.md
blegat: `@constraint(m, 2x == 2)` would otherwise print as `2 x == 2.0` right ?

blegat: Moved to a separate issue: https://github.com/jump-dev/JuMP.jl/issues/3339

odow:
If we only have `Float64` and `Complex`, then this makes more sense than what we had before. But I wonder if we should instead unify our coefficient printing and simplify: https://github.com/jump-dev/JuMP.jl/pull/3341.

blegat:
I think it's nice to have at least optional rounding. Now we can discuss whether we want it by default and I think would be in favor

odow:
I'll fix the conflict, since it's my fault for https://github.com/jump-dev/JuMP.jl/pull/3341.

odow (approved):

PR #3337, opened 2023-05-01
Replace normal equivalence with object equivalence in `memoize`

franckgaga:
It seems that object equivalence is more robust with some specific corner case on Duals. See my reply on discourse : https://discourse.julialang.org/t/cache-results-between-function-calls-in-optimization-jl/94964/12?u=franckgaga The simple example of the doc was working well with normal equivalence but that's not always the case. I'm not able to pinpoint in which situation it fails, but it's happening when `x==last_dx` is `true` BUT `typeof(x)==typeof(last_dx)` is `false`.


odow:
Thanks! Yeah I just saw your discourse post and came here to make the same change :smile:

odow:
Thanks!

Issue #3319, opened 2023-04-12
[Containers] bugs in sum(::DenseAxisArray; dims)


slwu89:
Hello, I'm on JuMP v1.10.0 and Julia 1.8.5 and experiencing the following strange behavior. Here is my MWE which gives an error when I try to sum over one dimension of a DenseAxisArray:

```julia
using JuMP, HiGHS
Items = 1:20
Bags = 1:10

model = Model(HiGHS.Optimizer)
@variable(model, bag_assignment[Bags,Items], Bin)
sum(bag_assignment,dims=2)
```

If however I directly put a range in the variable declaration rather than defining it as a variable, everything works as expected.

```julia
Items = 20
Bags = 10

model = Model(HiGHS.Optimizer)
@variable(model, bag_assignment[1:Bags,1:Items], Bin)
sum(bag_assignment,dims=2)
```

odow:
The `DenseAxisArray` container has a number of limitations compared to `Array`.

If you know the indices can be represented by an array, the easiest solution for you is `@variable(model, bag_assignment[Bags,Items], Bin, container = Array)`. See https://jump.dev/JuMP.jl/stable/manual/variables/#variable_forcing


There's a few related bugs:
```julia
julia> Containers.@container(x[i=1:2, j=1:2], i+j, container = DenseAxisArray)
2-dimensional DenseAxisArray{Int64,2,...} with index sets:
    Dimension 1, Base.OneTo(2)
    Dimension 2, Base.OneTo(2)
And data, a 2×2 Matrix{Int64}:
 2  3
 3  4

julia> sum(x; dims = 2)
ERROR: KeyError: key CartesianIndex(1,) not found
Stacktrace:
  [1] getindex(#unused#::JuMP.Containers._AxisLookup{Base.OneTo{Int64}}, key::CartesianIndex{1})
    @ JuMP.Containers ~/.julia/dev/JuMP/src/Containers/DenseAxisArray.jl:18
[...]

julia> A = 1:2
1:2

julia> Containers.@container(x[i=A, j=A], i+j, container = DenseAxisArray)
2-dimensional DenseAxisArray{Int64,2,...} with index sets:
    Dimension 1, 1:2
    Dimension 2, 1:2
And data, a 2×2 Matrix{Int64}:
 2  3
 3  4

julia> sum(x; dims = 2)
ERROR: ArgumentError: No method is implemented for reducing index range of type UnitRange{Int64}. Please implement
[...]

PR #3338, opened 2023-05-02
[Containers] throw nice error on sum(::DenseAxisArray; dims)

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3319

Taking a deeper look at fixing #3319, we actually run into the problem that `Base` doesn't drop flattened dimensions. So sum on a matrix still returns a `Matrix`:
```Julia
julia> x = [3 4; 4 5]
2×2 Matrix{Int64}:
 3  4
 4  5

julia> sum(x; dims = 1)
1×2 Matrix{Int64}:
 7  9

julia> sum(x; dims = 2)
2×1 Matrix{Int64}:
 7
 9
```
This happens even if the flattened dimension is in the middle:
```Julia
julia> x = rand(2, 2, 2)
2×2×2 Array{Float64, 3}:
[:, :, 1] =
 0.997592  0.424427
 0.673536  0.116184

[:, :, 2] =
 0.19444    0.853501
 0.0667024  0.725102

julia> sum(x; dims = 2)
2×1×2 Array{Float64, 3}:
[:, :, 1] =
 1.4220191127794288
 0.7897194338439628

[:, :, 2] =
 1.047941194519167
 0.7918042254294069
```

Our problem, then, is that what is the key for a flattened dimension of `DenseAxisArray`? We don't have a good answer, so I've opted to throw a nicer error asking the user to convert to a regular Array.

### DIFF
--- a/test/Containers/test_DenseAxisArray.jl
+++ b/test/Containers/test_DenseAxisArray.jl
@@ -846,7 +846,11 @@ function test_containers_denseaxisarrayview_kwarg_setindex()
 end
 
 function test_sum_dims()
-    Containers.@container(x[i=1:2, j=1:2], i+j, container = DenseAxisArray)
+    Containers.@container(
+        x[i=1:2, j=1:2],
+        i+j,
+        container = DenseAxisArray,
+    )
     @test_throws(
         ErrorException(
             "`sum(x::DenseAxisArray; dims)` is not supported. Convert the array " *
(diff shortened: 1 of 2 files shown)


test/Containers/test_DenseAxisArray.jl
```diff
@@ -845,4 +845,17 @@ function test_containers_denseaxisarrayview_kwarg_setindex()
     return
 end
 
+function test_sum_dims()
+    Containers.@container(x[i=1:2, j=1:2], i+j, container = DenseAxisArray)
```
odow, line 849:
```suggestion
    Containers.@container(
        x[i=1:2, j=1:2],
        i+j,
        container = DenseAxisArray,
    )
```

PR #3340, opened 2023-05-03
Prep for v1.11.0

odow:
- [x] extension-tests https://github.com/jump-dev/JuMP.jl/actions/runs/4867072087

### DIFF
--- a/docs/src/tutorials/linear/multi_objective_examples.jl
+++ b/docs/src/tutorials/linear/multi_objective_examples.jl
@@ -30,11 +30,7 @@ set_silent(model)
 @objective(model, Min, [3x1 + x2, -x1 - 2x2])
 @constraint(model, 3x1 - x2 <= 6)
 set_optimizer(model, () -> MOA.Optimizer(HiGHS.Optimizer))
-set_attribute(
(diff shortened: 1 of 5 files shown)


odow:
cc @pulsipher it's good having the `extension-tests.yml` now.

docs/src/tutorials/linear/multi_objective_examples.jl
```diff
@@ -30,11 +30,7 @@ set_silent(model)
 @objective(model, Min, [3x1 + x2, -x1 - 2x2])
 @constraint(model, 3x1 - x2 <= 6)
 set_optimizer(model, () -> MOA.Optimizer(HiGHS.Optimizer))
-set_attribute(
```
mlubin, line 33: Intended diff?

mlubin (approved):

docs/src/tutorials/linear/multi_objective_examples.jl
odow: Yip. I just noticed when I checked the versions we were using. Didn't seem worth a separate PR when I was doing this one.

PR #3341, opened 2023-05-03
Use _string_round when printing sets

### LINKED ISSUE #3339: Rounding constant in printing
We should probably call `_string_round` here:

https://github.com/jump-dev/JuMP.jl/blob/e87f315c7d3b429ef36a94dc70ea4f6b2a6c56a0/src/print.jl#L660-L679

See https://github.com/jump-dev/JuMP.jl/pull/3332#discussion_r1179699623

odow:
Closes #3339 

I don't know about this. If we're thinking of supporting arbitrary number types in JuMP, I'd lean the other way, and _remove_ the rounding when printing?

### DIFF
--- a/docs/src/developers/extensions.md
+++ b/docs/src/developers/extensions.md
@@ -214,7 +214,7 @@ julia> function JuMP.parse_constraint_head(
 
 julia> @constraint(model, x + x := 1.0)
 Rewriting := as ==
-2 x = 1.0
+2 x = 1
 ```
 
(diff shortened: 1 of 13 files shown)


blegat (approved):
I'm in favor of this. In the future, we can have printing options and make it optional but I would round by default. Even if power users might be interested to see that these are floating point numbers, most users like the abstraction and prefer to thing about math instead of having to worry about technical details.

odow:
I guess we can aways revisit when we add generic number support.

PR #2107, opened 2019-11-23
Replace parse_expr and destructive_add by MutableArithmetics

blegat:
**BREAKING CHANGES**

* `@SDconstraint(model, A >= 1)` throws `ERROR: Operation `+` between `Array{VariableRef,2}` and `Int64` is not allowed. You should use broadcast.` instead of broadcasting, `@SDconstraint(model, A >= 0)` is equivalent to `@contsraint(model, A in PSDCone())`
* https://github.com/JuliaOpt/JuMP.jl/issues/2106

JuMP implements an arithmetic between the following types:
* `Number`s (that are eventually converted to `Float64`)
* `AbstractVariableRef`
* `GenericAffExpr`
* `GenericQuadExpr`.

This arithmetic has two key features:
1) The return type of an operation is delicate to compute and we want it to be as "simple" as possible.
2) `GenericAffExpr` and `GenericQuadExpr` are mutable objects and creating new ones is costly.

The consequence of 1) is that many `Base` functions do not work out of the box (see e.g. https://github.com/JuliaLang/julia/pull/26344 and https://github.com/JuliaLang/julia/pull/26344).
The reason is that these are usually tested with `Float64` types which are stable under multiplications, addition, ...
The common mistakes are
* either the resulting type which is not inferred correctly (the code just promotes the type but does not take into account that there will be operations on them)
* either the different operands are first promoted to the same type. Which is a problem, if for instance we take the product of a `VariableRef` and a `Float64`, because of the promotion we will get at `QuadExpr` as result. This is the case in many places in `SparseArrays`.

For these reasons, many Base functions need to be rewritten for JuMP types to make them work.

The consequence of 2) is that even if generic functions are implemented correctly, they cannot exploit the mutability of the JuMP expression without have JuMP as a dependency. This is obviously not possible for `Base` but it may also be cumbersome in other cases (e.g. MultivariatePolynomials supports any coefficient type and don't want to have JuMP as a dependency just because the coefficients could be JuMP expressions).

JuMP contains code to:
* Fix `Base` function for JuMP types and make them efficient (addresses 1) and 2)).
* Rewrite expressions in macros into a code that exploits the mutability of the JuMP expressions.

This code is quite involved and could benefit other mutable types as well (e.g. `BigInt`, polynomials, MOI functions, ...).

[...]

PR #3349, opened 2023-05-09
Speed up vectorization of symmetric matrices

blegat:
Before https://github.com/jump-dev/JuMP.jl/pull/2107, we didn't do this conversion, I guess i just did it like to to make it similar to the `SquareMatrixShape` case but this gives unnecessary allocations if the matrix is already `Symmetric` and even more if the matrix is sparse as in https://discourse.julialang.org/t/out-of-memory-when-constructing-large-sparse-sdp-in-jump/98332/2

## Benchmark

Consider the following benchmark:
```julia
n = 1000
p = 0.01
C = sprandn(Float64, n, n, p)  
model = Model()
@variable(model, y[1:n])
@time @constraint(model, Symmetric(C - spdiagm(y)) in PSDCone())
```
On master, this gives
```
0.582551 seconds (10.06 M allocations: 696.425 MiB, 53.21% gc time)
```
After this PR, this gives
```
0.131655 seconds (2.58 M allocations: 186.525 MiB, 33.82% gc time)
```
The MOI way:
```julia
@time begin
func = MOI.VectorAffineFunction(
    [MOI.VectorAffineTerm(i, MOI.ScalarAffineTerm(1.0, index(y[i]))) for i in 1:n],
    JuMP.vectorize(C, SymmetricMatrixShape(n)),
)
set = MOI.PositiveSemidefiniteConeTriangle(n)
MOI.add_constraint(backend(model), func, set)
end
```
it gives
```
0.006370 seconds (33 allocations: 9.093 MiB)
```

### DIFF
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -180,7 +180,7 @@ function reshape_set(
 )
     return PSDCone()
 end
-function vectorize(matrix::Matrix, ::SymmetricMatrixShape)
+function vectorize(matrix, ::SymmetricMatrixShape)
     n = LinearAlgebra.checksquare(matrix)
     return [matrix[i, j] for j in 1:n for i in 1:j]
 end
@@ -219,7 +219,7 @@ function reshape_vector(
     return matrix
 end
 
-function vectorize(matrix::Matrix, ::SkewSymmetricMatrixShape)
+function vectorize(matrix, ::SkewSymmetricMatrixShape)
     n = LinearAlgebra.checksquare(matrix)
     return [matrix[i, j] for j in 1:n for i in 1:j-1]
 end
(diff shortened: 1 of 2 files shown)


odow:
Could we add a test of some sort?

odow:
Does this improve that benchmark?

src/sd.jl
```diff
@@ -246,7 +246,7 @@ function reshape_set(::MOI.PositiveSemidefiniteConeSquare, ::SquareMatrixShape)
 end
 vectorize(matrix::Matrix, ::SquareMatrixShape) = vec(matrix)
 
-function vectorize(matrix, shape::Union{SymmetricMatrixShape,SquareMatrixShape})
+function vectorize(matrix, shape::SquareMatrixShape)
```
odow, line 249: I also wonder if it's necessary to call `Matrix(vec(matrix))`. (This method and the one above.) Can we get away with just `vec(matrix)`?

blegat:
> Could we add a test of some sort?

Yes, I think we could have some sort of custom JuMP scalar that count the number of times it is copied.

src/sd.jl
blegat: If it's not a `Matrix` then it's going to build a `ReshapedArray`

odow:
> After this PR, this gives
> 0.131655 seconds (2.58 M allocations: 186.525 MiB, 33.82% gc time)

> [MOI] gives
> 0.006370 seconds (33 allocations: 9.093 MiB)

I guess the underlying problem is that JuMP doesn't have a way of passing sparse vectors to MOI.

PR #3362, opened 2023-05-18
[docs] add COPT to the documentation

odow:
https://jump.dev/JuMP.jl/previews/PR3362/packages/COPT/

- [ ] https://github.com/COPT-Public/COPT.jl/pull/25


odow:
Adding for now. Can update in a future commit.

PR #3369, opened 2023-05-21
Fix support for matrix in HermitianPSDCone

### LINKED ISSUE #3367: Error adding Matrix in HermitianPSDCone
A user sent me this example:
```Julia
julia> using JuMP, LinearAlgebra

julia> model = Model();

julia> @variable(model, X[1:2, 1:2] in HermitianPSDCone())
2×2 Hermitian{GenericAffExpr{ComplexF64, VariableRef}, Matrix{GenericAffExpr{ComplexF64, VariableRef}}}:
 real(X[1,1])                    real(X[1,2]) + imag(X[1,2]) im
 real(X[1,2]) - imag(X[1,2]) im  real(X[2,2])a

julia> @variable(model, t)
t

julia> X + I(2) * t
2×2 Matrix{GenericAffExpr{ComplexF64, VariableRef}}:
 real(X[1,1]) + t                real(X[1,2]) + imag(X[1,2]) im
 real(X[1,2]) - imag(X[1,2]) im  real(X[2,2]) + t

julia> @constraint(model, X + I(2) * t in JuMP.HermitianPSDCone())
ERROR: At REPL[20]:1: `@constraint(model, X + I(2) * t in JuMP.HermitianPSDCone())`: Unrecognized constraint building format. Tried to invoke `build_constraint(error, GenericAffExpr{ComplexF64, VariableRef}[t + real(X[1,1]) real(X[1,2]) + imag(X[1,2]) im; real(X[1,2]) - imag(X[1,2]) im t + real(X[2,2])], HermitianPSDCone())`, but no such method exists. This is due to specifying an unrecognized function, constraint set, and/or extra positional/keyword arguments.

If you're trying to create a JuMP extension, you need to implement `build_constraint` to accomodate these arguments.
Stacktrace:
 [1] error(::String, ::String)
   @ Base ./error.jl:42
 [2] _macro_error(macroname::Symbol, args::Tuple{Symbol, Expr}, source::LineNumberNode, str::String)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:1850
 [3] (::JuMP.var"#_error#80"{Tuple{Symbol, Expr}, Symbol, LineNumberNode})(str::String)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:897
 [4] build_constraint(::JuMP.var"#_error#80"{Tuple{Symbol, Expr}, Symbol, LineNumberNode}, ::Matrix{GenericAffExpr{ComplexF64, VariableRef}}, ::HermitianPSDCone; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:687
 [5] build_constraint(::Function, ::Matrix{GenericAffExpr{ComplexF64, VariableRef}}, ::HermitianPSDCone)
   @ JuMP ~/.julia/dev/JuMP/src/macros.jl:683
 [6] macro expansion
   @ ~/.julia/dev/JuMP/src/macros.jl:987 [inlined]
 [7] top-level scope
   @ REPL[20]:1

julia> @constraint(model, Hermitian(X + I(2) * t) in JuMP.HermitianPSDCone())
[real(X[1,1]) + t                real(X[1,2]) + imag(X[1,2]) im;
 real(X[1,2]) - imag(X[1,2]) im  real(X[2,2]) + t] ∈ HermitianPSDCone()
```

The error is bad, and they reasonably expected this to work since
```julia
julia> Y = X + I(2) * t
2×2 Matrix{GenericAffExpr{ComplexF64, VariableRef}}:
 real(X[1,1]) + t                real(X[1,2]) + imag(X[1,2]) im
 real(X[1,2]) - imag(X[1,2]) im  real(X[2,2]) + t

julia> ishermitian(Y)
true
```

odow:
Closes #3367

### DIFF
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -679,6 +679,23 @@ function build_constraint(
     )
 end
 
+function build_constraint(
+    _error::Function,
+    Q::AbstractMatrix{<:AbstractJuMPScalar},
+    cone::HermitianPSDCone,
+)
+    if !LinearAlgebra.ishermitian(Q)
+        _error(
+            "Unable to add matrix in HermitianPSDCone because the matrix is " *
+            "not Hermitian. If this error was unexpected, check the matrix " *
+            "for errors, or wrap the matrix `H` in " *
+            "`LinearAlgebra.Hermitian(H)` to force JuMP to treat the " *
+            "matrix as Hermitian.",
+        )
+    end
+    return build_constraint(_error, LinearAlgebra.Hermitian(Q), cone)
+end
+
 function build_constraint(_error::Function, H::LinearAlgebra.Hermitian, ::Zeros)
     n = LinearAlgebra.checksquare(H)
     shape = HermitianMatrixShape(n)
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -684,16 +684,11 @@ function build_constraint(
     Q::AbstractMatrix{<:AbstractJuMPScalar},
     cone::HermitianPSDCone,
 )
-    if !LinearAlgebra.ishermitian(Q)
-        _error(
-            "Unable to add matrix in HermitianPSDCone because the matrix is " *
-            "not Hermitian. If this error was unexpected, check the matrix " *
-            "for errors, or wrap the matrix `H` in " *
-            "`LinearAlgebra.Hermitian(H)` to force JuMP to treat the " *
-            "matrix as Hermitian.",
-        )
-    end
-    return build_constraint(_error, LinearAlgebra.Hermitian(Q), cone)
+    return _error(
+        "Unable to add matrix in HermitianPSDCone because the matrix is " *
+        "not a subtype of `LinearAlgebra.Hermitian`. To fix, wrap the matrix " *
+        "`H` in `LinearAlgebra.Hermitian(H)`.",
+    )
 end
 
 function build_constraint(_error::Function, H::LinearAlgebra.Hermitian, ::Zeros)
--- a/test/test_complex.jl
+++ b/test/test_complex.jl
@@ -292,7 +292,7 @@ function test_numerically_hermitian_matrix_error()
     Y = X + LinearAlgebra.I(2) * t
     @test !LinearAlgebra.ishermitian(Y)
     @test !(Y isa LinearAlgebra.Hermitian)
-    @test_macro_throws(
+    @test_throws_strip(
         ErrorException(
             "In `@constraint(model, Y in HermitianPSDCone())`: " *
             "Unable to add matrix in HermitianPSDCone because the matrix is " *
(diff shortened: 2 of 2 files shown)


test/test_complex.jl
```diff
@@ -273,4 +273,37 @@ function test_isreal()
...
+end
+
+function test_numerically_hermitian_matrix_error()
+    model = Model()
+    @variable(model, X[1:2, 1:2])
+    @variable(model, t)
+    Y = X + LinearAlgebra.I(2) * t
+    @test !LinearAlgebra.ishermitian(Y)
+    @test !(Y isa LinearAlgebra.Hermitian)
+    @test_macro_throws(
```
odow, line 295:
```suggestion
    @test_throws_strip(
```

src/sd.jl
```diff
@@ -679,6 +679,23 @@ function build_constraint(
...
+function build_constraint(
+    _error::Function,
+    Q::AbstractMatrix{<:AbstractJuMPScalar},
+    cone::HermitianPSDCone,
+)
+    if !LinearAlgebra.ishermitian(Q)
+        _error(
+            "Unable to add matrix in HermitianPSDCone because the matrix is " *
+            "not Hermitian. If this error was unexpected, check the matrix " *
+            "for errors, or wrap the matrix `H` in " *
```
blegat, line 691: It's unclear what you mean by "check the matrix for errors"

blegat: "because the matrix is of type `LinearAlgebra.Hermitian`. Even if the matrix is `H`, this need to be made explicit by wrapping it as `LinearAlgebra.Hermtian(H)`."

odow: Note that this method does check for `LinearAlgebra.ishermitian`

blegat: Ah, didn't catch that, I would just error all the time. This is a remake of https://github.com/jump-dev/JuMP.jl/pull/3264

odow: Ah. I thought we'd done it before, but I couldn't find.

Remind me again why? Just because it's inconsistent if you query `dual(con)` that you get a `Hermitian` matrix back when you passed a non-Hermitian square matrix? 

It really seems weird that the case in https://github.com/jump-dev/JuMP.jl/issues/3367 doesn't work.

blegat: If we want it to work, we can add the square cone in MOI. The issue with this PR is that when we do that then the type returned by `dual` will change in JuMP so that will be breaking.
So it's best to have an issue here and implement the square cone in MOI if there is a need for it.

odow: Updated to just throw the error

blegat (approved):

PR #3377, opened 2023-05-23
Add GenericModel{T} and GenericVariableRef{T}

blegat:
Following https://github.com/jump-dev/JuMP.jl/pull/3191#issuecomment-1558175946, this PR takes out from https://github.com/jump-dev/JuMP.jl/pull/3191 the part of **src** that should be boilerplate and not prone to generate argument to allow the important parts to stand out from https://github.com/jump-dev/JuMP.jl/pull/3191 and also to be able to merge this big part that is prone to create conflicts.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -204,19 +204,23 @@ jump_api_reference = DocumenterReference.automatic_reference_documentation(;
     subdirectory = "api",
     modules = [
         JuMP => [
-            "Base.empty!(::Model)" => DocumenterReference.DOCTYPE_FUNCTION,
-            "Base.isempty(::Model)" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Base.empty!(::GenericModel)" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Base.isempty(::GenericModel)" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
             "Base.copy(::AbstractModel)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
-            "Base.write(::IO, ::Model; ::MOI.FileFormats.FileFormat)" =>
+            "Base.write(::IO, ::GenericModel; ::MOI.FileFormats.FileFormat)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
-            "Base.read(::IO, ::Type{Model}; ::MOI.FileFormats.FileFormat)" =>
+            # Documenter tries to add `read(::IO, ::Type)` instead of this
+            # specific method
+            # "Base.read(::IO, ::Type{<:GenericModel}; ::MOI.FileFormats.FileFormat)" =>
+            #     DocumenterReference.DOCTYPE_FUNCTION,
+            "MOI.Utilities.reset_optimizer(::GenericModel)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
-            "MOI.Utilities.reset_optimizer(::Model)" =>
+            "MOI.Utilities.drop_optimizer(::GenericModel)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
-            "MOI.Utilities.drop_optimizer(::Model)" =>
-                DocumenterReference.DOCTYPE_FUNCTION,
-            "MOI.Utilities.attach_optimizer(::Model)" =>
+            "MOI.Utilities.attach_optimizer(::GenericModel)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
         ],
         JuMP.Containers => [
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -247,7 +247,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Update versions of the documentation (#3185)
  - Tidy the import of packages and remove unnecessary prefixes (#3186) (#3187)
  - Refactor `src/JuMP.jl` by moving methods into more relevant files (#3188)
- - Fix docstring of [`Model`](@ref) not appearing in the documentation (#3198)
+ - Fix docstring of `Model` not appearing in the documentation (#3198)
 
 ## Version 1.6.0 (January 1, 2023)
 
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -247,7 +247,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Update versions of the documentation (#3185)
  - Tidy the import of packages and remove unnecessary prefixes (#3186) (#3187)
  - Refactor `src/JuMP.jl` by moving methods into more relevant files (#3188)
- - Fix docstring of `Model` not appearing in the documentation (#3198)
+ - Fix docstring of [`Model`](@ref) not appearing in the documentation (#3198)
 
 ## Version 1.6.0 (January 1, 2023)
 
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -44,7 +44,7 @@ julia> @variable(model, x)
 x
 
 julia> typeof(x)
-VariableRef
+VariableRef (alias for GenericVariableRef{Float64})
 
 julia> num_variables(model)
 1
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -140,7 +140,7 @@ end
 
 Create a new instance of a JuMP model.
 
-If `optimizer_factory` is provided, the model is initialized with thhe optimizer
+If `optimizer_factory` is provided, the model is initialized with the optimizer
 returned by `MOI.instantiate(optimizer_factory)`.
 
 If `optimizer_factory` is not provided, use [`set_optimizer`](@ref) to set the
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -73,7 +73,15 @@ An abstract type that should be subtyped for users creating JuMP extensions.
 """
 abstract type AbstractModel end
 
-mutable struct Model <: AbstractModel
+"""
+    value_type(::Type{<:Union{AbstractModel,AbstractVariableRef}})
+
+Return the return type of [`value`](@ref) for variables of that model. It
+defaults to `Float64` if it is not implemented.
+"""
+value_type(::Type{<:AbstractModel}) = Float64
+
+mutable struct GenericModel{T<:Real} <: AbstractModel
     # In MANUAL and AUTOMATIC modes, CachingOptimizer.
     # In DIRECT mode, will hold an AbstractOptimizer.
     moi_backend::MOI.AbstractOptimizer
@@ -84,7 +92,7 @@ mutable struct Model <: AbstractModel
     # same bridge may be added many times so we store them in a `Set` instead
     # of, e.g., a `Vector`.
     bridge_types::Set{Any}
-    # Hook into a solve call...function of the form f(m::Model; kwargs...),
+    # Hook into a solve call...function of the form f(m::GenericModel; kwargs...),
     # where kwargs get passed along to subsequent solve calls.
     optimize_hook::Any
     # TODO: Document.
@@ -105,7 +113,9 @@ mutable struct Model <: AbstractModel
     set_string_names_on_creation::Bool
 end
 
-function Base.getproperty(model::Model, name::Symbol)
+value_type(::Type{GenericModel{T}}) where {T} = T
+
+function Base.getproperty(model::GenericModel, name::Symbol)
     if name == :nlp_data
         error(
             "The internal field `.nlp_data` was removed from `Model` in JuMP " *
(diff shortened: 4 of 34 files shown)


src/constraints.jl
```diff
@@ -94,7 +94,7 @@ function _dual_start(
             <:MOI.AbstractScalarSet,
         },
     },
-)::Union{Nothing,Float64}
+)::Union{Nothing,Number}
```
odow, line 97: So this is an interesting one. What is the dual start of a `ScalarAffineFunction{Complex{Float64}}-in-EqualTo{Complex{Float64}` constraint?

src/JuMP.jl
```diff
@@ -73,7 +73,15 @@ An abstract type that should be subtyped for users creating JuMP extensions.
...
 abstract type AbstractModel end
 
-mutable struct Model <: AbstractModel
+"""
+    value_type(::Type{<:Union{AbstractModel,AbstractVariableRef}})
+
+Return the return type of [`value`](@ref) for variables of that model. It
+defaults to `Float64` if it is not implemented.
+"""
+value_type(::Type{<:AbstractModel}) = Float64
```
odow, line 82: I wondered about even moving this to the test PR

src/JuMP.jl
```diff
@@ -166,18 +179,21 @@ julia> import MultiObjectiveAlgorithms as MOA
...
+    add_bridges::Bool = true,
+) where {T}
+    model = GenericModel{T}()
     set_optimizer(model, optimizer_factory; add_bridges = add_bridges)
     return model
 end
 
 """
-    direct_model(backend::MOI.ModelLike)
+    direct_model(backend::MOI.ModelLike, T::Type = Float64)
```
odow, line 192: Positional argument, or `direct_model(model; value_type = Float64)`?

src/constraints.jl
```diff
@@ -577,8 +577,10 @@ representing the function and the `set` field contains the MOI set.
 See also the [documentation](@ref Constraints) on JuMP's representation of
 constraints for more background.
 """
-struct ScalarConstraint{F<:AbstractJuMPScalar,S<:MOI.AbstractScalarSet} <:
-       AbstractConstraint
+struct ScalarConstraint{
+    F<:Union{Number,AbstractJuMPScalar},
```
odow, line 581: Revert. Let's make this change later.

src/constraints.jl
```diff
@@ -612,7 +614,7 @@ See also the [documentation](@ref Constraints) on JuMP's representation of
 constraints.
 """
 struct VectorConstraint{
-    F<:AbstractJuMPScalar,
+    F<:Union{Number,AbstractJuMPScalar},
```
odow, line 617: Ditto

src/constraints.jl
```diff
@@ -1020,7 +1025,8 @@ function _constraint_primal(
         },
     },
     result::Int,
-)::Float64
+    ::Type{T},
+)::T where {T}
```
odow, line 1029: This is a bug: https://github.com/jump-dev/JuMP.jl/issues/3380

odow:
The last commit, https://github.com/jump-dev/JuMP.jl/pull/3377/commits/825459b9d177ed85c01b239ae1260ef6e08b835a, fixes a number of cases that still used `Model`, `VariableRef`, `AffExpr`, or `QuadExpr`. 

Since this PR is extracted from https://github.com/jump-dev/JuMP.jl/pull/3191, which is passing tests, I assume that this means we don't have good test coverage for non-Float64 number types.

src/constraints.jl
odow: Fixed in https://github.com/jump-dev/JuMP.jl/pull/3381. We should merge that first, and then rebase this PR on top.

docs/src/manual/constraints.md
```diff
@@ -924,7 +924,7 @@ julia> start_value(con)
 
 julia> dual_start_value(con)
 
-julia> set_dual_start_value(con, 2)
+julia> set_dual_start_value(con, 2.0)
```
odow, line 927:
```suggestion
julia> set_dual_start_value(con, 2)
```

src/JuMP.jl
odow: Or `; with_value_type = T`? To match `with_bridge_type` and `with_cache_type`?

src/constraints.jl
```diff
@@ -973,9 +973,9 @@ into account in the primal value of the constraint. For instance, the constraint
 evaluation of `2x + 3y`.
 """
 function value(
-    con_ref::ConstraintRef{<:AbstractModel,<:MOI.ConstraintIndex};
+    con_ref::ConstraintRef{M,<:MOI.ConstraintIndex};
```
odow, line 976:
```suggestion
    con_ref::ConstraintRef{<:AbstractModel,<:MOI.ConstraintIndex};
```

src/constraints.jl
```diff
@@ -973,9 +973,9 @@ into account in the primal value of the constraint. For instance, the constraint
...
     result::Int = 1,
-)
+) where {M<:AbstractModel}
```
odow, line 978:
```suggestion
)
```

src/constraints.jl
```diff
@@ -1037,9 +1039,9 @@ Use `has_dual` to check if a result exists before asking for values.
 See also: [`result_count`](@ref), [`shadow_price`](@ref).
 """
 function dual(
-    con_ref::ConstraintRef{<:AbstractModel,<:MOI.ConstraintIndex};
+    con_ref::ConstraintRef{M,<:MOI.ConstraintIndex};
```
odow, line 1042:
```suggestion
    con_ref::ConstraintRef{<:AbstractModel,<:MOI.ConstraintIndex};
```

src/constraints.jl
```diff
@@ -1037,9 +1039,9 @@ Use `has_dual` to check if a result exists before asking for values.
...
     result::Int = 1,
-)
+) where {M<:AbstractModel}
```
odow, line 1044:
```suggestion
)
```

odow (approved):
I think we should still hold off merging this for now. But I'm a lot more confident in these changes than when it was in the monolithic PR.

src/feasibility_checker.jl
```diff
@@ -74,10 +74,10 @@ end
 """
     primal_feasibility_report(
         point::Function,
-        model::Model;
-        atol::Float64 = 0.0,
+        model::GenericModel{t};
```
odow, line 77:
```suggestion
        model::GenericModel{T};
```

src/variables.jl
```diff
@@ -214,23 +216,40 @@ with variables of type `V<:AbstractVariableRef` and coefficients of type `T`
...
+
 Base.conj(v::AbstractVariableRef) = v
 Base.real(v::AbstractVariableRef) = v
 Base.imag(v::AbstractVariableRef) = zero(v)
 Base.abs2(v::AbstractVariableRef) = v^2
 Base.isreal(::AbstractVariableRef) = true
 
 """
-    VariableRef <: AbstractVariableRef
+    GenericVariableRef <: AbstractVariableRef
```
odow, line 238:
```suggestion
    GenericVariableRef{T} <: AbstractVariableRef
```

src/JuMP.jl
```diff
@@ -73,7 +73,15 @@ An abstract type that should be subtyped for users creating JuMP extensions.
...
-mutable struct Model <: AbstractModel
+"""
+    value_type(::Type{<:Union{AbstractModel,AbstractVariableRef}})
+
+Return the return type of [`value`](@ref) for variables of that model. It
+defaults to `Float64` if it is not implemented.
+"""
+value_type(::Type{<:AbstractModel}) = Float64
+
+mutable struct GenericModel{T} <: AbstractModel
```
odow, line 84: Can we use a more descriptive name instead of `T`

mlubin, line 84: Could we start off by allowing a fixed list of type parameters? Like `Float32`, `Float64`, `Rational{...}`, `BigFloat`?

odow: Do you mean for this particular PR, or the entire sequence of PRs?

mlubin: For this PR. I mean adding a constructor that will reject type parameters outside this list.

odow: Would we remove it in https://github.com/jump-dev/JuMP.jl/pull/3385?

It's fair to even restrict it to only `Float64` for this PR.

blegat: The issue is that one use case is to use solvers with https://github.com/JuliaMath/DoubleFloats.jl or https://github.com/JuliaMath/Quadmath.jl (especially for tricky SDPs). We don't want to use Julia's package extension for that :smile:

blegat: Maybe it's enough if it's in the docstring ? Having a small name might make the user thing there is a simple interpretation and no need to look up the docstring. At least with `T` it forces the user to read the docstring and get the full explanation and not get a wrong understanding ^^

mlubin: Could we at least document which types are expected to work?
What about the method error principle when someone uses a weird type?

odow: How about a docstring for `Model` and a separate one for `GenericModel`. We can keep the current one for `Model`. 

I think it's hard for JuMP to make assumptions about what will work. Can we just push this onto the solver? And say something along the lines of 

```
The direct use of `GenericModel` for a value type other than `Float64` is an
advanced operation. The value type of `GenericModel` must match that expected by
the chosen optimizer. Consult the optimizers documentation for details. If not
documented, assume that the optimizer supports only `GenericModel{Float64}`.
Choosing an unsupported value type will throw an [`MOI.UnsupportedConstraint`](@ref)
or an [`MOI.UnsupportedAttribute`](@ref) error, the timing of which (during
the model construction or during a call to [`optimize!`](@ref)) depends on how
the solver is interfaced to JuMP.
```

mlubin: That sounds good for documentation. What about restricting `T<:Number` or `T<:Real`?

blegat: `T<:Real` sounds good indeed.

kalmarek: I'd prefer `T<:Number` if we want to use `Arb`s in solvers

blegat: I talked about it offline with @kalmarek 
So `Arblib.Arf` is like `BigFloat`, `Arblib.Arb` is like `IntervalArithmetics.Interval{BigFloat}` but `Arblib.Acb` is different from `Complex{IntervalArithmetics.Interval{BigFloat}}` because `Acb` is a Euclidean ball while `Complex` would be a rectangular / norm infinity ball. Since multiplication by complex number is a rotation, it turns out that Euclidean balls work better than rectangular ball (where you loose accuracy by outer-approximating rotated rectangle by non-rotated rectangles).
The current design of the complex JuMP interface does not really support Acb at the moment so I think `Real` is fine for now.

blegat:
I guess it's a bit weird to have `GenericModel{Float64}(value_type = BigFloat)`.
Maybe `model(value_type = BigFloat)` and `direct_model(value_type = BigFloat)` ?

odow:
> `GenericModel{Float64}(value_type = BigFloat)`

Yeah this is weird. But people would really do
```Julia
julia> GenericModel{Float64}(value_type = Int)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> Model(value_type = Int)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
```

> model(value_type = BigFloat)

We cannot do this. It would export `model` and conflict with everyone using `model` as the name of their `Model`.

docs/src/reference/models.md
```diff
@@ -8,6 +8,7 @@ the manual.
 ```@docs
 Model
 direct_model
+GenericModel
```
odow, line 11:
```suggestion
GenericModel{T}
```

blegat:
> > model(value_type = BigFloat)

> We cannot do this. It would export `model` and conflict with everyone using `model` as the name of their `Model`.

Then what about `generic_model(value_type = BigFloat)` and `direct_model(value_type = BigFloat)` ?

odow:
The documented API detail is `Model(; value_type)`. It's the most JuMP-like API, and I don't think that's confusing, or that there is any ambiguity.

It's an implementation detail that `Model` is also a synonym for `GenericModel{Float64}`. We'd never actually document people using the `GenericModel{T}` constructor.

docs/src/changelog.md
```diff
@@ -205,7 +205,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Update versions of the documentation (#3185)
  - Tidy the import of packages and remove unnecessary prefixes (#3186) (#3187)
  - Refactor `src/JuMP.jl` by moving methods into more relevant files (#3188)
- - Fix docstring of [`Model`](@ref) not appearing in the documentation (#3198)
+ - Fix docstring of `Model` not appearing in the documentation (#3198)
```
odow, line 208:
```suggestion
 - Fix docstring of [`Model`](@ref) not appearing in the documentation (#3198)
```

blegat:
We need a keyword argument for `direct_model` as well. At the moment `Model` is hardcoded in `direct_model`. I always found `direct_model` vs `Model`  a bit inconsistent so I thought it was a good occasion to have `direct_model` vs `generic_model` but the `generic` vs `direct` is maybe not the best way to describe the difference between `direct_model` and `Model`.

I guess we should just check whether having the constructor `Model` not create a `Model` violates one of the Julia rules for constructors

src/JuMP.jl
```diff
@@ -195,9 +227,12 @@ in mind the following implications of creating models using this *direct* mode:
 * The optimizer used cannot be changed the model is constructed.
 * The model created cannot be copied.
 """
-function direct_model(backend::MOI.ModelLike)
+function direct_model(
+    backend::MOI.ModelLike;
+    value_type::Type{T} = Float64,
```
odow, line 232: @blegat I already added the kwarg for `direct_model`

odow:
> I guess we should just check whether having the constructor Model not create a Model violates one of the Julia rules for constructors

I don't think there's any technical issue with it. Stylistically it's weird only if you know that `Model` is an alias. We just never mention the `GenericModel` constructor.

blegat:
@mlubin Any opinion on this ?

docs/src/manual/variables.md
```diff
@@ -1230,7 +1230,7 @@ julia> y = Union{VariableRef,AffExpr}[
            j > i ? first(keys(x[i, j].terms)) : x[i, j]
            for i in 1:size(x, 1), j in 1:size(x, 2)
        ]
-2×2 Matrix{Union{VariableRef, AffExpr}}:
+2×2 Matrix{Union{AffExpr, VariableRef}}:
```
odow, line 1233:
```suggestion
2×2 Matrix{Union{VariableRef, AffExpr}}:
```

mlubin:
`Model(; value_type)` and `direct_model(; value_type)` seems reasonable to me. It's the least disruptive. Exporting `model` is a no-go because of so many local variables named `model`. . Will Julia specialize on keyword argument types in this context (i.e., will the return value of `Model(; value_type=T)` infer as `GenericModel{T}`)? 

Could we make `GenericModel{Float64}(value_type = BigFloat)` an explicit error?

odow:
- [ ] Check type inference of `Model()` and `Model(; value_type = BigFloat)`

odow:
So I guess the inference failure is a kicker:

image

Let me revert to what we had before.

docs/make.jl
```diff
@@ -204,19 +204,21 @@ jump_api_reference = DocumenterReference.automatic_reference_documentation(;
...
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Base.isempty(::GenericModel)" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
             "Base.copy(::AbstractModel)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
-            "Base.write(::IO, ::Model; ::MOI.FileFormats.FileFormat)" =>
+            "Base.write(::IO, ::GenericModel; ::MOI.FileFormats.FileFormat)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
-            "Base.read(::IO, ::Type{Model}; ::MOI.FileFormats.FileFormat)" =>
+            "Base.read(::IO, ::Type{GenericModel}; ::MOI.FileFormats.FileFormat)" =>
```
odow, line 215:
```suggestion
            "Base.read(::IO, ::Type{GenericModel{T}}; ::MOI.FileFormats.FileFormat)" =>
```

docs/make.jl
```diff
@@ -204,19 +204,21 @@ jump_api_reference = DocumenterReference.automatic_reference_documentation(;
...
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Base.isempty(::GenericModel)" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
             "Base.copy(::AbstractModel)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
-            "Base.write(::IO, ::Model; ::MOI.FileFormats.FileFormat)" =>
+            "Base.write(::IO, ::GenericModel; ::MOI.FileFormats.FileFormat)" =>
                 DocumenterReference.DOCTYPE_FUNCTION,
-            "Base.read(::IO, ::Type{Model}; ::MOI.FileFormats.FileFormat)" =>
+            "Base.read(::IO, ::Type{GenericModel{T}}; ::MOI.FileFormats.FileFormat)" =>
```
odow, line 215:
```suggestion
            "Base.read(::IO, ::Type{<:GenericModel}; ::MOI.FileFormats.FileFormat)" =>
```

blegat:
Looks good to merge

odow:
> Looks good to merge

Let's wait for @mlubin to take another look.

src/JuMP.jl
```diff
@@ -123,62 +133,67 @@ function Base.getproperty(model::Model, name::Symbol)
...
-[`optimize!`](@ref).
-"""
-function Model()
-    caching_opt = MOIU.CachingOptimizer(
-        MOIU.UniversalFallback(MOIU.Model{Float64}()),
-        MOIU.AUTOMATIC,
-    )
-    return direct_model(caching_opt)
-end
+If `optimizer_factory` is provided, the model is initialized with thhe optimizer
```
mlubin, line 143:
```suggestion
If `optimizer_factory` is provided, the model is initialized with the optimizer
```

mlubin (approved):

Issue #1350, opened 2018-06-19
modeling with SI units (via Unitful.jl)


chriscoey:
Something that has been mentioned before in passing and might be worth discussing down the track: allow modeling with SI units using https://github.com/ajkeller34/Unitful.jl. It may be fairly trivial, I'm not sure.

DifferentialEquations.jl seems to allow this, for example [here](https://github.com/JuliaDiffEq/DiffEqTutorials.jl/blob/719c21cd76d8133b9f6a62ecceb10dfcccd24123/ExtraODEFeatures/.ipynb_checkpoints/Unit%20Checked%20Arithmetic%20via%20Unitful-checkpoint.ipynb). 

GPKit is a geometric programming toolkit in Python and it [supports specification of units](https://gpkit.readthedocs.io/en/latest/gettingstarted.html?highlight=units) and automatically does dimensional analysis to convert or verify units in expressions/constraints. From the GPKit docs, an example of a variable declaration:
```
# Declare a variable, z, with units of meters, and a description
z = Variable("z", "m", "A variable called z with units of meters")
```
and a simple model:
```
# consider a block with dimensions x, y, z less than 1
# constrain surface area less than 1.0 m^2
x = Variable("x", "m")
y = Variable("y", "m")
z = Variable("z", "m")
S = Variable("S", 1.0, "m^2")
c = (2*x*y + 2*x*z + 2*y*z <= S)
```
(There are other ideas we could borrow from GPKit to make JuMP even more user friendly. I like the little solution tables it can print after a solve.)

blegat:
You can do this by extending the `@variable` macro as follows
```julia
using JuMP
using Unitful
m = Model()
function unitvar(vref::JuMP.VariableRef, u::Unitful.FreeUnits)
    JuMP.GenericAffExpr(0.0 * u, vref => 1.0 * u)
end
struct Unit
    u::Unitful.FreeUnits
end
struct UnitVariable <: JuMP.AbstractVariable
    v::JuMP.ScalarVariable
    u::Unitful.FreeUnits
end
function JuMP.buildvariable(_error::Function, info::JuMP.VariableInfo, u::Unit)
    UnitVariable(JuMP.ScalarVariable(info), u.u)
end
function JuMP.addvariable(m::Model, v::UnitVariable, name::String="")
    vref = JuMP.addvariable(m, v.v, name)
    # add bridge to `m.moibackend` if its does not support constraints with Unitful coefficients
[...]

PR #3383, opened 2023-05-24
DimensionalData extension

longemen3000:
adds `DimensionaData` as an extension, closes https://github.com/rafaqz/DimensionalData.jl/pull/487


Project.toml
```diff
@@ -19,8 +19,15 @@ OrderedCollections = "1"
...
 [extras]
+DimensionalData = "0703355e-b756-11e9-17c0-8b28908087d0"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test"]
+test = ["DimensionalData", "Test"]
+
+[weakdeps]
+DimensionalData = "0703355e-b756-11e9-17c0-8b28908087d0"
```
odow, line 33:
```suggestion
DimensionalData = "0703355e-b756-11e9-17c0-8b28908087d0"

```

odow, line 33: So what happens with this in Julia 1.6? It just gets ignored?

longemen3000: yes

longemen3000: the options for Backwards compatibility is by `Requires` or as a regular dependency

Project.toml
```diff
@@ -19,8 +19,15 @@ OrderedCollections = "1"
 SnoopPrecompile = "1"
 julia = "1.6"
 
+[extensions]
+JuMPDimensionalDataExt = "DimensionalData"
+
 [extras]
+DimensionalData = "0703355e-b756-11e9-17c0-8b28908087d0"
```
odow, line 26: TODO: add a [compat] entry

test/Containers/test_dimensionaldata.jl
```diff
@@ -0,0 +1,29 @@
...
+    @test_throws(
+        ArgumentError,
+        @variable(model, z[i = 1:3, j = i:2], container = DimArray),
+    )
+    @test_throws(
+        ArgumentError,
+        @variable(model, w[1:3, j = 1:2], container = DimArray),
+    )
+    return
+end
```
odow, line 29:
```suggestion
end
```

Project.toml
odow: I think I'm okay if this is a Julia 1.9+ feature only. We don't need backwards compatibility for new features.

One issue is that the docs are currently build using Julia 1.6. So we'd need to update that, and then add some big compat warnings.

odow:
I guess a heads up: it's not certain that we'll merge an extension like this just yet. It's a pretty big change for JuMP, not because of this PR, but because it opens the door for others like it. But we have a monthly call on Thursday that I'll discuss this in.

See also: https://github.com/jump-dev/JuMP.jl/issues/1350

ext/JuMPDimensionalDataExt.jl
```diff
@@ -0,0 +1,37 @@
...
+
+import DimensionalData
+import JuMP
+
+function JuMP.Containers.container(
+    f::F, 
+    indices::JuMP.Containers.VectorizedProductIterator,
+    c::Type{DimensionalData.DimArray},
+    names,
+) where {F}
```
odow, line 16:
```suggestion
) where {F<:Function}
```

test/Containers/test_dimensionaldata.jl
```diff
@@ -0,0 +1,31 @@
...
+    
+function test_dimension_data_integration()
+    if !isdefined(Base, :get_extension)
+        return
+    end
+    model = Model()
+    @variable(model, x[i = 2:4, j = ["a", "b"]], container = DimArray)
+    @test x isa DimArray
+    @test_throws(
+        ArgumentError,
```
odow, line 21:
```suggestion
        MethodError,
```

ext/JuMPDimensionalDataExt.jl
```diff
@@ -0,0 +1,37 @@
...
+        if i isa Symbol && !Base.isidentifier(i)
+            # A symbol was created automatically by JuMP.
+            throw(ArgumentError("Cannot build DimArray. invalid indices."))
+        end
+    end
+    dims = NamedTuple(i => j for (i, j) in zip(names, indices.prod.iterators))
+    return DimensionalData.DimArray(map(i -> f(i...), indices), dims)
+end
+
+# TODO: imporve error for nested iterators
```
odow, line 27:
```suggestion
# TODO: improve error for nested iterators
```

odow (commented):
Overall, this is pretty slick.

Before merging, we need a bunch more tests, nicer error messages, and some documentation.

But I'm in favor of adding this to JuMP. The main issue I see is a discoverability one. We'll likely need some careful documentation about how and when different extensions are loaded and what features they enable.

test/Containers/test_dimensionaldata.jl
```diff
@@ -0,0 +1,31 @@
...
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+module TestContainersDimensionalData
+
+using Test
+
+using DimensionalData
+using JuMP
+
```
odow, line 12:
```suggestion

```

test/Containers/test_dimensionaldata.jl
```diff
@@ -0,0 +1,31 @@
...
+module TestContainersDimensionalData
+
+using Test
+
+using DimensionalData
+using JuMP
+    
+function test_dimension_data_integration()
+    if !isdefined(Base, :get_extension)
+        return
```
odow, line 15:
```suggestion
        return  # Skip this test on Julia versions before v1.9
```

odow:
Seems like this is working:

Julia 1.6
image

Julia 1.9
image

odow:
Some notes from the JuMP call

 - [ ] What if extension makes a breaking release
 - [ ] Add documentation
 - [ ] Necessary criteria for acceptance
      - Bar for acceptance is high. Qualitative decision by JuMP developers
      - "Significant" package in the ecosystem
      - A useful extension for JuMP models
      - Mature: 1.0 release, _or_ battle tested for a long time
      - Self-contained

longemen3000:
>  What if extension makes a breaking release
You can add `compat` annotations for extensions (but aqua seems to have problems with those, In general, Aqua needs to be updated to support 1.9 extensions and weakdeps).

odow:
The compat issue was this:

 * What if DimensionalData releases a breaking `v0.25` that we don't want to (or can't) support? 
 * Are we stuck on v0.24 for ever? 
 * Can users install JuMP and v0.25 together and have the extension not load? 
 * Is it a breaking change if JuMP removes an extension?

rafaqz:
This is an interesting point about extensions in general for the ecosystem... we are making our dependency trees larger and that does have costs besides compile time.

FWIW DimensionalData.jl is unlikely to do anything you cant support or change the basics of the dimarray interface much in the next few years.

odow:
Based on my testing, it seems that the compat dependency issue is a hard requirement:
```julia
(dd) pkg> st
Status `/private/tmp/dd/Project.toml`
  [0703355e] DimensionalData v0.24.12
  [4076af6c] JuMP v1.11.1 `~/.julia/dev/JuMP`

(dd) pkg> add DimensionalData@0.23
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package DimensionalData [0703355e]:
 DimensionalData [0703355e] log:
 ├─possible versions are: 0.1.0-0.24.12 or uninstalled
 ├─restricted to versions 0.24 by JuMP [4076af6c], leaving only versions: 0.24.0-0.24.12 or uninstalled
 │ └─JuMP [4076af6c] log:
 │   ├─possible versions are: 1.11.1 or uninstalled
 │   └─JuMP [4076af6c] is fixed to version 1.11.1
 └─restricted to versions 0.23 by an explicit requirement — no versions left
```

Thus, I think we need to treat extensions as if we were adopting the dependency into JuMP proper. The lazy loading is almost an implementation detail, rather than an magic thing that gets loaded if the versions align.

This makes me a lot more hesitant to add extensions for packages which are not version 1, or which have large dependency trees.

In this case, DimensionalData is borderline, because it's pre-1.0, but it is fairly stable at v0.24.12, and it has a fairly reasonable set of dependencies: https://juliahub.com/ui/Packages/DimensionalData/Unppe/0.24.12?page=1

odow:
Closing in favor of https://github.com/jump-dev/JuMP.jl/pull/3413

PR #3386, opened 2023-05-25
[docs] refactor API reference

odow:
I've been talking to @mortenpi about this. Very rough and ready. I need to think about it a bit more.

https://jump.dev/JuMP.jl/previews/PR3386


odow:
@mortenpi take a look. I think I like it. It's better than what we currently have. The hard part is organizing. But maybe macro/function/struct/constant is okay. 

It's nice to be able to link to a specific API page: https://jump.dev/JuMP.jl/previews/PR3386/api/list_of_constraint_types/

I think this should really force us to write nicer docstrings.

docs/make.jl
```diff
@@ -420,6 +504,11 @@ _add_moi_pages()
 #  Check that we have included all the markdown files in _PAGES!
 # ==============================================================================
 
+# For Documenter.hide arguments
+function _add_to_set(set::Set{String}, arg::Tuple{Bool,String,String,Vector{Any}})
```
odow, line 508:
```suggestion
function _add_to_set(
    set::Set{String},
    arg::Tuple{Bool,String,String,Vector{Any}},
)
```

docs/.gitignore
```diff
@@ -10,3 +10,4 @@ src/packages/*.md
 !src/packages/solvers.md
 !src/packages/extensions.md
 !src/packages/NLopt.md
+src/api.md
```
odow, line 13:
```suggestion
```

docs/make.jl
```diff
@@ -186,6 +186,101 @@ sort!(_LIST_OF_EXTENSIONS; by = x -> (!startswith(x[1], "jump-dev/"), x[1]))
...
+    )
+    return sort(contents; by = x -> (order[x[2]], "$(x[1])"))
+end
+
+function build_api_reference(mod, src_dir, sub_dir; extras = Any[])
+    ref_io = open(joinpath(src_dir, sub_dir, "$mod.md"), "w")
+    println(ref_io, "# API")
+    println(ref_io)
+    println(ref_io, "| NAME | KIND |")
+    println(ref_io, "| ---- | ---- |")
```
odow, line 231:
```suggestion
    println(ref_io, "| :--- | :--- |")
```

docs/make.jl
```diff
@@ -186,6 +186,101 @@ sort!(_LIST_OF_EXTENSIONS; by = x -> (!startswith(x[1], "jump-dev/"), x[1]))
...
+                else
+                    error("Documentation missing for $key")
+                end
+            end
+        end
+        open(joinpath(src_dir, sub_dir, "$key.md"), "w") do io
+            return write(io, "```@docs\n$key\n```")
+        end
+        push!(reference, Documenter.hide("`$key`" => "$sub_dir/$key.md"))
+        println(ref_io, " | [`$key`](@ref) | `$type` | ")
```
odow, line 249:
```suggestion
        println(ref_io, "| [`$key`](@ref) | `$type` |")
```

docs/make.jl
```diff
@@ -505,7 +597,7 @@ function _remove_literate_footer(dir)
     return
 end
 
-if _PDF
+if false # _PDF
```
odow, line 600:
```suggestion
if _PDF
```

odow:
This is nice: https://jump.dev/JuMP.jl/previews/PR3386/api/JuMP/

I've chosen to follow the FastAI.jl docs and list the API as a table with "name" and "kind" columns: https://fluxml.ai/FastAI.jl/dev/FastAI@dev/ref/FastAI.html

I think any more complicated grouping (like putting `lower_bound` and `set_lower_bound` together) just gets difficult. We can link docstrings with "See also."

odow:
I wonder which docstring breaks the PDF build...

mortenpi:
> I wonder which docstring breaks the PDF build...

Something I find useful here is to upload the LaTeX logs as artifacts, e.g. in Documenter we do this: https://github.com/JuliaDocs/Documenter.jl/blob/233a7822d670b063e894f62aea00729bcb3a28dd/.github/workflows/CI.yml#L142-L147

I don't remember if that also uploaded the generated TeX source, but if it doesn't, it should.

mortenpi:
This is a cool PoC! We should also keep thinking about a more integrated solution in Documenter (e.g. as a plugin), to e.g. to remove the box from the docstring and such. Can you think of any other information that you would like to include on the reference page that we could generate automatically?

One thing I wonder if this would also require some solution to https://github.com/JuliaDocs/Documenter.jl/issues/1079, so that you could have both inline and API docstrings.

odow:
The "backlinks" thing in Pollen is pretty neat.

image

Let me learn how to write a Documenter plug-in... The overall design of Pollen is nice and something to replicate:

image

It'd be nice to have different method clearly separated with links to the source visible. Rather than as button on hover.

.github/workflows/documentation.yml
```diff
@@ -22,7 +22,12 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
           DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # For authentication with SSH deploy key
+          DOCUMENTER_LATEX_DEBUG: ${{ github.workspace }}/latex-debug-logs
         run: julia --color=yes --project=docs/ docs/make.jl
+      - uses: actions/upload-artifact@v3
```
odow, line 27:
```suggestion
      - uses: actions/upload-artifact@v3
        if: ${{ always() }}
```

odow:
Hmm. Seems like it's a table of contents thing?

The last run is:
```
------------
Run number 3 of rule 'pdflatex'
------------
------------
Running 'lualatex  -interaction=batchmode -halt-on-error -shell-escape -recorder  "JuMP.tex"'
------------
tput: No value for $TERM and no -T specified
Latexmk: Missing input file: 'JuMP.toc' from line
  'No file JuMP.toc.'
Latexmk: List of undefined refs and citations:
  Label `10320957702592950026' multiply defined
  Label `10452768864725067872' multiply defined
  Label `11049904912934130928' multiply defined
  Label `1190479243439829324' multiply defined
  Label `12078371379441951361' multiply defined
  Label `12204569949998619624' multiply defined
  Label `13119822296912292823' multiply defined
  Label `13626937967155666566' multiply defined
  Label `14602231903852437277' multiply defined
  Label `14868284870739381643' multiply defined
  Label `15167481340731409537' multiply defined
  Label `15274036632127375564' multiply defined
  Label `16402249899589529833' multiply defined
  Label `16879307210612674453' multiply defined
  Label `16881230196786944906' multiply defined
  Label `17764947666624653333' multiply defined
  Label `191277205180970345' multiply defined
  Label `1924664524817847375' multiply defined
  Label `207608259042424899' multiply defined
  Label `219752489661273253' multiply defined
  Label `2504400552717476215' multiply defined
  Label `304762286158531042' multiply defined
  Label `3807517405601295309' multiply defined
  Label `4911071156002605034' multiply defined
  Label `5236547219531420915' multiply defined
  Label `5292908469261308310' multiply defined
  Label `6467694110721589801' multiply defined
  Label `6515500382253974441' multiply defined
  Label `7395396977737699532' multiply defined
  Label `9048419797204458298' multiply defined
Latexmk: Summary of warnings:
  Latex found 79 multiply defined reference(s)
Failure to make 'JuMP.pdf'
Collected error summary (may duplicate other messages):
  pdflatex: Command for 'pdflatex' gave return code 256
```

docs/make.jl
```diff
@@ -186,6 +186,101 @@ sort!(_LIST_OF_EXTENSIONS; by = x -> (!startswith(x[1], "jump-dev/"), x[1]))
...
+                if type == "constant"
+                    @warn("Skipping $key")
+                    continue
+                else
+                    error("Documentation missing for $key")
+                end
+            end
+        end
+        open(joinpath(src_dir, sub_dir, "$key.md"), "w") do io
+            return write(io, "```@docs\n$key\n```")
```
odow, line 246:
```suggestion
            return write(io, "# `$key`\n\n```@docs\n$key\n```")
```

mortenpi:
It looks like we might be generating the same `\label` for different headings if they contain the same string?

https://github.com/JuliaDocs/Documenter.jl/blob/233a7822d670b063e894f62aea00729bcb3a28dd/src/latex/LaTeXWriter.jl#L295-L301

odow:
I bisection-searched the `.tex` to see if it'd compile. The problem iiiiiiiiid

```
% Useful variables
\newcommand{\DocMainTitle}{JuMP}
\newcommand{\DocVersion}{}
\newcommand{\DocAuthors}{The JuMP core developers and contributors}
\newcommand{\JuliaVersion}{1.6.7}

% ---- Insert preamble
\input{preamble.tex}

\chapter{`@build_constraint`}

\end{document}
```

Should be ```\chapter{`@build\_constraint`}```

odow:
Probable fix: https://github.com/JuliaDocs/Documenter.jl/pull/2134

docs/make.jl
```diff
@@ -1,5 +1,5 @@
 import Pkg
-Pkg.pkg"add Documenter#633a95a"
+Pkg.pkg"add Documenter#d264a780e0f2d3b39997aed91dccba75da50bf1d"
```
odow, line 2:
```suggestion
Pkg.pkg"add Documenter#2c30b477ad27602464f66744f6b3cea95840ba78"
```

odow:
The multiple labels are because Documenter just hashes the string of the label. So if I have two headers like `# Foo` and `## Foo`, they'll get the same label. It's only a problem when I got to actually `@ref` one of them and find out that they're duplicate. And that can't actually happen or the HTML will complain.

docs/make.jl
```diff
@@ -514,13 +606,19 @@ if _PDF
     pop!(_PAGES)        # remove /Extensions
     pop!(_PAGES)        # remove /Solvers
     push!(_PAGES, moi)  # Re-add /MathOptInterface
+    section_title, contents = _PAGES[4]
+    @assert section_title == "API Reference"
+    # `contents` is a big list of docstrings. By default, they'll
+    # show up at the `\chapter` level. That's too high.
+    _PAGES[4] = section_title => ["Docstrings" => [contents[2:end]]]
```
odow, line 613:
```suggestion
    _PAGES[4] = section_title => ["Docstrings" => ["JuMP" => contents[2:end]]]
```

odow:
@mortenpi are "plugins" actually a thing? It isn't obvious to me how to make a new ` ```@foo` block without editing the Documenter source code.

odow:
Even as-is, I'm in favor of merging. It's an improvement over what we currently have. I particularly like the ability to link to a function: https://jump.dev/JuMP.jl/previews/PR3386/api/set_start_values/. And since we're backwards compatible, these links aren't going to rot.

joaquimg:
The single page per function is mandatory for the new style?
I prefer a large page in which I can do a ctrl+F easily.

odow:
There's https://jump.dev/JuMP.jl/previews/PR3386/api/JuMP/

But if you want to search within the docstrings... then I guess you need to use the search function.

joaquimg:
The issue with the search function is that it is a 2-click thing.
You search, get a one liner, click again to see if its really that.
If that is what you want great, otherwise you have a bunch of clicks again.

With ctrl+F you can loop through everything (functions and their description).

So the search function is ok if you know what function you are looking for, you just dont know where it is in the manual.

But if you are looking for something you dont know exactly what it is, I think it is bad... I do a lot of ctrl+F in this case.

mortenpi:
> @mortenpi are "plugins" actually a thing? It isn't obvious to me how to make a new `` ```@foo `` block without editing the Documenter source code.

Eee... so about that :sweat_smile: The short answer is sort of, but not officially? I want proper plugin support, but right now you just have to add the appropriate method overloads, basically. In general, I want to encourage more experimentation with plugins, and then once we have some use cases, we can formulate proper, official APIs in Documenter for them (that adhere to semver and have strict interfaces and all that).

It is doable as a separate package though. Here's a simple package that matches the `mermaid` code blocks and treats them appropriately (including an overload of `Documenter.HTMLWriter.domify` to actually render HTML): https://github.com/JuliaDocs/DocumenterMermaid.jl/blob/main/src/DocumenterMermaid.jl

The API reference, I imagine, would hook into the build process somehow differently though. I imagine it would actually push new pages dynamically into the `Document` object.

odow:
@joaquimg what sorts of things are you CRTL+F'ing for? Currently the docstrings are over a few separate pages, so it's not like you get to search everything. And a single page with all of the docstrings seems excessive and a bit overwhelming for the average user.

docs/DocumenterReference.jl
```diff
@@ -0,0 +1,180 @@
...
+    subdir = config.subdirectory
+    overview_md = """
+    # [API](@id DocumenterReference_$(config.current_module))
+
+    | NAME | KIND |
+    | :--- | :--- |
+    """
+    _iterate_over_symbols(config) do key, type
+        _add_page(document, "$subdir/$key.md", "```@docs\n$key\n```\n")
+        overview_md *= "| [`$key`](@ref) | `$(_to_string(type))` |"
```
odow, line 163:
```suggestion
        overview_md *= "| [`$key`](@ref) | `$(_to_string(type))` |\n"
```

docs/DocumenterReference.jl
```diff
@@ -0,0 +1,182 @@
...
+
+function _add_page(document::Documenter.Document, filename, contents)
+    mdpage = Markdown.parse(contents)
+    document.blueprint.pages[filename] = Documenter.Page(
+        filename, # source,
+        "$(document.user.build)/$filename", # build,
+        "$(document.user.build)/", # workdir,
+        mdpage.content,
+        Documenter.Globals(),
+        convert(MarkdownAST.Node, mdpage),
```
odow, line 149: cc @mortenpi very tacky. Did I do things correctly?

docs/DocumenterReference.jl
```diff
@@ -0,0 +1,182 @@
...
+    )
+    return
+end
+
+function _build_api_page(document::Documenter.Document, config::_Config)
+    subdir = config.subdirectory
+    overview_md = """
+    # [API](@id DocumenterReference_$(config.current_module))
```
mortenpi, line 157: You might be getting nonsense edit URL for the pages right now? But you should be able to add a

````markdown
```@meta
EditURL = nothing
```
````

to the top here (or bottom, or wherever, really).

File: docs/DocumenterReference.jl (hunk 11 above)
Comment (mortenpi) [reply docs/DocumenterReference.jl:149]: Oh, `order = 0.0`, interesting choice, I didn't even consider this. But this should definitely also work. I was imagining adding explicitly `Documenter.Page` objects later.. but this is probably easier. It can always be changed around in the future if need be, so I think this is good.

Comment (odow) [reply docs/DocumenterReference.jl:149]: > interesting choice, I didn't even consider this

Haha. It wasn't obvious how I could add a page at the `Expander` section, and then have it re-run the expander on the contents of the new page. But perhaps I missed something obvious.

File: docs/DocumenterReference.jl (hunk 12 above)
Comment (odow) [reply docs/DocumenterReference.jl:157]: Ahhhhh. I knew I was missing something.

File: docs/DocumenterReference.jl (hunk 11 above)
Comment (mortenpi) [reply docs/DocumenterReference.jl:149]: So I would have probably tried to hack it in as a `Builder` phase at `1.1 < order < 2.0`. But I did just noticed that the `SetupBuildDirectory` phase also takes care of some of the navigation stuff (the linear order of the pages etc). So it's actually convenient to do it before `SetupBuildDirectory` as you did, and let `SetupBuildDirectory` handle that for you.

File: docs/DocumenterReference.jl [hunk 13]
```diff
@@ -0,0 +1,198 @@
...
+end
+
+function _iterate_over_symbols(f, config)
+    current_module = config.current_module
+    for (key, type) in vcat(_exported_symbols(current_module), config.extras)
+        if key isa Symbol
+            doc = Base.Docs.doc(Base.Docs.Binding(current_module, key))
+            if occursin("No documentation found.", string(doc))
+                if type == DOCTYPE_CONSTANT
+                    continue  # It's okay not to document every constant.
```
odow, line 116: I'd like to remove this. Everything in the public API should be documented. But it requires https://github.com/jump-dev/MathOptInterface.jl/pull/2186 first.

odow:
So this needs some more eyes on. @mlubin @blegat @jd-foster 

Pros: a very explicit list of our public API: https://jump.dev/JuMP.jl/previews/PR3386/api/JuMP/

Cons: @joaquimg cannot CTRL+F though a giant page of the docstrings.

Thoughts?

docs/make.jl
```diff
@@ -186,6 +193,53 @@ sort!(_LIST_OF_EXTENSIONS; by = x -> (!startswith(x[1], "jump-dev/"), x[1]))
...
+            "Containers.default_container" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.nested" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.vectorized_product" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.build_ref_sets" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.container_code" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.AutoContainerType" =>
+                DocumenterReference.DOCTYPE_STRUCT,
+            "Containers.NestedIterator" => DocumenterReference.DOCTYPE_STRUCT,
```
odow, line 236:
```suggestion
            "Containers.NestedIterator" =>
                DocumenterReference.DOCTYPE_STRUCT,
```

docs/make.jl
```diff
@@ -186,6 +193,53 @@ sort!(_LIST_OF_EXTENSIONS; by = x -> (!startswith(x[1], "jump-dev/"), x[1]))
...
+            "Containers.@container" => DocumenterReference.DOCTYPE_MACRO,
+            "Containers.container" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.rowtable" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.default_container" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.nested" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.vectorized_product" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.build_ref_sets" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.container_code" => DocumenterReference.DOCTYPE_FUNCTION,
```
odow, line 233:
```suggestion
            "Containers.container_code" =>
                DocumenterReference.DOCTYPE_FUNCTION,
```

docs/make.jl
```diff
@@ -186,6 +193,53 @@ sort!(_LIST_OF_EXTENSIONS; by = x -> (!startswith(x[1], "jump-dev/"), x[1]))
...
+            # TODO(odow): consider exporting these from JuMP.Containers.
+            "Containers.@container" => DocumenterReference.DOCTYPE_MACRO,
+            "Containers.container" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.rowtable" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.default_container" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.nested" => DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.vectorized_product" =>
+                DocumenterReference.DOCTYPE_FUNCTION,
+            "Containers.build_ref_sets" => DocumenterReference.DOCTYPE_FUNCTION,
```
odow, line 232:
```suggestion
            "Containers.build_ref_sets" =>
                DocumenterReference.DOCTYPE_FUNCTION,
```

jd-foster:
I think it has the advantage of encouraging more documentation on the docstrings when they have a whole page to fill.

> @joaquimg what sorts of things are you CRTL+F'ing for? Currently the docstrings are over a few separate pages, so it's not like you get to search everything. And a single page with all of the docstrings seems excessive and a bit overwhelming for the average user.

Agree, but for the more advanced user it can be pretty useful; for example, Plotly does a single-page reference: https://plotly.com/julia/reference/

odow:
The latest commit should build a single page reference. Let's see how it looks.

odow:
How's this? https://jump.dev/JuMP.jl/previews/PR3386/api/JuMP/

joaquimg:
I like the single page api reference.
I would be fine with a handful of pages with some clear separation if there is a justification.
I really do not like the single page per item because of the ctrl+f.

odow:
Table or no table? (If no table, there'd just be the navigation sidebar with a list of functions)

jd-foster:
+1 for _not_ having the table (the Overview: NAME/KIND table, I assume) at the top of the page, and using the side bar to perform its function.

odow:
Ha. If we're going to have a one-pager without the table (let's see how it looks) then I can greatly simplify the Documenter side of things.

joaquimg:
I dont find the table helpful.

odow:
The single page is might be better for experts. But I see this all the time:

https://discourse.julialang.org/t/are-revise-packagecompiler-and-pkg-part-of-julia-and-how-do-we-better-share-them/99491/39?u=odow

image

joaquimg:
I find it hard to see why having a single page might be bad.
I think the user is referring to something else:
They want *very* detailed docs with multiple examples of usage for every function. And also many links to related functions.
Also, good classification makes sense. But that could be solved with some intermediary pages that list specific functions in some groups with non-empty intersection.

mlubin:
What's this infamous table? It's a bit hard to follow the discussion because the preview only has the latest version. The single page seems reasonable to me, given that the docstrings aren't as elaborate as MATLAB's function documentation.

odow:
> What's this infamous table?

It was a table at the top of the page with one row for each symbol, and a column for the name and type (function, macro, struct, constant) of the symbol. Similar to the one here: https://fluxml.ai/FastAI.jl/dev/FastAI@dev/ref/FastAI.html

It's necessary if you have a separate page for each docstring. But a little overboard if you have a giant page. 

Although I'm still in favor of it because it makes discovery a little easier. (Some people might miss the navigation sidebar.)

odow:
Table looked like this: https://jump.dev/MathOptInterface.jl/previews/PR2187/api/MathOptInterface/

mlubin:
The FastAI table takes up limited vertical space and has a scrollbar. That seems ok. Without a scrollbar it just takes up too much space without being that useful. That said, I'm not convinced we need a table at all.

odow:
So, ignoring the table, suggestions for changes to how things are currently? 

https://jump.dev/JuMP.jl/previews/PR3386/api/JuMP/

mlubin:
Maybe a sentence at the top directing readers to the manual for a more structured overview of the JuMP API? We don't want to suggest that people should read this page end to end before reading the manual.

.github/workflows/documentation.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
           DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # For authentication with SSH deploy key
           DOCUMENTER_LATEX_DEBUG: ${{ github.workspace }}/latex-debug-logs
+          TERM: dumb
```
odow, line 25:
```suggestion
```

odow:
Any other comments?

blegat (approved):
I like it. It's now easier to find docstring (just go to the single page and Ctrl+F) and easy for us to maintain (we don't have to do anyting).
About MATLAB, they have very long docstring with a lot of examples but we have tutorials for that. We could try to create a list of all tutorials using a function and add it to the docstring but that would be a lot for `optimize!` and people can use Documenter.jl's search function for that so I think it's good like that

odow:
I'll merge this tomorrow then, assuming no the comments. We can always make more changes once it's merged.

PR #3392, opened 2023-05-25
[docs] Color the Julia logs on CI

mortenpi:
I noticed that they're pretty gray: https://github.com/jump-dev/JuMP.jl/actions/runs/5085355976/jobs/9138739849?pr=3386


odow:
This would make it easier to see the doctest failures... 😆

odow (approved):

PR #3394, opened 2023-05-28
Add coefficient_type argument to add_bridge and remove_bridge

### LINKED ISSUE #3393: Bridges parametrized by complex type
In `add_bridge`, JuMP adds the `Float64` version of the bridge:
https://github.com/jump-dev/JuMP.jl/blob/e7f1bd1f311efb867f4737efab188ee72bed52cc/src/JuMP.jl#L497
This is an issue with PolyJuMP for instance which needs some bridge to act with complex coefficients as well:
https://github.com/jump-dev/SumOfSquares.jl/pull/300
One solution would be to have `add_typed_bridge` that adds the bridge already parametrized so that JuMP does not do `{Float64}`. Another option would be to have `add_complex_bridge`. It can also be keyword arguments to `add_bridge`.
Then we would need similar changes to `BridgeableConstraint`.

odow:
Closes #3393

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -449,7 +449,7 @@ abstract type AbstractConstraint end
 """
     BridgeableConstraint(
         constraint::C,
-        bridge_type::B;
+        bridge_type::Type{B};
         coefficient_type::Type{T} = Float64,
     ) where {C<:AbstractConstraint,B<:MOI.Bridges.AbstractBridge,T}
 
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -447,23 +447,25 @@ to be attached to a model.
 abstract type AbstractConstraint end
 
 """
-    struct BridgeableConstraint{C, B} <: AbstractConstraint
-        constraint::C
-        bridge_type::B
-    end
+    BridgeableConstraint(
+        constraint::C,
+        bridge_type::B;
+        value_type::Type{T} = Float64,
+    ) where {C<:AbstractConstraint,B<:MOI.Bridges.AbstractBridge,T}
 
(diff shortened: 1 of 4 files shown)


blegat:
`value_type` might be confusing with the `value_type` we're adding in https://github.com/jump-dev/JuMP.jl/pull/3377
In here, it's really the `coefficient_type` used with `MOI.Utilities` with functions

odow:
Updated to `coefficient_type`.

src/constraints.jl
```diff
@@ -447,33 +447,37 @@ to be attached to a model.
...
 """
-    struct BridgeableConstraint{C, B} <: AbstractConstraint
-        constraint::C
-        bridge_type::B
-    end
+    BridgeableConstraint(
+        constraint::C,
+        bridge_type::B;
+        coefficient_type::Type{T} = Float64,
+    ) where {C<:AbstractConstraint,B<:MOI.Bridges.AbstractBridge,T}
```
blegat, line 454: `B<:MOI.Bridges.AbstractBridge`  is not true, since `B` is a `Type`

src/constraints.jl
```diff
@@ -447,33 +447,37 @@ to be attached to a model.
 abstract type AbstractConstraint end
 
 """
-    struct BridgeableConstraint{C, B} <: AbstractConstraint
-        constraint::C
-        bridge_type::B
-    end
+    BridgeableConstraint(
+        constraint::C,
+        bridge_type::B;
```
odow, line 452:
```suggestion
        bridge_type::Type{B};
```

blegat (approved):

PR #3382, opened 2023-05-24
[WIP] [docs] Complex number AC Optimal Power Flow tutorial

jd-foster:
Still plenty to add and fix here, but the first pass at #3363 .

It currently is a nonlinear AC-OPF formulation, but the idea is to add to it the original intent, SDP relaxation of AC-OPF using `HermitianPSDCone`.

PR #3395, opened 2023-05-30
[docs] Complex number AC Optimal Power Flow tutorial

### LINKED ISSUE #3363: [docs] Tutorial on SDP relaxation of AC-OPF using `HermitianPSDCone`
This got somewhat lost: https://github.com/jump-dev/JuMP.jl/pull/3109#issuecomment-1287043601
* Happy for this issue to just be a new checkbox here:
https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-1341644412

odow:
Rebase of https://github.com/jump-dev/JuMP.jl/pull/3382

@jd-foster you should be able to pull and push from this branch now.

Preview: https://jump.dev/JuMP.jl/previews/PR3395/tutorials/applications/optimal_power_flow/

Closes #3363


docs/src/tutorials/conic/optimal_power_flow.jl
```diff
@@ -0,0 +1,504 @@
...
+
+@constraint(model, sum(P_G) >= sum(P_Demand))
+
+optimize!(model)
+first_relaxation_lower_bound = round(objective_value(model); digits = 2)
+println("Objective value (W relax. lower bound): $first_relaxation_lower_bound")
+
+Test.@test termination_status(model) == OPTIMAL    #src
+Test.@test primal_status(model) == FEASIBLE_POINT  #src
+Test.@test isapprox(first_relaxation_lower_bound, 2733.35; atol = 1e-2)  #src
```
odow, line 412:
```suggestion
Test.@test isapprox(first_relaxation_lower_bound, 2733.35; rtol = 1e-2)  #src
```

docs/src/tutorials/conic/optimal_power_flow.jl
```diff
@@ -0,0 +1,546 @@
...
+# ``P^D`` and ``Q^D`` respectively (these are fixed at 0 in the case of
+# trans-shipment and generator nodes).
+
+# The cost of operating each generator is modeled as a quadratic function of
+# its real power output; in our specific test case, the objective function to
+# minimize is:
+# ```math
+# \begin{align}
+#     \min \;\; & \; 0.11 \;\; (P^G_1)^2 +   5 P^G_1 + 150  \\
+#        + \;\; & 0.085 \; (P^G_2)^2 + 1.2 P^G_2 + 600  \\
```
odow, line 95:
```suggestion
#               & + 0.085 \; (P^G_2)^2 + 1.2 P^G_2 + 600  \\
```

docs/src/tutorials/conic/optimal_power_flow.jl
```diff
@@ -0,0 +1,546 @@
...
+#          \;\; & 0.1225 \;  (P^G_3)^2 +     P^G_3 + 335 \\
```
odow, line 96:
```suggestion
#               & + 0.1225 \;  (P^G_3)^2 +     P^G_3 + 335 \\
```

docs/src/tutorials/conic/optimal_power_flow.jl
```diff
@@ -0,0 +1,546 @@
...
+@variable(model, V[1:N] in ComplexPlane(), start = 1.0 + 0.0im)
+@constraint(model, imag(V[1]) == 0)
+
+@constraint(model, LinearAlgebra.Hermitian([1 V'; V W]) in HermitianPSDCone())
+
+## 2 x 2 minor inequalities:
+@constraint(
+    model,
+    [i = 1:N],
+    [1 // 2; real(W[i, i]); [real(V[i]); imag(V[i])]] in
```
odow, line 464:
```suggestion
    [0.5, real(W[i, i]), real(V[i]), imag(V[i])] in
```

docs/src/tutorials/conic/optimal_power_flow.jl
```diff
@@ -0,0 +1,546 @@
...
+#     \min \;\; & \; 0.11 \;\; (P^G_1)^2 +   5 P^G_1 + 150  \\
+#        + \;\; & 0.085 \; (P^G_2)^2 + 1.2 P^G_2 + 600  \\
+#          \;\; & 0.1225 \;  (P^G_3)^2 +     P^G_3 + 335 \\
+# \end{align}
+# ```
+# Let's create an initial JuMP model with some of this data:
+
+model = Model(Ipopt.Optimizer)
+set_silent(model)
+@variable(model, P_Gen_lb[i] <= P_G[i = 1:N] <= P_Gen_ub[i])
```
odow, line 103:
```suggestion
@variable(model, P_Gen_lb[i] <= P_G[i in 1:N] <= P_Gen_ub[i])
```
I don't know how well I've enforced this in the docs, and JuliaFormatter typically leaves it along because it's in a macro, but the JuMP style guide would prefer ` in ` over ` = `.

odow (commented):
This should definitely be moved to the `/applications` directory. I can see it being useful for EE people, but it might be too detailed for a general conic example to expect people to read.

jd-foster:
Ok, moved to the `/applications` directory and removing the WIP label.

docs/Project.toml
```diff
@@ -1,5 +1,6 @@
 [deps]
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+Clarabel = "61c947e1-3e6d-4ee4-985a-eec8c727bd6e"
```
odow, line 3: Needs a compat

docs/src/tutorials/applications/optimal_power_flow.jl
```diff
@@ -0,0 +1,545 @@
...
+# to obtaining a good estimate of the objective value at the global optimum
+# through the use of semidefinite programming.
+
+# The purpose of this tutorial is to highlight JuMP's ability to directly
+# formulate problems involving complex-valued decision variables and complex
+# matrix cones such as the  [`HermitianPSDCone`](@ref) object.
+
+# For another example of modeling with complex decision variables, see the
+# [Quantum state discrimination](@ref) tutorial, and see the
+# [Complex number support](@ref) section of the manual for more details.
```
odow, line 24: You could also link to PowerModels here to make it clear that people don't have to explicitly model using JuMP

docs/src/tutorials/applications/optimal_power_flow.jl
```diff
@@ -0,0 +1,545 @@
...
+# following question: what is the most cost-effective operation of electricity
+# generators while meeting constraints on the safe limits of network components?
+
+# The graph of the network shown here has nine nodes, with three each used for
+# the different purposes of generation ``G`` (nodes 1, 2, and 3), trans-shipment
+# (nodes 4, 6, and 8), and demand ``D`` (nodes 5, 7, and 9).
+
+# ![Nine Nodes](../../assets/case9mod.png)
+
+# We're using the 9-_bus_ network test case `case9mod` to explore this problem.
```
odow, line 50:
```suggestion
```

Paragraph before mentions nine nodes and next sentence says `case9`.

docs/src/tutorials/applications/optimal_power_flow.jl
```diff
@@ -0,0 +1,545 @@
...
+
+y_sh = 1 / 2 * (im * df_br.BR_Bc) * baseMVA
+
+# and leads to the shunt admittance matrix
+
+Y_sh = SparseArrays.spdiagm(
+    LinearAlgebra.diag(A * SparseArrays.spdiagm(y_sh) * A'),
+)
+
+# (This second one looks more complicated because we only want to add the
```
odow, line 208: Not obvious what "This second one" refers to

docs/src/tutorials/applications/optimal_power_flow.jl
jd-foster: Sorry, not sure what you're asking for.

jd-foster:
@frederikgeth @ccoffrin  I would appreciate any comments or improvement suggestions if you get a chance. 

Preview: https://jump.dev/JuMP.jl/previews/PR3395/tutorials/applications/optimal_power_flow/

docs/src/tutorials/applications/optimal_power_flow.jl
odow: I just thought the sentence was superfluous.

ccoffrin:
I didn't have a chance to look at the original version of this but I am a little surprised to see that line flow limits are not included (or did I miss them?).  I also generally find the matrix-centric version of formulation difficult to follow.  I prefer explicitly writing the line flow equation like `conj(Y+Y_fr)/(T*conj(T))*V_fr*conj(V_fr) - conj(Y)/T*V_fr*conj(V_to)` but the matrix form is worth it to make the SDP constraint easer to setup.

I saw a place or two where the use of `abs2` might further simplify the presentation.

FWIW, here is the formulation I came up with based on my preferred formulation style, https://github.com/lanl-ansi/rosetta-opf/blob/main/variants/jump-rectangular-complex.jl

Looking at the JuMP docs, I think my example could be improved further by defining the variables in the complex plain directly.

jd-foster:
@ccoffrin Thanks Carleton, that's useful feedback. 

> original version of this

This has been written from scratch; was there another tutorial I missed? Sorry if I'm not understanding you.

frederikgeth:
Thanks for this great contribution to the community. I have a few comments, to help reach the maximum audience. 

- I prefer the terminology _reference (angle) bus_ for what you're refering to as slack bus. You need the define one bus with a fixed voltage angle somewhere in the system, otherwise you have a degenerate problem. The concept of _slack_ bus originates from having to balance out active and reactive power in power flow (not OPF). By setting `imag(V[1]) == 0` you define the voltage angle at one bus to be 0 degrees, although in principle it could be any angle (offset) value. Note that you should in principle also restrict the sign of the real part `real(V[1]) >= 0`, as you otherwise also allow for a  second solution with a 180 degree angle, and from experience in unbalanced OPF I know this can cause unexpected numerical issues.
- You could mention in the beginning what the general plan of action is before getting into the details, i.e. develop 1) a rectangular formulation in power-voltage variables 2) based on the nodal admittance matrix of the system. In my experience, the approach where you don't build a nodal admittance matrix build construct the system of equations in a scalar fashion (like in PowerModels) is easier from the educational perspective, so it could be worth mentioning that these are two distinct approaches. 
- For the sparse SDP form, you could refer the readers to  
@ARTICLE{6064917,
  author={R. A. Jabr},
  title={Exploiting Sparsity in SDP Relaxations of the OPF Problem},
  journal={IEEE Transactions on Power Systems},
  volume={27},
  number={2},
  pages={1138-1139},
  year={2012},
  month={May},
  doi={10.1109/TPWRS.2011.2170772},
  ISSN={0885-8950}
}
- The W+V SDP relaxation section I find a bit unintuitive. You may want to mention why you're note defining W as PSD anymore at time of construction, and why this then requires you to set `imag(W[i, i]) == 0` and ` real(W[i, i]) >= 0`. Also, discuss why `sum(P_G) >= sum(P_Demand)` appears out of nowhere? You could also straight-away define `@constraint(model, [i in 1:N], 0.9^2 <= real(W[i, i]) <= 1.1^2)` where you define `@constraint(model, [i = 1:N], real(W[i, i]) >= 0)`? I'm not even sure why the gap would be worse? Similarly, the convergence shouldn't depend on start values so that's strange to see them? The significant difference in values or the voltage magnitude also goes unexplained. Similarly, you do the posprocessing of the voltage magnitude still based on the W variables instead of the V variables, which sort of disagrees with your statement of 'has the advantage that we can work directly with complex voltages to extend the formulation'. 
- You could mention that if the solution has a rank-1 W matrix, that the result would be exact, and that you could then uniquely post-process the angles?

jd-foster:
@frederikgeth Thanks Fred. I will definitely incorporate your first three points directly.
Your fourth point is insightful: I was already thinking of just deleting that last W+V section entirely for all those reasons as it wasn't adding much as a result.  Unless I can use it to add something distinctive, I still might.

frederikgeth:
I agree, at the moment the SDP W+V section doesn't really add much to the story or goals of this tutorial. It may be worth fleshing out at a later stage. 

If you prefer to keep SDP W+V, I would consider putting it before SDP W. It is easier to explain the lifting and reformulation if you keep both V+W initially. And then you can tell the story of how you can get rid of the V variables and still recover the same solution (for the SDP W section).

jd-foster:
Good, that story makes a lot more sense to me.
The alternative I was considering was to _make use_ of the voltage variables with observation that the voltage angles are "small" (between -5 and 15 degrees) in the global nonlinear solution, and use that information to bound them away from the origin (above the chord that joins rays at those angles and the lower voltage magnitude "circle"), recovering the voltage constraints that bind at the global optimum. Of course that's a whole other layer of complexity.

docs/src/tutorials/applications/optimal_power_flow.jl
jd-foster: Ok, will remove the redundancy.

odow:
LGTM. I don't really have an opinion on the SDP formulations. We can always merge and wait to see if we get any feedback from students etc.

jd-foster:
Great, I'm excited to get this merged!

jd-foster:
(I couldn't help doing it myself this time 😂)

odow:
> I couldn't help doing it myself this time

Haha. I was going to wait to see if anyone had other comments. But no harm no foul.

jd-foster:
Thanks! 
If anyone has any comments, please do post them.

ccoffrin:
I'm hoping to do another pass.

ccoffrin:
Old link to rendered preview version is broken, where can I view the latest version?

jd-foster:
https://jump.dev/JuMP.jl/dev/tutorials/applications/optimal_power_flow/

PR #3398, opened 2023-06-02
[docs] build on Julia v1.9 to support Pkg extensions

odow:
A necessary precursor for https://github.com/jump-dev/JuMP.jl/pull/3383. Haven't run locally, so let's see if any changes need to be made.


odow:
Blocked by the need to first update the MOI docs to Julia 1.9...

odow:
@jd-foster the SDP example is pretty flakey. Any suggestions for how to make it more robust?

jd-foster:
Yeah, it keeps killing the CI. I see you increased the tolerances, that's one option. Maybe we need try SCS again (usually these types of problems get solved with Mosek)?

PR #3399, opened 2023-06-02
[docs] Add Lovász number section to simple SDP examples.

jd-foster:
A classic SDP example:
https://en.wikipedia.org/wiki/Lov%C3%A1sz_number#Lov%C3%A1sz_%22sandwich_theorem%22

### DIFF
--- a/docs/src/tutorials/conic/simple_examples.jl
+++ b/docs/src/tutorials/conic/simple_examples.jl
@@ -359,7 +359,6 @@ function example_theta_problem()
     model = Model(SCS.Optimizer)
     set_silent(model)
     E = [(1, 2), (2, 3), (3, 4), (4, 5), (5, 1)]
-
     @variable(model, X[1:5, 1:5], PSD)
     for i in 1:5
         for j in (i+1):5
--- a/docs/src/tutorials/conic/simple_examples.jl
+++ b/docs/src/tutorials/conic/simple_examples.jl
@@ -373,7 +373,6 @@ function example_theta_problem()
         end
     end
     @constraint(model, LinearAlgebra.tr(LinearAlgebra.I * X) == 1)
-
     J = ones(Int, 5, 5)
     @objective(model, Max, LinearAlgebra.dot(J, X))
     optimize!(model)
--- a/docs/src/tutorials/conic/simple_examples.jl
+++ b/docs/src/tutorials/conic/simple_examples.jl
@@ -372,7 +372,7 @@ function example_theta_problem()
             end
         end
     end
-    @constraint(model, LinearAlgebra.dot(LinearAlgebra.I, X) == 1)
+    @constraint(model, LinearAlgebra.tr(LinearAlgebra.I * X) == 1)
 
     J = ones(Int, 5, 5)
     @objective(model, Max, LinearAlgebra.dot(J, X))
--- a/docs/src/tutorials/conic/simple_examples.jl
+++ b/docs/src/tutorials/conic/simple_examples.jl
@@ -306,6 +306,86 @@ end
 
 example_minimum_distortion()
 
+# ## Lovász numbers
+
+# The [Lovász number](https://en.wikipedia.org/wiki/Lov%C3%A1sz_number)
+# of a graph, also known as Lovász's theta-function, is a number that
+# lies between two important and related numbers that are computationally
+# hard to determine, namely the chromatic and clique numbers of the graph.
+# It is possible however to efficient compute the Lovász number as the
+# optimal value of a semidefinite program.
+
+# Consider the pentagon graph:
+# ```raw
+#      [5]
+#     /   \    
+#    /     \     
+#  [1]     [4]
+#   |       |
+#   |       |  
+#  [2] --- [3]
+# ```
+# with five vertices and edges. Its Lovász number is known to be precisely
+# ``\sqrt{5} \approx 2.236``, lying between 2 (the largest clique size) and
+# 3 (the smallest number needed for a vertex coloring).
+
+# Let ``i, \, j`` be integers such that ``1 \leq i < j \leq 5``.
+# We define ``A^{ij}`` to be the ``5 \times 5`` symmetric
+# matrix with entries ``(i,j)`` and ``(j,i)`` equal to 1, with all other entries 0.
+# Let  ``E`` be the graph's edge set; in this example, ``E`` contains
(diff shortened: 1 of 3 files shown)


blegat:
The problem is in geometric conic form so `Dualization.dual_optimizer(SCS.Optimizer)` should be used instead.

odow:
Preview: https://jump.dev/JuMP.jl/previews/PR3399/tutorials/conic/simple_examples/#Lovász-numbers

> The problem is in geometric conic form so Dualization.dual_optimizer(SCS.Optimizer) should be used instead.

I think adding `Dualization` to the JuMP documentation is a separate PR that probably deserves its own tutorial.

docs/src/tutorials/conic/simple_examples.jl
```diff
@@ -306,6 +306,86 @@ end
...
+# Let ``i, \, j`` be integers such that ``1 \leq i < j \leq 5``.
+# We define ``A^{ij}`` to be the ``5 \times 5`` symmetric
+# matrix with entries ``(i,j)`` and ``(j,i)`` equal to 1, with all other entries 0.
+# Let  ``E`` be the graph's edge set; in this example, ``E`` contains
+# (1,2), (2,3), (3,4), (4,5), (5,1)
+# and their transposes. The Lovász number can be computed from the program 
+# ```math
+# \begin{align}
+#     \text{max}      & \quad    J • X & \\
+#     \text{    s.t.} & \quad A^{ij} • X = 0 \text{ for all } (i,j) \notin E   \\
```
odow, line 341:
```suggestion
#     \text{    s.t.} & \quad A^{ij} • X = 0 \quad \forall (i,j) \notin E   \\
```

docs/src/tutorials/conic/simple_examples.jl
```diff
@@ -306,6 +306,86 @@ end
...
+            if !((i, j) in E || (j, i) in E)
+                A = zeros(Int, 5, 5)
+                A[i, j] = 1
+                A[j, i] = 1
+                @constraint(model, LinearAlgebra.dot(A, X) == 0)
+            end
+        end
+    end
+    @constraint(model, LinearAlgebra.tr(LinearAlgebra.I * X) == 1)
+
```
odow, line 376:
```suggestion
```

docs/src/tutorials/conic/simple_examples.jl
```diff
@@ -306,6 +306,86 @@ end
...
+# [_The sandwich theorem_](https://doi.org/10.37236%2F1193), 
+# Electronic Journal of Combinatorics, Volume 1, Issue 1, A1.
+
+function example_theta_problem()
+    model = Model(SCS.Optimizer)
+    set_silent(model)
+    E = [(1, 2), (2, 3), (3, 4), (4, 5), (5, 1)]
+
+    @variable(model, X[1:5, 1:5], PSD)
+
```
odow, line 364:
```suggestion
```

docs/src/tutorials/conic/simple_examples.jl
```diff
@@ -306,6 +306,86 @@ end
...
+#
+# Knuth, D. E. (1994), 
+# [_The sandwich theorem_](https://doi.org/10.37236%2F1193), 
+# Electronic Journal of Combinatorics, Volume 1, Issue 1, A1.
+
+function example_theta_problem()
+    model = Model(SCS.Optimizer)
+    set_silent(model)
+    E = [(1, 2), (2, 3), (3, 4), (4, 5), (5, 1)]
+
```
odow, line 362:
```suggestion
```

odow (commented):
LGTM. Style suggestions are optional. For these simple examples I tried to go with the fewest lines of code.

jd-foster:
It seems (ironically?) the OPF example is failing and killing the documentation build here...

blegat:
> I think adding `Dualization` to the JuMP documentation is a separate PR that probably deserves its own tutorial.

It does deserve it's own tutorial but we should also always use it in conic tutorials when appropriate. Especially when it's purely geometric conic form or geometric standard form.

joaquimg:
Yeah,
I agree we should strive to use what we would recommend in the tutorials.
We can either formulate in the way the solver wants the problem or use this tutorial to introduce dualization.
We certainly dont want users complaining things are slow because they followed the tutorial example.

odow:
> We certainly dont want users complaining things are slow because they followed the tutorial example.

It's a 5x5 PSD. They aren't going to complain it's slow. They might equally wonder why we need Dualization in this example. The current version is fine for the "simple examples" tutorial. Dualization can be a separate thing in which we actually explain the different conic formulations.

blegat:
> It's a 5x5 PSD. They aren't going to complain it's slow.

They are going to copy-paste this and apply it to a larger scale example.

> Dualization can be a separate thing in which we actually explain the different conic formulations.

If we only talk about Dualization in one tutorial it's not enough, we should use in every conic tutorials when it a clear win.

odow:
If we want to add dualization to various conic tutorials we can add it once it has a standalone tutorial. We don't need to do it now or in this PR.

joaquimg:
I would be fine not adding Dualization here, but, in this case, I think we should use the right formulation for the solver and highlight it.

jd-foster:
How do you determine the right formulation for a given solver? (I genuinely don't know.)

blegat:
> How do you determine the right formulation for a given solver? (I genuinely don't know.)

https://jump.dev/Dualization.jl/stable/manual/#DualOptimizer-1

> I would be fine not adding Dualization here, but, in this case, I think we should use the right formulation for the solver and highlight it.

Do you mean manually dualizing or manually adding bridging to free variables and vector affine conic constraints ?

> If we want to add dualization to various conic tutorials we can add it once it has a standalone tutorial. We don't need to do it now or in this PR.

I mentioned this example in the issue: https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-1575556301 to make sure we don't forget.

odow:
> How do you determine the right formulation for a given solver? (I genuinely don't know.)

:+1: This is my point on why we need a separate Dualization tutorial. The nuance is totally lost on 99% of JuMP users. Using Dualization in a few places is confusing. It's really an extra-for-experts option.

blegat:
> It's really an extra-for-experts option.

It shouldn't be an option for expert though. Everyone conic user should use it and we should try to educate users to do so

PR #3400, opened 2023-06-04
Update MPC paper citation

joaquimg:


blegat (approved):

blegat:
The failure on CI is the optimal power flow example: https://github.com/jump-dev/JuMP.jl/actions/runs/5166563401/jobs/9306814066?pr=3400#step:6:308

jd-foster:
I've relaxed the tolerances in the theta function PR due to the same, but probably should have made it separately

blegat:
> I've relaxed the tolerances in the theta function PR due to the same, but probably should have made it separately

It's fine, I understand the desire to save CI resources ^^

PR #3402, opened 2023-06-05
[docs] add Dualization tutorial

odow:
x-ref https://github.com/jump-dev/JuMP.jl/issues/2348

https://jump.dev/JuMP.jl/previews/PR3402/tutorials/conic/dualization/


docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
+# Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors    #src
+# This Source Code Form is subject to the terms of the Mozilla Public License   #src
+# v.2.0. If a copy of the MPL was not distributed with this file, You can       #src
+# obtain one at https://mozilla.org/MPL/2.0/.                                   #src
+
+# # Dualization
+
+# The purpose of this tutorial is to explain how to use [Dualization.j](@ref) to
```
odow, line 8:
```suggestion
# The purpose of this tutorial is to explain how to use [Dualization.jl](@ref) to
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+
+# This tutorial uses the following packges
+
+using JuMP
+import Dualization
+import SCS
+
+# ## Background
+
+# Conic optimization solvers typically accept one of two input formulations:
```
odow, line 26:
```suggestion
# Conic optimization solvers typically accept one of two input formulations.
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+# ## Primal and dual formulations
+
+# Duality plays a large role in conic optimization. For a detailed description
+# of conic duality, see [Duality](@ref).
+
+# A useful observation is that if the primal problem is in standard conic form,
+# then the dual problem is in geometric conic form, and vice versa. Morever, the
+# primal and dual may have a different number of variables and constraints,
+# although which one is smaller depends on the problem. Therefore, instead of
+# reformulating the problem problem from one form to the other, it can be more
```
odow, line 91:
```suggestion
# reformulating the problem from one form to the other, it can be more
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+
+# ## `dual_optimizer`
+
+# Manually deriving the conic dual is difficult and error-prone. The package
+# [Dualization.jl](@ref) provides the `Dualization.dual_optimizer` meta-solver,
+# which wraps any MathOptInterface-compatible solver in an interface that
+# automatically formulates the dual of an input problem, solves the dual
+# problem, and then reports the primal solution back to the user.
+
+# To demonstrate, we use `Dualization.dual_optimizer` to solve `model_primal`;
```
odow, line 179:
```suggestion
# To demonstrate, we use `Dualization.dual_optimizer` to solve `model_primal`:
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+#               \;\;\text{s.t.} & [A -I] [x; y] = b \\
+#                               & [x; y] \in \mathbb{R}^n \cross \mathcal{K}
+# \end{align}
+# ```
+# and to go from the standard conic form to the geometric conic form, we can
+# rewrite the equality constraint as a function belonging to the `{0}` cone:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
+#               \;\;\text{s.t.} & [A; I] x - [b; 0] \in {0}\cross \mathcal{K}
```
jd-foster, line 64:
```suggestion
#               \;\;\text{s.t.} & [A; I] x - [b; 0] \in \{0\} \cross \mathcal{K}
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+# variables are free.
+
+# It is trivial to convert between these two representations, for example, to go
+# from the geometric conic form to the standard conic form we introduce slack
+# variables $y$:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
+#               \;\;\text{s.t.} & [A -I] [x; y] = b \\
+#                               & [x; y] \in \mathbb{R}^n \cross \mathcal{K}
```
odow, line 56:
```suggestion
#                               & [x; y] \in \mathbb{R}^n \times \mathcal{K}
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+#                               & x \in \mathcal{K}
+# \end{align}
+# ```
+# in which we have a set of linear equality constraints $Ax = b$ and the
+# variables belong to a cone $\mathcal{K}$.
+#
+# The second is the _geometric_ conic form:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
```
jd-foster, line 42:
```suggestion
#     \min_{x \in \mathbb{R}^n} \; & c^\top x \\
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+import SCS
+
+# ## Background
+
+# Conic optimization solvers typically accept one of two input formulations.
+
+# The first is the _standard_ conic form:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
```
jd-foster, line 31:
```suggestion
#     \min_{x \in \mathbb{R}^n} \; & c^\top x \\
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+# ```
+# in which an affine function $Ax - b$ belongs to a cone $\mathcal{K}$ and the
+# variables are free.
+
+# It is trivial to convert between these two representations, for example, to go
+# from the geometric conic form to the standard conic form we introduce slack
+# variables $y$:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
```
jd-foster, line 54:
```suggestion
#     \min_{x \in \mathbb{R}^n} \; & c^\top x \\
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+#               \;\;\text{s.t.} & [A -I] [x; y] = b \\
+#                               & [x; y] \in \mathbb{R}^n \times \mathcal{K}
+# \end{align}
+# ```
+# and to go from the standard conic form to the geometric conic form, we can
+# rewrite the equality constraint as a function belonging to the `{0}` cone:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
```
jd-foster, line 63:
```suggestion
#     \min_{x \in \mathbb{R}^n} \; & c^\top x \\
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+# improve the performance of some conic optimization models. There are two
+# important takeaways:
+#
+#  1. JuMP reformulates problems to meet the input requirements of the
+#     solver, potentially increasing the problem size by adding slack variables
+#     and constraints.
+#  2. Solving the dual of a conic model can be more efficient than solving the
+#     primal.
+
+# This tutorial uses the following packges
```
jd-foster, line 18:
```suggestion
# This tutorial uses the following packages
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+# these cases, JuMP reformulates only the variables and constraints as necessary
+# to convert the problem into the desired form.
+
+# ## Primal and dual formulations
+
+# Duality plays a large role in conic optimization. For a detailed description
+# of conic duality, see [Duality](@ref).
+
+# A useful observation is that if the primal problem is in standard conic form,
+# then the dual problem is in geometric conic form, and vice versa. Morever, the
```
jd-foster, line 88:
```suggestion
# then the dual problem is in geometric conic form, and vice versa. Moreover, the
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+# \end{align}
+# ```
+# in which we have a set of linear equality constraints $Ax = b$ and the
+# variables belong to a cone $\mathcal{K}$.
+#
+# The second is the _geometric_ conic form:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
+#               \;\;\text{s.t.} & A x - b \in \mathcal{K}
```
jd-foster, line 43:
```suggestion
#               \;\;\text{s.t.} \; & A x - b \in \mathcal{K}
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+
+# ## Background
+
+# Conic optimization solvers typically accept one of two input formulations.
+
+# The first is the _standard_ conic form:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
+#               \;\;\text{s.t.} & A x = b  \\
```
jd-foster, line 32:
```suggestion
#               \;\;\text{s.t.} \; & A x = b  \\
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+# in which an affine function $Ax - b$ belongs to a cone $\mathcal{K}$ and the
+# variables are free.
+
+# It is trivial to convert between these two representations, for example, to go
+# from the geometric conic form to the standard conic form we introduce slack
+# variables $y$:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} & c^\top x \\
+#               \;\;\text{s.t.} & [A -I] [x; y] = b \\
```
jd-foster, line 55:
```suggestion
#               \;\;\text{s.t.} \; & [A -I] [x; y] = b \\
```

jd-foster (changes requested):

jd-foster (commented):
Some small tweaks

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+
+set_optimizer(model_dual, Dualization.dual_optimizer(SCS.Optimizer))
+optimize!(model_dual)
+dual.(dual_c)
+value.(y)
+
+# ## When to use `dual_optimizer`
+
+# Because it can make the problem larger or smaller, depending on the problem
+# and the choice of solver, there is is no definitive rule on when you should
```
jd-foster, line 204:
```suggestion
# and the choice of solver, there is no definitive rule on when you should
```

jd-foster (approved):

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,208 @@
...
+#     If you haven't seen conic duality before, try deriving the dual problem
+#     based on the description in [Duality](@ref). You'll need to know that the
+#     dual cone of [`PSDCone`](@ref) is the [`PSDCone`](@ref).
+
+# When we solve `model_primal` with `SCS.Optimizer`, SCS reports three variables
+# (`variables n: 3`), five rows in the constraint matrix (`constraints m: 5`),
+# and five non-zeros in the matrix (`nnz(A): 5`):
+
+set_optimizer(model_primal, SCS.Optimizer)
+optimize!(model_primal)
```
odow, line 129:
```suggestion
optimize!(model_primal)
nothing
```

odow:
Seems like this is an SCS-specific bug:
```Julia
julia> using JuMP

julia> import Documenter

julia> import Clarabel

julia> import SCS

julia> function foo(optimizer)
           model_primal = Model(optimizer)
           @variable(model_primal, X[1:2, 1:2], PSD)
           @objective(model_primal, Max, sum([1 -1; -1 1] .* X))
           @constraint(model_primal, primal_c[i = 1:2], 1 - X[i, i] == 0)
           optimize!(model_primal)
       end
foo (generic function with 2 methods)

julia> c = Documenter.Expanders.IOCapture.capture(() -> foo(SCS.Optimizer))
(value = nothing, output = "", error = false, backtrace = Ptr{Nothing}[])

julia> c = Documenter.Expanders.IOCapture.capture(() -> foo(Clarabel.Optimizer))
(value = nothing, output = "-------------------------------------------------------------\n           Clarabel.jl v0.5.1  -  Clever Acronym              \n                   (c) Paul Goulart                          \n                University of Oxford, 2022                   \n-------------------------------------------------------------\n\nproblem:\n  variables     = 3\n  constraints   = 5\n  nnz(P)        = 0\n  nnz(A)        = 5\n  cones (total) = 2\n    : Zero        = 1,  numel = 2\n    : PSDTriangle = 1,  numel = 3\n\nsettings:\n  linear algebra: direct / qdldl, precision: Float64\n  max iter = 200, time limit = Inf,  max step = 0.990\n  tol_feas = 1.0e-08, tol_gap_abs = 1.0e-08, tol_gap_rel = 1.0e-08,\n  static reg : on, ϵ1 = 1.0e-08, ϵ2 = 4.9e-32\n  dynamic reg: on, ϵ = 1.0e-13, δ = 2.0e-07\n  iter refine: on, reltol = 1.0e-13, abstol = 1.0e-12, \n               max iter = 10, stop ratio = 5.0\n  equilibrate: on, min_scale = 1.0e-04, max_scale = 1.0e+04\n               max iter = 10\n\niter    pcost        dcost       gap       pres      dres      k/t        μ       step      \n---------------------------------------------------------------------------------------------\n  0  -2.0000e+00  -2.0000e+00  0.00e+00  2.03e-02  3.85e-01  1.00e+00  1.70e+00   ------   \n  1  -3.3229e+00  -3.3043e+00  5.64e-03  3.30e-03  7.28e-02  2.02e-01  3.19e-01  8.43e-01  \n  2  -3.9930e+00  -3.9922e+00  1.83e-04  4.17e-05  9.53e-04  3.36e-03  4.57e-03  9.90e-01  \n  3  -3.9999e+00  -3.9999e+00  1.83e-06  4.16e-07  9.52e-06  3.36e-05  4.57e-05  9.90e-01  \n  4  -4.0000e+00  -4.0000e+00  1.83e-08  4.16e-09  9.52e-08  3.36e-07  4.57e-07  9.90e-01  \n  5  -4.0000e+00  -4.0000e+00  1.83e-10  4.16e-11  9.52e-10  3.36e-09  4.57e-09  9.90e-01  \n---------------------------------------------------------------------------------------------\nTerminated with status = solved\nsolve time = 1.77ms\n", error = false, backtrace = Ptr{Nothing}[])
```

odow:
Pushed a local fix here. But I'll fix upstream SCS.jl

odow:
So I'm going to push back on the suggestion that we should use Dualization in other tutorials. This is an extra for experts thing, and the addition of `Dualization.dual_optimizer` distracts from the point of other tutorials. It's also dependent on the solver and the problem. So I can see a user getting confused why we use it for some things and not others.**

I think a compromise might be to drop some `!!! tip`s around the place that they might consider solving the dual problem, rather than actually using it.

**Pointing them to a technical page on conic duality and the implementation of a specific solver is not an answer. Even the YALMIP docs, https://yalmip.github.io/command/dualize/, https://yalmip.github.io/tutorial/automaticdualization, point out that the user will likely be confused reading their documentation:

image

blegat:
> This is an extra for experts thing, and the addition of `Dualization.dual_optimizer` distracts from the point of other tutorials. It's also dependent on the solver and the problem. So I can see a user getting confused why we use it for some things and not others.

We shouldn't distract the user, we can just refer to the dualization tutorial saying that it's not specific to this tutorial so it's not a prerequisite for understanding the tutorial. It is however very important. We already had many users on discourse that were wondering why JuMP is slower than YALMIP. And that's only a fraction of users encountering this issue. Throwing users at conic formulation without mentioning that they should check dualization is like throwing someone in a fight without a sword. The argument should rather be whether we should mention dualization on tutorials that happen to be in the SCS format so we don't need it. Moreover, if we show the SCS log in the tutorial, the user will see that the dimensions are much larger than he would expect. When you read the documentation of SCS, it does not say in which format you should input it, dual or primal. We made an arbitrary decision when writing the MOI wrapper and the user cannot guess what this is. If we don't say anything, the user should expect that we dynamically choose the right one. But we don't, we just bridge to the format that we arbitrarily chose. Since we're not doing the right thing and leave to the user the job to dualize when needed we should at least let the user know about this responsibility. Because, intuitively, the user would think that we're doing it for him.

> Even the YALMIP docs, https://yalmip.github.io/command/dualize/, https://yalmip.github.io/tutorial/automaticdualization, point out that the user will likely be confused reading their documentation:

I don't think this is what is meant. He is just referred to the previous 2 sentences that are purposely meant to be confusing (by using twice the words primal and dual) to emphasize that YALMIP is doing to nice things under the hood.
It is actually very simple. For performance reason, Dualization allows the solver to solve the dual form of the problem but it's all transparent so you don't have to understand more than that.
The same YALMIP doc page also says "this can lead to a much faster solution" and "This will however be very inefficient in YALMIP" so the whole page is rather advocating that the user should always check if this option makes the problem be solved faster.

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,218 @@
...
+
+# The purpose of this tutorial is to explain how to use [Dualization.jl](@ref) to
+# improve the performance of some conic optimization models. There are two
+# important takeaways:
+#
+#  1. JuMP reformulates problems to meet the input requirements of the
+#     solver, potentially increasing the problem size by adding slack variables
+#     and constraints.
+#  2. Solving the dual of a conic model can be more efficient than solving the
+#     primal.
```
blegat, line 16: These two takeaways are a bit negative and scary, maybe we should end with a positive note "3. Dualization allows you to transparently solve the dual simply by changing one line of your program"

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,218 @@
...
+# in which an affine function $Ax - b$ belongs to a cone $\mathcal{K}$ and the
+# variables are free.
+
+# It is trivial to convert between these two representations, for example, to go
+# from the geometric conic form to the standard conic form we introduce slack
+# variables $y$:
+# ```math
+# \begin{align}
+#     \min_{x \in \mathbb{R}^n} \; & c^\top x \\
+#               \;\;\text{s.t.} \; & [A -I] [x; y] = b \\
```
blegat, line 55:
```suggestion
#               \;\;\text{s.t.} \; & \begin{bmatrix}A & -I\end{bmatrix} [x; y] = b \\
```

blegat: Currently it looks like `I` is substracted from `A`

jd-foster:
See also #3408

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,218 @@
...
+#               \;\;\text{s.t.} & [A; I] x - [b; 0] \in \{0\} \times \mathcal{K}
+# \end{align}
+# ```
+
+# From a theoretical perspective, the two formulations are equivalent, and if
+# you implement a model in the standard conic form and pass it to a geometric
+# conic form solver (or vice versa), then JuMP will automatically reformulate
+# the problem into the correct formulation.
+
+# From a practical perspective though, the geometric to conic reformulation is
```
blegat, line 73:
```suggestion
# From a practical perspective though, the geometric to standard reformulation is
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,218 @@
...
+# particularly problematic, because the additional slack variables and
+# constraints can make the problem much larger and therefore harder to solve.
```
blegat, line 75: The standard to geometric is as problematic. It's confusing to say one is more problematic because everything that is happening on the primal side also appears in the dual side.
For instance, if you have `x in K` that you convert to geometric, you need to add `dimension(K)` free variables and a constraint in `K` while if you dualize, you don't need to add any free variable, only a constraint in `K`.

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,218 @@
...
+# because it contains fewer variables and constraints. The difference is
+# particularly noticeable on large-scale optimization problems.
+
+# ## `dual_optimizer`
+
+# Manually deriving the conic dual is difficult and error-prone. The package
+# [Dualization.jl](@ref) provides the `Dualization.dual_optimizer` meta-solver,
+# which wraps any MathOptInterface-compatible solver in an interface that
+# automatically formulates the dual of an input problem, solves the dual
+# problem, and then reports the primal solution back to the user.
```
blegat, line 179: "reports the primal solution back to the user." is confusing, one might think you mean that Dualization reports the primal solver solution to the solver as primal solution without transforming it to dual solution

odow:
Copying your comment from: https://github.com/jump-dev/JuMP.jl/pull/3399#issuecomment-1577987780

> In view of how important it can be for a conic problem to solve well, it shouldn't be an option for expert though. Everyone conic user should use it and we should try to educate users to do so

Ah! So this is our source of disagreement then. If this is the case, then yes, it should be in every conic tutorial. And we should run examples with both SCS and Clarabel (for example) to show the difference. And it isn't sufficient just to sprinkle `dual_optimizer` in a few places.

But stepping back, if it's that important, shouldn't we choose to dualize automatically? (I think cvxpy does this?) This feels like the wrong decision to force users to learn.

odow:
> we're like "try to figure it out, we know which form it is but we won't tell you". The user might also think we are not telling because it's so obvious so if it's not for him, he might think JuMP is not for him

But it is 'try it out' because if the use SCS then they should, but if they use a different solver then maybe they shouldn't. And we don't have good documentation for each solver on whether it expects the standard or geometric forms so there's no way for the average user to tell.

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,218 @@
...
+#     based on the description in [Duality](@ref). You'll need to know that the
+#     dual cone of [`PSDCone`](@ref) is the [`PSDCone`](@ref).
+
+# When we solve `model_primal` with `SCS.Optimizer`, SCS reports three variables
+# (`variables n: 3`), five rows in the constraint matrix (`constraints m: 5`),
+# and five non-zeros in the matrix (`nnz(A): 5`):
+
+set_optimizer(model_primal, SCS.Optimizer)
+optimize!(model_primal)
+Base.Libc.flush_cstdio()  #hide
```
odow, line 130: TODO: rebase this PR on latest master and remove these hacks

blegat:
> But stepping back, if it's that important, shouldn't we choose to dualize automatically? (I think cvxpy does this?) This feels like the wrong decision to force users to learn.

We probably should but unless it's pure standard or pure geometric, we don't really know which form to use. One idea would be to see the total sum of bridge cost for dualized and non-dualized version and choose that one. Then should be multiply the bridge cost by the number of constraints of that type ? So we could have an automatic mode doing this and then a manual mode where the user sets whether he want it dualized or not.

> But it is 'try it out' because if the use SCS then they should, but if they use a different solver then maybe they shouldn't. And we don't have good documentation for each solver on whether it expects the standard or geometric forms so there's no way for the average user to tell.

Yes, it is "try it out". The message is "always try both", I think that's what your tutorial conveys when you say that if they plan to solve it multiple times or for a large instance then first try both and then choose and then run the large computational workload. Understanding why one is better is for the curious users but they don't need to understand why (if there was a need to understand why something works then ML/AI wouldn't be a thing :D)

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,257 @@
...
+
+#-
+
+dual.(dual_c)
+
+#-
+
+value.(y)
+
+# ## A mixed example
```
odow, line 209: I added this additional section with a non-obvious example

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,257 @@
...
+
+# ## When to use `dual_optimizer`
+
+# Because it can make the problem larger or smaller, depending on the problem
+# and the choice of solver, there is no definitive rule on when you should
+# use `dual_optimizer`. However, you should try `dual_optimizer` if your conic
+# optimization problem takes a long time to solve, or if you need to repeatedly
+# solve similarly structured problems with different data. In some cases solving
+# the dual instead of the primal can make a large difference.
+
```
odow, line 256:
```suggestion
```

docs/src/tutorials/conic/dualization.jl
```diff
@@ -0,0 +1,255 @@
...
+#               \;\;\text{s.t.} & \begin{bmatrix}A\\I\end{bmatrix} x - \begin{bmatrix}b\\0\end{bmatrix} \in \{0\} \times \mathcal{K}
+# \end{align}
+# ```
+
+# From a theoretical perspective, the two formulations are equivalent, and if
+# you implement a model in the standard conic form and pass it to a geometric
+# conic form solver (or vice versa), then JuMP will automatically reformulate
+# the problem into the correct formulation.
+
+# From a practical perspective though, the reformulations are because the
```
blegat, line 76:
```suggestion
# From a practical perspective though, the reformulations are problematic because the
```

docs/src/tutorials/conic/dualization.jl
blegat: I like it!

blegat (approved):
LGTM once the missing word is added

PR #3406, opened 2023-06-06
[docs] fix SCS log in start values tutorial

odow:
Can remove once https://github.com/jump-dev/SCS.jl/pull/275 is tagged.

### DIFF
--- a/docs/src/tutorials/conic/start_values.jl
+++ b/docs/src/tutorials/conic/start_values.jl
@@ -67,16 +67,17 @@ model = Model(SCS.Optimizer)
 @constraint(model, sum(x) <= 1)
 @objective(model, Max, sum(i * x[i] for i in 1:3))
 optimize!(model)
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/conic/start_values.jl
```diff
@@ -67,16 +67,17 @@ model = Model(SCS.Optimizer)
 @constraint(model, sum(x) <= 1)
 @objective(model, Max, sum(i * x[i] for i in 1:3))
 optimize!(model)
+Base.Libc.flush_cstdio()  #hide
 
-# By looking at the log (not shown in Documenter due to a bug), we can see that
-# SCS took 100 iterations to find the optimal solution. Now we set the optimal
-# solution as our starting point:
+# By looking at the log, we can see that SCS took 100 iterations to find the optimal
```
odow, line 72:
```suggestion
# By looking at the log, we can see that SCS took 75 iterations to find the optimal
```

PR #3412, opened 2023-06-07
[docs] tidy parts of OPF tutorial and reduce tolerances for stability

odow:
The docs are a little flakily because of numerical error solving the SDPs in the OPF tutorial.

This PR relaxes the tolerances by a lot to minimize the risk of solve failure, and also includes a number of formatting changes for style consistency.

### DIFF
--- a/docs/src/tutorials/applications/optimal_power_flow.jl
+++ b/docs/src/tutorials/applications/optimal_power_flow.jl
@@ -26,8 +26,8 @@
 # This tutorial takes a matrix-oriented approach focused on network nodes
 # that simplifies the construction of semidefinite programs.
 # Another approach is to formulate the problem focusing on network lines
-# where it is easier to work with flow constraints. A general approach is provided by
-# the Julia/JuMP package [PowerModels.jl](https://lanl-ansi.github.io/PowerModels.jl/stable/),
+# where it is easier to work with flow constraints. A general approach is
+# provided by the Julia/JuMP package [PowerModels.jl](https://lanl-ansi.github.io/PowerModels.jl/stable/),
 # an open-source framework to a broad range of power flow model formulations
 # along with utilities for working with detailed network data.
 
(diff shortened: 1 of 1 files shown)


jd-foster:
Thanks, looks goods and more streamlined. Definitely needed to back off the tolerances (revisit at a later stage re: Clarabel).

jd-foster:
Didn't think the Ipopt ref was relevant?

docs/src/tutorials/applications/optimal_power_flow.jl
```diff
@@ -564,12 +536,3 @@ DataFrames.DataFrame(;
...
 # [MATPOWER manual](https://matpower.org/docs/MATPOWER-manual.pdf); see especially Appendix B "Data File Format" and Table B-3.
-
-# **PowerModels.jl**:
-# the Julia/JuMP package [PowerModels.jl](https://lanl-ansi.github.io/PowerModels.jl/stable/) provides
-# an open-source framework for a broad range of power flow formulations along with utilities for working with detailed network data.
-
-# **Ipopt solver**:
-# Wächter, A., Biegler, L.
-# [_On the implementation of an interior-point filter line-search algorithm for large-scale nonlinear programming._](https://doi.org/10.1007/s10107-004-0559-y)
-# Math. Program. 106, 25—57 (2006).
```
odow, line 575: I removed PowerModels because we already link to it above. And I removed the paper reference to Ipopt because I don't know if it's relevant for the tutorial? Users don't need to read the paper to use Ipopt.

jd-foster: That's fine. I guess we would be referencing every solver paper if in every tutorial otherwise.

odow: There's an argument for "you should cite the paper if you use Ipopt," but that's a different discussion.

jd-foster: Yes, that was in my mind when I added it. But it's a broader discussion for the whole docs.

Issue #3214, opened 2023-02-10
[Containers] support ArrayInterface.jl traits


ParadaCarleton:
As far as I'm aware, JuMP's containers/labeled arrays aren't used outside of JuMP itself. This can cause a lot of problems and incompatibilities when trying to work with data from packages other than JuMP.

DimensionalData.jl is more widely-used, e.g. throughout the Julia Geostats ecosystem and in ArviZ.jl. Would it make sense to switch JuMP's containers out with DimensionalData.jl DimArrays for compatibility?

jd-foster:
I guess the short answer is that the principle in JuMP / MathOptInterface is to try to keep external dependencies to a minimum. In this regard, there seems to be small gains from "compatibility" in the abstract considering that the required functionality is fully implemented within the package with an `AbstractArray` interface.

blegat:
What we need is that the array directly generalizes `Array` so that the user does not have to know about whether the container `Array` or `DenseAxisArray` or `SparseAxisArray`.
This is what prevents us to use a `Dict` for `SpaceAxisArray`, because a `Dict` does not just behave like an `Array` with custom indices with missing entries. What we want there is really like a `SparseArray` with custom indexing.
The issue with `AxisArrays` is that if you pass integer indices, it assumes that you are directly indexing the underlying array:
```julia
julia> using AxisArrays

julia> A = AxisArray(1:2; i = 2:3)
1-dimensional AxisArray{Int64,1,...} with axes:
    :i, 2:3
And data, a 2-element UnitRange{Int64}:
 1
 2

julia> A[2]
2
```
We would need `A[2]` to be `1` here.
There is the same issue with `DimensionalData`:
```julia
julia> A = ones(X(2:3), Y(3:4))
2×2 DimArray{Float64,2} with dimensions: 
  X Sampled{Int64} 2:3 ForwardOrdered Regular Points,
  Y Sampled{Int64} 3:4 ForwardOrdered Regular Points
    3    4
 2  1.0  1.0
 3  1.0  1.0

julia> A[2, 3]
ERROR: BoundsError: attempt to access 2×2 Matrix{Float64} at index [2, 3]
Stacktrace:
 [1] getindex
   @ ./array.jl:925 [inlined]
[...]

PR #3413, opened 2023-06-07
Add DimensionalData extension

odow:
This PR is a replacement for https://github.com/jump-dev/JuMP.jl/pull/3383. I've added a new branch to the `jump-dev` remote so that we can generate documentation previews.

https://jump.dev/JuMP.jl/previews/PR3413/extensions/DimensionalData/

I'm going to mark this as closing #3214. I don't think we need to implement methods for DenseAxisArray, we just need to make it easier to integrate with other array packages.

Closes https://github.com/jump-dev/JuMP.jl/issues/3214


docs/make.jl
```diff
@@ -186,11 +186,16 @@ for (solver, data) in TOML.parsefile(joinpath(@__DIR__, "packages.toml"))
...
+    _LIST_OF_EXTENSIONS,
+    "rafaqz/DimensionalData.jl" => "extensions/DimensionalData.md",
+)
+
 # Sort, with jump-dev repos at the start.
 sort!(_LIST_OF_SOLVERS; by = x -> (!startswith(x[1], "jump-dev/"), x[1]))
 sort!(_LIST_OF_EXTENSIONS; by = x -> (!startswith(x[1], "jump-dev/"), x[1]))
 pushfirst!(_LIST_OF_SOLVERS, "Introduction" => "packages/solvers.md")
-pushfirst!(_LIST_OF_EXTENSIONS, "Introduction" => "packages/extensions.md")
+pushfirst!(_LIST_OF_EXTENSIONS, "Introduction" => "extensions/extensions.md")
```
odow, line 198:
```suggestion
pushfirst!(_LIST_OF_EXTENSIONS, "Introduction" => "extensions/introduction.md")
```

odow:
So this PR now has tests and documentation. 

It seems reasonable to merge, but it's worth a discussion on the long-term maintenance cost vs the additional features that are offered.

odow:
Discussed breaking changes in DimensionalData here: https://github.com/rafaqz/DimensionalData.jl/issues/491. None are immediately planned, but there might be a couple before v1.0.

Any changes are very unlikely to impact the code in JuMP, so consensus from the monthly call seemed to be that we should move forward with this case, and re-evaluate other extensions on a case-by-case basis.

I am interested to see how the Pkg extensions go more generally, and whether they lead to questions on the forum.

odow:
Discussed and approved during the hackathon.

PR #3408, opened 2023-06-06
Add basic dualization comparison to simple examples

jd-foster:
Happy for this be a separate PR: produces this table at the end of `simple_examples`:
```
Model Name         #PrimalVars #Cons #DualVars #DualCons
------------------------------------------------------------
theta_problem           15      7       6       1
minimum_distortion      11      15      14      15
max_cut                 10      5       4       1
k_means_clustering      21      29      28      22
```

https://jump.dev/JuMP.jl/previews/PR3408/tutorials/conic/simple_examples/

PR #3416, opened 2023-06-08
Add basic dualization comparison to simple examples

blegat:
Alternative to https://github.com/jump-dev/JuMP.jl/pull/3408

Output is
```
Model Name         #PrimalVars #Cons #DualVars #DualCons
------------------------------------------------------------
Maximum cut             10      5       4       1
K-means                 21      29      28      22
Correlation             6       8       7       5
Minimum distortion      11      15      14      15
Theta                   15      7       6       1
Robust                  12      4       24      10
```

It's kind of a proof of concept, just counting the number of constraints is a bit deceiving, we should have one column for each constraint type.
I makes me think that it would be nice to have a nice way to print the inner model the same way as `show(::AbstractModel)`.
We're still throwing the user in a fight with a sword and a bow without a good way to see which one fits best.
If we had a way to have the same output as `show(::AbstractModel)` but for the inner optimizer, it's be great.
Could be something like
```julia
julia> show_inner_models(model)
A Bridge layer
Minimization problem with:
Variables: 2
`MOI.ScalarAffineFunction{Float64}`-in-`MOI.EqualTo{Float64}`: 10 constraints
...

A CachingOptimizer{SCS.OptimizerCache,SCS.Optimizer}
Minimization problem with:
Variables: 2
`MOI.VectorAffineFunction{Float64}`-in-`MOI.Zeros`: 10 constraints
...
```
and then
```julia
julia> show_inner_models(dualized_model)
A Bridge layer
Minimization problem with:
Variables: 2
`MOI.ScalarAffineFunction{Float64}`-in-`MOI.EqualTo{Float64}`: 10 constraints
...

A DualOptimizer
Minimization problem with:
Variables: 2
`MOI.VectorAffineFunction{Float64}`-in-`MOI.Zeros`: 10 constraints
...

A CachingOptimizer{SCS.OptimizerCache,SCS.Optimizer}
Minimization problem with:
Variables: 2
`MOI.VectorAffineFunction{Float64}`-in-`MOI.Zeros`: 10 constraints
...
```
To walk through layers, we would do something like
```julia
MOI.inner_optimizer(::MOI.ModelLike) = nothing
# We implement it for CachingOptimizer, LazyBridgeOptimizer and Dualization.DualOptimizer

function show_inner_models(model::MOI.ModelLike)
    _show_model(model) # Could do some try catch here in case the solver does not support the getters
    show_inner_models(MOI.inner_optimizer(model))
end
```
and `MOI.inner_optimizer` would also be useful for `JuMP.unsafe_backend`. It would also be a way for the user to inspect the list of layers.
Meta-solver could also implement `MOI.inner_optimizer` unless they have two inner optimizer in which case they will just not implement it.

### DIFF
--- a/docs/src/tutorials/conic/simple_examples.jl
+++ b/docs/src/tutorials/conic/simple_examples.jl
@@ -11,6 +11,7 @@
 # This tutorial makes use of the following packages:
 
 using JuMP
+import Dualization
 import LinearAlgebra
 import Plots
 import Random
@@ -56,11 +57,11 @@ function svd_cholesky(X::AbstractMatrix)
     return (F.U * D)'
 end
 
-function solve_max_cut_sdp(weights)
+function solve_max_cut_sdp(weights, solver; verbose = true)
     N = size(weights, 1)
     ## Calculate the (weighted) Laplacian of the graph: L = D - W.
     L = LinearAlgebra.diagm(0 => weights * ones(N)) - weights
-    model = Model(SCS.Optimizer)
+    model = Model(solver)
     set_silent(model)
     @variable(model, X[1:N, 1:N], PSD)
     for i in 1:N
@@ -76,9 +77,11 @@ function solve_max_cut_sdp(weights)
     cut = [LinearAlgebra.dot(r, V[:, i]) > 0 for i in 1:N]
     S = findall(cut)
     T = findall(.!cut)
-    println("Solution:")
-    println(" (S, T) = ({", join(S, ", "), "}, {", join(T, ", "), "})")
-    return S, T
+    if verbose
+        println("Solution:")
+        println(" (S, T) = ({", join(S, ", "), "}, {", join(T, ", "), "})")
+    end
+    return S, T, model
 end
 
 # Given the graph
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/conic/simple_examples.jl
```diff
@@ -416,7 +435,58 @@ function example_robust_uncertainty_sets()
...
+
+function _num_variables(model::MOI.ModelLike)
+    return MOI.get(model, MOI.NumberOfVariables())
+end
+function _num_constraints(model::MOI.ModelLike)
+    return sum(
+        MOI.get(model, MOI.NumberOfConstraints{F,S}())
+        for (F, S) in MOI.get(model, MOI.ListOfConstraintTypesPresent())
+    )
+end
```
odow, line 458: This is too complicated for a "simple examples" tutorial

odow:
I don't think we need this. The Dualization tutorial explains how to solve the primal and dual with SCS, and how to interpret the log to see which problem was smaller. I don't want to complicate the "simple examples" tutorial with this. And we definitely don't want to encourage general users to explore the inner workings of a JuMP model. "Try both" is okay.

jd-foster:
This could be the basis for a nice blog post introducing the Dualization package, or adapted into its package documentation.

odow:
> or adapted into its package documentation

:+1: Nice idea. This makes a lot more sense in the Dualization docs than in JuMP.

odow:
The JuMP docs should just hint at the "try this out, sometimes it's better." The Dualization docs have a lot more scope to go into the details for users who want that.

PR #3420, opened 2023-06-15
SCIP licensed under Apache 2.0 since v8.0.3

spatil47:
SCIP has been [relicensed to Apache 2.0](https://scipopt.org/index.php#license) from v8.0.3 onward.


jd-foster:
Thanks; in fact, this is also stated in the SCIP.jl readme.

Issue #3270, opened 2023-03-07
Starting values and Objective.SlackBridge


blegat:
While working on https://github.com/jump-dev/DiffOpt.jl/pull/231, I found out that there might be some unexpected behavior to
```julia
julia> using SCS, JuMP

julia> model = Model(SCS.Optimizer);

julia> @variable(model, x);

julia> @objective(model, Min, x^2);

julia> optimize!(model)
------------------------------------------------------------------
	       SCS v3.2.1 - Splitting Conic Solver
	(c) Brendan O'Donoghue, Stanford University, 2012
------------------------------------------------------------------
problem:  variables n: 2, constraints m: 3
cones: 	  q: soc vars: 3, qsize: 1
settings: eps_abs: 1.0e-04, eps_rel: 1.0e-04, eps_infeas: 1.0e-07
	  alpha: 1.50, scale: 1.00e-01, adaptive_scale: 1
	  max_iters: 100000, normalize: 1, rho_x: 1.00e-06
	  acceleration_lookback: 10, acceleration_interval: 10
lin-sys:  sparse-direct-amd-qdldl
	  nnz(A): 3, nnz(P): 0
------------------------------------------------------------------
 iter | pri res | dua res |   gap   |   obj   |  scale  | time (s)
------------------------------------------------------------------
     0| 2.07e+01  1.00e+00  2.83e+01 -1.41e+01  1.00e-01  1.37e-03 
    25| 1.14e-05  2.44e-08  1.61e-05  8.04e-06  1.00e-01  1.39e-03 
------------------------------------------------------------------
status:  solved
timings: total: 1.39e-03s = setup: 1.35e-03s + solve: 4.27e-05s
	 lin-sys: 3.43e-06s, cones: 2.57e-06s, accel: 7.72e-07s
------------------------------------------------------------------
objective = 0.000008
------------------------------------------------------------------

julia> set_start_values(model)

julia> model.moi_backend.optimizer.model.model_cache.varattr
Dict{MathOptInterface.AbstractVariableAttribute, Dict{MathOptInterface.VariableIndex, Any}} with 1 entry:
  VariablePrimalStart() => Dict(VariableIndex(1)=>0.0)

julia> model.moi_backend.optimizer.model.model_cache.conattr
Dict{MathOptInterface.AbstractConstraintAttribute, Dict{MathOptInterface.ConstraintIndex, Any}}()

julia> optimize!(model)
------------------------------------------------------------------
[...]

PR #3422, opened 2023-06-23
Set SlackBridgePrimalDualStart in set_start_values

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3270


src/optimizer_interface.jl
```diff
@@ -1151,6 +1151,8 @@ function set_start_values(
     for (ci, dual_start) in constraint_dual
         set_dual_start_value(ci, dual_start)
     end
+    # Needed for models which bridge `min f(x)` into `min t; t >= f(x)`.
+    MOI.set(model, MOI.Bridges.Objective.SlackBridgePrimalDualStart(), true)
```
odow, line 1155:
```suggestion
    MOI.set(model, MOI.Bridges.Objective.SlackBridgePrimalDualStart(), nothing)
```

blegat:
Just need a test

odow:
> Just need a test

Our issue now is that all the solvers in the docs support quadratic objectives :smile:

So not really sure how easy it is to test this.

Issue #2093, opened 2019-11-02
Ease addition of matrix cones


blegat:
The `PSDCone` allows to easily add `MOI.PositiveSemidefiniteConeTriangle` and `MOI.PositiveSemidefiniteConeSquare`.
There are several other matrix cones, that should work exactly like `PSDCone` for consistency:
* `SemidefiniteInterval`, `SemidefiniteGreaterThan` and `SemidefiniteLessThan` that create `MOI....Triangle` and `MOI....Square`, see https://github.com/JuliaOpt/MathOptInterface.jl/issues/942
* `EigenvalueInterval`, `EigenvalueGreaterThan` and `EigenvalueLessThan` that create `MOI....Triangle` and `MOI....Square`, see https://github.com/JuliaOpt/MathOptInterface.jl/issues/942
* `SumOfSquares.DDCone` that create `SumOfSquares.DiagonalyDominantConeTriangle` and `SumOfSquares.DiagonalyDominantConeSquare`
* `SumOfSquares.SDDCone` that create `SumOfSquares.ScaledDiagonalyDominantConeTriangle` and `SumOfSquares.ScaledDiagonalyDominantConeSquare`

To allow reusing the code for `PSDCone` between these different sets we can create a `AbstractSymmetricMatrixCone` set and the function `triangle_vectorized_cone` and `square_vectorized_cone` and then all the code should work for any `AbstractSymmetricMatrixCone`.

Requires https://github.com/JuliaOpt/MathOptInterface.jl/issues/942

odow:
@blegat: We haven't made any progress on this. I see what we closed #942, so I'm not sure what `Requires #942` is for. Can this be closed? We seem to be getting on okay without it.

blegat:
I'll take a look whilr adding affine hermitian constraints. It would still be helpful

odow:
Bump. Is there anything to do here?

blegat:
Yes, I still need to get at it

odow:
Can you point me in the right direction? I still don't really understand what you're after.

blegat:
I'm going to take a look

odow:
Can you be a little more explicit in what you want here? For example, links to the bits in other packages that would be de-duplicated?

blegat:
These are the cones: https://github.com/jump-dev/SumOfSquares.jl/blob/8476c96d0fc182377a3a585222714061925092ab/src/diagonally_dominant.jl
[...]

PR #3424, opened 2023-06-26
Add fallback for MOI.AbstractSymmetricMatrixSet{Triangle,Square}

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/2093

@blegat is this what you had in mind?

It should allow something like
```julia
@variable(model, X[1:2, 1:2], Symmetric)
@constraint(model, X in SumOfSquares.ScaledDiagonallyDominantConeTriangle(2))
```

But what is a good model to test?

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -781,7 +781,7 @@ end
 
 function build_constraint(
     _error::Function,
-    x::Matrix,
+    x::AbstractMatrix,
     set::MOI.AbstractVectorSet,
 )
     return _error(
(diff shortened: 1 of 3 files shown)


blegat:
I find it quite neat!
Once we have that, why no `@constraint(model, (t, Symmetric(ones(2, 2))) in RootDetConeTriangle(2))` ?

odow:
The main reason against `(t, X)` would be that we couldn't recover the function from the MOI model alone. We could add a new shape though. 

@matbesancon suggested this a while ago: https://github.com/jump-dev/MathOptInterface.jl/issues/1624

But it was in the context of MOI. It makes a bit more sense for JuMP. It'd simplify using RootDetConeTriangle if you didn't need to manually construct the triangular vector.

odow:
I have a proof of concept for the tuple function. I'll merge this, then open a new PR.

PR #3427, opened 2023-06-27
Add support for Tuple functions with epigraph matrix sets

odow:
@blegat suggested this in #3424. Previously suggested by @matbesancon in https://github.com/jump-dev/MathOptInterface.jl/issues/1624.

Preview: https://jump.dev/JuMP.jl/previews/PR3427/tutorials/conic/tips_and_tricks/#RootDetCone

TODO:

 - [x] A nicer error message when a tuple passed instead of a vector


odow:
This certainly makes a lot of sense. It make the log/root det cones a lot nicer to use:

https://jump.dev/JuMP.jl/previews/PR3427/tutorials/conic/tips_and_tricks/#RootDetCone

But I worry about special casing the syntax to only a few cones.

You could wonder why the rest of the epigraph cones don't support the same syntax, or why cones with fixed input dimension like `Exponential` require a vector instead of a tuple.

We could add support one at a time, but I don't know if there's a generic way to support them all.

matbesancon:
> We could add support one at a time, but I don't know if there's a generic way to support them all.

I would say this is fairly natural for all epigraph-type of cones

src/sd.jl
```diff
@@ -768,3 +768,97 @@ function build_variable(
...
+
+## Example
+
+```jldoctest
+julia> shape = TupleShape(ScalarShape() => 1, SymmetricMatrixShape(2) => 3);
+```
+"""
+struct TupleShape{T} <: AbstractShape
+    shapes::T
+    TupleShape(args::Pair{<:AbstractShape,Int}...) = new{typeof(args)}(args)
```
blegat, line 787: Why not a tuple of shapes ?

test/test_constraint.jl
```diff
@@ -1684,4 +1684,93 @@ function test_eltype_for_constraint_primal_complex_float64()
...
+    @test reshape_set(obj.set, obj.shape) == obj.set
+    return
+end
+
+function test_tuple_shape_unsupported_error()
+    model = Model()
+    @variable(model, t)
+    @variable(model, x[1:2])
+    err = ErrorException(
+        "In ` @constraint(model, (t,) in MOI.ZeroOne())`: " *
```
odow, line 1759:
```suggestion
        "In `@constraint(model, (t,) in MOI.ZeroOne())`: " *
```

src/sd.jl
odow: We need to be able to tell the dimension of each component when converting back from the vector form to the tuple form.

test/test_constraint.jl
```diff
@@ -1684,4 +1684,93 @@ function test_eltype_for_constraint_primal_complex_float64()
...
+    )
+        c = @constraint(model, (u, v, w) in set)
+        obj = constraint_object(c)
+        @test obj.func == [u, v, w]
+        @test reshape_vector(obj.func, obj.shape) == (u, v, w)
+        @test reshape_set(obj.set, obj.shape) == obj.set
+    end
+    for set in
+        (MOI.SecondOrderCone(4), SecondOrderCone(), MOI.GeometricMeanCone(4))
+        c = @constraint(model, (t, x) in set)
```
odow, line 1708: This is now a generic implementation that works for arbitrary cones. If it has a downside, it's that it's too permissive. Effectively, the tuple vectorized by concatenating the elements, and doesn't concern itself with whether the structure of the tuple makes sense for the set. So these three constraints are all valid:
```julia
@constraint(model, (t, x) in SecondOrderCone())
@constraint(model, (t, x[1], x[2], x[3]) in SecondOrderCone())
@constraint(model, (t, x[1], x[2:3]) in SecondOrderCone())
```
We don't have a way of saying that `SecondOrderCone` is an epigraph with `Scalar,Vector`, and that `RotatedSecondOrderCone` is a perspective epigraph with `Scalar,Scalar,Vector`. We just know the full dimension.

src/sd.jl
blegat: Can't the dimension be inferred from the shape ? I would rather do that than store the same info twice with the risk of inconsistency

odow: For scalar and the matrix shapes, yes. `VectorShape` doesn't have the information.

So I guess we could drop the dimension, but we'd need to limit things to exactly one `VectorShape` in the tuple. It seemed simpler just to record the input dimensions.

docs/src/tutorials/conic/tips_and_tricks.jl
```diff
@@ -232,7 +232,7 @@ set_silent(model)
 @objective(model, Max, -t)
 @constraint(model, sum(x) == 1)
 @constraint(model, A * x .<= b)
-@constraint(model, [t; ones(n); x] in MOI.RelativeEntropyCone(2n + 1))
+@constraint(model, (t, ones(n), x) in MOI.RelativeEntropyCone(2n + 1))
```
odow, line 235: @blegat here's a case where we wouldn't be able to convert from the vector form to the tuple without the dimension of `VectorShape` provided on input because there are two `VectorShape` elements.

blegat (approved):

blegat:
This looks really nice, now what about `@variable(model, (t, X[1:3,1:3]) in MOI.RootDetConeTriangle(3))`, this will address your comment in https://github.com/jump-dev/MathOptInterface.jl/pull/2198#issuecomment-1585211284

odow:
This is a fairly big syntax change, so @mlubin better take a look.

odow:
```julia
@variable(model, (t, X[1:3,1:3]) in MOI.RootDetConeTriangle(3))
```
This seems. Hard. It requires a massive change to how we currently do parsing:
```Julia
julia> using JuMP

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, (t, x[1:2]) in SecondOrderCone())
ERROR: LoadError: Expression (t, x[1:2]) cannot be used as a name.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] _get_name(c::Expr)
   @ JuMP.Containers ~/.julia/packages/JuMP/Gra9Y/src/Containers/macro.jl:14
 [3] var"@variable"(__source__::LineNumberNode, __module__::Module, args::Vararg{Any, N} where N)
   @ JuMP ~/.julia/packages/JuMP/Gra9Y/src/macros.jl:2619
in expression starting at REPL[96]:1
```

docs/src/tutorials/conic/ellipse_approx.jl
```diff
@@ -109,7 +109,7 @@ m, n = size(S)
     [i in 1:m],
     S[i, :]' * Z * S[i, :] - 2 * S[i, :]' * z + s <= 1,
 )
-@constraint(model, [t; vec(Z)] in MOI.RootDetConeSquare(n))
```
mlubin, line 112: We use the syntax in examples, but what about documenting it in the general case, along with its limitations?
> Effectively, the tuple vectorized by concatenating the elements, and doesn't concern itself with whether the structure of the tuple makes sense for the set.

odow: I guess my concern was more: is this a good idea and should we move forward with it. It adds two ways to write each constraint (vector or tuple) which might be confusing for users.

mlubin:
I wonder if tuples are too magical and we should do something more explicit, e.g.,
```
@constraint(model, concat(t, x) in SecondOrderCone())
@constraint(model, concat(t, x[1], x[2], x[3]) in SecondOrderCone())
@variable(model, concat_define(t, X[1:3,1:3]) in MOI.RootDetConeTriangle(3))
```
`concat_define` is weird, but we're running up against the issue of `@variable` being weird syntax. We never resolved https://github.com/jump-dev/JuMP.jl/issues/692.

odow:
I think the tuple syntax in `@constraint` is the most natural. 

I'd be against `concat`. You can already use `vcat`
```julia
@constraint(model, vcat(t, x) in SecondOrderCone())
@constraint(model, vcat(t, x[1], x[2], x[3]) in SecondOrderCone())
```
I guess the only benefit would be that we could define `concat(::Variable, ::Symmetric{Matrix})` which would return the triangle. But we could also define a `upper_triangular` which would achieve the same thing:
```julia
@constraint(model, vcat(t, upper_triangular(X)) in SecondOrderCone())
```

The `@variable` syntax is going to break regardless. We _could_ parse the tuples at macro-expansion time. It's just a bit of work.

There'd also be some nasty edge-cases that aren't or the form `(::Variable, ::Vector{Variable})`. We probably want the first one only of these:
```julia
@variable(model, (t, w[1:2], x[1:2]) in MOI.RelativeEntropyCone(5))
@variable(model, (t, w, x) in MOI.RelativeEntropyCone(5))
@variable(model, (t, x) in MOI.RelativeEntropyCone(5))
@variable(model, (t, x[1:5]) in MOI.RelativeEntropyCone(5))
```

odow:
I'd still like to push forward with this. I think it makes some of the conic examples a lot nicer.

docs/src/tutorials/conic/experiment_design.jl
```diff
@@ -210,11 +210,8 @@ set_silent(dOpt)
 @variable(dOpt, t)
 @objective(dOpt, Max, t)
 @constraint(dOpt, sum(np) <= n)
-E = V * LinearAlgebra.diagm(0 => np ./ n) * V'
-@constraint(
-    dOpt,
-    [t, 1, (E[i, j] for i in 1:q for j in 1:i)...] in MOI.LogDetConeTriangle(q)
-)
+E = LinearAlgebra.Symmetric(V * LinearAlgebra.diagm(0 => np ./ n) * V')
+@constraint(dOpt, (t, 1.0, E) in MOI.LogDetConeTriangle(q))
```
odow, line 214: What is the error thrown if `E` is not symmetric

docs/src/tutorials/conic/experiment_design.jl
```diff
@@ -210,11 +210,8 @@ set_silent(dOpt)
 @variable(dOpt, t)
 @objective(dOpt, Max, t)
 @constraint(dOpt, sum(np) <= n)
-E = V * LinearAlgebra.diagm(0 => np ./ n) * V'
-@constraint(
-    dOpt,
-    [t, 1, (E[i, j] for i in 1:q for j in 1:i)...] in MOI.LogDetConeTriangle(q)
```
odow, line 216: Maybe this PR is really just asking for a nicer way to `vec` a symmetric matrix

odow:
Closing in favor of https://github.com/jump-dev/JuMP.jl/pull/3456

PR #3430, opened 2023-07-05
improve problem statement in cannery example of tutorial

SimonEnsemble:
feel free to reject without hurting my feelings.
* decision variables were not defined. now they are.
* having `d_{p,m}` be a distance makes less sense than a per-case shipment cost (as stated in [Dantzig's textbook](https://www.rand.org/content/dam/rand/pubs/reports/2007/R366part1.pdf.) pg. 36)

### DIFF
--- a/docs/src/tutorials/linear/cannery.jl
+++ b/docs/src/tutorials/linear/cannery.jl
@@ -31,8 +31,10 @@ import Test  #src
 # Each production plant ``p`` has a capacity, ``c_p``, and each market ``m``
 # has a demand ``d_m``. The distance from plant to market is ``d_{p,m}``.
 
-# With a little effort, we can formulate our problem as the following linear
-# program:
+# We wish to find the distribution plan ``x_{p,m}`` (the quantity of cans
+# shipped from plant `p` to market `m`) for ``p \in P`` and ``m \in M``  
+# that minimizes the shipping costs. We can formulate our problem as the
+# following linear program:
 # ```math
 # \begin{aligned}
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/linear/cannery.jl
```diff
@@ -25,14 +25,17 @@ import Test  #src
...
 
-# Each production plant ``p`` has a capacity, ``c_p``, and each market ``m``
-# has a demand ``d_m``. The distance from plant to market is ``d_{p,m}``.
+# Each production plant ``p`` has a capacity ``c_p``, and each market ``m``
+# has a demand ``d_m``. The shipping cost per case of cans from plant ``p`` 
+# to market ``m`` is ``d_{p,m}``.
 
-# With a little effort, we can formulate our problem as the following linear
-# program:
+# We wish to find the distribution plan ``x_{p,m}`` (the # of cases of cans
```
jd-foster, line 35:
```suggestion
# We wish to find the distribution plan ``x_{p,m}``, the number of cases of cans
```

docs/src/tutorials/linear/cannery.jl
```diff
@@ -25,14 +25,17 @@ import Test  #src
...
+# to ship from plant ``p`` to market ``m``) for ``p \in P`` and ``m \in M``
```
jd-foster, line 36:
```suggestion
# to ship from plant ``p`` to market ``m``, for ``p \in P`` and ``m \in M``  
```

jd-foster (commented):
Looks good to me apart from the small suggestion.

odow (approved):

PR #3439, opened 2023-07-30
Add Grisu dependency

blegat:
It seems JuMP v0.18 is broken on Julia v1.6 because we are depending on Grisu which was moved as a separate package.
See https://discourse.julialang.org/t/jumper-and-grisu/102258/4


mlubin:
Do tests pass with this fix?

blegat:
I was hoping the ci would run but apparently it does not. Maybe we should first add the ci to the branch and then make a PR against that

odow:
When I try the tests locally, I get:

image

I don't really want to go down the 0.18 rabbit hole. There's also the problem of getting a solver that works, etc

blegat:
The Julia v1.x releases were supposed to be non-breaking so I thought it would be an easy fix

odow:
I didn't look into the details. If I remember correctly, there were some breaking changes in the AST that we had to work around. There used to be a bunch of `if VERSION >= v"1.X"` that we removed when we updated the lower bound to 1.6?

odow:
Here is the culprit:
```Julia
julia> dump(:(.>=))
Symbol .>=

julia> VERSION
v"1.0.5"
```

```Julia
julia> dump(:(.>=))
Expr
  head: Symbol .
  args: Array{Any}((1,))
    1: Symbol >=

julia> VERSION
v"1.6.7"
```

Tis a lie that nothing broke between versions. I don't remember if we caught this and worked around it, or if we'd already changed the offending code and so didn't notice.

odow:
Oh gosh this is subtle:
```Julia
julia> dump(:.>=)
Symbol .>=

julia> dump(:(.>=))
Expr
  head: Symbol .
  args: Array{Any}((1,))
    1: Symbol >=

julia> VERSION
v"1.6.7"
```

At one point I probably just removed the parentheses.

odow:
My next problem is 

image

but I think this is a bug in MPB.

It increments the `numconstr` field when a quadratic constraint is added:
https://github.com/JuliaOpt/MathProgBase.jl/blob/094eddc51ee960e8a5f260a411f90053d2830d71/src/SolverInterface/nonlinear_to_lpqp.jl#L267-L269

but elsewhere, assumes that `numconstr` refers only to the number of linear constraints:
https://github.com/JuliaOpt/MathProgBase.jl/blob/094eddc51ee960e8a5f260a411f90053d2830d71/src/SolverInterface/nonlinear_to_lpqp.jl#L81-L83

Weirdly, this fails even on Julia 1.0. So when was the last time we actually tested this stuff? With what packages?

odow:
Looks like this was due to one of the last ever PRs to MPB: https://github.com/JuliaOpt/MathProgBase.jl/pull/229

It was committed on October 27, 2018, while we were still working on v0.18. Last commit to v0.18.6 was May 25, 2019. But we didn't tag the release of MPB until February 16, 2020. So it slipped through silently.

So, if we use MathProgBase@0.7.7 instead of MathProgBase@0.7.8, then that test passes. But now there are other failures with SCS...

This is turning into: things will only run if you use the versions that were available on or about May 2019.

mlubin:
Why are we spending time on this?

odow:
> Why are we spending time on this?

I wanted to understand the root cause of why the tests failed moving from Julia 1.0 to 1.6. And why the tests failed just on 1.0, when supposedly we had passed tests at one point in time.

But I don't think we need to release a new version of v0.18.7. We can now just declare that JuMP 0.18 works only on Julia 1.0, and that you'll need to install older versions of packages to get it working, and that there is no support. We can also point people to this PR if needed.

blegat:
If you got it working, it would be helpful to release. I'm sure people still have legacy code on JuMP v0.18 and having it work on the latest Julia is helpful for upgrading it to JuMP v1

odow:
I haven't heard any support requests for 0.18, other than people finding and trying to use JuMPeR. No one has mentioned that the tests fail on 1.0 either...

odow:
Talked to @blegat offline. We won't be tagging a new release.

blegat:
He convinced me with the argument that solvers won't work anyway since the MPB support has been dropped a while ago so getting binaries will be complicated

PR #3440, opened 2023-07-30
Add ci for release v0.18

blegat:
Same as for master but I added v1.0. We cannot drop it from v0.18 since you cannot drop a Julia version without doing a breaking release.


blegat:
If we merge this and rebase https://github.com/jump-dev/JuMP.jl/pull/3439, ci might run

mlubin (approved):

PR #3441, opened 2023-08-03
WIP: Add DataFrames as a weak dependency

### LINKED ISSUE #3438: Add DataFrames extension
To make this syntax a bit nicer

image

odow:
Closes #3438 

I started working on this, but I don't know whether I like it or not. 

I think I prefer the approach of explicitly constructing a data frame, and then adding a new column which is a vector of variables. It's more explicit and gives the user control over what the call the new column. This current approach would likely also be used naively to construct sparse sets which doesn't fix the loop problem.

### DIFF
--- a/ext/test_DataFrames.jl
+++ b/ext/test_DataFrames.jl
@@ -1,0 +1,21 @@
+#  Copyright 2017, Iain Dunning, Joey Huchette, Miles Lubin, and contributors
+#  This Source Code Form is subject to the terms of the Mozilla Public
+#  License, v. 2.0. If a copy of the MPL was not distributed with this
+#  file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+module TestContainersDataFrames
+
+using Test
+
+using DataFrames
+using JuMP
+
+function test_dimension_data_variable()
+    model = Model()
+    @variable(model, x[i = 2:4, j = 1:2], container = DataFrame)
+    @variable(model, y[i = 2:4, j = 1:2; isodd(i+j)], container = DataFrame)
+
+    return
+end
+
+end
--- a/Project.toml
+++ b/Project.toml
@@ -13,12 +13,15 @@ SnoopPrecompile = "66db9d55-30c0-4569-8b51-7e840670fc0c"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [weakdeps]
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 DimensionalData = "0703355e-b756-11e9-17c0-8b28908087d0"
 
 [extensions]
+JuMPDataFramesExt = "DataFrames"
 JuMPDimensionalDataExt = "DimensionalData"
 
 [compat]
+DataFrames = "1"
 DimensionalData = "0.24"
 MathOptInterface = "1.18"
 MutableArithmetics = "1"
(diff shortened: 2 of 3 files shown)


ext/test_DataFrames.jl
```diff
@@ -0,0 +1,21 @@
...
+
+using Test
+
+using DataFrames
+using JuMP
+
+function test_dimension_data_variable()
+    model = Model()
+    @variable(model, x[i = 2:4, j = 1:2], container = DataFrame)
+    @variable(model, y[i = 2:4, j = 1:2; isodd(i+j)], container = DataFrame)
```
odow, line 16: The more I think about this, the less I like it. I prefer the explicit "add a new column to an existing dataframe" than the "construct a dataframe from these indices." The latter approach is just going to encourage the same nested loops as the GAMS example.

blegat:
Ideally, we should parse the condition of the `JuMP.Containers.NestedIterator` then it would be an expression tree of conditions with `&&` and `||`. Then we put it into *conjunctive normal form* `c1 && c2 && ... && cm`. Then for each condition, we should check:

1. which indices does it involve ?
2. is this an equality `lhs == rhs`

If it involves a subset of the indices then we should filter with this condition earlier and not in the most nested level of the for loops.
If it is an equality, we could do the indexing ourself (as in https://discourse.julialang.org/t/is-it-possible-to-incorporate-the-relational-algebra-technique-into-jump-in-the-future-to-expedite-model-generation/101343/4?u=blegat) or use DataFrames but that's kind of an heavy dependency for just this.
Now what's a bit tricky is that you could have `[i = I, j = J, k = K, l = L; f(i, j, k) == g(j, k, l)]`. Then, it means you need to build the set of `(i, j, k)` and the set of `(j, k, l)` and then index them with the Dict to take the intersection. One could wonder if it wouldn't be better to just to the 4-nested for loops instead of having to do twice a 3-nested for loops. I think however than unless `K` has only 2 elements, it is always better to do two 3-nested for-loops than a 4-nested for loops so doing this intersection with dictionaries is always better.

In some cases, it might be better to do the nested for loops in an order that is different from the order of the indices used by the user but that can be left as future work since I don't think this is easy to find that. But detecting the two things I mentioned above and improving the iteration shouldn't be too complicated and it would solve the issue in the GAMS example without the user to do anything. The `SparseAxisArray` would automatically build itself efficiently.

odow:
> detecting the two things I mentioned above and improving the iteration shouldn't be too complicated

This is getting a bit too magical. It also assumes that constructing the index sets and a dictionary solves the performance problem.

I don't think we need to "fix" the GAMS example. We need to encourage people to rethink how they view models. Nested for-loops are not the best way to conceptualize the IJKLM model.

blegat:
> This is getting a bit too magical. It also assumes that constructing the index sets and a dictionary solves the performance problem.

It makes the complexity depends linearly on the number of tuples satisfying the condition. It might be surprising to the user that we build the `SparseAxisArray` so naively while he gave us all the information in a macro that are necessary to take into account its sparsity.

Following this PR, how would the user solve the IJKLM issue ?

odow:
> Following this PR, how would the user solve the IJKLM issue ?

They wouldn't. That's why I said I don't really like this PR.

odow:
Closing for now. But I'll leave https://github.com/jump-dev/JuMP.jl/issues/3438 open, and I'll add @blegat's comment to the issue. He has some CS things in mind.

Issue #3442, opened 2023-08-07
Doctests aren't failing


odow:
I think my change here https://github.com/jump-dev/JuMP.jl/pull/3386 means that the doctests are not running (or failing correctly).

As one example:

https://github.com/jump-dev/JuMP.jl/blob/a325eb638d9470204edb2ef548e93e59af56cc19/src/sets.jl#L323-L333

But running in the REPL gives:

```Julia
julia> model = Model();

julia> @variable(model, x in Semiinteger(3, 5))
x

julia> print(model)
Feasibility
Subject to
 x ∈ MathOptInterface.Semiinteger{Int64}(3, 5)
```

PR #3443, opened 2023-08-07
Add syntax for Parameter

odow:
<s>Still a WIP. Got distracted by #3442</s>

Part of #1185

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -944,6 +944,13 @@ function model_convert(model::AbstractModel, con::VectorConstraint)
     )
 end
 
+function model_convert(model::AbstractModel, x::VariableConstrainedOnCreation)
+    return VariableConstrainedOnCreation(
+        x.scalar_variable,
+        model_convert(model, x.set),
+    )
+end
+
 # TODO: update 3-argument @constraint macro to pass through names like @variable
 
 """
@@ -2737,12 +2744,9 @@ macro variable(args...)
                 )
             end
         end
-        buildcall = :(build_variable(
-            $_error,
-            $scalar_variables,
-            model_convert($model, $set),
-        ))
+        buildcall = :(build_variable($_error, $scalar_variables, $set))
     end
+    buildcall = :(model_convert($model, $buildcall))
     new_name_code = if isempty(set_string_name_kw_args)
         Expr(:if, :(set_string_names_on_creation($model)), name_code, "")
     else
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -2737,7 +2737,11 @@ macro variable(args...)
                 )
             end
         end
-        buildcall = :(build_variable($_error, $scalar_variables, $set))
+        buildcall = :(build_variable(
+            $_error,
+            $scalar_variables,
+            model_convert($model, $set),
+        ))
     end
     new_name_code = if isempty(set_string_name_kw_args)
         Expr(:if, :(set_string_names_on_creation($model)), name_code, "")
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -344,3 +344,30 @@ end
 function moi_set(set::Semiinteger{T}) where {T}
     return MOI.Semiinteger{T}(set.lower, set.upper)
 end
+
+"""
+    Parameter(value)
+
+A short-cut for the [`MOI.Parameter`](@ref) set.
+
+## Example
+
+```jldoctest
+julia> model = Model();
+
+julia> @variable(model, x in Parameter(2.0))
+x
+
+julia> print(model)
+Feasibility
+Subject to
+ x ∈ MathOptInterface.Parameter{Float64}(2.0)
+```
+"""
+struct Parameter{T} <: AbstractScalarSet
+    value::T
+end
+
+function moi_set(set::Parameter{T}) where {T}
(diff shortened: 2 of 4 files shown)


src/sets.jl
```diff
@@ -344,3 +344,30 @@ end
...
+    Parameter(value)
+
+A short-cut for the [`MOI.Parameter`](@ref) set.
+
+## Example
+
+```jldoctest
+julia> model = Model();
+
+julia> @variable(model, x in Parameter(2.0))
```
blegat, line 358: The reader might think it's needed to wrote `2.0` while actually `2` is sufficient since we convert automatically as he will see in the `print` below

blegat, line 358:
```suggestion
julia> @variable(model, x in Parameter(2))
```

joaquimg:
What about a macro:
`@parameter(model, x == 2)`

It would be more natural for beginners to create a parameter instead of a variable constrained in the parameter set.

src/variables.jl
```diff
@@ -1091,6 +1091,55 @@ function BinaryRef(v::GenericVariableRef)
...
+    set_parameter(x::GenericVariableRef, value)
+
+Update the parameter constraint on the variable `x` to `value`.
+
+See also [`ParameterRef`](@ref), [`is_parameter`](@ref).
+"""
+function set_parameter(x::GenericVariableRef, value)
+    model = owner_model(x)
+    model.is_model_dirty = true
+    set = MOI.Parameter(value)
```
joaquimg, line 1138: An error message might be needed above this line...

blegat:
> It would be more natural for beginners to create a parameter instead of a variable constrained in the parameter set.

We already have `@variable(model, x in ComplexPlane())` and no `@complex(model, x)`.  Using the `@variable` macro might be more consistent with the rest.

test/test_variable.jl
```diff
@@ -1471,4 +1471,20 @@ function test_variable_dimension_mismatch()
     return
 end
 
+function test_parameter()
+    model = Model()
+    @variable(model, x in Parameter(2))
+    p = ParameterRef(x)
+    @test p isa ConstraintRef
+    F, S = MOI.VariableIndex,MOI.Parameter{Float64}
```
odow, line 1479:
```suggestion
    F, S = MOI.VariableIndex, MOI.Parameter{Float64}
```

src/variables.jl
```diff
@@ -1091,6 +1091,73 @@ function BinaryRef(v::GenericVariableRef)
...
+    model.is_model_dirty = true
+    set = MOI.Parameter{T}(convert(T, value))
+    MOI.set(backend(model), MOI.ConstraintSet(), _parameter_index(x), set)
+    return
+end
+
+"""
+    parameter_value(x::GenericVariableRef, value)
+
+Update the parameter constraint on the variable `x` to `value`.
```
odow, line 1146:
```suggestion
Return the value of the parameter `x`.
```

src/variables.jl
```diff
@@ -1091,6 +1091,73 @@ function BinaryRef(v::GenericVariableRef)
...
+    T = value_type(typeof(x))
+    model = owner_model(x)
+    model.is_model_dirty = true
+    set = MOI.Parameter{T}(convert(T, value))
+    MOI.set(backend(model), MOI.ConstraintSet(), _parameter_index(x), set)
+    return
+end
+
+"""
+    parameter_value(x::GenericVariableRef, value)
```
odow, line 1144:
```suggestion
    parameter_value(x::GenericVariableRef)
```

src/variables.jl
```diff
@@ -1288,7 +1355,7 @@ function add_variable(
     var_index = _moi_add_constrained_variable(
         backend(model),
         variable.scalar_variable,
-        variable.set,
+        model_convert(model, variable.set),
```
odow, line 1358: @blegat I had to add a couple more `model_convert`

src/variables.jl
```diff
@@ -1391,7 +1458,7 @@ function add_variable(
     var_indices = _moi_add_constrained_variables(
         backend(model),
         variable.scalar_variables,
-        variable.set,
+        model_convert(model, variable.set),
```
blegat, line 1461: This is inconsistent with the constraint case where the `model_convert` is called by the macro and not inside `add_constraint`.
So it should be applied to
https://github.com/jump-dev/JuMP.jl/blob/8c205287191a345eb85db096deb845114a9ee38f/src/macros.jl#L2740
like in
https://github.com/jump-dev/JuMP.jl/blob/8c205287191a345eb85db096deb845114a9ee38f/src/macros.jl#L1054-L1058

odow: Done

src/macros.jl
```diff
@@ -2737,7 +2737,11 @@ macro variable(args...)
                 )
             end
         end
-        buildcall = :(build_variable($_error, $scalar_variables, $set))
+        buildcall = :(build_variable(
```
blegat, line 2740: For the constraints, we call `model_convert` on the output of `build_variable`, would is it different here ?

odow: Updated

odow:
Any other comments?

mlubin (approved):

PR #3447, opened 2023-08-08
Fix printing of models with large integers

LebedevRI:
Consider:
```
using JuMP, Ipopt
opt_model = Model(Ipopt.Optimizer, add_bridges = false)
@variable(opt_model, p)
@constraint(opt_model, p * 1e20 == 42)
print(opt_model)
```
```
InexactError: trunc(Int64, 1.0e20)

Stacktrace:
  [1] trunc
    @ ./float.jl:893 [inlined]
  [2] round
    @ ./float.jl:384 [inlined]
  [3] _string_round
    @ ~/.julia/packages/JuMP/jZvaU/src/print.jl:75 [inlined]
```

This did come up in an non-artificial situation.

In general, FP->Int cast, where the source FP value is larger than the largest integer of the target type is UB, and julia rightfully notices that.

Tests seem to suggest that edge cases at least don't cause UB.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -72,8 +72,8 @@ end
 # e.g.   5.3  =>  5.3
 #        1.0  =>  1
 function _string_round(x::Union{Float32,Float64})
-    if isinteger(x) && typemin(Int) <= x <= typemax(Int)
-        return string(round(Int, x))
+    if isinteger(x) && typemin(Int64) <= x <= typemax(Int64)
+        return string(round(Int64, x))
     end
     return string(x)
 end
--- a/src/print.jl
+++ b/src/print.jl
@@ -72,7 +72,10 @@ end
 # e.g.   5.3  =>  5.3
 #        1.0  =>  1
 function _string_round(x::Union{Float32,Float64})
-    return isinteger(x) ? string(round(Int, x)) : string(x)
+    if isinteger(x) && typemin(Int) <= x <= typemax(Int)
+        return string(round(Int, x))
+    end
+    return string(x)
 end
 
 _string_round(::typeof(abs), x::Real) = _string_round(abs(x))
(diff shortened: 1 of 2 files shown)


odow:
Thanks! I've tweaked the implementation slightly. 

Something to be aware of: you should never use large values like `1e20` in a JuMP model. Most commonly, it will just cause numerical issues, but it could also lead to wildly incorrect results.

In this example, the solution is `p = 4.2e-19`, but Ipopt has a tolerance of ~1e-6, so it might give you a solution of `p = 1e-6`, which is many orders of magnitude away from optimal.

LebedevRI:
> Thanks! I've tweaked the implementation slightly.

Thanks!
It seems i have missed the formatting nit, let me get that.

> Something to be aware of: you should never use large values like `1e20` in a JuMP model. Most commonly, it will just cause numerical issues, but it could also lead to wildly incorrect results.
> 
> In this example, the solution is `p = 4.2e-19`, but Ipopt has a tolerance of ~1e-6, so it might give you a solution of `p = 1e-6`, which is many orders of magnitude away from optimal.

Right, sure, i didn't expect that to work.
In that particular case it occurred in something like
`@constraint(model, fit(10kHz) == fit(100kHz))`, just as an experiment.

odow (approved):

src/print.jl
```diff
@@ -72,7 +72,10 @@ end
 # e.g.   5.3  =>  5.3
 #        1.0  =>  1
 function _string_round(x::Union{Float32,Float64})
-    return isinteger(x) ? string(round(Int, x)) : string(x)
+    if isinteger(x) && typemin(Int) <= x <= typemax(Int)
```
odow, line 75:
```suggestion
    if isinteger(x) && typemin(Int64) <= x <= typemax(Int64)
```

src/print.jl
```diff
@@ -72,7 +72,10 @@ end
...
+        return string(round(Int, x))
```
odow, line 76:
```suggestion
        return string(round(Int64, x))
```

odow:
Thanks!

LebedevRI:
Thank you!

PR #3451, opened 2023-08-15
Allow array in nonlinear expressions

blegat:
This PR is a proof of concept to show the possibility of storing arrays in the JuMP nonlinear expression and MOI nonlinear function.
Of course, it won't work with the AD but it can be supported by a solver or by a Disciplined Convex Programming transformation.
```julia
julia> model = Model();

julia> @variable(model, X[1:3, 1:3], Symmetric)
3×3 LinearAlgebra.Symmetric{VariableRef, Matrix{VariableRef}}:
 X[1,1]  X[1,2]  X[1,3]
 X[1,2]  X[2,2]  X[2,3]
 X[1,3]  X[2,3]  X[3,3]

julia> using LinearAlgebra

julia> @constraint(model, log(det(X)) >= 0)
log(det(VariableRef[X[1,1] X[1,2] X[1,3]; X[1,2] X[2,2] X[2,3]; X[1,3] X[2,3] X[3,3]])) - 0.0 ≥ 0
```

### DIFF
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -575,6 +575,8 @@ function moi_function(f::GenericNonlinearExpr{V}) where {V}
     for i in length(f.args):-1:1
         if f.args[i] isa GenericNonlinearExpr{V}
             push!(stack, (ret, i, f.args[i]))
+        elseif f.args[i] isa AbstractArray
+            ret.args[i] = moi_function.(f.args[i])
         else
             ret.args[i] = moi_function(f.args[i])
         end
@@ -586,6 +588,8 @@ function moi_function(f::GenericNonlinearExpr{V}) where {V}
         for j in length(arg.args):-1:1
             if arg.args[j] isa GenericNonlinearExpr{V}
                 push!(stack, (child, j, arg.args[j]))
+            elseif arg.args[j] isa AbstractArray
+                child.args[j] = moi_function.(arg.args[j])
             else
                 child.args[j] = moi_function(arg.args[j])
             end
@@ -611,6 +615,8 @@ function jump_function(model::GenericModel, f::MOI.ScalarNonlinearFunction)
             for child in reverse(arg.args)
                 push!(stack, (new_ret, child))
             end
+        elseif arg isa AbstractArray
+            push!(parent.args, jump_function.(model, arg))
         else
             push!(parent.args, jump_function(model, arg))
         end
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -330,14 +330,16 @@ Base.isreal(::GenericNonlinearExpr) = true
 
 # Univariate operators
 
-_is_real(::Any) = false
-_is_real(::Real) = true
-_is_real(::AbstractVariableRef) = true
-_is_real(::GenericAffExpr{<:Real}) = true
-_is_real(::GenericQuadExpr{<:Real}) = true
-_is_real(::GenericNonlinearExpr) = true
-_is_real(::NonlinearExpression) = true
-_is_real(::NonlinearParameter) = true
+_is_real(::Type) = false
+_is_real(::Type{<:Real}) = true
+_is_real(::Type{<:AbstractVariableRef}) = true
+_is_real(::Type{<:GenericAffExpr{<:Real}}) = true
+_is_real(::Type{<:GenericQuadExpr{<:Real}}) = true
+_is_real(::Type{<:GenericNonlinearExpr}) = true
+_is_real(::Type{<:NonlinearExpression}) = true
+_is_real(::Type{<:NonlinearParameter}) = true
+_is_real(::Type{<:AbstractArray{T}}) where {T} = _is_real(T)
+_is_real(x) = _is_real(typeof(x))
 
 function _throw_if_not_real(x)
     if !_is_real(x)
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -571,14 +571,19 @@ end
 
 moi_function(x::Number) = x
 
+# MOI.Nonlinear.ReverseAD does not support arrays but ArrayDiff.jl and
+# Convex.jl do.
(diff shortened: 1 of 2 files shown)


odow:
Some things need to be worked through:

 * How can solvers declare at the JuMP level what operations they support
 * How do we avoid type piracy without requiring JuMP to implement the full DCP atoms
 * How do we avoid overloading a method that could otherwise be traced

One thing to keep in mind is that we no longer parse `Symbol`. Every call is an overloaded Julia function. So we can't make up `:quad_over_lin` without making `function quad_over_lin(x, y) end`.

One option that @chriscoey and I discussed is that something like `det(X)` is not `LinearAlgebra.det(X)`, but `DCP.det(X)`, which creates `NonlinearExpr(:dcp_det, vec(X))` or `NonlinearExpr(:dcp_det, Any[X])`, but is very clearly not supported by the AD systems.

blegat:
> How can solvers declare at the JuMP level what operations they support

Why would they need to do that ? The function we overload are going to fail anyway so we might as well add them in the expression tree and we'll let the solver throw an error if it does not support what it is getting.

> How do we avoid type piracy without requiring JuMP to implement the full DCP atoms

DCP would be in MOI so we can do like:
https://github.com/jump-dev/JuMP.jl/pull/3106/files#diff-38284a9c301dbce5b78214aec3ec0172987aec8d7a85770c5a75c6f801fabee0R274

> How do we avoid overloading a method that could otherwise be traced

What do you mean by "traced" ?

> One thing to keep in mind is that we no longer parse `Symbol`. Every call is an overloaded Julia function. So we can't make up `:quad_over_lin` without making `function quad_over_lin(x, y) end`.

Yes, we would need to define functions. When they get numbers they return a number but if they get JuMP expressions, they return an `NonlinearExpr`.

> One option that @chriscoey and I discussed is that something like `det(X)` is not `LinearAlgebra.det(X)`, but `DCP.det(X)`, which creates `NonlinearExpr(:dcp_det, vec(X))` or `NonlinearExpr(:dcp_det, Any[X])`, but is very clearly not supported by the AD systems.

But then if someone calls `LinearAlgebra.det`, it won't work. Not sure I like that

odow:
I just worry that we're going to end in a situation where it's not obvious what functions can be used with which solver. I'd view DCP as a exploratory research project. It's not obvious that the first approach will be the right one.

blegat:
> I just worry that we're going to end in a situation where it's not obvious what functions can be used with which solver.

Still not sure what you mean. Could you give an example ? Is this a concern for a function that could be written in a more elementary form ? For example, `norm` would be rewritten into `sqrt(prod(...))` while DCP would prefer to keep it as a `norm` ?

odow:
> Why would they need to do that ?

How does the user find out what functions they can call for DCP?

> DCP would be in MOI 

I did not assume this. To begin with, it should be a JuMP extension or an MOI solver.

> What do you mean by "traced" ?

Evaluating the function returns the symbolic expression. If `f(x) = sin(x)^2`, then calling `f(x::VariableRef)` traces the function.

> But then if someone calls LinearAlgebra.det, it won't work. Not sure I like that

I think it's easier to keep all the DCP atoms in a strict namespace, rather than playing the game of overloading various methods from different places.

odow:
My takeaway is that this is a research question. Not a solved problem requiring an implementation.

odow:
> This can start as a separate repo if we want to start having something easier to play around than an pair of open PRs in JuMP and MOI.

Yes please.

src/nlp_expr.jl
```diff
@@ -506,6 +510,8 @@ function moi_function(f::GenericNonlinearExpr{V}) where {V}
     for i in length(f.args):-1:1
         if f.args[i] isa GenericNonlinearExpr{V}
             push!(stack, (ret, i, f.args[i]))
+        elseif arg.args[i] isa AbstractArray
+            child.args[i] = moi_function.(arg.args[i])
```
odow, line 514:
```suggestion
            ret.args[i] = moi_function.(arg.args[i])
```

odow (commented):
Should we instead just define
```julia
moi_function(x::AbstractArray) = moi_function.(x)
jump_function(x::AbstractArray) = jump_function.(x)
```
You could implement this as a pirated method in DCP.jl to begin with.

blegat:
I updated the PR. The scope is not to overload `norm`, `det`, etc... but to allow `NonlinearOperator` to work with array inputs. At the moment, it would just call the function and not build the nonlinear expression, even if one of the arguments is an array of JuMP expressions.
If this is used with AD then the AD will error because an array is in the expression graph which is the expected behavior.

odow:
I'm not in favor of merging this until we have DCP.jl mocked up end-to-end. I'd like to know what the alternatives are.

One option is for DCP.jl to use a separate type entirely to begin with. They don't need to (mis)use `NonlinearOperator`.

Supporting vectors in nonlinear expressions is not something we should rush into, even if it's something we eventually want to support.

The only thing that might need implementing in JuMP is `moi_function` and `jump_function`, but these could be done with pirated methods to begin with.

blegat:
This requires changing how the `NonlinearOperator` act on arrays. If we consider this a bug-fix then it's fine to wait. Otherwise, if we consider it as a breaking change then we should decide now.

odow:
Closing in favor of #3489

blegat:
Reopening because it is useful for https://github.com/blegat/ArrayDiff.jl/

blegat:
This is actually not needed, found a way to work around it in https://github.com/blegat/ArrayDiff.jl/pull/39

blegat:
I think we now have enough motivation to merge this.

- I need a JuMP extension full of type-piracy for this: https://github.com/NexOR-Optimization/Hexaly.jl/pull/10
- ArrayDiff is currently doing the same type-piracy
- Array syntax for Convex.jl is also because of this (this was the initial motivation)

odow (approved):

PR #3450, opened 2023-08-14
Add support for VectorNonlinearFunction

odow:
Tests won't pass because requires new tag of MOI.

But this is working with `PATHSolver#od/nlp-expr`

```julia
julia> using JuMP, PATHSolver

julia> model = Model(PATHSolver.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: EMPTY_OPTIMIZER
Solver name: Path 5.0.03

julia> @variable(model, 0 <= x[1:4] <= 10)
4-element Vector{VariableRef}:
 x[1]
 x[2]
 x[3]
 x[4]

julia> @constraints(model, begin
                              -x[3]^2 -  x[4]   + 2 ⟂ x[1]
                               x[3]^3 - 2x[4]^2 + 2 ⟂ x[2]
           x[1]^5 -  x[2]   + 2x[3]   - 2x[4]   - 2 ⟂ x[3]
           x[1]   + 2x[2]^3 - 2x[3]   + 4x[4]   - 6 ⟂ x[4]
       end)
([-x[3]² - x[4] + 2, x[1]] ∈ MathOptInterface.Complements(2), [((x[3] ^ 3) - (2 x[4]²)) + 2.0, +(x[2])] ∈ MathOptInterface.Complements(2), [((((x[1] ^ 5) - x[2]) + (2 x[3])) - (2 x[4])) - 2.0, +(x[3])] ∈ MathOptInterface.Complements(2), [(((x[1] + (2.0 * (x[2] ^ 3))) - (2 x[3])) + (4 x[4])) - 6.0, +(x[4])] ∈ MathOptInterface.Complements(2))

julia> optimize!(model)
Path 5.0.03 (Fri Jun 26 09:58:07 2020)
Written by Todd Munson, Steven Dirkse, Youngdae Kim, and Michael Ferris


Crash Log
major  func  diff  size  residual    step       prox   (label)
    0     0             9.3949e+00             0.0e+00 (f[    4])
    1     4     3     2 7.7898e+00  4.1e-01    0.0e+00 (f[    1])
    2     5     1     3 1.4635e+00  1.0e+00    0.0e+00 (f[    3])
    3     6     0     3 1.1831e+00  1.0e+00    0.0e+00 (f[    3])
pn_search terminated: no basis change.

Major Iteration Log
major minor  func  grad  residual    step  type prox    inorm  (label)
[...]

PR #3452, opened 2023-08-15
Add support for constant LHS in complements constraint

odow:
This came up when experimenting with #3450

### DIFF
--- a/test/test_complement.jl
+++ b/test/test_complement.jl
@@ -129,7 +129,7 @@ function test_vector_error_F_F()
     return
 end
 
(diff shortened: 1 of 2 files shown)


test/test_complement.jl
```diff
@@ -116,6 +129,20 @@ function test_vector_error_F_F()
     return
 end
 
+function test_vector_error_F_F()
```
odow, line 132:
```suggestion
function test_vector_error_0_F()
```

odow:
Merging as minor and non-controversial.

Issue #3469, opened 2023-08-31
StackOverflow with operate!! and NonlinearExpr


blegat:
This should be investigated, I wouldn't expect this to cause a StackOverflow
https://github.com/jump-dev/JuMP.jl/blob/a573de22a3c52160dcfc2e3bf06fe58b5243a110/src/mutable_arithmetics.jl#L289-L291

odow:
I think this is expected behavior. See https://github.com/jump-dev/JuMP.jl/pull/3106#discussion_r1306726303

```Julia
julia> import MutableArithmetics as MA

julia> struct Foo <: MA.AbstractMutable
           x::Int
       end

julia> Base.copy(f::Foo) = Foo(f.x)

julia> function MA.add_mul(lhs::Foo, x::Union{Int,Foo}, y::Union{Int,Foo})
           return MA.operate!!(MA.add_mul, copy(lhs), x, y)
       end

julia> lhs = Foo(2)
Foo(2)

julia> x = Foo(3)
Foo(3)

julia> MA.add_mul(lhs, x, 3)
ERROR: StackOverflowError:
Stacktrace:
     [1] operate_fallback!!(::MutableArithmetics.IsNotMutable, ::typeof(MutableArithmetics.add_mul), ::Foo, ::Foo, ::Int64)
       @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:581
     [2] operate!!(::typeof(MutableArithmetics.add_mul), ::Foo, ::Foo, ::Int64)
       @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/rewrite.jl:89
     [3] add_mul(lhs::Foo, x::Foo, y::Int64)
       @ Main ./REPL[26]:2
     [4] operate(op::typeof(MutableArithmetics.add_mul), x::Foo, y::Foo, args::Int64)
       @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:206
     [5] operate_fallback!!(::MutableArithmetics.IsNotMutable, ::typeof(MutableArithmetics.add_mul), ::Foo, ::Foo, ::Int64)
       @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:586
--- the last 4 lines are repeated 19994 more times ---
 [79982] operate!!(::typeof(MutableArithmetics.add_mul), ::Foo, ::Foo, ::Int64)
[...]

PR #3470, opened 2023-08-31
Improve performance of sum of products with nonlinear expressions

blegat:
Closes https://github.com/jump-dev/JuMP.jl/issues/3469

### DIFF
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -368,16 +368,20 @@ for f in (:+, :-, :*, :^, :/, :atan)
 end
 
 function _MA.operate!!(
-    ::typeof(_MA.add_mul),
+    op::_MA.AddSubMul,
     x::GenericNonlinearExpr,
-    y::AbstractJuMPScalar,
-)
+    args::Vararg{Any,N},
+) where {N}
     _throw_if_not_real(x)
     if x.head == :+
-        push!(x.args, y)
+        arg = *(args...)
+        if _MA.add_sub_op(op) != +
+            arg = _MA.add_sub_op(op)(arg)
+        end
+        push!(x.args, arg)
         return x
     end
-    return +(x, y)
+    return _MA.add_sub_op(op)(x, *(args...))
 end
 
 """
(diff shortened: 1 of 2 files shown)


blegat:
Looking into an example where it gives an improvement to implement `operate!!` for more than one factor, I found this weird issue:
https://github.com/jump-dev/MutableArithmetics.jl/issues/224

src/nlp_expr.jl
```diff
@@ -368,16 +368,20 @@ for f in (:+, :-, :*, :^, :/, :atan)
...
-    y::AbstractJuMPScalar,
-)
+    args::Vararg{Any,N},
+) where {N}
     _throw_if_not_real(x)
     if x.head == :+
-        push!(x.args, y)
+        arg = *(args...)
+        if _MA.add_sub_op(op) != +
+            arg = _MA.add_sub_op(op)(arg)
```
odow, line 379: I really think we should avoid this. If the user writes `x + y - 2 * z`, we should make `-(+(x, y), *(2, z))`, not `+(x, y, -(*(2, z))). 

Julia has a parsing precedence order. Let's follow that where possible.

odow:
I also don't see how this fixes your issue. It just defines a separate method so that the other issue isn't hit. We'd still need the `if` statement if we added another `AbstractJuMPScalar` that was `IsNotMutable`.

src/nlp_expr.jl
```diff
@@ -368,16 +368,20 @@ for f in (:+, :-, :*, :^, :/, :atan)
 end
 
 function _MA.operate!!(
-    ::typeof(_MA.add_mul),
+    op::_MA.AddSubMul,
     x::GenericNonlinearExpr,
-    y::AbstractJuMPScalar,
```
odow, line 373: I don't remember why I only wrote the binary case 😆

src/nlp_expr.jl
blegat: I thought `flatten!` was doing that anyway

PR #3472, opened 2023-08-31
[docs] distinguish Nonlinear Modeling sections in documentation

### LINKED ISSUE #3471: Nonlinear interfaces are not well distinguished in the docs
In the manual it is difficult to distinguish which tab leads to which interface since they are both "Nonlinear Modeling":
tempsnip

Also, the new interface comes last, so users will likely see the old API first.

odow:
Closes #3471 

http://jump.dev/JuMP.jl/previews/PR3472

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -343,10 +343,10 @@ const _PAGES = [
         "manual/objective.md",
         "manual/containers.md",
         "manual/solutions.md",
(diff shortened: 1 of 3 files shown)


odow:
@pulsipher is this better?

image

pulsipher:
Yep, that's much more clear.

PR #3474, opened 2023-08-31
Fix integer powers to return appropriate type instead of QuadExpr

### LINKED ISSUE #3454: Consider changing x^1 to return x
We currently have these methods:
https://github.com/jump-dev/JuMP.jl/blob/8a856bb677fdc1ec1e9a40a8a41e64089f7b6437/src/operators.jl#L201-L232

https://github.com/jump-dev/JuMP.jl/pull/3106 will make them somewhat more general by supporting `3`+ exponents. In doing so, it makes the return type unstable.

Once merged, we should consider changing `x^0` to return `one(T)` and `x^1` to return `x`.

See also https://github.com/jump-dev/JuMP.jl/pull/3106/files#r1295483658

odow:
Closes #3454 

This is technically breaking, since it returns a different type from before, but I'd also argue that it is a bug fix, because `x^0` and `x^1` should not return a quadratic expression.

Needs:

 - [x] https://github.com/jump-dev/JuMP.jl/actions/runs/6042779366

### DIFF
--- a/test/test_mutable_arithmetics.jl
+++ b/test/test_mutable_arithmetics.jl
@@ -117,11 +117,8 @@ function test_extension_quadratic(
     VariableRefType = VariableRef,
 )
     model = ModelType()
-    @variable(model, w)
-    @variable(model, x)
-    @variable(model, y)
-    @variable(model, z)
-    JuMP._MA.Test.quadratic_test(w, x, y, z)
+    @variable(model, x[1:4])
+    JuMP._MA.Test.quadratic_test(x...; exclude = ["quadratic_add_canonical"])
     return
 end
 
--- a/test/test_mutable_arithmetics.jl
+++ b/test/test_mutable_arithmetics.jl
@@ -118,6 +118,7 @@ function test_extension_quadratic(
 )
     model = ModelType()
     @variable(model, x[1:4])
+    # Test is excluded because of https://github.com/jump-dev/MutableArithmetics.jl/issues/227
     JuMP._MA.Test.quadratic_test(x...; exclude = ["quadratic_add_canonical"])
     return
 end
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -197,24 +197,24 @@ end
 
 function Base.:^(lhs::AbstractVariableRef, rhs::Integer)
     T = value_type(typeof(lhs))
-    if rhs == 2
-        return lhs * lhs
+    if rhs == 0
+        return one(T)
     elseif rhs == 1
-        return convert(GenericQuadExpr{T,variable_ref_type(lhs)}, lhs)
-    elseif rhs == 0
-        return one(GenericQuadExpr{T,variable_ref_type(lhs)})
+        return lhs
+    elseif rhs == 2
+        return lhs * lhs
     else
         return GenericNonlinearExpr(:^, Any[lhs, rhs])
     end
 end
 
 function Base.:^(lhs::GenericAffExpr{T}, rhs::Integer) where {T}
-    if rhs == 2
-        return lhs * lhs
+    if rhs == 0
+        return one(T)
     elseif rhs == 1
-        return convert(GenericQuadExpr{T,variable_ref_type(lhs)}, lhs)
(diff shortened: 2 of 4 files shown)


test/test_mutable_arithmetics.jl
```diff
@@ -117,11 +117,8 @@ function test_extension_quadratic(
     VariableRefType = VariableRef,
 )
     model = ModelType()
-    @variable(model, w)
-    @variable(model, x)
-    @variable(model, y)
-    @variable(model, z)
-    JuMP._MA.Test.quadratic_test(w, x, y, z)
+    @variable(model, x[1:4])
+    JuMP._MA.Test.quadratic_test(x...; exclude = ["quadratic_add_canonical"])
```
odow, line 121: MA assumes that `^` has the same return type as `x * x`.

odow:
There are quite a few failures, but it is't obvious if it's because of this PR.

Here's the latest `master`: https://github.com/jump-dev/JuMP.jl/actions/runs/6042942250

test/test_mutable_arithmetics.jl
blegat: This should be fixed at some point, can you open an issue and drop a comment here referencing this issue ?

pulsipher:
> There are quite a few failures, but it is't obvious if it's because of this PR.
> 
> Here's the latest `master`: https://github.com/jump-dev/JuMP.jl/actions/runs/6042942250

To be clear, this test is the current `master` without this PR, right? The InfiniteOpt failures are just because a few expressions a slightly different now with the use of `move_factors_into_sums` and all this is fixed with https://github.com/infiniteopt/InfiniteOpt.jl/pull/308. 

For the UnitJuMP and SumOfSquares, it appears the new operator overloading leads to some minor bugs based on assumptions made with the old API.

odow:
I think this PR is safe to merge.

 * PolyJuMP and SumOfSquares are https://github.com/JuliaAlgebra/MultivariatePolynomials.jl/pull/273
 * UnitJuMP is https://github.com/trulsf/UnitJuMP.jl/issues/19

test/test_mutable_arithmetics.jl
odow, line 121:
```suggestion
    # Test is excluded because of https://github.com/jump-dev/MutableArithmetics.jl/issues/227
    JuMP._MA.Test.quadratic_test(x...; exclude = ["quadratic_add_canonical"])
```

mlubin:
What about the performance impact of no longer being type stable? Should we add a note on the performance tips page?

odow:
It's already not type stable because of the case when `n > 2`. I don't think this needs an explicit callout. I find it hard to believe a bottleneck in someone's code is going to be writing `x^2` instead of `x * x`.

mlubin:
> It's already not type stable because of the case when n > 2.

But it is type stable in JuMP 1.14, right?

odow:
> But it is type stable in JuMP 1.14, right?

Yes. We needed to give up type stability, or else something like `x^3` wouldn't work and you'd need to write `x^3.0`.

For a real-world motivation for this change, see:
https://github.com/lanl-ansi/PowerModels.jl/pull/858/files#diff-7a544f21e32c63ffcff0a5cc9ba4bbf8218aca07ed057b21a238980bbdf4d5fb
I think getting the objective type "correct" is more important than type stability.

x-ref https://github.com/jump-dev/JuMP.jl/pull/3106#discussion_r1179843223

mlubin:
> I think getting the objective type "correct" is more important than type stability.

I agree with the code change. I do expect at least one person to be surprised either about the change in return types or the lack of type stability. It at least warrants an explicit statement in the release notes.

odow:
> It at least warrants an explicit statement in the release notes.

Yip. I think the release notes will need to be quite detailed anyway, because it's quite a big change. Let me start a separate PR for that and we can iterate.

odow:
I've made a note in https://github.com/jump-dev/JuMP.jl/pull/3485 to mention in the changelog.

Issue #3482, opened 2023-09-03
The Vector{AbstractJuMPScalar} problem


odow:
A variety of operations can construct `Vector{AbstractJuMPScalar}`. For example:
```julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, x)
x

julia> typeof([x, sin(x)])
Vector{AbstractJuMPScalar} (alias for Array{AbstractJuMPScalar, 1})
```
It also happens in UnitJuMP when the objectives have different unit types.

JuMP chooses to intercept these as `VectorNonlinearFunction`:
https://github.com/jump-dev/JuMP.jl/blob/2e0b076962b92034077679ed9b58b41a8cc05827/src/nlp_expr.jl#L1100-L1113

unfortunately, so too does UnitJuMP:
https://github.com/trulsf/UnitJuMP.jl/blob/c55366ff1c4f090105aa50ec7d6285368b5c7bed/src/units.jl#L178-L181

The UnitJuMP case it type piracy, but we should probably double check that the current behavior in JuMP is what we want to happen.

cc @trulsf

PR #3483, opened 2023-09-03
Fix promotion rules for GenericNonlinearExpr

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3482

https://github.com/jump-dev/JuMP.jl/actions/runs/6067289679

### DIFF
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -1119,7 +1119,7 @@ function moi_function(f::AbstractVector{<:GenericNonlinearExpr})
     return MOI.VectorNonlinearFunction(f)
 end
 
-function MOI.VectorNonlinearFunction(f::AbstractVector{<:GenericNonlinearExpr})
+function MOI.VectorNonlinearFunction(f::Vector{<:AbstractJuMPScalar})
     return MOI.VectorNonlinearFunction(map(moi_function, f))
 end
 
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -673,11 +673,25 @@ function _evaluate_expr(
     end
 end
 
-# MutableArithmetics.jl
+# MutableArithmetics.jl and promotion
 
-# These converts are used in the {add,sub}mul definition for AbstractJuMPScalar.
+function Base.promote_rule(
+    ::Type{GenericNonlinearExpr{V}},
+    ::Type{V},
+) where {V<:AbstractVariableRef}
+    return GenericNonlinearExpr{V}
+end
 
-Base.convert(::Type{<:GenericNonlinearExpr}, x::AbstractVariableRef) = x
+function Base.promote_rule(
+    ::Type{GenericNonlinearExpr{V}},
+    ::Type{<:Union{GenericAffExpr{C,V},GenericQuadExpr{C,V}}},
+) where {C,V<:AbstractVariableRef}
+    return GenericNonlinearExpr{V}
+end
+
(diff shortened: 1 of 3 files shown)


odow:
This fixes the UnitJuMP test failure, although UnitJuMP still commits type piracy.

mlubin (approved):

odow:
It's still possible to generate `Vector{AbstractJuMPScalar}`:
```Julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, x);

julia> y = [i == 1 ? x : sin(x) for i in 1:2]
2-element Vector{AbstractJuMPScalar}:
 x
 sin(x)

julia> @objective(model, Min, [i == 1 ? x : sin(x) for i in 1:2])
ERROR: MethodError: no method matching moi_function(::Vector{AbstractJuMPScalar})
Closest candidates are:
  moi_function(::AbstractConstraint) at /Users/oscar/.julia/dev/JuMP/src/constraints.jl:550
  moi_function(::AbstractVariableRef) at /Users/oscar/.julia/dev/JuMP/src/variables.jl:575
  moi_function(::Vector{var"#s65"} where var"#s65"<:AbstractVariableRef) at /Users/oscar/.julia/dev/JuMP/src/variables.jl:584
  ...
Stacktrace:
 [1] set_objective_function(model::Model, func::Vector{AbstractJuMPScalar})
   @ JuMP ~/.julia/dev/JuMP/src/objective.jl:146
 [2] set_objective(model::Model, sense::MathOptInterface.OptimizationSense, func::Vector{AbstractJuMPScalar})
   @ JuMP ~/.julia/dev/JuMP/src/objective.jl:175
 [3] macro expansion
   @ ~/.julia/dev/JuMP/src/macros.jl:1850 [inlined]
 [4] top-level scope
   @ REPL[5]:1

julia> @objective(model, Min, y)
ERROR: MethodError: no method matching moi_function(::Vector{AbstractJuMPScalar})
```

So perhaps we need a better error for the fallback.

odow:
Thoughts @pulsipher?

odow:
@trulsf fixed UnitJuMP, but I think this is the right thing regardless. We can always add back support for `Vector{AbstractJuMPScalar}` later.

src/nlp_expr.jl
```diff
@@ -1097,18 +1111,15 @@ function jump_function(
...
     return MOI.VectorNonlinearFunction
 end
 
-function moi_function(f::Vector{<:AbstractJuMPScalar})
+function moi_function(f::AbstractVector{<:GenericNonlinearExpr})
     return MOI.VectorNonlinearFunction(f)
 end
 
-function MOI.VectorNonlinearFunction(f::Vector{<:AbstractJuMPScalar})
+function MOI.VectorNonlinearFunction(f::AbstractVector{<:GenericNonlinearExpr})
```
blegat, line 1122: I don't agree. Here, you are explicit about the return type so it should work with a vector of anything

blegat: But it can be added later in a non breaking way so I won't dispute this

blegat (approved):

src/nlp_expr.jl
odow: Ah true. This is reasonable.

odow, line 1122:
```suggestion
function MOI.VectorNonlinearFunction(f::Vector{<:AbstractJuMPScalar})
```

PR #3500, opened 2023-09-11
Use let model=model in variable macro to improve type stability

### LINKED ISSUE #3499: `@variable` declarations returning empty container should have `VariableRef` element type
On latest tagged, `@variable(m, x[1:0])` returns a `Vector{Any}`.  Because of this, code which uses `x` as an input cannot infer the containers element type.  An example of how this can go wrong is if one calls `sum(x)` it will throw an error because there is no `zero(Any)`, where as `zero(VariableRef)` gives a zero `AffExpr`.

The simplest solution would be to check for empty returns and convert them to `Array{VariableRef}`.  However, I realize that this ignores the possibility of other variable types, so, alternatively, it might be nice to add a keyword argument.  Preferably, this would still default to something other than `Any`.

You can `convert` the result, but it seems undesirable to have to do the check.

**Additional context**
I think previous versions of JuMP returned a `VariableRef` container, but I haven't verified this.

odow:
Closes #3499

We probably need something similar in the other macros as well.

This is quite hard to test, because wrapping things in a function fixes the problem...

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1334,7 +1334,13 @@ function _constraint_macro(
 
     creation_code =
         Containers.container_code(idxvars, indices, code, requested_container)
-
+    # Wrap the entire code block in a let statement to make the model act as
+    # a type stable local variable.
+    creation_code = quote
+        let $model = $model
+            $creation_code
+        end
+    end
     if anonvar
         # Anonymous constraint, no need to register it in the model-level
         # dictionary nor to assign it to a variable in the user scope.
@@ -1939,6 +1945,13 @@ macro expression(args...)
     end
     code =
         Containers.container_code(idxvars, indices, code, requested_container)
+    # Wrap the entire code block in a let statement to make the model act as
+    # a type stable local variable.
+    creation_code = quote
+        let $m = $m
+            $creation_code
+        end
+    end
     # don't do anything with the model, but check that it's valid anyway
     if anonvar
         macro_code = code
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1947,9 +1947,9 @@ macro expression(args...)
         Containers.container_code(idxvars, indices, code, requested_container)
     # Wrap the entire code block in a let statement to make the model act as
     # a type stable local variable.
-    creation_code = quote
+    code = quote
         let $m = $m
-            $creation_code
+            $code
         end
     end
     # don't do anything with the model, but check that it's valid anyway
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1204,6 +1204,17 @@ end
 
 # TODO: update 3-argument @constraint macro to pass through names like @variable
 
+function _wrap_let(model, code)
+    if Meta.isexpr(model, :escape) && model.args[1] isa Symbol
+        return quote
+            let $model = $model
+                $code
+            end
+        end
+    end
+    return code
+end
+
 """
     _constraint_macro(
         args, macro_name::Symbol, parsefun::Function, source::LineNumberNode
@@ -1331,16 +1342,11 @@ function _constraint_macro(
         $parsecode
         $constraintcall
     end
-
     creation_code =
         Containers.container_code(idxvars, indices, code, requested_container)
     # Wrap the entire code block in a let statement to make the model act as
     # a type stable local variable.
-    creation_code = quote
-        let $model = $model
-            $creation_code
-        end
-    end
+    creation_code = _wrap_let(model, creation_code)
     if anonvar
         # Anonymous constraint, no need to register it in the model-level
         # dictionary nor to assign it to a variable in the user scope.
@@ -1947,11 +1953,7 @@ macro expression(args...)
         Containers.container_code(idxvars, indices, code, requested_container)
     # Wrap the entire code block in a let statement to make the model act as
     # a type stable local variable.
-    code = quote
-        let $m = $m
-            $code
-        end
-    end
+    code = _wrap_let(m, code)
     # don't do anything with the model, but check that it's valid anyway
     if anonvar
         macro_code = code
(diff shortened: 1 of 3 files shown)


odow:
@ExpandingMan could you try this branch?

ExpandingMan:
This does seem to work yes.  To be honest, I'm rather confused about what's going on here, I'm running inside a function where there should be no problems with inference.  Guess it can be chalked up to some weird compiler quirk? :shrug:

Regardless, looks fixed, thanks.

odow:
Do you have a reproducible example with the `Any`?

What's going on here is a little complicated. The `@variable` macro ends up creating a function that looks like `map(index -> add_variable(model, index...), 1:0)` where `model` is captured in the closure. But when run in global scope (or anywhere that the compiler doesn't know the type of `model`, it can only prove that `add_variable` returns `GenericVariableRef` (or `Any`, in your case) instead of `VariableRef` if it knew the type of `model`.

The solution is to add a `let model = model` before this `map` so that the compiler can figure out the type.

ExpandingMan:
I did get the `Any` when I was trying in global scope, I did not create an MWE for it happening inside a function.

Since I'm not entirely sure what's going on here, I should resist the urge to be sanctimonious, but, for whatever it's worth, I think this package's macros should output a lot less code, and the vast majority of macros should be exchangeable with simple function calls.  Elaborate macro code is way too hard to diagnose and reason about.

odow:
> I think this package's macros should output a lot less code, and the vast majority of macros should be exchangeable with simple function calls. Elaborate macro code is way too hard to diagnose and reason about.

I think we all agree on this. It's an item on our roadmap: https://jump.dev/JuMP.jl/stable/developers/roadmap/

test/test_hygiene.jl
```diff
@@ -57,4 +57,10 @@ Test.@test ex[3] == 6
 Test.@test i == 10
 Test.@test j == 10
 
+# Test that `model` is inferred correctly inside macros, despite being a
+# non-const global.
+model = JuMP.Model()
+JuMP.@variable(model, x[1:0])
+Test.@test x == JuMP.VariableRef[]
```
blegat, line 64: Maybe add a test where it's not named `model`

src/macros.jl
```diff
@@ -3016,7 +3016,13 @@ macro variable(args...)
             requested_container,
         )
     end
-
+    # Wrap the entire code block in a let statement to make the model act as
+    # a type stable local variable.
+    creation_code = quote
```
blegat, line 3021: Should this be done in `_finalize_macro` to fix other macros as well ?

odow: So it has to go here, because it if we move it to `_finalize_macro`, it'll create something like:
```Julia
quote
    JuMP._valid_model(m, :m)
    let m = m
        begin
            JuMP._error_if_cannot_register(m, :x)
            var"#3###277" = begin
                (JuMP.Containers.container)(((var"##278",)->begin
                            JuMP.add_variable(m, JuMP.model_convert(m, JuMP.build_variable(JuMP.var"#_error#115"{LineNumberNode}(:(#= REPL[5]:1 =#), Core.Box((:m, :(x[1:0])))), JuMP.VariableInfo(false, NaN, false, NaN, false, NaN, false, NaN, false, false))), if JuMP.set_string_names_on_creation(m)
                                    JuMP.string("x", "[", JuMP.string(var"##278"), "]")
                                else
                                    ""
                                end)
                        end), (JuMP.Containers).vectorized_product((JuMP.Base).OneTo(1:0)), JuMP.Containers.AutoContainerType, Any[Symbol("##278")])
            end
            m[:x] = var"#3###277"
            x = var"#3###277"
        end
    end
end
```
and then the `x = var"#3###277"` is a local variable inside the `let` scope.

I've added it to the other macros.

src/macros.jl
```diff
@@ -1939,6 +1945,13 @@ macro expression(args...)
     end
     code =
         Containers.container_code(idxvars, indices, code, requested_container)
+    # Wrap the entire code block in a let statement to make the model act as
+    # a type stable local variable.
+    creation_code = quote
```
odow, line 1950:
```suggestion
    code = quote
```

src/macros.jl
```diff
@@ -1939,6 +1945,13 @@ macro expression(args...)
...
+        let $m = $m
+            $creation_code
```
odow, line 1952:
```suggestion
            $code
```

odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/6179053153

odow:
Ah
```julia
julia> using JuMP

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, x[1:0])
VariableRef[]

julia> m = (; m = model)
(m = A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Names registered in the model: x,)

julia> @variable(m.m, y[1:0])
ERROR: syntax: invalid let syntax around /Users/oscar/.julia/dev/JuMP/src/macros.jl:3035
Stacktrace:
 [1] top-level scope
   @ REPL[8]:1
```

odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/6180324219

odow:
Okay, updated.

We can fix something like `@variable(model, x[1:0])`, but we can't fix something like `@variable(get_model(), x[1:0])`.

This leads to some weird behavior, like `get_model()` returning `Any[]`, but I don't know what else we can do. We can't know the return type of the vector if we don't know what `get_model()` returns.

blegat:
> We can fix something like `@variable(model, x[1:0])`, but we can't fix something like `@variable(get_model(), x[1:0])`.

I think fixing `@variable(model, x[1:0])` is enough

blegat (approved):

PR #3501, opened 2023-09-11
[docs] update to Documenter@1.0.0

odow:
x-ref https://github.com/JuliaDocs/Documenter.jl/pull/2247

preview: https://jump.dev/JuMP.jl/previews/PR3501/


odow:
@mortenpi Looks like it didn't make a difference. Should I try changing the example size threshold?

This PR

image

Last `master`

image

mortenpi:
> Should I try changing the example size threshold?

Nah, that's set to 1 KiB. But yeah, I think the issue is that the figures are using the `show` for `text/html` MIME, which we can't really write to a file..

docs/make.jl
```diff
@@ -549,6 +549,7 @@ write(joinpath(@__DIR__, "src", "JuMP.pdf"), "")
         sidebar_sitename = false,
         # Do no check for large pages.
         size_threshold = nothing,
+        example_size_threshold = 0,
```
odow, line 552:
```suggestion
        example_size_threshold = 1,
```

docs/make.jl
```diff
@@ -4,7 +4,7 @@
 #  file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import Pkg
-Pkg.pkg"add Documenter#2fabe9c3bd40b6288c3d86d96c6031728b2d92ad"
+Pkg.pkg"add Documenter#afeb75d1b2abc7166752c5e00675c8b62ee0bea3"
```
mortenpi, line 7: @odow, could we try again with https://github.com/JuliaDocs/Documenter.jl/pull/2247/commits/b8ea2c45f09b2205c71b56518714b4ea7e0b44b6 ?

```suggestion
Pkg.pkg"add Documenter#b8ea2c45f09b2205c71b56518714b4ea7e0b44b6"
```

odow:
Now there are a bunch of
image

mortenpi:
That's progress! I think? The only pages that now seem to warn about being over the size threshold are the API pages and changelog!

> `but no supported image representation is present as an alternative.`

These ones should be debugged -- why isn't there an alternative MIME for these ones?

> `@example block is above the threshold, falling back to 'image/svg+xml' representation.`

I am not sure about the utility of these warning. My point is that you should be aware that we're picking a suboptimal representation due to size (otherwise users will complain "why is Documenter not making pretty plots"). But also, real-world docs like JuMP will have a lot of these. I think in an ideal world, I would like the user to somehow disable the `text/html` MIME if they can't use it anyway. But I am not convinced that this is the right way to go.

One option to improve this is to collect all of these and print just a single warning saying that `N at-example blocks got replaced with mime X on pages y, z...`.

docs/make.jl
```diff
@@ -549,6 +549,7 @@ write(joinpath(@__DIR__, "src", "JuMP.pdf"), "")
         sidebar_sitename = false,
         # Do no check for large pages.
         size_threshold = nothing,
+        example_size_threshold = 1,
```
odow, line 552:
```suggestion
        # example_size_threshold = 1,
```

odow:
> These ones should be debugged -- why isn't there an alternative MIME for these ones?

I think they're the DataFrames? There are no plots in geographic clustering

https://jump.dev/JuMP.jl/previews/PR3501/tutorials/linear/geographic_clustering/

odow:
> One option to improve this is to collect all of these and print just a single warning saying that 

Yeah that sounds good. 

Although I'd also be happy with it just happening. In what situation is `text/html` a better representation than `image/svg+xml`?

I wonder if documenter could introduce something that we could pirate. Like `default_mime_type(::Any) = MIME("text/html")`, and then we could add something like `Documenter.default_mime_type(::Plots.Plot) = MIME("image/svg+xml")`. Pirating doesn't matter because this is only getting run once. I guess it could also be some sort of argument to `Documenter.HTML()`.

mortenpi:
> I think they're the DataFrames?

Yeah, that would make sense. It looks like some of them are pretty big as well though, going over the 8K limit.

> In what situation is `text/html` a better representation than `image/svg+xml`?

I did not check explicitly, but e.g. with PlotlyJS? Since you could get SVG + interactivity with `text/html`.

> I wonder if documenter could introduce something that we could pirate. Like `default_mime_type(::Any) = MIME("text/html")`, and then we could add something like `Documenter.default_mime_type(::Plots.Plot) = MIME("image/svg+xml")`. Pirating doesn't matter because this is only getting run once. I guess it could also be some sort of argument to `Documenter.HTML()`.

Yeah, something like that could be an option actually. I'll try to remember to open a follow-up issue with this.

docs/make.jl
```diff
@@ -4,7 +4,7 @@
 #  file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import Pkg
-Pkg.pkg"add Documenter#2fabe9c3bd40b6288c3d86d96c6031728b2d92ad"
+Pkg.pkg"add Documenter#b8ea2c45f09b2205c71b56518714b4ea7e0b44b6"
```
odow, line 7:
```suggestion
Pkg.pkg"add Documenter#975a0d3bc5aaca33148d76059a65433be2a3df11"
```

mortenpi:
This looks a bit better!

image

odow:
Yip. Just trying the excludes now.

odow:
Thoughts @mortenpi? I think this is good to merge

mortenpi:
If you wait a day, you might be able to get rid of the `Pkg.pkg"add Documenter#2fabe9c3bd40b6288c3d86d96c6031728b2d92ad"` line too :stuck_out_tongue: But yeah, I think this looks good to me!

odow:
> If you wait a day

Cool cool. That would be good. So close!

docs/Project.toml
```diff
@@ -40,7 +39,7 @@ Clarabel = "=0.5.1"
 DataFrames = "1"
 DimensionalData = "0.24"
 Distributions = "0.25"
-Documenter = "0.27.9, 0.28"
+Documenter = "1"
```
odow, line 42:
```suggestion
Documenter = "=1.0.0"
```

odow: We're accessing some of the Documenter internals (like adding a build step), so it's probably best if we lock the exact version. x-ref: https://documenter.juliadocs.org/stable/lib/semver/#What-is-not-covered-by-SemVer

Issue #3498, opened 2023-09-11
Should we simplify NonLinearExpr similar to AffExpr or QuadExpr?


mitchphillipson:
## Explanation of Issue
I'll start this with an example of the issue

```
julia> using JuMP;
julia> A = [:a,:b];
julia> B = [:c,:d,:e];
julia> p = JuMP.Containers.DenseAxisArray([1 0 3; 0 5 0], A,B);
julia> σ = JuMP.Containers.DenseAxisArray([.5, .75, .25], B);
julia> M = Model();
julia> @variable(M, x[A,B]);
julia> @expression(M, test[a=A], sum(x[a,b]^2*p[a,b] for b∈B));
julia> @expression(M, test_nl[a=A], sum(x[a,b]^(σ[b])*p[a,b] for b∈B));
```

The issue is the difference between
```
julia> test
1-dimensional DenseAxisArray{QuadExpr,1,...} with index sets:
    Dimension 1, [:a, :b]
And data, a 2-element Vector{QuadExpr}:
 x[a,c]² + 3 x[a,e]²
 5 x[b,d]²
```
and
```
julia> test_nl
1-dimensional DenseAxisArray{NonlinearExpr,1,...} with index sets:
    Dimension 1, [:a, :b]
And data, a 2-element Vector{NonlinearExpr}:
 ((x[a,c] ^ 0.5) * 1.0) + ((x[a,d] ^ 0.75) * 0.0) + ((x[a,e] ^ 0.25) * 3.0)
 ((x[b,c] ^ 0.5) * 0.0) + ((x[b,d] ^ 0.75) * 5.0) + ((x[b,e] ^ 0.25) * 0.0)
```

Notice both `test` and `test_nl` differ only in powers, but `test` drops any term multiplied by 0 whereas `test_nl` does not. This is deeper than just the `@show` method, as can be seen here:

```
julia> test[:b].terms
OrderedCollections.OrderedDict{UnorderedPair{VariableRef}, Float64} with 1 entry:
  UnorderedPair{VariableRef}(x[b,d], x[b,d]) => 5.0
```

```
julia> test_nl[:b].args
3-element Vector{Any}:
 (x[b,c] ^ 0.5) * 0.0
 (x[b,d] ^ 0.75) * 5.0
 (x[b,e] ^ 0.25) * 0.0
```
The zero terms no longer exist in `test`. 

[...]

PR #3502, opened 2023-09-11
Simplify NonlinearExpr with + or * and 1 or 0

odow:
One possible fix for https://github.com/jump-dev/JuMP.jl/issues/3498

@mitchphillipson could you try out this branch? (`] add JuMP#od/nlp-simplify`).

It'd be useful if you had the test cases using sparse matrices.

### DIFF
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -409,11 +409,7 @@ function _MA.operate!!(
     return +(x, *(args...))
 end
 
-function _MA.operate(
-    ::typeof(*),
-    x::Real,
-    y::GenericNonlinearExpr{V},
-) where {V}
+function _MA.operate(::typeof(*), x::Real, y::GenericNonlinearExpr{V}) where {V}
     if iszero(x)
         return zero(value_type(V))
     elseif isone(x)
@@ -423,11 +419,7 @@ function _MA.operate(
     end
 end
 
-function _MA.operate(
-    ::typeof(*),
-    x::GenericNonlinearExpr{V},
-    y::Real,
-) where {V}
+function _MA.operate(::typeof(*), x::GenericNonlinearExpr{V}, y::Real) where {V}
     if iszero(y)
         return zero(value_type(V))
     elseif isone(y)
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -383,19 +383,58 @@ for f in (:+, :-, :*, :^, :/, :atan)
     end
 end
 
+function Base.:+(x::Real, y::GenericNonlinearExpr{V}) where {V}
+    if iszero(x)
+        return y
+    end
+    return GenericNonlinearExpr{V}(:+, Any[convert(Float64, x), y])
+end
+
+_is_zero(::_MA.Zero) = true
+_is_zero(x::Number) = iszero(x)
+_is_zero(::Any) = false
+
 function _MA.operate!!(
     ::typeof(_MA.add_mul),
     x::GenericNonlinearExpr,
     args::Vararg{Any,N},
 ) where {N}
     _throw_if_not_real(x)
-    if any(isequal(_MA.Zero()), args)
+    if any(_is_zero, args)
         return x
     elseif x.head == :+
         push!(x.args, *(args...))
         return x
     end
     return +(x, *(args...))
+end
+
+function _MA.operate(
+    ::typeof(*),
+    x::Real,
+    y::GenericNonlinearExpr{V},
+) where {V}
+    if iszero(x)
+        return zero(value_type(V))
+    elseif isone(x)
+        return y
+    else
+        return x * y
+    end
+end
+
+function _MA.operate(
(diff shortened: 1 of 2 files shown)


mitchphillipson:
I'm still testing this, but have observed the following:

```
julia> using JuMP

julia> M = Model();

julia> A = [1, 0, 3]

julia> @variable(M, y[1:3])

julia> sum(y[i]^10*A[i] for i=1:3) #not as expected
(((y[1] ^ 10) * 1.0) + ((y[2] ^ 10) * 0.0)) + ((y[3] ^ 10) * 3.0)

julia> @expression(M, tmp, sum(y[i]^10*A[i] for i=1:3)) #As expected
(y[1] ^ 10) + ((y[3] ^ 10) * 3.0)

julia> T = sum(y[i]^10*A[i] for i=1:3)
(((y[1] ^ 10) * 1.0) + ((y[2] ^ 10) * 0.0)) + ((y[3] ^ 10) * 3.0)

julia> @expression(M, tmpT, T)
((y[1] ^ 10) * 1.0) + ((y[3] ^ 10) * 3.0) + ((y[2] ^ 10) * 0.0)
```
The simplification only takes hold inside an `@expression`. And as a second check, I constructed the expression first then placed it inside an `@expression`, no simplification.

I'm going to test this more tomorrow morning.

odow:
> The simplification only takes hold inside an @expression

Correct

odow:
I don't really know how I feel about this. Is there any evidence that keeping the `+0` and `*0` is a problem? Do we just want to remove them because it looks nice?

The overhead of walking every expression just to see if there is a `+0` might be larger than continuing with one. One problem here is that it makes `*` unstable, because it might return `0.0` or it might return a nonlinear expression.

mitchphillipson:
I'm tending towards agreeing that it's mainly aesthetic and probably adds
unnecessary overhead to model creation.

I'm fine if this gets shelved. I may build a more specific solution for my
use case.

odow:
I do think this is a nice to have, and I can see people requesting it in future.

Maybe I should try what you suggested, and add `drop_zeros!`, which wouldn't be type stable and would be opt-in by users.

mitchphillipson:
Having an optional function would be nice. 90% of the time that operation
isn't necessary, but it's nice to have for that 10%.

The type instability is intriguing to me. I'm guessing this comes from
something like $0*x^5+x^2$ becoming quadratic after `drop_zeros!` is
applied. Is there an issue forcing it to remain nonlinear?

odow:
> The type instability is intriguing to me.

It would come from a definition like this:
```julia
Base.:*(x::Real, y::NonlinearExpr) = iszero(x) ? x : NonlinearExpr(:*, Any[x, y])
```

Now in `a = 2.0; y = a * sin(x)` Julia will infer `y` as `Union{Float64,NonlinearExpr}`.

odow:
So one option would be:
```julia
_is_zero(x::Real) = iszero(x)
_is_zero(::Any) = false

_is_one(x::Real) = isone(x)
_is_one(::Any) = false

simplify(x) = x

simplify(x::Union{GenericAffExpr,GenericQuadExpr}) = drop_zeros!(x)

function simplify(x::GenericNonlinearExpr{V}) where {V}
    if x.head == :* && any(_is_zero, x.args)
        return zero(value_type(V))
    end
    for i in 1:length(x.args)
        x.args[i] = simplify(x.args[i])
    end
    if x.head == :*
        filter!(!_is_one, x.args)
    elseif x.head == :+
        filter!(!_is_zero, x.args)
    end
    if x.head == :+ || x.head == :* && length(x.args) == 1
        return x.args[1]
    end
    return x
end
```

So it's a fairly simple feature addition. But I think it could wait for a bit more usage. It feels like we might want a much more robust expression rewriting system that deals with more than just `+ 0`, `* 0`, and `* 1`.

mitchphillipson:
`simplify` resonates with me, but I would want it to be more robust. My background in Mathematics so my mind always goes to CAS, and I ask "What would Maple/Mathematica do with simplify?" The answer is that it would do much more. However, the "much more" is complicated and requires extensive manipulation of the expression tree and various heuristics of what is "simple". 

The Symbolics.jl (or SymbolicUtils.jl) package has code that does this, [simplify.jl](https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/master/src/simplify.jl). But then they have a helper, [simplify_rules.jl](https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/master/src/simplify_rules.jl), that has a long list of identities and special algebra rules. This is a lot to add right now, especially for cosmetics.

odow:
Yeah the expectations that `simplify` might be CAS-like is what is stopping me from adding it, and why I called the other function `flatten!` instead of `simplify`.

I think we can wait for the current stuff to be out in the world for a while before we decide to add anything.

odow:
I'm going to close this PR for now, but I'll leave the issue open.

PR #3509, opened 2023-09-14
Add support for min and max operators in GenericNonlinearExpr

### LINKED ISSUE #3508: New nonlinear expression stuff doesn't support `min` properly
On `master` with the new non linear expression, one cannot call `min` on two `AffExpr` because `min` internally just calls `isless`, and that is one of the functions that doesn't have operator overloading for the JuMP expression types.

I think the solution is presumably that JuMP needs to add a `min` method that internally uses `JuMP.op_strictly_less_than` (or one of its variants), right?

odow:
Closes #3508

### DIFF
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -354,13 +354,13 @@ end
 
 # Multivariate operators
 
-# The multivariate operators in MOI are +, -, *, ^, /, ifelse, atan
+# The multivariate operators in MOI are +, -, *, ^, /, ifelse, atan, min, max
 #
 # However, ifelse is a builtin, so we can't add methods to it.
 
 # We need only very generic fallbacks for these, because all other cases are
 # caught with more specific methods.
-for f in (:+, :-, :*, :^, :/, :atan)
+for f in (:+, :-, :*, :^, :/, :atan, :min, :max)
     op = Meta.quot(f)
     @eval begin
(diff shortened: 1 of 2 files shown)


davidanthoff:
Wow, that was fast, thanks a lot :)

odow:
I'm about to tag v1.15, so I wanted it in

davidanthoff:
@odow I just ran into a corner case that is not yet handled by this PR: `min` and `max` can also be called with just one argument, and the current implementation doesn't solve that. 

For now I can work around this by adding
```
# TODO Remove this once this is fixed in JuMP
function Base.min(x::JuMP.AbstractJuMPScalar)
    JuMP._throw_if_not_real(x)
    return JuMP.GenericNonlinearExpr{JuMP.variable_ref_type(x)}(:min, x)
end
```
to our own codebase, but I think this should really live here in JuMP. I would add a PR, but not entirely sure how you want to handle this kind of thing as you seem to use macros a bit to add these permutations, right?

odow:
Argh. Why would `min` and `max` be unary methods :cry: Let me take a look

odow:
I think this is better
```julia
Base.min(x::JuMP.AbstractJuMPScalar) = x
Base.max(x::JuMP.AbstractJuMPScalar) = x
```

PR #3516, opened 2023-09-19
Fix typo in the objective of the TSP lazy constraints example.

soppera:
An extra backslash was added in the text.


odow:
Thanks!

PR #3522, opened 2023-09-22
Fix unary min and max operators

odow:
Re-closes https://github.com/jump-dev/JuMP.jl/pull/3509#issuecomment-1730598325

### DIFF
--- a/test/test_nlp_expr.jl
+++ b/test/test_nlp_expr.jl
@@ -968,7 +968,6 @@ function test_operator_max()
     for f in (x, 1.0 * x + 2.0, x^2, sin(x))
         @test max(f) === f
     end
-
     return
 end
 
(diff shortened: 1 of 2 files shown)


test/test_nlp_expr.jl
```diff
@@ -962,6 +965,10 @@ function test_operator_max()
     @variable(model, x)
     @test isequal_canonical(max(x, 1), NonlinearExpr(:max, Any[x, 1.0]))
     @test isequal_canonical(max(1, x, x^2), max(max(1.0, x), x^2))
+    for f in (x, 1.0 * x + 2.0, x^2, sin(x))
+        @test max(f) === f
+    end
+
```
odow, line 971:
```suggestion
```

PR #3529, opened 2023-09-28
[docs] add MiniZinc.jl to installation table

odow:
cc @chriscoey


chriscoey:
Thanks Oscar. I wonder if at some point we might be able to say e.g. SCIP.jl also supports the cp-sat category (perhaps this would be pretty tough to implement though)

odow:
We've generally gone for descriptions of what the `.jl` package supports, not what the underlying solver does. 

The categories are also very coarse, so I don't know if they're super useful.

PR #3530, opened 2023-09-28
Add := operator for Boolean satisfiability problems

odow:
x-ref #2227

This is one improvement suggested by today's call, motivated by the fact that `==` has lower precedence than `&&` and `||`, so `x || y == true` parses as `x || (y == true)`. The new syntax, `x || y := true` parses as `(x || y) := true`.

The key difference of `:=` compared with `==` is that `==` uses `lhs - rhs in EqualTo(false)`, whereas `:=` lowers to the NonlinearExpr` `(lhs == rhs) in EqualTo(true)`. Thus, it should pretty much only be used for Boolean problems that support ScalarNonlinearFunction.

cc @chriscoey thoughts?

See the examples in `test/test_constraint.jl`

TODO: 

 - [x] Documentation

### DIFF
--- a/docs/src/developers/extensions.md
+++ b/docs/src/developers/extensions.md
@@ -200,11 +200,11 @@ julia> model = Model(); @variable(model, x);
 
 julia> function JuMP.parse_constraint_head(
            _error::Function,
-           ::Val{:(:=)},
+           ::Val{:≡},
            lhs,
            rhs,
        )
-           println("Rewriting := as ==")
+           println("Rewriting ≡ as ==")
            new_lhs, parse_code = MutableArithmetics.rewrite(lhs)
            build_code = :(
                build_constraint($(_error), $(new_lhs), MOI.EqualTo($(rhs)))
@@ -212,8 +212,8 @@ julia> function JuMP.parse_constraint_head(
            return false, parse_code, build_code
        end
 
-julia> @constraint(model, x + x := 1.0)
-Rewriting := as ==
+julia> @constraint(model, x + x ≡ 1.0)
+Rewriting ≡ as ==
 2 x = 1
 ```
 
--- a/docs/src/developers/extensions.md
+++ b/docs/src/developers/extensions.md
@@ -200,11 +200,11 @@ julia> model = Model(); @variable(model, x);
 
 julia> function JuMP.parse_constraint_head(
            _error::Function,
-           ::Val{:≝},
+           ::Val{:≔},
            lhs,
            rhs,
        )
-           println("Rewriting ≝ as ==")
+           println("Rewriting ≔ as ==")
            new_lhs, parse_code = MutableArithmetics.rewrite(lhs)
            build_code = :(
                build_constraint($(_error), $(new_lhs), MOI.EqualTo($(rhs)))
@@ -212,8 +212,8 @@ julia> function JuMP.parse_constraint_head(
            return false, parse_code, build_code
        end
 
-julia> @constraint(model, x + x ≝ 1.0)
-Rewriting ≝ as ==
+julia> @constraint(model, x + x ≔ 1.0)
+Rewriting ≔ as ==
 2 x = 1
 ```
 
--- a/docs/src/developers/extensions.md
+++ b/docs/src/developers/extensions.md
@@ -200,11 +200,11 @@ julia> model = Model(); @variable(model, x);
 
 julia> function JuMP.parse_constraint_head(
            _error::Function,
-           ::Val{:≡},
+           ::Val{:≝},
            lhs,
            rhs,
        )
-           println("Rewriting ≡ as ==")
+           println("Rewriting ≝ as ==")
            new_lhs, parse_code = MutableArithmetics.rewrite(lhs)
            build_code = :(
                build_constraint($(_error), $(new_lhs), MOI.EqualTo($(rhs)))
@@ -212,8 +212,8 @@ julia> function JuMP.parse_constraint_head(
            return false, parse_code, build_code
        end
 
-julia> @constraint(model, x + x ≡ 1.0)
-Rewriting ≡ as ==
+julia> @constraint(model, x + x ≝ 1.0)
+Rewriting ≝ as ==
 2 x = 1
 ```
 
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -1534,3 +1534,50 @@ julia> q = [5, 6]
 julia> @constraint(model, M * y + q ⟂ y)
 [y[1] + 2 y[2] + 5, 3 y[1] + 4 y[2] + 6, y[1], y[2]] ∈ MathOptInterface.Complements(4)
 ```
+
+## Boolean constraints
+
+Add a Boolean constraint (a [`MOI.EqualTo{Bool}`](@ref) set) using the `:=`
+operator with a `Bool` right-hand side term:
+
+```jldoctest
+julia> model = GenericModel{Bool}();
+
+julia> @variable(model, x[1:2]);
+
+julia> @constraint(model, x[1] || x[2] := true)
+x[1] || x[2] = true
+
+julia> @constraint(model, x[1] && x[2] := false)
+x[1] && x[2] = false
+
+julia> model
+A JuMP Model
+Feasibility problem with:
+Variables: 2
+`GenericNonlinearExpr{GenericVariableRef{Bool}}`-in-`MathOptInterface.EqualTo{Bool}`: 2 constraints
+Model mode: AUTOMATIC
+CachingOptimizer state: NO_OPTIMIZER
+Solver name: No optimizer attached.
+Names registered in the model: x
+```
+
+Boolean constraints should not be added using the `==` operator because JuMP
+will rewrite the constraint as `lhs - rhs = 0`, and because constraints like
+`a == b == c` require parentheses to diambiguate between `(a == b) == c` and
+`a == (b == c)`. In constrast, `a == b := c` is equivalent to `(a == b) := c`:
+
+```jldoctest
+julia> model = Model();
+
+julia> @variable(model, x[1:2]);
+
+julia> rhs = false
+false
+
+julia> @constraint(model, (x[1] == x[2]) == rhs)
+(x[1] == x[2]) - 0.0 = 0
+
+julia> @constraint(model, x[1] == x[2] := rhs)
+x[1] == x[2] = false
+```
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -1564,8 +1564,8 @@ Names registered in the model: x
 
 Boolean constraints should not be added using the `==` operator because JuMP
 will rewrite the constraint as `lhs - rhs = 0`, and because constraints like
-`a == b == c` require parentheses to diambiguate between `(a == b) == c` and
-`a == (b == c)`. In constrast, `a == b := c` is equivalent to `(a == b) := c`:
+`a == b == c` require parentheses to disambiguate between `(a == b) == c` and
+`a == (b == c)`. In contrast, `a == b := c` is equivalent to `(a == b) := c`:
 
 ```jldoctest
 julia> model = Model();
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -1543,6 +1543,20 @@ function relax_with_penalty!(
     return relax_with_penalty!(model, Dict(); default = default)
 end
 
+struct _BooleanEqualTo end
+
+function build_constraint(::Function, lhs, rhs, ::_BooleanEqualTo)
+    return ScalarConstraint(op_equal_to(lhs, rhs), MOI.EqualTo(true))
+end
+
+function build_constraint(::Function, lhs::Bool, rhs, ::_BooleanEqualTo)
+    return ScalarConstraint(rhs, MOI.EqualTo(lhs))
+end
+
+function build_constraint(::Function, lhs, rhs::Bool, ::_BooleanEqualTo)
+    return ScalarConstraint(lhs, MOI.EqualTo(rhs))
+end
+
 function parse_constraint_head(error_fn::Function, ::Val{:(:=)}, lhs, rhs)
     new_lhs, parse_code_lhs = _rewrite_expression(lhs)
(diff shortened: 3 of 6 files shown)


test/test_macros.jl
```diff
@@ -89,7 +89,7 @@ end
 
 struct CustomType end
 
-function JuMP.parse_constraint_head(_error::Function, ::Val{:(:=)}, lhs, rhs)
+function JuMP.parse_constraint_head(_error::Function, ::Val{:≝}, lhs, rhs)
```
odow, line 92:
```suggestion
function JuMP.parse_constraint_head(_error::Function, ::Val{:≔}, lhs, rhs)
```

test/test_macros.jl
```diff
@@ -326,7 +326,7 @@ function test_extension_custom_expression_test(
 )
     model = ModelType()
     @variable(model, x)
-    @constraint(model, con_ref, x := CustomType())
+    @constraint(model, con_ref, x ≝ CustomType())
```
odow, line 329:
```suggestion
    @constraint(model, con_ref, x ≔ CustomType())
```

docs/src/developers/extensions.md
```diff
@@ -200,20 +200,20 @@ julia> model = Model(); @variable(model, x);
 
 julia> function JuMP.parse_constraint_head(
            _error::Function,
-           ::Val{:(:=)},
+           ::Val{:≝},
```
odow, line 203:
```suggestion
           ::Val{:≔},
```

docs/src/developers/extensions.md
```diff
@@ -200,20 +200,20 @@ julia> model = Model(); @variable(model, x);
...
            lhs,
            rhs,
        )
-           println("Rewriting := as ==")
+           println("Rewriting ≝ as ==")
```
odow, line 207:
```suggestion
           println("Rewriting ≔ as ==")
```

docs/src/developers/extensions.md
```diff
@@ -200,20 +200,20 @@ julia> model = Model(); @variable(model, x);
...
            new_lhs, parse_code = MutableArithmetics.rewrite(lhs)
            build_code = :(
                build_constraint($(_error), $(new_lhs), MOI.EqualTo($(rhs)))
            )
            return false, parse_code, build_code
        end
 
-julia> @constraint(model, x + x := 1.0)
-Rewriting := as ==
+julia> @constraint(model, x + x ≝ 1.0)
```
odow, line 215:
```suggestion
julia> @constraint(model, x + x ≔ 1.0)
```

docs/src/developers/extensions.md
```diff
@@ -200,20 +200,20 @@ julia> model = Model(); @variable(model, x);
...
+Rewriting ≝ as ==
```
odow, line 216:
```suggestion
Rewriting ≔ as ==
```

chriscoey:
Looks ok! No strong opinions from me. Thanks for experimenting.

odow:
So this minizinc
```
var bool: x;
var bool: y;
constraint x - y == false;
solve maximize x + y;
```
makes the flatzinc
```
var bool: x:: output_var;
var bool: y:: output_var;
var 0..2: X_INTRODUCED_0_:: is_defined_var;
var 0..1: X_INTRODUCED_2_ ::var_is_introduced :: is_defined_var;
var 0..1: X_INTRODUCED_3_ ::var_is_introduced :: is_defined_var;
constraint int_eq(X_INTRODUCED_3_,X_INTRODUCED_2_);
constraint int_lin_eq([1,1,-1],[X_INTRODUCED_3_,X_INTRODUCED_2_,X_INTRODUCED_0_],0):: ctx_pos:: defines_var(X_INTRODUCED_0_);
constraint bool2int(y,X_INTRODUCED_2_):: defines_var(X_INTRODUCED_2_);
constraint bool2int(x,X_INTRODUCED_3_):: defines_var(X_INTRODUCED_3_);
solve  maximize X_INTRODUCED_0_;
```

So the minizinc compiler:

 * Re-interprets `bool` variables as `0..1` integer variables
 * Compiles `x - y == false` into `x == y`

I guess any arithmetic between `bool` is converted into integers.

I don't know what that means for us. But perhaps we don't need this special syntax.

odow:
```
var bool: x;
var bool: y;
constraint x - y == false;
solve satisfy;
```

This is with `Gecode`
```
var bool: x:: output_var;
var bool: y:: output_var;
var 0..1: X_INTRODUCED_1_ ::var_is_introduced :: is_defined_var;
var 0..1: X_INTRODUCED_2_ ::var_is_introduced :: is_defined_var;
constraint int_eq(X_INTRODUCED_2_,X_INTRODUCED_1_);
constraint bool2int(y,X_INTRODUCED_1_):: defines_var(X_INTRODUCED_1_);
constraint bool2int(x,X_INTRODUCED_2_):: defines_var(X_INTRODUCED_2_);
solve  satisfy;
```

odow:
There's also a general problem with `VariableRef{Bool}` and arithmetic:
```julia
julia> model = GenericModel{Bool}()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

2-element Vector{GenericVariableRef{Bool}}:
 x[1]
 x[2]

julia> x[1] - x[2]
ERROR: MethodError: no method matching _build_aff_expr(::Bool, ::Bool, ::GenericVariableRef{Bool}, ::Int64, ::GenericVariableRef{Bool})

Closest candidates are:
  _build_aff_expr(::V, ::V, ::K, ::V, ::K) where {V, K}
   @ JuMP ~/.julia/dev/JuMP/src/aff_expr.jl:87
  _build_aff_expr(::V, ::V, ::K) where {V, K}
   @ JuMP ~/.julia/dev/JuMP/src/aff_expr.jl:81

Stacktrace:
 [1] -(lhs::GenericVariableRef{Bool}, rhs::GenericVariableRef{Bool})
   @ JuMP ~/.julia/dev/JuMP/src/operators.jl:120
 [2] top-level scope
   @ REPL[14]:1
```

I think this just means that you can't write `@constraint(model, x[1] == x[2])`. 

You have to write `@constraint(model, (x[1] == x[2]) == true)`, which is an argument for `@constraint(model, x[1] := x[2])`.

src/constraints.jl
```diff
@@ -1542,3 +1542,22 @@ function relax_with_penalty!(
...
+
+function parse_constraint_head(error_fn::Function, ::Val{:(:=)}, lhs, rhs)
+    new_lhs, parse_code_lhs = _rewrite_expression(lhs)
+    new_rhs, parse_code_rhs = _rewrite_expression(rhs)
+    parse_code = quote
+        $parse_code_lhs
+        $parse_code_rhs
+    end
+    build_code = quote
+        if $new_rhs isa Bool
```
blegat, line 1554: Since we do it at the parsing level, it won't work if `rhs` is a Julia variable (not a JuMP decision variable) for which the value is a `Bool` right ?

odow: This part is at runtime. See the tests.

test/test_constraint.jl
```diff
@@ -1721,4 +1721,80 @@ function test_triangle_vec()
...
+    @test isequal_canonical(o.func, op_and(x[1], x[2]))
+    @test o.set == MOI.EqualTo(false)
+    # x[1] && x[2] := true
+    c = @constraint(model, x[1] && x[2] := true)
+    o = constraint_object(c)
+    @test isequal_canonical(o.func, op_and(x[1], x[2]))
+    @test o.set == MOI.EqualTo(true)
+    # x[1] || x[2] := y
+    y = true
+    c = @constraint(model, x[1] || x[2] := y)
```
odow, line 1749: @blegat this works

src/constraints.jl
blegat: Ah indeed. I guess it's best to add code inside the `build_constraint` that add code that is generated by the macro if possible

blegat: But in this case, it might be difficule

blegat: We would need a new JuMP set `Equal` with and do `build_constraint(error, (a, b), JuMP.Equal()`

mlubin:
The asymmetry of the symbol `:=` is a bit annoying/misleading. 
```
@constraint(model, x[1] || x[2] := y)
```
and
```
@constraint(model, y := x[1] || x[2])
```
should be equivalent, but superficially they don't seem to be.

odow:
We could use `===`, but that could be confusing with `==`. The `<-->` syntax is used for reified constraints, which isn't quite the same.

Ideally, people would only write `lhs := true` or `lhs := false`.

odow:
I had a play to try and come up with a better syntax, but I couldn't really think of anything

pulsipher:
This would be helpful for defining logical constraints in DisjunctiveProgramming.jl.

odow:
> This would be helpful for defining logical constraints in DisjunctiveProgramming.jl.

Any syntax suggestions?

odow:
But it'd be `EqualTo{Bool}`, is that not sufficient?

odow:
Here are our options for operators with a higher precedence than `||` `&&` `==` etc:
https://github.com/JuliaLang/julia/blob/342e39433c6f28ce0cbf7b6ad5478ab8a8ac2031/src/julia-parser.scm#L7-L13
```
  (append! (add-dots '(= += -= −= *= /= //= |\\=| ^= ÷= %= <<= >>= >>>= |\|=| &= ⊻= ≔ ⩴ ≕))
           (add-dots '(~))
           '(:= $=)))
;; comma - higher than assignment outside parentheses, lower when inside
(define prec-pair (add-dots '(=>)))
(define prec-conditional '(?))
(define prec-arrow       (add-dots '(← → ↔ ↚ ↛ ↞ ↠ ↢ ↣ ↦ ↤ ↮ ⇎ ⇍ ⇏ ⇐ ⇒ ⇔ ⇴ ⇶ ⇷ ⇸ ⇹ ⇺ ⇻ ⇼ ⇽ ⇾ ⇿ ⟵ ⟶ ⟷ ⟹ ⟺ ⟻ ⟼ ⟽ ⟾ ⟿ ⤀ ⤁ ⤂ ⤃ ⤄ ⤅ ⤆ ⤇ ⤌ ⤍ ⤎ ⤏ ⤐ ⤑ ⤔ ⤕ ⤖ ⤗ ⤘ ⤝ ⤞ ⤟ ⤠ ⥄ ⥅ ⥆ ⥇ ⥈ ⥊ ⥋ ⥎ ⥐ ⥒ ⥓ ⥖ ⥗ ⥚ ⥛ ⥞ ⥟ ⥢ ⥤ ⥦ ⥧ ⥨ ⥩ ⥪ ⥫ ⥬ ⥭ ⥰ ⧴ ⬱ ⬰ ⬲ ⬳ ⬴ ⬵ ⬶ ⬷ ⬸ ⬹ ⬺ ⬻ ⬼ ⬽ ⬾ ⬿ ⭀ ⭁ ⭂ ⭃ ⥷ ⭄ ⥺ ⭇ ⭈ ⭉ ⭊ ⭋ ⭌ ￩ ￫ ⇜ ⇝ ↜ ↝ ↩ ↪ ↫ ↬ ↼ ↽ ⇀ ⇁ ⇄ ⇆ ⇇ ⇉ ⇋ ⇌ ⇚ ⇛ ⇠ ⇢ ↷ ↶ ↺ ↻ --> <-- <-->)))
```

For ASCII operators, we really only have `:=` or `~`. 

Is `@constraint(model, x[1] || x[2] ~ y)` and `@constraint(model, true ~ x[1] || x[2])` to horrific?

odow:
`===` is ruled out because of
```julia
julia> dump(:(x == y === true))
Expr
  head: Symbol comparison
  args: Array{Any}((5,))
    1: Symbol x
    2: Symbol ==
    3: Symbol y
    4: Symbol ===
    5: Bool true
```

odow:
I've changed it so that `:=` returns `EqualTo{Bool}`, even for other value type models.

I wonder if we should enforce that the `func := rhs` where `rhs` is a `Bool`, and not support others. Then we won't have the non-symmetric problem, and it's more obvious when to use `:=` than `==`.

pulsipher:
I think restricting the RHS to be a `Bool` is for the best. It satisfies the use case for DisjunctiveProgramming and it alings with @chriscoey's original idea of specifying logic constraints without a set (implying the RHS is true).

odow:
Yeah, I've changed to only rhs Bool and it makes more sense.

pulsipher (approved):
This looks great.

docs/src/manual/constraints.md
```diff
@@ -1534,3 +1534,50 @@ julia> q = [5, 6]
...
+Model mode: AUTOMATIC
+CachingOptimizer state: NO_OPTIMIZER
+Solver name: No optimizer attached.
+Names registered in the model: x
+```
+
+Boolean constraints should not be added using the `==` operator because JuMP
+will rewrite the constraint as `lhs - rhs = 0`, and because constraints like
+`a == b == c` require parentheses to diambiguate between `(a == b) == c` and
+`a == (b == c)`. In constrast, `a == b := c` is equivalent to `(a == b) := c`:
```
odow, line 1568:
```suggestion
`a == (b == c)`. In contrast, `a == b := c` is equivalent to `(a == b) := c`:
```

docs/src/manual/constraints.md
```diff
@@ -1534,3 +1534,50 @@ julia> q = [5, 6]
...
+`GenericNonlinearExpr{GenericVariableRef{Bool}}`-in-`MathOptInterface.EqualTo{Bool}`: 2 constraints
+Model mode: AUTOMATIC
+CachingOptimizer state: NO_OPTIMIZER
+Solver name: No optimizer attached.
+Names registered in the model: x
+```
+
+Boolean constraints should not be added using the `==` operator because JuMP
+will rewrite the constraint as `lhs - rhs = 0`, and because constraints like
+`a == b == c` require parentheses to diambiguate between `(a == b) == c` and
```
odow, line 1567:
```suggestion
`a == b == c` require parentheses to disambiguate between `(a == b) == c` and
```

blegat (approved):

mlubin (approved):

odow:
We may want to expose the `_DoNotConvertSet` as a public API, but that can wait for a request.

Issue #3531, opened 2023-10-02
2 problems found in doc `Benders Decomposition`


WalterMadelim:
In [Callback method](https://jump.dev/JuMP.jl/dev/tutorials/algorithms/benders_decomposition/#Callback-method),
1. line 4 in the first code block, the objective should be `θ + c_1' * x`, not `θ` ? I don't understand the reason not adding `c1 * x` 
2. line 4 counting from bottom, in the second code block, inside function my_callback(cb_data), `model` should change to `lazy_model` ? Is `model` meant to be the one in [iterative method](https://jump.dev/JuMP.jl/dev/tutorials/algorithms/benders_decomposition/#benders_iterative)? I don't know the motivation.

After making these 2 changes, I got the reasonable answer below
```julia
julia> x_optimal = value.(x)
2-element Vector{Float64}:
 -0.0
  1.0
julia> y_optimal = optimal_ret.y
2-element Vector{Float64}:
 0.0
 0.0
```
which is from the complete source code below, please try it
```julia
using JuMP
import Gurobi
import Printf

function solve_subproblem(x)
    model = Model(Gurobi.Optimizer)
    @variable(model, y[1:dim_y] >= 0)
    con = @constraint(model, A_2 * y .<= b - A_1 * x)
    @objective(model, Min, c_2' * y)
    optimize!(model)
    @assert termination_status(model) == OPTIMAL
    return (obj = objective_value(model), y = value.(y), π = dual.(con))
end

function print_iteration(k, args...)
    f(x) = Printf.@sprintf("%12.4e", x)
    println(lpad(k, 9), " ", join(f.(args), " "))
    return
end

c_1 = [1, 4]
c_2 = [2, 3]
dim_x = length(c_1)
dim_y = length(c_2)
b = [-2; -3]
A_1 = [1 -3; -1 -3]
A_2 = [1 -2; -1 -1]
M = -1000;
MAXIMUM_ITERATIONS = 100
ABSOLUTE_OPTIMALITY_GAP = 1e-6

lazy_model = Model(Gurobi.Optimizer)
@variable(lazy_model, x[1:dim_x] >= 0, Int)
@variable(lazy_model, θ >= M)
@objective(lazy_model, Min, θ + c_1' * x) # problem here
print(lazy_model)

k = 0

"""
    my_callback(cb_data)

A callback that implements Benders decomposition. Note how similar it is to the
inner loop of the iterative method.
"""
function my_callback(cb_data)
[...]

PR #3532, opened 2023-10-04
[docs] fix benders_decomposition.jl

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3531

Thanks for catching this @WalterMadelim. I'm surprised that it didn't error!

### DIFF
--- a/docs/src/tutorials/algorithms/benders_decomposition.jl
+++ b/docs/src/tutorials/algorithms/benders_decomposition.jl
@@ -253,7 +253,7 @@ function my_callback(cb_data)
     end
     x_k = callback_value.(cb_data, x)
     θ_k = callback_value(cb_data, θ)
-    number_of_subproblem_solves += 1
+    global number_of_subproblem_solves += 1
     ret = solve_subproblem(x_k)
     if θ_k < (ret.obj - 1e-6)
         ## Only add the constraint if θ_k violates the constraint
--- a/docs/src/tutorials/algorithms/benders_decomposition.jl
+++ b/docs/src/tutorials/algorithms/benders_decomposition.jl
@@ -273,7 +273,7 @@ optimize!(lazy_model)
 
 number_of_subproblem_solves
 
-# But for larger problem, you can expect the callback algorithm to be more
+# But for larger problems, you can expect the callback algorithm to be more
 # efficient than the iterative algorithm.
 
 # Finally, we can obtain the optimal solution:
--- a/docs/src/tutorials/algorithms/benders_decomposition.jl
+++ b/docs/src/tutorials/algorithms/benders_decomposition.jl
@@ -244,8 +244,7 @@ print(lazy_model)
 
 # What differs is that we write a callback function instead of a loop:
 
-k = 0
-
+number_of_subproblem_solves = 0
 function my_callback(cb_data)
     status = callback_node_status(cb_data, lazy_model)
     if status != MOI.CALLBACK_NODE_STATUS_INTEGER
@@ -254,6 +253,7 @@ function my_callback(cb_data)
     end
     x_k = callback_value.(cb_data, x)
     θ_k = callback_value(cb_data, θ)
+    number_of_subproblem_solves += 1
     ret = solve_subproblem(x_k)
     if θ_k < (ret.obj - 1e-6)
         ## Only add the constraint if θ_k violates the constraint
@@ -268,6 +268,13 @@ set_attribute(lazy_model, MOI.LazyConstraintCallback(), my_callback)
 # Now when we optimize!, our callback is run:
 
 optimize!(lazy_model)
+
+# For this model, the callback algorithm required more solves of the subproblem:
+
+number_of_subproblem_solves
+
+# But for larger problem, you cann expect the callback algorithm to be more
+# efficient than the iterative algorithm.
 
 # Finally, we can obtain the optimal solution:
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/algorithms/benders_decomposition.jl
```diff
@@ -266,7 +266,7 @@ function my_callback(cb_data)
         return
     end
     cut = @build_constraint(θ >= ret.obj + -ret.π' * A_1 * (x .- x_k))
-    MOI.submit(model, MOI.LazyConstraint(cb_data), cut)
+    MOI.submit(lazy_model, MOI.LazyConstraint(cb_data), cut)
```
odow, line 269: I should look at why this didn't error. It should have...

WalterMadelim: image
It has indeed, after changing GLPK to Gurobi. (I have not installed GLPK)

odow: Ah, so this might be an issue with GLPK, not with JuMP. Thanks for looking.

odow:
Now I don't understand what's going on here:
image

odow:
GLPK callbacks are weird

WalterMadelim:
I think the output in your figure is normal. There are multiple "Terminating with the optimal solution", because this is a callback function, which will be called frequently during solving, and the solver's terminating criterion doesn't depend on user's callback function. As for the Jumping up - Jumping down lower bound, this is also reasonable because the problem is changing during solving. The puzzling point is why GLPK can return the right answer, even if you submit to `model` not `lazy_model`.

But it should be fine after revising. I agree with the new callback function.

WalterMadelim:
> GLPK callbacks are weird

Actually I would prefer to use a solver-dependent callback. (For example, I would refer to Gurobi.jl directly when I have a need)

docs/src/tutorials/algorithms/benders_decomposition.jl
```diff
@@ -239,34 +239,27 @@ y_optimal = optimal_ret.y
...
-    global k += 1
+    status = callback_node_status(cb_data, lazy_model)
+    if status != MOI.CALLBACK_NODE_STATUS_INTEGER
+        ## Only add the constraint if `x` is an integer feasible solution
+        return
+    end
     x_k = callback_value.(cb_data, x)
     θ_k = callback_value(cb_data, θ)
-    lower_bound = c_1' * x_k + θ_k
+    number_of_subproblem_solves += 1
```
odow, line 256:
```suggestion
    global number_of_subproblem_solves += 1
```

odow:
I don't understand why GLPK calls so many times: 
https://jump.dev/JuMP.jl/previews/PR3532/tutorials/algorithms/benders_decomposition/#Callback-method

image

WalterMadelim:
I thought there might be no one who is more familiar with this than you, because it seems that you are the main contributer of GLPK.jl.🥲

From my point of view, the `function my_callback(cb_data)` is special, because it is controlled by the solver rather than Julia. I tried to put break points `@bp`'s inside it, but they were never hitted. Therefore this function is more likely to be a component of the branch and cut module: when the main algorithm need a user's callback, then a lazy cut is retrieved and stored.

Therefore, I think the name `number_of_subproblem_solves` is not proper. It's the number of times that callback module in GLPK is executed when `status == MOI.CALLBACK_NODE_STATUS_INTEGER`. Subproblems modeling and branching should reside in a deeper module than callback module in GLPK (my comprehension).

This adapted version of TSP callback code (using MOI without JuMP) might be a comparison. Solver is Gurobi, the problem is larger and the counter reaches `20`, compared with your result `341`. (you can try it)

https://github.com/WalterMadelim/p8f.jl/blob/main/src/TSP_GRB_callback.jl

odow:
> I thought there might be no one who is more familiar with this than you, because it seems that you are the main contributer of GLPK.jl

Yes :smile: I meant this as "given I understand exactly how it works, it is confusing that it is called so many times."

docs/src/tutorials/algorithms/benders_decomposition.jl
```diff
@@ -276,8 +269,12 @@ set_attribute(lazy_model, MOI.LazyConstraintCallback(), my_callback)
 
 optimize!(lazy_model)
 
-# Note how this problem also takes 4 iterations to converge, but the sequence
-# of bounds is different compared to the iterative method.
+# For this model, the callback algorithm required more solves of the subproblem:
+
+number_of_subproblem_solves
+
+# But for larger problem, you cann expect the callback algorithm to be more
```
odow, line 276:
```suggestion
# But for larger problem, you can expect the callback algorithm to be more
```

docs/src/tutorials/algorithms/benders_decomposition.jl
```diff
@@ -276,8 +269,12 @@ set_attribute(lazy_model, MOI.LazyConstraintCallback(), my_callback)
 
 optimize!(lazy_model)
 
-# Note how this problem also takes 4 iterations to converge, but the sequence
-# of bounds is different compared to the iterative method.
+# For this model, the callback algorithm required more solves of the subproblem:
+
+number_of_subproblem_solves
+
+# But for larger problem, you can expect the callback algorithm to be more
```
odow, line 276:
```suggestion
# But for larger problems, you can expect the callback algorithm to be more
```

PR #3536, opened 2023-10-06
DNM: Fix and test support for Σ, ∑,and ∏

### LINKED ISSUE #3533: Use of Σ is not consistent
Someone complained (rightly) about:
```julia
julia> model = Model();

 x[1]
 x[2]

julia> @expression(model, Σ(x[i] for i in 1:2))
x[1] + x[2]

julia> @expression(model, Σ(x))
ERROR: UndefVarError: `Σ` not defined
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/MutableArithmetics/cnvbo/src/rewrite.jl:321 [inlined]
 [2] macro expansion
   @ ~/.julia/dev/JuMP/src/macros.jl:717 [inlined]
 [3] top-level scope
   @ REPL[5]:1
```

cc @dourouc05

odow:
Closes #3533 

We've always supported Σ, ∑,and ∏, but these weren't tested, and the only worked for generators, not with a single itterable argument.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -684,6 +684,12 @@ function _rewrite_to_jump_logic(x)
             return Expr(:call, op_greater_than_or_equal_to, x.args[2:end]...)
         elseif x.args[1] == :(==)
             return Expr(:call, op_equal_to, x.args[2:end]...)
+        elseif x.args[1] == :Σ
+            return Expr(:call, sum, x.args[2:end]...)
+        elseif x.args[1] == :∑
+            return Expr(:call, sum, x.args[2:end]...)
+        elseif x.args[1] == :∏
+            return Expr(:call, prod, x.args[2:end]...)
         end
     elseif Meta.isexpr(x, :||)
         return Expr(:call, op_or, x.args...)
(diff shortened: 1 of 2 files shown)


blegat:
Is that an issues with MutableArithmetics, it seems that it is always using `_is_sum`

odow:
Doc failure is https://github.com/jump-dev/JuMP.jl/pull/3538

blegat:
I see, when it's going through this code then it works but when it's just gets executed as is then it fails.
I guess the issue with this PR is if the user does
```julia
julia> Σ = 1
1
```
and then uses that in its macro. Then if you replace it with `sum`, that's an issue because it's not `1` anymore.
Maybe MA shouldn't do what it does. I think that comes from a copy-paste of the old rewriting code from JuMP.

odow:
> Maybe MA shouldn't do what it does. I think that comes from a copy-paste of the old rewriting code from JuMP.

Yeah

odow:
I think I'm going to close this as won't-fix. The `\Sigma` isn't documented, so it's mostly a legacy quirk. We should encourage people not to use it.

PR #3537, opened 2023-10-06
Fix text/latex printing of Interval sets

### LINKED ISSUE #3535: Printing constraints with intervals escapes the bounds once too much
When running:
https://jump.dev/JuMP.jl/stable/tutorials/linear/diet/

but printing the model constraints:
```julia
@constraint(
    model,
    [row in eachrow(limits)],
    row.min <= sum(foods[!, row.nutrient] .* foods.x) <= row.max,
)
```
the interval square brackets `[lb, ub]` gets printed with an additional backslash around the square brackets

cc @odow

odow:
Closes #3535 

This has been around for aaaaaaages:
https://github.com/jump-dev/JuMP.jl/pull/250/files#diff-daf635fc60966c7826dcb301089f0408e7834d17ce0ec46018ff674814ba6186R72-R73

I guess we've just never noticed.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -741,7 +741,7 @@ end
 
 function in_set_string(::MIME"text/latex", set::MOI.Interval)
     lower, upper = _string_round(set.lower), _string_round(set.upper)
-    return string("\\in \\[", lower, ", ", upper, "\\]")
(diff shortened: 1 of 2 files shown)


odow:
Doc failure is https://github.com/jump-dev/JuMP.jl/pull/3538

PR #3545, opened 2023-10-18
Add PolyJuMP KKT and QCQP solvers

blegat:
We could debate whether it should be two of one rows since they both live in the same package. However, they are really two different solvers, the only thing they share is that they supports `ScalarPolynomialFunction` constraints so I think it's really two rows.


odow:
I don't really have an opinion. I'm okay with either.

We could/should also update the extension: https://jump.dev/JuMP.jl/previews/PR3545/packages/PolyJuMP/

But it's okay if I do this in a separate PR.

blegat:
Good point, I bumped it here

odow:
I guess I need to figure out how to change the vale error threshold for pages in the package directory.

PR #3546, opened 2023-10-20
Add MINOTAUR to solver list

jd-foster:
MINOTAUR can run from an .nl stub (see "Running solvers" section of https://minotaur-solver.github.io/downloads.html)

BSD-like licence: https://en.wikipedia.org/wiki/BSD_licenses

Might be too early to add this without a JLL, but binaries are available from the first link above.


docs/src/installation.md
```diff
@@ -127,6 +127,7 @@ The link in the `Solver` column is the corresponding Julia package.
 | [Loraine.jl](https://github.com/kocvara/Loraine.jl)                            |                                                                                  |        | MIT      | LP, SDP                   |
 | [MadNLP.jl](https://github.com/sshin23/MadNLP.jl)                              |                                                                                  |        | MIT      | LP, QP, NLP               |
 | [MiniZinc](https://www.minizinc.org/)                                          | [MiniZinc.jl](https://github.com/jump-dev/MiniZinc.jl)                           | Manual | MPL-2    | CP-SAT                    |
+| [MINOTAUR](https://github.com/coin-or/minotaur)                                | [AmplNLWriter.jl](https://github.com/jump-dev/AmplNLWriter.jl)                   | Manual | BSD-like | (MI)NLP                   |
```
blegat, line 130:
```suggestion
| [Minotaur](https://github.com/coin-or/minotaur)                                | [AmplNLWriter.jl](https://github.com/jump-dev/AmplNLWriter.jl)                   | Manual | BSD-like | (MI)NLP                   |
```

blegat: Apart from the README, I see it mostly spelled with lowercase, e.g., https://minotaur-solver.github.io/downloads.html
@krushna33 @mustafa712 what's the correct case ?

mustafa712: Hi @blegat,
We are sorry for the inconsistency in the naming. We would like it spelled with lowercase as in "Minotaur" instead of "MINOTAUR". Currently, I have updated the README file with "Minotaur". There are some other instances in our source code where we it is still "MINOTAUR". We will update it in the near future.

We are planning many changes and updates for the solver. Please let us know if you need some specific updates in the solver with respect to JuMP. We would be keen to do so.

Regards
Minotaur Team

blegat: Thanks for your reply and for the development of Minotaur :) We will call it "Minotaur" in the our solver table then

PR #3551, opened 2023-10-29
Add start_value, lower_bound, and upper_bound support for GenericAffExpr

### LINKED ISSUE #3550: set_start_value for complex variables
x-ref https://discourse.julialang.org/t/wram-start-of-hermitianpsdcone-in-jump/105491

At minimum, we need a better error message and some better documentation

odow:
Closes #3550 

This PR is up for debate, but our problem is that some usages of `@variable` return expressions of variables, and then common operations like `lower_bound` or `set_start_value` do not work.

In most common cases we _can_ make this work, so we probably should. This would improve the user-experience, particularly for those using `Complex` number support.

### DIFF
--- a/src/aff_expr.jl
+++ b/src/aff_expr.jl
@@ -810,197 +810,71 @@ function moi_function_type(::Type{<:Vector{<:GenericAffExpr{T}}}) where {T}
     return MOI.VectorAffineFunction{T}
 end
 
+"""
+    _eval_as_variable(f::F, x::GenericAffExpr, args...) where {F}
+
+In many cases, `@variable` can return a `GenericAffExpr` instead of a
+`GenericVariableRef`. This is particularly the case for complex-valued
+expressions. To make common operatons like `lower_bound(x)` work, we should
+forward the method if and only if `x` is convertable to a `GenericVariableRef`.
+"""
+function _eval_as_variable(f::F, x::GenericAffExpr, args...) where {F}
+    if length(x.terms) != 1
+        error(
+            "Cannot call $f with $x because it is not a real-valued affine " *
+            "expression of one variable.",
+        )
+    elseif !isone(first(keys(x.terms)))
+        error(
+            "Cannot call $f with $x because it is not a real-valued affine " *
+            "expression of one variable with a coefficient of `+1`.",
+        )
+    end
+    return f(first(keys(x.terms)), args...)
+end
+
+function _eval_as_variable(
+    f::F,
+    x::GenericAffExpr{T},
+    args...,
+) where {F,T<:Complex}
+    return error(
+        "Cannot call $f with $x because it is not a real-valued affine " *
+        "expression of one variable with a coefficient of `+1`. Use " *
+        "`real(x)` or `imag(x)` to obtain the real and imaginary part and " *
+        "pass that instead.",
+    )
+end
+
 # start_value(::GenericAffExpr)
 
-function start_value(x::GenericAffExpr)
-    # Unset start values return `nothing`, but we cannot do arithmetic on
-    # `nothing`. So convert any nothings to `missing`, then compute the `value`
-    # and then re-convert back to `nothing` if the result is `missing`.
-    to_missing_start(x) = something(start_value(x), missing)
-    return coalesce(value(to_missing_start, x), nothing)
-end
+start_value(x::GenericAffExpr) = _eval_as_variable(start_value, x)
 
-function set_start_value(x::GenericAffExpr, ::Nothing)
-    set_start_value.(keys(x.terms), nothing)
-    return
-end
-
-function set_start_value(x::GenericAffExpr, value::Number)
-    if isempty(x.terms) && x.constant == value
-        return
-    elseif length(x.terms) != 1
-        error(
-            "Cannot set the start value of $x because it does not contain " *
-            "exactly one variable",
-        )
-    end
-    # a * x + b == value --> x == (value - b) / a
-    a = first(values(x.terms))
-    if !isone(abs(a))
-        error(
-            "Cannot set the start value of `$x` because it does not contain " *
-            "exactly one variable with a coefficient of `+1` or `-1`",
-        )
-    end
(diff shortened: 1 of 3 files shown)


src/aff_expr.jl
```diff
@@ -809,3 +809,100 @@ moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
...
+end
+
+function set_start_value(x::GenericAffExpr, ::Nothing)
+    set_start_value.(keys(x.terms), nothing)
+    return
+end
+
+function set_start_value(x::GenericAffExpr, value::Number)
+    if isempty(x.terms) && iszero(value)
+        return
```
blegat, line 830: What if `x.constant` is not zero ?

src/aff_expr.jl
```diff
@@ -809,3 +809,100 @@ moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
...
+function set_start_value(x::GenericAffExpr{T}, value::Number) where {T<:Complex}
+    value_T = convert(T, value)
+    set_start_value(real(x), real(value_T))
+    set_start_value(imag(x), imag(value_T))
+    return
+end
+
+# lower_bound(::GenericAffExpr)
+
+has_lower_bound(x::GenericAffExpr) = all(has_lower_bound, keys(x.terms))
```
blegat, line 852: For the terms with negative coefficients, it should be `has_upper_bound`

src/aff_expr.jl
```diff
@@ -809,3 +809,100 @@ moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
...
+
+lower_bound(x::GenericAffExpr) = value(lower_bound, x)
```
blegat, line 854: Does not work for negative coefficients

src/aff_expr.jl
```diff
@@ -809,3 +809,100 @@ moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
...
+
+delete_lower_bound(x::GenericAffExpr) = delete_lower_bound.(keys(x.terms))
```
blegat, line 856: Not sure we want to allow this when there is more than one var. If there is one variable only, we should delete the upper bound if there is only one term

src/aff_expr.jl
odow: Oops. Yeah. This makes things a lot more complicated.

odow:
Hmm. There are actually a bunch of edge cases once you dig in:
```Julia
    model = Model()
    @variable(model, x)
    y = (1 - 2im) * x
    @test !has_lower_bound(y)
    set_lower_bound(y, 1 + 2im)
```
What is `lower_bound(x)`? Maybe we should only support expressions that have `+/-1` or `+/-im` as coefficients.

mlubin:
My sense is that this might cause more confusion than it solves. Next we'll have a request to support `lower_bound` and `upper_bound` for nonlinear expressions, which is not trivial to implement.

blegat:
Setting bounds should only be allowed with one term and real coefficient. If complex, the user can be explicit by doing `set_lower_bound(imag(aff), ...)`

odow:
> My sense is that this might cause more confusion than it solves.

Yeah. But it's currently causing confusion because some outputs from `@variable` are actually `AffExpr`, and even something like `real(x)` is not a variable but an `AffExpr`.

> Setting bounds should only be allowed with one term and real coefficient

This is reasonable.

mlubin:
> Setting bounds should only be allowed with one term and real coefficient

That's a fair compromise.

odow:
How about this now

src/aff_expr.jl
```diff
@@ -809,3 +809,74 @@ moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
...
+
+In many cases, `@variable` can return a `GenericAffExpr` instead of a
+`GenericVariableRef`. This is particularly the case for complex-valued
+expressions. To make common operatons like `lower_bound(x)` work, we should
+forward the method if and only if `x` is convertable to a `GenericVariableRef`.
+"""
+function _eval_as_variable(f::F, x::GenericAffExpr, args...) where {F}
+    if length(x.terms) != 1
+        error(
+            "Cannot call $f with $x because it is not a real-valued affine " *
```
blegat, line 824: Not sure we need the `real-value`. the `+1` below is enough. Otherwise, the user might think the type needs to be `Real` while `one(ComplexF64)` actually works as coefficient. It's not about the type, it's just about being equal to `1`.

blegat: Ah we remove the complex below, never mind

test/test_expr.jl
```diff
@@ -451,4 +451,70 @@ function test_quadexpr_owner_model()
...
+    model = Model()
+    @variable(model, x[1:2, 1:2] in HermitianPSDCone())
+    @test_throws(
+        ErrorException(
+            "Cannot call $start_value with $(x[1, 1]) because it is not a real-valued affine " *
+            "expression of one variable with a coefficient of `+1`. Use " *
+            "`real(x)` or `imag(x)` to obtain the real and imaginary part and " *
+            "pass that instead.",
+        ),
+        start_value(x[1, 1]),
```
blegat, line 500: It may seem a bit weird that this does not work since the diagonal of a hermitian matrix is real

src/aff_expr.jl
```diff
@@ -809,3 +809,74 @@ moi_function(a::Vector{<:GenericAffExpr}) = MOI.VectorAffineFunction(a)
...
+        )
+    end
+    return f(variable, args...)
+end
+
+function _eval_as_variable(
+    f::F,
+    x::GenericAffExpr{T},
+    args...,
+) where {F,T<:Complex}
```
blegat, line 842: Maybe we can remove this method so that `set_start_value(one(ComplexF64)` * x, 1.0)` works. This would be useful for the diagonal of hermitian matrices which are real

odow:
Removed the method. How's the now

blegat (approved):

PR #3552, opened 2023-10-29
Add SkipModelConvertScalarSetWrapper

### LINKED ISSUE #3549: Make `_DoNotConvertSet` Public
We need this for DisjunctiveProgramming. We need to intercept the `add_constraint` call be dispatching on our expression type along with the set (`_DoNotConvertSet`), but this is not public. Sorry I failed to bring this up in #3530.

odow:
Closes #3549


pulsipher:
Thanks for this. I missed your comment on the issue asking for name suggestions. This name seems a little on the verbose side, but I don't really have a more concise suggestion.

blegat (approved):

odow:
Merging because I think we're okay with this.

Issue #3554, opened 2023-11-01
Provide better error message for Vector+VariableRef


odow:
https://discourse.julialang.org/t/methoderror-no-method-matching-zero-type-vector-affexpr/105639
```Julia
julia> using JuMP

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: NO_OPTIMIZER
Solver name: No optimizer attached.

julia> @variable(model, x[1:2])
2-element Vector{VariableRef}:
 x[1]
 x[2]

julia> @variable(model, y)
y

julia> @constraint(model, x + y .>= 0)
ERROR: MethodError: no method matching zero(::Type{Vector{VariableRef}})

Closest candidates are:
  zero(::Union{Type{P}, P}) where P<:Dates.Period
   @ Dates ~/.julia/juliaup/julia-1.9.3+0.x64.apple.darwin14/share/julia/stdlib/v1.9/Dates/src/periods.jl:51
  zero(::GenericQuadExpr)
   @ JuMP ~/.julia/dev/JuMP/src/quad_expr.jl:126
  zero(::AbstractVariableRef)
   @ JuMP ~/.julia/dev/JuMP/src/variables.jl:348
  ...

Stacktrace:
  [1] _instantiate_zero(#unused#::Type{Vector{VariableRef}})
    @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:28
  [2] promote_operation_fallback(op::typeof(+), #unused#::Type{Vector{VariableRef}}, #unused#::Type{VariableRef})
    @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:51
  [3] promote_operation(::typeof(+), ::Type, ::Type)
    @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:113
  [4] promote_operation_fallback(op::typeof(MutableArithmetics.add_mul), T::Type, args::Type)
    @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:99
  [5] promote_operation(::typeof(MutableArithmetics.add_mul), ::Type, ::Type)
    @ MutableArithmetics ~/.julia/dev/MutableArithmetics/src/interface.jl:113
  [6] mutability(::Type, ::Function, ::Type, ::Type)
[...]

PR #3557, opened 2023-11-02
Add errors for Matrix ± AbstractJuMPScalar

jd-foster:
Give a more informative error when there an operation between a single JuMP variable and a Matrix type object is attempted.

Closes #3553

Issue #3553, opened 2023-10-31
Multiplying JuMP expression with UniformScaling absorbs the matrix


blegat:
```julia
julia> @variable(model, x)

julia> rand(2, 2) - x * I
ERROR: MethodError: no method matching -(::Matrix{Float64}, ::AffExpr)

Closest candidates are:
  -(::RationalPoly, ::Any)
   @ MultivariatePolynomials ~/.julia/packages/MultivariatePolynomials/6TYc8/src/rational.jl:73
  -(::V, ::GenericAffExpr{C, V}) where {C, V<:AbstractVariableRef}
   @ JuMP ~/.julia/packages/JuMP/D44Aq/src/operators.jl:146
  -(::GenericQuadExpr, ::GenericAffExpr)
   @ JuMP ~/.julia/packages/JuMP/D44Aq/src/operators.jl:361
  ...

Stacktrace:
 [1] top-level scope
   @ REPL[39]:1
```
`x * I` cannot be a `UniformScaling` since the scaling should be a `Number` but and we cannot create a `Diagonal([x, x, ...])` since we don't know yet what is the dimension we'll need.

odow:
I don't know what we can or should do here. The answer is to force people to use `I(2)`.

odow:
I don't really want to add `MutableArithmetics.UniformScaling{<:AbstractMutable}` and have to deal with all the consequences?

blegat:
Yes it seems like a rabbit hole. Maybe we could try to convince LinearAlgebra to accept Any instead of Number. I think they thought that they are only implementing * with Number. This is fine but they can still allow Any in the constructor when it's called explicitly. In a Matrix they didn't force the entries to be Number so why do it in UniformScaling ?

odow:
Perhaps open an issue? This will be easier with upgradable stdlibs.

odow:
Closing as too hard to fix. It can't be a common issue, because I don't remember it coming up on Discourse.

jd-foster:
Could this give a better error message? For example,
```
Base.:-(a::Matrix, b::Union{JuMP.GenericVariableRef, JuMP.GenericAffExpr}) = error(
   "Subtraction between a Matrix and a JuMP variable is not supported: instead of A - x, " *
   "prefer A .- x for element-wise subtraction, or if you are modifying the diagonal entries of the matrix " *
   " do A - x * LinearAlgebra.I(n), where n is the diagonal length."
)
```

[...]

PR #3558, opened 2023-11-03
Improve error message for non-broadcasted addition and subtraction

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3554

Follow-up to #3557 

cc @jd-foster thoughts on this more general error?


jd-foster:
Looks much more general. I wonder about the related case that also prompted this, #3553, that it might not be obvious that `x * LinearAlgebra.I == 1.0 * x`. I'm not sure this can be covered here, but might be documented somewhere along the lines of the deleted comment "if you are modifying the diagonal entries of the matrix, do `A - x * LinearAlgebra.I(n)`, where `n` is the diagonal length."

odow:
I think in retrospect that supporting `UniformScaling` was the wrong choice. But yeah, I don't have a good answer for where we should document this. Putting it in an error message that probably happens when the user didn't write `I` isn't great either.

blegat:
We can restrict the error message to square matrices. I think it's still nice to include it. It shows why it's ambiguous to add a scalar and a matrix because both would be reasonable interpretations

odow:
How's this

blegat (approved):
I like it

PR #3560, opened 2023-11-03
Add EAGO to solvers

DimitriAlston:


docs/packages.toml
```diff
@@ -126,7 +126,8 @@
     user = "darnstrom"
     rev = "v0.5.1"
 # [EAGO]
-#   user = "PSORLab"
+#     user = "PSORLab"
+#     rev = "v0.8.1"
```
odow, line 130: The previous doc build failed because there are typos. I'll merge your change to the `(MI)` part, and then I'll open this PR under my account so that you can see the documentation preview and I'll explain what needs to change.

DimitriAlston: Thanks!

PR #3563, opened 2023-11-06
[docs] add tutorial on piecewise linear

odow:
preview: https://jump.dev/JuMP.jl/previews/PR3563/tutorials/linear/piecewise_linear/


odow:
I think this is a good addition. I can imagine linking quite a few discourse posts to it.

jd-foster:
This is nice; I've wanted to write something like this for a while but hadn't got around to it.

jd-foster:
1. Are you interested in solving the problems _and_ displaying the optimal values?
2. I find it a bit confusing to put outer approximation first up when it's about PWL functions - the OA stuff is 'implicit' PWL while the linear approximation is more 'explicit'.
3. What do you think about an example showing combination PWL constraint + exact function (`x^2 >= 0`) constraint to give a tight envelope?
4. Could reinforce that `SOS2` gets you _on_ the line by showing an optimal solution 'before' vs 'after' add the  `SOS2()` constraint.

odow:
See https://github.com/jump-dev/JuMP.jl/pull/3565 for docstrings

odow:
> I find it a bit confusing to put outer approximation first up when it's about PWL functions

Should we just call it `# Approximating nonlinear functions` then?

odow:
> What do you think about an example showing combination PWL constraint + exact function (x^2 >= 0) constraint to give a tight envelope?

I don't understand. How does one represent `x^2 >= 0`?

odow:
> Are you interested in solving the problems and displaying the optimal values?

I wasn't quite sure of a good way to do this. Ideally, you'd have some sort of interactive slider, where you could move `x` around, but we can't do that in a static environment.

jd-foster:
> I don't understand. How does one represent `x^2 >= 0`?

No idea why I wrote that: I meant the epi-graph`{t : t >= x^2 }`.

odow:
I still don't really follow what you're suggesting?

jd-foster:
Broadly speaking, I was thinking that you can combine both sections of the tutorial: a non-convex inner approximation boundary with a convex function (or concave, depending on the direction of the inequality) constraint to get a tight envelope to a function.
```
@constraint(model, y >= f(x))
y <= sum(λ[i] * ŷ[i] for i in 1:n)
    sum(λ) == 1
    λ in SOS2() 
```
But I'll just wait to read the whole thing again once it settles, or just PR later.

odow:
@jd-foster I merged because this is an improvement. But feel free to make other edits in a separate PR.

PR #3571, opened 2023-11-12
Convert constraint starting values

blegat:
When writing [this post](https://discourse.julialang.org/t/warm-start-using-dualization-with-jump-and-cosmo/106006/6?u=blegat), I found out when when starting integer dual starting values, COSMO was failing here:
https://github.com/oxfordcontrol/COSMO.jl/blob/dd6a6fdbda91231e4c3a108d072389e50a5003ac/src/interface.jl#L167
because `y0` does not have the right element type.
As we are doing the conversion for variable starting values, I thought we might need to do it for constraints as well:
https://github.com/jump-dev/JuMP.jl/blob/b13812d9f7cb5c0c3fa781c3e60400279a2d8907/src/variables.jl#L1685
Now, we might want to use `MOI.Utilities.value_type` as the value type of the function may not be the value type of the variables.

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -85,17 +85,6 @@ function dual_start_value(
     return reshape_vector(_dual_start(con_ref), dual_shape(con_ref.shape))
 end
 
-function _set_converted(
-    attr::MOI.AbstractConstraintAttribute,
-    con_ref::ConstraintRef{<:AbstractModel,<:MOI.ConstraintIndex{F}},
-    value,
-) where {F<:MOI.AbstractFunction}
-    model = owner_model(con_ref)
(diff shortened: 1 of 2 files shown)


odow (approved):
I simplified the implementation a little

src/constraints.jl
```diff
@@ -94,6 +94,10 @@ end
 
 _value_type(::Any, ::Any) = Any
 
+function _value_type(::ConstraintRef{M,<:MOI.ConstraintIndex{F}}) where {M,F}
+    return _value_type(M,F)
```
odow, line 98:
```suggestion
    return _value_type(M, F)
```

PR #3575, opened 2023-11-16
Truncate printing of large expressions

### LINKED ISSUE #3574: Truncate printing expressions if they are very large
x-ref https://discourse.julialang.org/t/outofmemoryerror-on-large-multicommodity-flow-model/106286

It's likely never a good solution to produce MBs of string data from the `show` method

odow:
Closes #3574 

I don't know if this is the right approach, but we should do something.

### DIFF
--- a/test/test_print.jl
+++ b/test/test_print.jl
@@ -977,13 +977,13 @@ function test_truncated_printing()
         "x_{30} + [[\\ldots\\text{940 terms omitted}\\ldots]] + x_{971}",
         function_string(MIME("text/latex"), y),
     )
-    ret = TERM_LIMIT_FOR_PRINTING[]
-    TERM_LIMIT_FOR_PRINTING[] = 3
+    ret = _TERM_LIMIT_FOR_PRINTING[]
+    _TERM_LIMIT_FOR_PRINTING[] = 3
     @test function_string(MIME("text/plain"), y) ==
           "x[1] + x[2] + [[...997 terms omitted...]] + x[1000]"
     @test function_string(MIME("text/latex"), y) ==
           "x_{1} + x_{2} + [[\\ldots\\text{997 terms omitted}\\ldots]] + x_{1000}"
-    TERM_LIMIT_FOR_PRINTING[] = ret
+    _TERM_LIMIT_FOR_PRINTING[] = ret
     return
 end
 
--- a/test/test_print.jl
+++ b/test/test_print.jl
@@ -967,4 +967,24 @@ function test_print_text_latex_interval_set()
     return
 end
 
+function test_truncated_printing()
+    model = Model()
+    @variable(model, x[1:1000])
+    y = sum(x)
+    s = function_string(MIME("text/plain"), y)
+    @test occursin("x[30] + [[...940 terms omitted...]] + x[971]", s)
+    @test occursin(
+        "x_{30} + [[\\ldots\\text{940 terms omitted}\\ldots]] + x_{971}",
+        function_string(MIME("text/latex"), y),
+    )
+    ret = TERM_LIMIT_FOR_PRINTING[]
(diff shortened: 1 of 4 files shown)


pulsipher:
It would be nice to also truncate terms in flattened nonlinear sums. Should be straightforward to add for expressions of the form: `NonlinearExpr(:+, [nlp_sum_terms...])`.

odow:
Truncated the nonlinear stuff. 

My main hesitation is the use of `TERM_LIMIT_FOR_PRINTING` as a global constant. But maybe that's okay. It's really a debugging flag. Maybe we should just document that extensions should not modify this, leave it to the user. And that extensions should never rely on the output of `function_string` remaining stable.

pulsipher:
> My main hesitation is the use of `TERM_LIMIT_FOR_PRINTING` as a global constant. But maybe that's okay. It's really a debugging flag. Maybe we should just document that extensions should not modify this, leave it to the user. And that extensions should never rely on the output of `function_string` remaining stable.

This seems reasonable from an extension point of view, so long as `function_string` doesn't have significant changes that impact testing frequently. In this case, having the default term limit be `60` is plenty large likely leave any testing that relies on printing unchanged. For InfiniteOpt, I don't think this will break any tests since the string checks are for small expressions.

odow:
@blegat thoughts?

blegat:
I think it's yet another proof that we shouldn't unify printing of JuMP and MOI. In MOI we have `_PrintOptions` and this would be a good candidate of additional option for `_PrintOption`. This would be much cleaner than modifying a global constant.

odow:
Should or shouldn't?

blegat:
Indeed, I meant should

odow:
Should we make it a private constant for now then? Merge this PR, and open an issue to unify the printing?

odow:
Made it a private constant

test/test_print.jl
```diff
@@ -967,4 +967,24 @@ function test_print_text_latex_interval_set()
...
+    model = Model()
+    @variable(model, x[1:1000])
+    y = sum(x)
+    s = function_string(MIME("text/plain"), y)
+    @test occursin("x[30] + [[...940 terms omitted...]] + x[971]", s)
+    @test occursin(
+        "x_{30} + [[\\ldots\\text{940 terms omitted}\\ldots]] + x_{971}",
+        function_string(MIME("text/latex"), y),
+    )
+    ret = _TERM_LIMIT_FOR_PRINTING[]
```
odow, line 980:
```suggestion
    ret = JuMP._TERM_LIMIT_FOR_PRINTING[]
```

test/test_print.jl
```diff
@@ -967,4 +967,24 @@ function test_print_text_latex_interval_set()
...
+    _TERM_LIMIT_FOR_PRINTING[] = 3
```
odow, line 981:
```suggestion
    JuMP._TERM_LIMIT_FOR_PRINTING[] = 3
```

test/test_print.jl
```diff
@@ -967,4 +967,24 @@ function test_print_text_latex_interval_set()
...
+    @test function_string(MIME("text/plain"), y) ==
+          "x[1] + x[2] + [[...997 terms omitted...]] + x[1000]"
+    @test function_string(MIME("text/latex"), y) ==
+          "x_{1} + x_{2} + [[\\ldots\\text{997 terms omitted}\\ldots]] + x_{1000}"
+    _TERM_LIMIT_FOR_PRINTING[] = ret
```
odow, line 986:
```suggestion
    JuMP._TERM_LIMIT_FOR_PRINTING[] = ret
```

odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/7009559319

PR #3577, opened 2023-11-20
[docs] document how to support operators with several vector arguments

### LINKED ISSUE #3576: Operator with vector arguments
I often have questions on how to define an nonlinear operator with several vector inputs.
We have the splatting section:
https://jump.dev/JuMP.jl/dev/manual/nonlinear/
but it doesn't explain what to do with multiple inputs.
I would probably write something in the doc like:
If you have a function `f(x::Vector, y::Vector, z::Real)` of length `nx` and `ny`, you can do
```julia
@operator(model, op_f, nx + ny + 1, (z, args...) -> f(collect(args[1:nx]), collect(args[nx .+ (1:ny)]), z))
```

odow:
Closes #3576

### DIFF
--- a/docs/src/manual/nonlinear.md
+++ b/docs/src/manual/nonlinear.md
@@ -556,6 +556,22 @@ f(x::Vector) = sum(x[i]^i for i in 1:length(x))
 @objective(model, Min, op_f(x...))
 ```
 
+If the operator takes several vector inputs, write a function which takes the
+splatted arguments and reconstructs the required vector inputs:
+```@repl
+using JuMP
+model = Model();
+@variable(model, x[1:2]);
+@variable(model, y[1:2]);
+@variable(model, z);
+f(x::Vector, y::Vector, z) = sum((x[i] * y[i])^z for i in 1:2)
+f(x, y, z)
+f_splat(args...) = f(collect(args[1:2]), collect(args[3:4]), args[5])
+f_splat(x..., y..., z)
+@operator(model, op_f, 5, f_splat)
+@objective(model, Min, op_f(x..., y..., z))
+```
+
 ### Automatic differentiation
 
 JuMP does not support black-box optimization, so all user-defined operators must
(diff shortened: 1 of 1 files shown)


odow:
@blegat is this what you wanted:

https://jump.dev/JuMP.jl/previews/PR3577/manual/nonlinear/#User-defined-operators-with-vector-inputs

docs/src/manual/nonlinear.md
```diff
@@ -556,6 +556,22 @@ f(x::Vector) = sum(x[i]^i for i in 1:length(x))
 @objective(model, Min, op_f(x...))
 ```
 
+If the operator takes several vector inputs, write a function which takes the
```
blegat, line 559:
```suggestion
If the operator takes several vector inputs, write a function that takes the
```

blegat: My understanding is that we use "that" when it's a statement that additionally constrains the work and to use "which" when it is another information about it but it doesn't change the set of possibilities.

blegat:
It might be more helpful to show how to deal with functions that cannot be traced and to show how to also pass the derivative info

odow:
Is this part not sufficient? https://jump.dev/JuMP.jl/stable/manual/nonlinear/#Multivariate-functions

blegat:
It doesn't say that you can use AD

odow:
I'd rather not encourage people. We can already use ForwardDiff automatically. If anyone has a user-defined operator and wants to use a custom AD that isn't ForwardDiff, then I expect them to be very advanced users who can figure it out based on the signature of `grad_f(g::AbstractVector{T}, x::T...) where {T}`.

If we tell people to use Enzyme/Zygote/ReverseDiff.jl then we'll have to deal with the questions and quirks.

odow:
This is good enough for now. A question of whether to add a third-party AD example is best left for a different PR. I'll make a note in the issue.

PR #3589, opened 2023-11-27
Maybe improve `installation.md` by showing package's "popularity"

LebedevRI:
For me at least, picking the "right" solver is still a but daunting.
It would be nice to be actually able to filter by features (i mean,
other than Ctrl+F), but asides from that, perhaps showing
the "popularity" of each particular Julia package might make choosing easier?

This is how it looks: https://github.com/LebedevRI/JuMP.jl/blob/installation.md/docs/src/installation.md#supported-solvers

If this looks too ugly, or is not wanted, please just close the PR :)

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -98,57 +98,107 @@ use, so installation is often more complex.
 Solvers with a missing entry in the `Julia Package` column are written in Julia.
 The link in the `Solver` column is the corresponding Julia package.
 
-| Solver                                                                         | Julia Package                                                                    | Installation | License | Supports             |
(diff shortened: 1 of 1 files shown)


odow:
I think we need to be quite careful here. I agree that JuMP could make it easier to suggest solvers for a particular problem class, but we don't really want to pick winners.

One thing we do have is the README for each solver: https://jump.dev/JuMP.jl/stable/packages/solvers/

If we wanted to make the table a little easier to read, we could consider dropping the `solver` column in favor of just the Julia package and include the organization. So `jump-dev/HiGHS.jl`. But I don't know what we'd do with multiple solvers in the same package, like AmplNLWriter.

odow:
Perhaps to clarify though: I'm very open to suggestions for changes. I don't particularly like the table or think that it is very helpful...

PR #3590, opened 2023-11-29
Fix type instability in `set_objective_coefficient`

joaquimg:
This comparison is a call to `isa` that was giving me a performance hit on `set_objective_coefficient`.


odow:
Looks like this goes back as far as https://github.com/jump-dev/JuMP.jl/pull/2008.

We could just fallback to `MOI.modify` for all objective types? I don't see why we need to throw the unsupported error?

PR #3596, opened 2023-12-06
[docs] add DocumenterCitations

blegat:
I always found it quite difficult and time consuming to manage citations and ensure a uniform look. Just found this [DocumenterCitations](https://github.com/JuliaDocs/DocumenterCitations.jl) which seems promising so let's see how it looks. I started with just a few citations as a proof of concept.

Preview https://jump-dev.github.io/JuMP.jl/previews/PR3596


goerz:
I have a vague memory of having seen something like that before… I think it might have been if you forget to have a (canonical) bibliography block? That is, there’s nothing for the reference to link to?

Are you using the latest version? Because I also recall doing something do give a clear error message in such a situation (or maybe it was just something I wanted to do and then forgot to follow up on)

docs/src/index.md
```diff
@@ -122,3 +122,8 @@ and the [Mozilla FAQ](https://www.mozilla.org/en-US/MPL/2.0/FAQ/) for more
 information. In addition, JuMP is typically used in conjunction with solver
 packages and extensions which have their own licences. Consult their package
 repositories for the specific licenses that apply.
+
+## Bibliography
+
+```@bibliography
+```
```
odow, line 129: Here's the canonical block

odow:
> Are you using the latest version?

Yip. We're using 
```
Installed DocumenterCitations ────────── v1.3.2
```

goerz:
Hard to say… I tried to reproduce it locally after checking out that PR branch, with

```
julia --project=docs
(docs) pkg> dev .
(docs) pkg> instantiate
include("docs/make.jl")
```

but that never gets to the error with the bibliography, as it fails with during the doctests/tutorials with

```
...
Terminated with status = numerical error
solve time =  165ms
/Users/goerz/Documents/Programming/JuMP.jl/docs/src/tutorials/applications/optimal_power_flow.jl: Test Failed at /Users/goerz/Documents/Programming/JuMP.jl/docs/src/tutorials/applications/optimal_power_flow.jl:420
```

Do you have any special instructions for building to the documentation? I was trying this on my MacBook; should I use my Linux workstation instead?

(I'd maybe recommend creating a `Makefile` so that `make docs`, `make tests`, etc. works)

If I can reproduce it, I'll see if I can figure out what's going on with the citation.

odow:
What if you run with `ENV["CI"] = "true"` to that it runs with `prettyurls`

https://github.com/jump-dev/JuMP.jl/blob/cecc5ba11e0c13f2860550eb4f3b734257d804e2/docs/make.jl#L557

> As a matter of style, I would probably recommend putting the Bibliography on a separate page at the end of the documentation

Yeah will do. @blegat was just trying this as a proof of concept.

goerz:
Sure, I'll try the `CI=true`. If it fails in some weird way, I'd also recommend temporarily activating the DEBUG logging in the workflow: https://juliadocs.org/DocumenterCitations.jl/stable/internals/#Debugging

odow:
For some reason https://github.com/jump-dev/JuMP.jl/pull/3596/commits/50c3859378e31b801e6dedf5a1a4dd9f74c08f87 fixed it...

docs/src/background/bibliography.md
```diff
@@ -0,0 +1,4 @@
+## Bibliography
```
odow, line 1:
```suggestion
#
```

blegat:
> For some reason [50c3859](https://github.com/jump-dev/JuMP.jl/commit/50c3859378e31b801e6dedf5a1a4dd9f74c08f87) fixed it...

:open_mouth: I guess it's because we shouldn't use the same plugin instance for the pdf and the html, good catch

odow (approved):
LGTM. Can fix the various other references over time.

goerz:
> Just tried one with an accent with `Matousek2013`. I assume that for the LaTeX -> pdf build to work, we should escape the accent so I used `{\v{s}}` 

The escaped version is fully supported, but Unicode should work as well, both for HTML and TeX/PDFs. I’ve been using unicode in my .bib files for years, and haven’t had any problems. Officially, bibtex is still “ascii-only”, so I don’t know *why* it works, but any installation of LaTeX on Linux and macOS I’ve used since at least 2015 just does the right thing.

blegat:
@goerz Thanks! I guess both can be used then

PR #3597, opened 2023-12-06
Update `extension-tests` workflow to include DisjunctiveProgramming

pulsipher:
DisjunctiveProgramming has now reached a fairly mature state as an extension.


odow:
I'll also add it to the docs

PR #3598, opened 2023-12-06
[docs] add DisjunctiveProgramming

odow:
Probably going to need a few fixes 

cc @pulsipher 
cc @hdavid16 

https://jump.dev/JuMP.jl/previews/PR3598/packages/DisjunctiveProgramming/

### DIFF
--- a/docs/packages.toml
+++ b/docs/packages.toml
@@ -127,7 +127,7 @@
     rev = "v0.5.2"
 [DisjunctiveProgramming]
     user = "hdavid16"
-    rev = "5d814e9023acf25672977632bc9dc111d68ea8ca"
+    rev = "6a4d0ac4a7484e52b1dbff9cee310a73d2d23e81"
     filename = "docs/jump/README.md"
     extension = true
 [EAGO]
--- a/docs/packages.toml
+++ b/docs/packages.toml
@@ -127,7 +127,8 @@
     rev = "v0.5.2"
 [DisjunctiveProgramming]
     user = "hdavid16"
-    rev = "v0.5.0"
+    rev = "b71f715ba826fa63677da9590ae41b41ca5024e6"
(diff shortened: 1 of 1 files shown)


pulsipher:
Thanks @odow, I suspected this wouldn't quite work. I think the best way forward would be for me to make a JuMP specific readme like I did for InfiniteOpt. I should have time to do that this week.

joaquimg:
Another option would be to cleanup the readme since most of it is replicated in the docs.

pulsipher:
> Another option would be to cleanup the readme since most of it is replicated in the docs.

This along with improving the docs is on my TODO list, I just need to find the time :)

pulsipher:
I think I fixed the vale errors

docs/packages.toml
```diff
@@ -125,6 +125,11 @@
 [DAQP]
     user = "darnstrom"
     rev = "v0.5.2"
+[DisjunctiveProgramming]
+    user = "hdavid16"
+    rev = "b71f715ba826fa63677da9590ae41b41ca5024e6"
```
odow, line 130:
```suggestion
    rev = "5d814e9023acf25672977632bc9dc111d68ea8ca"
```

docs/packages.toml
```diff
@@ -125,6 +125,11 @@
 [DAQP]
     user = "darnstrom"
     rev = "v0.5.2"
+[DisjunctiveProgramming]
+    user = "hdavid16"
+    rev = "5d814e9023acf25672977632bc9dc111d68ea8ca"
```
odow, line 130:
```suggestion
    rev = "6a4d0ac4a7484e52b1dbff9cee310a73d2d23e81"
```

odow:
Merging because we can always update later. I don't see the typos @pulsipher is talking about.

Issue #3513, opened 2023-09-18
Refactor JuMP macros


odow:
We have a roadmap item for this: https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/developers/roadmap.md, but it'd be useful to have an issue tracking it (x-ref https://github.com/jump-dev/JuMP.jl/pull/3500#issuecomment-1716511987).

I don't have any suggestions for what a "good" `macros.jl` file would look like, but it would be cleaner, with less code being run in the macros, and no user-facing changes.

chriscoey:
Maybe precluded by the "no user-facing changes", but it would be great if macro changes could make it easier to use JuMP "programmatically". Possibly that could be a separate roadmap item. I don't know what needs to be done really because I'm no expert on macros etc, but I have had to resort to generating low level MOI code to build complex models programmatically.

odow:
I think the non-macro version is just a documentation issue. We already have all of the machinery available in JuMP.

Given:
```julia
using JuMP
model = Model()
@variable(model, 0 <= x[1:I, 1:T] <= 100)
@constraint(model, [i in 1:I, t in 2:T], x[i, t] - x[i, t-1] <= 10)
@constraint(model, [i in 1:I, t in 2:T], x[i, t] - x[i, t-1] >= -10)
@objective(model, Min, sum(x))
```

One version of the non-macro version is
```julia
model = Model()
x = [VariableRef(model) for i in 1:I, t in 1:T]
set_lower_bound.(x, 0)
set_upper_bound.(x, 1000)
for i in 1:I, t in 2:T
    add_constraint(model, ScalarConstraint(x[i, t] - x[i, t-1], MOI.LessThan(10.0)))
    add_constraint(model, ScalarConstraint(x[i, t] - x[i, t-1], MOI.GreaterThan(-10.0)))
end
set_objective(model, MOI.MIN_SENSE, sum(x))
```

It's also possible to minimally use the macros:
```julia
model = Model()
x = [@variable(model) for i in 1:I, t in 1:T]
set_lower_bound.(x, 0)
set_upper_bound.(x, 1000)
for i in 1:I, t in 2:T
    f, set = x[i, t] - x[i, t-1], MOI.LessThan(10.0)
    @constraint(model, f in set)
    f, set = x[i, t] - x[i, t-1], MOI.GreaterThan(-10.0)
    @constraint(model, f in set)
end
sense, f = MOI.MIN_SENSE, sum(x)
[...]

PR #3601, opened 2023-12-07
Fix error for unsupported objective sense

odow:
Apparently no one has ever noticed that this error is wrong.

Part of #3513

### DIFF
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -2131,4 +2131,18 @@ function test_issue_3514()
     return
 end
 
+function test_bad_objective_sense()
+    model = Model()
+    @variable(model, x)
+    @test_macro_throws(
+        ErrorException(
+            "In `@objective(model, :MinMax, x)`: unexpected sense `MinMax`. " *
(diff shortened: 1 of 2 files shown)


test/test_macros.jl
```diff
@@ -2131,4 +2131,18 @@ function test_issue_3514()
...
 end
 
+function test_bad_objective_sense()
+    model = Model()
+    @variable(model, x)
+    @test_macro_throws(
+        ErrorException(
+            "In `@objective(model, :MinMax, x)`: unexpected sense `MinMax`. " *
+            "The sense must be an `::MOI.OptimizatonSense`, or the symbol " *
+            "`:Min` or `:Max`."
```
odow, line 2141:
```suggestion
            "`:Min` or `:Max`.",
```

test/test_macros.jl
```diff
@@ -2131,4 +2131,18 @@ function test_issue_3514()
     return
 end
 
+function test_bad_objective_sense()
+    model = Model()
+    @variable(model, x)
+    @test_macro_throws(
```
odow, line 2137:
```suggestion
    @test_throws_strip(
```

PR #3602, opened 2023-12-07
WIP: refactor src/macros.jl for clarity and readability

### LINKED ISSUE #3513: Refactor JuMP macros

odow:
Closes #3513

This diff is bad, mainly because I've been moving things around. I don't really have a clear idea where we're headed, so this is just an exploration, and once I've settled on a nicer design we can figure out how to make a proper PR.


src/macros.jl
```diff
@@ -1603,106 +1595,19 @@ Subject to
...
-julia> value(x)
-10.0
-```
- """ :(@NLparameters)
-
-@doc """
-    @NLconstraints(model, args...)
+"""
+macro constraints(model, block)
+    return _plural_macro_code(model, block, Symbol("@constraint"))
```
odow, line 1600: this is a useful change. It gets rid of `eval` and no more floating docstrings.

odow: See https://github.com/jump-dev/JuMP.jl/pull/3603

odow:
Closing. I've started splitting this into other PRs.

PR #3607, opened 2023-12-08
Refactor src/macros.jl into separate files

odow:
Part of #3513

`src/macros.jl` is pretty overwhelming, partly because it is so long.

This PR beings to break up the code into separate files for each macro.

Longer-term, I wonder if it makes sense to move the various implementations of `build_variable` etc to the `macros/variable.jl` file.
https://github.com/jump-dev/JuMP.jl/blob/280f5fb53b994474a37628c19bbace4de57a827b/src/sd.jl#L302-L424
There's currently a bit of everything all over the place.


blegat (approved):

odow:
Once this is merged, we can start cleaning up each macro one-by-one. It should be a bit more manageable.

PR #3609, opened 2023-12-08
Fix text/latex printing of GenericNonlinearExpr

### LINKED ISSUE #3604: Improve LaTeX Printing for `GenericNonlinearExpr`s
While adding better printing to DisjunctiveProgramming, I noticed some shortcomings with the LaTeX printing of `GenericNonlinearExpr`s. 

First, indexed variables such as `x[1]` are printed as `x[1]` and not `x_{1}`, since `function_string` is not called:
```julia
m = Model(); @variable(m, x[1:2])
@expression(m, x[1]^2.3)
```
image

It would also be nice to use LaTeX versions of operators where possible, at least for the more common ones. For example, it would be nice to use `\\leq` instead of `<=`:
```julia
@expression(m, x[1] <= x[2])
```
image

This can even lead to problems for certain operators (e.g., `&&`):
```julia
@expression(m, x[1] && x[2])
```
image
Here, `&&` needs to be replaced with `\\wedge` to parse correctly.

odow:
Closes #3604

### DIFF
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -147,12 +147,31 @@ end
 _parens(::MIME) = "(", ")", "", "", ""
 _parens(::MIME"text/latex") = "\\left(", "\\right)", "{", "}", "\\textsf"
 
+_op_to_string(::MIME, op) = op
+
+function _op_to_string(::MIME"text/latex", op::Symbol)
+    if op == :&&
+        return "\\wedge"
+    elseif op == :||
+        return "\\vee"
+    elseif op == :(<=)
+        return "\\le"
+    elseif op == :(>=)
+        return "\\ge"
+    elseif op == :(==)
+        return "="
+    else
+        return string(op)
+    end
+ end
+
 function function_string(mime::MIME, x::GenericNonlinearExpr)
     p_left, p_right, p_open, p_close, p_textsf = _parens(mime)
     io, stack = IOBuffer(), Any[x]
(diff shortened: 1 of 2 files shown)


pulsipher:
Thanks for the quick turnaround.

src/nlp_expr.jl
```diff
@@ -147,12 +147,31 @@ end
...
+        return "\\le"
+    elseif op == :(>=)
+        return "\\ge"
+    elseif op == :(==)
+        return "="
+    else
+        return string(op)
+    end
+ end
+
```
pulsipher, line 167: This might be asking for too, but it would be nice to have some extensible control of how operators are printed. In DisjunctiveProgramming I would like to use the correct LaTeX representations of logical operators that we introduce (e.g., `\\neg`, `\\implies`). Currently, I hack around this via `replace`: https://github.com/hdavid16/DisjunctiveProgramming.jl/blob/d7e8e7a09b1cb39d628b25d9a0cded2345c9664d/src/print.jl#L74-L100. In this case, we have `GenericNonlinearExpr{LogicalVariableRef}` so I could overload `JuMP.function_string`, but that would duplicate a lot of the same code.

odow: How's this now

pulsipher: Perfect, thanks!

blegat (approved):

PR #3612, opened 2023-12-09
Tidy macros/@constraint.jl

odow:
Part of #3513

### DIFF
--- a/src/macros/@constraint.jl
+++ b/src/macros/@constraint.jl
@@ -258,8 +258,14 @@ function _constraint_macro(
     code = if is_vectorized
         quote
             $parse_code
-            build = model_convert.($model, $build_call)
-            add_constraint.($model, build, $name_expr)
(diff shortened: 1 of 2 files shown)


src/macros/@constraint.jl
```diff
@@ -192,33 +191,24 @@ called from in the user's code. One way of generating this is via the hidden
...
-            error_fn("Not enough arguments")
-        end
-    end
-    model = esc(pos_args[1])
-    y = pos_args[2]
-    extra = pos_args[3:end]
-    if Meta.isexpr(args[2], :block)
+    error_fn(str...) = _macro_error(macro_name, input_args, source, str...)
+    args, kwargs, container = Containers._extract_kw_args(input_args)
+    if length(args) < 2 &&!isempty(kwargs)
```
odow, line 201:
```suggestion
    if length(args) < 2 && !isempty(kwargs)
```

PR #3616, opened 2023-12-10
Add parse_macro_arguments to unify how we handle macro inputs

odow:
Part of #3513 

The main benefit of this is that it provides a public API via `Containers.parse_macro_arguments`, and it deals with the `parameter` stuff that happens when you have `;` keywords in a macro.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -323,37 +323,6 @@ function model_convert(
     return model_convert.(model, x)
 end
 
-"""
-    _add_kw_args(call, kw_args)
-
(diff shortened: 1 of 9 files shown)


src/macros.jl
```diff
@@ -375,7 +354,7 @@ julia> call
 :(f(1, $(Expr(:escape, :x)), a = 2))
 ```
 """
-function _add_positional_args(call, args)
+function _add_positional_args(call::Expr, args::Vector{Any})
```
odow, line 357:
```suggestion
function _add_positional_args(call::Expr, args::Vector)
```

Issue #3622, opened 2023-12-11
Performance downgrade in solving optimal control problems(OCP)


johnysyumich:
Hi developers,

We are currently "new" users to JuMP but we do have experience in using JuMP since we were users of NLOptControl: https://github.com/JuliaMPC/NLOptControl.jl. This package relies very heavily on JuMP and stopped upgrading about three years ago. We are thinking of upgrading it to let it compile with the current version of JuMP. The last version of NLOptControl uses JuMP v0.18.6 which is the version right before the change from MPB to MOI. However, we met some performance issues when we were upgrading it, and here is one of the dummy examples we created to benchmark with the old package.

Here is the code we run:
[example.zip](https://github.com/jump-dev/JuMP.jl/files/13641695/example.zip)

Basically, it is controlling a modeled vehicle to change to another lane.

On my computer (I7-12700K), the algorithm can solve the problem with around 130ms. However, with the old implementation (NLOptControl & JuMP v0.18.6) we can achieve around 45ms per optimization on the same computer. With our current modifications, we cannot achieve the same performance as we could before. It would be very helpful if you would provide any insights about the difference between performance and how we can modify the current dummy code to work as fast as the previous version.

Thanks for your help in advance!

odow:
I can't run your code because I don't know what `PCY1` is.

As a general comment, JuMP v0.18.6 was released over four years ago. In the last few years we have made big improvements to our nonlinear code, so that the expression and `eval` stuff is not needed. See https://jump.dev/JuMP.jl/stable/manual/nonlinear/ and tutorials like https://jump.dev/JuMP.jl/stable/tutorials/nonlinear/rocket_control/. I should write an explicit MPC tutorial (x-ref https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-1839210372).

You can ask for support and suggested code improvements on http://jump.dev/forum. We try to keep the GitHub issues for bug reports.

Here's how I'd write your code. I can't test it, but it should point you in the right direction:
```julia
using JuMP, Ipopt

KZX     = 289.5
KZYR    = 427.0
KZYF    = 277.2
FzF0    = 2670.0
FzR0    = 2670.0
PC1     = PCY1
PD1     = PDY1 - PDY2
PD2     = PDY2/FZ0
PE1     = PEY1 - PEY2
PE2     = PEY2/FZ0
PE3     = PEY3
PK1     = PKY1*FZ0
PK2     = 1/(PKY2*FZ0)
[...]

PR #3623, opened 2023-12-12
[docs] update should_i_use.md to remove optimal control

odow:
Thoughts? Especially @pulsipher.

With tracing, we have a nice interface now. I don't see any reason to recommend people go elsewhere.

Here's an example of a model that is now much nicer to write https://github.com/jump-dev/JuMP.jl/issues/3622

### DIFF
--- a/docs/src/should_i_use.md
+++ b/docs/src/should_i_use.md
@@ -95,21 +95,6 @@ Alternatives to consider are:
  * [Optimization.jl](https://github.com/SciML/Optimization.jl)
  * [NLopt.jl](https://github.com/JuliaOpt/NLopt.jl)
 
-### Optimal control problems
-
-JuMP supports formulating optimal control problems as large nonlinear programs
-(see, for example, [Optimal control for a Space Shuttle reentry trajectory](@ref)).
-However, the nonlinear interface has a number of limitations (for example, the
-need to write out the dynamics in algebraic form) that mean JuMP might not be
-the right tool for the job.
-
-Alternatives to consider are:
-
- * [CasADi [MATLAB/Python]](https://web.casadi.org),
-   [CasADi.jl](https://github.com/ichatzinikolaidis/CasADi.jl)
- * [InfiniteOpt.jl](https://github.com/infiniteopt/InfiniteOpt.jl)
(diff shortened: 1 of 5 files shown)


joaquimg:
Maybe instead of removing the section completely, keep it, remove the references to other tools and add a link to a tutorial.

odow:
I don't know about keeping the links here. We don't provide links to alternatives for other problem types that JuMP supports. This section is in the "Why shouldn't I use JuMP" bit. And solving an optimal control problem is not a reason not to use JuMP.

I have a TODO to add a MPC tutorial: https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-1839210372

People seem to be using (and enjoying) JuMP for optimal control problems:
https://discourse.julialang.org/t/setting-up-free-final-time-in-jump-optimization/107302/3

pulsipher:
Naturally, I like having the reference, but I understand the argument to remove it from the why not to use JuMP section. 

Perhaps an alternative suggestion would be to add some references in the tutorials to relevant JuMP extensions (e.g., InfiniteOpt, SDDP, BilevelJuMP). For instance, it is straightforward to implement an optimal control problem that is discretized with a fixed time step that uses explicit/implicit Euler as demonstrated by existing/future JuMP tutorials; however, using a more complex discretization scheme (e.g., orthogonal collocation over finite elements) is nontrivial to implement. Thus, referring the user to InfiniteOpt to handle these more difficult cases would make sense and highlight that JuMP ecosystem has several useful extensions.

odow:
Added some suggestions https://github.com/jump-dev/JuMP.jl/pull/3625.

I think they are a much better way of suggesting relevant extensions. I doubt many people actually read the "when should I not use JuMP" and then followed the link to an extension.

odow:
Updated to refer to InfiniteOpt and BilevelJuMP in the tutorials.

odow:
See

https://jump.dev/JuMP.jl/previews/PR3623/tutorials/nonlinear/rocket_control/
https://jump.dev/JuMP.jl/previews/PR3623/tutorials/nonlinear/nested_problems/

pulsipher:
I approve of the changes and support this PR. I will also note that InfiniteOpt supports 2-stage stochastic programs like https://jump.dev/JuMP.jl/stable/tutorials/applications/two_stage_stochastic/.

PR #3640, opened 2023-12-21
[docs] update to GA-4

odow:
This won't fix all of the existing HTML files in gh-pages branch. I'll need to fix that manually.

I've tested that this is all we need to do in SDDP.jl https://github.com/odow/SDDP.jl/pull/723

x-ref https://github.com/JuliaDocs/Documenter.jl/issues/2016


odow:
Working
image

PR #3641, opened 2023-12-26
[docs] add modulo to tips_and_tricks.jl

odow:
This comes up reasonably frequently: https://discourse.julialang.org/t/round-off-decision-variable/107988/3

### DIFF
--- a/docs/src/tutorials/linear/tips_and_tricks.jl
+++ b/docs/src/tutorials/linear/tips_and_tricks.jl
@@ -114,6 +114,28 @@ model = Model();
 
 # This reformulation does not work for ``t \le \max\{x, y\}``.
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/linear/tips_and_tricks.jl
```diff
@@ -114,6 +114,28 @@ model = Model();
 
 # This reformulation does not work for ``t \le \max\{x, y\}``.
 
+# ## Modulo
+
+# To model ``y = x % n``, where ``n`` is a constant modulus, we use the
```
odow, line 119:
```suggestion
# To model ``y = x \text{ mod } n``, where ``n`` is a constant modulus, we use the
```

PR #3642, opened 2023-12-26
Fix converting (Aff,Quad)Expr to NonlinearExpr

odow:
x-ref https://discourse.julialang.org/t/unable-to-create-vector-out-of-affexpr-in-some-cases/107968


odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/7333357911

odow:
This can change the return type, but it's a bug fix.

PR #3644, opened 2023-12-28
[docs] Add SDPLR to list of solvers

blegat:
I'm already opening this PR even if it's not released to see if the README passes the tests

## Basic

 - [x] Check that the solver is a registered Julia package, needs https://github.com/blegat/SDPLR.jl/pull/19
       documented.
       use the solver with JuMP

## Documentation

## Optional

cc @sburer


odow:
I wonder if we should move it to `jump-dev` before tagging.

docs/packages.toml
```diff
@@ -84,6 +84,8 @@
     rev = "v2.0.0"
 [SDPA]
     rev = "v0.5.0"
+[SDPLR]
+    rev = "29552b8da0982d61fe286f73d11f766cfb7361d6"
```
odow, line 88:
```suggestion
    rev = "v0.1.0"
```

Once tagged...

PR #3653, opened 2024-01-15
Throw informative error when parsing reference sets in macros

### LINKED ISSUE #3652: Diagnostics request: using variable in variable container bounds
```
using JuMP, SCIP

model = Model(SCIP.Optimizer)

max_a = 10

@variable(model, a, Int)

@variable(model, b[1:a], Int) 
# Should have been: @variable(model, b[1:max_a], Int) 
```
```
MethodError: no method matching (::Colon)(::Int64, ::VariableRef)

Closest candidates are:
  (::Colon)(::T, ::Any, ::T) where T<:Real
   @ Base range.jl:50
  (::Colon)(::A, ::Any, ::C) where {A<:Real, C<:Real}
   @ Base range.jl:10
  (::Colon)(::T, ::Any, ::T) where T
   @ Base range.jl:49
  ...


Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/JuMP/027Gt/src/macros.jl:375 [inlined]
 [2] top-level scope
   @ In[2]:6
```

I understand that this does not work, i meant to use `max_a` instead,
but the diagnostic was not as helpful. Might it be possible to improve it perhaps?

odow:
Closes #3652 

Error from #3652 is now:
```Julia
julia> using JuMP

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC

julia> @variable(model, a, Int)
a

julia> @variable(model, b[1:a], Int) 
ERROR: At REPL[67]:1: `@variable(model, b[1:a], Int)`: unexpected error parsing reference set: 1:a
Stacktrace:
 [1] error(::String, ::String, ::Expr)
   @ Base ./error.jl:44
 [2] (::JuMP.Containers.var"#error_fn#98"{String})(::String, ::Vararg{Any})
   @ JuMP.Containers ~/.julia/dev/JuMP/src/Containers/macro.jl:315
 [3] macro expansion
   @ ~/.julia/dev/JuMP/src/Containers/macro.jl:92 [inlined]
 [4] macro expansion
   @ ~/.julia/dev/JuMP/src/macros.jl:375 [inlined]
 [5] top-level scope
   @ REPL[67]:1

caused by: MethodError: no method matching (::Colon)(::Int64, ::VariableRef)

Closest candidates are:
  (::Colon)(::T, ::Any, ::T) where T<:Real
   @ Base range.jl:50
  (::Colon)(::A, ::Any, ::C) where {A<:Real, C<:Real}
   @ Base range.jl:10
  (::Colon)(::T, ::Any, ::T) where T
   @ Base range.jl:49
  ...

Stacktrace:
 [1] macro expansion
   @ ~/.julia/dev/JuMP/src/Containers/macro.jl:90 [inlined]
 [2] macro expansion
   @ ~/.julia/dev/JuMP/src/macros.jl:375 [inlined]
 [3] top-level scope
   @ REPL[67]:1
```

cc @LebedevRI

### DIFF
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -2254,4 +2254,70 @@ function test_escaping_of_set_kwarg()
     return
 end
 
+function test_error_parsing_reference_sets()
+    model = Model()
+    @variable(model, a)
+    @test_throws_runtime(
+        ErrorException(
+            "In `@variable(model, b[1:a])`: unexpected error parsing reference set: 1:a",
+        ),
+        @variable(model, b[1:a]),
+    )
+    @test_throws_runtime(
+        ErrorException(
+            "In `@variable(model, b[1:2, 1:a])`: unexpected error parsing reference set: 1:a",
+        ),
+        @variable(model, b[1:2, 1:a]),
+    )
+    @test_throws_runtime(
+        ErrorException(
+            "In `@variable(model, b[i = 1:a, 1:i])`: unexpected error parsing reference set: 1:a",
+        ),
+        @variable(model, b[i = 1:a, 1:i]),
+    )
+    @test_throws_runtime(
+        ErrorException(
+            "In `@variable(model, b[i = 1:2, a:i])`: unexpected error parsing reference set: a:i",
+        ),
+        @variable(model, b[i = 1:2, a:i]),
+    )
+    @test_throws_runtime(
+        ErrorException(
+            "In `@variable(model, b[i = 1:2; i < a])`: unexpected error parsing condition: i < a",
+        ),
+        @variable(model, b[i = 1:2; i < a]),
+    )
+    @test_throws_runtime(
+        ErrorException(
+            "In `@expression(model, b[1:a], a + 1)`: unexpected error parsing reference set: 1:a",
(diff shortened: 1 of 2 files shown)


test/test_macros.jl
```diff
@@ -2254,4 +2254,70 @@ function test_escaping_of_set_kwarg()
...
+        @expression(model, b[i = 1:2, a:i], a + 1),
+    )
+    @test_throws_runtime(
+        ErrorException(
+            "In `@expression(model, b[i = 1:2; i < a], a + 1)`: unexpected error parsing condition: i < a",
+        ),
+        @expression(model, b[i = 1:2; i < a], a + 1),
+    )
+    return
+end
```
LebedevRI, line 2321: This tests both `@variable` and `@expression`.
Should `@constraint` also be tested?

LebedevRI:
Nice! Thank you!

test/test_macros.jl
odow: It also uses the same code paths. but added anyway

PR #3658, opened 2024-01-23
Add support for modifying quadratic coefficients

### LINKED ISSUE #3657: Add support for MOI.ScalarQuadraticCoefficientChange
MOI v1.21 added `MOI.ScalarQuadraticCoefficientChange`: https://github.com/jump-dev/MathOptInterface.jl/pull/2296

We should expose this at the JuMP level, but maybe wait for https://github.com/jump-dev/HiGHS.jl/pull/195 to be merged.

odow:
Closes #3657

Looks like there's an issue in MOI normalizing the difference between `x[1] * x[2]` and `x[2] * x[1]`.

TODO:

 - [x] Wait for https://github.com/jump-dev/MathOptInterface.jl/pull/2410 and update Project.toml` to contain min 1.25.2 so we can't use this with the buggy MOI


odow:
Merging because I don't think this is controversial, and because this is blocking a refactor of some related things.

PR #3666, opened 2024-02-05
Bump compat of DimensionalData

odow:
So it actually took all of 2 months after merging https://github.com/jump-dev/JuMP.jl/pull/3413#issuecomment-1603425251 for a breaking release to happen.

x-ref https://discourse.julialang.org/t/jump-dimensionaldata-compatibility/109736

cc @Ethan-russell


mlubin:
Would a compat helper script that notifies us of new releases of dependencies work here?

odow:
Yeah we could add https://github.com/JuliaRegistries/CompatHelper.jl

I think I avoided it in the past because it tended to be more annoying than useful.

PR #3668, opened 2024-02-05
Add is_solved_and_feasible

odow:
x-ref https://discourse.julialang.org/t/jump-behaviour-of-value-for-infeasible-models/109662/4

Perhaps one reason the question keeps coming up is that the docs don't actually uniformly implement the checks (because we "know" that HiGHS will find the right answer) :smile: I'll take the blame for that.


odow:
See https://discourse.julialang.org/t/jump-behaviour-of-value-for-infeasible-models/109662/5 for some additional thoughts.

docs/src/tutorials/getting_started/getting_started_with_JuMP.jl
```diff
@@ -198,6 +198,12 @@ dual_status(model)
 # Now we know that our solver found an optimal solution, and that it has a
 # primal and a dual solution to query.
 
+# It is good practice in your code to include statements like this so that an
```
joaquimg, line 201: maybe not even "good practice" but something stronger like "very important" or "imperative"

odow: Yeah.

odow:
I think I like my suggestion from discourse:
```julia
"""
    has_solution(model::Model; dual::Bool = false)

Return `true` if the model has a feasible primal solution to return and the
solver terminated normally. If `dual`, additionally check that a feasible
dual solution is available.

If this function returns `false`, use [`termination_status`](@ref), [`result_count`](@ref),
[`primal_status`](@ref) and [`dual_status`](@ref) to understand what solutions
are available (if any).
"""
function has_solution(model::Model; dual::Bool = false)
    ret = termination_status(model) in (OPTIMAL, LOCALLY_SOLVED) &&
          primal_status(model) == FEASIBLE_POINT
    if dual
        ret &= dual_status(model) == FEASIBLE_POINT
    end
    return ret
end
```

Then these would all just become `@assert has_solution(model)` and `@assert has_solution(model; dual = true)`.

We have `has_values` and `has_duals`, but they don't check the termination status.

docs/src/tutorials/getting_started/getting_started_with_JuMP.jl
gdalle: Note that assertions can be disabled: https://docs.julialang.org/en/v1/base/base/#Base.@assert

gdalle: Not sure it matters much for docs though

docs/src/tutorials/linear/cannery.jl
```diff
@@ -121,10 +121,9 @@ solution_summary(model)
 
 # What's the optimal shipment?
 
+Test.@test termination_status(model) == OPTIMAL
+Test.@test primal_status(model) == FEASIBLE_POINT
+Test.@test objective_value(model) == 1_680.0           #src
```
gdalle, line 126: maybe `isapprox`?

gdalle (commented):
This is a very welcome addition! I would also add some details on the various problem statuses, as mentioned on discourse
https://discourse.julialang.org/t/jump-behaviour-of-value-for-infeasible-models/109662/6

src/optimizer_interface.jl
```diff
@@ -573,6 +573,45 @@ function dual_status(model::GenericModel; result::Int = 1)
...
+[`result_count`](@ref), [`primal_status`](@ref) and [`dual_status`](@ref) to
+understand what solutions are available (if any).
+"""
+function has_optimal_solution(
+    model::GenericModel;
+    dual::Bool = false,
+    allow_local::Bool = true,
+    result::Int = 1,
+)
+    status = termination_status(model)
```
blegat, line 606: We could probably return `true` with a warning if it's `ALMOST_...` with a keyword argument to disable the warning or to make it an error

src/optimizer_interface.jl
```diff
@@ -573,6 +573,62 @@ function dual_status(model::GenericModel; result::Int = 1)
     return MOI.get(model, MOI.DualStatus(result))::MOI.ResultStatusCode
 end
 
+"""
+    has_optimal_solution(
+        model::Model;
```
odow, line 578:
```suggestion
        model::GenericModel;
```

src/optimizer_interface.jl
```diff
@@ -573,6 +573,62 @@ function dual_status(model::GenericModel; result::Int = 1)
...
+        dual::Bool = false,
+        allow_local::Bool = true,
+        allow_almost::Bool = false,
+        result::Int = 1,
+    )
+
+Return `true` if the model has an optimal primal solution associated with result
+index `result`.
+
+If `dual`, additionally check that a feasible dual solution is available.
```
odow, line 588:
```suggestion
If `dual`, additionally check that an optimal dual solution is available.
```

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,34 @@ Subject to
  y[b] ≤ 1
 ```
 
+## Check if optimal solution exists
+
+Use [`has_optimal_solution`](@ref) to check if the solver found an optimal
+solution:
+```julia solutions
```
odow, line 39:
```suggestion
```jldoctest solutions
```

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,34 @@ Subject to
...
+julia> has_optimal_solution(model)
+true
+```
+
+By default [`has_optimal_solution`](@ref) returns `true` for both global and
+local optima. Pass `allow_local = false` to check if the solver found a globally
+optimal solution:
+```julia solutions
```
odow, line 47:
```suggestion
```jldoctest solutions
```

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,34 @@ Subject to
...
+julia> has_optimal_solution(model; allow_local = false)
+true
+```
+
+Pass `dual = true` to check if the solver found an optimal dual solution in
+addition to an optimal primal solution:
+```solutions
```
odow, line 54:
```suggestion
```jldoctest solutions
```

docs/src/tutorials/getting_started/getting_started_with_JuMP.jl
```diff
@@ -180,6 +180,11 @@ optimize!(model)
 
 # Now let's see what information we can query about the solution.
```
odow, line 181:
```suggestion
# Now let's see what information we can query about the solution,
# starting with [`has_optimal_solution`](@ref):
```

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,34 @@ Subject to
  y[b] ≤ 1
 ```
 
+## Check if optimal solution exists
```
odow, line 35:
```suggestion
## Check if an optimal solution exists
```

docs/src/manual/solutions.md
```diff
@@ -32,6 +32,34 @@ Subject to
...
+
+Use [`has_optimal_solution`](@ref) to check if the solver found an optimal
+solution:
+```jldoctest solutions
+julia> has_optimal_solution(model)
+true
+```
+
+By default [`has_optimal_solution`](@ref) returns `true` for both global and
```
odow, line 44:
```suggestion
By default, [`has_optimal_solution`](@ref) returns `true` for both global and
```

docs/src/manual/solutions.md
```diff
@@ -267,11 +295,37 @@ And data, a 2-element Vector{Float64}:
...
+
+julia> solve_and_print_solution(model)
+Solution is optimal
+  objective value = -205.14285714285714
+  primal solution: x = 15.428571428571429
+  dual solution: c1 = 1.7142857142857142
+```
+
+You can also use a more advanced workflow that deals with a broader range of
+statues:
```
odow, line 325:
```suggestion
statuses:
```

odow:
See https://jump.dev/JuMP.jl/previews/PR3668/manual/solutions/

src/optimizer_interface.jl
```diff
@@ -573,6 +573,65 @@ function dual_status(model::GenericModel; result::Int = 1)
...
+[`NEARLY_FEASIBLE_POINT`](@ref).
+
+If this function returns `false`, use [`termination_status`](@ref),
+[`result_count`](@ref), [`primal_status`](@ref) and [`dual_status`](@ref) to
+understand what solutions are available (if any).
+"""
+function has_optimal_solution(
+    model::GenericModel;
+    dual::Bool = false,
+    allow_local::Bool = true,
```
blegat, line 610: It's quite counter-intuitive that non-optimal solution are considered optimal by default. I would use `allow_local = false` by default.

odow: Hard disagree. This is an entry point function for users who don't need to know the details. They'll complain if they need to write `has_optimal_soution(model; allow_local = true)` when solving an LP with Ipopt but not with HiGHS.

"optimal" is widely used for both global and local minima. I don't really want to get into the nitty gritty. (If we did, we should probably change `allow_local` to be `allow_stationary_point`...)

odow: @mlubin your thoughts would be useful

blegat: I don't think it's a technical detail. From the user that does not know anything about how the problem is solved, he will rightfully think that `has_optimal_solution` means the solution is optimal. If you solve an LP with Ipopt, Maybe Ipopt could detect the problem is convex and return `MOI.OPTIMAL` then.

odow: Ipopt cannot detect convexity. 

I don't think people care about the difference between global and local optima. Either problem is convex and we could a global, or the problem is non-convex, and they probably want only a local. If the problem is non-convex and they want the global, they're doing something special and can opt-out.

Even then, `AmplNLWriter` returns `LOCALLY_SOLVED` even if you use a global solver like Couenne (because we can't know what the executable at `"/path/to/random/ampl.exe"` is).

I don't want to have to explain to new users why they can use `has_optimal_solution(model)` for some solvers and `has_optimal_solution(model; allow_local = true)` for others.

mlubin: I think it's confusing to break with MOI and start saying that `LOCALLY_SOLVED` is "optimal". Can we use a different word like "solved" instead of "optimal"? `is_solved`?

odow: Some alternatives:

 * `has_solution(model; dual = true)`
 
 * `is_solved(model)`
 * `was_solved(model)`
 * `solved_successfully(model)`
 * `solved_successfully(model; require_global = true)` <-- an new option

 * `is_optimize_success(model)`
 * `optimized_successfully(model)`
 * `optimize_succeeded(model)`

joaquimg: I think we thought carefully about MOI statuses.
So the new function should take longer to bikeshed, otherwise we might commit to something that will prove to be bad later.

I vote for fixing the entire manual to use termination_status and has_primal all around. So that people can start doing it right.

I would leave the new function for a larger and careful discussion.

joaquimg: Still `termination_success(model)` might be more compatible with MOI.

odow: > I vote for fixing the entire manual to use termination_status and has_primal all around. So that people can start doing it right.

I did that in the first commit to this PR. But it didn't feel right. Even I wouldn't want to write that in my code.

I really want a single "all is good" call, and if not, then I can dig into the details.

I'd be okay with `is_termination_success(model)`

blegat: `solved_successfully(model)` seems consistent with `LOCALLY_SOLVED`. It means that infeasible or unbounded is considered as unsuccessful but that's ok

mlubin: `normal_termination`?
The issue with "success" is that, from the solver's perspective, detecting infeasibility or unboundedness is a success because it did exactly what you asked it to do.
What about `ALMOST_OPTIMAL` and `ALMOST_LOCALLY_SOLVED`?

odow: There's an `allow_almost::Bool = false` option.

What about `has_converged_solution`, `has_solved_solution`.

Or something like
```
has_feasible_solution(
    model;
    primal::Bool = true,
    dual::Bool = false,
    allow_limit_termination::Bool = false,
    allow_local_optima::Bool = true,
    allow_almost::Bool = false,
    result::Int = 1,
)
```

blegat: The solution could be feasible even if the status is NUMERICAL_ERROR. `has_converged_solution` is nice even though yoi could still converge to a certificate. `solved_solution` is twice `solve`. What about `is_feasibly_solved` ?

blegat: Because it basically means that it's solved (consistent with `LOCALLY_SOLVED`) and it's feasible (consistent with the primal status). Having a function name combining these two words would make sure this is not misinterpreted

odow: `is_solved_and_feasible`?

odow:
I've changed to `is_solved_and_feasible`

blegat (approved):

docs/src/manual/solutions.md
```diff
@@ -267,11 +295,37 @@ And data, a 2-element Vector{Float64}:
...
+
+A simple approach is to use [`is_solved_and_feasible`](@ref):
+
 ```jldoctest solutions
-julia> begin
-           if termination_status(model) == OPTIMAL
+julia> function solve_and_print_solution(model)
+           optimize!(model)
+           if !is_solved_and_feasible(model; dual = true)
+               error("The model was not solved correctly.")
```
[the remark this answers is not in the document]
blegat: It's also useful to include the `raw_status` here

docs/src/manual/solutions.md
```diff
@@ -267,25 +295,103 @@ And data, a 2-element Vector{Float64}:
...
+A simple approach for small scripts and notebooks is to use
+[`is_solved_and_feasible`](@ref):
+
 ```jldoctest solutions
-julia> begin
-           if termination_status(model) == OPTIMAL
+julia> function solve_and_print_solution(model)
+           optimize!(model)
+           if !is_solved_and_feasible(model; dual = true)
+               term = termination_status(model)
```
odow, line 308:
```suggestion
```

docs/src/manual/solutions.md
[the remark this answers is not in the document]
odow: Done

mlubin (approved):

Issue #3678, opened 2024-02-15
Iterating SparseAxisArray does not preserve order


odow:
x-ref https://discourse.julialang.org/t/unexpected-jump-variable-order-when-indexed-by-tuple/110262

```julia
julia> using JuMP

julia> model = Model();

julia> N = 1:2;

julia> M = [[1.0, 11.0, 100.0], [0.0, 17.0, 104.0]];

julia> @variable(model, foo[n in N, M[n]])
JuMP.Containers.SparseAxisArray{VariableRef, 2, Tuple{Int64, Float64}} with 6 entries:
  [1, 1.0  ]  =  foo[1,1.0]
  [1, 100.0]  =  foo[1,100.0]
  [1, 11.0 ]  =  foo[1,11.0]
  [2, 0.0  ]  =  foo[2,0.0]
  [2, 104.0]  =  foo[2,104.0]
  [2, 17.0 ]  =  foo[2,17.0]

julia> foo.data
Dict{Tuple{Int64, Float64}, VariableRef} with 6 entries:
  (2, 17.0)  => foo[2,17.0]
  (1, 11.0)  => foo[1,11.0]
  (1, 1.0)   => foo[1,1.0]
  (1, 100.0) => foo[1,100.0]
  (2, 0.0)   => foo[2,0.0]
  (2, 104.0) => foo[2,104.0]

julia> @constraint(model, foo[1, :] in SecondOrderCone())
[foo[1,11.0], foo[1,1.0], foo[1,100.0]] ∈ MathOptInterface.SecondOrderCone(3)
```

User could reasonably expect constraint function to be `[foo[1,1.0], foo[1,11.0], foo[1,100.0]]`.

This is because we store the data in a `Dict`.

We could swap to an `OrderedDict`, or throw an error or a warning when we convert a `SparseAxisArray` to a vector.

odow:
This is actually pretty nasty.

We fallback to `AbstractVector` methods:

```julia
julia> using JuMP

julia> A = [[1, 2, 10], [2, 3, 30]]
2-element Vector{Vector{Int64}}:
 [1, 2, 10]
 [2, 3, 30]

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: NO_OPTIMIZER
Solver name: No optimizer attached.

julia> @variable(model, x[i in 1:2, j in A[i]])
[...]

PR #3680, opened 2024-02-16
[breaking] throw error if Containers.SparseAxisArray used in constraint function

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3678

This PR is the nuclear option of throwing an error. Technically, this is breaking change, but it blocks a high risk of incorrect usage.

We should look at the tests and `extension-tests.yml` before deciding to merge this: https://github.com/jump-dev/JuMP.jl/actions/runs/7924586714

We could make this less breaking by adding methods like:
```julia
function build_constraint(
    ::Function,
    f::Containers.SparseAxisArray{<:Union{Number,AbstractJuMPScalar},1},
    s::MOI.Nonnegatives,
)
    return VectorConstraint(f, s)
end

function build_constraint(
    ::Function,
    f::Containers.SparseAxisArray{<:Union{Number,AbstractJuMPScalar},1},
    s::MOI.Nonpositives,
)
    return VectorConstraint(f, s)
end

function build_constraint(
    ::Function,
    f::Containers.SparseAxisArray{<:Union{Number,AbstractJuMPScalar},1},
    s::MOI.Zeros,
)
    return VectorConstraint(f, s)
end
```

but even then, things like the constraint dual might be in the "wrong" order.

### DIFF
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -2358,4 +2358,19 @@ function test_op_or_short_circuit()
     return
 end
 
+function test_sparseaxisarray_constraint_zeros()
+    A = [[1, 2, 10], [2, 3, 30]]
+    model = Model()
+    @variable(model, x[i in 1:2, j in A[i]])
+    @test_throws_runtime(
+        ErrorException,
+        @constraint(model, x[1, :] == 0),
+    )
+    @test_throws_runtime(
+        ErrorException,
+        @constraint(model, x[1, :] in SecondOrderCone()),
+    )
+    return
+end
+
 end  # module
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -2362,13 +2362,24 @@ function test_sparseaxisarray_constraint_zeros()
     A = [[1, 2, 10], [2, 3, 30]]
     model = Model()
     @variable(model, x[i in 1:2, j in A[i]])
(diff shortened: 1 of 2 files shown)


test/test_macros.jl
```diff
@@ -2358,4 +2358,30 @@ function test_op_or_short_circuit()
...
+            "In `@constraint(model, x[1, :] in SecondOrderCone())`: $msg",
+        ),
+        @constraint(model, x[1, :] in SecondOrderCone()),
+    )
+    @test_throws_runtime(
+        ErrorException("In `@constraint(model, x[1, :] in SOS1())`: $msg"),
+        @constraint(model, x[1, :] in SOS1()),
+    )
+    @test_throws_runtime(
+        ErrorException("In @constraint(model, x[1, :] in SOS2())`: $msg"),
```
odow, line 2381:
```suggestion
        ErrorException("In `@constraint(model, x[1, :] in SOS2())`: $msg"),
```

odow:
So this is a tricky decision. We didn't use it in any tests, and it was never used by any extension that we test.

I wonder how common it is in user-land.

odow:
An alternative is to fix the iteration ordering with an `OrderedDict`.

odow:
Closing in favor of #3681

PR #3681, opened 2024-02-17
[Containers] use OrderedDict as the data structure for SparseAxisArray

### LINKED ISSUE #3678: Iterating SparseAxisArray does not preserve order

odow:
Closes #3678 
Closes #3680

The issue is still really that `SparseAxisArray` is not a regular n-dimensional Array. Each slice can have a different length with different axes. It doesn't make sense to support a lot of operations with it.

The only thing that we should support are vectors.

This PR has the problem that the current iteration order is the transpose of what it "should" be (row-major, instead of column-major). 

This is needed because the second and subsequent dimensions can depend on the first and prior. So it isn't really meaningful to iterate "down" the first index holding all else constant.

I can see this being a constant source of future problems.


src/Containers/SparseAxisArray.jl
```diff
@@ -448,7 +464,6 @@ function Base.show(io::IOContext, x::SparseAxisArray)
         (i, (key, value)) in enumerate(x.data) if
         i < half_screen_rows || i > length(x) - half_screen_rows
     ]
-    sort!(key_strings; by = x -> x[1])
```
odow, line 451: I didn't run the full tests locally, but from what I can tell, there are few, if any, printing changes, because we used to sort the keys, and most tests/examples provided the keys in a nice sorted order.

src/Containers/SparseAxisArray.jl
```diff
@@ -5,28 +5,32 @@
...
+julia> dict = OrderedDict((:a, 2) => 1.0, (:a, 3) => 2.0, (:b, 3) => 3.0)
+OrderedDict{Tuple{Symbol, Int64}, Float64} with 3 entries:
+  (:a, 2) => 1.0
   (:a, 3) => 2.0
   (:b, 3) => 3.0
-  (:a, 2) => 1.0
 
 julia> array = Containers.SparseAxisArray(dict)
-SparseAxisArray{Float64, 2, Tuple{Symbol, Int64}} with 3 entries:
+JuMP.Containers.SparseAxisArray{Float64, 2, Tuple{Symbol, Int64}} with 3 entries:
```
odow, line 33:
```suggestion
SparseAxisArray{Float64, 2, Tuple{Symbol, Int64}} with 3 entries:
```

joaquimg:
Shouldnt the iteration ordering comment be part of docs?

odow:
Yeah. I guess my first question is: should we do this? 

If we do, we need more docs on the specifics of the order and slicing, and when things can go wrong.

Also: https://github.com/jump-dev/JuMP.jl/actions/runs/7944311147

joaquimg:
The slicing issues you showed seem pretty dangerous.

The only downside would be the performance penalty? Did you benchmark it?

odow:
> The only downside would be the performance penalty? Did you benchmark it?

I didn't benchmark it, but I think the risk of incorrect usage outweighs the cost.

It's probably slower because creating an ordered dict is slower. But it shouldn't be too bad.

joaquimg:
I agree correctness must come first.

Is there any other downside?

odow:
I added a warning.

I guess the main downside is that existing iteration orders may change. But that could have happened anyway, because we used a `Dict`.

blegat (approved):
The change of order of `eachindex` is indeed sub-optimal since we want `SparseAxisArrays` to match the behavior of `Array` it is dense and it's indices are `Base.OneTo`. However, it doesn't seem that this could be done in general and it's still a strict improvement to the behavior before this PR.

docs/src/manual/containers.md
```diff
@@ -315,13 +315,28 @@ JuMP.Containers.SparseAxisArray{Tuple{Int64, Symbol}, 1, Tuple{Int64}} with 2 en
...
            println(y[key])
        end
 (2, :B)
 (3, :B)
 ```
 
+!!! warning
+    In contrast to `Base.Array`s, which iterate in column-major order, that is,
+    by varying indices from left to right, `SparseAxisArray` iterates in
+    row-major order, that is, by varying the indices from right to left. As an
```
mlubin, line 336: Is this true in general or only when the `SparseAxisArray` is created by a macro? If you construct using a dict then the iteration order will match the given dict, right?

odow: Only when created by a macro. Dict order is as given.

odow: Tweaked the wording slightly.

odow:
I audited all the open issues. I think we're safe from the `delete` segfault because you cannot delete elements in a `SparseAxisArray`.

The most up for debate issue is: https://github.com/JuliaCollections/OrderedCollections.jl/issues/82

```Julia
julia> Containers.@container(x[k in 1:2], k, container = SparseAxisArray)
JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 2 entries:
  [1]  =  1
  [2]  =  2

julia> Containers.@container(y[k in 2:-1:1], k, container = SparseAxisArray)
JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 2 entries:
  [2]  =  2
  [1]  =  1

julia> x == y
true
```

But this coincidentally just reproduces the behavior of the current JuMP:
```julia
julia> Containers.@container(x[k in 1:2], k, container = SparseAxisArray)
JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 2 entries:
  [1]  =  1
  [2]  =  2

julia> Containers.@container(y[k in 2:-1:1], k, container = SparseAxisArray)
JuMP.Containers.SparseAxisArray{Int64, 1, Tuple{Int64}} with 2 entries:
  [1]  =  1
  [2]  =  2

julia> x == y
true
```

mlubin:
https://github.com/JuliaCollections/OrderedCollections.jl/issues/82 isn't too concerning to me.

odow:
So good to merge then?

mlubin (approved):

PR #3683, opened 2024-02-26
[docs] clarify section on automatic differentiation in nonlinear.md

odow:
x-ref https://discourse.julialang.org/t/nonlinear-optimization-with-many-constraints-autodifferentiation-which-julia-solution/110678/20

Thoughts @gdalle

Preview: https://jump.dev/JuMP.jl/previews/PR3683/manual/nonlinear/#jump_user_defined_operators


docs/src/manual/nonlinear.md
```diff
@@ -417,8 +417,12 @@ using the [`@operator`](@ref) macro.
 ## [User-defined operators](@id jump_user_defined_operators)
 
 In addition to a standard list of univariate and multivariate operators
-recognized by the `MOI.Nonlinear` submodule, JuMP supports *user-defined*
-Julia operators.
+recognized by the `MOI.Nonlinear` submodule, JuMP supports user-defined
+operators.
```
gdalle, line 421: At this point it's not clear what one can do with an operator that cannot be done with function tracing

docs/src/manual/nonlinear.md
```diff
@@ -622,53 +626,166 @@ f_splat(x..., y..., z)
...
-### Automatic differentiation
+### Common mistakes when writing a user-defined operator
 
-JuMP does not support black-box optimization, so all user-defined operators must
-provide derivatives in some form. Fortunately, JuMP supports automatic
-differentiation of user-defined operators.
+When nonlinear expressions are provided algebraically (either via direct input
+into the macros or via [Function tracing](@ref)), JuMP computes first- and
+second-order derivatives using sparse reverse-mode automatic differentiation.
+For details, see [ReverseAD](@ref).
```
gdalle, line 634: I would put this earlier because it's not about operators per se

docs/src/manual/nonlinear.md
```diff
@@ -622,53 +626,166 @@ f_splat(x..., y..., z)
...
+Instead, unless you provide your own derivatives with [Gradients and Hessians](@ref),
 JuMP uses [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl) to
-perform automatic differentiation of user-defined operators; see the ForwardDiff.jl
-[documentation](https://www.juliadiff.org/ForwardDiff.jl/v0.10.2/user/limitations.html)
-for a description of how to write a function suitable for automatic
-differentiation.
+compute derivatives.
 
-#### Common mistakes when writing a user-defined operator
+ForwardDiff has a number of limitations that you should be aware of when
```
gdalle, line 643: Maybe wise to mention that gradients are very slow in high dimension cause forward mode AD?

odow:
How about this now.  I don't really know where the trade-off is between providing lots of information one the details for expert users, and trying to hide irrelevant information for new users.

docs/src/manual/nonlinear.md
```diff
@@ -161,6 +161,14 @@ julia> sin(sin(1.0))
 0.7456241416655579
 ```
 
+## Automatic differentiation
+
+JuMP computes first- and second-order derivatives using sparse reverse-mode
+automatic differentiation. For details, see [ReverseAD](@ref).
+
+For a tutorial on how to construct and query the derivatives, see
+[Querying hessians](@ref)
```
odow, line 170:
```suggestion
[Computing Hessians](@ref)
```

docs/src/manual/nonlinear.md
```diff
@@ -161,6 +161,14 @@ julia> sin(sin(1.0))
 0.7456241416655579
 ```
 
+## Automatic differentiation
```
gdalle, line 164: There is a second section about AD at the bottom, I would move this down with the rest

docs/src/manual/nonlinear.md
```diff
@@ -407,22 +415,12 @@ ERROR: Cannot evaluate `>` between a variable and a number.
...
 
 In addition to a standard list of univariate and multivariate operators
 recognized by the `MOI.Nonlinear` submodule, JuMP supports user-defined
-operators.
-
-Unless [Gradients and Hessians](@ref) are explicitly provided, user-defined
-operators must support automatic differentiation. If you encounter an error
-adding the operator, see [Common mistakes when writing a user-defined operator](@ref).
+operators, which let you represent nonlinear functions that cannot (or should
+not) be traced.
```
gdalle, line 423:
```suggestion
not) be traced, for instance because they rely on non-Julia subroutines.
```

docs/src/manual/nonlinear.md
```diff
@@ -464,6 +462,29 @@ model = Model();
 op_square_2 = model[:op_square]
 ```
 
+## Automatic differentiation
```
gdalle, line 465: This section is great!

gdalle (changes requested):

docs/src/manual/nonlinear.md
odow: I kept the section. I wanted to distinguish general AD from the AD specific to operators. The bottom section was `##` instead of `###` by mistake, so it looked more prominent than I intended.

PR #3686, opened 2024-02-28
Add _CONSTRAINT_LIMIT_FOR_PRINTING to control printing of many constraints

### LINKED ISSUE #3651: Model printing is rather verbose
Consider something as simple as
```
model = Model();
@variable(model, 0 <= x[1:100, 1:100] <= 10, Int)
print(model)
```
This produces ~30k lines of output.

Given that each `x[i]` can even have it's own type,
i can't imagine solving this is too simple,
but surely this can be improved?

(I've encountered this the hard way, with chrome tab crashing
(and, you can't even increase per-tab memory limits nowadays!),
and had to fix it by manually going into `.ipynb` notebook...)

odow:
Closes #3651

We need to decide whether this is okay. It is conceivably breaking for people who rely on printing the model to a string, even if they really shouldn't be doing that.

One justification for doing this is that we added https://github.com/jump-dev/JuMP.jl/pull/3575 and I haven't heard any complaints, and it is pretty easy to suddenly spam a lot of content to stdout if your model is large.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -371,10 +371,11 @@ function _print_latex(io::IO, model::AbstractModel)
     if !isempty(constraints)
         print(io, "\\text{Subject to} \\quad")
         m = div(_CONSTRAINT_LIMIT_FOR_PRINTING[], 2)
-        skip_start, skip_stop = m + 1, length(constraints) - m
-        n = skip_stop - skip_start
+        skip_start = _CONSTRAINT_LIMIT_FOR_PRINTING[] - m + 1
+        skip_stop = length(constraints) - m
+        n = skip_stop - skip_start + 1
         for (i, constraint) in enumerate(constraints)
-            if i <= skip_start || skip_stop <= i
+            if i < skip_start || skip_stop < i
                 println(io, " & ", constraint, "\\\\")
             end
             if n > 0 && i == skip_start
--- a/src/print.jl
+++ b/src/print.jl
@@ -283,6 +283,24 @@ function show_backend_summary(io::IO, model::GenericModel)
 end
 
 """
+    const _CONSTRAINT_LIMIT_FOR_PRINTING = Ref{Int}(100)
+
+A global constant used to control when constraints are omitted when printing
+the model.
+
+Get and set this value using `_CONSTRAINT_LIMIT_FOR_PRINTING[]`.
+
+```julia
+julia> _CONSTRAINT_LIMIT_FOR_PRINTING[]
+100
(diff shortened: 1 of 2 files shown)


odow:
An alternative might be to print at most N constraints of each function-set type. That would give a more representative print of the model.

src/print.jl
```diff
@@ -282,6 +282,24 @@ function show_backend_summary(io::IO, model::GenericModel)
...
+
+```julia
+julia> _CONSTRAINT_LIMIT_FOR_PRINTING[]
+100
+
+julia> _CONSTRAINT_LIMIT_FOR_PRINTING[] = 10
+10
+```
+"""
+const _CONSTRAINT_LIMIT_FOR_PRINTING = Ref{Int}(100)
```
joaquimg, line 301: Is it too bad if it is a field in the model struct?

odow: That would be an alternative. I'm open to discussion more generally on the right approach for this. I don't particularly like the way it currently is.

src/print.jl
```diff
@@ -282,6 +282,24 @@ function show_backend_summary(io::IO, model::GenericModel)
...
+A global constant used to control when constraints are omitted when printing
+the model.
+
+Get and set this value using `_CONSTRAINT_LIMIT_FOR_PRINTING[]`.
+
+```julia
+julia> _CONSTRAINT_LIMIT_FOR_PRINTING[]
+100
+
+julia> _CONSTRAINT_LIMIT_FOR_PRINTING[] = 10
```
blegat, line 297: Documented this means it would be breaking to remove this ?

blegat:
I like MOI's `_PrintOptions`. I think we unify MOI and JuMP printing and we should start by having the same option struct.

LebedevRI:
It would be nice if these kind of knobs be mentioned in the docs,
otherwise they are kind-of hard to find.

odow:
We have purposefully chosen not to mention these in the docs. They are currently part of the internal API.

Is there an issue with the current defaults? Why do you want to change them?

LebedevRI:
> We have purposefully chosen not to mention these in the docs. They are currently part of the internal API.
> 
> Is there an issue with the current defaults? Why do you want to change them?

(I agree that the current default is better than nothing, as discussed.)
I wanted to view some of the constraints that got omitted from the output

odow:
There are probably better tools for viewing a subset of constraints. For example, you could loop through `all_constraints(model)`.

PR #3693, opened 2024-02-29
Fix LinearAlgebra.hermitian(::AbstractJuMPScalar, ::Symbol)

### LINKED ISSUE #3692: LinearAlgebra.hermitian is incorrect
https://github.com/jump-dev/JuMP.jl/blob/119569556f5fd28d12b012f4c94d8dfdaca6d65a/src/JuMP.jl#L1058
doesn't seem to match the docstring of `LinearAlgebra.hermitian`:

> If `A` is a Number, its real part is returned converted back to the input type.

odow:
Closes #3692

Docstrinng is:
```julia
help?> LinearAlgebra.hermitian
  hermitian(A, uplo=:U)

  Construct a hermitian view of A. If A is a matrix, uplo controls whether the upper (if uplo = :U)
  or lower (if uplo = :L) triangle of A is used to implicitly fill the other one. If A is a Number,
  its real part is returned converted back to the input type.

  If a hermitian view of a matrix is to be constructed of which the elements are neither matrices
  nor numbers, an appropriate method of hermitian has to be implemented. In that case,
  hermitian_type has to be implemented, too.
```

The important sentence is: `If A is a Number, its real part is returned converted back to the input type.`

### DIFF
--- a/test/test_complex.jl
+++ b/test/test_complex.jl
@@ -312,9 +312,9 @@ end
 
 function test_hermitian_jump_scalar()
     model = Model()
-    @variable(model, X[1:2,1:2] in HermitianPSDCone())
+    @variable(model, x[1:2, 1:2] in HermitianPSDCone())
     Y = im * x.data
-    Z = [0 real(x[1,2])*im-imag(x[1,2]); -real(x[1,2])*im-imag(x[1,2]) 0]
+    Z = [0 real(x[1, 2])*im-imag(x[1, 2]); -real(x[1, 2])*im-imag(x[1, 2]) 0]
     @test !isequal_canonical(Y, Z)
     @test isequal_canonical(
         LinearAlgebra.Hermitian(Y),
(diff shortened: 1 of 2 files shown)


test/test_complex.jl
```diff
@@ -310,4 +310,17 @@ function test_hermitian_generic_number_type()
     return
 end
 
+function test_hermitian_jump_scalar()
+    model = Model()
+    @variable(model, X[1:2,1:2] in HermitianPSDCone())
```
odow, line 315:
```suggestion
    @variable(model, x[1:2, 1:2] in HermitianPSDCone())
```

test/test_complex.jl
```diff
@@ -310,4 +310,17 @@ function test_hermitian_generic_number_type()
...
+    Y = im * x.data
+    Z = [0 real(x[1,2])*im-imag(x[1,2]); -real(x[1,2])*im-imag(x[1,2]) 0]
```
odow, line 317:
```suggestion
    Z = [0 real(x[1, 2])*im-imag(x[1, 2]); -real(x[1, 2])*im-imag(x[1, 2]) 0]
```

PR #3695, opened 2024-02-29
Fix *(::AbstractJuMPScalar{<:Real}, ::Hermitian) to return Hermitian

### LINKED ISSUE #3694: *(::Real, ::Hermitian) is not hermitian
```Julia
julia> using JuMP, LinearAlgebra

julia> A = LinearAlgebra.Hermitian([1 1+1im; 1-1im 2])
2×2 Hermitian{Complex{Int64}, Matrix{Complex{Int64}}}:
 1+0im  1+1im
 1-1im  2+0im

julia> model = Model();

julia> @variable(model, x)
x

julia> x * A
2×2 Matrix{GenericAffExpr{ComplexF64, VariableRef}}:
 x           (1 + im) x
 (1 - im) x  2 x
```

x-ref https://github.com/jump-dev/MutableArithmetics.jl/issues/264#issuecomment-1971966048

odow:
Closes #3694

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -516,7 +516,9 @@ function Base.:*(
 )
     c = LinearAlgebra.sym_uplo(A.uplo)
     B = c == :U ? _mult_upper(x, A) : _mult_lower(x, A)
-    return LinearAlgebra.Hermitian(B, c)
+    # Intermediate conversion to `Matrix` is needed to work around
+    # https://github.com/JuliaLang/julia/issues/52895
+    return LinearAlgebra.Hermitian(Matrix(LinearAlgebra.Hermitian(B, c)), c)
 end
 
 function Base.:*(
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -502,3 +502,30 @@ function _MA.operate!!(
 )
     return _throw_operator_error(op, x)
 end
+
+_mult_upper(α, A) = parent(α * LinearAlgebra.UpperTriangular(parent(A)))
+_mult_lower(α, A) = parent(α * LinearAlgebra.LowerTriangular(parent(A)))
+
+function Base.:*(
+    x::Union{
+        GenericVariableRef{<:Real},
+        GenericAffExpr{<:Real},
+        GenericQuadExpr{<:Real},
+    },
+    A::LinearAlgebra.Hermitian,
+)
+    c = LinearAlgebra.sym_uplo(A.uplo)
+    B = c == :U ? _mult_upper(x, A) : _mult_lower(x, A)
+    return LinearAlgebra.Hermitian(B, c)
+end
+
+function Base.:*(
+    A::LinearAlgebra.Hermitian,
+    x::Union{
+        GenericVariableRef{<:Real},
+        GenericAffExpr{<:Real},
+        GenericQuadExpr{<:Real},
+    },
+)
+    return x * A
+end
--- a/test/test_complex.jl
+++ b/test/test_complex.jl
@@ -324,15 +324,15 @@ function test_hermitian_jump_scalar()
 end
 
 function test_mul_real_hermitian()
-    A = [1 1+1im; 1-1im 2]
+    A = [1+1im 1+1im; 1-2im 3+1im]
     model = Model()
     @variable(model, x)
     for s in (:L, :U), f in (x, x + 1, x^2)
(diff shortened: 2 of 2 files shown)


araujoms:
Could you also do the Symmetric case? It's actually much easier because you don't need to test whether the scalar is real.

test/test_complex.jl
```diff
@@ -323,4 +323,18 @@ function test_hermitian_jump_scalar()
...
+function test_mul_real_hermitian()
+    A = [1 1+1im; 1-1im 2]
+    model = Model()
+    @variable(model, x)
+    for s in (:L, :U), f in (x, x + 1, x^2)
+        B = LinearAlgebra.Hermitian(A, s)
+        @test f * B isa LinearAlgebra.Hermitian
+        @test isequal_canonical(f * B, LinearAlgebra.Hermitian(f * A))
+        @test B * f isa LinearAlgebra.Hermitian
+        @test isequal_canonical(B * f, LinearAlgebra.Hermitian(A * f))
```
araujoms, line 335: I'd suggest using an underlying non-Hermitian matrix for a more stringent test. I've seen related bugs appearing in `LinearAlgebra`.
```suggestion
    A = [1+1im 1+1im; 1-2im 3+1im]
    model = Model()
    @variable(model, x)
    for s in (:L, :U), f in (x, x + 1, x^2)
        B = LinearAlgebra.Hermitian(A, s)
        @test f * B isa LinearAlgebra.Hermitian
        @test isequal_canonical(f * B, LinearAlgebra.Hermitian(f * A, s))
        @test B * f isa LinearAlgebra.Hermitian
        @test isequal_canonical(B * f, LinearAlgebra.Hermitian(A * f, s))
```

odow:
This one we can fix in MA: https://github.com/jump-dev/MutableArithmetics.jl/pull/268

araujoms:
I've just realized that with this definition of multiplication you are generating hidden `undef`s in the lower triangular, that is, you are relying on the bug JuliaLang/julia#52895 being fixed, which I think is only the case for Julia 1.11 onwards. So the following code will fail for all previous version of Julia:
```julia
using JuMP
prob = Model()
@variable(prob,x)
m = x*LinearAlgebra.Hermitian(randn(2,2))
m+m
```

odow:
Perhaps we should close this then. The currently behavior is correct. The new PR would introduce the risk of incorrect behavior due to bugs in Base.

If JuMP generates a `Matrix`, you can always explicitly wrap it in `Hermitian`.

It's really hard to uniformly support all of LinearAlgebra, SparseArrays, MutableArithmetics, and JuMP.

araujoms:
Or you could just return instead
```julia
LinearAlgebra.Hermitian(Matrix(LinearAlgebra.Hermitian(B, c)),c)
```
This gets rids of the undefs, so no bugs are possible, you keep the performance gain, and don't require the users to always rewrap everything in Hermitian.

src/operators.jl
```diff
@@ -502,3 +502,30 @@ function _MA.operate!!(
...
+    x::Union{
+        GenericVariableRef{<:Real},
+        GenericAffExpr{<:Real},
+        GenericQuadExpr{<:Real},
+    },
+    A::LinearAlgebra.Hermitian,
+)
+    c = LinearAlgebra.sym_uplo(A.uplo)
+    B = c == :U ? _mult_upper(x, A) : _mult_lower(x, A)
+    return LinearAlgebra.Hermitian(B, c)
```
araujoms, line 519:
```suggestion
    return LinearAlgebra.Hermitian(Matrix(LinearAlgebra.Hermitian(B, c)),c)
```

araujoms:
Thanks a lot for solving all of the problems I pointed out! I'm always amazed by how quick you and @blegat are.

odow:
No problem. A lot of the problems are because you might be the only person actively using JuMP for complex Hermitian stuff... 😆

araujoms:
Well, one might say only, but I'd rather say first. :wink: 

There are also [these guys](https://arxiv.org/abs/2310.05844), but they invented their own reformulation of the complex Hermitian PSD cone, so they are only using JuMP for the real one.

odow:
> There are also [these guys](https://arxiv.org/abs/2310.05844)

You might recognize an author

araujoms:
Well, I know some of them personally, but there is one that I'm sure you do. :laughing:

PR #3696, opened 2024-03-01
Add the MINLP solver MAiNGO to the documentation

MAiNGO-github:
We have recently built a wrapper around our open-source global MINLP solver MAiNGO, which is is written in C++.
The wrapper initially took a lot of inspiration from the Baron.jl repository.

Looking at the [checklist](https://jump.dev/JuMP.jl/stable/developers/checklists/#Adding-a-new-solver-to-the-documentation), the conditions for including the solver in the documentation should be fulfilled.

We would be very excited to be included in the list of compatible solvers.

https://github.com/MAiNGO-github/MAiNGO.jl

## Basic

       documented.
       use the solver with JuMP

## Documentation

## Optional

### DIFF
--- a/docs/.vale.ini
+++ b/docs/.vale.ini
@@ -20,6 +20,9 @@ Google.Periods = NO
 [docs/src/packages/Manopt.md]
 Google.Exclamation = NO
 
+[docs/src/packages/MAiNGO.md]
+Vale.Spelling = NO
+
 [docs/src/packages/Optim.md]
 Google.EmDash = NO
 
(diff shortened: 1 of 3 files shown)


odow:
Nice! Let me take a look

odow (changes requested):
I'm going to mark this as blocked by https://github.com/MAiNGO-github/MAiNGO.jl/issues/2, but it looks nice!

Pleased you figured out the Yggdrasil stuff :smile:

MAiNGO-github:
@odow 
I added the Github actions.
I also incorporated a recent improvement from Baron.jl

There are were spellchecking warnings. I fixed those that were correct. Some are caused by how our README is set up.
In case this is blocking or annoying, I would suggest not adding to docs/packages.toml, but only to the overview table.

odow (approved):

PR #3697, opened 2024-03-02
add matlab tutorial

araujoms:
Here is my attempt at the tutorial I proposed in #2348. Feel free to criticize and modify it.

PR #3698, opened 2024-03-03
[docs] add Transitioning from MATLAB tutorial

odow:
Cherry-picked from https://github.com/jump-dev/JuMP.jl/pull/3697

<s>Preview: https://jump.dev/JuMP.jl/previews/PR3698/tutorials/getting_started/transitioning_from_matlab/</s>

https://jump.dev/JuMP.jl/previews/PR3698/tutorials/transitioning/transitioning_from_matlab/

### DIFF
--- a/docs/src/tutorials/getting_started/transitioning_from_matlab.jl
+++ b/docs/src/tutorials/getting_started/transitioning_from_matlab.jl
@@ -1,0 +1,243 @@
+# # Transitioning from MATLAB
+
+# Several JuMP users have previous experience with optimization on MATLAB, in
+# particular with the modellers [YALMIP](https://yalmip.github.io/) or [CVX](https://cvxr.com/cvx/).
+# This tutorial is aimed to those users.
+
+# ## Namespaces
+
+# Julia has namespaces, which MATLAB lacks. Therefore one needs to either use the
+# command
+
+ using JuMP
+
+# in order bring all names exported by JuMP into scope, or
+
+ import JuMP
+
+# in order to merely make the JuMP package available. This requires prefixing
+# everything you use from JuMP with `JuMP.`, so it gets old really fast. In this
+# tutorial we'll use the former.
+
+# ## Starting a problem
+
+# With YALMIP and CVX you have a single, implicit optimization problem you're
+# working on. With JuMP you have to create one explicitly, and when you
+# declare variables, constraints, or the objective function you have to 
+# specify to which problem they belong. You create a problem with the command
+
+ prob = Model()
+ 
+# You can optionally also set the optimizer and options as an argument to the
+# function `Model()`, but we'll do this later.
+
+# ## Declaring variables
+
+# In most cases there is a direct translation between variable declarations.
+# The following table shows some common examples:
+
+# | variable | JuMP | YALMIP | CVX |
+# | -------- | ------- | ------- | ------- |
+# | real variable  | `@variable(prob, x)` | `x = sdpvar` | `variable x` |
+# | real vector    | `@variable(prob, v[1:d])` | `v = sdpvar(d,1)` | `variable v(d)` |
+# | real matrix    | `@variable(prob, m[1:d,1:d])` | `m = sdpvar(d,d,'full')` | `variable m(d,d)` |
+# | complex matrix    | `@variable(prob, m[1:d,1:d] in ComplexPlane())` | `m = sdpvar(d,d,'full','complex')` | `variable m(d,d) complex` |
+# | real symmetric matrix    | `@variable(prob, m[1:d,1:d], Symmetric)` | `m = sdpvar(d)` | `variable m(d,d) symmetric` |
+# | complex Hermitian matrix    | `@variable(prob, m[1:d,1:d], Hermitian)` | `m = sdpvar(d,d,'hermitian','complex')` | `variable m(d,d) hermitian` |
+
+# A more interesting case is when you want to declare for example `n` real 
+# symmetric matrices. Both YALMIP and CVX allow you to put the matrices as the 
+# slices of a 3-dimensional array, via the commands `m = sdpvar(d,d,n)` and 
+# `variable m(d,d,n) symmetric`, respectively. With JuMP this is not possible. 
+# Instead, to achieve the same result one needs to declare a vector of `n` 
+# matrices: `m = [@variable(prob, [1:d,1:d], Symmetric) for i=1:n]`.
+# The analogous construct in MATLAB would be a cell array containing the 
+# optimization variables, which every discerning programmer avoids as cell arrays
+# are rather slow. This is not a problem in Julia: a vector of matrices is almost
+# as fast as a 3-dimensional array.
+
+# ## Declaring constraints
+
+# As in the case of variables, in most cases there is a direct translation between
+# the syntaxes:
+
+# | constraint | JuMP | YALMIP | CVX |
+# | -------- | ------- | ------- | ------- |
+# | equality constraint  | `@constraint(prob, v == c)` | `v == c` | `v == c` |
+# | nonnegative cone  | `@constraint(prob, v >= 0)` | `v >= 0` | `v >= 0` |
+# | real PSD cone    | `@constraint(prob, m in PSDCone())` | `m >= 0` | `m == semidefinite(length(m))` |
+# | complex PSD cone    | `@constraint(prob, m in HermitianPSDCone())` | `m >= 0` | `m == hermitian_semidefinite(length(m))` |
+# | second order cone  | `@constraint(prob, [t; v] in SecondOrderCone())` | `cone(v,t)` | `{v,t} == lorentz(length(v))` |
+# | exponential cone  | `@constraint(prob, [x, y, z] in MOI.ExponentialCone())` | `expcone([x,y,z])` | `{x,y,z} == exponential(1)` |
+
+# A subtlety appears when declaring equality constraints for matrices. In general
+# JuMP uses `@constraint(prob, m .== c)`, with the dot meaning broadcasting in 
+# Julia, except when `m` is `Symmetric` or `Hermitian`: in this case `@constraint(prob, m == c)`
+# is allowed, and much better, as JuMP is smart enough to not generate redundant
+# constraints for the lower diagonal and the imaginary part of the diagonal (in
+# the complex case). Both YALMIP and CVX are also smart enough to do this and 
+# the syntax is always just `m == c`.
+
+# Experienced YALMIP users will probably be relieved to see that `>=` is only ever used to make a vector nonnegative, never to make a matrix positive semidefinite, as this ambiguity is reliable source of bugs.
+
+# Like CVX, but unlike YALMIP, JuMP can also constrain variables upon creation:
+
+# | variable | JuMP | CVX |
+# | -------- | ------- | ------- |
+# | nonnegative vector    | `@variable(prob, v[1:d] >= 0)` | `variable v(d) nonnegative` |
+# | real PSD matrix    | `@variable(prob, m[1:d,1:d] in PSDCone())` | `variable m(d,d) semidefinite` |
+# | complex PSD matrix    | `@variable(prob, m[1:d,1:d] in HermitianPSDCone())` | `variable m(d,d) complex semidefinite` |
+
+# ## Setting objective
+
+# Like CVX, but unlike YALMIP, JuMP has a specific command for setting an
+# objective function: `@objective(prob, Min, obj)`. Here `obj` is any expression
+# you want to optimize, and `Min` the direction you want. The other possibility is
+# of course `Max`.
+
+# ## Setting solver and options
+
+#  In order to set an optimizer with JuMP you can use at any point the command
+# `set_optimizer(prob, Hypatia.Optimizer)`, where "Hypatia" is an example
+# solver. The list of supported solvers is available [here](https://jump.dev/JuMP.jl/stable/installation/#Supported-solvers).
+# To configure the solver options you use the command `set_attribute(prob, 
+# "verbose", true)` after you set the optimizer, where as an example we were 
+# setting the option verbose to true.
+#
+# A crucial difference is that with JuMP you always have to explicitly choose a
+# solver before optimizing. Both YALMIP and CVX allow you to leave it empty and
+# will try to guess an appropriate solver for the problem.
+
+# ## Optimizing
+
+# Like YALMIP, but unlike CVX, with JuMP you need to explicitly start the
+# optimization, with the command `optimize!(prob)`. The exclamation mark here is
+# a Julia-ism that means the function is modifying its argument, `prob`. 
+
+# ## Extracting variables
+
+# Like YALMIP, but unlike CVX, with JuMP you need to explicitly ask for the value
+# of your variables after optimization is done, with the function call `value(x)`
+# to obtain the value of variable `x`. A subtlety is that unlike YALMIP the 
+# function `value` is only defined for scalars, so for vectors and matrices you
+# need to use Julia broadcasting: `value.(v)`. There's also a specialized function
+# for extracting the value of the objective, `objective_value(prob)`, which is
+# useful if your objective doesn't have a convenient expression.
+
+# ## Dual variables
+
+# Like YALMIP and CVX, JuMP allows you to recover the dual variables. In order to
+# do that the simplest method is to name the constraint you're interested in,
+# e.g., `bob = @constraint(prob, sum(v) == 1)` and then after the optimzation is
+# done call `dual(bob)`. For the general method see [here](https://jump.dev/JuMP.jl/stable/manual/constraints/#Duals-of-variable-bounds).
+
+# ## Reformulating problems
+
+# Perhaps the biggest difference between JuMP and YALMIP and CVX is how far the
+# modeller is willing to go in reformulating the problems you give to it. CVX is
+# happy to reformulate anything it can, even using approximations if your solver
+# cannot handle the problem. YALMIP will only do exact reformulations, but is
+# still fairly adventurous, being willing to reformulate a nonlinear objective
+# in terms of conic constraints, for example. JuMP does no such thing: it only
+# reformulates objectives into objectives, and constraints into constraints, and
+# is fairly conservative at that. As a result, you might need to do some 
+# reformulations manually, for which a good guide is available [here](https://jump.dev/JuMP.jl/stable/tutorials/conic/tips_and_tricks/).
+
+# An interesting possibility that JuMP offers is turning off reformulations
+# altogether, which will reduce latency if they're in fact not needed, or
+# produce an error if they are. To do that, set your optimizer with the option
+# `add_bridges = false`, as e.g. in `set_optimizer(prob, Hypatia.Optimizer, add_bridges=false)`
+
+# ## Vectorization
+
+# In MATLAB it is absolutely essential to "vectorize" your code to obtain
+# acceptable performance. This is because MATLAB is a very slow interpreted
+# language, which sends your commands to extremely fast libraries. When you
+# "vectorize" your code you're minimizing the MATLAB part of the work and sending
+# it to the libraries instead. There's no such duality with Julia. Everything you
+# write and most libraries you use will compile down to LLVM, so "vectorization"
+# has no effect.
+
+# For example, if you are writing a linear program in MATLAB and instead of the
+# usual `constraints = [v >= 0]` you write
+# ```matlab
+# for i=1:n
+#    constraints = [constraints, v(i) >= 0];
+# end
+# ```
+# performance will be horrible. With Julia, on the other hand, there's hardly any
+# difference between `@constraint(prob, v >= 0)` and 
+for i=1:n
+    @constraint(prob, v[i] >= 0)
+end
+# 
+# ## Rosetta stone
+
+# To finish, we'd like to show a complete example of the same optimization 
+# problem being solved with JuMP, YALMIP, and CVX. It is an SDP computing a 
+# simple lower bound on the random robustness of entanglement using the partial
+# transposition criterion. The code is complete apart from the function that does
+# partial transposition. With JuMP we use the function `partialtranspose` from 
+# [Convex.jl](https://jump.dev/Convex.jl/stable/). With both YALMIP and CVX we
+# use the function `PartialTranspose` from [QETLAB](https://github.com/nathanieljohnston/QETLAB).
+
+using JuMP
+using LinearAlgebra
+import Hypatia
+import Convex
+
+function random_state_pure(d)
+    x = randn(ComplexF64, d)
+    y = x*x'
+    return Hermitian(y/tr(y))
+end
+
+function robustness_jump(d)
+    prob = Model()
+    @variable(prob,λ)
+    ρ = random_state_pure(d^2)
+	id = Hermitian(I(d^2))
+	ρT = Hermitian(Convex.partialtranspose(ρ,1,[d,d]))
+    PPT = @constraint(prob, ρT + λ*id in HermitianPSDCone())
+    @objective(prob, Min, λ)
+    set_optimizer(prob, Hypatia.Optimizer, add_bridges = false)
+    set_attribute(prob, "verbose", true)
+    optimize!(prob)
+	WT = dual(PPT)
+	display(value(λ))
+	display(dot(WT,ρT))
+end
+
+# ```matlab
+# function robustness_yalmip(d)
+#     rho = random_state_pure(d^2);
+#     rhoT = PartialTranspose(rho,1,[d d]);
+#     lambda = sdpvar;
+#     constraints = [(rhoT + lambda*eye(d^2) >= 0):'PPT'];
+#     ops = sdpsettings(sdpsettings,'verbose',1,'solver','sedumi');
+#     optimize(constraints,lambda,ops);
+#     WT = dual(constraints('PPT'));
+#     value(lambda)
+#     real(WT(:).'*rhoT(:))
+# end
+# 
+# function robustness_cvx(d)
+#     rho = random_state_pure(d^2);
+#     rhoT = PartialTranspose(rho,1,[d d]);
+#     cvx_begin
+#         variable lambda
+#         dual variable WT
+#         WT : rhoT + lambda*eye(d^2) == hermitian_semidefinite(d^2)
+#         minimise lambda    
+#     cvx_end
+#     lambda
+#     real(WT(:)'*rhoT(:))
+# end
+# 
+# function rho = random_state_pure(d)
+#     x = randn(d,1) + 1i*randn(d,1);
+#     y = x*x';
+#     rho = y/trace(y);
+# end
+# ```
+
--- a/docs/src/tutorials/getting_started/transitioning_from_matlab.jl
+++ b/docs/src/tutorials/getting_started/transitioning_from_matlab.jl
@@ -27,7 +27,7 @@
 # The purpose of this tutorial is to help new users to JuMP who have previously
 # used YALMIP or CVX by comparing and contrasting their different features.
 
-# !!!tip
+# !!! tip
 #    If you have not used Julia before, read the [Getting started with Julia](@ref)
 #    tutorial.
 
--- a/docs/src/tutorials/getting_started/transitioning_from_matlab.jl
+++ b/docs/src/tutorials/getting_started/transitioning_from_matlab.jl
@@ -270,6 +270,19 @@ objective_value(model)
 # end
 # ```
 
+# ## Primal versus dual form
+
+# When you translate some optimization problems from YALMIP or CVX to JuMP, you 
+# might be surprised to see it get much faster or much slower, even if you're 
+# using exactly the same solver. The most likely reason is that YALMIP will
+# always interpret the problem as the dual form, whereas CVX and JuMP will try to
+# interpret the problem in the form most appropriate to the solver. If the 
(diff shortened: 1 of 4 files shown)


docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,280 @@
...
+# ever used to make a vector nonnegative, never to make a matrix positive
+# semidefinite, as this ambiguity is reliable source of bugs.
+
+# Like CVX, but unlike YALMIP, JuMP can also constrain variables upon creation:
+
+# |                    | JuMP                                                 | CVX                                    |
+# | ------------------ | ---------------------------------------------------- | -------------------------------------- |
+# | nonnegative vector | `@variable(model, v[1:d] >= 0)`                      | `variable v(d) nonnegative`            |
+# | real PSD matrix    | `@variable(model, m[1:d,1:d] in PSDCone())`          | `variable m(d,d) semidefinite`         |
+# | complex PSD matrix | `@variable(model, m[1:d,1:d] in HermitianPSDCone())` | `variable m(d,d) complex semidefinite` |
```
odow, line 110: What about mentioning variable bounds, integrality, starting points, etc?

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,280 @@
...
+# still fairly adventurous, being willing to reformulate a nonlinear objective
+# in terms of conic constraints, for example. JuMP does no such thing: it only
+# reformulates objectives into objectives, and constraints into constraints, and
+# is fairly conservative at that. As a result, you might need to do some
+# reformulations manually, for which a good guide is available [here](@ref conic_tips_and_tricks).
+
+# An interesting possibility that JuMP offers is turning off reformulations
+# altogether, which will reduce latency if they're in fact not needed, or
+# produce an error if they are. To do that, set your optimizer with the option
+# `add_bridges = false`, as for example, in `set_optimizer(model, Hypatia.Optimizer, add_bridges=false)`
```
odow, line 173: We probably shouldn't tell people about this just yet. It's an extra for expert thing, not for new users.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,280 @@
...
+    model = Model()
+    @variable(model, λ)
+    ρ = random_state_pure(d^2)
+	id = LinearAlgebra.Hermitian(LinearAlgebra.I(d^2))
+	ρT = LinearAlgebra.Hermitian(Convex.partialtranspose(ρ, 1,[d ,d]))
+    PPT = @constraint(model, ρT + λ * id in HermitianPSDCone())
+    @objective(model, Min, λ)
+    set_optimizer(model, Hypatia.Optimizer; add_bridges = false)
+    set_attribute(model, "verbose", true)
+    optimize!(model)
```
odow, line 232: One thing this example doesn't touch on is solver and result statuses. We should either add `@assert is_solved_and_feasible(model)`, or explain how to query statuses in YALMIP and CVX.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,280 @@
...
+
+# ## Rosetta stone
+
+# To finish this tutorial, we show a complete example of the same optimization
+# problem being solved with JuMP, YALMIP, and CVX. It is an SDP computing a
+# simple lower bound on the random robustness of entanglement using the partial
+# transposition criterion. The code is complete apart from the function that does
+# partial transposition. With JuMP we use the function `partialtranspose` from
+# [Convex.jl](https://jump.dev/Convex.jl/stable/). With both YALMIP and CVX we
+# use the function `PartialTranspose` from [QETLAB](https://github.com/nathanieljohnston/QETLAB).
```
odow, line 207: Using `PartialTranspose` is too complicated. We want something that works out-of-the-box for JuMP and MATLAB.

araujoms: I really want to use this example for three reasons:

1. It illustrates a major advantage of JuMP over Convex, in that you can use whatever functions you want to build your problems, you don't need them to be specifically coded for JuMP. This is why I gave up using CVXPY, fwiw.
2. It tells people where to find a good implementation of partial transposition. It is the thing everyone asks me when I tell them to switch to Julia.
3. It is the canonical example I use to teach duality in my classes and in my (review)[https://arxiv.org/abs/2307.02551]

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
araujoms: I had in mind users that are already experienced with optimization, just not JuMP and Julia. Feel free to remove it, though.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
araujoms: That's a good point, we should add a section on querying solver status, this is something everyone will need.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
araujoms: I don't know what you mean. Variable bound? Like `x <= 1`? Sure I could mention it explicitly. Integrality is what, constraining variables to be bits or integers? That's a single line to add in the table. Starting point is setting the starting point of the optimizer? I wouldn't do that, it hardly ever works.

araujoms:
I don't have write access anymore since this PR is now in your branch.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,280 @@
...
+# A crucial difference is that with JuMP you always have to explicitly choose a
+# solver before optimizing. Both YALMIP and CVX allow you to leave it empty and
+# will try to guess an appropriate solver for the problem.
+
+# ## Optimizing
+
+# Like YALMIP, but unlike CVX, with JuMP you need to explicitly start the
+# optimization, with the command `optimize!(model)`. The exclamation mark here
+# is a Julia-ism that means the function is modifying its argument, `model`.
+
```
araujoms, line 138:
```suggestion

# ## Querying solution status

# After the optimization is done, it's sensible to test for the solution status.
# Like YALMIP and CVX, JuMP provides a solver-independent way to check it, via 
# the command `is_solved_and_feasible(model)`, that returns a Boolean. If it's
# false, you should either look at the status returned by the solver with the 
# command `raw_status(model)`, or re-run the optimization with `verbose` set to
# `true` to get more detail.
```

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,280 @@
...
+#     rhoT = PartialTranspose(rho, 1, [d d]);
+#     cvx_begin
+#         variable lambda
+#         dual variable WT
+#         WT : rhoT + lambda*eye(d^2) == hermitian_semidefinite(d^2)
+#         minimise lambda
+#     cvx_end
+#     lambda
+#     real(WT(:)'*rhoT(:))
+# end
```
araujoms, line 279:
```suggestion
# ```matlab
# function robustness_cvx(d)
#     rho = random_state_pure(d^2);
#     rhoT = PartialTranspose(rho, 1, [d d]);
#     cvx_begin
#         variable lambda
#         dual variable WT
#         WT : rhoT + lambda*eye(d^2) == hermitian_semidefinite(d^2)
#         minimise lambda
#     cvx_end
#     if strcmp(cvx_status,'Solved')
#         lambda
#         real(WT(:)'*rhoT(:))
#     else
#         display('Something went wrong.')
#     end
# end
# 
# function rho = random_state_pure(d)
#     x = randn(d, 1) + 1i * randn(d, 1);
#     y = x * x';
#     rho = y / trace(y);
# end
```

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,280 @@
...
+#     rhoT = PartialTranspose(rho, 1, [d d]);
+#     lambda = sdpvar;
+#     constraints = [(rhoT + lambda*eye(d^2) >= 0):'PPT'];
+#     ops = sdpsettings(sdpsettings, 'verbose', 1, 'solver', 'sedumi');
+#     optimize(constraints, lambda, ops);
+#     WT = dual(constraints('PPT'));
+#     value(lambda)
+#     real(WT(:).'*rhoT(:))
+# end
+# ```
```
araujoms, line 257:
```suggestion
# ```matlab
# function robustness_yalmip(d)
#     rho = random_state_pure(d^2);
#     rhoT = PartialTranspose(rho, 1, [d d]);
#     lambda = sdpvar;
#     constraints = [(rhoT + lambda*eye(d^2) >= 0):'PPT'];
#     ops = sdpsettings(sdpsettings, 'verbose', 1, 'solver', 'sedumi');
#     sol = optimize(constraints, lambda, ops);
#     if sol.problem == 0
#         WT = dual(constraints('PPT'));
#         value(lambda)
#         real(WT(:).'*rhoT(:))
#     else
#         display(['Something went wrong: ', sol.info])
#     end
# end
# 
# function rho = random_state_pure(d)
#     x = randn(d, 1) + 1i * randn(d, 1);
#     y = x * x';
#     rho = y / trace(y);
# end
# ```
```

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,280 @@
...
+    ρ = random_state_pure(d^2)
+	id = LinearAlgebra.Hermitian(LinearAlgebra.I(d^2))
+	ρT = LinearAlgebra.Hermitian(Convex.partialtranspose(ρ, 1,[d ,d]))
+    PPT = @constraint(model, ρT + λ * id in HermitianPSDCone())
+    @objective(model, Min, λ)
+    set_optimizer(model, Hypatia.Optimizer; add_bridges = false)
+    set_attribute(model, "verbose", true)
+    optimize!(model)
+	WT = dual(PPT)
+	return value(λ), LinearAlgebra.dot(WT, ρT)
```
araujoms, line 234:
```suggestion
    if is_solved_and_feasible(model)
        WT = dual(PPT)
        return value(λ), LinearAlgebra.dot(WT, ρT)
    else
        return "Something went wrong: "*raw_status(model)
    end
```

araujoms (commented):
I've added the commands querying the solution status to the JuMP, YALMIP, and CVX codes. Also, note that in MATLAB the main function needs to be the first one, so I changed back the order.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
joaquimg: Some extra details could go into an session by the end of the tutorial to do break the flow.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
joaquimg: Again, instead of removing we could just move it to the end in a "advanced" session

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,300 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE    #src
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER         #src
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,  #src
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE  #src
+# SOFTWARE.                                                                      #src
+
+# # Transitioning from MATLAB
+
+# The purpose of this tutorial is to explain the similarities and differences of
+# JuMP and [YALMIP](https://yalmip.github.io/) or [CVX](https://cvxr.com/cvx/).
```
mlubin, line 24: Please provide some basic introduction like what is MATLAB and what are yalmap/cvx. Make it clear that these are in no way affiliated with JuMP. Use the registered trademark symbol when first mentioning MATLAB, e.g., MATLAB®.

odow:
> I don't have write access anymore since this PR is now in your branch.

Yeah. I changed it here so that we can get a preview: https://jump.dev/JuMP.jl/previews/PR3698/tutorials/getting_started/transitioning_from_matlab/

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,348 @@
...
+        PPT,
+        LinearAlgebra.Hermitian(rhoT + λ * id) in HermitianPSDCone(),
+    )
+    @objective(model, Min, λ)
+    set_optimizer(model, Clarabel.Optimizer)
+    set_attribute(model, "verbose", true)
+    optimize!(model)
+    if is_solved_and_feasible(model)
+        WT = dual(PPT)
+        return value(λ), LinearAlgebra.dot(WT, rhoT)
```
odow, line 287: It's been a while since I did MATLAB, but it's not obvious to me that this is the same as `real(WT(:).' * rhoT(:))`. Do you need a `real()` around the dot? Or even just `real(WT[:]' * rhoT[:])`?

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,380 @@
...
+# ```julia
+# @constraint(model, v >= 0)
+# ```
+#  and
+# ```julia
+# for i in 1:n
+#     @constraint(model, v[i] >= 0)
+# end
+# ```
+
```
araujoms, line 259:
```suggestion

# ## Arbitrary precision

# JuMP is capable of both specifying the optimization problem and solving it in a
# great variety of types, allowing it to work in single, double, quadruple, or 
# arbitrary precision, for example. In contrast, neither YALMIP nor CVX can do
# this, as they are restricted by MATLAB's archaic type system (with the package
# [mpYALMIP](https://github.com/aeroimperial-optimization/mpYALMIP) one can
# modify YALMIP to support solving in multiple precision, but still the problem
# can only be specified in double precision).

# In order to do that, one needs to create the model with the command `model = GenericModel{T}()`,
# where `T` is the type you want to use, for example `BigFloat`, and then set the
# optimizer with the command `set_optimizer(model, Hypatia.Optimizer{T})`. At the
# time of writing the solvers supporting generic types are [Clarabel](https://github.com/oxfordcontrol/Clarabel.jl),
# [COSMO](https://github.com/oxfordcontrol/COSMO.jl), and [Hypatia](https://github.com/jump-dev/Hypatia.jl).

# See [here](https://jump.dev/JuMP.jl/stable/tutorials/conic/arbitrary_precision/)
# for a complete guide on using JuMP with generic types.

```

odow: I'm not sure about this part. We should focus on the core features of the packages that are similar.

There's the arbitrary precision tutorial for experts. But that's not something they need to know when they first transition from MATLAB.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
araujoms: It is not the same thing. `dot(a,b)` is the same thing as `a(:)'*b(:)`. The `real` is only for neatness, the imaginary part is numerical noise. In Julia it's [not necessary](https://github.com/JuliaLang/julia/pull/52318). But you might notice that on the YALMIP code I'm doing `WT(:).' * rhoT(:)`, whereas in the CVX code I'm doing `WT(:)' * rhoT(:)`. This is a workaround for a [bug in YALMIP](https://github.com/yalmip/YALMIP/issues/1375). I didn't remark it here because I thought it would be just noise.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
araujoms: Both the similarities and the differences are necessary, and this is a key difference. MATLAB users won't dream that this this is possible, so they won't find the arbitrary precision tutorial. This section is just intended to take them there.

araujoms:
Maybe it is worth it adding a paragraph explaining `Symmetric` and `Hermitian`?

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
odow: I removed it. I don't want to add it to an advanced section. It doesn't belong in a "transitioning from MATLAB" tutorial.

blegat:
It's also worth mentioning the difference between YALMIP and JuMP according to dualization. YALMIP is always in image form / geometric conic form while in JuMP it depends on the solver. See section 2.1 of the MOI paper for instance. You can refer to https://jump.dev/JuMP.jl/stable/tutorials/conic/dualization/

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,389 @@
...
+    set_optimizer(model, Clarabel.Optimizer)
+    set_attribute(model, "verbose", true)
+    optimize!(model)
+    if is_solved_and_feasible(model)
+        WT = dual(PPT)
+        return value(λ), LinearAlgebra.dot(WT, rhoT)
+    else
+        return "Something went wrong: $(raw_status(model))"
+    end
+end
```
araujoms, line 328: With  #3695 merged maybe you could go back to the previous version? I think it's much nicer.
```suggestion
function robustness_jump(d)
    rho = random_state_pure(d^2)
    id = LinearAlgebra.Hermitian(LinearAlgebra.I(d^2))
    rhoT = LinearAlgebra.Hermitian(partial_transpose(rho, 1, [d, d]))
    model = Model()
    @variable(model, λ)
    @constraint(model, PPT, rhoT + λ * id in HermitianPSDCone())
    @objective(model, Min, λ)
    set_optimizer(model, Clarabel.Optimizer)
    set_attribute(model, "verbose", true)
    optimize!(model)
    if is_solved_and_feasible(model)
        WT = dual(PPT)
        return value(λ), LinearAlgebra.dot(WT, rhoT)
    else
        return "Something went wrong: $(raw_status(model))"
    end
end
```

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
odow: I added rows for binary and integer, and a section for bounds. I decided not to add starting points.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
odow: Done

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
odow: Done

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
odow: I don't want to add this to a a "Transitioning from MATLAB" tutorial. Also, the arbitrary precision is not a panacea. Users should not just use a solver with more precision if they run into numerical difficulties.

odow:
> Maybe it is worth it adding a paragraph explaining Symmetric and Hermitian?

I'd like to keep this tutorial pretty tightly scoped to focus on JuMP/YALMIP/CVX.

It is not the place for this tutorial to be a cheatsheet of every difference between Julia and MATLAB. Symmetric and Hermitian should be reasonably self-evident. If it isn't users can always look up the Julia documentation.

odow:
> It's also worth mentioning the difference between YALMIP and JuMP according to dualization.

Is this necessary for users transitioning from MATLAB? It seems in the weeds. I'd prefer we left it out and got feedback first.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
araujoms: The point of arbitrary precision is not to solve numerical problems, but to obtain higher precision. In my [latest paper](https://arxiv.org/abs/2311.18707) I had to use an arbitrary precision solver because MOSEK just couldn't give me 8 digits, even when I tweaked its precision, and I really needed 8 digits. With SDPA-GMP I could easily get more than 20 digits. Ironically enough, SDPA-GMP was much more vulnerable than MOSEK to lack of strict feasibility.

odow: I understand your point. But it doesn't belong in this tutorial. If there is an argument to be made, it is that we should make the arbitrary precision more widely known for _all_ users, not just those coming from MATLAB.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,389 @@
...
+# # Transitioning from MATLAB
+
+# [YALMIP](https://yalmip.github.io/) and [CVX](https://cvxr.com/cvx/) are two
+# packages for mathematical optimization in [MATLAB®](https://mathworks.com/products/matlab.html).
+# They are independently developed and are in no way affiliated with JuMP.
+
+# The purpose of this tutorial is to help new users to JuMP who have previously
+# used YALMIP or CVX by comparing and contrasting their different features.
+
+# !!!tip
```
odow, line 30:
```suggestion
# !!! tip
```

araujoms:
> It is not the place for this tutorial to be a cheatsheet of every difference between Julia and MATLAB. Symmetric and Hermitian should be reasonably self-evident. If it isn't users can always look up the Julia documentation.

The difference is that with YALMIP and CVX one needs to do `a = 0.5*(a+a')` over and over again, to make sure your matrices are *exactly* Hermitian, whereas with JuMP this doesn't matter, you just have to wrap them in `Hermitian`.

blegat:
> Is this necessary for users transitioning from MATLAB? It seems in the weeds. I'd prefer we left it out and got feedback first.

This is the n°1 comment people have when coming from YALMIP (see for instance https://discourse.julialang.org/t/yalmip-vs-jump/30776/10 and https://discourse.julialang.org/t/jump-sumofsquares-generate-sdps-with-18x-more-constraints-than-yalmip/65377/15). It can just be a small note refering to the relevant part of the doc.

araujoms:
I also think it would be worth it such a remark, it has a large impact on performance.

docs/src/tutorials/getting_started/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,389 @@
...
+# ```julia
+# @constraint(model, v >= 0)
+# ```
+#  and
+# ```julia
+# for i in 1:n
+#     @constraint(model, v[i] >= 0)
+# end
+# ```
+
```
araujoms, line 272:
```suggestion

# ## Primal versus dual form

# When you translate some optimization problem from YALMIP or CVX to JuMP you 
# might be surprised to see it get much faster or much slower, even if you're 
# using exactly the same solver. The most likely reason is that YALMIP will
# always interpret the problem as the dual form, whereas CVX and JuMP will try to
# interpret the problem in the form most appropriate to the solver. If the 
# problem is more naturally formulated in the primal form it is likely that
# YALMIP's performance will suffer, or if JuMP gets it wrong its performance will
# suffer. It might be worth trying both primal and dual forms if you're having
# trouble, which can be done automatically with the package [Dualization](https://jump.dev/JuMP.jl/stable/packages/Dualization/#Dualization.jl).
#
# For an in-depth explanation of this issue see the [dualization](https://jump.dev/JuMP.jl/stable/tutorials/conic/dualization/) tutorial.
```

docs/src/tutorials/transitioning/transitioning_from_matlab.jl
```diff
@@ -0,0 +1,447 @@
...
+    model = Model()
+    @variable(model, λ)
+    @constraint(model, PPT, rhoT + λ * id in HermitianPSDCone())
+    @objective(model, Min, λ)
+    set_optimizer(model, Clarabel.Optimizer)
+    set_attribute(model, "verbose", true)
+    optimize!(model)
+    if is_solved_and_feasible(model)
+        WT = dual(PPT)
+        return value(λ), real(LinearAlgebra.dot(WT, rhoT))
```
araujoms, line 382: Why did add `real` here? It's redundant, `dot` already returns a real number from Julia 1.11 onwards.

odow: Because the docs currently build with Julia 1.10 and we support Julia 1.6. Also, v1.11 isn't released yet.

araujoms: What's the problem? It doesn't cause any error with older versions of Julia, the output is just slightly less neat.

docs/src/tutorials/transitioning/transitioning_from_matlab.jl
```diff
@@ -270,13 +270,58 @@ objective_value(model)
...
+
+x_sym = LinearAlgebra.Symmetric(x, :L)
+
+#-
+
+x_sym = LinearAlgebra.Symmetric(x, :U)
+
+# The same applies for Hermitian matrices, using `LinearAlgebra.Hermitian` and
+# `LinearAlgebra.ishermitian`.
+
```
araujoms, line 317: What I wanted to point out here is that with MATLAB you have to be paranoid about making your matrices exactly Hermitian, in order to not get subtle errors, whereas with JuMP you just need to wrap them and you're good to go.

But now I've tested it, and it turns out that it's not true. If you try to constrain a non-Hermitian matrix to be in HermitianPSDCone() you get an error message, as I expected, but if you try to constrain a non-Symmetric matrix to be in PSDCone() JuMP will let you (and fail afterwards, of course). Is that intentional?

docs/src/tutorials/transitioning/transitioning_from_matlab.jl
odow: `dot` currently returns a `Complex{Float64}` instead of a `Float64`. Your MATLAB examples return a `Float64`.

```julia
julia> import LinearAlgebra

julia> A = LinearAlgebra.Hermitian(rand(ComplexF64, (3, 3)))
3×3 LinearAlgebra.Hermitian{ComplexF64, Matrix{ComplexF64}}:
 0.468242+0.0im       0.296799+0.49972im   0.621408+0.148878im
 0.296799-0.49972im   0.413853+0.0im       0.677598+0.875476im
 0.621408-0.148878im  0.677598-0.875476im   0.59341+0.0im

julia> LinearAlgebra.dot(A, A)
4.6860981128420285 + 0.0im

julia> real(LinearAlgebra.dot(A, A))
4.6860981128420285
```

araujoms: I know. This is not an error, and will change anyway in Julia 1.11.

odow: > I know

I know because I saw your PR :smile:

My point is that I personally would argue that changing from `ComplexF64` to `Float64` is a breaking change, and the fact it will change in Julia v1.11 is not compelling because we still have many JuMP users on a range of old versions. I want something that works across all versions, not just some upcoming one. We cannot assume that new users from MATLAB will use a recent version of Julia. See, for example, a [discourse user on Julia v1.5](https://discourse.julialang.org/t/fail-to-update-gurobi-jl-for-using-gurobi-11-0/111225).

docs/src/tutorials/transitioning/transitioning_from_matlab.jl
araujoms: Why? Is there any case where you would need that? As far as I can see this can either generate less efficient code or give you an error.

docs/src/tutorials/transitioning/transitioning_from_matlab.jl
araujoms: I really don't understand the problem. It works. It's just slightly uglier. And moreover it's just example code in a tutorial. Nothing depends on whether its return value is `Float64` or `ComplexF64`.

odow: It doesn't work. One return a `ComplexF64`, and the other returns a `Float64`.

araujoms: Why is this a problem?

docs/src/tutorials/transitioning/transitioning_from_matlab.jl
araujoms: I was just curious, I wasn't suggesting to change this, it would clearly be a breaking change.

Although now that you brought it up, I will definitely suggest removing it when the time comes for JuMP 2.0.

odow: > I will definitely suggest removing it when the time comes for JuMP 2.0

My goal is not to release JuMP 2.0 if at all possible. If we did, it would be for a minimal set of breaking changes, not just to remove some things we don't like. The only candidate was the nonlinear changes, but we seem to have managed to sneak them in with minimal disruption. The JuMP developers have no plans for JuMP 2.0.

odow:
Merging because this is good enough for now. If needed, we can make tweaks in smaller PRs, instead of hashing everything out here. (There are already 56 comments!)

Issue #3701, opened 2024-03-05
Documentation Request: List whether a solver supports Indicator Constraints


schlichtanders:
Using DisjunctiveProgramming.jl I was told that Indicator Constraints may actually be what I am looking for.

However unfortunately, Indicator Constraints don't have a bridge to Big-M notation, not are they listed on the solver's summary page whether it is supported or not.

It would be great to either 
- have a bridge
- or update the documentation to name whether a solver supports Indicator Constraints

blegat:
You should find it in this list: https://jump.dev/JuMP.jl/stable/packages/Gurobi/#MathOptInterface-API
However, it seems to be supported in the MOI wrapper of Gurobi but not in this list. @odow will be able to help

schlichtanders:
I was just pointed to the fact, that a bridge actually exists.

I was just falsely assuming there is no bridge. Probably because I read the Tips and Tricks section where Indicator constraints are mentioned to be only supported by a couple of solvers.

https://jump.dev/JuMP.jl/stable/tutorials/linear/tips_and_tricks/#Indicator-constraints

It would be great if this section is updated/deleted. Plus that the [section about Disjunctions](https://jump.dev/JuMP.jl/stable/tutorials/linear/tips_and_tricks/#Disjunctions) gets an alternative implementation using Indicator Constraints.

odow:
> not are they listed on the solver's summary page whether it is supported or not.

This is an oversight on my part. I guess the script I ran didn't catch indicators :cry:

> It would be great if this section is updated/deleted

To clarify, what section?

> Plus that the [section about Disjunctions](https://jump.dev/JuMP.jl/stable/tutorials/linear/tips_and_tricks/#Disjunctions) gets an alternative implementation using Indicator Constraints.

I will add

schlichtanders:
Thank you for your help.

I mean that the [Indicator section](https://jump.dev/JuMP.jl/stable/tutorials/linear/tips_and_tricks/#Indicator-constraints) does not need the Trick2 part, the manual workaround in case a solver does not support it. Instead mention that the MOI bridge already does that, as soon as one restricts the variables to a finite domain.

PR #3703, opened 2024-03-06
[docs] clarify that JuMP can reformulate indicator constraints

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3701

### DIFF
--- a/docs/src/tutorials/linear/tips_and_tricks.jl
+++ b/docs/src/tutorials/linear/tips_and_tricks.jl
@@ -234,26 +234,28 @@ M = 100
 
 # ### Trick 1
 
-# Some solvers have native support for indicator constraints.
+# Some solvers have native support for indicator constraints. In addition, if
+# the variables involved have finite domains, then JuMP can automatically
+# reformulate an indicator into a mixed-integer program.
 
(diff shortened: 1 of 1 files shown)


odow:
I guess the `M` is chosen by the user. Whereas JuMP cannot automatically perform the same reformulation only if we know the bounds to infer a suitable `M`.

In practice, all decision variables in reality have bounds (you cannot have infinite anything). It's just that the user may not directly specify them, or know an appropriate value.

remi-garcia:
I just saw this with the new release. The ``Trick 2'' bothers me as it is actually not possible to use big M constraints if variables have not a finite domain. Doing so actually add bounds.

MWE for which this change the result:
```julia
model = Model()
@variable(model, x[1:2])
@variable(model, z, Bin)
@constraint(model, z --> {sum(x) <= 1})
@objective(model, Max, sum(x))
```
Termination status: unbounded while
```julia
model = Model()
@variable(model, x[1:2])
@variable(model, z, Bin)
M = 100   # the specific value does not matter
@constraint(model, sum(x) <= 1 + M * (1 - z))
@objective(model, Max, sum(x))
```
Termination status: optimal, objective value: M+1

PR #3707, opened 2024-03-10
Bump compat of DimensionalData

odow:
We can fix the printing. But https://github.com/rafaqz/DimensionalData.jl/issues/664 is a blocker.


rafaqz:
I forgot to up the lower bound. Its fixed and the other version will be yanked.

odow:
> I forgot to up the lower bound

I don't really understand why this would help? It's failing on nightly, not on an older version like 1.6

> the other version will be yanked.

There is no need to yank patch versions

odow:
Your `master` is still failing: https://github.com/rafaqz/DimensionalData.jl/actions/runs/8224367439/job/22488102829

rafaqz:
Thats nightly failing

odow:
> Thats nightly failing

Sure, but I don't want to bump the compat because then JuMP will fail tests.

It looks like you added the problematic https://github.com/rafaqz/DimensionalData.jl/pull/585 only yesterday? We shouldn't touch `Base` internals.

rafaqz:
To clarify the yank is for 1.8 and 1.9... they will also break for unrelated reasons (like effects system macros). I just forgot about them.

I misunderstood your comment as nightly will fail anyway for other reasons even if I fix that particular bug.

rafaqz:
DD unavoidably hits a few internals where Base doesnt let us pass through dimension types.

But that new one was probably a mistake, its just very tricky to get dispatch working. Im working around it now.

(Also this version bump is huge, like 40 closed issues. So might be messy for a few days.)

odow:
> I misunderstood your comment as nightly will fail anyway for other reasons even if I fix that particular bug.

It's okay if your tests fail because of printing, etc. But ideally, they shouldn't cause the basic tests in JuMP to fail:

https://github.com/jump-dev/JuMP.jl/blob/d7c45a979e01c6d91cd489873bbd8982ad2518d1/ext/test_DimensionalData.jl

rafaqz:
For sure. I will fix that asap.

Hopefuly the other ones out if my control dont affect your tests.

Project.toml
```diff
@@ -20,7 +20,7 @@ DimensionalData = "0703355e-b756-11e9-17c0-8b28908087d0"
 JuMPDimensionalDataExt = "DimensionalData"
 
 [compat]
-DimensionalData = "0.24, 0.25"
+DimensionalData = "0.24, 0.25, 0.26"
```
odow, line 23:
```suggestion
DimensionalData = "0.24, 0.25, 0.26.2"
```

docs/Project.toml
```diff
@@ -41,7 +41,7 @@ CDDLib = "=0.9.2"
 CSV = "0.10"
 Clarabel = "=0.6.0"
 DataFrames = "1"
-DimensionalData = "0.24"
+DimensionalData = "0.26"
```
odow, line 44:
```suggestion
DimensionalData = "0.26.2"
```

PR #3711, opened 2024-03-13
[docs] add deprecation notes to the legacy nonlinear interface docstrings

odow:
I think this has been in enough use now that we can recommend it more strongly.

Preview: https://jump.dev/JuMP.jl/previews/PR3711/api/JuMP/#@NLconstraint


blegat (approved):

odow:
I wonder if we should also remove the warning:

image

We've had lots of people use this now, and I've only heard good things.

Issue #3566, opened 2023-11-07
Type instability in `set_normalized_rhs`


franckgaga:
Hi,

I'm trying to get rid of the type instabilities in `ModelPredictiveControl.jl` to improve the performance (it's important for real-time control). I'm not sure if the impact on speed is major here, but I noticed that my code spend a lot of time in `set_normalized_rhs` function (with `@profview` in VS Code). I need to call this function at each discrete-time, before calling `optimize!`.

I found that there is a type instability in the function using `JET.jl`. Here is a MWE:
```julia
using JuMP, DAQP, JET
model = Model(DAQP.Optimizer)
@variable(model, x1)
@variable(model, x2)
@objective(model, Min, 0.5*(x1^2+x2^2)+x1+x2)
A = [1 1; -1 -1; 1 -1; -1 1]
b = [3; 3; 4; 4] 
@constraint(model, linconstraint, A*[x1;x2] .<= b)
b = [5; 5; 5; 5]
main(linconstraint, b) = set_normalized_rhs.(linconstraint, b)
@report_opt vscode_console_output=stdout main(linconstraint, b)
```
the result is:
```julia
═════ 1 possible error found ═════
┌ main(linconstraint::Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.LessThan{Float64}}, ScalarShape}}, b::Vector{Int64}) @ Main /home/francis/Dropbox/Programmation/Julia/TestMPC/src/yo.jl:10
│┌ materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(set_normalized_rhs), Tuple{Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.LessThan{Float64}}, ScalarShape}}, Vector{Int64}}}) @ Base.Broadcast ./broadcast.jl:873
[...]

PR #3716, opened 2024-03-22
Add batched modification methods

joaquimg:
Modify problems bit by bit can be surprisingly slow in some solvers (like Xpress). Batch modifications like this solve the problems. I was working on a model that changes from 750s to 150s build+modify+solve time after using batch modifications.

`rhs` would be weird in plural as `rhss`, I don't know what to do about it.

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -784,9 +784,9 @@ function set_normalized_rhs(
     F<:Union{MOI.ScalarAffineFunction{T},MOI.ScalarQuadraticFunction{T}},
 }
     MOI.set(
-        owner_model(first(constraints)),
+        backend(owner_model(first(constraints))),
         MOI.ConstraintSet(),
-        constraints,
+        index.(constraints),
         S.(convert.(T, values)),
     )
     return
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -794,10 +794,10 @@ julia> model = Model();
 julia> @variable(model, x);
 
 julia> @constraint(model, con1, 2x + 1 <= 2)
-con : 2 x ≤ 1
+con1 : 2 x ≤ 1
 
 julia> @constraint(model, con2, 3x + 2 <= 4)
-con : 3 x ≤ 2
+con2 : 3 x ≤ 2
 
 julia> set_normalized_rhs([con1, con2], [4, 5])
 
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -805,7 +805,7 @@ julia> con1
 con1 : 2 x ≤ 4
 
 julia> con2
-con1 : 3 x ≤ 5
+con2 : 3 x ≤ 5
 ```
 """
 function set_normalized_rhs(
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -773,6 +773,41 @@ function set_normalized_rhs(
     return
 end
 
+"""
+    set_normalized_rhs(
+        constraints::AbstractVector{<:ConstraintRef},
+        values::AbstractVector{<:Number}
+    )
+
+Set the right-hand side terms of all `constraints` to `values`.
+
+Note that prior to this step, JuMP will aggregate all constant terms onto the
+right-hand side of the constraint. For example, given a constraint `2x + 1 <=
+2`, `set_normalized_rhs([con], [4])` will create the constraint `2x <= 4`, not `2x +
+1 <= 4`.
+
+## Example
+
+```jldoctest; filter=r"≤|<="
+julia> model = Model();
+
+julia> @variable(model, x);
+
+julia> @constraint(model, con1, 2x + 1 <= 2)
+con : 2 x ≤ 1
+
+julia> @constraint(model, con2, 3x + 2 <= 4)
+con : 3 x ≤ 2
+
+julia> set_normalized_rhs([con1, con2], [4, 5])
+
+julia> con1
+con1 : 2 x ≤ 4
+
+julia> con2
+con1 : 3 x ≤ 5
+```
+"""
 function set_normalized_rhs(
     constraints::AbstractVector{
         <:ConstraintRef{<:AbstractModel,MOI.ConstraintIndex{F,S}},
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -731,7 +731,7 @@ function add_constraint(
 end
 
 """
-    set_normalized_rhs(constraint::ConstraintRef, value)
+    set_normalized_rhs(constraint::ConstraintRef, value::Number)
 
 Set the right-hand side term of `constraint` to `value`.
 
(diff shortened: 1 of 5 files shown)


src/objective.jl
```diff
@@ -491,11 +493,103 @@ function _set_objective_coefficient(
     return
 end
 
+"""
+    set_objective_coefficients(
```
odow, line 497:
```suggestion
    set_objective_coefficient(
```

Shouldn't we just keep the same name and provide a new method?

joaquimg: That is certainly an option. I would be totally fine with it. It might even reduce the number of typos in the code, with which I suffered myself.

odow: Yeah. We have dispatch. Let's use it instead of adding more functions.

src/variables.jl
```diff
@@ -2596,6 +2653,8 @@ julia> set_normalized_coefficients(con, x, [(1, 2.0), (2, 5.0)])
 julia> con
 con : [2 x, 5 x] ∈ MathOptInterface.Nonnegatives(2)
 ```
+
+See also [`set_normalized_coefficient`](@ref).
```
joaquimg, line 2657:
```suggestion
```

src/constraints.jl
```diff
@@ -773,6 +773,25 @@ function set_normalized_rhs(
     return
 end
 
+function set_normalized_rhs(
```
odow, line 776: Docstring?

odow (approved):

src/objective.jl
```diff
@@ -491,11 +491,87 @@ function _set_objective_coefficient(
...
+"""
+function set_objective_coefficient(
+    model::GenericModel{T},
+    variables::AbstractVector{<:GenericVariableRef{T}},
+    coeffs::AbstractVector{<:Real},
+) where {T}
+    if _nlp_objective_function(model) !== nothing
+        error("A nonlinear objective is already set in the model")
+    elseif length(variables) != length(coeffs)
+        msg = "The number of variables and coefficients must match"
```
blegat, line 532: Give the number of variables and coefficients, it can spare some debugging time to people :)

src/objective.jl
```diff
@@ -572,3 +648,91 @@ function _set_objective_coefficient(
...
+function set_objective_coefficient(
+    model::GenericModel{T},
+    variables_1::AbstractVector{<:GenericVariableRef{T}},
+    variables_2::AbstractVector{<:GenericVariableRef{T}},
+    coeffs::AbstractVector{<:Real},
+) where {T}
+    if _nlp_objective_function(model) !== nothing
+        error("A nonlinear objective is already set in the model")
+    elseif !(length(variables_1) == length(variables_2) == length(coeffs))
+        msg = "The number of variables and coefficients must match"
```
blegat, line 690: Same here

src/objective.jl
```diff
@@ -491,11 +491,87 @@ function _set_objective_coefficient(
...
+    coeffs::AbstractVector{<:Real},
+) where {T}
+    if _nlp_objective_function(model) !== nothing
+        error("A nonlinear objective is already set in the model")
+    elseif length(variables) != length(coeffs)
+        msg = "The number of variables and coefficients must match"
+        throw(DimensionMismatch(msg))
+    end
+    F = objective_function_type(model)
+    _set_objective_coefficient(model, variables, convert.(T, coeffs), F)
```
blegat, line 536: `convert` will allocate a new vector even if it's already the right type, maybe we can have some utility `_lazy_convert` that avoids that

src/objective.jl
```diff
@@ -572,3 +648,91 @@ function _set_objective_coefficient(
...
+    variables_2::AbstractVector{<:GenericVariableRef{T}},
+    coeffs::AbstractVector{<:Real},
+) where {T}
+    if _nlp_objective_function(model) !== nothing
+        error("A nonlinear objective is already set in the model")
+    elseif !(length(variables_1) == length(variables_2) == length(coeffs))
+        msg = "The number of variables and coefficients must match"
+        throw(DimensionMismatch(msg))
+    end
+    coeffs_t = convert.(T, coeffs)::AbstractVector{<:T}
```
blegat, line 693:
```suggestion
    coeffs_t = convert.(T, coeffs)::AbstractVector{T}
```

blegat: Here we want to keep the copy since we may modify it for diagonal entries

blegat (approved):
Looks good, only minor comments

src/objective.jl
```diff
@@ -491,11 +491,87 @@ function _set_objective_coefficient(
...
+5 x + 4 y + 1
+```
+"""
+function set_objective_coefficient(
+    model::GenericModel{T},
+    variables::AbstractVector{<:GenericVariableRef{T}},
+    coeffs::AbstractVector{<:Real},
+) where {T}
+    if _nlp_objective_function(model) !== nothing
+        error("A nonlinear objective is already set in the model")
```
odow, line 530:
```suggestion
        error("A nonlinear objective is set in the model")
```

pedroripper:
Hey everyone! 

I tried to use this new dispatch for `set_normalized_coefficient` but got an error... here is a MWE

```julia
using JuMP
using HiGHS

@variable(model, x[1:3] >= 0)
@constraint(model, c[i in 1:3], x[i] == i)

constraint_coeffs = Vector{JuMP.ConstraintRef}()
variables = Vector{JuMP.VariableRef}()
coefficients = Vector{Float64}()

for i in 1:3
    push!(constraint_coeffs, c[i])
    push!(variables, x[i])
    push!(coefficients, 5)
end

set_normalized_coefficient(constraint_coeffs, variables, coefficients)
``` 

And when I am more specific about the type of the constraint reference vector, it works
```julia
using JuMP
using HiGHS

@variable(model, x[1:3] >= 0)
@constraint(model, c[i in 1:3], x[i] == i)

constraint_coeffs =  Vector{JuMP.ConstraintRef{JuMP.Model, MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, MOI.EqualTo{Float64}}, ScalarShape}}()
variables = Vector{JuMP.VariableRef}()
coefficients = Vector{Float64}()

for i in 1:3
    push!(constraint_coeffs, c[i])
    push!(variables, x[i])
    push!(coefficients, 5)
end

set_normalized_coefficient(constraint_coeffs, variables, coefficients)

```

src/objective.jl
odow: We really shouldn't care about copies like this. I don't want to complicate things. It's a very minor cost.

franckgaga:
@joaquimg @odow @blegat I just wanted to thank you for this excellent contribution, you are doing amazing work! 

`ModelPredictiveControl.jl` uses `OSQP.jl` by default and it also has this issue. I raises an issue (#3566) in which I was suspecting a type instability for the bad performance. It was not the culprit at all. Using the new `set_normalized_rhs` and `set_objective_coefficient` methods with vectors drastically improved the performance ! :partying_face:

PR #3720, opened 2024-04-01
[docs] add bounds to big-M examples

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3703#issuecomment-2029566988

cc @remi-garcia


remi-garcia:
Thanks for the quick commit.

However, this is not consistent anymore with "If the solver doesn't support indicator constraints and the variables do not have a finite domain, you an use the big-M trick."

I believe that something like "If the solver doesn't support indicator constraints, you can use the big-M trick (which requires variables to have a finite domain)." would be better suited.

PR #3721, opened 2024-04-02
[docs] fix bounds in big-M examples of tips_and_tricks.jl

odow:
Follow-up to #3720

@remi-garcia how is this?

### DIFF
--- a/docs/src/tutorials/linear/tips_and_tricks.jl
+++ b/docs/src/tutorials/linear/tips_and_tricks.jl
@@ -255,7 +255,7 @@ model = Model();
 # ### Trick 2
 
 # If the solver doesn't support indicator constraints and the variables do not
-# have a finite domain, you an use the big-M trick.
+# have a finite domain, you can use the big-M trick.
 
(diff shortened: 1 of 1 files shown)


remi-garcia:
It is better, it is consistent that way. But I am wondering why in
> In addition, if the variables involved have *finite domains*, then JuMP can automatically reformulate an indicator into a mixed-integer program.

variables must have a finite domain and not just upper bounds in case of `<=` and lower bounds in case of `>=`.

I will try to have a look at the JuMP reformulation as I think that it is possible to have JuMP to automatically reformulate an indicator into a mixed-integer program whenever there is the right bound and not just with finite domains.

odow:
> I will try to have a look at the JuMP reformulation as I think that it is possible to have JuMP to automatically reformulate an indicator into a mixed-integer program whenever there is the right bound and not just with finite domains.

Yes, that's possible. I just went with the simplest thing.

The relevant bridge is https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Bridges/Constraint/bridges/indicator_to_milp.jl

PR #3724, opened 2024-04-08
Fix adjoint of GenericNonlinearExpr

### LINKED ISSUE #3723: Y' when Y is of type ::Matrix{NonlinearExpr}
Hi,

I encountered an issue which might be a bug (or not).

```julia

using JuMP

model = Model()

@variables(model, begin
    x[1:2]
    X[1:2, 1:2]
end)

@expression(model, Y, [cos(x[1]) 0 ; 0 x[2]])

@expression(model, XY, X*Y') # this returns an error
# @expression(model, YX, Y*X') # this is ok

```

> ERROR: MethodError: promote_operation(::typeof(*), ::Type{VariableRef}, ::Type{Union{}}) is ambiguous.
Candidates:
  promote_operation(op::typeof(*), ::Type{S}, A::Type{<:AbstractArray{T}}) where {S, T}
    @ MutableArithmetics ~/.julia/packages/MutableArithmetics/xVyia/src/implementations/LinearAlgebra.jl:179
  promote_operation(::Union{typeof(*), typeof(+), typeof(-)}, V::Type{<:AbstractVariableRef}, C::Type{<:Union{Number, LinearAlgebra.UniformScaling}})
    @ JuMP ~/.julia/packages/JuMP/glJ0u/src/mutable_arithmetics.jl:38
Possible fix, define
  promote_operation(::typeof(*), ::Type{S}, ::Type{Union{}}) where S<:AbstractVariableRef

Note that with a transpose it works well.

Actually, Y' is not defined:

```julia
Y'
```

> 2×2 adjoint(::Matrix{NonlinearExpr}) with eltype Union{}:
Error showing value of type LinearAlgebra.Adjoint{Union{}, Matrix{NonlinearExpr}}:
ERROR: MethodError: no method matching conj(::NonlinearExpr)

I'm on julia v1.10.2 and JuMP v1.21.0

odow:
Closes #3723

So I didn't really think through the implications of this prior.

We kinda punted on the question of whether `ScalarNonlinearFunction` is `<:Real` or `<:Complex`. But I think it's time to pick `<:Real`, and if we need it, we can introduce something like:
```julia
struct ComplexScalarNonlinearFunction
    real::ScalarNonlinearFunction
    imag::ScalarNonlinearFunctionn
end
```

### DIFF
--- a/test/test_nlp_expr.jl
+++ b/test/test_nlp_expr.jl
@@ -1059,4 +1059,37 @@ function test_nlp_matrix_adjoint()
     return
 end
 
+function test_error_complex_literal_pow()
+    model = Model()
+    @variable(model, x in ComplexPlane())
+    @test_throws(
+        ErrorException(
+            "Cannot build `GenericNonlinearExpr` because a term is complex-" *
+            "valued: `($x)::$(typeof(x))`",
+        ),
+        x^3,
+    )
+    return
+end
+
+function test_error_nonlinear_expr_complex_constructor()
+    model = Model()
+    @variable(model, x in ComplexPlane())
+    @test_throws(
+        ErrorException(
+            "Cannot build `GenericNonlinearExpr` because a term is complex-" *
+            "valued: `($x)::$(typeof(x))`",
+        ),
+        NonlinearExpr(:^, Any[x, 3]),
+    )
+    @test_throws(
+        ErrorException(
+            "Cannot build `GenericNonlinearExpr` because a term is complex-" *
+            "valued: `($x)::$(typeof(x))`",
+        ),
+        NonlinearExpr(:^, x, 3),
+    )
+    return
+end
+
 end  # module
--- a/test/test_nlp_expr.jl
+++ b/test/test_nlp_expr.jl
@@ -1035,4 +1035,28 @@ function test_convert_float_nonlinear_expr()
     return
 end
 
+function test_nlp_adjoint()
+    model = Model()
+    @variable(model, x)
+    y = sin(x)
+    @test y' === y
+    @test conj(y) === y
+    @test real(y) === y
+    @test isequal_canonical(imag(y), zero(y))
+    @test isequal_canonical(abs2(y), y^2)
+    @test isreal(y)
+    return
+end
+
+function test_nlp_matrix_adjoint()
+    model = Model()
+    @variable(model, x[1:2])
+    y = sin.(x)
+    @test isequal_canonical(
+        @expression(model, expr, y' * y),
+        NonlinearExpr(:+, Any[0.0, y[2] * y[2], y[1] * y[1]]),
+    )
+    return
+end
+
 end  # module
--- a/test/test_nlp_expr.jl
+++ b/test/test_nlp_expr.jl
@@ -1054,7 +1054,7 @@ function test_nlp_matrix_adjoint()
     y = sin.(x)
     @test isequal_canonical(
         @expression(model, expr, y' * y),
-        NonlinearExpr(:+, Any[0.0, y[2] * y[2], y[1] * y[1]]),
+        NonlinearExpr(:+, Any[0.0, y[2]*y[2], y[1]*y[1]]),
     )
     return
 end
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -291,6 +291,12 @@ function Base.one(::Type{GenericNonlinearExpr{V}}) where {V}
     return GenericNonlinearExpr{V}(:+, 1.0)
 end
 
+Base.conj(x::GenericNonlinearExpr) = x
+Base.real(x::GenericNonlinearExpr) = x
+Base.imag(x::GenericNonlinearExpr) = zero(x)
+Base.abs2(x::GenericNonlinearExpr) = x^2
+Base.isreal(::GenericNonlinearExpr) = true
+
 # Univariate operators
 
 _is_real(::Any) = false
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -334,6 +334,7 @@ for f in MOI.Nonlinear.DEFAULT_UNIVARIATE_OPERATORS
         end
     elseif isdefined(Base, f)
         @eval function Base.$(f)(x::AbstractJuMPScalar)
+            _throw_if_not_real(x)
             return GenericNonlinearExpr{variable_ref_type(x)}($op, x)
         end
     elseif isdefined(MOI.Nonlinear, :SpecialFunctions)
@@ -341,6 +342,7 @@ for f in MOI.Nonlinear.DEFAULT_UNIVARIATE_OPERATORS
         SF = MOI.Nonlinear.SpecialFunctions
         if isdefined(SF, f)
             @eval function $(SF).$(f)(x::AbstractJuMPScalar)
+                _throw_if_not_real(x)
                 return GenericNonlinearExpr{variable_ref_type(x)}($op, x)
             end
         end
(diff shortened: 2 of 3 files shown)


test/test_nlp_expr.jl
```diff
@@ -1035,4 +1035,28 @@ function test_convert_float_nonlinear_expr()
...
+    return
+end
+
+function test_nlp_matrix_adjoint()
+    model = Model()
+    @variable(model, x[1:2])
+    y = sin.(x)
+    @test isequal_canonical(
+        @expression(model, expr, y' * y),
+        NonlinearExpr(:+, Any[0.0, y[2] * y[2], y[1] * y[1]]),
```
odow, line 1057:
```suggestion
        NonlinearExpr(:+, Any[0.0, y[2]*y[2], y[1]*y[1]]),
```

blegat:
Note sure about this. Here, you could even have a `GenericAffExpr{ComplexF64}` inside so it's weird to say it's real.
Why not implement these so that it works for the most common operators and error for user-defined operators or even for our own operators for which we haven't done anything yet ? I think erroring is better than unexpected behavior here.

blegat:
Actually, we could also do
```julia
function Base.real(x::GenericNonlinearExpr{V}) where {V}
    return GenericNonlinearExpr{V}(:real, x)
end
```
and support it in the AD

blegat:
In view of https://github.com/jump-dev/MathOptInterface.jl/pull/2468, we probably indeed need a way to tell if we have a guarantee that a `MOI.ScalarNonlinearFunction` will always have real output or if it may be complex. So maybe we also need `MOI.ComplexScalarNonlinearFunction` which is has the same fields has `ScalarNonlinearFunction` and behaves exactly the same way except for `is_maybe_real`. At the JuMP level, we can have only one function type. In the conversion from the JuMP nonlinear expression to the MOI nl function, we need to decide if we can guarantee real-ness. There are three scenarios
* We know it will be real: so we use `MOI.ScalarNonlinearFunction`
* We know it may be complex, e.g., there are complex coefficients: so we use `MOI.ComplexScalarNonlinearFunction`
* There is some user-defined operator so we don't know whether it's real, what do we do here ? I think it's safer to return a `MOI.ComplexScalarNonlinearFunction`

In the third case, we should use a `MOI.ComplexScalarNonlinearFunction`.

We should have a bridge that converts `MOI.ComplexScalarNonlinearFunction`-in-`MOI.EqualTo` into `MOI.ScalarNonlinearFunction`-in-`MOI.EqualTo` by adding `real` and `imag` at the root of the expression tree (actually, maybe we could make it work with `SplitComplexEqualToBridge` ?).

For the user-defined function, we could add a way for the user to say that it's always real (and maybe it should be the default). Whenever the operator is evaluated, we should check that it's indeed real with a type assert probably then. For complex ones, we could have a `ComplexNonlinearOperator` or a trait `may_be_complex(::NonlinearOperator{F})` that returns `true` by default (that's probably simpler). If, when we evaluate a nonlinear operator that is said to be real-valued by the user, we see a complex value as output, then we error and tell the user he should implement `may_be_complex` / or use `ComplexNonlinearOperator`.

odow:
I guess the issue is that `x'` is a very common operation when building models, and people probably expect that it does `transpose` instead of `adjoint`.

Let's ignore `real` and `imag` for now, and focus on `conj`. Can we actually define this for arbitrary nonlinear expressions?

blegat:
We probably replied at the same time. I think in the case we can guarantee that the expression is real, we `adjoint` is a no-op. Otherwise, we should add it in the expression tree. It's find if the AD errors for now saying `adjoint` is not a supported operator, it's still an improvement over incorrect behavior. I also expect that we can assert that the expression is real most of the time, especially if we consider that `NonlinearOperator` is real by default.

odow:
The other option is `NonlinearExpr(:complex, Any[real_part, imag_part])`.

But it's also not obvious what to do for `sqrt(-x)` where `x >= 0`.

Part of the problem is that we don't have a solver, or an AD system that is capable of `Complex`. So perhaps it's best just to state that `ScalarNonlinearFunction <: Real` for now.

I'll check through MOI, but we shouldn't have any users relying on this yet. 

Here's an email I sent to @ccoffrin in August:
```
So this was sufficient

struct ComplexExpr{F<:AbstractJuMPScalar}
    real::F
    imag::F
end

Base.zero(::Type{ComplexExpr{F}}) where {F} = ComplexExpr(zero(F), zero(F))

Base.real(x::ComplexExpr) = x.real

Base.imag(x::ComplexExpr) = x.imag

function Base.:*(x::Complex, y::ComplexExpr)
    return ComplexExpr(
        real(x) * real(y) - imag(x) * imag(y),
        imag(x) * real(y) + real(x) * imag(y),
    )
end

function Base.:*(x::Complex, y::NonlinearExpr)
    return ComplexExpr(real(x) * y, imag(x) * y)
end

Base.:*(y::NonlinearExpr, x::Complex) = x * y

function Base.:+(x::ComplexExpr, y::ComplexExpr)
    return ComplexExpr(real(x) + real(y), imag(x) + imag(y))
end

function Base.:+(x::AbstractJuMPScalar, y::ComplexExpr)
    return ComplexExpr(x + real(y), imag(y))
end

to get this to work

using JuMP, LinearAlgebra
model = Model()
@variable(model, vm[1:2, 1:3])
@variable(model, va[1:2, 1:3])
@variable(model, p[1:3])
@variable(model, q[1:3])
VV_real = vm[1,:] .* vm[2,:]' .* cos.(va[1,:] .- va[2,:]')
VV_imag = vm[1,:] .* vm[2,:]' .* sin.(va[1,:] .- va[2,:]')
Y = [
    1 + 1im  2 + 2im 3 + 3im;
    4 + 4im  5 + 5im 6 + 6im;
    5 + 5im  6 + 6im 7 + 7im;
]
VV = VV_real .+ VV_imag .* im
@constraint(model, p .== real.(diag(Y * VV)))
@constraint(model, q .== imag.(diag(Y * VV)))

So I'm willing to declare this as "will work, given time to implement and test."
```

We don't have complex-valued nonlinear expressions, but we have complex values with nonlinear components.

We wouldn't support something like `cos(::ScalarAffineFunction{<:Complex})` because there's no application need.

blegat:
My only worry is that users might input complex expressions into nonlinear expression and it would lead here to silent bugs with this PR. Unless the AD throws an error later if it encounters complex values ?

odow:
Yes, currently AD errors on things it doesn't understand

odow:
Here's a extension-tests: https://github.com/jump-dev/JuMP.jl/actions/runs/8623134630

blegat:
Yes but other solvers like Convex.jl might not error. We may be passing incorrect (because of methods of this PR) models with complex expressions to a solver assuming it will error of there are complex expressions. I think it would be safer to error in the conversion from the JuMP expression to the MOI function as well

odow:
> I think it would be safer to error in the conversion from the JuMP expression to the MOI function as well

Do you mean erroring if the input is complex in the `ScalarNonlinearFunction` constructor of MOI?

You can't create complex nonlinear expressions without manually constructing them:
https://github.com/jump-dev/JuMP.jl/blob/05d48766a933c1e4d48fb5e23b28eb4d3dcbace1/src/nlp_expr.jl#L296-L334

blegat:
The check doesn't seem to be always called:
```julia
julia> model = Model();

julia> @variable(model, x in ComplexPlane())
real(x) + imag(x) im

julia> cos(x)
ERROR: Cannot build `GenericNonlinearExpr` because a term is complex-valued: `(real(x) + imag(x) im)::GenericAffExpr{ComplexF64, VariableRef}`
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] _throw_if_not_real(x::GenericAffExpr{ComplexF64, VariableRef})
   @ JuMP ~/.julia/dev/JuMP/src/nlp_expr.jl:307
 [3] cos(x::GenericAffExpr{ComplexF64, VariableRef})
   @ JuMP ~/.julia/dev/JuMP/src/nlp_expr.jl:325
 [4] top-level scope
   @ REPL[23]:1

julia> x^3
(real(x) + imag(x) im) ^ 3

julia> @constraint(model, x^3 == 1)
((real(x) + imag(x) im) ^ 3.0) - 1.0 = 0
```

odow:
Ah. I think you found the one method that has a bug:
https://github.com/jump-dev/JuMP.jl/blob/05d48766a933c1e4d48fb5e23b28eb4d3dcbace1/src/operators.jl#L210-L218

blegat:
You can also still do this:
```julia
julia> model = Model();

julia> @variable(model, x in ComplexPlane())
real(x) + imag(x) im

julia> GenericNonlinearExpr(:+, x)
+(real(x) + imag(x) im)
```

odow:
If people are manually constructing expressions then we don't check anything, correct.

blegat (approved):

odow:
Checking https://github.com/jump-dev/JuMP.jl/actions/runs/8637031676 again before merging

PR #3725, opened 2024-04-10
Fix throwing error in ^(::GenericAffExpr{<:Complex}, ::Integer)

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3724#issuecomment-2046772222


odow:
Closing in favor of #3724

Issue #3729, opened 2024-04-15
Poor performance in complementarity models


odow:
@mitchphillipson sent me a variation off this example, which runs much slower than I would expect:
```julia
using JuMP
import PATHSolver

function main()
    R = [Symbol("R_$i") for i in 1:51]
    S = G = [Symbol("S_$i") for i in 1:71]
    ys0 = Containers.DenseAxisArray(rand(length(R), length(S), length(G)), R, S, G)
    ys0_n = length(ys0)
    ys0.data[rand(1:ys0_n, floor(Int, 0.8 * ys0_n))] .= 0.0
    id0 = Containers.DenseAxisArray(rand(length(R), length(S), length(G)), R, S, G)
    ld0 = Containers.DenseAxisArray(rand(length(R), length(S)), R, S)
    ld0.data[rand(1:length(ld0), 20)] .= 0.0
    kd0 = Containers.DenseAxisArray(rand(length(R), length(S)), R, S)
    tk0 = Containers.DenseAxisArray(fill(0.32, length(R)), R)
    household = Model(PATHSolver.Optimizer)
    @variables(household, begin
        ty[r in R, s in S], (start = 0.5)
        tk[R, S]
        Y[R, S] >= 0,	(start = 1)
        PA[R, G] >= 0,	(start = 1)
        PY[R, G] >= 0,	(start = 1)
        RK[R, S] >= 0,	(start = 1)
        PL[R] >= 0,	    (start = 1)
    end)
    @expressions(household, begin
        theta_PL_Y_VA[r in R, s in S],
            ifelse(
                ld0[r,s] != 0,
                ld0[r,s] / (ld0[r,s] + kd0[r,s] * (1 + tk0[r])),
                0.0,
            )
        PI_Y_VA[r in R, s in S],
            PL[r]^theta_PL_Y_VA[r,s] * (RK[r,s] * (1 + tk[r,s]) / (1 + tk0[r]))^(1 - theta_PL_Y_VA[r,s])
        I_PL_Y[r in R, s in S],
            ld0[r,s] * PI_Y_VA[r,s] / PL[r]
        I_RK_Y[r in R, s in S],
            kd0[r,s] * (1 + tk0[r]) * PI_Y_VA[r,s] / (RK[r,s] * (1 + tk[r,s]))
    end)
    @constraint(
        household,
        [r in R, s in S; sum(ys0[r, s, g] for g in G) > 0],
[...]

PR #3732, opened 2024-04-16
Add @nonlinear macro for modifying how expressions are parsed

odow:
x-ref https://github.com/jump-dev/JuMP.jl/issues/3498
x-ref https://github.com/jump-dev/JuMP.jl/issues/3729

 - [x] I'm open to bikeshedding the name (and whether we should even add this).

A core problem is that @mitchphillipson's models from #3729 contain `(x * (1 + y)) / c`, which creates an AffExpr `1 + y`, a QuadExpr `x * (1 + y)`, and then a new QuadExpr `(x * (1 + y)) / 2` (because `/(::Quad, ::Real)` is not in-place.

Since everything is about to be converted to `NonlinearExpr` anyway, it'd be nicer if we just kept things as `NonlinearExpr`.

cc @Downsite


blegat:
> Since everything is about to be converted to `NonlinearExpr` anyway, it'd be nicer if we just kept things as `NonlinearExpr`.

In this case yes, but this performance issue could be hit in cases where a quadratic expression is the desired typed.
If the rewriting could would rewrite `(x * (1 + y)) / 2` into `MA.operate!!(/, x * (1 + y),  2)` then the performance issue would be fixed (provided the corresponding methods are defined in MA). I'd prefer fixing the performance issue, especially since it falls exactly within realms of issues MA was designed to fix.

odow:
> but this performance issue could be hit in cases where a quadratic expression is the desired typed.

Exactly. Which is why this `@nl` is opt-in. We want to default to building affine and quadratic expressions. But sometimes people might want nonlinear ones.

It's not really about this specific performance issue. Even if `operate!(/, Float64, lhs, rhs)` was implemented, there would still be more temporaries allocated because creating `AffExpr` and QuadExpr` require allocating `OrderedDict` which are expensive.

blegat:
I don't see any objection then, I can see this being useful. About the naming, once we drop the legacy nonlinear interface, what would be left of the `nl` shortcut ? If not much then maybe the `nl` shortcut is not so standard anymore in JuMP and we can use `@nonlinear`

odow:
Yeah I have no strong opinions on the name. I don't like `@nl(...)`. `@nonlinear` could work.

odow:
Bikeshed

 * `@_nl` ?
 * `@nl_esc`
 * `@scary_nl`
 * `@nl_boo` (@matbesancon)
 * `@nlexpr(...)`
 * `@nl_expr(...)`
 * `@raw_nl_expr(...)`
 * `@raw_nonlinear_expr(...)`
 * `@raw_nl`
 * ...

odow:
This could use a review. I've updated to `@force_nonlinear` as discussed.

pulsipher (approved):
Looks good to me.

matbesancon (approved):

docs/src/manual/nonlinear.md
```diff
@@ -355,6 +355,34 @@ julia> expr.args
...
+
+julia> @variable(model, x);
+
+julia> f = (x - 0.1)^2
+x² - 0.2 x + 0.010000000000000002
+
+julia> typeof(f)
+QuadExpr (alias for GenericQuadExpr{Float64, GenericVariableRef{Float64}})
+```
+To over-ride this behavior, use the [`@force_nonlinear`](@ref) macro:
```
matbesancon, line 373: nitpick: spelling override without hyphen?

src/macros/@force_nonlinear.jl
```diff
@@ -0,0 +1,118 @@
...
+                return Expr(:call, _op_div, x.args[2:end]...)
+            elseif x.args[1] == :^
+                return Expr(:call, _op_pow, x.args[2:end]...)
+            end
+        end
+        return x
+    end
+    return quote
+        r = $(esc(ret))
+        if !(r isa $GenericNonlinearExpr)
```
blegat, line 113: I'd rather call a `_check_nonlinear` than complicate the generated code ? Less worry about hygiene issue of `r`, etc...

odow: Done

blegat (approved):

MAiNGO-github:
@odow Is it expected that @force_nonlinear does not work when the quadratic expression comes from a user-defined function?

`@force_nonlinear(p.funcs.g(x, w[j, :])): expression did not produce a GenericNonlinearExpr. Got a QuadExpr: w[1,1]*w[1,2] - x[1]*w[1,2] + w[1,1] + 0 x[1]")`

Downsite:
@odow Is it expected that @force_nonlinear does not work when the quadratic expression comes from a user-defined function?

`@force_nonlinear(p.funcs.g(x, w[j, :])): expression did not produce a GenericNonlinearExpr. Got a QuadExpr: w[1,1]*w[1,2] - x[1]*w[1,2] + w[1,1] + 0 x[1]")`

odow:
Yes, because the macro cannot rewrite the expressions inside `p.funcs.g`. 

We considered automatically converting the result, but decided against it.

You should modify `p.funcs.g` to create the appropriate expression type.

PR #3734, opened 2024-04-17
Add methods for Base.complex

### LINKED ISSUE #3733: Add a method for Complex(a,b)
Could you please add a method `Complex(a,b) = a + im*b` for real JuMP variables `a` and `b`?

The reason is that some time ago I generalized the function `__svec_to_smat_complex!` from https://github.com/jump-dev/Hypatia.jl/blob/master/src/Cones/arrayutilities.jl , to work with JuMP variables as well by replacing `Complex(a,b)` with `a + im*b`. But as it turns out this slightly degrades performance for numerical types, so I'd like to undo my hack.

odow:
Closes #3733

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -531,3 +531,32 @@ function Base.:*(
 )
     return x * A
 end
+
+function Base.complex(
+    r::Union{
+        GenericVariableRef{<:Real},
+        GenericAffExpr{<:Real},
+        GenericQuadExpr{<:Real},
+    },
+    i::Real,
+)
+    return r + im * i
+end
+
+function Base.complex(
+    r::Real,
+    i::Union{
+        GenericVariableRef{<:Real},
+        GenericAffExpr{<:Real},
+        GenericQuadExpr{<:Real},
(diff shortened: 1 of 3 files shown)


araujoms:
I was going to do this now, you were too fast for me. Thanks a lot!

src/operators.jl
```diff
@@ -531,3 +531,32 @@ function Base.:*(
...
+        GenericQuadExpr{<:Real},
+    },
+)
+    return r + im * i
+end
+
+function Base.complex(
+    r::Union{GenericVariableRef{<:Real},GenericAffExpr{<:Real}},
+    i::Union{GenericVariableRef{<:Real},GenericAffExpr{<:Real}},
+)
```
araujoms, line 560: Why are you not allowing `GenericQuadExpr{<:Real}`here?

odow: Because I can't copy-paste correctly haha (and then I added tests for the methods that exist, not thinking)

blegat (approved):

PR #3743, opened 2024-05-03
Support vector-valued constraints in normalized_coefficient

### LINKED ISSUE #3742: Support normalized_coefficient for vector-valued constraints
See https://discourse.julialang.org/t/jump-constraint-coefficients/113754

odow:
Closes #3742 

One naming question:

We actually have `set_normalized_coefficiets` for this case:
https://jump.dev/JuMP.jl/stable/api/JuMP/#set_normalized_coefficients

So we should probably either call this `normalized_coefficients`, or rename `set_normalized_coefficients` to `set_normalized_coefficient` (but leave the current method for backwards compatibility).

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -2801,7 +2801,7 @@ end
 Return the coefficient associated with `variable` in `constraint` after JuMP has
 normalized the constraint into its standard form.
 
-See also [`set_normalized_coefficient`](@ref).
+See also [`set_normalized_coefficient`](@ref) and [`set_normalized_coefficients`](@ref)..
 
 ## Example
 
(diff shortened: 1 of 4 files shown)


src/variables.jl
```diff
@@ -2816,6 +2832,15 @@ con : 5 x ≤ 2
...
 5.0
+
+julia> @constraint(model, con_vec, [x, 2x + 1, 3] >= 0)
+con_vec : [x, 2 x + 1, 3] ∈ MathOptInterface.Nonnegatives(3)
+
+julia> normalized_coefficient(con_vec, x)
+3-element Vector{Float64}:
+ 1.0
+ 2.0
+ 0.0
```
odow, line 2843: Another option is whether the return for this should be `[(1, 1.0), (2, 2.0)]` to match the expectation of `set_normalized_coefficient`.

blegat: I would rather have a `SparseArrays.SparseVector`

odow: I don't really want a `SparseVector`. This would be perhaps the only place in all of JuMP that returns such an object. We should either return `Vector` or return the input needed by `set_normalized_coefficient`.

blegat: The same input as `set_normalized_coefficient` then I'd say

odow: Okay, how's this?

odow: It makes sense that `set_normalized_coefficient(c, x, normalized_coefficient(c, x))` preserves the current model.

blegat (approved):

joaquimg (approved):

Issue #3746, opened 2024-05-16
Precompilation in CI fails on v1.22


hexaeder:
Hello,
in the docs of some package I have JuMP as an indirect dependency. I experience [precompile errors](https://github.com/MakieOrg/GraphMakie.jl/actions/runs/9076105666/job/24938107012?pr=187#step:5:1207) in the GH Actions CI on JuMP v1.22, by forcing v1.21 it works. I was not able to reproduce the problem locally.

The stacktrace 
```julia
Stacktrace:
  [1] top-level scope
    @ ~/.julia/packages/JuMP/Y2NoH/src/macros/@NL.jl.3746.cov:6
```
implies, that this code

https://github.com/jump-dev/JuMP.jl/blob/71c7764b2861164f7e668120a8d9b700ab3e399a/src/macros.jl#L470-L472

tries to include some coverage file (don't now why it ends up therein my CI) and fails. Maybe it would already help to filter and only include `*.jl` files?

odow:
My suggestion would be to move the documentation build to a new GitHub action job. At the moment it runs after testing, which populates the .cov coverage files:
https://github.com/MakieOrg/GraphMakie.jl/blob/01c606eb7963fdad67ac91847f88a0e938ab3e91/.github/workflows/CI.yml#L44-L62

See JuMP's separate `documentation.yml` job: https://github.com/jump-dev/JuMP.jl/blob/master/.github/workflows/documentation.yml

odow:
I thought the loop and include was a common pattern, but it looks like I must have hit something similar, because elsewhere we also have the `endswith(file, ".jl")` pattern:
https://github.com/jump-dev/MathOptInterface.jl/blob/8742ceb9f2a6853672f9a5c999d75ea3b162ac03/src/Test/Test.jl#L629-L630

odow:
Here's another case

https://github.com/jump-dev/MultiObjectiveAlgorithms.jl/blob/26a9a6462d3e948e5c8ed4c89bdfffe59bc6ca10/src/MultiObjectiveAlgorithms.jl#L633-L635

odow:
I've audited `jump-dev`, and I think that's it: https://github.com/search?q=org%3Ajump-dev+readdir&type=code&p=1

odow:
Prepping a new release: https://github.com/jump-dev/JuMP.jl/pull/3749

[...]

PR #3747, opened 2024-05-16
only autoinclude files ending with `.jl` in `macros`

hexaeder:
possible fix for #3746


odow (approved):
Thanks. I made a slight change and added a comment.

PR #3751, opened 2024-05-20
Improve error message for unsupported kwargs in variable macro

### LINKED ISSUE #3750: Fuzzy search on unrecognized keyword arguments to give suggestions
Welcome. Use this GitHub issue to suggest a new feature for JuMP.

When using JuMP, there are many keyword arguements.
It is sometimes hard to remember exactly how they are spelt.
E.g. casing, pluralization, typos etc.

When a keyword argument is not recognized display a list of similar keyword arguments as per fuzzy string distence.

Implement this in Julia itself: https://github.com/JuliaLang/julia/issues/54521
however, JuMP isn't actually hitting the usual path for this, this is a manually thrown error, so implementing it in julia itself would not work.

**Additional context**

Here is an example:

```julia
julia> @variable(model, [1:3], basename=:foo)
ERROR: At REPL[4]:1: `@variable(model, [1:3], basename = :foo)`: Unrecognized keyword argument: basename.

If you're trying to create a JuMP extension, you need to implement `build_variable`. Read the docstring for more details.
Stacktrace:
  [1] error(::String, ::String)
    @ Base ./error.jl:44
  [2] (::JuMP.Containers.var"#error_fn#98"{String})(str::String)
    @ JuMP.Containers ~/.julia/packages/JuMP/Gwn88/src/Containers/macro.jl:325
  [3] build_variable(::JuMP.Containers.var"#error_fn#98"{…}, ::VariableInfo{…}; kwargs::@Kwargs{…})
    @ JuMP ~/.julia/packages/JuMP/Gwn88/src/macros/@variable.jl:677
  [4] macro expansion ...
```

I would like the output to include:
```
 Unrecognized keyword argument: basename.
 Did you mean: base_name ?
 ```

odow:
Closes #3750

@oxinabox is this enough?

### DIFF
--- a/src/macros/@variable.jl
+++ b/src/macros/@variable.jl
@@ -675,9 +675,25 @@ function build_variable(
             )
         end
         error_fn(
-            "Unrecognized keyword argument: $key.\n\nIf you're trying " *
-            "to create a JuMP extension, you need to implement " *
-            "`build_variable`. Read the docstring for more details.",
+            """
+            Unrecognized keyword argument: $key.
+
+            The supported keyword arguments are:
+
+             * `base_name`
+             * `binary`
+             * `container`
+             * `integer`
+             * `lower_bound`
+             * `set`
+             * `set_string_name`
+             * `start`
+             * `upper_bound`
+             * `variable_type`
+
+            If you're trying to create a JuMP extension, you need to implement `JuMP.build_variable`.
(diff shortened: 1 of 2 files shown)


blegat (approved):

oxinabox (approved):
Yes, this will fill the need.

PR #3758, opened 2024-05-27
[docs] improve various docstrings

odow:
Motivated by https://discourse.julialang.org/t/cpu-time-vs-wallclock/114776

https://jump.dev/JuMP.jl/previews/PR3758/api/JuMP/#solve_time


odow:
There are still a bunch that can be improved, but this is a good start.

odow:
The original motivation goes from
image
to
image

Issue #3762, opened 2024-05-30
Deleting part of containerized constraints makes whole constraint unprintable


LebedevRI:
At least, i would hope this quirk should be documented,
although since the model is still printable, it seems like a bug,
especially since it is recommended/ to print separate constraints for debugging.

```
using JuMP

function main()
    model = Model() 

    @variable(model, v[1:2,1:2,1:2])
    @constraint(model, c[i=1:2,j=1:2,k=1:2], v[i,j,k] >= 42)

    display(c)
    print(c)
    print(model)
    delete(model, c[1,1,1])

    # display(c)
    # ^ This fails with:
    # The index MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.GreaterThan{Float64}}(1) is invalid. Note that an index becomes invalid after it has been deleted.

    #print(c) # Same

    print(model) # Works!
end

main()
```

odow:
MWE:
```Julia
julia> using JuMP

julia> model = Model() ;

julia> @variable(model, x);

julia> @constraint(model, c[i=1:2], x <= i);

julia> delete(model, c[1])

julia> c
2-element Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.LessThan{Float64}}, ScalarShape}}:
Error showing value of type Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.LessThan{Float64}}, ScalarShape}}:
ERROR: The index MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.LessThan{Float64}}(1) is invalid. Note that an index becomes invalid after it has been deleted.
Stacktrace:
  [1] throw_if_not_valid
```

@LebedevRI I will commend you. You somehow keep finding the rough edges of JuMP!

LebedevRI:
@odow thank you!

> @LebedevRI I will commend you. You somehow keep finding the rough edges of JuMP!

Thank you, i guess? It's a bit of a curse :[ I would very much prefer to
never encounter bugs anywhere instead, but the exact opposite happens in practice...

PR #3763, opened 2024-05-31
Fix printing when show called on invalid variable or constraint

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3762

 - [x] Test extensions: https://github.com/jump-dev/JuMP.jl/actions/runs/9312266581

### DIFF
--- a/test/test_print.jl
+++ b/test/test_print.jl
@@ -1070,4 +1070,21 @@ function test_print_omit_vector()
     return
 end
 
+function test_invalid_references()
+    model = Model()
+    @variable(model, x[1:2])
+    @constraint(model, c[i=1:2], x[i] <= i)
+    delete(model, c[1])
+    delete(model, x[1])
+    mime = MIME("text/plain")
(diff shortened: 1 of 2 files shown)


test/test_print.jl
```diff
@@ -1070,4 +1070,21 @@ function test_print_omit_vector()
     return
 end
 
+function test_invalid_references()
+    model = Model()
+    @variable(model, x[1:2])
+    @constraint(model, c[i=1:2], x[i] <= i)
```
odow, line 1076:
```suggestion
    @constraint(model, c[i in 1:2], x[i] <= i)
```

blegat (approved):
We can ignore PolyJuMP and SumOfSquares failing, I'm doing a big refactoring

odow:
So you're okay with just printing "InvalidXXX"?

blegat:
Yes, the variable is invalid so using it in a model should throw but we could allow printing it and `InvalidXXX` seems reasonable.

Issue #3765, opened 2024-05-31
Inequality constraints over symmetric matrices are cumbersome


LebedevRI:
Unless i'm missing something very obvious, there is no nice way to exploit the symmetry of matrices
(to avoid adding duplicate constraints) when using `<=` and `>=` constraints.
Here's what i've tried:
```julia
using JuMP
import LinearAlgebra
import MathOptInterface as MOI

function main()
    model = Model() 

    @variable(model, v[1:3,1:3], Symmetric)

    # Results in nice triangular matrix in Zeros()
    @constraint(model, c0, LinearAlgebra.Symmetric(v.-1)  == LinearAlgebra.Symmetric(fill(41, 3,3))) # Yay!

    # Unsupported matrix in vector-valued set.
    # @constraint(model, c1, LinearAlgebra.Symmetric(v.-1)  >= LinearAlgebra.Symmetric(fill(41, 3,3))) # No.

    # Works, but results in 9 constraints, and all 9 elements are used, instead of always using upper-triangular elements.
    @constraint(model, c2, LinearAlgebra.Symmetric(v.-1) .>= LinearAlgebra.Symmetric(fill(41, 3,3))) # Yes, but no symmetry.
    
    # Unsupported matrix in vector-valued set.
    # @constraint(model, c3, LinearAlgebra.Symmetric(v.-42) >= LinearAlgebra.Symmetric(fill(0, 3,3))) # No.

    # Unsupported matrix in vector-valued set.
    #@constraint(model, c4, LinearAlgebra.Symmetric(v.-42) >= 0) # No

    # Works, but still no notion of symmetry.
    @constraint(model, c5, vec(LinearAlgebra.Symmetric(v.-42)) >= 0)

    # MethodError: no method matching triangular_form
    # @constraint(model, c6, vec(MOI.triangular_form(LinearAlgebra.Symmetric(v.-42))) in MOI.Nonnegatives(3*3)) # No

    # RHS is not lower-triangular, you get `-42 in Nonnegatives` elts. Oops.
    @constraint(model, c7, vec(LinearAlgebra.LowerTriangular(v.-1)) >= vec(fill(41, 3,3))) # Yes, kind of.

    # Finally, this works, no duplicate constraints, but the vector is still 9-element (padded with zeros)
[...]

PR #3766, opened 2024-06-03
Add set inequality syntax for matrices

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3765

Documentation preview: https://jump.dev/JuMP.jl/previews/PR3766/manual/constraints/#Symmetric-matrices

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -1474,6 +1474,44 @@ julia> @constraint(model, LinearAlgebra.Symmetric(X) >= 0, PSDCone())
 ```
 (Note that no error is thrown, even though `X` is not symmetric.)
 
+## Symmetric matrices
+
+Inequalities between symmetric matrices are not supported, due to the common
+ambiguity between elementwise inequalities and a [`PSDCone`](@ref) constraint.
+```jldoctest symmetric_matrix
+julia> model = Model();
+
+julia> @variable(model, x[1:2, 1:2], Symmetric);
+
+julia> @variable(model, y[1:2, 1:2], Symmetric);
+
+julia> @constraint(model, x >= y)
+ERROR: At REPL[91]:1: `@constraint(model, x >= y)`: Unsupported matrix in vector-valued set. Did you mean to use the broadcasting syntax `.>=` instead? Alternatively, perhaps you are missing a set argument like `@constraint(model, X >= 0, PSDCone())` or `@constraint(model, X >= 0, HermmitianPSDCone())`.
+Stacktrace:
+[...]
+```
+
+Instead, use the [Set inequality syntax](@ref) to specify either a [`PSDCone`](@ref),
+or a elementwise inequality set like [`Nonnegatives`](@ref):
+
+```jldoctest symmetric_matrix
+julia> @constraint(model, x >= y, PSDCone())
+[x[1,1] - y[1,1]  x[1,2] - y[1,2]
+ x[1,2] - y[1,2]  x[2,2] - y[2,2]] ∈ PSDCone()
+
+julia> @constraint(model, x >= y, Nonnegatives())
+[x[1,1] - y[1,1]  x[1,2] - y[1,2]
+ x[1,2] - y[1,2]  x[2,2] - y[2,2]] ∈ Nonnegatives()
+
+julia> @constraint(model, x >= y, Nonpositives())
+[x[1,1] - y[1,1]  x[1,2] - y[1,2]
+ x[1,2] - y[1,2]  x[2,2] - y[2,2]] ∈ Nonpositives()
+
+julia> @constraint(model, x >= y, Zeros())
+[x[1,1] - y[1,1]  x[1,2] - y[1,2]
+ x[1,2] - y[1,2]  x[2,2] - y[2,2]] ∈ Zeros()
+```
+
 ## Complementarity constraints
 
 A mixed complementarity constraint `F(x) ⟂ x` consists of finding `x` in the
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -188,19 +188,19 @@ Instead, use the [Set inequality syntax](@ref) to specify a set like
 ```jldoctest symmetric_matrix
 julia> @constraint(model, x >= y, PSDCone())
 [x[1,1] - y[1,1]  x[1,2] - y[1,2]
- x[1,2] - y[1,2]  x[2,2] - y[2,2]] ∈ PSDCone()
+ ⋯                x[2,2] - y[2,2]] ∈ PSDCone()
 
 julia> @constraint(model, x >= y, Nonnegatives())
 [x[1,1] - y[1,1]  x[1,2] - y[1,2]
- x[1,2] - y[1,2]  x[2,2] - y[2,2]] ∈ Nonnegatives()
+ ⋯                x[2,2] - y[2,2]] ∈ Nonnegatives()
 
 julia> @constraint(model, x >= y, Nonpositives())
 [x[1,1] - y[1,1]  x[1,2] - y[1,2]
- x[1,2] - y[1,2]  x[2,2] - y[2,2]] ∈ Nonpositives()
+ ⋯                x[2,2] - y[2,2]] ∈ Nonpositives()
 
 julia> @constraint(model, x >= y, Zeros())
 [x[1,1] - y[1,1]  x[1,2] - y[1,2]
- x[1,2] - y[1,2]  x[2,2] - y[2,2]] ∈ Zeros()
+ ⋯                x[2,2] - y[2,2]] ∈ Zeros()
 ```
 
 ### Special cases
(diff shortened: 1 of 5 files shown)


docs/src/manual/constraints.md
```diff
@@ -1474,6 +1474,44 @@ julia> @constraint(model, LinearAlgebra.Symmetric(X) >= 0, PSDCone())
...
+ambiguity between elementwise inequalities and a [`PSDCone`](@ref) constraint.
+```jldoctest symmetric_matrix
+julia> model = Model();
+
+julia> @variable(model, x[1:2, 1:2], Symmetric);
+
+julia> @variable(model, y[1:2, 1:2], Symmetric);
+
+julia> @constraint(model, x >= y)
+ERROR: At REPL[91]:1: `@constraint(model, x >= y)`: Unsupported matrix in vector-valued set. Did you mean to use the broadcasting syntax `.>=` instead? Alternatively, perhaps you are missing a set argument like `@constraint(model, X >= 0, PSDCone())` or `@constraint(model, X >= 0, HermmitianPSDCone())`.
```
odow, line 1489:
```suggestion
ERROR: At none:1: `@constraint(model, x >= y)`: Unsupported matrix in vector-valued set. Did you mean to use the broadcasting syntax `.>=` instead? Alternatively, perhaps you are missing a set argument like `@constraint(model, X >= 0, PSDCone())` or `@constraint(model, X >= 0, HermmitianPSDCone())`.
```

odow:
I can see us getting asked to support this for _all_ matrices, not just symmetric ones. But perhaps that' a PR for another day.

LebedevRI:
> I can see us getting asked to support this for _all_ matrices, not just symmetric ones. But perhaps that' a PR for another day.

FWIW, yes. The `vec()` escape hatch is meh, and may make issues w/ matrix shape mismatch (== comparing wrong things) less easy to discover.

docs/src/manual/constraints.md
```diff
@@ -164,27 +164,51 @@ julia> @constraint(model, A * x .>= b)
...
 
-julia> @variable(model, X[1:2, 1:2])
-2×2 Matrix{VariableRef}:
- X[1,1]  X[1,2]
- X[2,1]  X[2,2]
+julia> @variable(model, x[1:2, 1:2], Symmetric);
+
+julia> @variable(model, y[1:2, 1:2], Symmetric);
 
+<<<<<<< od/symmetric-inequality
```
LebedevRI, line 179: rebase marker

blegat:
What we are doing for `@constraint(model, x <= y, set)` is a bit weird, it will give a `value` and `dual` that are inconsistent with the scalar `@constraint(model, x <= y)` (similar to what we did for Convex.jl). I'm wondering if, instead of interpreting it as `y - x in set`, we shouldn't rather see it as `x - y in -set`. We would still encode it as `y - x in set` but we could keep the `-1` flip in the shape so that it's recovered when you query `value` or `dual`.

blegat:
As this is breaking then maybe we should continue to do what we do but we should add a warning in the docs that `@constraint(model, x >= y, set)` is preferred since `@constraint(model, x <= y, set)` may have a behavior that's unexpected or confusing which makes the things less readable or error-prone if `value` or `dual` are used.

LebedevRI:
> What we are doing for `@constraint(model, x <= y, set)` is a bit weird, it will give a `value` and `dual` that are inconsistent with the scalar `@constraint(model, x <= y)` (similar to what we noticed and fixed for Convex.jl).

> I'm wondering if, instead of interpreting it as `y - x in set`,

Peanut gallery comment: the docs and disscussion in the issue make it seem
like `x <= y` is treated exactly as `x >= y`, i.e. it would have still resulted in `x-y in set`?

> we shouldn't rather see it as `x - y in -set`. We would still encode it as `y - x in set` but we could keep the `-1` flip in the shape so that it's recovered when you query `value` or `dual`.

This kinda sounds like either solution would address my concerns from https://github.com/jump-dev/JuMP.jl/issues/3765, i think.

> As this is breaking then maybe we should continue to do what we do but we should add a warning in the docs that `@constraint(model, x >= y, set)` is preferred since `@constraint(model, x <= y, set)` may have a behavior that's unexpected or confusing which makes the things less readable or error-prone if `value` or `dual` are used.

Peanut gallery comment: the concern here is external non JuMP/MOI sets/code, correct? Since i don't think this syntax currently works on native JuMP/MOI stuff, so it would only affect extensions?

odow:
Let's just throw an error for these `<=` cases. I'll document the `PSDCone` case in a separate PR.

blegat:
But it already does it for AbstractVector so wouldn't adding an error be breaking ?

LebedevRI:
Peanut gallery comment:
it's going to be unfortunate that there's difference in behavior in these inequality constraints.
From outside, what would //i// think the best, no-unexpected-behavior, be, is:
```
@constraint(model, x >= y)
```
is always the same as
```
@constraint(model, x - y >= 0)
@constraint(model, x - y in Nonnegatives)
```
likewise

```
@constraint(model, x <= y)
```
is always the same as
```
@constraint(model, x - y <= 0)
@constraint(model, x - y in Nonpositives)
```

... for any [common] type of `x` and `y`, be it either a scalar, a vector, a symmetric matrix, or a non-symmetric matrix.

Currently
```
@constraint(model, x >= y)
@constraint(model, x <= y)
```
are not valid, and the code is always expected to explicitly spell out the cone type.
They already do that, so that change would greatly improve the QoL for the "normal" comparisons,
make bugs basically impossible in them (have i mentioned that manually spelling the set is a bit mouthful?),
while not particularly affecting the cone constraints - one can't automatically guess
which constraint was meant there, wherein that is trivial and obvious in "normal" case...

As i see it, it's an obvious trade-off, and the current solution is not obviously optimal.

I'll shut up now.

odow:
> From outside, what would //i// think the best, no-unexpected-behavior

@blegat agrees, and that's what we have been discussing. It's just a breaking change to make that work, so we won't be doing it in JuMP 1.x.

odow:
I've added an error for these new cases. No change to the existing vector support. I've added a warning to https://github.com/jump-dev/JuMP.jl/pull/3769, and we don't mention `x <= y, Set()` in the documentation.

odow:
I actually think this is less problematic that it seems. You're opting into a new syntax.

In Boyd's book, they have:

$x \preceq_K y \iff y - x \in K$

And they write

> We also write $x \succeq_K y$ for $y \preceq_K x$

It's not obvious that flipping $\preceq$ to $\succeq$ should change the sign of the primal or dual value. That's just not the way you interpret it.

LebedevRI:
FWIW i believe this (& its non-symmetric friend) can proceed without waiting for the understanding in #3770 to be reached.

test/test_constraint.jl
```diff
@@ -1887,4 +1887,58 @@ function test_set_normalized_coefficient_quadratic_batch()
...
 
+function test_symmetric_matrix_inequality()
+    model = Model()
+    @variable(model, x[1:2, 1:2], Symmetric)
+    @variable(model, y[1:2, 1:2], Symmetric)
+    set_start_value.(x, [1 2; 2 3])
+    set_start_value.(y, [6 4; 4 7])
+    g = [x[1, 1] - y[1, 1], x[1, 2] - y[1, 2], x[2, 2] - y[2, 2]]
+    for set in (Nonnegatives(), Nonpositives(), Zeros())
+        c = @constraint(model, x >= y, set)
```
odow, line 1898: Add a test for `x - y in set`

odow: Ah. @mlubin, I remember now. 

We can't support `x - y in Nonnegatives()` because then we can't tell it apart from `@constraint(model, x >= y)` which was the whole point in the first place.

blegat: > We can't support `x - y in Nonnegatives()`

What do you mean ? You mean `x >= y, Nonnegatives()` ?

odow: No I mean `@constraint(model, x - y in Nonnegatives())` because this hits the same codepath as `@constraint(model, x >= y)`.

odow: Unless I intercept the `GreaterThan(false)` method, but then I think the last time I looked that makes ambiguities.

odow:
Let's merge this as a strict improvement for now. I can come back to the `x - y in set` case in a separate PR.

test/test_constraint.jl
blegat: Ah, that's the part I didn't get. It does make sense to intersect the `GreaterThan` then because having `AbstractArray`-in-`GreaterThan` is ambiguous and shouldn't be allowed while `AbstractArray`-in-`Nonnegatives` should be allowed.
So neither `@constraint(model, x - y in GreaterThan(false))` nor `@constraint(model, x >= y)` will work.

odow: Ah. I forgot. We switched from `GreaterThan(false)` _to_ `Nonnegatives`. So we can't even intercept. `x >= y` is exactly equal to `x - y in Nonnegatives`

blegat: Ouch, that was probably not ideal. Using `Nonnegatives` by default kind of means we interpret `>=` as elementwise by default.

blegat: Maybe we could add a layer of indirection to fix that without being breaking

odow:
Merging this for now, but I'll hold off tagging a new release until I've had a go at `x - y in Nonnegatives`.

PR #3768, opened 2024-06-04
Improve printing of symmetric matrices when used in constraints

### LINKED ISSUE #3764: function_string of Symmetric matrix includes the entire matrix
```julia
using JuMP
import LinearAlgebra
import MathOptInterface as MOI

function main()
    model = Model() 

    @variable(model, v[1:3,1:3], Symmetric)

    @constraint(model, c0, LinearAlgebra.Symmetric(v.-1)  == LinearAlgebra.Symmetric(fill(41, 3,3))) # Yay!
    @show(c0) # Uhh, why is it not symmetric?
    print(c0) # Uhh, why is it not symmetric?
    display(c0) # Oh, but it is symmetric here.
    print(model) # And it is symmetric here too.
end

main()
```
```
c0 = c0 : [v[1,1] - 42  v[1,2] - 42  v[1,3] - 42;
 v[1,2] - 42  v[2,2] - 42  v[2,3] - 42; ∈ Zeros()
 v[1,3] - 42  v[2,3] - 42  v[3,3] - 42]
```
image
This is even worse for text-only output, looks like only pretty-printing does it right.

odow:
Closes #3764

So there is an argument for this, especially if we merge https://github.com/jump-dev/JuMP.jl/pull/3766, because it will help distinguish when JuMP exploits symmetry.

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -908,9 +908,9 @@ function function_string(
                 line *= " & "
             end
             if A isa LinearAlgebra.Symmetric && i > j
-                line *= "\\cdot"
+                line *= "\\cdots"
             elseif A isa LinearAlgebra.UpperTriangular && i > j
-                line *= "\\cdot"
+                line *= "\\cdots"
             else
                 line *= function_string(mode, A[i, j])
             end
@@ -935,7 +935,9 @@ function function_string(
     constraint::VectorConstraint{F,S,SymmetricMatrixShape},
 ) where {F,S}
     f = reshape_vector(jump_function(constraint), shape(constraint))
-    return function_string(mode, LinearAlgebra.UpperTriangular(f))
+    str = function_string(mode, LinearAlgebra.UpperTriangular(f))
+    return replace(str, "⋅" => "⋯")
+
 end
 
 function function_string(mode::MIME, p::NonlinearExpression)
--- a/src/print.jl
+++ b/src/print.jl
@@ -909,6 +909,8 @@ function function_string(
             end
             if A isa LinearAlgebra.Symmetric && i > j
                 line *= "\\cdot"
+            elseif A isa LinearAlgebra.UpperTriangular && i > j
+                line *= "\\cdot"
             else
                 line *= function_string(mode, A[i, j])
             end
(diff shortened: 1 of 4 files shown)


src/print.jl
```diff
@@ -909,6 +909,8 @@ function function_string(
             end
             if A isa LinearAlgebra.Symmetric && i > j
                 line *= "\\cdot"
```
odow, line 911: We could consider reverting the `Symmetric` case.

odow:
What about some other options:
```julia
julia> c = @constraint(model, x == 0)
[x[1,1]  x[1,2]  x[1,3]  x[1,4]
 ↙       x[2,2]  x[2,3]  x[2,4]
 ↙       ↙       x[3,3]  x[3,4]
 ↙       ↙       ↙       x[4,4]] ∈ Zeros()

julia> c = @constraint(model, x == 0)
[x[1,1]  x[1,2]  x[1,3]  x[1,4]
 ∘       x[2,2]  x[2,3]  x[2,4]
 ∘       ∘       x[3,3]  x[3,4]
 ∘       ∘       ∘       x[4,4]] ∈ Zeros()

julia> c
[x[1,1]  x[1,2]  x[1,3]  x[1,4]
 ⋰       x[2,2]  x[2,3]  x[2,4]
 ⋰       ⋰       x[3,3]  x[3,4]
 ⋰       ⋰       ⋰       x[4,4]] ∈ Zeros()

julia> c
[x[1,1]  x[1,2]  x[1,3]  x[1,4]
 ⤯       x[2,2]  x[2,3]  x[2,4]
 ⤯       ⤯       x[3,3]  x[3,4]
 ⤯       ⤯       ⤯       x[4,4]] ∈ Zeros()

```

src/print.jl
```diff
@@ -923,6 +925,21 @@ function function_string(mode, constraint::AbstractConstraint)
...
+# LinearAlgebra.UpperTriangular. There shouldn't be any cases in which the
+# constraint function becomes ambiguous.
+function function_string(
+    mode,
+    constraint::VectorConstraint{F,S,SymmetricMatrixShape},
+) where {F,S}
+    f = reshape_vector(jump_function(constraint), shape(constraint))
+    str = function_string(mode, LinearAlgebra.UpperTriangular(f))
+    return replace(str, "⋅" => "⋯")
+
```
odow, line 940:
```suggestion
```

odow:
Use of `\cdot` was added in https://github.com/jump-dev/JuMP.jl/pull/1857 without discussion. I think we can use `⋯` (`\cdots`) for now.

It hints at "something else is here" and avoids confusion with `⋅` and `0.0`.

odow:
Merging because I think this is a win, and it is needed for https://github.com/jump-dev/JuMP.jl/pull/3766

PR #3769, opened 2024-06-05
[docs] add warning about Y <= X, Set() syntax

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3766

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -1271,7 +1271,7 @@ julia> @constraint(model, x .- 1 >= 0, MOI.Nonnegatives(2))
 !!! warning
     The syntax `@constraint(model, y <= x, Set())` is supported, but it is not
     recommended because the value of the primal and dual solutions associated
-    with the connstraint may be the negative of what you expect.
+    with the constraint may be the negative of what you expect.
 
 ## Second-order cone constraints
 
@@ -1446,7 +1446,7 @@ julia> @constraint(model, X >= Y, PSDCone())
 !!! warning
     The syntax `@constraint(model, Y <= X, PSDCone())` is supported, but it is
     not recommended because the value of the primal and dual solutions associated
-    with the connstraint may be the negative of what you expect.
+    with the constraint may be the negative of what you expect.
 
 ### Symmetry
 
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -1238,8 +1238,9 @@ julia> @constraint(model, x in MOI.ExponentialCone())
 ## Set inequality syntax
 
 For modeling convenience, the syntax `@constraint(model, x >= y, Set())` is
-short-hand for `@constraint(model, x - y in Set())`. Therefore, the following
-calls are equivalent:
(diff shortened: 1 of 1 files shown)


docs/src/manual/constraints.md
```diff
@@ -1270,6 +1268,11 @@ julia> @constraint(model, x .- 1 >= 0, MOI.Nonnegatives(2))
 [x[1] - 1, x[2] - 1] ∈ MathOptInterface.Nonnegatives(2)
 ```
 
+!!! warning
+    The syntax `@constraint(model, y <= x, Set())` is supported, but it is not
+    recommended because the value of the primal and dual solutions associated
+    with the connstraint may be the negative of what you expect.
```
odow, line 1274:
```suggestion
    with the constraint may be the negative of what you expect.
```

docs/src/manual/constraints.md
```diff
@@ -1440,6 +1443,11 @@ julia> @constraint(model, X >= Y, PSDCone())
  X[2,1] - 2  X[2,2] - 1] ∈ PSDCone()
 ```
 
+!!! warning
+    The syntax `@constraint(model, Y <= X, PSDCone())` is supported, but it is
+    not recommended because the value of the primal and dual solutions associated
+    with the connstraint may be the negative of what you expect.
```
odow, line 1449:
```suggestion
    with the constraint may be the negative of what you expect.
```

docs/src/manual/constraints.md
```diff
@@ -1238,8 +1238,9 @@ julia> @constraint(model, x in MOI.ExponentialCone())
 ## Set inequality syntax
 
 For modeling convenience, the syntax `@constraint(model, x >= y, Set())` is
```
LebedevRI, line 1240: Note that as per https://docs.julialang.org/en/v1/manual/unicode-input/
`⪰` can be "trivially" inputted as `\succeq`.

Wouldn't it be possible to:
```
@constraint(model, x \succeq y)
# error: specify one of the conic sets
```
```
@constraint(model, x >= y, <... Some conic set ...>)
# warning: use \succeq instead
```
```
@constraint(model, x \succeq y, <... Some non-conic set ...>)
# warning: use >= instead
```
```
@constraint(model, x >= y)
# becomes @constraint(model, x - y, Nonnegatives)
```
```
@constraint(model, x <= y)
# becomes @constraint(model, x - y, Nonpositives)
```
Is the concern that someone writes `@constraint(model, x >= y)` thinking it means `PSDCone`,
it silently works, and doesn't do what's expected? Or is there some other concern?

odow: > Is the concern that someone writes @constraint(model, x >= y) thinking it means PSDCone,
it silently works, and doesn't do what's expected? 

Yes.

We won't be adding support for `\succeq`. There is too much risk of confusion.

LebedevRI: :/ I really feel such a decision should be made with a larger headcount than i see here.

odow: I'll open a new issue

PR #3778, opened 2024-06-25
Improve matrix inequality support

odow:
- [x] Requires #3797 to be merged first

Follow-up to #3766.

In our discussion of https://github.com/jump-dev/JuMP.jl/pull/3766, we never really talked about the fact that `==` is not ambiguous.

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -116,7 +116,7 @@ julia> b = [5, 6]
  6
 
 julia> @constraint(model, con_vector, A * x == b)
-con_vector : [x[1] + 2 x[2] - 5, 3 x[1] + 4 x[2] - 6] ∈ MathOptInterface.Zeros(2)
+con_vector : [x[1] + 2 x[2] - 5, 3 x[1] + 4 x[2] - 6] ∈ Zeros()
 
 julia> @constraint(model, con_scalar, A * x .== b)
 2-element Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.EqualTo{Float64}}, ScalarShape}}:
@@ -148,7 +148,7 @@ constraint.
 
 ```jldoctest con_vector
 julia> @constraint(model, A * x <= b)
-[x[1] + 2 x[2] - 5, 3 x[1] + 4 x[2] - 6] ∈ MathOptInterface.Nonpositives(2)
+[x[1] + 2 x[2] - 5, 3 x[1] + 4 x[2] - 6] ∈ Nonpositives()
 
 julia> @constraint(model, A * x .<= b)
 2-element Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.LessThan{Float64}}, ScalarShape}}:
@@ -156,7 +156,7 @@ julia> @constraint(model, A * x .<= b)
  3 x[1] + 4 x[2] ≤ 6
 
 julia> @constraint(model, A * x >= b)
-[x[1] + 2 x[2] - 5, 3 x[1] + 4 x[2] - 6] ∈ MathOptInterface.Nonnegatives(2)
+[x[1] + 2 x[2] - 5, 3 x[1] + 4 x[2] - 6] ∈ Nonnegatives()
 
 julia> @constraint(model, A * x .>= b)
 2-element Vector{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.GreaterThan{Float64}}, ScalarShape}}:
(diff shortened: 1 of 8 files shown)


test/test_macros.jl
```diff
@@ -1977,15 +1977,17 @@ function test_matrix_in_vector_set()
         ),
         @constraint(model, X <= A),
     )
+    Y = Containers.DenseAxisArray(X, 2:3, 2:3)
+    B = Containers.DenseAxisArray(A, 2:3, 2:3)
     @test_throws_runtime(
         ErrorException(
-            "In `@constraint(model, X == A)`: " *
+            "In `@constraint(model, X == B)`: " *
```
odow, line 1984:
```suggestion
            "In `@constraint(model, Y == B)`: " *
```

docs/src/manual/constraints.md
```diff
@@ -169,18 +169,37 @@ julia> @constraint(model, A * x .>= b)
 Inequalities between matrices are not supported, due to the common ambiguity
 between elementwise inequalities and a [`PSDCone`](@ref) constraint.
 
-```jldoctest symmetric_matrix
+````jldoctest symmetric_matrix
```
blegat, line 172:
```suggestion
```jldoctest symmetric_matrix
```

docs/src/manual/constraints.md
```diff
@@ -169,18 +169,37 @@ julia> @constraint(model, A * x .>= b)
...
+
+```julia
+@constraint(model, x >= y, Nonnegatives())
+# or
+@constraint(model, x .>= y)
+```
 Stacktrace:
 [...]
-```
+````
```
blegat, line 202:
```suggestion
```
```

**Comment by blegat** [comment]
Can we just replace `Nonnegatives` here:
https://github.com/jump-dev/JuMP.jl/blob/301d46e81cb66c74c6e22cd89fb89ced740f157b/src/macros/%40constraint.jl#L583
What are the other cases that you'd like to intercept ?

File: docs/src/manual/constraints.md (hunk 2 above)
Comment (odow) [reply docs/src/manual/constraints.md:172]: Nah. This really needs four ticks to disambiguate the inner triple.

**Comment by odow** [comment]
I didn't do this, because it could be interpreted as a breaking change.

**Comment by odow** [comment]
Bump @blegat

**Comment by blegat** [comment]
> I didn't do this, because it could be interpreted as a breaking change.

Breaking change for which function ? We are changing it to `_OpGreaterThan` before giving it to `build_constraint` in this PR as well anyway.

**Comment by odow** [comment]
Changes the return of `operator_to_set`, but perhaps this is minor: https://juliahub.com/ui/Search?type=code&q=operator_to_set

**Comment by blegat** [comment]
Indeed, the function could be used outside of a JuMP macro. I'd say it is minor. `_intercept_operator` is complicated or code too much just to accomplish to accomplish exactly the same thing as changing the returned value of `operator_to_set`

**Comment by odow** [comment]
I made a slight tweak to get rid of `_intercept_operator` but add a new method to `operator_to_set`.

**Comment by blegat** [comment]
I still think this complicating things too much just for such a small breaking change. Users are expected to implement `operator_to_set`, we never said that calling it would always return the same thing. I can't believe we haven't made something that could qualify as breaking for the returned value of such function since JuMP v1.0

**Comment by odow** [comment]
Okay how about now

File: src/macros/@constraint.jl [hunk 4]
```diff
@@ -792,6 +816,86 @@ function parse_constraint_call(
...
+        # or
+        @constraint(model, x .>= y)
+        ```""",
+    )
+end
+
+function build_constraint(
+    error_fn::Function,
+    f,
+    ::LessThanMacroSet
```
odow, line 862:
```suggestion
    ::LessThanMacroSet,
```

src/macros/@constraint.jl
```diff
@@ -792,6 +816,86 @@ function parse_constraint_call(
...
+    args...;
+    kwargs...,
+)
+    return build_constraint(error_fn, f, Nonpositives(), args...; kwargs...)
+end
+
+function build_constraint(
+    error_fn::Function,
+    ::Union{Matrix,LinearAlgebra.Symmetric,LinearAlgebra.Hermitian},
+    ::LessThanMacroSet
```
odow, line 872:
```suggestion
    ::LessThanMacroSet,
```

blegat:
I'd prefer to name things for what they are rather than what they are used for.
What about calling the set `Nonnegative` and saying that
`@constraint(model, a - b in Nonnegative())` and `@constraint(model, a >= b)` are synonymous.

odow:
Let's please not have `Nonnegative` and `Nonnegatives` for the same but slightly different thing.

blegat:
How about `GreaterThanZero` ?

odow:
This needs something similar to https://github.com/jump-dev/JuMP.jl/pull/3797

blegat:
> This needs something similar to #3797

What do you mean ?

odow:
It needed the adjoint dual for the new constraints, which is why this PR is
based against #3797

odow:
Okay, this is good to go on my end.

blegat (approved):

PR #3782, opened 2024-07-14
Add Convex.jl to solver list

blegat:
The MOI wrapper was added in Convex.jl v0.16


mlubin:
This needs more caveats, like those mentioned at https://github.com/jump-dev/Convex.jl?tab=readme-ov-file#using-with-jump.

mlubin:
I'm also having trouble thinking of which important use cases of DCP are covered without vector input functions like `norm`. Is it really worth advertising?

odow:
Agree. I'm still treating this as somewhat experimental.

blegat:
I agree there is maybe not point advertising this yet, let's close this

PR #3794, opened 2024-07-30
[docs] add warning to docs about tolerances in Bin and Int variables

### LINKED ISSUE #3793: `value` returns both `0.0` and `-0.0` on `Bin`ary values
The output here should be a BitMatrix, or, if you care about type stability, which is quite reasonable, a `Matrix{Float64}` with entries `0.0` and `1.0`. Some stray `-0.0`s aren't "wrong" in the sense that `-0.0 == 0.0`, but unless they convey some meaningful information that sign bit should be normalized to `0.0`:

```julia-repl
julia> using JuMP

julia> import HiGHS

julia> function bin_pack(capacities::AbstractVector{<:Real}, sizes::AbstractVector{<:Real})

           @variable(model, x[eachindex(sizes), eachindex(capac
           @objective(model, Max, sum(x))
           @constraint(model, sum(x, dims=2) .<= 1)
           @constraint(model, sizes' * x .<= capacities')

           set_silent(model)
           optimize!(model)

           value.(x)
       end
bin_pack (generic function with 1 method)

julia> bin_pack([4,5], [3,2,2,2])
4×2 Matrix{Float64}:
  0.0   1.0
  1.0  -0.0
  1.0  -0.0
 -0.0   1.0
```

odow:
Closes #3793

### DIFF
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -562,6 +562,14 @@ julia> @variable(model, x, Bin)
 x
 ```
 
+!!! warning
+    Solvers use tolerances to decide whether a variable satisfies the binary
+    constraint. Thus, the true feasible region is
+    $[-\varepsilon, \varepsilon] \cup [1 - \varepsilon, 1 + \varepsilon]$,
+    where $\varepsilon$ is solver-specific, but typically `1e-6`. As a result,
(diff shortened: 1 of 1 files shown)


odow:
image

image

PR #3797, opened 2024-08-02
Fix querying dual of symmetric equality constraints

### LINKED ISSUE #3796: Incorrect dual variable extracted from equality constraint between Hermitian matrices
When I use a constraint of the form `@constraint(model, A == B)`, where `A` and `B` are Hermitian matrices, for some reason the dual variable associated to this constraint has the off-diagonal elements multiplied by 2. I've tested a couple of different solvers, so the bug must be inside JuMP itself. A MWE follows below. Note the text-to-last line, where I'm dividing the off-diagonal elements of the dual variable by 2 to make it match the primal result.

```julia

using JuMP
using LinearAlgebra
import Hypatia
import Clarabel
import Ket
import Random

function dual_test(d)
    Random.seed!(1)
    ρ = Ket.random_state(d^2, 1)

    model = Model()
    @variable(model, σ[1:d^2, 1:d^2] in HermitianPSDCone())
    @variable(model, λ)
    noisy_state = Hermitian(ρ + λ * I(d^2))
    @constraint(model, witness_constraint, σ == noisy_state)
    @objective(model, Min, λ)
    @constraint(model, Ket.partial_transpose(σ, 2, [d, d]) in HermitianPSDCone())

    #set_optimizer(model, Clarabel.Optimizer)
    optimize!(model)

    W = dual(witness_constraint)
    W = Hermitian(Diagonal(W) + 0.5(W - Diagonal(W))) #workaround
    return objective_value(model), dot(ρ, W)
end
```

odow:
Closes #3796

I need to think about whether there is a better option. I don't think we should just fix the default `reshape_vector` because we don't need to modify anything for the primal? Just the dual?

### DIFF
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -141,15 +141,50 @@ MathOptInterface.PositiveSemidefiniteConeTriangle(2)
 struct PSDCone end
 
 """
-    SymmetricMatrixShape
+    SymmetricMatrixShape(
+        side_dimension::Int;
+        needs_adjoint_dual::Bool = false,
+    )
 
-Shape object for a symmetric square matrix of `side_dimension` rows and columns.
+The shape object for a symmetric square matrix of `side_dimension` rows and
+columns.
+
 The vectorized form contains the entries of the upper-right triangular part of
 the matrix given column by column (or equivalently, the entries of the
 lower-left triangular part given row by row).
+
+## `needs_adjoint_dual`
+
+By default, the [`dual_shape`](@ref) of [`SymmetricMatrixShape`](@ref) is also
+[`SymmetricMatrixShape`](@ref). This is true for cases such as a
+`LinearAlgebra.Symmetric` matrix in [`PSDCone`](@ref).
+
+However, JuMP also supports `LinearAlgebra.Symmetric` matrix in `Zeros`, which
+is interpreted as an element-wise equality constraint. By exploiting symmetry,
+we pass only the upper triangle of the equality constraints. This works for the
+primal, but it leads to a factor of 2 difference in the off-diagonal dual
+elements. (The dual value of the `(i, j)` element in the triangle formulation
+should be divided by 2 when spread across the `(i, j)` and `(j, i)` elements in
+the square matrix formulation.) If the constraint has this dual inconsistency,
+set `needs_adjoint_dual = true`.
 """
 struct SymmetricMatrixShape <: AbstractShape
     side_dimension::Int
+    needs_adjoint_dual::Bool
+
+    function SymmetricMatrixShape(
+        side_dimension::Int;
+        needs_adjoint_dual::Bool = false,
+    )
+        return new(side_dimension, needs_adjoint_dual)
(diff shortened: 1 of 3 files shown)


odow:
Okay, this looks like the problem. We never had any tests for this.

blegat:
You can use `JuMP.dual_shape`.
This is similar to what we do in
https://github.com/jump-dev/MathOptInterface.jl/blob/a15b67fe47ad20caf316eb1bba0369a46ceb5a34/src/Bridges/Constraint/bridges/square.jl#L401-L411
The rule in dual is to multiply by the adjoint of the linear transformation we did.
In case the matrix was symmetric, we assume that the transformation was taking the average of the two off-diagonal entries hence the **transpose** would set half the dual to each entry. But, because the triangular dot product multiplies by 2, the **adjoint** does not divide by 2 and just set the same value to both off-diagonal.
In here, because `Zeros` has the standard dot product, the adjoint and transpose are the same so assuming that the transformation takes the average, this PR is the adjoint. Could you add a small explanation in the comments explaining that we are simply doing the adjoint (assuming we used the average).

odow:
> You can use JuMP.dual_shape

I don't think so, or at least, not for `SymmetricMatrixShape` and `HermitianMatrixShape`, because these are a special case of us interpreting `Matrix -in- Zeros`. If we implemented `dual_shape` it would also modify things like Matrix in PSD.

Or we need a new matrix shape just for these Matrix-in-Zeros constraints.

blegat:
> Or we need a new matrix shape just for these Matrix-in-Zeros constraints.

Yes, maybe we need to do that.

odow:
Name suggestions for the primal and dual shapes?

odow:
Okay, instead of adding more shapes, I've added `needs_adjoint_dual` kwarg.

blegat (approved):
I agree a keyword argument is simpler

araujoms:
I'm afraid this PR introduced a bug: before if I tried to constrain `A == B` for Symmetric `A` and Hermitian `B` it would give me an error. Now it silently fails.

```julia
using JuMP
using LinearAlgebra
import Hypatia
import Clarabel
import Ket
import Random

function dual_test(d)
    Random.seed!(1)
    ρ = Ket.random_state(Float64, d^2, 1)

    model = Model()
    @variable(model, σ[1:d^2, 1:d^2] in PSDCone())
    @variable(model, λ)
    noisy_state = Hermitian(ρ + λ * I(d^2))
    @constraint(model, witness_constraint, σ == noisy_state)
    @objective(model, Min, λ)
    @constraint(model, Ket.partial_transpose(σ, 2, [d, d]) in PSDCone())

    #set_optimizer(model, Clarabel.Optimizer)
    optimize!(model)

    W = dual(witness_constraint)
    return W
end
```

blegat:
Oops, can you open an issue so that we can track this ?

odow:
> Now it silently fails

No, now it silently succeeds :smile:

PR #3798, opened 2024-08-04
Update power_systems.jl - `wind.` completed to `wind_generator.`

justbyoo:
Example code had incomplete variable name, `wind.`. Corrected to complete variable name, `wind_generator.`.


odow:
I don't think this is correct. Note that `wind` is an argument to the function?

odow:
On line 130, we have:
```
solution = solve_economic_dispatch(generators, wind_generator, scenario);
```
which calls this function with `wind_generator`.

odow:
Closing this as won't-fix since the current implementation is correct, but please feel free to open any others PRs that you think might improve the docs :smile: 

You can also comment here if you have some suggestions: https://github.com/jump-dev/JuMP.jl/issues/2348

PR #3805, opened 2024-08-11
Fix various bugs with Real-valued Hermitian matrices

### LINKED ISSUE #3804: Constraining Symmetric matrix to be equal to a Hermitian matrix is not Hermitian
Bug introduced in #3797: before if I tried to constrain A == B for Symmetric A and Hermitian B it would give me an error. Now it silently fails.

```julia

using JuMP
using LinearAlgebra
import Hypatia
import Ket
import Random

function dual_test(d)
    Random.seed!(1)
    ρ = Ket.random_state(Float64, d^2, 1)

    model = Model()
    @variable(model, σ[1:d^2, 1:d^2] in PSDCone())
    @variable(model, λ)
    noisy_state = Hermitian(ρ + λ * I(d^2))
    @constraint(model, witness_constraint, σ == noisy_state)
    @objective(model, Min, λ)
    @constraint(model, Ket.partial_transpose(σ, 2, [d, d]) in PSDCone())

    optimize!(model)

    W = dual(witness_constraint)
    return W
end
```

odow:
Closes #3804 

Now the dual of #3804 is Hermitian
```Julia
julia> using JuMP

julia> using LinearAlgebra

julia> import Hypatia

julia> import Ket

julia> import Random

julia> function dual_test(d)
           Random.seed!(1)
           ρ = Ket.random_state(Float64, d^2, 1)
           model = Model()
           @variable(model, σ[1:d^2, 1:d^2] in PSDCone())
           @variable(model, λ)
           noisy_state = Hermitian(ρ + λ * I(d^2))
           @constraint(model, witness_constraint, σ == noisy_state)
           @objective(model, Min, λ)
           @constraint(model, Ket.partial_transpose(σ, 2, [d, d]) in PSDCone())
           set_silent(model)
           optimize!(model)
           W = dual(witness_constraint)
           return W
       end
dual_test (generic function with 1 method)

julia> dual_test(2)
4×4 Hermitian{ComplexF64, Matrix{ComplexF64}}:
 -0.119631+0.0im   0.213316+0.0im  -0.213316+0.0im   0.380369+0.0im
  0.213316-0.0im  -0.380369+0.0im  -0.119631+0.0im   0.213316+0.0im
 -0.213316-0.0im  -0.119631-0.0im  -0.380369+0.0im  -0.213316+0.0im
  0.380369-0.0im   0.213316-0.0im  -0.213316-0.0im  -0.119631+0.0im
```

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -573,7 +573,7 @@ function Base.complex(
 end
 
 function Base.:+(
-    A::LinearAlgebra.Symmetric{V},
+    A::LinearAlgebra.Symmetric{V,Matrix{V}},
     B::LinearAlgebra.Hermitian,
 ) where {
     V<:Union{
@@ -587,7 +587,7 @@ end
 
 function Base.:+(
     A::LinearAlgebra.Hermitian,
-    B::LinearAlgebra.Symmetric{V},
+    B::LinearAlgebra.Symmetric{V,Matrix{V}},
 ) where {
     V<:Union{
         GenericVariableRef{<:Real},
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -570,6 +570,8 @@ function Base.complex(
     return r + im * i
 end
 
+# These methods exist in LinearAlgebra for subtypes of Real. Without them, we
+# return a `Matrix` which looses the Hermitian information.
 function Base.:+(
     A::LinearAlgebra.Symmetric{V,Matrix{V}},
     B::LinearAlgebra.Hermitian,
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -599,7 +599,7 @@ function Base.:+(
 end
 
 function Base.:-(
-    A::LinearAlgebra.Symmetric{V},
+    A::LinearAlgebra.Symmetric{V,Matrix{V}},
     B::LinearAlgebra.Hermitian,
 ) where {
     V<:Union{
@@ -613,7 +613,7 @@ end
 
 function Base.:-(
     A::LinearAlgebra.Hermitian,
-    B::LinearAlgebra.Symmetric{V},
+    B::LinearAlgebra.Symmetric{V,Matrix{V}},
 ) where {
     V<:Union{
         GenericVariableRef{<:Real},
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -571,3 +571,55 @@ function Base.complex(
 )
     return r + im * i
 end
+
+function Base.:+(
+    A::LinearAlgebra.Symmetric{V},
+    B::LinearAlgebra.Hermitian,
+) where {
+    V<:Union{
+        GenericVariableRef{<:Real},
+        GenericAffExpr{<:Real},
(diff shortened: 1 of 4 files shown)


odow:
We now get:
```julia
julia> using JuMP

julia> using LinearAlgebra

julia> import Hypatia

julia> import Ket

julia> import Random

julia> function dual_test(d)
           Random.seed!(1)
           ρ = Ket.random_state(Float64, d^2, 1)
           model = Model()
           @variable(model, σ[1:d^2, 1:d^2] in PSDCone())
           @variable(model, λ)
           noisy_state = Hermitian(ρ + λ * I(d^2))
           @constraint(model, witness_constraint, σ == noisy_state)
           @objective(model, Min, λ)
           @constraint(model, Ket.partial_transpose(σ, 2, [d, d]) in PSDCone())
           set_silent(model)
           optimize!(model)
           W = dual(witness_constraint)
           return W
       end
dual_test (generic function with 1 method)

julia> dual_test(2)
4×4 Symmetric{Float64, Matrix{Float64}}:
 -0.119631   0.213316  -0.213316   0.380369
  0.213316  -0.380369  -0.119631   0.213316
 -0.213316  -0.119631  -0.380369  -0.213316
  0.380369   0.213316  -0.213316  -0.119631
```

odow:
I'm not fully convinced about this one. You could argue that we should be able to return `Hermitian{Float64,Matrix{Float64}}`, which would mean being able to distinguish real- and complex-valued `HermitianMatrixSpace`.

@blegat should weigh in before we merge.

odow:
My underlying problem is that Julia has both `Symmetric` and `Hermitian{<:Real}`, which are slightly different things.

src/operators.jl
```diff
@@ -571,3 +569,55 @@ function Base.complex(
 )
     return r + im * i
 end
+
+function Base.:+(
```
blegat, line 573: What's wrong with the default implementation of `LinearAlgebra` ? Add a comment that the fault returns a `Matrix` and we overload to return a `Hermitian`

blegat:
> I'm not fully convinced about this one. You could argue that we should be able to return `Hermitian{Float64,Matrix{Float64}}`, which would mean being able to distinguish real- and complex-valued `HermitianMatrixSpace`.

It might make more sense to return `Hermitian{Float64}` and we could add a `RealHermitianMatrixSpace`. However, it could also be confusing so I think `Symmetric` has advantages. We've made an effort that the shape only depends on the type of the input but here the type is `Hermitian{<:Real}` so we see that it's synonymous to `Symmetric` so `Symmetric` seems to be the best option.

src/operators.jl
odow, line 573:
```suggestion
# These methods exist in LinearAlgebra for subtypes of Real. Without them, we
# return a `Matrix` which looses the Hermitian information.
function Base.:+(
```

odow:
> so Symmetric seems to be the best option.

Sounds good.

blegat (approved):

PR #3820, opened 2024-09-09
[docs] add example to index.md

odow:
Our landing page has a lot of text with no example: https://jump.dev/JuMP.jl/v1.23.1/

image


odow:
Now it is
image

I think it's an improvement to have an example of using JuMP on the first page of the docs.

PR #3822, opened 2024-09-10
[docs] restructure conic optimization tutorials

odow:
Our tutorials could be improved by unifying them to have a statement of purpose at the start, and by making the title what the tutorial is useful for, rather than the problem that it is solving.


blegat (approved):

odow:
So this is now
image
which I think is better than
image

PR #3828, opened 2024-09-13
Add Info box at start of OPF example

jd-foster:
The Info box highlights the link to PowerModels and its branch model approach
cf. https://discourse.julialang.org/t/variables-for-power-flow-with-condition-in-constraint/116343/3?u=jd-foster

### DIFF
--- a/docs/src/tutorials/applications/optimal_power_flow.jl
+++ b/docs/src/tutorials/applications/optimal_power_flow.jl
@@ -26,7 +26,7 @@
 #     that simplifies the construction of semidefinite programs.
 #     Another approach is to formulate the problem focusing on network lines
 #     (known as a _branch model_) where it is easier to work with flow 
-#     constraints. A general approach is provided by the Julia/JuMP package 
+#     constraints. A general approach is provided by
 #     [PowerModels.jl](https://lanl-ansi.github.io/PowerModels.jl/stable/),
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/applications/optimal_power_flow.jl
```diff
@@ -23,13 +21,15 @@
...
-# where it is easier to work with flow constraints. A general approach is
-# provided by the Julia/JuMP package [PowerModels.jl](https://lanl-ansi.github.io/PowerModels.jl/stable/),
-# an open-source framework to a broad range of power flow model formulations
-# along with utilities for working with detailed network data.
+# !!! info
+#     This tutorial takes a matrix-oriented approach focused on network nodes
+#     that simplifies the construction of semidefinite programs.
+#     Another approach is to formulate the problem focusing on network lines
+#     (known as a _branch model_) where it is easier to work with flow 
+#     constraints. A general approach is provided by the Julia/JuMP package
```
odow, line 29:
```suggestion
#     constraints. A general approach is provided by
```

odow: I don't know who started the Julia/JuMP thing. I dislike it, but it seems pretty popular. Since we're in the JuMP docs though, readers should be familiar with Package.jl for a Julia package.

odow (approved):

PR #3830, opened 2024-09-20
Small tweaks to Tolerances tutorial

jd-foster:
Nice write-up Oscar.


odow (approved):
I still need to find out why vale thinks `eample` is a word. It's happened before

odow:
Yeah I liked it. I don't really recall seeing such an explicit write up in other places. The Gurobi docs have it in pieces, but they're often just throw-away sentences like "this could happen". Seeing it in code from HiGHS is something else.

odow:
I like the example at the end where HiGHS says it is infeasible. But then if you provide a starting point it's like "hey that is actually feasible!"

jd-foster:
> I still need to find out why vale thinks eample is a word. It's happened before

Random thought: because it contains `ampl`??

PR #3833, opened 2024-10-02
[docs] add feasibility cuts to Benders decomposition

odow:
I've changed to a more realistic example, since the Matrix input was a bit misleading. I've also added feasibility cuts (at the end, to keep the initial developments simpler).

Preview: https://jump.dev/JuMP.jl/previews/PR3833/tutorials/algorithms/benders_decomposition/

x-ref https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-2103440611

### DIFF
--- a/docs/src/tutorials/algorithms/benders_decomposition.jl
+++ b/docs/src/tutorials/algorithms/benders_decomposition.jl
@@ -68,21 +68,22 @@ import Test  #src
 # V_2(\bar{x}) = & \text{min}  \ & c_2(y)\\
 #          & \text{subject to} \ & f_2(y) \in S_2 \\
 #          &                     & f_3(x, y) \in S_3 \\
-#          &                     & x = \bar{x} \\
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/algorithms/benders_decomposition.jl
```diff
@@ -202,29 +245,32 @@ for k in 1:MAXIMUM_ITERATIONS
...
 
 optimize!(model)
 @assert is_solved_and_feasible(model)
-Test.@test value.(x) == [0.0, 1.0]  #src
 x_optimal = value.(x)
+optimal_ret = solve_subproblem(x_optimal)
+iterative_solution = optimal_flows(optimal_ret.y)
 
-#-
+# which is the same as the monolithic_solution:
```
odow, line 267:
```suggestion
# which is the same as the monolithic solution:
```

Issue #3806, opened 2024-08-14
Where is JuMP's sparse AD code, and can DifferentiationInterface help?


gdalle:
Hi there @odow and friends!

Looking at our work on sparse autodiff with @adrhill and @amontoison, I've been wondering how it could be useful to JuMP. We have developed a combination of three new packages:
- for sparsity detection ([SparseConnectivityTracer.jl](https://github.com/adrhill/SparseConnectivityTracer.jl))
- for matrix colorings and decompression ([SparseMatrixColorings.jl](https://github.com/gdalle/SparseMatrixColorings.jl))
- for backend-independent gradients, (sparse) Jacobians and (sparse) Hessians ([DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl))

From what I understand, JuMP's current sparse AD engine is [`Nonlinear.ReverseAD`](https://jump.dev/JuMP.jl/stable/moi/submodules/Nonlinear/overview/#ReverseAD), but there are also experiments going on in [MathOptSymbolicAD.jl](https://github.com/lanl-ansi/MathOptSymbolicAD.jl). Did I miss anything?

I'm not suggesting that DifferentiationInterface and friends should replace your own default AD solution. But since you say in the docs that testing other AD backends would be nice, maybe there's an angle there?
In any case, perhaps DI could be a nice addition to the [docs page on autodiff of user-defined operators](https://jump.dev/JuMP.jl/stable/tutorials/nonlinear/operator_ad/)?

odow:
> Did I miss anything?

Nope. That is the situation. All nonlinear/AD stuff is in the MOI.Nonllinear module. There is nothing in JuMP.

> But since you say in the docs that testing other AD backends would be nice, maybe there's an angle there?

New AD backends need to be implemented as an oracle like this:

https://github.com/lanl-ansi/MathOptSymbolicAD.jl/blob/e5d3eaa1f56346d53a5ef8f895743ab3e89ff38c/src/nonlinear_oracle.jl#L278-L496

> In any case, perhaps DI could be a nice addition to the [docs page on autodiff of user-defined operators](https://jump.dev/JuMP.jl/stable/tutorials/nonlinear/operator_ad/)?

Feel free to open a PR 😄

amontoison:
What could be useful for JuMP is star coloring for the Hessian of the Lagrangian.
[...]

PR #3836, opened 2024-10-06
[docs] update autodiff tutorial

gdalle:
Two modifications to the tutorial on autodiff of user-defined operators:

1. Upgrade to Enzyme v0.13 (no need for `autodiff_deferred` anymore). I didn't necessarily go for the most efficient solutions, you may want to ask @wsmoses for that, I just performed the minimal changes to gain compatibility with the breaking release.
2. Add a section demonstrating DifferentiationInterface. For some reason it doesn't work with `AutoEnzyme()` but it's a start.

Related issues:

- #3806
- https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-2295445264


gdalle:
I had a question about the use of DI. For many backends, performance is greatly improved by the preparation mechanism, which allows you to record a tape / preallocate a cache / etc. But to apply it, you need an example input that has the same type and size as the ones you'll encounter during optimization. Is there any way we can use this mechanism in JuMP?

docs/src/tutorials/nonlinear/operator_ad.jl
```diff
@@ -324,3 +325,105 @@ function enzyme_rosenbrock()
...
 enzyme_rosenbrock()
+
+# ## DifferentiationInterface
+
+# Julia offers [many different autodiff packages](https://juliadiff.org/).
+# [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl)
+# is a package that provides an abstraction layer across multiple underlying
+# autodiff libaries.
+
+# All the necessary information about your choice of underlyingg autodiff
```
odow, line 336:
```suggestion
# All the necessary information about your choice of underlying autodiff
```

odow:
You could just use the first call to build the tape, and then error if there are calls from different inputs. JuMP doesn't provide anything special. You can do everything in user-land.

docs/src/tutorials/nonlinear/operator_ad.jl
```diff
@@ -324,3 +325,105 @@ function enzyme_rosenbrock()
...
 
 enzyme_rosenbrock()
+
+# ## DifferentiationInterface
+
+# Julia offers [many different autodiff packages](https://juliadiff.org/).
+# [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl)
+# is a package that provides an abstraction layer across multiple underlying
+# autodiff libaries.
+
```
wsmoses, line 335: Perhaps add a disclaimer:

While DI makes it easier to experiment with different autodiff tools, this comes at a cost. In particular, DI may produce slower derivatives than using an AD tool directly, and sometimes fail to work at all. For example, in this code passing AutoEnzyme to DI fails to differentiate the code which is differentiable with Enzyme directly like above.

gdalle: By the way do you have any idea why it fails here? As usual it's a segfault so not particularly user friendly

wsmoses: No idea, I haven’t even tried to run it or seen what the error message, just took your word that it failed xD.

open an issue with the mwe and error and can take a look

odow: > and sometimes fail to work at all

I don't want to mention something in our documentation if this is the case. 

What is the failure mode? Will we gracefully report errors? Segfaults are not good.

wsmoses: Oh agreed, in that case @gdalle just drop DI docs then?

gdalle: > open an issue with the mwe and error and can take a look

The MWE is literally this docs page but sure, I'll open an issue tomorrow on the Enzyme repo.

> I don't want to mention something in our documentation if this is the case.

To be fair, Enzyme will also sometimes fail to differentiate code, and throw very unpleasant errors. 

> What is the failure mode? Will we gracefully report errors? Segfaults are not good.

The fact that it is a segfault is not due to DI per se, it's Enzyme not catching the error gracefully. Sure, DI might be using Enzyme wrong in this particular case (perhaps because of a function returning two closures), but I too would very much prefer a clean error message from Enzyme to a segfault. Will try to reproduce tomorrow.

gdalle: > Oh agreed in that segfaults are bad, in that case @gdalle just drop DI docs then?

I'll try to debug the MWE with Enzyme, but it is by far the pickiest of backends, so I'm reasonably sure that it would work effortlessly if you plugged in Zygote or FiniteDiff. So I think there is still value in showcasing DI, with the proper warnings of course. @odow your call.

docs/src/tutorials/nonlinear/operator_ad.jl
```diff
@@ -324,3 +325,105 @@ function enzyme_rosenbrock()
 end
 
 enzyme_rosenbrock()
+
+# ## DifferentiationInterface
+
+# Julia offers [many different autodiff packages](https://juliadiff.org/).
+# [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl)
+# is a package that provides an abstraction layer across multiple underlying
+# autodiff libaries.
```
odow, line 334:
```suggestion
# autodiff libraries.
```

docs/src/tutorials/nonlinear/operator_ad.jl
```diff
@@ -331,9 +331,9 @@ enzyme_rosenbrock()
 # Julia offers [many different autodiff packages](https://juliadiff.org/).
 # [DifferentiationInterface.jl](https://github.com/gdalle/DifferentiationInterface.jl)
 # is a package that provides an abstraction layer across multiple underlying
```
odow, line 333:
```suggestion
# is a package that provides an abstraction layer across a few underlying
```

docs/src/tutorials/nonlinear/operator_ad.jl
odow: I realized my changes mean we never actually mention Enzyme in the context of DI, so perhaps that's okay.

wsmoses: > The fact that it is a segfault is not due to DI per se, it's Enzyme not catching the error gracefully. Sure, DI might be using Enzyme wrong in this particular case (perhaps because of a function returning two closures), but I too would very much prefer a clean error message from Enzyme to a segfault. Will try to reproduce tomorrow.

Again haven't seen the error message so not clear what's up, but I'm unaware of any segfaults caused by Enzyme atm. Most the time people saw segfaults in the past were actually bugs in the Julia compiler itself (e.g. https://github.com/JuliaLang/julia/pull/55306 ). 

That said unfortunately even if you're not the underlying cause of an error, the top level package you use is going to be the one people blame for issues (especially if code works otherwise without it). Analogously, there's a ton of code in Enzyme working around bugs in the Julia and LLVM compiler errors =/. Perhaps something similar would be good in DI?

gdalle: It is not DI's job to work around LLVM errors, it's Enzyme's. DI is a fancy argument-passer

gdalle: And here's the MWE: https://github.com/EnzymeAD/Enzyme.jl/issues/1942

gdalle: For the record, pure Enzyme also fails on this one when one uses `splat(f)`

PR #3840, opened 2024-10-07
Convert AbstractString to String in set_attribute

### LINKED ISSUE #3839: String valued MOI attributes: Improved error messages or `AbstractString` instead of `String`?
The following code produces an error that I thought could be worth pointing out, since I assume it could be either

- caught somewhere in a "fallback" to produce a hint towards the actual mistake, or
- be solved by moving from `String` to, e.g., `AbstractString` and internally just doing `string(value)`.

However I didn't manage to find the actual location where the `value` in some `MOI.set(...)` function is typed as `String`, and I'm unsure if there are other attributes where allowing to pass an `AbstractString` would negatively influence something.

It took me embarrassingly long to get that this was just caused by the `SubString` that made it's way into the model name, because I got fooled by the error and stacktrace ...

### Code

```julia
using JuMP

model_name = split("foo.bar", ".")[1]

model = Model()
set_attribute(model, MOI.Name(), model_name)          # error
set_attribute(model, MOI.Name(), string(model_name))  # works
```

### Error message

```console
ERROR: MathOptInterface.SetAttributeNotAllowed{MathOptInterface.Name}: Setting attribute MathOptInterface.Name()
cannot be performed. You may want to use a `CachingOptimizer` in `AUTOMATIC` mode or you may need to call
`reset_optimizer` before doing this operation if the `CachingOptimizer` is in `MANUAL` mode.
```

odow:
Closes #3839

<s>I want to do an extension test. Anyone expecting some subtype of `AbstractString` will now get a `String`. But it seems that this is a bug fix that addresses the most common use-cases...</s> This is now safe.

https://github.com/jump-dev/JuMP.jl/actions/runs/11223971621

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@latest
         with:
-          version: '1'
+          version: '1.10'
       - name: Install dependencies
         shell: julia --color=yes --project=docs/ {0}
         run: |
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -1251,9 +1251,25 @@ function get_attribute(
     return get_attribute(model, String(name))
 end
 
-_string_if_abstract_string(value) = value
-_string_if_abstract_string(value::String) = value
-_string_if_abstract_string(value::AbstractString) = String(value)
+# Some MOI attributes have a strict value type, like ::String or ::Float64, but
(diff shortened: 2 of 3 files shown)


.github/workflows/documentation.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@latest
         with:
-          version: '1'
+          version: '1.10'
```
odow, line 17: Docs look like they broke on 1.11

blegat (approved):

PR #3841, opened 2024-10-08
[docs] enable printing in TSP tutorial

odow:
https://jump.dev/JuMP.jl/previews/PR3841/tutorials/algorithms/tsp_lazy_constraints/#Lazy-constraint-method


odow:
My motivation for adding the printing is that it appeared to fail in the 1.11 build :smile:

Looks like a few weird issues going from 1.10 to 1.11.

PR #3842, opened 2024-10-08
[Docs] Add missing warning from #3836

wsmoses:
Looks like the warning got dropped before merging

### DIFF
--- a/docs/src/tutorials/nonlinear/operator_ad.jl
+++ b/docs/src/tutorials/nonlinear/operator_ad.jl
@@ -333,6 +333,11 @@ enzyme_rosenbrock()
 # is a package that provides an abstraction layer across a few underlying
 # autodiff libraries.
 
+# !!! warning
+#     While DI makes it easier to experiment with different autodiff tools, this comes at a cost.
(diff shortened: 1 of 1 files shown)


odow:
I decided not to add this in that PR because we never mention AutoEnzyme. I'm reticent to specifically call out DI for failing to work with one example with Enzyme, and "May be slower" is not important (performance is not the most important consideration) for JuMP users.

We also already have a native Enzyme code example in the tutorial above, which seems preferable to me.

How about something more generic:
```Julia
# !!! warning
#     The JuMP developers cannot help you debug error messages related to
#     DifferentiationInterface. If the operator works, it works. If not, we
#     suggest you try a different automatic differentiation library. See
#     [juliadiff.org](https://juliadiff.org/) for details.
```

odow:
Doc build failed because of the switch in the last few hours from Julia 1.10 to Julia 1.11. It's fixed (by pinning 1.10) on `master`.

wsmoses:
sure/done

odow (approved):

PR #3843, opened 2024-10-08
[docs] debug failure on v1.11

odow:
Doc build is currently failing on 1.11

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@latest
         with:
-          version: '1.10'
+          version: '1'
       - name: Install dependencies
         shell: julia --color=yes --project=docs/ {0}
         run: |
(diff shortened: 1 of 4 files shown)


odow:
So it seems like the TSP tutorial has broken. I have no idea why

odow:
My conclusion is that the lazy constraint algorithm in GLPK adds local lazy, so it's just pretty terrible at repeating existing work.

blegat:
Maybe we should use SCIP instead of GLPK but looking at SCIP.jl, it supports `MOI.HeuristicCallback` and `MOI.UserCutCallback` but not `MOI.LazyConstraintCallback`. @matbesancon what is blocking SCIP.jl from adding support to the solver-agnostic lazy callback ?

matbesancon:
Nothing really, there is a constraint handler mechanism that should be a
superset of the lazy constraint, so one could implement a restricted
version for LazyConstraintCallback's sake. And this would keep OSS solvers
for the tests, for people to be able to try things out.

PR #3853, opened 2024-10-18
Add SHOT to list of supported solvers

blegat:
This doesn't check most of the item but it's the same as Couenne and BONMIN, I don't think the list applies to the solvers available through AMPLNLWriter

## Basic

 - [ ] Check that the solver is a registered Julia package
 - [ ] Check that the solver supports the long-term support release of Julia
 - [ ] Check that the solver has a MathOptInterface wrapper
 - [ ] Check that the tests call `MOI.Test.runtests`. Some test excludes are
       documented.
 - [ ] Check that the README and/or documentation provides an example of how to
       use the solver with JuMP

## Documentation

## Optional

 - [ ] Add package metadata to `docs/packages.toml`


odow:
I've previously held off adding it because of 
https://github.com/jump-dev/AmplNLWriter.jl/blob/69220aff78bd769b164e960926a5947f568bff6d/test/MINLPTests/run_minlptests.jl#L72-L99

odow:
It is what it is

PR #3856, opened 2024-10-22
Fix broken PDF format link and create UpdatedIndexforPDF.md

Akshit222:
This pull request addresses several issues in the documentation related to PDF generation.
  -Fixed broken PDF link in index.md.
  -Created UpdatedIndexforPDF.md to serve as a clean version for PDF generation, avoiding HTML syntax issues present in the 
    original index.md.
  -Addressed numerous broken links in index.md, which necessitated the creation of UpdatedIndexforPDF.md to ensure the PDF 
    documentation routes correctly.
  -Enhanced the new UpdatedIndexforPDF.md to include underlines for text with attached links, improving usability.
  -Plan to submit a separate pull request to fix all broken links in the original index.md.
  -As this is my first pull request, I appreciate any feedback or suggestions for improvement.

PR #3857, opened 2024-10-22
This pull request addresses broken URLs in the JuMP documentation in index.md 

Akshit222:
The following sections have been updated:

- JuMP.pdf broken  URL is updated but it will only work if my another pull request fixing the pdf format of index.md file gets accepts -Installation Guide URL was fixed
- Getting Started with Julia URL was fixed
- Getting Started with JuMP URL was fixed

These changes ensure that users can access the correct resources and improve the overall usability of the documentation.
There were a few other broken URL in index.md like :
      The diet problem
      Maximum likelihood estimation
But i didn't know what is the correct URL for them

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -19,7 +19,8 @@
 Welcome to the documentation for JuMP.
 
 !!! note
-    This documentation is also available in PDF format: [JuMP.pdf](JuMP.pdf).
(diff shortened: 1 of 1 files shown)


odow:
Closing for the same reason as https://github.com/jump-dev/JuMP.jl/pull/3856#issuecomment-2430110131

PR #3859, opened 2024-10-23
Clarify usage of `add_to_expression!` in docs

joaquimg:
There were comments about combining two expressions but no examples, so this was a bit more hidden.

### DIFF
--- a/docs/src/manual/expressions.md
+++ b/docs/src/manual/expressions.md
@@ -126,6 +126,32 @@ julia> add_to_expression!(ex, 1.0, y)
 2 x + y - 1
 ```
 
+This is a efficient way to combine previously created expressions.
+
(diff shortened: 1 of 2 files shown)


docs/src/manual/expressions.md
```diff
@@ -126,6 +126,32 @@ julia> add_to_expression!(ex, 1.0, y)
 2 x + y - 1
 ```
 
+This is a efficient way to combine previously created expressions.
```
blegat, line 129:
```suggestion
This is an efficient way to combine previously created expressions.
```

blegat: You should stress here that it modifies the first expression

odow (approved):

PR #3865, opened 2024-11-03
Use add_constrained_variable with two sets when adding bounded variables

odow:
Requires 
 - [x] https://github.com/jump-dev/MathOptInterface.jl/pull/2574
 - [x] MOI v1.34


src/variables.jl
```diff
@@ -1981,6 +1981,15 @@ function _moi_add_constrained_variable(
     return x
 end
 
+function _moi_add_constrained_variable(
+    moi_backend::MOI.ModelLike,
+    ::Nothing,
+    set::Tuple{MOI.GreaterThan{T},MOI.LessThan{T}}
+) where {T}
```
odow, line 1988:
```suggestion
) where {T<:Real}
```

src/variables.jl
```diff
@@ -1981,6 +1981,15 @@ function _moi_add_constrained_variable(
     return x
 end
 
+function _moi_add_constrained_variable(
+    moi_backend::MOI.ModelLike,
+    ::Nothing,
+    set::Tuple{MOI.GreaterThan{T},MOI.LessThan{T}}
```
odow, line 1987:
```suggestion
    set::Tuple{MOI.GreaterThan{T},MOI.LessThan{T}},
```

odow:
cc @joaquimg

PR #3872, opened 2024-11-06
Simplify how we copy docstrings for MOI enums

odow:


odow:
I don't know what I was thinking the first time around. This one is much simpler, and we now link explicitly to the MOI source code.

Issue #3738, opened 2024-04-25
Nonlinear subexpressions


odow:
Background

 * JuMP has long supported nonlinear subexpressions created with `@NLexpression`.
 * The AD system in `MOI.Nonlinear` has support for representing expressions, and dealing with them efficiently
 * When we created `ScalarNonlinearFunction`, we did not create an analogous `ScalarNonlinearExpression` object.
 * In most cases, nonlinear expressions make little difference, or are a minor win.
 * An upside and a downside is that it's up to users to decide which expressions should be common. I have seen, anecdotally, that users on the forum either use no `@NLexpression`, or they made every possible expression a `@NLexpression`
 * Best performance is to judiciously use common subexpressions where it would make a difference.
 * The best real-world example we have is @mitchphillipson's MCP model: https://github.com/jump-dev/JuMP.jl/issues/3729#issuecomment-2060586587

The question is how to achieve this in JuMP.

I opened an issue in MOI: https://github.com/jump-dev/MathOptInterface.jl/issues/2488

Short of rewriting much of the `MOI.Nonlinear` module to use a single DAG of expressions (instead of the current tree), we _could_ pass the expressions through to MOI, and then attempt to detect common subexpressions. This would rely on a heuristic of when it was beneficial to do so.

A simpler approach would be to add a "nonlinear expression" set to MOI (and JuMP), just like we've done for Parameter.

A crude API would be:
```Julia
@variable(model, y in CommonSubexpression())
@constraint(model, [y; f(x)] in CommonSubexpression())
```
with the fallback to a bridge
```Julia
@variable(model, y)
@constraint(model, y == f(x))
```
and maybe one for MCP:
```Julia
@variable(model, y)
@constraint(model, f(x) - y ⟂ y)
```

We could come up with nicer syntax, for example:
```julia
@expression(model, y, f(x), NonlinearSubexpression())
```

odow:
`@common_expression(model, y, f(x))`

```julia
@variable(model, y)
@constraint(model, y :== f(x))
```

Robbybp:
What is stopping `@expression(model, y, f(x))` from adopting the common-subexpression behavior you're describing? Would this require breaking changes to MOI?

odow:
> What is stopping @expression(model, y, f(x)) from adopting the common-subexpression behavior you're describing? 

[...]

PR #3878, opened 2024-11-12
Add subexpression kwarg to expression macro

odow:
Trying out an idea from #3738. We don't necessarily need to merge this. Open to thoughts/feedback.

The benefit of this approach is simplicity.

The downsides are that it doesn't solve the issue in MOI, it doesn't utilise the subexpression support in `MOI.Nonlinear`, and we don't have a way for the user to query the actual subexpression.

Perhaps we should return a tuple of the `VariableRef` and the expression, so people could use either-or?

### DIFF
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -2487,4 +2487,92 @@ function test_array_scalar_sets()
     return
 end
 
+function test_subexpression_kwarg()
+    model = Model()
+    @variable(model, x)
+    @expression(model, ex, sin(x), subexpression = true)
+    @test ex isa VariableRef
+    @test model[:ex] isa VariableRef
+    @test model[:ex] === ex
+    @test occursin("ex - sin(x) = 0", sprint(print, model))
+    @test num_variables(model) == 2
+    return
+end
+
+function test_subexpression_kwarg_array()
+    model = Model()
+    @variable(model, x[1:2])
+    @expression(model, ex[i in 1:2], sin(x[i]), subexpression = true)
+    @test ex isa Vector{VariableRef}
+    @test model[:ex] === ex
+    @test occursin("ex[1] - sin(x[1]) = 0", sprint(print, model))
+    @test occursin("ex[2] - sin(x[2]) = 0", sprint(print, model))
+    @test num_variables(model) == 4
+    return
+end
+
+function test_subexpression_kwarg_dense_axis_array()
+    model = Model()
+    @variable(model, x[2:3])
+    @expression(model, ex[i in 2:3], sin(x[i]), subexpression = true)
+    @test ex isa Containers.DenseAxisArray{VariableRef}
+    @test model[:ex] === ex
+    @test occursin("ex[2] - sin(x[2]) = 0", sprint(print, model))
+    @test occursin("ex[3] - sin(x[3]) = 0", sprint(print, model))
+    @test num_variables(model) == 4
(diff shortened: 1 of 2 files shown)


src/macros.jl
```diff
@@ -474,3 +474,30 @@ for file in readdir(joinpath(@__DIR__, "macros"))
...
+function _build_subexpression(
+    ::Function,
+    model::AbstractModel,
+    expr::AbstractJuMPScalar,
+    name::String,
+)
+    y = @variable(model)
+    set_name(y, name)
+    @constraint(model, y == expr)
+    return y
```
odow, line 490:
```suggestion
    return y, expr
```
This would enable
```julia
julia> model = Model()
A JuMP Model
├ solver: none
├ num_variables: 0
├ num_constraints: 0
└ Names registered in the model: none

julia> @variable(model, x)
x

julia> @expression(model, ex, sin(x), subexpression = true)
(ex, sin(x))

julia> ex
(ex, sin(x))

julia> model[:ex]
(ex, sin(x))

julia> model = Model()
A JuMP Model
├ solver: none
├ num_variables: 0
├ num_constraints: 0
└ Names registered in the model: none

 x[1]
 x[2]

julia> @expression(model, ex[i in 1:2], sin(x[i]), subexpression = true)
2-element Vector{Tuple{VariableRef, NonlinearExpr}}:
 (ex[1], sin(x[1]))
 (ex[2], sin(x[2]))

julia> first.(ex)
 ex[1]
 ex[2]

julia> last.(ex)
2-element Vector{NonlinearExpr}:
 sin(x[1])
 sin(x[2])
```

which seems quite nice.

test/test_macros.jl
```diff
@@ -2487,4 +2487,92 @@ function test_array_scalar_sets()
...
+    @test num_variables(model) == 2
+    return
+end
+
+function test_subexpression_kwarg_dict_element()
+    model = Model()
+    @variable(model, x[i in 1:2])
+    @test_throws_runtime(
+        ErrorException(
+            "In `@expression(model, ex, Dict((i => x[i] for i = 1:2)), subexpression = true)`: Unable to build a subexpression for the type Dict{Int64, VariableRef}",
```
odow, line 2566:
```suggestion
            "In `@expression(model, ex, Dict((i => x[i] for i = 1:2)), subexpression = true)`: Unable to build a subexpression for the type $(Dict{Int,VariableRef})",
```

test/test_macros.jl
```diff
@@ -2487,4 +2487,92 @@ function test_array_scalar_sets()
...
 end
 
+function test_subexpression_kwarg()
+    model = Model()
+    @variable(model, x)
+    @expression(model, ex, sin(x), subexpression = true)
+    @test ex isa VariableRef
+    @test model[:ex] isa VariableRef
+    @test model[:ex] === ex
+    @test occursin("ex - sin(x) = 0", sprint(print, model))
```
odow, line 2497:
```suggestion
    @test occursin(r"ex - sin\(x\) ==? 0", sprint(print, model))
```

joaquimg:
I don't like `@expression` adding variables and constraints. I would force users to do both steps or create a new macro `@subexpression` (which still seems to be a very suboptimal solution, rethinking the name might be needed).

joaquimg:
Having it in the docs (as is) seems to be a better way.

PR #3879, opened 2024-11-13
[docs] add section and example on common subexpression elimination

odow:
Part of #3738 

https://jump.dev/JuMP.jl/previews/PR3879/manual/nonlinear/#Common-subexpressions

### DIFF
--- a/docs/src/manual/nonlinear.md
+++ b/docs/src/manual/nonlinear.md
@@ -153,6 +153,52 @@ julia> sin(sin(1.0))
 0.7456241416655579
 ```
 
+## Common subexpressions
+
+JuMP does not perform common subexpression elimination. Instead, if you re-use
+an expression in multiple places, JuMP will insert a copy of the expression.
+
+JuMP's lack of common subexpression elimination is a common cause of performance
+problems, particularly in nonlinear models with a pattern like
+`sum(t / common_term for t in terms)`. One example is the logistic loss:
+
+```jldoctest
+julia> model = Model();
+
+julia> @variable(model, x[1:2]);
+
+julia> @expression(model, expr, sum(exp.(x)))
+0.0 + exp(x[2]) + exp(x[1])
+
+julia> @objective(model, Min, sum(exp(x[i]) / expr for i in 1:2))
+(exp(x[1]) / (0.0 + exp(x[2]) + exp(x[1]))) + (exp(x[2]) / (0.0 + exp(x[2]) + exp(x[1])))
+```
+In this model, JuMP will compute the value (and derivatives) of the denominator
+twice, without realizing that the same expression appears twice.
+
+As a work-around, create a new [`@variable`](@ref) and use an `==`
+[`@constraint`](@ref) to constrain the value of the variable to the
+subexpression.
+
+```jldoctest
+julia> model = Model();
+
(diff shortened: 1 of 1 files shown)


docs/src/manual/nonlinear.md
```diff
@@ -153,6 +153,52 @@ julia> sin(sin(1.0))
...
+julia> @constraint(model, expr == sum(exp.(x)))
+expr - (0.0 + exp(x[2]) + exp(x[1])) = 0
+
+julia> @objective(model, Min, sum(exp(x[i]) / expr for i in 1:2))
+(exp(x[1]) / expr) + (exp(x[2]) / expr)
+```
+
+The reason JuMP does not perform common subexpression elimination automatically
+is for simplicity, and because there is a trade-off: for simple expressions, the
+extra complexity of detecting and merging common subexpressions may outweight
```
odow, line 199:
```suggestion
extra complexity of detecting and merging common subexpressions may outweigh
```

docs/src/manual/nonlinear.md
```diff
@@ -153,6 +153,52 @@ julia> sin(sin(1.0))
...
+julia> @variable(model, x[1:2]);
+
+julia> @expression(model, expr, sum(exp.(x)))
+0.0 + exp(x[2]) + exp(x[1])
+
+julia> @objective(model, Min, sum(exp(x[i]) / expr for i in 1:2))
+(exp(x[1]) / (0.0 + exp(x[2]) + exp(x[1]))) + (exp(x[2]) / (0.0 + exp(x[2]) + exp(x[1])))
+```
+In this model, JuMP will compute the value (and derivatives) of the denominator
+twice, without realizing that the same expression appears twice.
```
odow, line 177:
```suggestion
twice, without realizing that it is the same expression.
```

docs/src/manual/nonlinear.md
```diff
@@ -153,6 +153,52 @@ julia> sin(sin(1.0))
...
+expr - (0.0 + exp(x[2]) + exp(x[1])) = 0
+
+julia> @objective(model, Min, sum(exp(x[i]) / expr for i in 1:2))
+(exp(x[1]) / expr) + (exp(x[2]) / expr)
+```
+
+The reason JuMP does not perform common subexpression elimination automatically
+is for simplicity, and because there is a trade-off: for simple expressions, the
+extra complexity of detecting and merging common subexpressions may outweight
+the cost of computing them independently.
```
odow, line 200:
```suggestion
the cost of computing them independently. Instead, we leave it to the user to
decide which expressions to extract as common subexpressions.
```

odow:
Doc failure is https://github.com/jump-dev/JuMP.jl/pull/3880

docs/src/manual/nonlinear.md
```diff
@@ -153,6 +153,53 @@ julia> sin(sin(1.0))
 0.7456241416655579
 ```
 
+## Common subexpressions
+
+JuMP does not perform common subexpression elimination. Instead, if you re-use
```
odow, line 158:
```suggestion
JuMP does not perform [common subexpression elimination](https://en.wikipedia.org/wiki/Common_subexpression_elimination).
Instead, if you re-use
```

PR #3883, opened 2024-11-14
Fix bug which mutates user expressions in constraint macro

### LINKED ISSUE #3882: Bug mutating expression in interval constraint
Found by https://discourse.julialang.org/t/unexpected-behavior-containerized-expression-automatically-redefined-when-included-in-interval-constraint/122648

```Julia
julia> using JuMP

julia> model = Model()
A JuMP Model
├ solver: none
├ num_variables: 0
├ num_constraints: 0
└ Names registered in the model: none

julia> @variable(model, x)
x

julia> @expression(model, a[1:1], x+1)
1-element Vector{AffExpr}:
 x + 1

julia> a
1-element Vector{AffExpr}:
 x + 1

julia> @constraint(model, -1 <= a[1] <= 1)
x ∈ [-2, 0]

julia> a
1-element Vector{AffExpr}:
 x
```

odow:
Closes #3882

This is a temporary fix. A longer-term approach is to return the `is_mutable::Bool` value from MutableArithemtics:
https://github.com/jump-dev/MutableArithmetics.jl/blob/d47000d5e02100458aff33670684a95f105ec40a/src/rewrite.jl#L341
which would let JuMP decide whether it needed to call `copy_if_mutable`.

### DIFF
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -2490,7 +2490,7 @@ end
 function test_do_not_mutate_expression_double_sided_comparison()
     model = Model()
     @variable(model, x)
-    @expression(model, a[1:1], x+1)
+    @expression(model, a[1:1], x + 1)
     @constraint(model, -1 <= a[1] <= 1)
     @test isequal_canonical(a[1], x + 1)
     return
@@ -2499,7 +2499,7 @@ end
 function test_do_not_mutate_expression_single_sided_comparison()
     model = Model()
     @variable(model, x)
-    @expression(model, a[1:1], x+1)
+    @expression(model, a[1:1], x + 1)
     @constraint(model, a[1] >= 1)
     @test isequal_canonical(a[1], x + 1)
     return
@@ -2508,7 +2508,7 @@ end
 function test_do_not_mutate_expression_in_set()
     model = Model()
     @variable(model, x)
-    @expression(model, a[1:1], x+1)
+    @expression(model, a[1:1], x + 1)
     @constraint(model, a[1] in MOI.Interval(-1, 1))
     @test isequal_canonical(a[1], x + 1)
     return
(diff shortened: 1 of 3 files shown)


test/test_macros.jl
```diff
@@ -2487,4 +2487,31 @@ function test_array_scalar_sets()
     return
 end
 
+function test_do_not_mutate_expression_double_sided_comparison()
+    model = Model()
+    @variable(model, x)
+    @expression(model, a[1:1], x+1)
```
odow, line 2493:
```suggestion
    @expression(model, a[1:1], x + 1)
```

test/test_macros.jl
```diff
@@ -2487,4 +2487,31 @@ function test_array_scalar_sets()
...
+    @constraint(model, -1 <= a[1] <= 1)
+    @test isequal_canonical(a[1], x + 1)
+    return
+end
+
+function test_do_not_mutate_expression_single_sided_comparison()
+    model = Model()
+    @variable(model, x)
+    @expression(model, a[1:1], x+1)
```
odow, line 2502:
```suggestion
    @expression(model, a[1:1], x + 1)
```

test/test_macros.jl
```diff
@@ -2487,4 +2487,31 @@ function test_array_scalar_sets()
...
+    @constraint(model, a[1] >= 1)
+    @test isequal_canonical(a[1], x + 1)
+    return
+end
+
+function test_do_not_mutate_expression_in_set()
+    model = Model()
+    @variable(model, x)
+    @expression(model, a[1:1], x+1)
```
odow, line 2511:
```suggestion
    @expression(model, a[1:1], x + 1)
```

blegat (approved):
Suprising this went unnoticed

Issue #1654, opened 2018-12-01
Performance of small constraints


blegat:
Creating small constraints like
```julia
@variable(model, x)
@variable(model, y)
@constraint(model, x >= y)
```
is rather costly compared to JuMP v0.18. The reason is that creating a `OrderedDict` of two elements is a lot slower than creating a `Vector` of two elements:
```julia
julia> using DataStructures

julia> od() = OrderedDict{Int, Int}()
od (generic function with 1 method)

julia> d() = Dict{Int, Int}()
d (generic function with 1 method)

julia> v() = Int[]
v (generic function with 1 method)

julia> using BenchmarkTools

julia> @btime od()
  73.591 ns (4 allocations: 352 bytes)
OrderedDict{Int64,Int64} with 0 entries

julia> @btime d()
  91.941 ns (4 allocations: 608 bytes)
Dict{Int64,Int64} with 0 entries

julia> @btime v()
  19.997 ns (1 allocation: 80 bytes)
0-element Array{Int64,1}

julia> od2() = OrderedDict{Int, Int}(1 => 2, 2 => 3)
od2 (generic function with 1 method)

julia> d2() = Dict{Int, Int}(1 => 2, 2 => 3)
d2 (generic function with 1 method)

julia> v2() = Int[2, 3]
v2 (generic function with 1 method)

julia> @btime od2()
  188.935 ns (9 allocations: 656 bytes)
OrderedDict{Int64,Int64} with 2 entries:
  1 => 2
  2 => 3

julia> @btime d2()
  116.193 ns (6 allocations: 672 bytes)
Dict{Int64,Int64} with 2 entries:
  2 => 3
  1 => 2

julia> @btime v2()
  20.386 ns (1 allocation: 96 bytes)
2-element Array{Int64,1}:
 2
 3
```

Maybe we could create a custom dict optimized for a small number of elements that would not create the internal dictionary if there is 2 elements or less.
```julia
struct CrazyDict{K, V}
    data::Union{Nothing, OrderedDict{K, V}}
    key1::Union{Nothing, K}
    value1::Union{Nothing, V}
    key2::Union{Nothing, K}
    value2::Union{Nothing, V}
end
```
That would avoid creating a dictionary for small number of elements.

ccoffrin:
[...]

PR #3890, opened 2024-11-22
Parametrized container

blegat:
Still very much a draft. I'll add more details later.
When the user uses the container syntax to create constraints, MOI receives each constraint independently so ExaModels.jl has to recover the repeated structure. We discussed during the JuMP-dev hackathon of a way to communicate the repeated structure through MOI more explicitly and this PR is a POC the resulting design.
As a byproduct, this also gives a solution for https://github.com/jump-dev/JuMP.jl/issues/1654 since the constraint will be expanded at the MOI level with a bridge from IteratedFunction in Zeros to Scalar?Function in EqualTo (we need a way to know the scalar function type but that should be doable)
We can also recover the pretty printing of containers that we had in JuMP v0.18

### DIFF
--- a/it.jl
+++ b/it.jl
@@ -1,0 +1,196 @@
+using JuMP
+model = Model()
+@variable(model, x)
+
+struct Iterator
+    length::Int
+    values::Vector{Float64}
+    function Iterator(length::Int)
+        return new(length, Float64[])
+    end
+end
+
+function num_values(it::Iterator)
+    return div(length(it.values), it.length)
+end
+
+function values_at(it::Iterator, i)
+    return it.values[(1 + it.length * (i - 1)):it.length * i]
+end
+
+struct IteratorValues{I}
+    iterator::Iterator
+    values::I
+end
+
+struct IteratorInExpr
+    iterator::Iterator
+    index::Int
+end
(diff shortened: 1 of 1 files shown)


mlubin:
Does this change the semantics of when and how often terms are evaluated?
What if you write something like:
```
@constraint(model, [i in [:a, :b]], x + d1[database_lookup(i)] >= d2[database_lookup(i)], container = ParametrizedArray)
```

blegat:
Good question, this will not change the semantic, your subexpressions depending only on `i` but not any JuMP expressions could be any code (although this doesn't work yet but that's the idea). Once you start interacting with JuMP expressions, we record the value (so here the value of `database_lookup(i)`) obtained for each value of the iterator and this value has to be a `Float64`. At the MOI level, we pass the value obtained when we evaluated `database_lookup(i)` so MOI receives the `Float64` value of each sub-expression depending on this iterator. I assume you don't have sub-expression depending on two iterators at the same time yet ^^

blegat:
Moved to https://github.com/blegat/GeneratorOptInterface.jl as this is very experimental so it will be easier to experiment as a separate package

PR #3895, opened 2024-11-27
Add ellipse fitting example

langestefan:
Draft PR for the ellipse fitting example, will add code.

For this demo I took an image of the cartwheel galaxy captured by the James Webb space telescope:
[James Webb Space Telescope images](https://www.newscientist.com/article/2452297-10-stunning-james-webb-space-telescope-images-show-the-beauty-of-space/) 

It's a very general setup and multiple ellipse model definitions, optimization criteria and demo images can be defined.

To efficiently fit the ellipses I need to do some image preprocessing, some filtering and some clustering. For these tasks I used `Clustering.jl`, `Images.jl` and `DSP.jl`. Would it be okay to use these dependencies? I can come up with alternative implementations but it will be more messy. 

After fitting the ellipses the final result looks like this:

afbeelding


odow:
This looks like a super cool (and visual) example.

No problem adding the extra packages. Just ensure they have a compat:
https://github.com/jump-dev/JuMP.jl/blob/488a0dbf0a0d34fcbc37344653edb7fdcf82794c/docs/Project.toml#L42

Also: we need to make sure that the image we use can be redistributed. If possible, find the original source, instead of new scientist?

langestefan:
Yeah JWST really creates some breathtaking images! 

> Also: we need to make sure that the image we use can be redistributed. If possible, find the original source, instead of new scientist?

Found the source: https://webbtelescope.org/contents/media/images/2022/039/01G8JXN0K2VBQP112RNSQWTCTH

> Unless otherwise stated, all material on the site was produced by NASA and the Space Telescope Science Institute (STScI). Material on this site produced by STScI was created, authored, and/or prepared for NASA under Contract NAS5-03127. Unless otherwise specifically stated, no claim to copyright is being asserted by STScI and material on this site may be freely used as in the [public domain](https://webbtelescope.org/copyright.html#h4-62a9d944-60a7-4120-914a-01cb6b72174b) in accordance with NASA's contract. However, it is requested that in any subsequent use of this work NASA and STScI be given appropriate acknowledgement.

I'll make sure we add the acknowledgment.

langestefan:
@odow I tested it locally, it should run. I have tried to follow the style guide based on other examples but might have missed something.

Maybe good to check if all the MOI calls are correct, and you will probably spot some improvements in `JuMP` syntax as well.

docs/Project.toml
```diff
@@ -31,13 +35,15 @@ ParametricOptInterface = "0ce4ce61-57bf-432b-a095-efac525d185e"
...
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 SQLite = "0aa819cd-b072-5ff4-a722-6bc24af294d9"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Wavelets = "29a6e085-ba6d-5f35-a997-948ac2efa89a"
 
 [compat]
```
odow, line 48: These need entries in the `[compat]` section

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
+# # Example: least-squares fitting of circles and ellipses
```
odow, line 1: This needs a license header

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+import SCS
+import LinearAlgebra
+import Images
+import Wavelets
+using LinearAlgebra
+using DSP
+using Clustering
+using Plots
+using LinearOperatorCollection
+using RegularizedLeastSquares
```
odow, line 21: Please use `import` for all packages other than JuMP.

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+# ```
+# The value of ``r_0`` is positive if the point is outside the ellipse, zero if it is on
+# the ellipse and negative if it is inside the ellipse. We also see we only need 6 
+# parameters to uniquely define an ellipse.
+
+# ## Helper functions
+
+# We define some helper functions to help us visualize the results.
+# <details>
+# <summary>Functions</summary>
```
odow, line 66: Remove these. Just be explicit.

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+function plot_dwt(x, sz = (500, 500))
+    return plt = heatmap(
```
odow, line 68:
```suggestion
    return Plots.heatmap(
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+# The value of ``r_0`` is positive if the point is outside the ellipse, zero if it is on
+# the ellipse and negative if it is inside the ellipse. We also see we only need 6 
+# parameters to uniquely define an ellipse.
+
+# ## Helper functions
+
+# We define some helper functions to help us visualize the results.
+# <details>
+# <summary>Functions</summary>
+function plot_dwt(x, sz = (500, 500))
```
odow, line 67:
```suggestion
function plot_dwt(x; sz = (500, 500))
```
The formatter is going to complain to you about a bunch of things

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+        ylims = (0, size(x, 1)),
+        size = sz,
+        dpi = 300,
+    )
+end
+
+function normalize(x)
+    x = (x .- minimum(x)) ./ (maximum(x) - minimum(x))
+    return x
+end
```
odow, line 83:
```julia
function normalize(x::AbstractVector)
    l, u = extrema(x)
    return (x .- l) ./ (u - l)
end
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+# ## Reading the test image
+
+# To test our ellipse-fitting algorithm we need a test image with elliptical features.
+# For our test image we will use an image of the cartwheel galaxy, 
+# [captured by the James Webb Space Telescope](https://webbtelescope.org/contents/media/images/2022/039/01G8JXN0K2VBQP112RNSQWTCTH).
+# Galaxies [come in many shapes and sizes](https://science.nasa.gov/universe/galaxies/types/), elliptical being one of them.
+
+# This is just a toy problem with little scientific value, but you can imagine how the 
+# rotation and position of elliptical galaxies can be useful information to astronomers.
+img = load("docs/src/assets/cartwheel_galaxy.png");
```
odow, line 95:
```suggestion
img = load("../../assets/cartwheel_galaxy.png");
```
This has to be relative to the current directory. It's something like:

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+# galaxies. We will use the [Sobel operator](https://en.wikipedia.org/wiki/Sobel_operator) for this task.
+# <details>
+# <summary>Sobel operator</summary>
+function edge_detector(
+    f_smooth::Matrix{Float64},
+    d1::Float64 = 0.1,
+    d2::Float64 = 0.1,
+)
+    rows, cols = size(f_smooth)
+
```
odow, line 178: Remove all blank lines within functions.

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+    gradient_x = conv(f_smooth, sobel_x)
+    gradient_y = conv(f_smooth, sobel_y)
+
+    gradient_magnitude = sqrt.(gradient_x .^ 2 + gradient_y .^ 2)
+
+    gradient_xx = conv(f_smooth, sobel_xx)
+    gradient_yy = conv(f_smooth, sobel_yy)
+    laplacian_magnitude = sqrt.(gradient_xx .^ 2 + gradient_yy .^ 2)
+
+    for i in 1:rows
```
odow, line 197:
```suggestion
    for i in 1:rows, j in 1:cols
```
You can drop one level of indentation with this.

Or just
```julia
    return (gradient_magnitude .> d1) .| (laplacian_magnitude .< d2) 
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+        levels = [0.0],
+        linewidth = 2,
+        color = :red,
+        cbar = false,
+    )
+end
+scatter!([0], [0]; color = :blue, label = "Squared (Objective 1)")
+scatter!([0], [0]; color = :red, label = "Min-Max (Objective 2)")
+display(plt)
+# </details>
```
odow, line 416:
```suggestion

```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+        Z_sq;
+        levels = [0.0],
+        linewidth = 2,
+        color = :red,
+        cbar = false,
+    )
+end
+scatter!([0], [0]; color = :blue, label = "Squared (Objective 1)")
+scatter!([0], [0]; color = :red, label = "Min-Max (Objective 2)")
+display(plt)
```
odow, line 415:
```suggestion
plt
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,416 @@
...
+solver = createLinearSolver(OptISTA, Φ; reg = reg, iterations = 15);
+
+# The sampled image in wavelet domain is given by:
+b = Φ * vec(x);
+
+# We can now solve the optimization problem to find the sparse representation of the image.
+x_approx = solve!(solver, b)
+x_approx = reshape(x_approx, size(x));
+x_final = normalize(x_approx)
+
```
odow, line 155:
```suggestion
```

Remove all blank lines in the middle of code blocks

docs/src/tutorials/conic/ellipse_fitting.jl
langestefan: Which license exactly do you mean?

blegat: You can copy-paste this:
https://github.com/jump-dev/JuMP.jl/blob/00c3b35d8f0d894e0ff70768b58ee200631cc6a2/docs/src/tutorials/conic/dualization.jl#L1-L5

langestefan: got it, thanks

langestefan:
All comments should be addressed now.

docs/Project.toml
```diff
@@ -71,3 +77,9 @@ SQLite = "1"
 SpecialFunctions = "2"
 StatsPlots = "0.15"
 Tables = "1"
+Wavelets = "0.10.0"
+RegularizedLeastSquares = "0.16.6"
+LinearOperatorCollection = "2.0.7"
+Images = "0.26.1"
+Clustering = "0.15.7"
+DSP = "0.7.10"
```
odow, line 85: Can we keep these in alphabetical order

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,383 @@
...
+using JuMP
+import SCS
+import Images
+import Wavelets
+import Clustering
+import LinearAlgebra
+import LinearOperatorCollection as LOC
+import RegularizedLeastSquares as RLS
+import Plots
+import DSP
```
odow, line 25: Ditto. Put `using JuMP` first, and then the `import` in alphabetical order.

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,383 @@
...
+# ## Reading the test image
+
+# To test our ellipse-fitting algorithm we need a test image with elliptical features.
+# For our test image we will use an image of the cartwheel galaxy, 
+# [captured by the James Webb Space Telescope](https://webbtelescope.org/contents/media/images/2022/039/01G8JXN0K2VBQP112RNSQWTCTH).
+# Galaxies [come in many shapes and sizes](https://science.nasa.gov/universe/galaxies/types/), elliptical being one of them.
+
+# This is just a toy problem with little scientific value, but you can imagine how the 
+# rotation and position of elliptical galaxies can be useful information to astronomers.
+img = load("../../assets/cartwheel_galaxy.png");
```
odow, line 96: This looks like it needs `Images.load`

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,383 @@
...
+
+# We convert the image to gray scale so that we can work with a single channel.
+
+img_gray = Images.Gray.(img)
+mosaicview(img, img_gray; nrow = 1)
```
odow, line 101: Ditto here

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,383 @@
...
+
+# The result of the clustering is a list of clusters to which we will assign a unique 
+# color. Each cluster is a list of points that belong to the same galaxy.
+clusters = result.clusters
+N_clusters = length(clusters)
+colors = Plots.distinguishable_colors(N_clusters + 1)[2:end]
+plt = plot_dwt(img_roi)
+for (i, cluster) in enumerate(clusters)
+    p_cluster = points[:, cluster.core_indices]
+    Plots.scatter!(
```
odow, line 217:
```suggestion
    Plots.scatter!(
        plt,
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,383 @@
...
+# \end{align}
+# ```
+# And hence can be modelled as a second-order cone program (SOCP) using 
+# [`MOI.RotatedSecondOrderCone`](@ref) as follows:
+
+ellipses_C1 = Vector{Dict{Symbol,Any}}()
+for (i, cluster) in enumerate(clusters)
+    p_cluster = points[:, cluster.core_indices]
+    Ξ = [(point[1], point[2]) for point in eachcol(p_cluster)]
+
```
odow, line 283:
```suggestion
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,383 @@
...
+    @variable(model, ζ >= 0)
+    @constraint(
+        model,
+        [1 / 2; ζ; model[:r]] in
+        MOI.RotatedSecondOrderCone(2 + length(model[:r]))
+    )
+    obj = @objective(model, Min, ζ)
+    set_objective_function(model, obj)
+    optimize!(model)
+
```
odow, line 294:
```suggestion
```

docs/Project.toml
```diff
@@ -55,19 +63,23 @@ ForwardDiff = "0.10"
...
 PATHSolver = "=1.7.8"
 ParametricOptInterface = "0.8.1"
 Plots = "1"
+RegularizedLeastSquares = "0.16.6"
 SCS = "=2.0.1"
 SQLite = "1"
 SpecialFunctions = "2"
 StatsPlots = "0.15"
 Tables = "1"
+Wavelets = "0.10.0"
```
odow, line 85:
```suggestion
Wavelets = "0.10.0"

```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,377 @@
...
+# ```math
+# \begin{align}
+#     \min_{Q, d, e, ζ} &ζ \\
+# \text{s.t.} \quad ζ \geq d^2_{\text{res}}(\xi_i, \mathcal{E}) &\quad \forall i \in N
+# \end{align}
+# ```
+# And hence can be modelled as a second-order cone program (SOCP) using 
+# [`MOI.RotatedSecondOrderCone`](@ref) as follows:
+
+ellipses_C1 = Vector{Dict{Symbol,Any}}()
```
odow, line 279:
```suggestion
ellipses_C1 = Dict{Symbol,Any}[]
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,377 @@
...
+# ```math	
+# \begin{equation}
+#     \min_{Q, d, e} \max_{\xi_i \in \mathcal{F}} d_\text{res}(\xi_i, \mathcal{E}) = 
+#     \min_{Q, d, e} ||d_\text{res}||_\infty
+# \end{equation}
+# ```
+
+# This objective can be implemented in JuMP using [`MOI.NormInfinityCone`](@ref) as 
+# follows:
+ellipses_C2 = Vector{Dict{Symbol,Any}}()
```
odow, line 343:
```suggestion
ellipses_C2 = Dict{Symbol,Any}[]
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,377 @@
...
+    model = ellipse(Ξ)
+    N = length(Ξ)
+    @variable(model, ζ)
+    @constraint(
+        model,
+        [ζ; model[:r]] in MOI.NormInfinityCone(1 + length(model[:r]))
+    )
+    obj = @objective(model, Min, ζ)
+    set_objective_function(model, obj)
+    optimize!(model)
```
odow, line 357:
```suggestion
    optimize!(model)
    @assert is_solved_and_feasible(model)
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,390 @@
...
+# features. For our test image we will use an image of the cartwheel galaxy,
+# [captured by the James Webb Space Telescope](https://webbtelescope.org/contents/media/images/2022/039/01G8JXN0K2VBQP112RNSQWTCTH).
+# Galaxies [come in many shapes and sizes](https://science.nasa.gov/universe/galaxies/types/),
+# elliptical being one of them.
+
+# This is just a toy problem with little scientific value, but you can imagine
+# how the rotation and position of elliptical galaxies can be useful information
+# to astronomers.
+
+img = Images.load("../../assets/cartwheel_galaxy.png");
```
odow, line 100:
```suggestion
filename = joinpath(@__DIR__, "..", "..", "assets", "cartwheel_galaxy.png")
img = Images.load(filename);
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,391 @@
...
+# will use a conic optimization approach to do so since it is a very natural way
+# to represent ellipses.
+
+# First, we define the residual distance definition (6) of a point to an ellipse
+# in JuMP:
+
+function create_ellipse_model(Ξ::Array{Tuple{Int,Int},1}, ϵ = 1e-5)
+    N = length(Ξ)
+    model = Model(SCS.Optimizer)
+    set_silent(model)
```
odow, line 265:
```suggestion
    # set_silent(model)
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,391 @@
...
+    model = create_ellipse_model(Ξ)
+    N = length(Ξ)
+    @variable(model, ζ)
+    @constraint(
+        model,
+        [ζ; model[:r]] in MOI.NormInfinityCone(1 + length(model[:r]))
+    )
+    @objective(model, Min, ζ)
+    optimize!(model)
+    @assert is_solved_and_feasible(model)
```
odow, line 372:
```suggestion
    @assert is_solved_and_feasible(model; allow_almost = true)
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,391 @@
...
+# Now that we have all the ingredients we can finally start fitting ellipses. We
+# will use a conic optimization approach to do so since it is a very natural way
+# to represent ellipses.
+
+# First, we define the residual distance definition (6) of a point to an ellipse
+# in JuMP:
+
+function create_ellipse_model(Ξ::Array{Tuple{Int,Int},1}, ϵ = 1e-5)
+    N = length(Ξ)
+    model = Model(SCS.Optimizer)
```
odow, line 264:
```suggestion
    model = Model(Clarabel.Optimizer)
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,391 @@
...
+# Ellipse fitting is a common task in data analysis and computer vision and is
+# of key importance in many application areas. In this tutorial we show how to
+# fit an ellipse to a set of points using a conic optimization approach.
+
+# ## Required packages
+
+# This tutorial uses the following packages:
+
+using JuMP
+import Clustering
```
odow, line 17:
```suggestion
import Clarabel
import Clustering
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,392 @@
...
+using JuMP
+import Clarabel
+import Clustering
+import DSP
+import Images
+import LinearAlgebra
+import LinearOperatorCollection as LOC
+import Plots
+import RegularizedLeastSquares as RLS
+import SCS
```
odow, line 25:
```suggestion
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,391 @@
...
+# will use a conic optimization approach to do so since it is a very natural way
+# to represent ellipses.
+
+# First, we define the residual distance definition (6) of a point to an ellipse
+# in JuMP:
+
+function create_ellipse_model(Ξ::Array{Tuple{Int,Int},1}, ϵ = 1e-5)
+    N = length(Ξ)
+    model = Model(Clarabel.Optimizer)
+    # set_silent(model)
```
odow, line 265:
```suggestion
    set_silent(model)
```

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,391 @@
...
+    model = create_ellipse_model(Ξ)
+    N = length(Ξ)
+    @variable(model, ζ)
+    @constraint(
+        model,
+        [ζ; model[:r]] in MOI.NormInfinityCone(1 + length(model[:r]))
+    )
+    @objective(model, Min, ζ)
+    optimize!(model)
+    @assert is_solved_and_feasible(model; allow_almost = true)
```
odow, line 372:
```suggestion
    @assert is_solved_and_feasible(model)
```

langestefan:
@odow Thanks for pushing it across the finishline! Seems like it's almost ready, let me know if I can do anything else.

Just curious, but why did the solver change to Clarabel?

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,391 @@
...
+    model = create_ellipse_model(Ξ)
+    N = length(Ξ)
+    @variable(model, ζ)
+    @constraint(
+        model,
+        [ζ; model[:r]] in MOI.NormInfinityCone(1 + length(model[:r]))
+    )
+    @objective(model, Min, ζ)
+    optimize!(model)
+    @assert is_solved_and_feasible(model)
```
odow, line 372:
```suggestion
    @assert is_solved_and_feasible(model; allow_almost = true)
```

odow:
I changed to Clarabel because SCS wasn't solving the final problem; it was hitting an iteration limit. (As an aside, you must **always** check the termination condition of a solver before accessing the solution.) But it seems like even Clarabel has trouble.

langestefan:
You can try changing epsilon from 1e-5. This is the numerical issue with the SPD matrix constraint I was talking about on slack.

odow:
What happens if we just drop the constraint? Then you might end up with an ellipse with zero area?

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,391 @@
...
+# in JuMP:
+
+function create_ellipse_model(Ξ::Array{Tuple{Int,Int},1}, ϵ = 1e-5)
+    N = length(Ξ)
+    model = Model(Clarabel.Optimizer)
+    set_silent(model)
+    @variable(model, Q[1:2, 1:2], PSD)
+    @variable(model, d[1:2])
+    @variable(model, e)
+    @constraint(model, Q >= ϵ * LinearAlgebra.I(2), PSDCone())
```
odow, line 269:
```suggestion
```

langestefan:
Mathematically that doesn't make sense: https://en.m.wikipedia.org/wiki/Matrix_representation_of_conic_sections

I believe I tried this and it just produced trivial solutions of the quadratic equation. 

I think what we should do is fix the scale to 1. This means we set the determinant of Q to 1.

There is probably a way to do that with MOI? Maybe MOI.logdetconetriangle or similar?

langestefan:
I just checked the results with the SPD constraint removed.

Using SCS:
afbeelding

Using Clarabel:
afbeelding

This is quite puzzling to me. I am not suprised SCS now fails, but that Clarabel succeeds I don't understand.

docs/src/tutorials/conic/ellipse_fitting.jl
```diff
@@ -0,0 +1,378 @@
...
+    laplacian_magnitude = sqrt.(gradient_xx .^ 2 + gradient_yy .^ 2)
+    return (gradient_magnitude .> d1) .& (laplacian_magnitude .< d2)
+end
+
+# We apply the Sobel operator to the binary image:
+
+edges = edge_detector(convert(Matrix{Float64}, x_bin), 1e-1, 1e2)
+edges = Images.thinning(edges; algo = Images.GuoAlgo())
+
+# And finally we cluster the edges using dbscan so we can fit ellipses to
```
odow, line 204:
```suggestion
# And finally we cluster the edges using `dbscan` so we can fit ellipses to
```

odow:
> This is quite puzzling to me. I am not suprised SCS now fails

It's not that SCS now fails. It didn't work before (it hit an iteration limit). It was just luck that the primal solution that SCS terminated with looked okay.

>  but that Clarabel succeeds I don't understand.

Different solvers different algorithms :smile:

odow:
Thanks! That took a while to get right, but we got there in the end :smile:

langestefan:
Yay! Thanks for finishing it. I learned a lot again.

PR #3899, opened 2024-12-04
[docs] call optimize! in solve_and_print_solution

lamont-granquist:
Just missing a line to do the thing.


odow (approved):
Thanks! The tests passed previously because we'd already solved the model in the previous cell :smile:

Issue #3912, opened 2025-01-15
Type error for primal feasibility report in non-Float64 precision


mtanneau:
```julia
using JuMP

model = JuMP.GenericModel{Float32}()
@variable(model, x)
# linear constraints are OK
@constraint(model, x == 1f0)
# quadratic constraints are OK too (it seems)
@constraint(model, x*x == 0f0)

JuMP.primal_feasibility_report(model, Dict(x => 0f0))  # no error so far

# general nonlinear constraints throw an error
# note: a similar error occurs 
@constraint(model, cos(x) == 1f0)
@constraint(model, sqrt(x) == 1f0)

JuMP.primal_feasibility_report(model, Dict(x => 0f0))  # 💥
```

the stacktrace below suggests that something got converted
```jl
ERROR: distance_to_set using the distance metric MathOptInterface.Utilities.ProjectionUpperBoundDistance() for set type MathOptInterface.EqualTo{Float32} has not been implemented yet.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] distance_to_set(d::MathOptInterface.Utilities.ProjectionUpperBoundDistance, ::Float64, set::MathOptInterface.EqualTo{Float32})
    @ MathOptInterface.Utilities ~/.julia/packages/MathOptInterface/PUD05/src/Utilities/distance_to_set.jl:75
  [3] distance_to_set(point::Float64, set::MathOptInterface.EqualTo{Float32})
    @ MathOptInterface.Utilities ~/.julia/packages/MathOptInterface/PUD05/src/Utilities/distance_to_set.jl:71
  [4] _distance_to_set(point::Float64, set::MathOptInterface.EqualTo{Float32}, ::Type)
    @ JuMP ~/.julia/packages/JuMP/CU7H5/src/feasibility_checker.jl:172
  [5] _add_infeasible_constraints(model::GenericModel{Float32}, ::Type{GenericNonlinearExpr{GenericVariableRef{Float32}}}, ::Type{MathOptInterface.EqualTo{Float32}}, violated_constraints::Dict{Any, Float32}, point_f::Function, atol::Float32)
    @ JuMP ~/.julia/packages/JuMP/CU7H5/src/feasibility_checker.jl:143
  [6] primal_feasibility_report(point::Function, model::GenericModel{Float32}; atol::Float32, skip_missing::Bool)
[...]

PR #3913, opened 2025-01-15
Fix primal_feasibility_report with non-Float64 number types

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3912


test/test_feasibility_checker.jl
```diff
@@ -197,4 +197,22 @@ function test_nonlinear_missing()
     )
 end
 
+function test_nonlinear_Float32()
+    model = GenericModel{Float32}()
+    @variable(model, x >= 2.0f0)
+    @constraint(model, c1, x == 1.0f0)
+    @constraint(model, c2, x*x == 0.0f0)
```
odow, line 204:
```suggestion
    @constraint(model, c2, x * x == 0.0f0)
```

PR #3914, opened 2025-01-15
[docs] use tectonic_jll

odow:
Trying out an alternative to the docker image.

A proper fix is probably to sort https://github.com/JuliaDocs/Documenter.jl/pull/2349


odow:
Saves a few minutes but not much

PR #3915, opened 2025-01-16
Support TIME_LIMIT in is_solved_and_feasible

odow:
Originally added in https://github.com/jump-dev/JuMP.jl/pull/3668, doesn't look like we discussed `TIME_LIMIT`.

But this came up last year in a discussion I had with Simon Bowly, where I thought we did support `TIME_LIMIT`.

And then @ccoffrin just asked me about it as well.

I tend to think that most people who set a time limit and get given a feasible solution would say that `is_solved_and_feasible` is `true`.

If we decide to add, there's an open question whether we should fold this into `allow_local`, or whether we add a new `allow_time_limit` kwarg.

### DIFF
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -808,6 +808,7 @@ end
         model::GenericModel;
         allow_local::Bool = true,
         allow_almost::Bool = false,
+        allow_time_limit::Bool = false,
         dual::Bool = false,
         result::Int = 1,
     )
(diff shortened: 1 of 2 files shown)


odow:
TIME_LIMIT seems like the one most commonly set by users who expect a solution. ITERATION_LIMIT can happen in interior point-type solvers and tends to indicate failure, not success. NODE_LIMIT and OBJECTIVE_LIMIT are up for debate.

> In the use-case I encountered, what I needed is an is_feasible or has_solution function, which just checks for primal feasibility. Do we already have a shorthand for that type of test?

Just define:
```julia
has_values(model) = primal_status(model) == FEASIBLE_POINT
```
We have
https://github.com/jump-dev/JuMP.jl/blob/99cdc4cd56df7ba60f56f66624a4c8cacb0284de/src/variables.jl#L1916-L1951
but it's not quite the same thing.

odow:
Call conclusions were somewhat inconclusive.

@ccoffrin suggested that he really needed only `primal_status(model) == FEASIBLE_POINT`, which didn't seem to justify a new function.

We also see value in "teaching" users that they need to analyze the statuses in more detail. `is_solved_and_feasible` is okay for scripts and notebooks, but not really for library code or some sort of meta algorithm.

This PR cannot be merged as-is.

An alternative is:
```julia
is_solved_and_feasible(model; allow_time_limit::Bool = false)
```
but this doesn't help cases like with other limits like `SOLUTION_LIMIT`.

A second alternative is:
```julia
is_solved_and_feasible(model; allowed_termination_status = [OPTIMAL, LOCALLY_INFEASIBLE])
is_solved_and_feasible(model; allowed_termination_status = [OPTIMAL, LOCALLY_INFEASIBLE, TIME_LIMIT])
```
but this requires the user to know the various statuses, and it's a bit cumbersome. And something like this is painful:
```Julia
is_solved_and_feasible(model; additionally_allowed_termination_status = [TIME_LIMIT])
```
One option is:
```Julia
if is_solved_and_feasible(model) ||
   (termination_status(model) == TIME_LIMIT && primal_status(model) == FEASIBLE_POINT)
    # ...
end
```

@blegat also sugggested `is_solved(model)` and `is_feasible(model)` to split the logic, which would enable:
```julia
if (is_solved(model) || termination_status(model) == TIME_LIMIT) && is_feasible(model)
    # ...
end
```

But this is just adding more complexity when at this point the user really needs to write proper logic for handling the statuses. Yes, this means duplicated code, but that's not the end of the world because it makes things very clear about what is allowed etc.

We have https://jump.dev/JuMP.jl/stable/manual/solutions/#Recommended-workflow, which I think stands up pretty well.

odow:
Perhaps this PR can just be closed. #3925 tweaks the docstring of `is_solved_and_feasible`.

src/optimizer_interface.jl
```diff
@@ -808,29 +808,52 @@ end
         model::GenericModel;
         allow_local::Bool = true,
         allow_almost::Bool = false,
+        allow_time_limit::Bool = false,
```
joaquimg, line 811: I would prefer a list of accepted status instead of more keywords

joaquimg: `allow_status = ()`, and I would suggest a tuple since it is non-allocating. But accepting any collection that work with `in` should be fine

src/optimizer_interface.jl
```diff
@@ -858,14 +884,16 @@ function is_solved_and_feasible(
     dual::Bool = false,
     allow_local::Bool = true,
     allow_almost::Bool = false,
+    additional_termination_statuses::Vector{TerminationStatusCode} = TerminationStatusCode[],
```
joaquimg, line 887: I do not see a reason to force a Vector, we should at least allow tuples.

odow: I really want to minimise the number of input types we support. This is part of the MethodError principle. Is there a good argument not to restrict to `Vector`? The allocations here are minimal, and checking this status should be so minor in comparison to solving the problem.

odow, line 887:
```suggestion
    additional_termination_statuses::Vector{MOI.TerminationStatusCode} = MOI.TerminationStatusCode[],
```

joaquimg:
```suggestion
    additional_termination_statuses::Union{
        Vector{TerminationStatusCode},
        NTuple{N,TerminationStatusCode}} = (),
```

joaquimg: The Method error should be very clear with this suggestion. No weir tuples would be allowed.
Allocations are minimal, but there is no reason for them. Minimal allocations in a loop become big allocations.

odow: Who's calling this in a loop? And if they are, they cache the vector once outside the loop.

odow:
I'm also somewhat in favor of just closing this PR and leaving the behavior as-is for now.

odow:
Closing for now. I'll make a mental note to re-open this if the question comes up again in the future.

And if you, reader from the future, are interested in this feature, please comment below and I will re-open.

PR #3916, opened 2025-01-16
Update variables.md explanation of parameters

artkuo:
I was confused by this example at multiple levels. I didn't understand what "Parameters set" means, is there a whole set of parameters somewhere? I didn't understand `Parameter(i)` because `i` looks like an index, not a value. I didn't get how there can be `x in SecondOrderCone` and `x in Parameter(i)`, same syntax but seemingly different concepts. The values are eventually revealed later on in example, but even then I was confused.

This change is to try to be more explicit about what's going on, in case others are as obtuse as I am. I hope something like this will be clearer.

### DIFF
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -1318,24 +1318,19 @@ requires `MOI.add_constrained_variables`.
 Some solvers have explicit support for parameters, which are constants in the
 model that can be efficiently updated between solves.
 
-JuMP implements parameters by a decision variable constrained on creation to a set consisting of a single value, of type 
-[`Parameter`](@ref). For example, the following creates two variables `p[i]` with parameter values 1 and 2, respectively:
+JuMP implements parameters by a decision variable constrained on creation to a
+value of the [`Parameter`](@ref) set. For example, the following creates two
+parameters, `p[1]` and `p[2]`, with parameter values `2.0` and `4.0`:
 
 ```jldoctest nonlinear_parameters
 julia> model = Model();
 
 julia> @variable(model, x);
 
-julia> @variable(model, p[i = 1:2] in Parameter(i))
+julia> @variable(model, p[i = 1:2] in Parameter(2.0 * i))
 2-element Vector{VariableRef}:
  p[1]
  p[2]
-```
-
-Create anonymous parameters using the `set` keyword:
-```jldoctest nonlinear_parameters
-julia> anon_parameter = @variable(model, set = Parameter(1.0))
-_[4]
 ```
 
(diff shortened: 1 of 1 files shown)


artkuo:
The build failed because it failed a test on `builders_decomposition.jl`. I don't think it's related to this pull request, which is just changing some non-executing text in a markdown file.

odow:
Hi @artkuo, thanks for opening the PR! 

Don't worry about the failure. It is my fault.

To answer some of your questions;

> is there a whole set of parameters somewhere

JuMP uses "set" to mean the domains that constraint functions belong to. 

If you have time, you might want to read https://arxiv.org/abs/2002.03447. Or you can take a look at https://jump.dev/JuMP.jl/stable/moi/manual/standard_form/

So `SecondOrderCone()` is the set https://jump.dev/JuMP.jl/stable/api/JuMP/#SecondOrderCone.

`Parameter(value)` is the singleton mathematical set `{value}`. See

https://jump.dev/JuMP.jl/stable/api/JuMP/#Parameter

which is really just a synonym for:

https://jump.dev/JuMP.jl/stable/moi/reference/standard_form/#MathOptInterface.Parameter

> because i looks like an index, not a value

It's a value :smile:

odow:
I made a change. Is that better?

artkuo:
Yes, that looks great! Sorry about my questions, they were meant to be rhetorical but were eventually resolved. I should have put that under "This change is..." I think your revision is very nice, thanks. The `2.0*i` removes potential confusion between index and value.

artkuo:
I'm was confused, I clicked "close with comment" thinking your revision would be merged. But it looks like that just killed the pull request. I'm leaving open in case you need to see that your revision still needs to be merged. Please close this if and when appropriate.

odow:
> I'm was confused, I clicked "close with comment" thinking your revision would be merged.

I have the power to edit your PR directly :smile: When I merge it, it'll attribute you as the author.

docs/src/manual/variables.md
```diff
@@ -1318,40 +1318,35 @@ requires `MOI.add_constrained_variables`.
...
+value of the [`Parameter`](@ref) set. For example, the following creates two
+parameters, `p[1]` and `p[2]`, with parameter values `2.0` and `4.0`:
 
 ```jldoctest nonlinear_parameters
 julia> model = Model();
 
 julia> @variable(model, x);
 
-julia> @variable(model, p[i = 1:2] in Parameter(i))
+julia> @variable(model, p[i = 1:2] in Parameter(2.0 * i))
```
odow, line 1330:
```suggestion
julia> @variable(model, p[i in 1:2] in Parameter(2.0 * i))
```

PR #3925, opened 2025-01-25
Add assert_is_solved_and_feasible

### LINKED ISSUE #3923: assert_is_solved_and_feasible
We could have a function that checks is_solved_and_feasible and if false, throws a nice error message explaining what happened.

```julia
function assert_is_solved_and_feasible(model; kwargs...)
    if !is_solved_and_feasible(model; kwargs...)
        error(string(solution_summary(model)))
    end
end
```

```julia
if primal_status(model) == FEASIBLE_POINT
    # continue
end
```

odow:
Closes #3923 

Bikeshedding: the alternative is `is_solved_and_feasible(model; assert = true)`


src/optimizer_interface.jl
```diff
@@ -871,6 +881,58 @@ function is_solved_and_feasible(
...
+  Dual status        : NO_SOLUTION
+
+* Work counters
+
+Stacktrace:
+[...]
+```
+"""
+function assert_is_solved_and_feasible(model::GenericModel; kwargs...)
+    if is_solved_and_feasible(model; kwargs...)
```
odow, line 926:
```suggestion
    if !is_solved_and_feasible(model; kwargs...)
```

src/optimizer_interface.jl
```diff
@@ -871,6 +881,58 @@ function is_solved_and_feasible(
...
+
+Stacktrace:
+[...]
+```
+"""
+function assert_is_solved_and_feasible(model::GenericModel; kwargs...)
+    if !is_solved_and_feasible(model; kwargs...)
+        error(
+            "The model was not solved correctly. Here is a summary of the " *
+            "solution to help you debug why this happened:\n\n" *
```
mlubin, line 929:
```suggestion
            "solution to help debug why this happened:\n\n" *
```

mlubin:
I like it.

odow:
It's too late now, but for JuMP 2.0, a better solution might be for this to be the default behavior of `optimize!(model)`, and you have to explicitly pass something `optimize!(model; skip_status_check = true)` which is a sign that you will then explicitly check the problem statuses before querying the solution.

Right now we have the behavior that almost everyone should use
```julia
optimize!(model)
assert_is_solved_and_feasible(model)
```
and some people should do
```julia
optimize!(model)
if termination_status(model) == OPTIMAL
    # ...
else if termination_status(model) == TIME_LIMIT
    # ...
else
    # ...
end
```

blegat (approved):

odow:
Thoughts on new function instead of a keyword?

blegat:
This is something to be used in quick scripts so keystrokes matter. `assert_is_solved_and_feasible` will be easier to type, auto-complete etc... It also makes sense as another function since the behavior is quite different, I don't see it just as finetuning of `is_solved_and_feasible`

odow:
Okay. This is good to go by me then.

joaquimg (approved):

Issue #3951, opened 2025-02-26
Constraint associated with a variable set


odow:
Reported by @lopenguin at https://discourse.julialang.org/t/dual-to-psd-constraint-in-jump/126352

If I have
```julia
using JuMP
model = Model()
@variable(model, A[1:2, 1:2] >= 0, PSD)
@variable(model, B[1:2, 1:2] in HermitianPSDCone())
@variable(model, C[1:3] in SecondOrderCone())
```
it's not at all trivial to get the constraint reference so I can query the dual. 

One suggested workaround might be:
```Julia
using JuMP
model = Model()
@variable(model, A[1:2, 1:2] >= 0, Symmetric)
@constraint(model, A_ref, A in PSDCone())
@variable(model, B[1:2, 1:2], Hermitian)
@constraint(model, B_ref, B in HermitianPSDCone())
@variable(model, C[1:3])
@constraint(model, C_ref, C in SecondOrderCone())
```

For scalar variables, we have:
```Julia
model = Model()
@variable(model, x >= 0, Int)
LowerBoundRef(x)
IntegerRef(x)
```

So perhaps we need something similar.

odow:
The underlying issue is that we aren't doing anything with the `ConstraintIndex` returned here:

https://github.com/jump-dev/JuMP.jl/blob/bea72a0278c7acdb2f7990bc0603466cfd43b6c4/src/variables.jl#L1971
https://github.com/jump-dev/JuMP.jl/blob/bea72a0278c7acdb2f7990bc0603466cfd43b6c4/src/variables.jl#L1989
https://github.com/jump-dev/JuMP.jl/blob/bea72a0278c7acdb2f7990bc0603466cfd43b6c4/src/variables.jl#L2165
https://github.com/jump-dev/JuMP.jl/blob/bea72a0278c7acdb2f7990bc0603466cfd43b6c4/src/variables.jl#L2248

odow:
ConstrainedVariableRef
ConstrainedInSetRef
InSetConstraintRef
SetRef
VariableSetRef
VariableInSetRef **
ConstraintAtCreationRef
ConstrainedOnCreationRef
CreationConstraintRef

odow:
We discussed this on today's call, and agreed that we need a dictionary in model to map between variables and their corresponding set. 

We did not agree on a name for the ref. Conclusion is that I should prototype the PR with documentation etc, and we can decide then.

PR #3955, opened 2025-02-28
Add VariableInSetRef and has_variable_in_set

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3951

### DIFF
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -1308,6 +1308,40 @@ julia> x = @variable(model, [1:3], set = SecondOrderCone())
     You cannot delete the constraint associated with a variable constrained on
     creation.
 
+To check if a variable was constrained on creation, use [`has_variable_in_set`](@ref),
+and use [`VariableInSetRef`](@ref) to obtain the associated constraint reference:
+
+```jldoctest
+julia> model = Model();
+
+julia> @variable(model, x[1:2, 1:2], PSD)
+2×2 LinearAlgebra.Symmetric{VariableRef, Matrix{VariableRef}}:
+ x[1,1]  x[1,2]
+ x[1,2]  x[2,2]
+
+julia> has_variable_in_set(x)
+true
+
+julia> c = VariableInSetRef(x)
+[x[1,1]  x[1,2]
+ ⋯       x[2,2]] ∈ PSDCone()
+
+julia> @variable(model, y)
+y
+
+julia> has_variable_in_set(y)
+false
+
+julia> @variable(model, z in Semicontinuous(1, 2))
+z
+
+julia> has_variable_in_set(z)
+true
+
+julia> c_z = VariableInSetRef(z)
+z ∈ MathOptInterface.Semicontinuous{Int64}(1, 2)
+```
+
 ### Example: positive semidefinite variables
 
 An alternative to the syntax in [Semidefinite variables](@ref), declare a matrix
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -1752,6 +1752,188 @@ function parameter_value(x::GenericVariableRef)
     return set.value
 end
 
+# VariableInSetRef
+
+"""
+    has_variable_in_set(
+        model::GenericModel,
+        x::Union{AbstractJuMPScalar,AbstractArray{<:AbstractJuMPScalar}},
+    )::Bool
+
+Return a `Bool` if [`VariableInSetRef`](@ref) returns a valid constraint
+reference without erroring.
+
+## Exceptions
+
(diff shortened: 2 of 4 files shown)


docs/src/manual/variables.md
```diff
@@ -1308,6 +1308,40 @@ julia> x = @variable(model, [1:3], set = SecondOrderCone())
     You cannot delete the constraint associated with a variable constrained on
     creation.
 
+To check if a variable was constrained on creation, use [`has_variable_in_set`](@ref),
```
blegat, line 1311:
```suggestion
To check if a variable was constrained on creation, use [`is_variable_in_set`](@ref),
```

src/variables.jl
```diff
@@ -1752,6 +1752,188 @@ function parameter_value(x::GenericVariableRef)
...
+
+julia> c_z = VariableInSetRef(z)
+z ∈ MathOptInterface.Semicontinuous{Int64}(1, 2)
+```
+"""
+function VariableInSetRef(x::AbstractJuMPScalar)
+    model = owner_model(x)
+    ci = get(model.variable_in_set_ref, x, nothing)
+    if ci === nothing
+        error("VariableInSetRef does not exist for $x")
```
blegat, line 1923:
```suggestion
        error("`VariableInSetRef` does not exist for `$x`")
```

src/variables.jl
```diff
@@ -2215,16 +2399,21 @@ function add_variable(
...
         T,
     )
-    var_refs =
-        [GenericVariableRef{T}(model, var_index) for var_index in var_indices]
-    return reshape_vector(var_refs, variable.shape)
+    ret = reshape_vector(GenericVariableRef{T}.(model, x), variable.shape)
+    if ci !== nothing  # ci === nothing if variable.set isa MOI.Reals
+        model.variable_in_set_ref[ret] = ci
+        if variable.shape != ScalarShape() && variable.shape != VectorShape()
+            model.shapes[ci] = variable.shape
```
blegat, line 2413: It's nice that the shape can work as well

docs/src/manual/variables.md
odow: Is this a general request to change the name of the function? We have `is_integer` and `is_fixed` but `has_lower_bound` and `has_upper_bound`.

odow:
So thoughts on the naming of these?

docs/src/manual/variables.md
blegat: A variable has a lower bound but you cannot say that a variable has a variable in set. I know you mean "has the variable an in set constraints" but that's not how it reads.

odow: Changed to `is`

odow:
Nightly is failing because it seems to have lost the ability to infer
https://github.com/jump-dev/MutableArithmetics.jl/blob/cc1be0edc797c3dff5fadbd33371b6c3aec0e158/src/reduce.jl#L42-L52

```julia
function fused_map_reduce(op::F, args::Vararg{Any,N}) where {F<:Function,N}
    _check_same_length(args...)
    T = promote_map_reduce(op, _concrete_eltype.(args)...)
    accumulator = neutral_element(reduce_op(op), T)
    buffer = buffer_for(op, T, eltype.(args)...)
    for I in zip(eachindex.(args)...)
        accumulator =
            buffered_operate!!(buffer, op, accumulator, getindex.(args, I)...)
    end
    return accumulator
end
```

which is kinda-not-surprising. It's pretty weird code that relies on splatting and types and stuff.

blegat (approved):

joaquimg:
I mostly like this PR very much.
Name seems fine.

My only concern is the little inconsistency on the macro variable parameters.

All(?) explicit parameters like Bin (binary=true), Int (integer=trey), Bounds (lower...)
Have their own BinaryRef, IntegerRef, LowerBoundRef...

Except for PSD, although we do not have (posdef=true).
As a user I might expect PositiveSemidefiniteSetRef.

Maybe the solution for this is to acknowledge that the PSD key is legacy and remove is almost everywhere in the docs, and only keep it is the variable macro doc with a strong recommendation for the "set =" syntax.

blegat:
Yes, the PSD is mostly legacy and `in PSDCone()` should be used

odow:
The reason we don't have `PSDRef` is that it is not trivially inferable from the variable index in the way that scalar variable constraints are just the index of the variable.

I don't have strong opinions on `PSD`. I agree that it's a magic symbol though that doesn't actually exist which makes it hard for new users to reason about. I'll draft a PR that removes PSD from the docs and we can decide what to do.

joaquimg (approved):

Issue #3948, opened 2025-02-23
Add option to print the solvers in memory representation


joaquimg:
Writing?
```julia
model = Model(HiGHS.Optimizer)
@variable(model, p1, set = Parameter(1.0))
@variable(model, p2 == 1.0)
write_to_file(model, "model.mps")
optimize!(model)
highs = unsafe_backend(model).inner
tmp_filename = "model.mps"
HiGHS.Highs_writeModel(highs, tmp_filename)
```
is hard for non-super-heavy users.

This is helpful for people with minimal understanding of solution methods.

odow:
> is hard for non-super-heavy users.

Should these users be printing out the internal model? I wouldn't really be in favor of them doing that.

This is also quite solver-dependent. Maybe it just needs better documentation in each solver's README.

Having said that, perhaps we could make `MOI.write_to_file(::HiGHS.Optimizer, filename::String)` work, and have `write_to_file(model, "model.mps"; use_internal_solver_method = true)`

odow:
I looked into this. The current isn't too bad:
```Julia
using JuMP, HiGHS
model = Model(HiGHS.Optimizer)
@variable(model, x >= 1)
optimize!(model)
Highs_writeModel(unsafe_backend(model), "model.mps")
```

we could perhaps make it
```Julia
using JuMP, HiGHS
model = Model(HiGHS.Optimizer)
@variable(model, x >= 1)
optimize!(model)
MOI.write_to_file(unsafe_backend(model), "model.mps")
```

I'm not really sure who this method is aimed at. Someone who wants to deeply introspect the solver's model, knowing that there is a bug somewhere between JuMP and the solver, so `JuMP.write_to_file(model)` doesn't work, but also has no understanding of `Highs_writeModel`?

joaquimg:
I like:
`MOI.write_to_file(unsafe_backend(model), "model.mps")`
Documenting it should solve the issue.
Some places where is might be useful:
1 - debug tutorial
2 - transitioning tutorial
3 - reference in the `write_to_file` docstring
4 - the solver readme

odow:
Came up https://discourse.julialang.org/t/using-second-order-cone-constraints-with-logarithms-in-jump-and-gurobis-c-api/125769/7?u=odow

odow:
I think there's an argument for (3) and (4). Strong disagreement about (2), and moderate disagreement about (1).

[...]

PR #3956, opened 2025-03-02
[docs] improve docstrings of write_to_file and read_from_file

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3948

### DIFF
--- a/src/file_formats.jl
+++ b/src/file_formats.jl
@@ -90,12 +90,12 @@ solver's internal API to write a model to disk.
 For MPS files in particular, [`write_to_file`](@ref) may not support the full
 range of features that a solver's internal API supports. This is because some
 solvers have defined solver-specific extensions to the MPS format, whereas our
-Julia implementation supports only features where are standardized across
+Julia implementation supports only features which are standardized across
 multiple solvers.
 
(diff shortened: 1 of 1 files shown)


src/file_formats.jl
```diff
@@ -38,10 +38,89 @@ end
...
+## Solver-specific formats
+
+[`write_to_file`](@ref) calls a Julia function from the MathOptInterface package
+that is independent of the choice of solver. That is, it does not call a
+solver's internal API to write a model to disk.
+
+For MPS files in particular, [`write_to_file`](@ref) may not support the full
+range of features that a solver's internal API supports. This is because some
+solvers have defined solver-specific extensions to the MPS format, whereas our
+Julia implementation supports only features where are standardized across
```
odow, line 93:
```suggestion
Julia implementation supports only features which are standardized across
```

src/file_formats.jl
```diff
@@ -38,10 +38,89 @@ end
...
+multiple solvers.
+
+To write a file to disk using the solver's internal API, use
+[`direct_model`](@ref) and
```
odow, line 97:
```suggestion
[`direct_model`](@ref) and call the solver's C API. For example:
```

PR #3957, opened 2025-03-08
[docs] fix: Add correct parameters for set_attribute in installation.md

Prasanthi-Peram:
Reference : [https://github.com/jump-dev/JuMP.jl/blob/master/src/optimizer_interface.jl](url)

From the implementation of the  
`set_optimizer_attribute(
        model::Union{GenericModel,MOI.OptimizerWithAttributes},
        attr::Union{AbstractString,MOI.AbstractOptimizerAttribute},
        value)` method; 

The function expects three parameters however [Installation Guide](https://jump.dev/JuMP.jl/stable/installation/#Install-a-solver) uses a (model,key value pair) in the solvers' section which throws a Method Error at runtime. This fix would solve the issue.

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -78,9 +78,9 @@ julia> using HiGHS
 
 julia> model = Model(HiGHS.Optimizer);
 
-julia> set_attribute(model, "output_flag" => false)
+julia> set_attribute(model, "output_flag",false)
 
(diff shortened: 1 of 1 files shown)


odow (approved):
Ooops! That's an embarrassing mistake on my part.

Thanks for the PR. I've tweaked it slightly to make the code block a `jldoctest`. This will run the code when we build the documentation to check that it runs without error.

odow:
Thanks!

PR #3959, opened 2025-03-11
Softly deprecate PSD by removing it from the docs

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3955#issuecomment-2715508282

Thoughts?

### DIFF
--- a/docs/src/manual/variables.md
+++ b/docs/src/manual/variables.md
@@ -1082,13 +1082,13 @@ julia> model[:x]
 
 ## Semidefinite variables
 
-Declare a square matrix of JuMP variables to be positive semidefinite by passing
-`PSD` as a positional argument:
+Declare a square matrix of JuMP variables to be positive semidefinite using
+[`PSDCone`](@ref):
 
 ```jldoctest
 julia> model = Model();
 
-julia> @variable(model, x[1:2, 1:2], PSD)
+julia> @variable(model, x[1:2, 1:2] in PSDCone())
 2×2 LinearAlgebra.Symmetric{VariableRef, Matrix{VariableRef}}:
  x[1,1]  x[1,2]
  x[1,2]  x[2,2]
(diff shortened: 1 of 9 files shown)


docs/src/tutorials/conic/simple_examples.jl
```diff
@@ -158,7 +158,7 @@ function example_k_means_clustering()
     end
     model = Model(Clarabel.Optimizer)
     set_silent(model)
-    @variable(model, Z[1:m, 1:m] >= 0, PSD)
+    @variable(model, Z[1:m, 1:m] >= 0 in PSDCone())
```
odow, line 161:
```suggestion
    @variable(model, Z[1:m, 1:m] >= 0, set = PSDCone())
```

docs/src/tutorials/conic/simple_examples.jl
```diff
@@ -158,7 +158,7 @@ function example_k_means_clustering()
     end
     model = Model(Clarabel.Optimizer)
     set_silent(model)
-    @variable(model, Z[1:m, 1:m] >= 0, PSD)
+    @variable(model, Z[1:m, 1:m] >= 0, set = PSDCone())
```
odow, line 161: This one is a reason not to stop using `PSD`. It's clunky, and I can see people getting tripped up over it.

odow:
I'm not convinced that this is a good idea.

docs/src/tutorials/conic/simple_examples.jl
joaquimg: I like it with the "set", the PSD one does not look better to me

joaquimg: Would you think its clearer if we replace ">=0" by "lower_bound=0".
I would not mind.

joaquimg (approved):

joaquimg:
I would not mind adding a note to the variable macro docstring asking to use "set =..."

docs/src/tutorials/conic/simple_examples.jl
odow: No I want to keep this as it was... We have nice syntax. There's no need to use `lower_bound=0` like we're a Python library.

odow: Was the main reason to stop using PSD because `VariableInSet` is confusing?

joaquimg: I dont agree that
x >= 0, PSD
Is better than
x >= 0, set = ... 

We should decide if its legacy or not.
For me it felt like it was.
For @blegat the same.

We can keep it, but if we do, we should have PSDRef.

odow: > We can keep it, but if we do, we should have PSDRef.

We have VariableInSetRef though?

What about https://github.com/jump-dev/JuMP.jl/pull/3960

odow: I'm missing `x >= 0, PSD)`. I vote we leave as-is and wait for more people to ask how to get the dual of a variable constraint.

mlubin:
It's kinda weird to be completely removing syntax from the docs. What if someone is trying to read existing code?

odow:
> What if someone is trying to read existing code?

I think this is especially true because it is magic syntax in the macro. You can't look it up as `? PSD` or something like that.

odow:
Closing because I don't particularly want to do this. Let's see how `VariableInSetRef` beds in for a while.

joaquimg:
My suggestion was to remove PSD everywhere, except from the `@variable` docstring. So it is findable.
Moreover, in that docstring it would say that the syntax is not recommended and users should use "set = ..."

blegat:
We could at least use `in PSDCone()` most of the time in the docs as it's the preferred way.

docs/src/manual/variables.md
```diff
@@ -1082,13 +1082,13 @@ julia> model[:x]
 
 ## Semidefinite variables
 
-Declare a square matrix of JuMP variables to be positive semidefinite by passing
```
blegat, line 1085: We could just say here that there are the two possibilities at least, we should reopen and merge part of the changes of this PR.

PR #3967, opened 2025-03-30
Add set_name(model::GenericModel, name::AbstractString)

odow:
x-ref https://discourse.julialang.org/t/can-i-query-is-continuous-about-a-jump-model/127484/5?u=odow

### DIFF
--- a/src/print.jl
+++ b/src/print.jl
@@ -205,9 +205,32 @@ name(model::AbstractModel) = "An Abstract JuMP Model"
 function name(model::GenericModel)
     if MOI.supports(backend(model), MOI.Name())
         ret = MOI.get(model, MOI.Name())
-        return isempty(ret) ? "A JuMP Model" : ret
+        if !isempty(ret)
+            return ret
+        end
     end
     return "A JuMP Model"
+end
+
+"""
+    set_name(model::GenericModel, name::AbstractString)
+
+Set the [`MOI.Name`](@ref) attribute of `model`'s [`backend`](@ref) to `name`.
+
+## Example
+
+```jldoctest
+julia> model = Model();
+
+julia> set_name(model, "My Model")
+
+julia> name(model)
+"My Model"
+```
+"""
+function set_name(model::GenericModel, name::AbstractString)
+    MOI.set(model, MOI.Name(), name)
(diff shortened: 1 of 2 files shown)


WalterMadelim:
@odow 
I think `name(model)` can be added to `solution_summary(model)`?
Existing:
```julia
solution_summary(; result = 1, verbose = false)
├ solver_name          : Gurobi
├ Termination
```
Revamped:
```julia
solution_summary(; result = 1, verbose = false)
├ My_Model was solved by Gurobi
├ Termination
```
where `My_Model::String = name(model)`

odow:
The name is printed as part of `show(::IO, ::AbstractModel)`.

I don't think we need to print the model name because in most cases it will be the default. The model name is also unrelated to the solution.

If you're solving multiple models in various orders and you're printing the solution summary to some sort of log file, then you should add some additional information to identify the models. 

For example: 
```julia
julia> using JuMP, HiGHS

julia> begin
           set_silent(model)
           set_name(model, "My Model")
           @variable(model, x >= 0)
           optimize!(model)
           contents = sprint() do io
               show(io, model)
               println(io)
               show(io, solution_summary(model))
           end;
           print(contents)
       end
My Model
├ solver: HiGHS
├ num_variables: 1
├ num_constraints: 1
│ └ VariableRef in MOI.GreaterThan{Float64}: 1
└ Names registered in the model
  └ :x
solution_summary(; result = 1, verbose = false)
├ solver_name          : HiGHS
├ Termination
│ ├ termination_status : OPTIMAL
│ ├ result_count       : 1
│ ├ raw_status         : kHighsModelStatusOptimal
│ └ objective_bound    : 0.00000e+00
├ Solution (result = 1)
│ ├ primal_status        : FEASIBLE_POINT
│ ├ dual_status          : FEASIBLE_POINT
│ ├ objective_value      : 0.00000e+00
│ ├ dual_objective_value : 0.00000e+00
│ └ relative_gap         : 0.00000e+00
└ Work counters
  ├ solve_time (sec)   : 2.51860e-05
  ├ simplex_iterations : 0
  ├ barrier_iterations : 0
  └ node_count         : -1
```

PR #3970, opened 2025-04-01
(slow) operations tracker

joaquimg:
This adds the optional ability to mark all places where slow operations that allocate expression occur. Keeping the stack might be a lot of data, but it is a way to properly identify the issues instead of erroring one by one.

Possible additional functionality:
1 - also mark the operation type. A bit redundant with the stack, but might lead to summary reports.
2 - create a user friendly summary

TODO

- [ ] add tests
- [ ] add docs session, to be referenced in the performance issues.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -126,6 +126,8 @@ mutable struct GenericModel{T<:Real} <: AbstractModel
     set_string_names_on_creation::Bool
     #
     variable_in_set_ref::Dict{Any,MOI.ConstraintIndex}
+    # A cache for tracking ineficient use of operations.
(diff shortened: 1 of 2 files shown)


odow:
I think we decided not to do this. 

We have some better support with `set_macro_timing `, but this tutorial could still be improved:

https://jump.dev/JuMP.jl/dev/tutorials/getting_started/debugging/#Debugging-performance-problems

joaquimg:
I am not sure we should close yet.
I still think macros and this are very different things.
I am thinking about counting allocations here, it might be useful to rank operation types.

odow:
I don't know what the right approach is, but after thinking about this more I don't think that it's this PR.

odow:
Closing because I don't want to do this.

Issue #3973, opened 2025-04-04
Error when adding a Nonpositive constraint in a variable bounded above


joaquimg:
This seems to be a bridging issue, it is trying to interpret the constraint as a bounded variable or bound.

MWE:
```julia
model = Model(HiGHS.Optimizer)
set_silent(model)
@variable(model, x <= 1)
@constraint(model, c6, [x] in Nonpositives())
optimize!(model)
```
the error
```julia
ERROR: MathOptInterface.UpperBoundAlreadySet{MathOptInterface.LessThan{Float64}, MathOptInterface.LessThan{Float64}}: Cannot add `VariableIndex`-in-`MathOptInterface.LessThan{Float64}` constraint for variable MOI.VariableIndex(1) as a `VariableIndex`-in-`MathOptInterface.LessThan{Float64}` constraint was already set for this variable and both constraints set an upper bound.
Stacktrace:
  [1] _throw_if_existing_upper
    @ HiGHS ~/.julia/packages/HiGHS/6NuM6/src/MOI_wrapper.jl:1450 [inlined]
  [2] _update_info
    @ HiGHS ~/.julia/packages/HiGHS/6NuM6/src/MOI_wrapper.jl:103 [inlined]
  [3] add_constraint(model::HiGHS.Optimizer, f::MathOptInterface.VariableIndex, s::MathOptInterface.LessThan{Float64})
    @ HiGHS ~/.julia/packages/HiGHS/6NuM6/src/MOI_wrapper.jl:1470
  [4] add_constraint(b::MathOptInterface.Bridges.LazyBridgeOptimizer{…}, f::MathOptInterface.VariableIndex, s::MathOptInterface.LessThan{…})
    @ MathOptInterface.Bridges ~/.julia/packages/MathOptInterface/jGuEH/src/Bridges/bridge_optimizer.jl:1907
  [5] #normalize_and_add_constraint#51
    @ ~/.julia/packages/MathOptInterface/jGuEH/src/Utilities/constraints.jl:25 [inlined]
  [6] normalize_and_add_constraint
    @ ~/.julia/packages/MathOptInterface/jGuEH/src/Utilities/constraints.jl:19 [inlined]
  [7] bridge_constraint(::Type{…}, model::MathOptInterface.Bridges.LazyBridgeOptimizer{…}, f::MathOptInterface.VectorOfVariables, set::MathOptInterface.Nonpositives)
[...]

PR #3974, opened 2025-04-06
Catch and rethrow more informative error for BoundAlreadySet in optimize!

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/3973

### DIFF
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -598,20 +598,68 @@ function optimize!(
     try
         MOI.optimize!(backend(model))
     catch err
-        # TODO: This error also be thrown also in MOI.set() if the solver is
-        # attached. Currently we catch only the more common case. More generally
-        # JuMP is missing a translation layer from MOI errors to JuMP errors.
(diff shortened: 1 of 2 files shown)


src/optimizer_interface.jl
```diff
@@ -598,22 +598,70 @@ function optimize!(
...
 
+# TODO: These errors could also be thrown in MOI.add_constraint or MOI.set() if
+# the solver is attached. Currently we catch only the more common case. More
+# generally JuMP is missing a translation layer from MOI errors to JuMP errors.
+_rethrow_moi_error(err) = rethrow(err)
+
+function _rethrow_moi_error(::MOI.UnsupportedAttribute{MOI.NLPBlock})
+    return error(
+        "The solver does not support nonlinear problems " *
+        "(that is, NLobjective and NLconstraint).",
```
blegat, line 615: So it does support it if you use `@constraint` ? The message is unclear

odow: NLPBlock in the context of JuMP comes from the `@NL` macros. This was the old error message. I didn't change it. It's also pretty irrelevant now so I don't see much need to change.

PR #3987, opened 2025-04-30
[docs] add conventions page to the developer docs

odow:
Closes #3980

PR #3988, opened 2025-04-30
Fix reshaping ::Nothing for matrix shapes

odow:
Found writing #3987


blegat:
Can't we do `reshape_vector(::Nothing, ::AbstractShape) = nothing` ? If this edge-case was missed, it is surely missed also by extensions that defines shapes (e.g. PolyJuMP) so defining for `AbstractShape` would be nice

PR #3995, opened 2025-05-13
Do not set variable name when unsupported by solver in `_moi_add_constrained_variables`

### LINKED ISSUE #3994: Unexpected behavior when adding constrained variable to a solver that does not support name (direct model)
Hey,

I had a problem using Ipopt with direct model & parameters (MRE `nok1`).
I feel like the bug comes from JuMP because the behavior is inconsistent with "regular" variable addition (see `ok3`).

```julia

using JuMP, Ipopt, MathOptInterface

const MOI = MathOptInterface

function nok1() # Current: error ***** Expected : no error because ok3 works
    model = JuMP.direct_model(Ipopt.Optimizer())
    @variable(model, x in MOI.Parameter(0.0))
    return
end

function nok2() # Current: error ***** Expected : no error because ok3 works
    model = JuMP.direct_model(Ipopt.Optimizer())
    @variable(model, x in MOI.LessThan(1.0))
    return
end


function ok3()
    model = JuMP.direct_model(Ipopt.Optimizer())
    @variable(model, x <= 1)
    return
end
```

`nok1` & `nok2` throw the following error :

```
  MathOptInterface.UnsupportedAttribute{MathOptInterface.VariableName}: Attribute MathOptInterface.VariableName() is not supported by the model.
  Stacktrace:
    [1] throw_set_error_fallback(model::IpoptMathOptInterfaceExt.Optimizer, attr::MathOptInterface.VariableName, index::MathOptInterface.VariableIndex, value::String; error_if_supported::MathOptInterface.SetAttributeNotAllowed{MathOptInterface.VariableName})
      @ MathOptInterface ~/.julia/packages/MathOptInterface/98sse/src/attributes.jl:603
    [2] throw_set_error_fallback(model::IpoptMathOptInterfaceExt.Optimizer, attr::MathOptInterface.VariableName, index::MathOptInterface.VariableIndex, value::String)
      @ MathOptInterface ~/.julia/packages/MathOptInterface/98sse/src/attributes.jl:593
    [3] set(::IpoptMathOptInterfaceExt.Optimizer, ::MathOptInterface.VariableName, ::MathOptInterface.VariableIndex, ::String)
      @ MathOptInterface ~/.julia/packages/MathOptInterface/98sse/src/attributes.jl:553
    [4] _moi_add_constrained_variable(moi_backend::IpoptMathOptInterfaceExt.Optimizer, scalar_variable::JuMP.ScalarVariable{Float64, Float64, Float64, Float64}, set::MathOptInterface.Parameter{Float64}, name::String, ::Type{Float64})
      @ JuMP ~/.julia/packages/JuMP/RGIK3/src/variables.jl:2344
    [5] add_variable(model::JuMP.Model, variable::JuMP.VariableConstrainedOnCreation{MathOptInterface.Parameter{Float64}, JuMP.ScalarVariable{Float64, Float64, Float64, Float64}}, name::String)
      @ JuMP ~/.julia/packages/JuMP/RGIK3/src/variables.jl:2314
```

Note that if I use `JuMP.set_string_names_on_creation(model, false)`, the three functions work.

guimarqu:
Closes #3994

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -2346,7 +2346,8 @@ function _moi_add_constrained_variable(
 ) where {T}
     var_index, con_index = MOI.add_constrained_variable(moi_backend, set)
     _moi_constrain_variable(moi_backend, var_index, scalar_variable.info, T)
-    if !isempty(name) && MOI.supports(moi_backend, MOI.VariableName(), MOI.VariableIndex)
+    if !isempty(name) &&
+       MOI.supports(moi_backend, MOI.VariableName(), MOI.VariableIndex)
         MOI.set(moi_backend, MOI.VariableName(), var_index, name)
     end
     return var_index, con_index
@@ -2435,7 +2436,11 @@ function _moi_add_constrained_variables(
     end
     if names !== nothing
         for (var_index, name) in zip(var_indices, names)
-            if !isempty(name) && MOI.supports(backend(model), MOI.VariableName(), MOI.VariableIndex)
+            if !isempty(name) && MOI.supports(
+                backend(model),
+                MOI.VariableName(),
+                MOI.VariableIndex,
+            )
                 MOI.set(moi_backend, MOI.VariableName(), var_index, name)
             end
         end
@@ -2456,7 +2461,11 @@ function _moi_add_constrained_variables(
     end
     if names !== nothing
         for (var_index, name) in zip(var_indices, names)
-            if !isempty(name) && MOI.supports(backend(model), MOI.VariableName(), MOI.VariableIndex)
+            if !isempty(name) && MOI.supports(
+                backend(model),
+                MOI.VariableName(),
+                MOI.VariableIndex,
+            )
                 MOI.set(moi_backend, MOI.VariableName(), var_index, name)
             end
         end
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -2347,7 +2347,7 @@ function _moi_add_constrained_variable(
     var_index, con_index = MOI.add_constrained_variable(moi_backend, set)
     _moi_constrain_variable(moi_backend, var_index, scalar_variable.info, T)
     supports = MOI.supports(moi_backend, MOI.VariableName(), MOI.VariableIndex)
-    if !isempty(name) && supports       
+    if !isempty(name) && supports
         MOI.set(moi_backend, MOI.VariableName(), var_index, name)
     end
     return var_index, con_index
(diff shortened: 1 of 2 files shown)


blegat:
Thanks, it looks like the right fix
Actually, here `set_name` should be changed to calling `MOI.set(moi_backend, MOI.VariableName(), index, name)` to avoid the type instability of `backend(model)`
https://github.com/jump-dev/JuMP.jl/blob/f659f18bd3cd294514f91c58462d25d5fe5d2398/src/variables.jl#L2224-L2228
Can you add tests ?

guimarqu:
Change + format done

where should I put the tests ? In `TestVariable` ?

test/test_variable.jl
```diff
@@ -1778,4 +1778,28 @@ function test_variable_in_set_HermitianPSDCone()
...
+    inner = MOIU.MockOptimizer(MOIU.Model{Float64}(); supports_names = false)
+    model = direct_model(innner)
+    @variable(model, x in Parameter(0.0))
+    @test is_parameter(x)
+    return
+end
+
+function test_direct_model_variable_in_set_SOS1()
+    inner = MOIU.MockOptimizer(MOIU.Model{Float64}(); supports_names = false)
+    model = direct_model(mockoptimizer)
```
odow, line 1791:
```suggestion
    model = direct_model(inner)
```

test/test_variable.jl
```diff
@@ -1778,4 +1778,28 @@ function test_variable_in_set_HermitianPSDCone()
     return
 end
 
+function test_direct_model_variable_in_set_Parameter()
+    inner = MOIU.MockOptimizer(MOIU.Model{Float64}(); supports_names = false)
+    model = direct_model(innner)
```
odow, line 1783:
```suggestion
    model = direct_model(inner)
```

src/variables.jl
```diff
@@ -2346,7 +2346,8 @@ function _moi_add_constrained_variable(
 ) where {T}
     var_index, con_index = MOI.add_constrained_variable(moi_backend, set)
     _moi_constrain_variable(moi_backend, var_index, scalar_variable.info, T)
-    if !isempty(name)
+    supports = MOI.supports(moi_backend, MOI.VariableName(), MOI.VariableIndex)
+    if !isempty(name) && supports
```
odow, line 2350:
```suggestion
    if !isempty(name) && supports
```

PR #4000, opened 2025-05-16
Add value(::Function, ::X) and value(::X; result::Int) for Symmetric and Hermitian

### LINKED ISSUE #3998: `value` doesn't work for complex matrices
If I declare a real symmetric or real PSD variable, `value` works nicely for it, preserving the wrapper. But with complex Hermitian or complex PSD variables, it gives me an error. It would be great if it could work as well.

Example
```julia
using JuMP
using Hypatia

model = Model()
@variable(model, ρ[1:2,1:2] in HermitianPSDCone())
@variable(model, σ[1:2, 1:2] in PSDCone())
optimize!(model)
value(σ)
value(ρ)
```

araujoms:
Closes #3998 

I also took the liberty of generalizing `value(::Symmetric)` to work with `AbstractJuMPScalar`, as it was failing for `AffExpr` for no reason.

### DIFF
--- a/src/sd.jl
+++ b/src/sd.jl
@@ -232,7 +232,7 @@ function reshape_vector(
     matrix = Matrix{T}(undef, shape.side_dimension, shape.side_dimension)
     k = 0
     for j in 1:shape.side_dimension
-        for i in 1:j-1
+        for i in 1:(j-1)
             k += 1
             matrix[j, i] = matrix[i, j] = 0.5 * v[k]
         end
@@ -306,7 +306,7 @@ end
 
 function vectorize(matrix, ::SkewSymmetricMatrixShape)
     n = LinearAlgebra.checksquare(matrix)
-    return [matrix[i, j] for j in 1:n for i in 1:j-1]
+    return [matrix[i, j] for j in 1:n for i in 1:(j-1)]
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/sd.jl
```diff
@@ -432,7 +432,7 @@ end
 
 function value(
     Q::LinearAlgebra.Symmetric{V,Matrix{V}},
-) where {V<:AbstractVariableRef}
+) where {V<:AbstractJuMPScalar}
```
odow, line 435: This change needs a test

odow:
I made a couple of changes so these match the `value` methods for the various scalars.

odow:
Also: #4000!!!

blegat (approved):

araujoms:
I almost wrote a bullshit issue to get the number 4000 there as well.

What do you think of doing instead of `value(x::AbstractArray) = value.(x)`? Copying only the upper triangular?

odow:
I meant rethinking this error:

https://github.com/jump-dev/JuMP.jl/blob/704235b9474bec96e45d1eda8e89a7bb6359eda4/src/variables.jl#L2701-L2706

araujoms:
Ah yes, I always found that a bit annoying. The basic JuMP objects are vectors and matrices, so it makes sense to define `value` for them directly. More complicated objects can be left for the user to deal with.

blegat:
Yes we could, but maybe in a separate PR because we'll still need the special cases for Symmetric and Hermitian since IIRC the broadcasting don't preserve the shape

PR #4001, opened 2025-05-18
Update cutting_stock_column_generation.jl

WalterMadelim:
wait a minute. Let me see if I can find some other inconsistencies.


odow (approved):

WalterMadelim:
I think the struct defined is needless, which make the exposition lengthy.
And an additional question is, why `i in 1:5` have the same width `75.0`? Can't they be classified into one term?
```julia
Data for the cutting stock problem:
  W = 100.0
with pieces:
   i   w_i d_i
  ------------
   1  75.0  38
   2  75.0  44
   3  75.0  30
   4  75.0  41
   5  75.0  36
   6  53.8  33
```

odow:
The purpose of the various tutorials is also to show how you can use Julia structs instead of very basic Vector and Int. I'd like to keep that as-is.

PR #4009, opened 2025-05-21
use new syntax in state discrimination tutorial

araujoms:
I'll also update the MATLAB tutorial after `value` gets extended to vectors and matrices.


blegat (approved):
Much cleaner, thanks

PR #4021, opened 2025-06-25
Fix link to source code for latest release

blegat:
Actually, this correct, we just forgot we had post-release item : https://jump.dev/JuMP.jl/stable/developers/checklists/#Making-a-release I though confused the release-1.8 (which is outdate) branch with the release-1.0 branch.


odow:
Done

PR #4032, opened 2025-07-22
Cache subexpressions when building MOI.ScalarNonlinearExpression

blegat:
Here is a simplified reproducer of the performance issue identified in https://github.com/jump-dev/MathOptInterface.jl/issues/3015:
```julia
using JuMP

f(x, u) = [sin(x[1]) - x[1] * u, cos(x[2]) + x[1] * u]
function RK4(f, X, u)
    k1 = f(X     , u)
    k2 = f(X+k1/2, u)
    k3 = f(X+k2/2, u)
    k4 = f(X+k3  , u)
    X + (k1 + 2k2 + 2k3 + k4) / 6
end

model = Model()

@variable(model, q[1:2])
@variable(model, u)

x = q
for m = 1:4
    x = RK4(f, x, u)
end
@time moi_function.(x)
```
Before this PR, this gives
```
14.610110 seconds (207.45 M allocations: 6.927 GiB, 68.63% gc time)
```
After this PR, this gives
```
0.000103 seconds (2.16 k allocations: 74.250 KiB)
```

Note that the `MOI.ScalarNonlinearFunction` we generate now share common sub-expression by pointers! If I'm not mistaken, we don't support modifying these in-place so that shouldn't be an issue.

This fixes the slow model-building issue but ReverseAD will still be terribly slow. One thing we could do is to detect, in `MOI.Nonlinear`, the `MOI.ScalarNonlinearFunction` that share sub-functions correponding to the same object (with a dictionary). When it detects two functions with the same pointer, it can then create subexpressions and use its existing support for subexpressions.
The nice thing about it is that we're not doing any change to the MOI interface. Creating `MOI.ScalarNonlinearFunction` was already possible from the beginning, we just didn't treat it any differently. So this change will just be a performance optimization of the AD but the interface is as simple and non-breaking as it gets.

## Results on other benchmarks

For this benchmark, cherry-pick both https://github.com/jump-dev/MathOptInterface.jl/pull/2803 and https://github.com/jump-dev/MathOptInterface.jl/pull/3008. We don't know yet what's the right fix at the MOI level, these PRs just make sure we only see the JuMP-side in the timing.
Then, execute `test/perf/benchmark_cache.jl`.
The conclusion of the benchmark seems to be that a global dict is better than a new dict per constraint and also that we have a small win to hash the `ObjectId`.
The benchmark was generated by Claude as a neutral judge.

#### Current
```
BenchmarkTools.Trial: 149 samples with 1 evaluation per sample.
 Range (min … max):  10.813 ms … 176.359 ms  ┊ GC (min … max):  0.00% … 86.85%
 Time  (median):     26.873 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   33.761 ms ±  30.991 ms  ┊ GC (mean ± σ):  22.81% ± 20.14%

  ▁▂  ▆▆▂█▃                                                     
  ██▇▅█████▆▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆▅▁▁▁▆▁▅▅ ▅
  10.8 ms       Histogram: log(frequency) by time       168 ms <

 Memory estimate: 24.01 MiB, allocs estimate: 721053.
```
#### One dict per constraint
```
BenchmarkTools.Trial: 1850 samples with 1 evaluation per sample.
 Range (min … max):  2.579 ms …  4.681 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.689 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.704 ms ± 86.148 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

          ▂▂▄▄▄▆█▄▅▂▆▂▃▂▁▃▂▁                                  
  ▂▂▂▃▃▅▆███████████████████▇█▆▆▅▅▇▄▅▅▄▄▄▄▃▃▃▃▃▃▃▂▃▃▃▃▂▂▁▂▂▃ ▄
  2.58 ms        Histogram: frequency by time        2.92 ms <

 Memory estimate: 17.45 KiB, allocs estimate: 303.
```
#### Global dict
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):   9.567 μs … 150.842 ms  ┊ GC (min … max):  0.00% … 99.95%
 Time  (median):     12.009 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   28.061 μs ±   1.508 ms  ┊ GC (mean ± σ):  53.73% ±  1.00%

      ▃▆█▇▄▂                                                    
  ▁▂▃▇███████▆▅▆▆▆▇▇▆▆▅▅▄▃▂▂▂▃▃▃▂▃▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  9.57 μs         Histogram: frequency by time         22.6 μs <

 Memory estimate: 17.41 KiB, allocs estimate: 325.
```
#### Global dict + object id
```
BenchmarkTools.Trial: 10000 samples with 3 evaluations per sample.
 Range (min … max):   9.182 μs … 211.604 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     13.418 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   14.339 μs ±   5.920 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▂▅▆▇██▇▅▅▃▂▂▁                                            
  ▂▂▃▄▆███████████████▆▆▅▄▄▄▄▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▁▂▂▂ ▄
  9.18 μs         Histogram: frequency by time         30.1 μs <

 Memory estimate: 16.58 KiB, allocs estimate: 297.
```

#### Current
```
BenchmarkTools.Trial: 271 samples with 1 evaluation per sample.
 Range (min … max):   9.258 ms … 313.421 ms  ┊ GC (min … max):  0.00% … 95.58%
 Time  (median):     14.209 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   18.470 ms ±  34.692 ms  ┊ GC (mean ± σ):  25.40% ± 12.79%

  ▆█▁                                                           
  ███▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▁▁▄ ▅
  9.26 ms       Histogram: log(frequency) by time       251 ms <

 Memory estimate: 14.49 MiB, allocs estimate: 380142.
```
#### One dict per constraint
```
BenchmarkTools.Trial: 221 samples with 1 evaluation per sample.
 Range (min … max):  10.698 ms … 329.769 ms  ┊ GC (min … max):  0.00% … 94.39%
 Time  (median):     16.632 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   22.627 ms ±  41.113 ms  ┊ GC (mean ± σ):  29.84% ± 15.29%

  ▆█                                                            
  ██▇█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▄▁▁▁▄▁▁▄ ▅
  10.7 ms       Histogram: log(frequency) by time       267 ms <

 Memory estimate: 17.23 MiB, allocs estimate: 400142.
```
#### Global dict
```
BenchmarkTools.Trial: 236 samples with 1 evaluation per sample.
 Range (min … max):   7.718 ms … 338.331 ms  ┊ GC (min … max):  0.00% … 93.87%
 Time  (median):     18.529 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   21.190 ms ±  36.381 ms  ┊ GC (mean ± σ):  22.21% ± 12.21%

  ▅▅█                                                           
  ███▇▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▄ ▅
  7.72 ms       Histogram: log(frequency) by time       287 ms <

 Memory estimate: 17.37 MiB, allocs estimate: 418600.
```
#### Global dict + object id
```
BenchmarkTools.Trial: 156 samples with 1 evaluation per sample.
 Range (min … max):   6.855 ms …    2.565 s  ┊ GC (min … max):  0.00% … 99.34%
 Time  (median):     13.982 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   32.069 ms ± 204.190 ms  ┊ GC (mean ± σ):  50.93% ±  7.95%

            ▁▄▁▃█                                               
  ▃▁▁▃▃▅▃▁▁▃█████▆▄▃▁▆▃▃▃▁▃▃▄▃▃▁▁▃▁▃▃▁▁▄▃▃▃▁▁▃▁▁▃▁▁▁▁▁▃▃▁▁▁▁▁▃ ▃
  6.85 ms         Histogram: frequency by time         37.4 ms <

 Memory estimate: 15.81 MiB, allocs estimate: 360177.
```

#### Current
```
BenchmarkTools.Trial: 292 samples with 1 evaluation per sample.
 Range (min … max):   6.926 ms … 522.915 ms  ┊ GC (min … max):  0.00% … 97.27%
 Time  (median):     10.711 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   17.169 ms ±  41.033 ms  ┊ GC (mean ± σ):  19.03% ±  8.04%

  ▃▆▄█▄ ▄▃ ▂                                                    
  ████████▆█▅█▆▄▄▁▄▁▁▄▁▁▁▄▄▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▄▁▄▁▁▁▁▄▄▄▅ ▅
  6.93 ms       Histogram: log(frequency) by time      74.8 ms <

 Memory estimate: 20.51 MiB, allocs estimate: 449903.
```
#### One dict per constraint
```
BenchmarkTools.Trial: 160 samples with 1 evaluation per sample.
 Range (min … max):  18.589 ms … 618.022 ms  ┊ GC (min … max):  0.00% … 94.65%
 Time  (median):     22.911 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   31.275 ms ±  60.962 ms  ┊ GC (mean ± σ):  21.55% ± 10.59%

    █▃ ▄       ▅▂▅▇           ▄ ▃                               
  ▅▃████▆▇▁▆▅▅▇████▇▃▁▅▁▃▁▁▆▇▇███▇▅▁▁▁▁▁▁▁▁▁▁▁▃▁▃▁▁▆▆▆▃▃█▇▃▁▃▆ ▃
  18.6 ms         Histogram: frequency by time         34.7 ms <

 Memory estimate: 27.11 MiB, allocs estimate: 453303.
```
#### Global dict
```
BenchmarkTools.Trial: 6570 samples with 1 evaluation per sample.
 Range (min … max):  248.308 μs … 524.374 ms  ┊ GC (min … max):  0.00% … 99.79%
 Time  (median):     554.571 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   760.800 μs ±   8.279 ms  ┊ GC (mean ± σ):  18.82% ±  1.74%

           █▂   ▂▂▁                                              
  ▄▇▅▅▃▄▄▆███▇▄▄███▅▃▃▄▄▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  248 μs           Histogram: frequency by time         1.98 ms <

 Memory estimate: 707.58 KiB, allocs estimate: 13364.
```
#### Global dict + object id
```
BenchmarkTools.Trial: 3525 samples with 1 evaluation per sample.
 Range (min … max):  223.772 μs …   2.514 s  ┊ GC (min … max):  0.00% … 99.83%
 Time  (median):     736.434 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):     1.879 ms ± 42.358 ms  ┊ GC (mean ± σ):  37.89% ±  1.68%

  ▃█▃▁▅▇▄▄▃▃▃▃▃▃▃▃▂▂▂▂▂▂▂▁▁▁▁ ▁                                ▁
  ███████████████████████████████▇▇▇▇▇▇▇▅▁▅▅▅▆▃▅▅▆▅▃▁▅▄▁▄▁▃▁▃▆ █
  224 μs        Histogram: log(frequency) by time      5.68 ms <

 Memory estimate: 662.53 KiB, allocs estimate: 11724.
```

#### Current
```
BenchmarkTools.Trial: 91 samples with 1 evaluation per sample.
 Range (min … max):  14.584 ms …    1.458 s  ┊ GC (min … max):  0.00% … 98.41%
 Time  (median):     53.483 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):   69.033 ms ± 149.943 ms  ┊ GC (mean ± σ):  22.83% ± 10.32%

       ▁    █                                                   
  ▃▄▄▆▆█▇▆███▇▄▃▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃ ▁
  14.6 ms         Histogram: frequency by time          285 ms <

 Memory estimate: 19.29 MiB, allocs estimate: 574306.
```
#### One dict per constraint
```
BenchmarkTools.Trial: 1481 samples with 1 evaluation per sample.
 Range (min … max):  3.041 ms …   4.058 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.369 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.377 ms ± 196.428 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                   ▂▁▃▅▄▃█▄                                    
  ▁▁▂▄▆▅▆▅██▆▃▁▂▃▅██████████▃▁▁▁▁▁▁▂▁▂▂▂▂▂▁▂▁▂▁▁▁▁▁▂▁▁▁▂▃▂▃▂▂ ▃
  3.04 ms         Histogram: frequency by time           4 ms <

 Memory estimate: 1.19 MiB, allocs estimate: 26906.
```
#### Global dict
```
BenchmarkTools.Trial: 3090 samples with 1 evaluation per sample.
 Range (min … max):  954.119 μs … 986.423 ms  ┊ GC (min … max):  0.00% … 99.38%
 Time  (median):       1.246 ms               ┊ GC (median):     0.00%
 Time  (mean ± σ):     1.618 ms ±  17.724 ms  ┊ GC (mean ± σ):  19.61% ±  1.79%

    ▁█▅ ▁                                                        
  ▂▅██████▅▃▄▄▄▄▄▄▄▄▃▂▁▁▁▁▄██▅▆███▇▆▅▄▃▂▁▁▂▂▂▁▁▁▂▁▂▂▁▁▁▁▁▁▁▁▁▁▁ ▃
  954 μs           Histogram: frequency by time         2.08 ms <

 Memory estimate: 1.24 MiB, allocs estimate: 28180.
```
#### Global dict + object id
```
BenchmarkTools.Trial: 1992 samples with 1 evaluation per sample.
 Range (min … max):  845.061 μs …   2.848 s  ┊ GC (min … max):  0.00% … 99.98%
 Time  (median):     998.244 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):     2.510 ms ± 63.787 ms  ┊ GC (mean ± σ):  56.95% ±  2.24%

   ▃▁  █▅▃▂▁▁     ▁                                             
  ▅██▆▇██████▆▄▃▃▇█▅▃▃▃▂▂▃▃▄▄▄▃▃▄▅▄▃▂▂▂▂▂▂▃▃▃▂▂▂▂▂▂▂▁▁▂▂▂▂▂▂▂▂ ▃
  845 μs          Histogram: frequency by time         1.89 ms <

 Memory estimate: 1.15 MiB, allocs estimate: 25129.
```

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -15,6 +15,7 @@ BasedOnStyles = Vale, Google
 
 # TODO(odow): fix all of these
 Google.Ellipses = OFF
+Google.EmDash = OF
 Google.Exclamation = OFF
 Google.FirstPerson = OFF
 Google.OptionalPlurals = OFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -162,6 +162,8 @@ mutable struct GenericModel{T<:Real} <: AbstractModel
     # A dictionary to store timing information from the JuMP macros.
     enable_macro_timing::Bool
     macro_times::Dict{Tuple{LineNumberNode,String},Float64}
+    # We use `Any` as key because we haven't defined `GenericNonlinearExpr` yet
+    subexpressions::Dict{Any,MOI.ScalarNonlinearFunction}
 end
 
 value_type(::Type{GenericModel{T}}) where {T} = T
@@ -276,6 +278,7 @@ function direct_generic_model(
         Dict{Any,MOI.ConstraintIndex}(),
         false,
         Dict{Tuple{LineNumberNode,String},Float64}(),
+        Dict{Any,MOI.ScalarNonlinearFunction}(),
     )
 end
 
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -162,7 +162,7 @@ mutable struct GenericModel{T<:Real} <: AbstractModel
     # A dictionary to store timing information from the JuMP macros.
     enable_macro_timing::Bool
     macro_times::Dict{Tuple{LineNumberNode,String},Float64}
-    # We use `Any` as key because we haven't defined `GenericNonlinearExpr` yet
+    # A cache to track common subexpressions based on their `objectid`.
     subexpressions::Dict{UInt64,MOI.ScalarNonlinearFunction}
 end
 
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -163,7 +163,7 @@ mutable struct GenericModel{T<:Real} <: AbstractModel
     enable_macro_timing::Bool
     macro_times::Dict{Tuple{LineNumberNode,String},Float64}
     # We use `Any` as key because we haven't defined `GenericNonlinearExpr` yet
-    subexpressions::Dict{Any,MOI.ScalarNonlinearFunction}
+    subexpressions::Dict{UInt64,MOI.ScalarNonlinearFunction}
 end
 
 value_type(::Type{GenericModel{T}}) where {T} = T
@@ -278,7 +278,7 @@ function direct_generic_model(
         Dict{Any,MOI.ConstraintIndex}(),
         false,
         Dict{Tuple{LineNumberNode,String},Float64}(),
-        Dict{Any,MOI.ScalarNonlinearFunction}(),
+        Dict{UInt64,MOI.ScalarNonlinearFunction}(),
     )
 end
 
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -769,6 +769,10 @@ function moi_function(constraint::AbstractConstraint)
     return moi_function(jump_function(constraint))
 end
 
+function moi_function(constraint::AbstractConstraint, model)
+    return moi_function(jump_function(constraint), model)
+end
+
 """
     moi_set(constraint::AbstractConstraint)
 
@@ -1028,6 +1032,11 @@ function _moi_add_constraint(
     return MOI.add_constraint(model, f, s)
 end
 
+function moi_function(f, model)
+    check_belongs_to_model(f, model)
+    return moi_function(f)
+end
+
 """
     add_constraint(
         model::GenericModel,
@@ -1044,10 +1053,9 @@ function add_constraint(
     name::String = "",
 )
     con = model_convert(model, con)
+    func, set = moi_function(con, model), moi_set(con)
     # The type of backend(model) is unknown so we directly redirect to another
     # function.
-    check_belongs_to_model(con, model)
-    func, set = moi_function(con), moi_set(con)
     cindex = _moi_add_constraint(
         backend(model),
         func,
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -1036,6 +1036,7 @@ function check_belongs_to_model(f::Vector, model)
     for func in f
         check_belongs_to_model(func, model)
     end
+    return
 end
 
 function moi_function(model, f)
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -1032,6 +1032,12 @@ function _moi_add_constraint(
     return MOI.add_constraint(model, f, s)
 end
 
+function check_belongs_to_model(f::Vector, model)
+    for func in f
+        check_belongs_to_model(func, model)
+    end
+end
+
 function moi_function(f, model)
     check_belongs_to_model(f, model)
     return moi_function(f)
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -769,8 +769,8 @@ function moi_function(constraint::AbstractConstraint)
     return moi_function(jump_function(constraint))
 end
 
-function moi_function(constraint::AbstractConstraint, model)
-    return moi_function(jump_function(constraint), model)
+function moi_function(model, constraint::AbstractConstraint)
+    return moi_function(model, jump_function(constraint))
 end
 
 """
@@ -1038,7 +1038,7 @@ function check_belongs_to_model(f::Vector, model)
     end
 end
 
-function moi_function(f, model)
+function moi_function(model, f)
     check_belongs_to_model(f, model)
     return moi_function(f)
 end
@@ -1059,7 +1059,7 @@ function add_constraint(
     name::String = "",
 )
     con = model_convert(model, con)
-    func, set = moi_function(con, model), moi_set(con)
+    func, set = moi_function(model, con), moi_set(con)
     # The type of backend(model) is unknown so we directly redirect to another
     # function.
     cindex = _moi_add_constraint(
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -1241,6 +1241,7 @@ function moi_function(f::AbstractVector{<:GenericNonlinearExpr})
 end
 
 function MOI.VectorNonlinearFunction(f::Vector{<:AbstractJuMPScalar})
+    model = owner_model(first(f))
     return MOI.VectorNonlinearFunction(moi_function.(f, model))
 end
 
@@ -1287,7 +1288,7 @@ x
 ```
 """
 function simplify(model::GenericModel, f::AbstractJuMPScalar)
-    g = MOI.Nonlinear.SymbolicAD.simplify(moi_function(f))
+    g = MOI.Nonlinear.SymbolicAD.simplify(moi_function(f, model))
     return jump_function(model, g)
 end
 
@@ -1384,7 +1385,7 @@ julia> ∇f[y]
 ```
 """
 function gradient(model::GenericModel{T}, f::AbstractJuMPScalar) where {T}
-    g = moi_function(f)
+    g = moi_function(f, model)
     ∇f = Dict{GenericVariableRef{T},Any}()
     for xi in MOI.Nonlinear.SymbolicAD.variables(g)
         df_dx = MOI.Nonlinear.SymbolicAD.simplify!(
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -588,8 +588,9 @@ function moi_function(
     model::JuMP.GenericModel,
 ) where {V}
     cache = model.subexpressions
-    if haskey(cache, f)
-        return cache[f]
+    key = objectid(f)
+    if haskey(cache, key)
+        return cache[key]
     end
     ret = MOI.ScalarNonlinearFunction(f.head, similar(f.args))
     stack = Tuple{MOI.ScalarNonlinearFunction,Int,GenericNonlinearExpr{V}}[]
@@ -604,8 +605,9 @@ function moi_function(
     end
     while !isempty(stack)
         parent, i, arg = pop!(stack)
-        if haskey(cache, arg)
-            parent.args[i] = cache[arg]
+        arg_key = objectid(arg)
+        if haskey(cache, arg_key)
+            parent.args[i] = cache[arg_key]
             continue
         end
         child = MOI.ScalarNonlinearFunction(arg.head, similar(arg.args))
@@ -617,9 +619,9 @@ function moi_function(
                 child.args[j] = moi_function(arg.args[j])
             end
         end
-        cache[arg] = child
+        cache[arg_key] = child
     end
-    cache[f] = ret
+    cache[key] = ret
     return ret
 end
 
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -597,7 +597,7 @@ function moi_function(
         if f.args[i] isa GenericNonlinearExpr{V}
             push!(stack, (ret, i, f.args[i]))
         elseif f.args[i] isa AbstractJuMPScalar
-            ret.args[i] = moi_function(model, f.args[i])
+            ret.args[i] = moi_function(f.args[i], model)
         else
             ret.args[i] = moi_function(f.args[i])
         end
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -584,6 +584,10 @@ end
 moi_function(x::Number) = x
 
 function moi_function(f::GenericNonlinearExpr{V}) where {V}
+    model = owner_model(f)
+    if haskey(model.subexpressions, f)
+        return model.subexpressions[f]
+    end
     ret = MOI.ScalarNonlinearFunction(f.head, similar(f.args))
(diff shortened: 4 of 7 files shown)


odow:
> If I'm not mistaken, we don't support modifying these in-place so that shouldn't be an issue.

In theory we don't, but the _can_ be. This change needs very very careful consideration.

blegat:
Yes, it has nontrivial consequences. It may very well jeopardize our chances of adding a modification API for nonlinear constraints in the future.

odow:
I think we can simplify this to:
```julia
function my_moi_function(f::GenericNonlinearExpr{V}) where {V}
    cache = Dict{UInt64,MOI.ScalarNonlinearFunction}()
    ret = MOI.ScalarNonlinearFunction(f.head, similar(f.args))
    stack = Tuple{MOI.ScalarNonlinearFunction,Int,GenericNonlinearExpr{V}}[]
    for i in length(f.args):-1:1
        if f.args[i] isa GenericNonlinearExpr{V}
            push!(stack, (ret, i, f.args[i]))
        else
            ret.args[i] = moi_function(f.args[i])
        end
    end
    while !isempty(stack)
        parent, i, arg = pop!(stack)
        parent.args[i] = get!(cache, objectid(arg)) do
            child = MOI.ScalarNonlinearFunction(arg.head, similar(arg.args))
            for j in length(arg.args):-1:1
                if arg.args[j] isa GenericNonlinearExpr{V}
                    push!(stack, (child, j, arg.args[j]))
                else
                    child.args[j] = moi_function(arg.args[j])
                end
            end
            return child
        end
    end
    return ret
end
```
It doesn't hold state across multiple calls to `moi_function`, but if there are large nested expressions within a single expression (the common case), then we exploit that.

odow:
The example gives:
```Julia
julia> @time moi_function.(x);
  0.000142 seconds (1.53 k allocations: 106.688 KiB)
```

blegat:
If this was only about `moi_function` then I would just get the model from the first `JuMP.VariableRef` I would see and then use the dictionary inside it. The reason I have to change the API is to merge `moi_function` and `check_belongs_to_model` in view of that benchmark: https://github.com/jump-dev/MathOptInterface.jl/pull/2788#issuecomment-3102384794

odow:
We can fix `check_belongs_to_model` as well. I don't want to add the model-level cache. The cache should be within the function.

blegat:
It would be weird to have the subexpressions only work when they are on the same function. Especially since the AD backend supports them being on different ones. I guess we can still find out they are the same later in post-processing, it could be enough just for the sake of speeding up passing the model to the AD backend without the exponential blowup.
One issue would be that we start creating many small dictionaries if there are a lot of constraints, but that's probably negligible.
What's the reason for not having a model-level cache ?

odow:
> It would be weird to have the subexpressions only work when they are on the same function.

This is an implementation detail. Nothing should be observable at the JuMP level.

> Especially since the AD backend supports them being on different ones.

This is also an implementation detail.

> I guess we can still find out they are the same later in post-processing, it could be enough just for the sake of speeding up passing the model to the AD backend without the exponential blowup.

This is my strongly preferred option. I would like us to provide the full tape to an AD engine, and then it to turn everything into a single global DAG.

> One issue would be that we start creating many small dictionaries if there are a lot of constraints, but that's probably negligible.

Yes, we definitely need benchmarks before merging anything like this

> What's the reason for not having a model-level cache ?

It turns every nonlinear expression into a long-lived GC object that will never be freed until the model is.

Second, any expression occurring in multiple constraints seems like much less of an issue than the original example, where nested expressions are programmatically created.

Third, if we start with the internal cache, we can always change to a model-level cache later if needed.

blegat:
That makes sense, if we only take care of not duplicating aliased subexpression at the level of each function, it's indeed easier.
We then detect subexpressions in the AD backend using hashes so the aliases used by the user have no impact on the subexpressions used by the AD.
It means that the user has no control over subexpressions but on the other hand, relying on the sub-expression used by the user was probably not the ideal way to let the user control it as well.

blegat:
To avoid having the JuMP objects continue living, we can us their hash as keys instead of the objects themselves

blegat:
https://github.com/jump-dev/JuMP.jl/issues/3729#issuecomment-2060549713 is a good benchmark but we need more.
What about a big sum of `sin(x[i])`

blegat:
I added benchmarks on the top comment.

odow:
Merging. As discussed on the developer call: since this doesn't change any behaviour, we have the right to revert it if anything comes up in future.

PR #4033, opened 2025-07-22
Add vector version for setting constraint attributes

### LINKED ISSUE #4030: Vectorized set_lower_bound and set_upper_bound
See discussion in https://discourse.julialang.org/t/jump-vector-version-of-set-lower-bound-and-set-upper-bound/130703

Monthly call decides that this is something we should at least consider, and that there might already exist various parts for it in MOI and HiGHS.jl.

blegat:
The user could also write the `my_set_lower_bound` of https://github.com/jump-dev/JuMP.jl/issues/4030#issuecomment-3094854001
However, calling this version is a bit easier to write but also a bit safer since it still calls `check_belongs_to_model` and takes care of the `is_model_dirty` flag.
One caveat is that users might then try doing it with JuMP containers so we may need at some point to also add support for `AbstractArray` that then collects and redirect to this method.
We could also check that the `Vector` have the same length here.

Closes https://github.com/jump-dev/JuMP.jl/issues/4030

### DIFF
--- a/src/optimizer_interface.jl
+++ b/src/optimizer_interface.jl
@@ -1295,6 +1295,20 @@ function MOI.set(
     return
 end
 
+function MOI.set(
+    model::GenericModel,
+    attr::MOI.AbstractConstraintAttribute,
+    crs::Vector{<:ConstraintRef},
+    value::Vector,
+)
+    for con_ref in crs
+        check_belongs_to_model(con_ref, model)
+    end
+    model.is_model_dirty = true
+    MOI.set(backend(model), attr, index.(crs), value)
+    return
+end
+
 """
     get_attribute(model::GenericModel, attr::MOI.AbstractModelAttribute)
     get_attribute(x::GenericVariableRef, attr::MOI.AbstractVariableAttribute)


odow:
I'm not convinced this is needed. It would still require users to understand the relationship between JuMP and MOI. The would need to know that `set_lower_bound.(x, v)` is slow, and that they really needed to update `MOI.ConstraintSet` for `LowerBoundRef`. So all this does it help then by calling `backend(model)` and `index`?

We already have too many methods in JuMP and MOI. I'm hesitant to add yet more.

odow:
More concretely:
```julia
using JuMP, HiGHS
N = 1_000
model = direct_model(HiGHS.Optimizer())
@variable(model, x[1:N] >= 0)
bounds = rand(N)
# Current option
MOI.set(
    backend(model),
    MOI.ConstraintSet(),
    index.(LowerBoundRef.(x)),
    MOI.GreaterThan.(v),
)
# This PR:
MOI.set(
    model,
    MOI.ConstraintSet(),
    LowerBoundRef.(x),
    MOI.GreaterThan.(v),
)
```
I don't really see that as a win.

I think I'd prefer we had fewer special cases and a more documented way to lower from JuMP to MOI.

blegat:
Those two options are not equivalent since the first one does not check `check_belongs_to_model` nor set the dirty flag. This is the reason we cannot really advise the user to use the MOI level. Here, we're just adding a JuMP equivalent to an existing MOI method in order to allow the user have these checks.

odow:
`check_belongs_to` isn't really needed if the user is calling `MOI`. We assume that they know what they're doing.

Similarly with the `dirty` flag. The point of it was to catch differences in solvers. It someone reaches for `backend(` then all bets are off the table.

The current option also avoids an extra vector allocation since `index.(LowerBoundRef.(` will be fused into a single broadcast.

odow:
The more I think about this the less I like it.

blegat:
Actually, one issue is that if we start on this road, we'll also probably need to support containers as wellw extracting the vectorization of the underlying constraints

odow:
Agreed. Close this then? People can use MOI if they really need to, and that means reshaping into a vector if they need.

PR #4047, opened 2025-08-08
Add simplify, derivative, and gradient from MOI.Nonlinear.SymbolicAD

odow:


odow:
I'll merge for now. We're a while from another release

Issue #4053, opened 2025-08-14
Issues using DenseAxisArrays with typed indexing


quachpas:
Hi! I was looking over the [typed indices tutorial](https://jump.dev/JuMP.jl/stable/tutorials/linear/typed_indices/#Typed-indices) and thought I'd be able to get a `UnitRange{T}` axis working, because apparently I desperately want to be able to write `x[Factory(1):Factory(10)]`...

I think I stumbled on a few bugs in `DenseAxisArray` in the process when the axis is either `OneTo` or `UnitRange`, because these are treated as special cases in `DenseAxisArray.jl`.

As said by @odow, part of the problem is subtyping `Integer`, i.e., `Factory <: Integer`. I use another struct `Market`, which will always go to `AxisLookup{Dict}` and that absolutely works, but I would love to get the `<: Integer` case working.

I define them as follows with the minimal implementation to make `UnitRange{T}` work with `collect` and `getindex`.
```jl
struct Factory <: Integer
    x::Int
end
Factory(f::Factory) = f
Base.:<(a::Factory, b::Factory) = a.x < b.x
Base.:+(a::Factory, b::Factory) = Factory(a.x + b.x)
Base.:-(a::Factory, b::Factory) = Factory(a.x - b.x)
Base.:(==)(a::Factory, b::Factory) = a.x == b.x
Base.hash(a::Factory, h::UInt) = hash((Factory, a.x), h) # hash(1) != hash(Factory(1))
Base.:(==)(a::Factory, b::Integer) = false # isequal(1, Factory(1)) == false
Base.:(==)(a::Integer, b::Factory) = false
Base.promote_rule(T::Type{<:Integer}, ::Type{Factory}) = T
Base.Int(x::Factory) = x.x
```

Then we can write `Factory(1):Factory(3)` and `Base.OneTo(Factory(3))` as dimensions.

Without subtyping `Integer`, more methods need to be defined, but it remains similar
```jl
struct Market
    x::Int
end
Base.:<(a::Market, b::Market) = a.x < b.x
Base.isless(a::Market, b::Market) = isless(a.x, b.x)
Base.:+(a::Market, b::Market) = Market(a.x + b.x)
Base.:-(a::Market, b::Market) = Market(a.x - b.x)
Base.:÷(a::Market, b::Market) = Market(a.x ÷ b.x)
Base.one(::Market) = Market(1)
Base.:*(a::Integer, b::Market) = Market(a * b.x)
[...]

PR #4056, opened 2025-08-16
[Containers] fix getindex when a custom subtype of <:Integer is used

odow:
Closes https://github.com/jump-dev/JuMP.jl/issues/4053

I'm going to merge this with a view that this is a bug fix, but I reserve the right to revert it in a subsequent release if it turns out to be breaking.

It's pretty niche though, so I don't see it being a problem in the real world. It requires:

 1. Someone to have defined a new type `Foo <: Integer`
 2. To have defined `==` between `Foo` and `T<:Integer` to be `false`
 3. To have created a `DenseAxisArray` with `Foo` as an axis element
 4. Try to index into the `DenseAxisArray` with a `T` and expect things to work.

### DIFF
--- a/src/Containers/DenseAxisArray.jl
+++ b/src/Containers/DenseAxisArray.jl
@@ -114,6 +114,9 @@ function Base.getindex(
     x::_AxisLookup{Tuple{T,T}},
     key::Integer,
 ) where {T<:Integer}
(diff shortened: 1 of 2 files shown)


odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/17015191344

PR #4057, opened 2025-08-22
Add missing promote_operation when missing number types

blegat:
Follow up from https://github.com/jump-dev/JuMP.jl/pull/3730.
As https://github.com/jump-dev/JuMP.jl/pull/3730 was shown to improve performance issues in https://github.com/jump-dev/JuMP.jl/issues/3729, we should aim for complete coverage.
Complete coverage also does not seem to require so much work, we shouldn't miss much after this PR.


odow (approved):

odow:
One perk of 100% code coverage is that it's not muuuuch easier to notice methods like this that can't be reached anymore.

odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/17199239619

blegat:
Only SumOfSquares and PolyJuMP are failing so looks good to merge

PR #4059, opened 2025-08-25
[docs] document closure capture bug in threading

odow:
x-ref https://github.com/jump-dev/JuMP.jl/issues/2348#issuecomment-3218467117


docs/src/tutorials/algorithms/parallelism.md
```diff
@@ -124,6 +124,92 @@ each iteration of the for-loop. In addition, you must avoid race conditions in
 the rest of your Julia code, for example, by using a lock when pushing elements
 to a shared vector.
 
+### Thread safety and the closure capture bug
+
+!!! danger
+    This section is very important to understand. It is not specific to JuMP and
+    it applies to all multithreaded Julia programs.
+
+There is an upstream design issue in Julia ([julia#15276](https://github.com/julialang/julia/issues/15276))
```
odow, line 133:
```suggestion
There is an upstream design issue in Julia ([julia#14948](https://github.com/JuliaLang/julia/issues/14948))
```

PR #4062, opened 2025-08-27
Add cuOpt to list of solvers

blegat:
## Basic

       documented.
       use the solver with JuMP

## Documentation

## Optional

### DIFF
--- a/docs/packages.toml
+++ b/docs/packages.toml
@@ -46,6 +46,8 @@
     rev = "v1.1.1"
 [CSDP]
     rev = "v1.1.2"
+[cuOpt]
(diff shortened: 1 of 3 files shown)


docs/packages.toml
```diff
@@ -46,6 +46,8 @@
     rev = "v1.1.1"
 [CSDP]
     rev = "v1.1.2"
+[cuOpt]
+    rev = "v0.1.0"
```
odow, line 50:
```suggestion
    rev = "51d58b848a1743ba943fd4f887a2cf8db93e21b0"
```

odow (approved):

PR #4065, opened 2025-09-08
prevent adding methods to the functions `>` and `>=`

nsajko:
As documented, the intended way to implement `>` is to add a method to `<`. Similarly with `>=`. A package should never add a method to either `>` or `>=`.

### DIFF
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -3403,7 +3403,7 @@ There are three common mistakes that lead to this.
     )
 end
 
-for sym in (:(<=), :(>=), :(<), :(>))
+for sym in (:(<=), :(<))
     err = _logic_error_exception(sym)
     @eval begin
         Base.$(sym)(::GenericVariableRef, ::Number) = throw($err)


odow (approved):
Assuming the fallbacks work

docs/src/manual/nonlinear.md
```diff
@@ -476,7 +476,7 @@ julia> f(x::Vector{VariableRef}) = x[1] > 1 ? 0 : x[2]
 f (generic function with 1 method)
 
 julia> f(x)
-ERROR: Cannot evaluate `>` between a variable and a number.
+ERROR: Cannot evaluate `<` between a variable and a number.
```
odow, line 479: This is the only downside. We might talk about `<` when the user wrote `>`.

odow:
Aside from recommended practice, what is the motivation for this?

nsajko:
> Aside from recommended practice, what is the motivation for this?

This PR, together with PR JuliaDiff/ForwardDiff.jl#771, greatly decreases the amount of sysimage invalidation on `using JuMP`:

* Status before both PRs:

    * unique invalidation count: 2570

    * data: [without_open_PRs.json](https://github.com/user-attachments/files/22201587/without_open_PRs.json)

* Status after both PRs:

    * unique invalidation count: 1468

    * data: [with_open_PRs.json](https://github.com/user-attachments/files/22201598/with_open_PRs.json)

EDIT: the above is with nightly (v1.13) Julia, https://github.com/JuliaLang/julia/commit/865b8be7891c277e8be0c86135a2b7cbeec22592

nsajko:
Status with this PR, but without PR JuliaDiff/ForwardDiff.jl#771:

* unique invalidation count: 2262

* data: [with_open_JuMP_PR.json](https://github.com/user-attachments/files/22201655/with_open_JuMP_PR.json)

So an improvement, but not as much as with PR JuliaDiff/ForwardDiff.jl#771.

nsajko:
NB: if you want to reproduce the above results, do `using JuMP` without loading anything (other than SnoopCompileCore) before. Not even the REPL or Pkg (both load StyledStrings which does piracy so causes invalidation, changing the results).

odow:
Invalidation is a good reason

PR #4078, opened 2025-10-03
Nicer real and imag in LaTeX

blegat:
Open to suggestion, Wikipedia also suggest `Re` and `Im`.


odow:
What's wrong with current? Re and Im are okay, I guess.

blegat:
In LaTeX, italic *real* means that you are multiplying the variables *r*, *e*, *a* and *l* together which isn't the case here ^^
We can go for Re

odow:
> In LaTeX, italic real means that you are multiplying the variables r, e, a and l together which isn't the case here 

Well, I mean, that's one possible interpretation. But not if we have variable names with more than one character.

odow:
I'll take over. This doesn't account for stuff like `@variable(model, y[1:2] in ComplexPlane())`

PR #4087, opened 2025-10-28
[docs] add UnoSolver.jl to the documentation

odow:
## Basic

       documented.
       use the solver with JuMP

## Documentation

## Optional


docs/packages.toml
```diff
@@ -229,3 +229,7 @@
 [Tulip]
     user = "ds4dm"
     rev = "v0.9.8"
+[UnoSolver]
+    user = "cvanaret"
+    rev = "0.1.0"
```
odow:
```suggestion
    rev = "v2.3.0"
```

PR #4100, opened 2025-12-28
Call `_string_round` for numbers in nonlinear expressions

klamike:
I noticed numbers aren't rounded when printing nonlinear expressions, unlike aff/quad. Is fixing this breaking?

Before this PR:
```julia
julia> m = Model();

julia> @variable m x;

julia> @objective m Min x^3
x ^ 3

julia> objective_function_string(MIME("text/plain"), m)
"x ^ 3.0"    # <-- ugly 3.0
```

After this PR:
```julia
julia> m = Model();

julia> @variable m x;

julia> @objective m Min x^3
x ^ 3

julia> objective_function_string(MIME("text/plain"), m)
"x ^ 3"      # <-- nice 3
```

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -21,7 +21,7 @@ julia> op_ifelse(true, 1.0, 2.0)
 1.0
 
 julia> op_ifelse(x, 1.0, 2.0)
-ifelse(x, 1.0, 2.0)
+ifelse(x, 1, 2)
 
 julia> op_ifelse(true, x, 2.0)
 x


src/nlp_expr.jl
```diff
@@ -185,6 +185,9 @@ op_string(::MIME"text/latex", ::GenericNonlinearExpr, ::Val{:<=}) = "\\le"
 op_string(::MIME"text/latex", ::GenericNonlinearExpr, ::Val{:>=}) = "\\ge"
 op_string(::MIME"text/latex", ::GenericNonlinearExpr, ::Val{:(==)}) = "="
 
+_string_round_if_number(mode, x::Union{Float32,Float64,Complex}) = _string_round(mode, x)
```
odow:
```suggestion
function _string_round_if_number(mode::MIME, x::Union{Float32,Float64})
    return _string_round(mode, x)
end

```

klamike:
Ah, missed the doctests. I'll adjust those in a bit unless you want to do it yourself @odow

PR #4101, opened 2025-12-31
Parse `ℯ^x` as `exp(x)` instead of `2.718281828459045^x`

klamike:
First, happy new year! :champagne: Wishing you all the best in 2026 :smile:

___
This PR adds a special case for `^` where the LHS is `ℯ` (`\euler`) and the RHS is an `AbstractJuMPScalar`. It avoids a conversion from `Irrational{:ℯ}` to `Float64`. https://github.com/jump-dev/MathOptInterface.jl/issues/2776 is sort of related.

Before this PR:

```julia
julia> m = GenericModel{Float32}();

julia> @variable m x;

julia> expr = ℯ^x
2.718281828459045 ^ x

julia> expr.head, expr.args[1], typeof(expr.args[1])
(:^, 2.718281828459045, Float64)

julia> typeof(ℯ)
Irrational{:ℯ}
```

After this PR:

```julia
julia> m = GenericModel{Float32}();

julia> @variable m x;

julia> expr = ℯ^x
exp(x)

julia> expr.head, expr.args[1]
(:exp, x)
```

### DIFF
--- a/src/nlp_expr.jl
+++ b/src/nlp_expr.jl
@@ -398,6 +398,7 @@ for f in (:+, :-, :*, :^, :/, :atan, :min, :max)
         end
     end
 end
+
 function Base.:^(::Irrational{:ℯ}, y::AbstractJuMPScalar)
     # without this, ℯ^y becomes 2.718281828459045^y instead of exp(y)
     return exp(y)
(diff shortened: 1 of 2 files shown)


klamike:
> Are there any other methods like this that might cause similar troubles?

Not that I've run into

PR #4107, opened 2026-01-16
[docs] add a tutorial overview page

odow:


odow:
I need to fix `import package as alias`, but this is looking okay: https://jump.dev/JuMP.jl/previews/PR4107/tutorials/overview/

Nonlinear examples? look at https://jump.dev/JuMP.jl/previews/PR4107/tutorials/overview/#tutorial_Ipopt

DataFrame examples? Look at https://jump.dev/JuMP.jl/previews/PR4107/tutorials/overview/#tutorial_DataFrames

odow:
Then here, https://github.com/jump-dev/MultiObjectiveAlgorithms.jl/tree/master?tab=readme-ov-file#use-with-jump, we can say "See all the JuMP tutorials that use MultiObjectiveAlgorithms.jl", and ditto for any package

odow:
I think this is easier than coming up with a custom tag system for the tutorials. We'd have endless discussions about what tags to use.

blegat:
It's an elegant solution, with everything automated and the ability to refer to the JuMP documentation from a package. I like it.

PR #4110, opened 2026-01-19
Please using `objective_bound` in deriving the lower bound in Benders

WalterMadelim:
There are some other issues ([discourse](https://discourse.julialang.org/t/sometimes-no-objective-bound-info-from-an-lp-solve-by-gurobi/135080/9?u=waltermadelim)) related. But the core idea is briefly the title.

### DIFF
--- a/docs/src/tutorials/algorithms/benders_decomposition.jl
+++ b/docs/src/tutorials/algorithms/benders_decomposition.jl
@@ -250,7 +250,7 @@ println("Iteration  Lower Bound  Upper Bound          Gap")
 for k in 1:MAXIMUM_ITERATIONS
     optimize!(model)
     assert_is_solved_and_feasible(model)
-    lower_bound = objective_value(model)
+    lower_bound = objective_bound(model)
     x_k = value.(x)
     ret = solve_subproblem(x_k)
     upper_bound = (objective_value(model) - value(θ)) + ret.obj


odow:
This is not needed. We already check `assert_is_solved_and_feasible` on the previous line.

If there is a difference between `objective_value` and `objective_bound`, then either: a) there is a bug in the solver, or b) you have an allowable objective gap and you are solving a MIP.

## The MIP gap case

If you have a large MIP gap, say, 5%, then this `lower_bound` can exceed the true lower bound by 5% (ish; it depends how you define the gap, etc).

But that is _fine_. You've explicitly set that. The `ABSOLUTE_OPTIMALITY_GAP` is about how close we are approximating the cost-to-go function. That's the gap we care about and want to converge.

Put another way: if you have a MIP gap of 5% in the solver, but you use `objective_bound` to compute the `lower_bound`, you may _never_ converge the `gap` to zero because you are measuring the wrong thing.

## Solvers that don't define `objective_bound`

Assuming the first stage is convex and continuous, you most certainly can use Ipopt to solve the first-stage problem. If the first-stage is non-convex, then you can still use Ipopt, but it will likely converge to a local optimum, which it would without Benders anyway.

## Sub-optimally solving the first-stage

True, you don't need to solve the first-stage to optimality. In that case it _would_ be appropriate to use `objective_bound`. But we don't need that here because we check `assert_is_solved_and_feasible`.

WalterMadelim:
So you can keep the current status, if you're only pursuing an "approximate" solution.

My idea was that we need a valid upper and a valid lower bound, nothing else.

In literatures, researchers should make these details very explicit. Otherwise he can allege that "my algorithm can close the gap" in the paper. But when I look into it, I may find that his formula for calculating "gap" is not valid in the mathematical sense.

mlubin:
> We already check assert_is_solved_and_feasible on the previous line.

Should we pass `allow_local=False` to `assert_is_solved_and_feasible`? I don't think it's intended that we'd accept `LOCALLY_SOLVED` as a termination status, and if it prevents someone who copy-pasted the code from doing the wrong thing it will have served a purpose.

odow:
What MIP solver returns LOCALLY_SOLVED?

Also, if we relax the integrality for something like two-stage stochastic, then `LOCALLY_SOLVED` is a valid status (e.g., if you use Ipopt).

mlubin:
> What MIP solver returns LOCALLY_SOLVED?

None, it's just defensive coding. If we were explicitly checking the statuses instead of `assert_is_solved_and_feasible`, we wouldn't accept `LOCALLY_SOLVED`, so we shouldn't let the default `allow_local=True` guide us in the wrong direction. Someone using Ipopt for the benders master needs to understand what they're doing, and getting an error when copy-pasting the tutorial and switching the solver for Ipopt would nudge them to do so.

mlubin:
@WalterMadelim I believe our use of `objective_value` is reasonably consistent after #4111. The solver is claiming that it solved the problem to global optimality, so `objective_value` is the global optimal objective value, up to tolerances. Default tolerances are sufficiently tight in most cases. If you're relaxing tolerances or solving the master to a suboptimal solution, I agree you need to be more careful; however that's beyond the scope of an introductory tutorial.

WalterMadelim:
@mlubin Okay, it's of introductory level. In practice Benders Decomposition is mostly used in e.g. 2SSP with numerous scenarios, where the situation is quite different---there're a bunch of subproblems.

I saw you had a paper https://doi.org/10.1016/j.orl.2013.02.003 talking about a "regularized" cutting plane method. (I mentioned this in [discourse](https://discourse.julialang.org/t/sometimes-no-objective-bound-info-from-an-lp-solve-by-gurobi/135080/10?u=waltermadelim)) yesterday (the 3rd bullet point there in the link). I'm unsure if you have any thoughts or advanced experience that can share with us.

A recent update for practitioners like me is that Gurobi13 now supports PDHG (aka PDLP) to solve larger-scale LPs (on CPU _or_ GPU). Since you're one of the co-authors of the PDLP paper, and you had a julia package about this. Could you share any thoughts, e.g. how's it compared to other methods---primal/dual simplex, barrier (aka interior point) method?

I'd appreciate it. Thanks!

mlubin:
With the caveat that Github comments on a closed issue aren't the best venue for tangential discussions:

- To be honest I haven't thought about regularized cutting plane methods since that paper more than a decade ago and don't think I'd be more useful than your favorite LLM.
- I'm a fan of Ed Rothberg's talk which provided a very balanced take on PDLP: https://www.youtube.com/watch?v=B9rKByy2ym4. You've got it in Gurobi now, so the best way to see how it compares is to try it out!

PR #4118, opened 2026-02-16
Add SDPLRPlus

blegat:
cc @luotuoqingshan

## Basic

       documented.
       use the solver with JuMP

## Documentation

## Optional

### DIFF
--- a/docs/packages.toml
+++ b/docs/packages.toml
@@ -230,8 +230,8 @@
     has_html = true
     extension = true
 [SDPLRPlus]
-    user = "blegat"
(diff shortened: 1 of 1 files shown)


odow:
Update the README to follow the usual template, and remove their personal email etc. The README still mentions that it isn't registered, and there is no license or affiliation section.

odow:
And how was the Julia code written? Based on the paper of SDPLR? Or the source code? Because if the code, then the license should probably follow upstream instead of MIT.

blegat:
It follows the algorithm from the paper but not the source code itself. For instance SDPLRPlus adds some nice abstraction for the constraint matrices which make the code quite different.

PR #4119, opened 2026-02-25
Refactor to use ParallelTestRunner in the tests

odow:


odow:
We'd need to drop pre 1.10 support. But that could be okay. We've done it with MOI and many others.

odow:
If this looks good I'll split https://github.com/jump-dev/JuMP.jl/pull/4119/changes/baaf401e7f0c588b32b0dde7578bac2c0dc2cfb1 into a separate PR.

odow:
I'll do #4121 separately

odow:
Nice. There's a solid 2x speedup here.

PR #4120, opened 2026-02-25
Add dependabot

odow:


odow:
Doc failure is https://github.com/jump-dev/JuMP.jl/pull/4122

PR #4123, opened 2026-02-25
[docs] Update LinearOperatorCollection compat

nHackel:
This PR reverts the workaround #4122 needed to fix [the LinearOperatorCollection issue with the LinearOperators v2.13 struct changes](https://github.com/JuliaImageRecon/LinearOperatorCollection.jl/issues/21) and updates the collection to the next working version


odow:
Thanks for the fix! If you're interested, here's the tutorial where we use it: https://jump.dev/JuMP.jl/stable/tutorials/conic/ellipse_fitting/

nHackel:
Seeing all the packages that use both LinearOperatorCollection and RegularizedLeastSquares directly or for their docs was an interesting side-effect of the unfortunate breakage 😅 It's a really cool and illustrative example

PR #4130, opened 2026-03-08
Add CoolPDLP to list of solvers

blegat:
cc @gdalle @klamike 

## Basic

       documented.
       use the solver with JuMP

## Documentation

## Optional

### DIFF
--- a/docs/styles/config/vocabularies/JuMP/accept.txt
+++ b/docs/styles/config/vocabularies/JuMP/accept.txt
@@ -13,6 +13,7 @@ broadcasted
 camelCase
 [Cc]odecov
 compat
+CPU(?s)
 dataframe(?s)
 datastructure
 demuxer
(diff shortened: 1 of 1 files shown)


docs/packages.toml
```diff
@@ -142,6 +142,9 @@
     user = "oxfordcontrol"
     rev = "v0.11.0"
     has_html = true
+[CoolPDLP]
+    user = "JuliaDecisionFocusedLearning"
+    rev = "v0.1.0"
```
gdalle: Is it important here to keep the version up to date? What does this file do?

klamike (approved):
LGTM thanks @blegat!

klamike:
Vale is complaining about spelling mistakes on `Nationale`, `Agence`, `Dalle`, `Guillaume`, `CPUs`. These all seem okay to me, is there a way to tell Vale to ignore those?

gdalle:
I'm low-key offended

PR #4139, opened 2026-03-26
terminology on Benders

matbesancon:
Calling the master problem a subproblem is pretty confusing. First-stage is the terminology mostly in stochastic optimization I think?

And Taylor approximation is only for differentiable functions


odow:
How about https://github.com/jump-dev/JuMP.jl/pull/4141

PR #4150, opened 2026-04-23
Add example of building & solving JuMP models in parallel

WalterMadelim:
Maybe this can be adapted to a more general form. I just share my code as a proposal.

### DIFF
--- a/docs/src/tutorials/algorithms/parallelism.md
+++ b/docs/src/tutorials/algorithms/parallelism.md
@@ -194,6 +194,65 @@ each iteration of the for-loop. In addition, you must avoid race conditions in
 the rest of your Julia code, for example, by using a lock when pushing elements
 to a shared vector.
 
+### An example
+Here is an example that use multiple threads to build and solve MILP models.
+```
+import JuMP, Gurobi
+const CONFIG = Dict{String,Any}("OutputFlag"=>0,"Threads"=>1);
+Env() = Gurobi.Env(CONFIG);
+function Env(N::Int)
+    v = Vector{Gurobi.Env}(undef, N)
+    Threads.@threads for i=eachindex(v)
+        v[i] = Env()
+    end
+    v
+end;
+function Model!(mv::Vector{JuMP.Model}, i, ev::Vector{Gurobi.Env})
+    m = JuMP.direct_model(Gurobi.Optimizer(ev[i]))
+    JuMP.set_string_names_on_creation(m, false)
+    mv[i] = m
+end
+Model!(mv::Vector{JuMP.Model}, ev::Vector{Gurobi.Env}) = Threads.@threads for i=eachindex(mv)
+    Model!(mv, i, ev)
+end
+show_thread_config() = (th = map(Threads.nthreads, (:default, :interactive)); println("my_config> Threads=$th"))
+function build!(m; N = 80)
+    x = JuMP.@variable(m, [1:N], binary=true)
+    y = JuMP.@variable(m, [1:N], binary=true)
+    z = JuMP.@variable(m, [1:N, 1:N], lower_bound=0)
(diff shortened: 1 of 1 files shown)


odow:
Have you seen the examples lower down? https://jump.dev/JuMP.jl/dev/tutorials/algorithms/parallelism/#Example:-parameter-search-with-multi-threading

WalterMadelim:
I looked at it. I surmise that example is weak in terms of practical performance.

It's probably not efficient to parallelize tasks where task1 is building and solving model and task2 is building and solving model. This style is probably easier to code. But I suspect that at runtime, the multithreading execution will degrade highly, e.g. when you observe htop, you may find that for the most of the time there is only a single core running.

A better way should separate JuMP's modeling part and Gurobi's solving part entirely, since the nature of their jobs are different (JuMP allocates a lot, whereas MIP solving is computation heavy).

odow:
How about this:

```Julia
using JuMP
import HiGHS

function build_model(s::Int; N::Int = 80)
    set_silent(model)
    @variable(model, x[1:N], Bin)
    @variable(model, y[1:N], Bin)
    @variable(model, z[1:N, 1:N] >= 0)
    @constraint(model, [i in 1:N, j in 1:N], z[i, j] <= x[i])
    @constraint(model, [i in 1:N, j in 1:N], z[i, j] <= y[j])
    @constraint(model, [i in 1:N, j in 1:N], z[i, j] >= x[i] + y[j] - 1)
    @objective(model, Min, sum(rand(-10:10) * i for i in z))
    set_time_limit_sec(model, s)
    return model
end

struct Solution
    scenario::Int
    objective_value::Float64
end

function solve_model(ch::Channel{Solution}, s::Int, model::Model)
    optimize!(model)
    @assert has_values(model) # There is always the trivial solution
    put!(ch, Solution(s, objective_value(model)))
    return
end

function run_example(S::Int)
    models = Vector{Model}(undef, S)
    Threads.@threads for s in 1:S
        models[s] = build_model(s)
    end
    ch = Channel{Solution}()
    for (s, model) in enumerate(models)
        Threads.@spawn solve_model(ch, s, model)
    end
    for i in 1:S
        solution = take!(ch)
        println("s=$(solution) [solved $i/$S]")
    end
    return
end

run_example(15)
```

odow:
> This style is probably easier to code.

:+100: This shouldn't be understated.

> It's probably not efficient to parallelize tasks where task1 is building and solving model and task2 is building and solving model. .... But I suspect that at runtime, the multithreading execution will degrade highly

Why?

odow:
How's https://github.com/jump-dev/JuMP.jl/pull/4155

WalterMadelim:
I've been using multithreading utilities in julia for nearly a year.

According to my experience (or say, my personal feeling), the following 3-loop paradigm works the best, whether we're building allocation-heavy JuMP Models or solving hard MILPs in parallel. 
```julia
import JuMP
function build_s(ch, s::Int, m::JuMP.Model)
    # suppose a heavy JuMP modeling happens here, to `m`
    put!(ch, s)
end
function build(inn::Vector{JuMP.Model})
    W = Threads.nthreads()
    S = length(inn)
    @assert S >= W "we don't consider test cases with small S"
    ch = Channel{Int}()
    for s = 1:W
        Threads.@spawn(build_s(ch, s, inn[s]))
    end
    i = W
    while true
        i === S && break
        take!(ch)
        s = i += 1
        Threads.@spawn(build_s(ch, s, inn[s]))
    end
    for _ = 1:W
        take!(ch)
    end
end
```

* As for `Threads.@threads`, I think it works perfectly when `S <= W`, but I suspect there's some performance issue when `S >> W`.
* As for spawning all `S` tasks at once with `@spawn`, I'm not sure if it overwhelms the hardware (e.g. Memory bandwidth) so that performance is degraded.

PR #4158, opened 2026-05-04
[docs] add an example for defining callbacks in a package extension

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/issues/2964

### DIFF
--- a/docs/src/tutorials/linear/callbacks.jl
+++ b/docs/src/tutorials/linear/callbacks.jl
@@ -241,7 +241,7 @@ function example_solver_dependent_callback_ipopt()
     )
         push!(x_vals, callback_value(model, x))
         @test isapprox(obj_value, 1.0 * x_vals[end] + 0.5, atol = 1e-1)
-        # return `true` to keep going, or `false` to terminate the optimization.
+        ## return `true` to keep going, or `false` to terminate the optimization.
         return iter_count < 1
     end
     set_attribute(model, Ipopt.CallbackFunction(), my_callback)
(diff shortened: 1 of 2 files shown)


docs/src/tutorials/linear/callbacks.jl
```diff
@@ -204,4 +216,122 @@ function example_solver_dependent_callback()
...
+        mu::Float64,
+        d_norm::Float64,
+        regularization_size::Float64,
+        alpha_du::Float64,
+        alpha_pr::Float64,
+        ls_trials::Cint,
+    )
+        push!(x_vals, callback_value(model, x))
+        @test isapprox(obj_value, 1.0 * x_vals[end] + 0.5, atol = 1e-1)
+        # return `true` to keep going, or `false` to terminate the optimization.
```
odow:
```suggestion
        ## return `true` to keep going, or `false` to terminate the optimization.
```

docs/src/tutorials/linear/callbacks.jl
```diff
@@ -204,4 +216,122 @@ function example_solver_dependent_callback()
...
+        inf_du::Float64,
+        mu::Float64,
+        d_norm::Float64,
+        regularization_size::Float64,
+        alpha_du::Float64,
+        alpha_pr::Float64,
+        ls_trials::Cint,
+    )
+        push!(x_vals, callback_value(model, x))
+        @test isapprox(obj_value, 1.0 * x_vals[end] + 0.5, atol = 1e-1)
```
odow:
```suggestion
        Test.@test isapprox(obj_value, 1.0 * x_vals[end] + 0.5, atol = 1e-1)
```

PR #4164, opened 2026-05-20
[docs] change the parallelism tutorial to spawn only a limited set of tasks

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/4150#issuecomment-4498931336

We probably shouldn't be encouraging people to spawn a very large set of tasks. We might also consider pointing users to https://juliafolds2.github.io/OhMyThreads.jl/stable/ instead of manually rolling their own like this.

### DIFF
--- a/docs/src/tutorials/algorithms/parallelism.jl
+++ b/docs/src/tutorials/algorithms/parallelism.jl
@@ -359,11 +359,14 @@ a_correct_way_to_build_with_multithreading()
 
 # ## Example: using Channels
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/algorithms/parallelism.jl
```diff
@@ -359,11 +359,14 @@ a_correct_way_to_build_with_multithreading()
 
 # ## Example: using Channels
```
odow:
```suggestion
# ### Example: using Channels
```

PR #4173, opened 2026-06-19
Fix failing test on Julia nightly

odow:
The issue is that even on old versions, you cannot create a Symmetric matrix of Strings.
```julia
julia> using LinearAlgebra

julia> Symmetric(["a" "b"; "b" "c"])
ERROR: MethodError: no method matching symmetric_type(::Type{String})
```

The failure in nightly is because of a change so that broadcasting `name.(x)` over a symmetric matrix `x` now tries to create a symmetric matrix as the answer. (That seems pretty reasonable.)

Yet another problem caused by Julia's ill-defined interfaces.

### DIFF
--- a/test/test_macros.jl
+++ b/test/test_macros.jl
@@ -1647,7 +1647,7 @@ function test_set_string_name_PSDCone()
     set_string_names_on_creation(model, false)
     x = @variable(model, [1:2, 1:2] in PSDCone())
     for xi in x
-        @test isempty(name(x))
+        @test isempty(name(xi))
     end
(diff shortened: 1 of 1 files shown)


test/test_macros.jl
```diff
@@ -1644,7 +1646,9 @@ function test_set_string_name_PSDCone()
     model = Model()
     set_string_names_on_creation(model, false)
     x = @variable(model, [1:2, 1:2] in PSDCone())
-    @test name.(x) == fill("", 2, 2)
+    for xi in x
+        @test isempty(name(x))
```
odow:
```suggestion
        @test isempty(name(xi))
```

Issue #4175, opened 2026-06-19
Error message improvements


odow:
## Background

@mlubin and I have been talking about improving JuMP for coding agents. One area in which we can do a lot better is to provide more informative error messages, so that when things go wrong agents (and users) have more guidance on what to do next, or where to look in the documentation to understand why they hit an error.

As a recent example, see @DimitriAlston's recent issue #4171 and fix in #4172.

See also this talk: https://www.youtube.com/watch?v=bRDtplyhRYw

I don't know that there's a perfect template that we should apply.

One approach would be to follow commercial solvers and have a short unique identifier with a longer description in the docs about when and why it happens.

Another is to add more content to the error message.

At minimum, I think we need two-part error messages: the first says what (and perhaps why) the error happened. The second should prompt some sort of action item as the next step for how to fix the error.

I would have thought agents liked more verbose, well formatted error messages, but Claude is trying to tell me not. I don't know I believe it. Here's part of our conversation:
```
❯ Are longer, more descriptive error messages helpful for coding agents?

Yes, with some nuance.

Coding agents benefit from context that a human would infer from experience. A human seeing unsupported operator mentally fills in "oh, I probably wrote = instead of ==." An agent has to reason that out from the message alone, so a hint like "did you mean ==?" is more valuable to an agent than a human.

However, the Markdown and verbosity problems still apply — arguably more so. An agent reads the full message into its context window, so padding (circular explanations, redundant examples) wastes tokens and can dilute the signal.

The practical sweet spot for a library like JuMP, which is increasingly used in agentic workflows:

- Keep: the bad value, the exhaustive list of valid alternatives, and the fix hint (operator_to_set for custom sets)
- Add: a direct "did you mean X?" suggestion when the bad value is close to a valid one (e.g., = → ==)
- Drop: prose explanation of what the error is, redundant examples, section headers

So the right response to "should I write longer messages for agents" is: more precise and complete, not more verbose. An agent benefits from a message that removes ambiguity about the fix, not one that has more words.
```

## Action items

So, action items:

 * Do you have examples of packages/libraries/languages with very nice error messages? If so, please share them below.
 * Do other projects have style guides for how to write nice error messages? Should JuMP adopt such a style guide?
 * Crank through the 167 error messages in JuMP/src and improve them...

odow:
```
❯ What are some examples of Python packages with great error messages?

A few standouts:

[...]

PR #4176, opened 2026-06-19
Improve various error messages

odow:
x-ref #4175


src/feasibility_checker.jl
```diff
@@ -63,14 +53,34 @@ function primal_feasibility_report(
...
+                argument to `primal_feasibility_report`.
+                """,
             )
         end
         return fx
     end
 end
 
+function primal_feasibility_report(model::GenericModel{T}; kwargs...) where {T}
+     if !has_values(model)
```
odow:
```suggestion
    if !has_values(model)
```

PR #4179, opened 2026-06-22
Add ALGLIB Solver to installation docs

alglib-project:
This PR adds ALGLIB Solver to the supported solvers table.

ALGLIB.jl is registered in Julia General and provides a JuMP/MathOptInterface wrapper for the external ALGLIB Solver executable.

The checklist is below:

## Basic

       documented.
       use the solver with JuMP

## Documentation

## Optional

 - [NO] Add package metadata to `docs/packages.toml`


odow (approved):
I removed "Free" from the license column. ALGLIB is not a free solver.

PR #4183, opened 2026-06-27
Fix BridgeableConstraint with model_convert

blegat:
Tried fixing https://github.com/jump-dev/SumOfSquares.jl/issues/444 with Claude but he got really confused.
The root cause it that the way `model_convert` works is a bit unclear and inconsistent.
When you call `@constraint(model, ...)`, we first create an `AbstractConstraint` with `build_constraint`. There we don't have the model type so we cannot convert yet.
Then, in the macro code itself we call `model_convert`:
https://github.com/jump-dev/JuMP.jl/blob/44397c71edc7e65d85b55ce062f8b47dad0ea643/src/macros/%40constraint.jl#L188-L206
Then we pass the converted `AbstractConstraint` to `add_constraint` which then also calls `model_convert`:
https://github.com/jump-dev/JuMP.jl/blob/44397c71edc7e65d85b55ce062f8b47dad0ea643/src/constraints.jl#L1046
It a bit redundant. I checked and both were added in github.com/jump-dev/JuMP.jl/pull/3385.
The advantages of doing it in `add_constraint` are that 1) it's easier to maintain method implementations that macro code and 2) we don't need to do a broadcast in case of the vectorization.
In PolyJuMP, we create a constraint outside the macro so we rely on the `model_convert` in `add_constraint`.
https://github.com/jump-dev/PolyJuMP.jl/blob/571a9874547ea44efe38822e073a3aecffdb127c/src/constraint.jl#L282-L288
However, we didn't call `model_convert` there so that's an issue.

So there are a few ways to fix the issue in SumOfSquares. In all of them, we need the fix in `model_convert` of `BridgeableConstraint`. The question is, where should `model_convert` be called.
We should decide who's responsibility it is to call `model_convert`, is it

1) the caller of `add_constraint` or
2) the implementation of `add_constraint`.

In case

1) we should probably remove the `model_convert` in `add_constraint(::AbstractModel, ::AbstractConstraint)` (we may consider it breaking but that would only be for users that weren't using the macro so unclear if it's a big deal) and in case
2) we should add `model_convert` in `add_constraint(::GenericModel, ::BridgeableConstraint)` like done in this PR and then also remove the `model_convert` called from the macro.

### DIFF
--- a/test/test_model.jl
+++ b/test/test_model.jl
@@ -436,29 +436,28 @@ function test_macro_bridgeable()
     model = Model()
     @variable(model, x)
     @constraint(model, x in BridgeMe{Int,Nonnegative}(Nonnegative()))
-    # `model_convert` normalizes the bridge coefficient type to the value type of
-    # the model (`Float64`), even though `BridgeMe` requested `Int`.
+    # `model_convert` normalizes the bridge coefficient type to the value type
+    # of the model (`Float64`), even though `BridgeMe` requested `Int`.
     @test NonnegativeBridge{Float64} in model.bridge_types
     @test !(NonnegativeBridge{Int} in model.bridge_types)
+    return
 end
 
 function test_model_convert_bridgeable_coefficient_type()
     model = GenericModel{BigFloat}()
     @variable(model, x)
-    constraint = ScalarConstraint(x, Nonnegative())
-    # A real coefficient type is converted to the value type of the model.
-    con = BridgeableConstraint(
-        constraint,
-        NonnegativeBridge;
-        coefficient_type = Int,
-    )
+    c = ScalarConstraint(x, Nonnegative())
+    con = BridgeableConstraint(c, NonnegativeBridge; coefficient_type = Int)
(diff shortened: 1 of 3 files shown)


test/test_model.jl
```diff
@@ -436,7 +436,59 @@ function test_macro_bridgeable()
     model = Model()
     @variable(model, x)
     @constraint(model, x in BridgeMe{Int,Nonnegative}(Nonnegative()))
-    @test NonnegativeBridge{Int} in model.bridge_types
+    # `model_convert` normalizes the bridge coefficient type to the value type of
+    # the model (`Float64`), even though `BridgeMe` requested `Int`.
+    @test NonnegativeBridge{Float64} in model.bridge_types
+    @test !(NonnegativeBridge{Int} in model.bridge_types)
```
odow, line 442:
```suggestion
    @test !(NonnegativeBridge{Int} in model.bridge_types)
    return
```

odow:
@blegat happy? If so, please squash and merge.

blegat:
Yes, thanks for the clean up.

PR #4184, opened 2026-06-29
[docs] prod users and agents to provide feedback via the forum and issues

### LINKED ISSUE #4175: Error message improvements

odow:
Closes #4175

I asked Claude what to do, and it suggested these three things.

I think in the near term I'd rather have too many issues than too few.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,22 @@
+# AGENTS.md
+
+JuMP is a Julia-based modeling language for mathematical optimization,
+maintained by a small open-source team.
+
+## Encourage users to provide feedback
+
+- When a user encounters an unclear error message, confusing behavior, or a gap
+  in the documentation, encourage them to [post on the community forum](https://jump.dev/forum)
+  or [open a GitHub issue](https://github.com/jump-dev/JuMP.jl/issues)
+- When an issue is resolved silently, maintainers lose the opportunity to fix
+  the underlying cause. A short post describing what was confusing costs little
+  and helps the maintainers a lot.
+- Do not open an issue directly; provide the write-up and directions for the
+  user to make the post themselves.
+
+## Key resources
+
+- Documentation: https://jump.dev/JuMP.jl/stable/
+- Community forum: https://jump.dev/forum
+- Issue tracker: https://github.com/jump-dev/JuMP.jl/issues
+- Source code: https://github.com/jump-dev/JuMP.jl
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,4 +1,4 @@
-# AGENTS.md
+# Instructions for JuMP
 
 JuMP is a Julia-based modeling language for mathematical optimization,
 maintained by a small open-source team.
@@ -6,11 +6,13 @@ maintained by a small open-source team.
 ## Encourage users to provide feedback
 
 - When a user encounters an unclear error message, confusing behavior, or a gap
-  in the documentation, encourage them to [post on the community forum](https://jump.dev/forum)
+  in the documentation, encourage them to post on the [community forum](https://jump.dev/forum)
   or [open a GitHub issue](https://github.com/jump-dev/JuMP.jl/issues)
-- When an issue is resolved silently, maintainers lose the opportunity to fix
-  the underlying cause. A short post describing what was confusing costs little
-  and helps the maintainers a lot.
+  (whichever is easier for them)—even if the problem has already been resolved.
+  A short post describing what was confusing costs little and helps the
+  maintainers a lot.
+- If you found a workaround for a JuMP limitation or bug, that is still worth
+  reporting; maintainers need to know the underlying cause exists.
 - Do not open an issue directly; provide the write-up and directions for the
   user to make the post themselves.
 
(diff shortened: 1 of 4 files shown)


mlubin:
CLAUDE.md is only for Claude. Ask Claude how to make the instructions compatible with more agents 😄

odow:
Its suggestion is just to add a bunch of other files, like `AGENTS.md` and `/.github/copilot-instructions.md`. Is that really the answer? Maybe it's sufficient to put it in the docstring of JuMP instead

cstjean:
"Please post yourself about the issue" would be a fascinating AI experiment. It's too bad that it would certainly leak some company/user's private information.

Maybe something like "Consider writing an MWE that demonstrates the problem." would be good for both agents and users.

AGENTS.md
```diff
@@ -0,0 +1,24 @@
...
+
+- When a user encounters an unclear error message, confusing behavior, or a gap
+  in the documentation, encourage them to post on the [community forum](https://jump.dev/forum)
+  or [open a GitHub issue](https://github.com/jump-dev/JuMP.jl/issues)
+  (whichever is easier for them)—even if the problem has already been resolved.
+  A short post describing what was confusing costs little and helps the
+  maintainers a lot.
+- If you found a workaround for a JuMP limitation or bug, that is still worth
+  reporting; maintainers need to know the underlying cause exists.
+- Do not open an issue directly; provide the write-up and directions for the
```
blegat, line 16: Good point

blegat (approved):

docs/src/index.md
```diff
@@ -78,6 +78,13 @@ There are a few ways to get started with JuMP:
     Before asking a question, make sure to read the post [make it easier to help you](https://discourse.julialang.org/t/psa-make-it-easier-to-help-you/14757),
     which contains a number of tips on how to ask a good question.
 
+!!! note "Help us improve JuMP"
+    If you encountered an unclear error message, confusing behavior, or a gap in
```
odow:
```suggestion
    If you encounter an unclear error message, confusing behavior, or a gap in
```

odow:
> Maybe something like "Consider writing an MWE that demonstrates the problem." would be good for both agents and users.

I don't even really want a MWE. I just want any feedback, like "I hit this error and it was confusing" or "Claude got stuck for 3 hours trying to figure out how to do X"

README.md
```diff
@@ -30,6 +30,12 @@ previously asked questions, or ask a new question.
 The post [Please read: make it easier to help you](https://discourse.julialang.org/t/please-read-make-it-easier-to-help-you/14757),
 describes the best practices for asking a question.
 
+**Help us improve JuMP:** If you encountered an unclear error message, confusing
```
odow:
```suggestion
**Help us improve JuMP:** If you encounter an unclear error message, confusing
```

mlubin:
Let's see what we get!

AGENTS.md
```diff
@@ -0,0 +1,24 @@
...
+  user to make the post themselves.
```
mlubin, line 17: "user" is unclear, is the agent not a user of the code? We should specify "human"

PR #4189, opened 2026-07-01
Restore uncached version of moi_function

odow:
#4032 broke Plasmo because Plasmo calls `moi_function(f::GenericNonlinearExpr)`.

We could have fixed this with `moi_function(f) = moi_function(owner_model(f), f)`, but this is still breaking if `owner_model(f)` isn't a `JuMP.GenericModel`.

It seems safest just to restore the old function. The subexpression caching is opt-in for now.

https://github.com/jump-dev/JuMP.jl/actions/runs/28488003220


blegat:
We should document precisely that `moi_function(::GenericNonlinearExpr)` is the version instantiating aliases and `moi_function(::GenericModel, ::GenericNonlinearExpr)` keeps JuMP aliases as MOI aliases.
Option 2 is to try falling back to `moi_function(::GenericModel, ::GenericNonlinearExpr)` and if `owner_model` is `nothing` (constant expression ? Must be pretty rare) or it is not `GenericModel` then we don't do the cache, otherwise we do.

Having this very subtle behavior with aliases just be undocumented could be quite confusing. Especially since now agents will like tweaking with JuMP internals directly so they will probably like just calling `moi_function` directly so we should be crystal clear.

blegat:
One thing that worries me with option 1 is that we might forget to call `moi_function` with the model as first argument and this kind of but might be tricky to catch. I removed this method on purpose to easily catch all cases where this was needed. Now that we add this method, we might in the future just call `moi_function` in the JuMP call and forget about `moi_function(::GenericModel, ..)`. In thing Option 2 is probably safer.
We could also add a function that should return the cache and we make it default to return `nothing` for `JuMP.AbstractModel`.

odow:
See also #4190 

Agree that we need to think about this in a lot more detail.

odow:
Merging this since it fixes a breaking change. We can think more about it in #4190

PR #4190, opened 2026-07-01
Refactor moi_function and jump_function into separate file

odow:
x-ref #4189 

We don't necessarily need to merge this. But these functions were spread all over the place. I'll comment in-line on something.


src/callbacks.jl
```diff
@@ -123,11 +123,13 @@ function MOI.submit(
...
 )
-    return MOI.submit(backend(model), cb, moi_function(con.func), con.set)
+    f = moi_function(model, con.func)
+    return MOI.submit(backend(model), cb, f, con.set)
 end
 
 function MOI.submit(model::GenericModel, cb::MOI.UserCut, con::ScalarConstraint)
-    return MOI.submit(backend(model), cb, moi_function(con.func), con.set)
+    f = moi_function(model, con.func)
+    return MOI.submit(backend(model), cb, f, con.set)
```
odow, line 132: These ones were missing from #4032

src/constraints.jl
```diff
@@ -1045,11 +1023,6 @@ function check_belongs_to_model(f::Vector, model)
     return
 end
 
-function moi_function(model, f)
-    check_belongs_to_model(f, model)
-    return moi_function(f)
-end
-
```
odow, line 1052: So this is a very weird method. It's not just a fallback. I think the only reason it was needed was because of `check_belongs_to_model`?

src/constraints.jl
```diff
@@ -1066,6 +1039,7 @@ function add_constraint(
     name::String = "",
 )
     con = model_convert(model, con)
+    check_belongs_to_model(con, model)
```
odow, line 1042: We just do it explicitly here now instead.

src/jump_moi_function.jl
```diff
@@ -0,0 +1,474 @@
...
+
+julia> moi_function(f)
+1.0 + 2.0 MOI.VariableIndex(1)
+```
+"""
+function moi_function end
+
+# A default fallback for backwards compatibility. The first argument `model` was
+# introduced in JuMP@1.31.0.
+moi_function(model, f) = moi_function(f)
```
odow, line 53: And now I have a much more sensible fallback.

src/callbacks.jl
blegat: How did we miss them ? Removing the `moi_function` method was precisely done to catch them

src/constraints.jl
blegat: Yes, since we moved to requiring the user to call `moi_function`, I thought we might as well do `check_belongs_to_model` there since these two are always called one after the other

src/constraints.jl
blegat: If we're going to call `check_belongs_to_model` every time we call `moi_function` (which makes sense), we might as well just call it inside `moi_function` no ? Since `moi_function` takes 2 arguments, it makes sense to check that they model is indeed the correct one, doesn't it ?

src/jump_moi_function.jl
```diff
@@ -0,0 +1,480 @@
...
+        end
+        model.subexpressions[arg_key] = child
+    end
+    model.subexpressions[key] = ret
+    return ret
+end
+
+# A backwards-compatible function to preserve behavior prior to #4032. This was
+# used by Plasmo.jl
+function moi_function(f::GenericNonlinearExpr{V}) where {V}
```
blegat, line 207: I still think that this should just grab model from the expression graph and then redirect to `moi_function(model, f)`. Having such a subtle behavior (aliases kept or not), depend on whether you call two very similar methods.

blegat (changes requested):

src/jump_moi_function.jl
blegat: I think the definition should be: The user should **always** call `moi_function(model, f)`. When you implement, you can either implement `moi_function(model, f)` (and checking that the model match) or just implement `moi_function(f)`.
To me, it makes sense to check that the model match precisely when we transform the JuMP function to MOI since that's when the variable references are going to be turned into indices (so without the model). Implementing it in two steps 1) check membership and then 2) do a `moi_function(f)` is one way to do it but they could also just be done together, which is what we do for `NonlinearExpr`

src/callbacks.jl
odow: We don't have any nonlinear callbacks.

src/jump_moi_function.jl
odow: We can't do that because that's then a breaking change. People can and have been calling `moi_function(f)`.

src/jump_moi_function.jl
odow: There are trivial examples where the model doesn't exist. `NonlinearExpr(:+, Any[0.0])`

odow: Okay, updated.

odow:
Okay @blegat how is this.

src/jump_moi_function.jl
blegat: It's the new way forward but we don't have to enforce it. It's for clarifying in the comments& doc

src/jump_moi_function.jl
```diff
@@ -0,0 +1,484 @@
...
+    model = owner_model(f)
+    if model isa GenericModel
+        return moi_function(model, f)
+    end
```
blegat, line 213: Maybe add `@assert isnothing(model)` and a comment explaining this is for the rare case where the is no variable and the user call thos backward compatible method so even if there may be aliases it's ok

blegat (approved):
Looks good modulo my minor comments to improve readability

PR #4193, opened 2026-07-06
Adds CONOPT to solvers documentation

stephenjmaher:
Adding the CONOPT.jl package to the solvers documentation.

We would like to discuss the possibility of bringing CONOPT.jl under the JuMP-dev organisation, similar to KNITRO.jl. The intention would be for us to continue the maintenance of CONOPT.jl as part of the JuMP community. We would include the line "Contact GAMS Software if you encounter any problem with this interface or the solver." in the documentation, as is the case for KNITRO.jl.

## Basic

       documented.
       use the solver with JuMP

## Documentation

## Optional

### DIFF
--- a/docs/packages.toml
+++ b/docs/packages.toml
@@ -149,7 +149,7 @@
     rev = "v0.1.1"
 [CONOPT]
     user = "GAMS-dev"
-    rev = "93efb3badb311fcbf03806c475ec7075df3ed851"
+    rev = "9dc9a6c04897c56ac0f956899cc3e50643193ea3"
 [COPT]
     user = "COPT-Public"
     rev = "v1.1.33"
--- a/docs/packages.toml
+++ b/docs/packages.toml
@@ -147,6 +147,9 @@
 [CoolPDLP]
     user = "JuliaDecisionFocusedLearning"
     rev = "v0.1.1"
+[CONOPT]
+    user = "GAMS-dev"
+    rev = "93efb3badb311fcbf03806c475ec7075df3ed851"
 [COPT]
     user = "COPT-Public"
     rev = "v1.1.33"
--- a/docs/packages.toml
+++ b/docs/packages.toml
@@ -149,7 +149,7 @@
     rev = "v0.1.1"
 [CONOPT]
     user = "GAMS-dev"
-    rev = "9dc9a6c04897c56ac0f956899cc3e50643193ea3"
+    rev = "v0.1.0"
 [COPT]
     user = "COPT-Public"
     rev = "v1.1.33"
--- a/.vale.ini
+++ b/.vale.ini
@@ -39,6 +39,10 @@ Vale.Spelling = NO
 [docs/src/packages/Clarabel.md]
 Google.Latin = NO
 
+[docs/src/packages/CONOPT.md]
+Google.Latin = NO
+Google.OxfordComma = NO
+
 [docs/src/packages/CoolPDLP.md]
 Google.OptionalPlurals = NO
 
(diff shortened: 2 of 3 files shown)


odow:
See this checklist: https://jump.dev/JuMP.jl/stable/developers/checklists/#Adding-a-new-solver-to-the-documentation

You should also edit the installation table.

> We would like to discuss the possibility of bringing CONOPT.jl under the JuMP-dev organisation

I'm going to discuss this privately with the core contributors. Let's proceed with it in `gams-dev` and we'll get back to you.

> The intention would be for us to continue the maintenance of CONOPT.jl as part of the JuMP community

I think we'd need more than an intention. We'd probably also reserve the right to deprecate the package and archive it if GAMS stopped maintaining it.

stephenjmaher:
> > The intention would be for us to continue the maintenance of CONOPT.jl as part of the JuMP community
> 
> I think we'd need more than an intention. We'd probably also reserve the right to deprecate the package and archive it if GAMS stopped maintaining it.

We would be happy for a maintenance agreement to be put in place. Then we can include conditions about the expected maintenance. Setting out a maintenance agreement might serve well for other new and existing solvers in JuMP-dev as well.

stephenjmaher:
Regarding 
> Check that the solver is a registered Julia package

A PR has been created for this https://github.com/JuliaPackaging/Yggdrasil/pull/14008

You should add the checklist to https://jump.dev/JuMP.jl/stable/packages/solvers/, or at least a link. Its current location is not well connected to the "Adding new solvers" section.

odow:
> We would like to discuss the possibility of bringing CONOPT.jl under the JuMP-dev organisation

Our conclusion is no, and that we'd prefer to de-emphasise the distinction between packages in jump-dev and those outside it. 

I've opened a PR to update the docs: https://github.com/jump-dev/JuMP.jl/pull/4195. We can discuss more there.

From a user-perspective, they just do `] add CONOPT` or `] add KNITRO`. There's no signal to them on install that they are using a "JuMP" package or a "GAMS" package.

We have a policy for this: https://jump.dev/pages/governance/#transferring-repositories-to-jump-dev

The key part is

> the bar for transferring repositories to the JuMP-dev GitHub organization is high, and can be summarized by the following: The maintenance of the package would be simplified if core contributors have Admin permission to the repository.

Commercial wrappers like Gurobi.jl and KNITRO.jl exist in jump-dev as a historical accident; they were originally written by the JuMP community, and we have maintained them (to various levels of effort) since.  We would like more of the commercial solvers to move out of jump-dev and into the maintenance hands of others. One in-progress example is Xpress: https://github.com/fico-xpress/XpressAPI.jl.

odow (changes requested):
This is blocked by https://github.com/GAMS-dev/CONOPT.jl/issues/11. CONOPT is not yet a registered package.

stephenjmaher:
> Our conclusion is no, and that we'd prefer to de-emphasise the distinction between packages in jump-dev and those outside it.

This is reasonable.

The changes in #4195 are very good, since it puts all solvers on the same level. Previously, the JuMP-dev solvers looked more official than the contributed solvers.

stephenjmaher:
CONOPT.jl is now a registered package https://github.com/JuliaRegistries/General/pull/160410

docs/packages.toml
```diff
@@ -147,6 +147,9 @@
 [CoolPDLP]
     user = "JuliaDecisionFocusedLearning"
     rev = "v0.1.1"
+[CONOPT]
+    user = "GAMS-dev"
+    rev = "9dc9a6c04897c56ac0f956899cc3e50643193ea3"
```
odow: Can change this to the tag

odow (approved):

PR #4194, opened 2026-07-07
Fix comments for basis status explanation

WalterMadelim:
I think the exposition still needs to be improved. The current words don't read coherent (see the 2nd place of modification).

### DIFF
--- a/docs/src/tutorials/linear/basis.jl
+++ b/docs/src/tutorials/linear/basis.jl
@@ -64,7 +64,7 @@ value.(x)
 
 get_attribute(x[1], MOI.VariableBasisStatus())
 
-# the result is a [`MOI.BasisStatusCode`](@ref). Query all of the basis statuses
+# the result is a [`MOI.BasisStatusCode`](@ref). Query all of the basis status
 # with the broadcast `get_attribute.(`:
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/linear/basis.jl
```diff
@@ -64,7 +64,7 @@ value.(x)
 
 get_attribute(x[1], MOI.VariableBasisStatus())
 
-# the result is a [`MOI.BasisStatusCode`](@ref). Query all of the basis statuses
+# the result is a [`MOI.BasisStatusCode`](@ref). Query all of the basis status
```
odow, line 67: `statuses` is correct here. We mean the plural of status. We want to query multiple "status"

docs/src/tutorials/linear/basis.jl
```diff
@@ -137,7 +137,7 @@ c_basis = Dict(
     all_constraints(model; include_variable_in_set_constraints = false)
 )
 
-# Thus, the basis is formed by `x`, `y`, and the slack associated with `c3`.
+# Thus, the basis is formed by `x`, `y`, and the slack is associated with `c3`.
```
odow, line 140: `is` is not correct here. The basis is formed by three things: 

1. x
2. y
3. the slack associated with constraint c3

PR #4196, opened 2026-07-09
Maintain order of expression generators when calling flatten!

odow:
Erwin points out in https://yetanothermathprogrammingconsultant.blogspot.com/2026/03/experience-with-nlp-solvers-on-simple.html that we don't maintain the order. (This is more cosmetic than changing actual model behaviour, but it's a nice quality of life improvement for users looking at the expressions.)

The prior result here was `1, 3, 2` instead of `1, 2, 3`. 
```Julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, x[1:3]);

julia> f = sum(log(x[i]) for i in 1:3) |> flatten!
log(x[1]) + log(x[3]) + log(x[2])
```


odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/29481861017

PR #4199, opened 2026-07-20
Fix checking if problem is an LP in lp_sensitivity_report

odow:


src/lp_sensitivity2.jl
```diff
@@ -334,11 +334,15 @@ end
 Return `true` if `model` is a linear program.
 """
 function _is_lp(model::GenericModel)
+    AffType = Union{GenericVariableRef,GenericAffExpr}
+    if !(objective_function_type(model)  <: AffType)
```
odow:
```suggestion
    if !(objective_function_type(model) <: AffType)
```

PR #4204, opened 2026-07-22
Use Base.IdDict instead of objectid to store subexpressions

### LINKED ISSUE #4203: Potential hash collision for nonlinear subexpressions
Hi,

Thanks for your work and release. As my problems of interest have plenty of common subexpressions, I tried v1.31 to check for any speed improvement (for build time so far if I understood correctly).

However, that release broke my code (which worked with v1.30.1).
The bug is silent and JuMP just seems to build a different optimization problem.  

Here is a MWE (with the help of Claude).  

```julia 

using JuMP

n = 25_000 # 5_000 seems to be OK for both versions

model = Model()
@variable(model, x)
cons = [@constraint(model, sin(x) * i <= 0) for i in 1:n]

# The stored function is  -( *(sin(x), i), 0.0 ) ; recover that i.
function multiplier(model, c)
    f = MOI.get(JuMP.backend(model), MOI.ConstraintFunction(), JuMP.index(c))

    got = nothing
    walk(g) =
        if g isa MOI.ScalarNonlinearFunction
            g.head === :* && for a in g.args
                a isa Real && (got = a)
            end
            foreach(walk, g.args)
        end
    walk(f)
    return got
end

bad = [(i, m) for (i, c) in enumerate(cons) for m in (multiplier(model, c),) if m != i]

println("JuMP ", pkgversion(JuMP))
println("wrong constraints: ", length(bad), " / ", n)
for (i, got) in first(bad, 5)
    println("  constraint $i stored as sin(x) * $got")
end
```

Outputs:

JuMP 1.31.0
wrong constraints: 141 / 25000
  constraint 14989 stored as sin(x) * 10611.0
  constraint 15001 stored as sin(x) * 10600.0
  constraint 15029 stored as sin(x) * 10567.0
  constraint 15263 stored as sin(x) * 10292.0
  constraint 15280 stored as sin(x) * 10270.0

JuMP 1.30.1
wrong constraints: 0 / 25000

odow:
Closes #4203

I didn't realise `objectid` could be re-used, but I guess it makes sense. It's not documented very well.

But this is exactly what I didn't want to happen, because we don't want to keep every single subexpression alive for the life of the model. That could be a big increase in memory usage.

One option is to make the subexpression storage opt-in.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -163,7 +163,7 @@ mutable struct GenericModel{T<:Real} <: AbstractModel
     enable_macro_timing::Bool
     macro_times::Dict{Tuple{LineNumberNode,String},Float64}
     # A cache to track common subexpressions based on their `objectid`.
-    subexpressions::Dict{UInt64,MOI.ScalarNonlinearFunction}
+    subexpressions::Base.IdDict{Any,MOI.ScalarNonlinearFunction}
 end
 
 value_type(::Type{GenericModel{T}}) where {T} = T
@@ -278,7 +278,7 @@ function direct_generic_model(
         Dict{Any,MOI.ConstraintIndex}(),
         false,
         Dict{Tuple{LineNumberNode,String},Float64}(),
-        Dict{UInt64,MOI.ScalarNonlinearFunction}(),
+        Base.IdDict{Any,MOI.ScalarNonlinearFunction}(),
     )
 end
 
--- a/src/jump_moi_function.jl
+++ b/src/jump_moi_function.jl
@@ -167,10 +167,9 @@ end
 moi_function_type(::Type{<:GenericNonlinearExpr}) = MOI.ScalarNonlinearFunction
 
 function moi_function(model::GenericModel, f::GenericNonlinearExpr{V}) where {V}
-    # key = objectid(f)
-    # if haskey(model.subexpressions, key)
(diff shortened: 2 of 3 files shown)


odow:
I would have much preferred that we just exploited common subexpressions within a single call to `moi_function`: https://github.com/jump-dev/JuMP.jl/pull/4032#issuecomment-3148733897

Then there's no model-level state that we are tracking.

AayushSabharwal:
To avoid keeping expressions alive, something like a WeakKeyDict is useful. It uses `WeakRef` to store the keys, allowing the GC to collect them, and a finalizer to remove it from the dictionary when they get cleaned up. However, the implementation in Base is pretty slow iirc. SymbolicUtils uses WeakCacheSets.jl for hashconsing. It's a set not a dict, so it isn't directly applicable here, but it illustrates the sort of approach that we found to work well in such scenarios.

odow:
I need to:

 * Benchmark the current approach
 * Benchmark something based on https://docs.julialang.org/en/v1/base/base/#Core.WeakRef
 * Consider making it opt-in via `@expression`
 * Consider a global flag to turn off subexpression caching

odow:
Closing because I don't like this. At minimum, we can do something like #4207. Or maybe WeakRef + `@expression`.

PR #4207, opened 2026-07-24
Use WeakKeyDict for the subexpression dictionary

### LINKED ISSUE #4203: Potential hash collision for nonlinear subexpressions

odow:
A variant of #4204 that closes #4203

The downside to this approach is that we need to decide how frequently to filter the subexpression dictionary. If an element has been GC'd, it cannot be referenced in the future, so it is safe to remove.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -127,6 +127,27 @@ function value_type(::Type{T}) where {T}
     )
 end
 
+mutable struct _WeakCache{V}
+    data::Dict{WeakRef,V}
+    size_at_last_filter::Int
+    _WeakCache{V}() where {V} = new(Dict{WeakRef,V}(), 0)
+end
+
+Base.get(cache::_WeakCache, key, default) = get(cache.data, key, default)
+
+Base.getindex(cache::_WeakCache, key) = getindex(cache.data, WeakRef(key))
+
+Base.length(cache::_WeakCache) = length(cache.data)
+
+function Base.setindex!(cache::_WeakCache{V}, value::V, key) where {V}
+    cache.data[WeakRef(key)] = value
+    if length(cache.data) >= max(5 * cache.size_at_last_filter, 100)
+        filter!(kv -> !isnothing(first(kv).value), cache.data)
+        cache.size_at_last_filter = length(cache.data)
+    end
+    return value
+end
+
 mutable struct GenericModel{T<:Real} <: AbstractModel
     # In MANUAL and AUTOMATIC modes, CachingOptimizer.
     # In DIRECT mode, will hold an AbstractOptimizer.
@@ -163,7 +184,7 @@ mutable struct GenericModel{T<:Real} <: AbstractModel
     enable_macro_timing::Bool
     macro_times::Dict{Tuple{LineNumberNode,String},Float64}
     # A cache to track common subexpressions based on their `objectid`.
-    subexpressions::Dict{UInt64,MOI.ScalarNonlinearFunction}
+    subexpressions::_WeakCache{MOI.ScalarNonlinearFunction}
 end
 
 value_type(::Type{GenericModel{T}}) where {T} = T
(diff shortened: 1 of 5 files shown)


src/JuMP.jl
```diff
@@ -127,6 +127,27 @@ function value_type(::Type{T}) where {T}
     )
 end
 
+mutable struct _WeakCache{V}
+    data::Dict{WeakRef,V}
+    size_at_last_filter::Int
+    _WeakCache{V}() where {V} = new(Dict{WeakRef,V}(), 0)
+end
+
+Base.get(cache::_WeakCache, key, default) = get(cache.data, key, default)
```
blegat:
```suggestion
Base.get(cache::_WeakCache, key, default) = get(cache.data, WeakRef(key), default)
```

blegat: I'm a bit surprised we don't need this and the tests are passing

src/JuMP.jl
```diff
@@ -127,6 +127,27 @@ function value_type(::Type{T}) where {T}
...
+
+Base.getindex(cache::_WeakCache, key) = getindex(cache.data, WeakRef(key))
```
blegat: Where is this used ?

odow: It was used in the tests to simplify things. I guess I can remove.

odow:
I think I'm just re-implementing `Base.WeakKeyDict`. Except that instead of having a finaliser on the key, I have the clean-up pass.

blegat (approved):

src/macros/@force_nonlinear.jl
```diff
@@ -84,10 +84,10 @@ julia> @expression(model, @force_nonlinear(x * 2.0 * (1 + x) * x))
 x * 2 * (1 + x) * x
 
 julia> @allocated @expression(model, x * 2.0 * (1 + x) * x)
-3264
+3200
 
 julia> @allocated @expression(model, @force_nonlinear(x * 2.0 * (1 + x) * x))
-944
+784
```
blegat, line 90: Surprising that this decreases when we change `struct` to `mutable struct`

blegat:
Related https://github.com/JuliaLang/julia/pull/62831

odow:
I guess we should make a decision on this.

blegat:
I feel that we can merge as is. We can add a function later that cleans the dict if needed. Not sure that we should do this cleaning automatically. If it's documented, then it should be fine. Another option would be to clean it before `optimize!`, maybe with an option to disable it. But then it won't work as well if you call `optimize!` from the same scope these objects are created. And we also need to assume GC has been run. So it's better to have a custom function, where we document that the user is recommended to call `GC.gc()` before calling that function

PR #4214, opened 2026-08-09
[docs] improve docstrings for getting various result attributes

### LINKED ISSUE #4212: More discoverability of termination status codes
The docstring of `termination_status` is a bit indirect, and some agents don't easily figure out the names of the termination statuses. We can add a hint to the docstring along the lines of:

>  See `MOI.TerminationStatusCode` for the list of all possible statuses and their meanings.

odow:
Closes #4212


mlubin (approved):
Thanks!

PR #4237, opened 2026-09-03
Fix Base.empty!(::GenericModel) and document its troublesome behavior

odow:
I'll hold off merging this one until we get a few eyes on it. The behaviour of references before and after `empty!` was surprising to me, even though it's a direct consequence of what we're doing.

### DIFF
--- a/src/JuMP.jl
+++ b/src/JuMP.jl
@@ -128,40 +128,59 @@ function value_type(::Type{T}) where {T}
 end
 
 mutable struct GenericModel{T<:Real} <: AbstractModel
+    # !!! note
+    #     When adding new fields to this struct, you must also update
+    #     `Base.empty!(::GenericModel)`.
+
     # In MANUAL and AUTOMATIC modes, CachingOptimizer.
     # In DIRECT mode, will hold an AbstractOptimizer.
     moi_backend::MOI.ModelLike
+
     # List of shapes of constraints that are not `ScalarShape` or `VectorShape`.
     shapes::Dict{MOI.ConstraintIndex,AbstractShape}
+
     # List of bridges to add in addition to the ones added in
     # `MOI.Bridges.full_bridge_optimizer`. With `BridgeableConstraint`, the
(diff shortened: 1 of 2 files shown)


src/JuMP.jl
```diff
@@ -948,6 +967,29 @@ attributes but not optimizer attributes. Always return the argument.
...
+
+julia> @variable(model, y >= 2)
+y
+
+julia> lower_bound(x)
+2.0
+
+julia> x === y
+true
+```
```
odow, line 991: In particular, was anyone aware of this feature

blegat: I always found it annoying that in Julia it's difficult to know if a struct is stack allocated or heap allocated. In this case, because `GenericModel` is heap allocated, my understanding is that the compiler is going to make some decision but it's not always so clear. In earlier versions, `VariableRef` was heap allocated so `===` would return `false` here. Not sure at which version this changed but it sure looks confusing ^^

blegat: I would add a sentence explaining this is because `x` is stack-allocated (or maybe in other words), otherwise, it's a bit difficult to understand.

blegat (approved):
Looks good to me after we give a bit more explanation on why `x === y`. We could add seeds to variables to prevent this but that would worsen performance, we cannot protect the user from everything

src/JuMP.jl
odow: Nah this isn't about x being stack allocated per se. `x == y` would still be `true`, even if it was heap allocated. Maybe I should just do `==` instead.

src/JuMP.jl
```diff
@@ -948,6 +967,29 @@ attributes but not optimizer attributes. Always return the argument.
...
+
+julia> empty!(model);
+
+julia> @variable(model, y >= 2)
+y
+
+julia> lower_bound(x)
+2.0
+
+julia> x === y
```
odow:
```suggestion
julia> x == y
```

PR #4240, opened 2026-09-03
[docs] various fixes found by Claude

odow:
I finds so many little things that I have no idea to find otherwise.


docs/src/tutorials/linear/sports_scheduling.jl
```diff
@@ -88,7 +88,7 @@ model = Model(HiGHS.Optimizer);
 
 @constraint(model, [m in M, t in 1:T], sum(x[m, :, t]) + sum(x[:, m, t]) <= 1);
 
-# Constraint: every team `m` plays at least half home games
+# Constraint: every team `m` plays at approximately half home games
```
odow:
```suggestion
# Constraint: every team `m` plays approximately half home games
```

PR #4247, opened 2026-09-08
[docs] fix use of T_max in rocket control tutorial

odow:


docs/src/tutorials/nonlinear/rocket_control.jl
```diff
@@ -162,7 +162,7 @@ Plots.plot(
 # ## Next steps
 
 # * Experiment with different values for the constants. How does the solution
-#   change? In particular, what happens if you change `T_max`?
+#   change?
```
odow:
```suggestion
#   change? In particular, what happens if you change `T_max`?
```

PR #4254, opened 2026-09-14
Fuse check_belongs_to_model with moi_function

blegat:
As discussed in https://github.com/jump-dev/MathOptInterface.jl/pull/2788#issuecomment-3102384794, `check_belongs_to_model` needs to be merged with `moi_function` to be alias-aware.
See benchmark in https://github.com/jump-dev/MathOptInterface.jl/pull/2788#issuecomment-5664390751


blegat:
> we could save time by walking the tree only once?

The issue is not that we go through it twice, it is that by traversing naively like this, because it is not a tree but a DAG, we end up verifying the whether the same node belongs to the model exponentially many times. So we could create a dictionary again to track what we already check to avoid the exponential blowup but it might be a bit wasteful since we're already doing it in `moi_function` and these functions are always called together anyway.

blegat:
> So the other option is that we make `moi_function(model, f)` throw an error. It's arguable whether it is breaking/not-breaking, because doing it with the wrong model is a bug.

It amounts to saying that now `moi_function(model, f)` is equivalent to `moi_function` + `check_belongs_to_model` ? I think it's a good idea.

src/jump_moi_function.jl
```diff
@@ -48,10 +50,23 @@ julia> moi_function(model, f)
 """
 function moi_function end
 
-# A default fallback for backwards compatibility. The first argument `model` was
+# Default fallbacks for backwards compatibility. The first argument `model` was
 # introduced in JuMP@1.31.0.
 moi_function(model, f) = moi_function(f)
```
blegat: This is hit from `moi_function(::GenericNonlinearExpr, model)` for arguments that are constants. If a user forget to implement this method, we may miss the `check_belongs_to_model` check but we were already not doing the check in the preview `check_belongs_to_model(::GenericNonlinearExpr, model)` for these.
The only risk is for a `func` that is neither a `AbstractJuMPScalar` nor an `AbstractVector{<:AbstractJuMPScalar}` and is used directly as function in a constraint, we now won't call `check_belongs_to_model` on them.

src/jump_moi_function.jl
```diff
@@ -23,14 +23,16 @@ function moi_function_type end
     moi_function(model::GenericModel, x::AbstractJuMPScalar)
     moi_function(model::GenericModel, x::AbstractArray{<:AbstractJuMPScalar})
 
-Given a JuMP object `x`, return the MathOptInterface equivalent.
+Check that `x` belongs to `model`, then return the MathOptInterface equivalent.
+Specialized methods may fuse the ownership check and conversion into a single
+traversal.
```
odow: Isn't it more: if you implement this two-argument version you MUST also check ownership?

src/variables.jl
```diff
@@ -389,6 +389,8 @@ it belongs to a different model.
 """
 function check_belongs_to_model end
 
+check_belongs_to_model(::Any, ::AbstractModel) = nothing
```
odow: What is this method doing here

src/jump_moi_function.jl
```diff
@@ -369,6 +381,13 @@ function moi_function(f::AbstractVector{<:GenericNonlinearExpr})
     return MOI.VectorNonlinearFunction(f)
 end
 
+function moi_function(
+    model::GenericModel,
+    f::AbstractVector{<:GenericNonlinearExpr},
+)
+    return MOI.VectorNonlinearFunction([moi_function(model, row) for row in f],)
```
odow:
```suggestion
    return MOI.VectorNonlinearFunction([moi_function(model, row) for row in f])
```

odow (approved):
LGTM. I've gone and moved the `check_belongs_to_model` functions into `jump_moi_function.jl`, since it makes more sense to keep them close to `moi_function`.

odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/35303818496

odow:
This broke Plasmo

odow:
https://github.com/jump-dev/JuMP.jl/actions/runs/35549917869

blegat:
The SumOfSquares failure is https://github.com/oxfordcontrol/Clarabel.jl/pull/235

blegat:
The fact that we now rely on multiple dispatch and not an hardcoded `if`-`else`, is also helping extension, see https://github.com/blegat/GenOpt.jl/pull/36

PR #4258, opened 2026-09-17
Remove _isfinite

blegat:
They are now unused after https://github.com/jump-dev/JuMP.jl/pull/4252, it's good they are internal methods we can just remove


odow:
Oops yeah I missed this.

PR #4260, opened 2026-09-21
Allow compilation with JuliaC

blegat:
Allow creating a model with a concrete type for the MOI backend. This prevents using `set_optimizer` but on the other hand, this allows compiling a program solve a JuMP model with a given fixed optimizer with JuliaC !


src/JuMP.jl
```diff
@@ -127,14 +127,14 @@ function value_type(::Type{T}) where {T}
     )
 end
 
-mutable struct GenericModel{T<:Real} <: AbstractModel
+mutable struct ModelImpl{T<:Real,B<:MOI.ModelLike} <: AbstractModel
```
blegat: We might want to find a better name than `ModelImpl`

joaquimg (commented):
Can a juliac test run on CI?

blegat:
> Can a juliac test run on CI?

Yes and I think we should add one with this PR. **EDIT** added in `.github/workflows/juliac.yml`

.github/workflows/juliac.yml
```diff
@@ -0,0 +1,42 @@
...
+          Pkg.instantiate()
+      - name: Install JuliaC
+        run: julia --project=@juliac -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaC", version="0.3.10"))'
+      - name: Compile the LP
+        # Unsafe trimming matches the tested JuMPy backend profile. The next
+        # step executes the binary; this is not a safe-trimming guarantee.
+        run: >
+          julia --project=@juliac -m JuliaC
+          --output-exe jump_lp --project test/JuliaC
+          --trim=unsafe-warn --bundle test/JuliaC/build
```
blegat, line 39: Importantly, `--trim=safe` isn't working yet, there are a bunch of things that are still too dynamic for JuliaC to claim it's safe like JuMP's optimize hooks, HiGHS's default interrrupt's callback but unclear how easy it would be to make `safe` work. For this PR, `unsafe_warn` is enough.

